PR #1533, opened 2019-03-18
go: Use a faster ed25519 implementation

Yawning:
This should force everything to use our ed25519 and x25519.


Yawning:
For what it's worth, I'm of the opinion that this is good to go, the moment we get the ed25519 code cleared.

PR #1847, opened 2019-06-25
Add storage export debug subcommand

### LINKED ISSUE #1845: Add storage export debug subcommand
Requires #1758 to not duplicate work.

We need a way to dump a given storage root so we can keep state even while changing the underlying data structure.

kostko:
Fixes #1845.

Blocked on #1835.

TODO
* [ ] Merge with genesis dump tool to export all roots for the dumped roothash block(s).


Yawning:
This can be closed since #2495 has a newer stab at implementing this.

Issue #1793, opened 2019-06-07
Replace epochtime with ticker


ptrus:
Replace epochtime with a `ticker`-time. The general idea is that `ticker` will serve as a (configurable) base frequency. Everything currently relying on `epochtime.WatchEpochs` will be able to be configured as a multiplayer of the base frequency. 

This will also enable: https://github.com/oasislabs/ekiden/issues/1654
And deprecate: https://github.com/oasislabs/ekiden/issues/1762

## Details

- rename `epochtime` to `ticker`
- Extend api to include: `WatchTicks(multiplyer uint64) (<-chan api.Ticker, *pubsub.Subscription)`
- everything currently using `WatchEpochs` should be updated to use the configurable `WatchTciks()` (with configurable multiplyer)

PR #1889, opened 2019-07-02
epochtime refactor

ptrus:
Fixes: #1793 

Done:
- epochtime -> ticker
- enable mock_ticker via flag
- simplify mock_ticker (instead of setting whatever "epoch" only allow advancing ticks now)

TODO:
- Remove all epoch references from ticker 
  - can be a separate package that uses ticker as a timesource & has epoch interval defined
- potentially get rid of the notion of epoch altogether?
  - this would require changing a lot of code, as epoch's are used everywhere
  - instead, maybe we should for now leave: epoch==compute schedule interval
- fix all the flags/etc

Proposed service changes:
- beacon:
  - https://github.com/oasislabs/ekiden/pull/1899
- keymanager:
  - not sure if any changes needed for now?
- registry: 
  - on epoch change: remove expired nodes (= keep expiration per "global epoch")
  - https://github.com/oasislabs/ekiden/pull/1899
- scheduler: (separate PR probably)
  - configurable intervals for committee scheduling
  - shorten transaction leader 
  - increase storage committee interval


ptrus:
Closing this one as it will probably be easier to start this from scratch now with all the changes happening in mean time - also the approach will likely be different.

PR #2147, opened 2019-09-20
messages from roothash mux app

### LINKED ISSUE #2230: Registration panics if we stop the compute worker early
like if we exit really quickly. which, when we do, it closes the compute worker node's initCh without setting a workerHost

## Details

and then the registration callback tries to access the workerHost and segfaults.

## Acceptance Criteria
* that we handle this case

### LINKED ISSUE #2231: broadcastTx panics during shutdown
the when we try to create a subscription, we get a nil (internally, but as I learned, not `== nil`) Subscription if tendermint is shutting down

## Details

and then later we panic on `txSub.Out()`

## Acceptance Criteria
* that we handle this case

pro-wh:
This adds a "roothash messages" field to the compute pipeline. A runtime would be able to send them, and they propagate into a block header.

Roothash messages are stored inline in the block header, rather than in storage.

This implements the simple merging algorithm, where up to one compute committee can send roothash messages.

This adds the staking general balance adjustment message, which causes the staking mux app to adjust an account's general balance.

There's a dummy acceptable transfer peer policy to control what runtimes can send these staking general balance adjustment messages. It accesses a set of approved runtime IDs in the genesis document.

TODO:
- [x] switch dummy acceptable transfer peer policy to deny all
- [x] remove  dummy roothash message types
- [x] test unacceptable transfer peer
- [x] come up with a way to test without hacking up successfulRound

blocks https://github.com/oasislabs/ekiden/issues/2077
context https://github.com/oasislabs/private-rfcs/pull/144

fixes #2230
fixes #2231


pro-wh:
added an e2e test. please review

go/oasis-node/cmd/debug/byzantine/compute.go
```diff
@@ -161,6 +162,8 @@ func (cbc *computeBatchContext) createCommitment(id *identity.Identity, rak sign
 		PreviousHash: cbc.bd.Header.EncodedHash(),
 		IORoot:       cbc.newIORoot,
 		StateRoot:    cbc.newStateRoot,
+		// TODO: allow script to set roothash messages?
```
kostko: File issue for this?

go/oasis-test-runner/scenario/e2e/roothash_messages.go
```diff
@@ -0,0 +1,159 @@
...
+		return errors.Wrap(err, "setting epoch")
+	}
+	logger.Info("epoch transition done")
+
+	select {
+	case err = <-sc.net.Errors():
+		_ = client.Process.Kill()
+		return errors.Wrapf(err, "network")
+	case err = <-clientErrCh:
+		fmt.Println("client terminated")
```
kostko: Remove this (I noticed there is also a leftover print in the other scenario, can you remove that while you're here).

go/staking/api/api.go
```diff
@@ -352,6 +352,10 @@ const (
 	KindStorage   ThresholdKind = 3
 
 	KindMax = KindStorage
+
+	// LogEventGeneralAdjustment is a log event value that signals adjustment
+	// of an account's general balance due to a roothash message.
+	LogEventGeneralAdjustment = "staking/general_adjustment"
```
kostko: I would move this to the const block at the top as this one is really threshold-kind-related.

go/tendermint/apps/staking/state.go
```diff
@@ -35,6 +37,12 @@ var (
 	//
 	// Value is a little endian encoding of an uint64.
 	debondingIntervalKeyFmt = keyformat.New(0x54)
+	// acceptableTransferPeersKeyFmt is the key format used for the acceptable transfer peers set.
+	//
+	// Value is a CBOR-serialized map from acceptable runtime IDs the boolean true.
```
kostko:
```suggestion
	// Value is a CBOR-serialized map from acceptable runtime IDs to the boolean true.
```

go/tendermint/apps/staking/state.go
```diff
@@ -35,6 +37,12 @@ var (
...
+	acceptableTransferPeersKeyFmt = keyformat.New(0x55)
```
kostko: I assume the size of this map is expected to be small (on the order of number of runtimes)? Another option would be to have the entries be stored in the tree -- e.g., one key per entry:
```golang
keyformat.New(0x55, &signature.MapKey{})
```

go/worker/merge/committee/node.go
```diff
@@ -464,9 +464,20 @@ func (n *Node) startMergeLocked(commitments []commitment.ComputeCommitment, resu
 		defer cancel()
 
 		var ioRoots, stateRoots []hash.Hash
+		var roothashMessages []*block.RoothashMessage
 		for _, result := range results {
 			ioRoots = append(ioRoots, result.IORoot)
 			stateRoots = append(stateRoots, result.StateRoot)
+
+			// Merge roothash messages.
+			// The rule is that at most one result can have sent roothash messages.
```
kostko: Added to [list of slashing conditions](https://github.com/oasislabs/oasis-core/issues/2078) for the transaction scheduler.

runtime/src/transaction/context.rs
```diff
@@ -72,4 +76,8 @@ impl<'a> Context<'a> {
             .expect("tags is not empty")
             .push(Tag::new(key.as_ref().to_vec(), value.as_ref().to_vec()))
     }
+
+    pub fn send_roothash_message(&mut self, roothash_message: RoothashMessage) {
```
kostko: Add documentation comment.

go/oasis-node/cmd/debug/byzantine/compute.go
pro-wh: ok #2269

go/oasis-test-runner/scenario/e2e/roothash_messages.go
pro-wh: if we don't want to know, then ok

go/staking/api/api.go

go/tendermint/apps/staking/state.go
pro-wh: thanks

go/worker/merge/committee/node.go
pro-wh: thanks

runtime/src/transaction/context.rs
pro-wh: ok

go/tendermint/apps/staking/state.go
pro-wh: Right, that's the expectation. Thanks for posting another option. Let's keep that in mind so we can later support a large number of acceptable runtimes if we want to.

pro-wh:
`cargo audit` report shows now new problems compared to master. current problems, common with master:

- memoffset 0.2.1 https://rustsec.org/advisories/RUSTSEC-2019-0011
- protobuf 1.7.4 https://rustsec.org/advisories/RUSTSEC-2019-0003
- protobuf 2.4.2 https://rustsec.org/advisories/RUSTSEC-2019-0003
- smallvec 0.6.9 https://rustsec.org/advisories/RUSTSEC-2019-0009, https://rustsec.org/advisories/RUSTSEC-2019-0012
- spin 0.5.0 https://rustsec.org/advisories/RUSTSEC-2019-0013
- rust-crypto 0.2.36 https://rustsec.org/advisories/RUSTSEC-2016-0005

PR #2229, opened 2019-10-04
experimental e2e coverage hacks

pro-wh:
add a test that runs the node, so that we can get coverage in integration tests

somehow codecov doesn't merge the coverage files correctly, so add a step to merge them ourselves

### DIFF
--- a/.buildkite/go/build.sh
+++ b/.buildkite/go/build.sh
@@ -15,5 +15,5 @@ set -euxo pipefail
 # Build the Go parts
 ####################
 pushd go
-  make
+  make all oasis-node/integrationrunner/integrationrunner.test
 popd
--- a/.buildkite/go/build.sh
+++ b/.buildkite/go/build.sh
@@ -15,5 +15,5 @@ set -euxo pipefail
 # Build the Go parts
 ####################
 pushd go
-  make all oasis-node/integrationrunner/integrationrunner.test
+  make all integrationrunner
 popd
--- a/go/Makefile
+++ b/go/Makefile
@@ -30,6 +30,10 @@ build: generate
 	@echo "Building Oasis local network runner"
 	@env -u GOPATH $(OASIS_GO) build $(GOFLAGS) -o ./oasis-net-runner/oasis-net-runner ./oasis-net-runner
 
+# Oasis node with coverage.
+oasis-node/integrationrunner/integrationrunner.test: generate
+	env -u GOPATH $(OASIS_GO) test $(GOFLAGS) -c -covermode=atomic -coverpkg=./... -o $@ ./oasis-node/integrationrunner
+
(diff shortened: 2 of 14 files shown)


pro-wh:
>  Unable to find report content in the storage archive.

aw

pro-wh:
oh no it's running TestIntegration in the Go tests

pro-wh:
https://codecov.io/gh/oasislabs/oasis-core/tree/c450dcd86d5bed94b38374b10eceecd8a0a71c2d this isn't right, we use those functions

pro-wh:
oh hold for work that needs to be done in #2321 instead

pro-wh:
ok have a look

go/Makefile
```diff
@@ -30,6 +30,10 @@ build: generate
 	@echo "Building Oasis local network runner"
 	@env -u GOPATH $(OASIS_GO) build $(GOFLAGS) -o ./oasis-net-runner/oasis-net-runner ./oasis-net-runner
 
+# Oasis node with coverage.
+oasis-node/integrationrunner/integrationrunner.test: generate
```
kostko: Maybe a shorter (phony) target name?

go/oasis-node/integrationrunner/integration_test.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"flag"
+	"testing"
+)
+
+var run = flag.Bool("integration.run", false, "Run the node")
+
+func TestIntegration(t *testing.T) {
+	if !*run {
+		t.Skip("Pass -integration.run to run the node")
```
kostko: This is to avoid running this during normal test runs? Can you add some comments around here (maybe package-level) how this is used?

scripts/build-run-e2e-cov.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh -e
```
kostko: Where is this script used, is this just for testing?

pro-wh: for manual invocation

go/oasis-node/integrationrunner/integration_test.go
pro-wh: yeah. I'll add a comment

go/Makefile
pro-wh: k I'll add a phony target

.buildkite/go/build.sh
```diff
@@ -15,5 +15,5 @@ set -euxo pipefail
 # Build the Go parts
 ####################
 pushd go
-  make
+  make all oasis-node/integrationrunner/integrationrunner.test
```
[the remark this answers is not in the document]
pro-wh: eh fine

PR #2244, opened 2019-10-09
Delegation

### LINKED ISSUE #2212: Support for delegation
We need support for delegation.

As we have two ledgers (staking and runtime) there are multiple options:
* Support delegation via the staking ledger in the consensus layer (works even if the runtimes are not working).
* Support delegation via the runtimes (similar to #2077).
* Support both.

Most likely at least part of this support is required in the consensus layer.

*More requirements and options TBD.*

kostko:
Fixes #2212 

TODO
* [x] Dump and restore delegation descriptors.
* [x] Dump and restore debonding delegation descriptors.
* [x] Make debonding period epoch-granular.

### DIFF
--- a/go/tendermint/apps/staking/staking.go
+++ b/go/tendermint/apps/staking/staking.go
@@ -528,35 +528,21 @@ func (app *stakingApplication) reclaimEscrow(ctx *abci.Context, state *MutableSt
 // stake to meet the sum of the thresholds specified.  The thresholds vector
 // can have multiple instances of the same threshold kind specified, in which
 // case it will be factored in repeatedly.
-func EnsureSufficientStake(appState *abci.ApplicationState, ctx *abci.Context, id signature.PublicKey, thresholds []staking.ThresholdKind) error {
-	state := NewMutableState(ctx.State())
-
-	m, err := state.Thresholds()
+func EnsureSufficientStake(ctx *abci.Context, id signature.PublicKey, thresholds []staking.ThresholdKind) error {
+	sc, err := NewStakeCache(ctx)
 	if err != nil {
-		return errors.Wrap(err, "staking/tendermint: failed to query thresholds")
+		return err
 	}
-	escrowBalance := state.EscrowBalance(id)
-
-	var targetThreshold staking.Quantity
-	for _, v := range thresholds {
-		qty := m[v]
-		if err = targetThreshold.Add(&qty); err != nil {
-			return errors.Wrap(err, "staking/tendermint: failed to accumulate threshold")
-		}
-	}
-
-	if escrowBalance.Cmp(&targetThreshold) < 0 {
-		return staking.ErrInsufficientStake
-	}
-
-	return nil
+	return sc.EnsureSufficientStake(id, thresholds)
 }
 
-// Snapshot is a snapshot of the escrow balances and thresholds that can be
-// used in lieu of repeated queries to `EnsureSufficientStake` at a given
-// height.  This should be favored when repeated queries are going to
+// StakeCache is a lookup cache for escrow balances and thresholds that
+// can be used in lieu of repeated queries to `EnsureSufficientStake` at a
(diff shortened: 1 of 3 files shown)


go/tendermint/apps/epochtime_mock/epochtime_mock.go
```diff
@@ -149,7 +149,8 @@ func (app *epochTimeMockApplication) EndBlock(request types.RequestEndBlock) (ty
 	return types.ResponseEndBlock{}, nil
 }
 
-func (app *epochTimeMockApplication) FireTimer(ctx *abci.Context, timer *abci.Timer) {
+func (app *epochTimeMockApplication) FireTimer(ctx *abci.Context, timer *abci.Timer) error {
+	return nil
```
Yawning: Should this return an error?

go/tendermint/apps/staking/staking.go
```diff
@@ -158,40 +158,42 @@ func (app *stakingApplication) InitChain(ctx *abci.Context, request types.Reques
...
 			)
 			return errors.New("staking/tendermint: invalid genesis escrow balance")
 		}
-
-		account := &ledgerEntry{
-			GeneralBalance:  v.GeneralBalance,
-			EscrowBalance:   v.EscrowBalance,
-			DebondStartTime: v.DebondStartTime,
-			Nonce:           v.Nonce,
+		if !v.Escrow.DebondingShares.IsZero() {
```
Yawning: Whowa, is this an ok assumption to make?  I don't think it is...

go/tendermint/apps/staking/staking.go
```diff
@@ -323,15 +459,19 @@ func (app *stakingApplication) queryGenesis(s, r interface{}) ([]byte, error) {
...
 		acct := state.account(acctID)
-		ledger[acctID.ToMapKey()] = &staking.GenesisLedgerEntry{
-			GeneralBalance:  acct.GeneralBalance,
-			EscrowBalance:   acct.EscrowBalance,
-			DebondStartTime: acct.DebondStartTime,
-			Nonce:           acct.Nonce,
-		}
+		// All debonding stops on a dump/restore and needs to be restarted. This
+		// is also checked on chain init.
+		acct.Escrow.DebondingShares = *staking.NewQuantity()
```
Yawning: Ah I see.  I'm not sure people will be happy about this...

go/tendermint/apps/staking/staking.go
```diff
@@ -542,63 +712,73 @@ func (app *stakingApplication) reclaimEscrow(ctx *abci.Context, state *MutableSt
...
-			"now", now,
-			"debond_at", debondAt,
+	// Fetch escrow account.
+	from := state.account(reclaim.Account)
+
+	// Fetch delegation.
+	delegation := state.delegation(id, reclaim.Account)
+
+	// Update delegated shares.
+	if err := delegation.Shares.Sub(&reclaim.Shares); err != nil {
```
Yawning: Hm, so "reclaim all my shares" requires querying the current share count?  We could use `SubUpTo` to make this easier, but not sure if being convenient is better than being explicit (No opinion either way).

go/tendermint/apps/staking/staking.go
Yawning: Doesn't this eat tokens that are currently debonding?  The delegation's been updated, so the tokens held in debonding limbo need to either be put back, or this needs to be preserved no?

go/tendermint/apps/staking/staking.go
kostko: Fixed so dump/restore now also dumps and restores unbonding delegations.

go/tendermint/apps/staking/staking.go
kostko: Fixed so dump/restore now also dumps and restores unbonding delegations.

kostko:
Changed debonding so that it is now epoch-based.

Issue #965, opened 2018-09-21
Specific node parameters should be backed by consensus


kostko:
There are some node batch parameters which are currently command-line flags but in production should be backed by consensus.

These are:
* Maximum number of runtime calls in a batch.
* Maximum size in bytes of a batch.

The best way to implement this would be to have a "protocol parameters backend" which would expose methods for getting the values of these parameters for a given epoch. This way we could also change these (and other parameters) on a global scale easily.

PR #2262, opened 2019-10-14
runtime: check AVR quote for debug bit

matevz:
PR for https://github.com/oasislabs/ekiden/issues/2181

This PR:
- adds `OASIS_UNSAFE_ALLOW_DEBUG_ENCLAVES` env variable at compile time which checks AVR attributes Debug flag in rust,
- adds `ias.debug.allow_debug_enclaves` runtime switch for worker which checks AVR attributes Debug flag in go,
- if debug enclaves are not allowed, they are dropped; similarly, if debug enclaves are allowed, production enclaves are dropped.

### DIFF
--- a/go/common/sgx/ias/avr.go
+++ b/go/common/sgx/ias/avr.go
@@ -19,7 +19,10 @@ import (
 
 const nonceMaxLen = 32
 
-var unsafeSkipVerify bool
+var (
+	unsafeSkipVerify         bool
+	unsafeAllowDebugEnclaves bool
+)
 
 // TimestampFormat is the format of the AVR timestamp, suitable for use with
 // time.Parse.
@@ -404,6 +407,18 @@ func validateAVRSignature(data, encodedSignature, encodedCertChain []byte, trust
 // of the process' lifetime.
 func SetSkipVerify() {
 	unsafeSkipVerify = true
+}
+
+// SetAllowDebugEnclave will enable running and communicating with enclaves
+// with debug flag enabled in AVR for the remainder of the process' lifetime.
+func SetAllowDebugEnclaves() {
+	unsafeAllowDebugEnclaves = true
+}
+
+// UnsetAllowDebugEnclave will disable running and communicating with enclaves
+// with debug flag enabled in AVR for the remainder of the process' lifetime.
+func UnsetAllowDebugEnclaves() {
+	unsafeAllowDebugEnclaves = false
 }
 
 // BuildMrSignerBlacklist builds the MRSIGNER blacklist.
--- a/go/common/sgx/ias/avr_test.go
+++ b/go/common/sgx/ias/avr_test.go
@@ -14,6 +14,10 @@ func TestAVR(t *testing.T) {
 }
 
 func testAVRv2(t *testing.T) {
+	// TODO: Generate and test production AVR without debug bit.
+	SetAllowDebugEnclaves()
+	defer UnsetAllowDebugEnclaves()
+
 	raw, sig, certs := loadAVRv2(t)
 
 	avr, err := DecodeAVR(raw, sig, certs, IntelTrustRoots, time.Now())
--- a/go/common/sgx/ias/ias.go
+++ b/go/common/sgx/ias/ias.go
@@ -101,6 +101,10 @@ func (e *httpEndpoint) VerifyEvidence(ctx context.Context, quoteBinary, pseManif
 		return nil, nil, nil, fmt.Errorf("ias: invalid nonce length")
 	}
 
+	if err := quote.Verify(); err != nil {
+		return nil, nil, nil, err
+	}
+
 	// Encode the payload in the format that IAS wants.
 	reqPayload, err := json.Marshal(&iasEvidencePayload{
 		ISVEnclaveQuote: quoteBinary,
@@ -253,7 +257,8 @@ func NewIASEndpoint(cfg *EndpointConfig) (Endpoint, error) {
 	if cfg.DebugIsMock {
 		logger.Warn("DebugSkipVerify set, VerifyEvidence calls will be mocked")
 
-		SetSkipVerify() // Intel isn't signing anything.
+		SetSkipVerify()         // Intel isn't signing anything.
+		SetAllowDebugEnclaves() // Debug enclaves are used for testing.
 		return &mockEndpoint{
 			spidInfo: SPIDInfo{
 				SPID:               spidBin,
--- a/go/common/sgx/ias/quote.go
+++ b/go/common/sgx/ias/quote.go
@@ -85,11 +85,34 @@ func (b *Body) MarshalBinary() ([]byte, error) {
 	return bBin, nil
 }
 
+// AttributesFlags is attributes flags inside enclave report attributes.
+type AttributesFlags uint64
+
+// Predefined enclave report attributes flags.
+const (
+	AttributeInit          AttributesFlags = 0b0000_0001
+	AttributeDebug         AttributesFlags = 0b0000_0010
+	AttributeMode64Bit     AttributesFlags = 0b0000_0100
+	AttributeProvisionKey  AttributesFlags = 0b0001_0000
+	AttributeEInitTokenKey AttributesFlags = 0b0010_0000
+)
+
+// Attributes is a SGX enclave attributes value inside report.
+type Attributes struct {
+	Flags AttributesFlags
+	Xfrm  uint64
+}
+
+// GetFlagInit returns value of given flag attribute of the Report.
+func (a AttributesFlags) Contains(flag AttributesFlags) bool {
+	return (uint64(a) & uint64(flag)) != 0
+}
+
 // Report is an enclave report body.
-type Report struct {
+type Report struct { // nolint: maligned
 	CPUSVN     [16]byte
 	MiscSelect uint32
-	Attributes [16]byte
+	Attributes Attributes
 	MRENCLAVE  sgx.MrEnclave
 	MRSIGNER   sgx.MrSigner
 	ISVProdID  uint16
@@ -102,12 +125,16 @@ func (r *Report) MarshalBinary() ([]byte, error) {
 	rBin := []byte{}
 	uint16b := make([]byte, 2)
 	uint32b := make([]byte, 4)
+	uint64b := make([]byte, 8)
 
 	rBin = append(rBin, r.CPUSVN[:]...)
 	binary.LittleEndian.PutUint32(uint32b, r.MiscSelect)
 	rBin = append(rBin, uint32b[:]...)
 	rBin = append(rBin, make([]byte, 28)...) // 28 reserved bytes.
-	rBin = append(rBin, r.Attributes[:]...)
+	binary.LittleEndian.PutUint64(uint64b, uint64(r.Attributes.Flags))
+	rBin = append(rBin, uint64b[:]...)
+	binary.LittleEndian.PutUint64(uint64b, r.Attributes.Xfrm)
+	rBin = append(rBin, uint64b[:]...)
 	rBin = append(rBin, r.MRENCLAVE[:]...)
 	rBin = append(rBin, make([]byte, 32)...) // 32 reserved bytes.
 	rBin = append(rBin, r.MRSIGNER[:]...)
@@ -126,7 +153,8 @@ func (r *Report) MarshalBinary() ([]byte, error) {
 func (r *Report) UnmarshalBinary(data []byte) error {
 	copy(r.CPUSVN[:], data[0:])
 	r.MiscSelect = binary.LittleEndian.Uint32(data[16:])
-	copy(r.Attributes[:], data[48:])
+	r.Attributes.Flags = AttributesFlags(binary.LittleEndian.Uint64(data[48:]))
+	r.Attributes.Xfrm = binary.LittleEndian.Uint64(data[56:])
 	_ = r.MRENCLAVE.UnmarshalBinary(data[64 : 64+sgx.MrEnclaveSize])
 	_ = r.MRSIGNER.UnmarshalBinary(data[128 : 128+sgx.MrSignerSize])
 	r.ISVProdID = binary.LittleEndian.Uint16(data[256:])
@@ -146,6 +174,19 @@ func (q *Quote) Verify() error {
 	if mrSignerBlacklist[q.Report.MRSIGNER] {
 		return fmt.Errorf("ias/quote: blacklisted MRSIGNER")
 	}
+
+	if !unsafeAllowDebugEnclaves {
+		// Disallow debug enclaves, if we are in production mode.
+		if q.Report.Attributes.Flags.Contains(AttributeDebug) {
+			return fmt.Errorf("ias/avr: disallowed debug enclave since we are in production mode")
+		}
+	} else {
+		// Disallow non-debug enclaves, if we are in debug mode.
+		if !q.Report.Attributes.Flags.Contains(AttributeDebug) {
+			return fmt.Errorf("ias/avr: disallowed production enclave since we are in debug mode")
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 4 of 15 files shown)


go/common/sgx/ias/quote.go
```diff
@@ -146,6 +165,14 @@ func (q *Quote) Verify() error {
 	if mrSignerBlacklist[q.Report.MRSIGNER] {
 		return fmt.Errorf("ias/quote: blacklisted MRSIGNER")
 	}
+
+	// Drop enclaves compiled in debug mode.
+	if os.Getenv("OASIS_UNSAFE_ALLOW_DEBUG_ENLCAVES") == "" {
```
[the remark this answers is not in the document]
kostko: Also I'm not sure if we want to set this via an env var on the Go side? Shouldn't it be the same as "skip AVR verify" (e.g., a special flag)?

go/common/node/node.go
```diff
@@ -460,6 +460,11 @@ func (c *CapabilityTEE) Verify(ts time.Time) error {
 			return err
 		}
 
+		// Verify the quote.
+		if err = q.Verify(); err != nil {
+			return err
+		}
```
pro-wh: why didn't we do this before? will adding this break things?

go/common/sgx/ias/ias.go
```diff
@@ -101,6 +102,13 @@ func (e *httpEndpoint) VerifyEvidence(ctx context.Context, quoteBinary, pseManif
 		return nil, nil, nil, fmt.Errorf("ias: invalid nonce length")
 	}
 
+	// Drop enclaves compiled in debug mode.
+	if os.Getenv("OASIS_UNSAFE_ALLOW_DEBUG_ENCLAVES") == "" {
+		if quote.Report.Attributes.GetFlag(AttributeDebug) {
+			return nil, nil, nil, fmt.Errorf("ias: dropped enclave running in debug mode")
+		}
+	}
```
pro-wh: would it be okay to upgrade this to a `quite.Verify()`?

go/common/sgx/ias/quote.go
```diff
@@ -85,11 +86,29 @@ func (b *Body) MarshalBinary() ([]byte, error) {
 	return bBin, nil
 }
 
+// Attributes is a SGX enclave attributes value.
+type Attributes [16]byte
+type AttributeFlag uint8
```
pro-wh: why are we using uint8? other materials are modeling it as uint64

runtime/src/common/sgx/avr.rs
```diff
@@ -316,6 +324,22 @@ pub fn verify(avr: &AVR) -> Fallible<AuthenticatedAVR> {
         _ => return Err(AVRError::MalformedQuote.into()),
     };
 
+    #[cfg(target_env = "sgx")]
+    {
+        if option_env!("OASIS_UNSAFE_ALLOW_DEBUG_ENCLAVES").is_none() {
+            let af = match AttributesFlags::from_bits(u64::from_le_bytes(
+                quote_body.report_body.attributes[0..8].try_into()?,
```
pro-wh: sorry for the scope creep: would it be possible to use sgx_isa to do the parsing of the whole report?

.buildkite/pipeline.yml
```diff
@@ -39,6 +39,7 @@ docker_plugin_sgx_config: &docker_plugin_sgx_config
     - /dev/isgx
   environment:
     - "OASIS_TEE_HARDWARE=intel-sgx"
+    - "OASIS_UNSAFE_ALLOW_DEBUG_ENCLAVES=1"
```
pro-wh: why is this alone here? could this be the wrong place to put it?

go/common/node/node.go
matevz: It seems to work fine.

.buildkite/pipeline.yml
matevz: For go, `OASIS_UNSAFE_ALLOW_DEBUG_ENCLAVES` is a runtime parameter. We could use `-ldflags` switch to set a compile-time flag, but this seemed even worse.

go/common/sgx/ias/quote.go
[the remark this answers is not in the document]
kostko: As I mentioned above, this should not be configured via an environment variable on the Go side. It should be similar to how we handle "skip AVR verify".

runtime/src/common/sgx/avr.rs
matevz: done

.buildkite/pipeline.yml
pro-wh: ah, right. let's unify discussion on Jernej's comment about that

runtime/src/common/sgx/avr.rs
```diff
@@ -15,8 +12,9 @@ use serde_json;
 use untrusted;
 use webpki;
 
-#[cfg(target_env = "sgx")]
-use sgx_isa::Report;
+use byteorder::{LittleEndian, ReadBytesExt};
+use sgx_isa::{AttributesFlags, Report};
+use std::io::{Cursor, Read, Seek, SeekFrom};
```
pro-wh: now that these aren't conditioned under `target_env`, we can put them into the std -> 3rd party dependencies -> local dependencies order

runtime/src/common/sgx/avr.rs
pro-wh: thanks!

runtime/src/common/sgx/avr.rs
```diff
@@ -314,44 +287,59 @@ pub fn verify(avr: &AVR) -> Fallible<AuthenticatedAVR> {
...
+    {
+        // Force-ratchet the clock forward, to at least the time in the AVR.
+        update_insecure_posix_time(timestamp);
+
+        // Return dummy AVR.
+        Ok(AuthenticatedAVR {
+            report_data: vec![],
+            identity: EnclaveIdentity {
+                mr_enclave: MrEnclave::from(vec![]),
+                mr_signer: MrSigner::from(vec![]),
```
pro-wh: was this desired? that we not do the usual checks and return a dummy AVR? I thought we use this in the gateway

go/common/sgx/ias/quote.go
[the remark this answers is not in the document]
matevz: Agree, I added `ias.debug.allow_debug_enclaves` go runtime flag now instead. 
`OASIS_UNSAFE_ALLOW_DEBUG_ENLCAVES` is used in compile-time rust.

runtime/src/common/sgx/avr.rs
matevz: Yeah, this was a leftover of `Report` cleanup. SGX/non-SGX case can now be joined.

go/common/sgx/ias/avr.go
```diff
@@ -20,6 +20,7 @@ import (
 const nonceMaxLen = 32
 
 var unsafeSkipVerify bool
+var unsafeAllowDebugEnclaves bool
```
Yawning: Use a var block.

```
var (
   foo bool
   bar uint32
)
```

go/common/sgx/ias/quote.go
```diff
@@ -146,6 +164,14 @@ func (q *Quote) Verify() error {
 	if mrSignerBlacklist[q.Report.MRSIGNER] {
 		return fmt.Errorf("ias/quote: blacklisted MRSIGNER")
 	}
+
+	// Drop enclaves compiled in debug mode.
```
Yawning: Random thought.  Should the unsafe flag, mandate debug enclaves?  IMO it should be either production enclaves OR debug enclaves, not "set a flag and be allowed to mix both".

go/common/sgx/ias/avr_test.go
```diff
@@ -14,6 +14,9 @@ func TestAVR(t *testing.T) {
 }
 
 func testAVRv2(t *testing.T) {
+	// TODO: Generate another testdata AVR with debug bit disabled.
+	unsafeAllowDebugEnclaves = true
```
Yawning: Unset this in a defer so that other tests that depend on verification being strict do the right thing.

go/common/sgx/ias/quote_test.go
```diff
@@ -10,6 +10,9 @@ import (
 )
 
 func TestQuote(t *testing.T) {
+	// TODO: Generate another testdata AVR with debug bit disabled.
+	unsafeAllowDebugEnclaves = true
```
Yawning: Likewise, unset this in a defer.

go/common/sgx/ias/quote_test.go
```diff
@@ -38,12 +41,8 @@ func TestQuote(t *testing.T) {
 		"CPUSVN",
 	)
 	require.EqualValues(t, 0x00000000, quote.Report.MiscSelect, "MISCSELECT")
-	require.Equal(
-		t,
-		"07000000000000000700000000000000",
-		hex.EncodeToString(quote.Report.Attributes[:]),
-		"ATTRIBUTES",
-	)
+	require.EqualValues(t, 0x0000000000000007, quote.Report.Attributes.Flags, "ATTRIBUTES.FLAGS")
```
Yawning: God I hope that we don't have XFRM and the flags transposed.

matevz: We don't ;)

go/common/sgx/ias/quote.go
matevz: Hm.. but this will prevent us from doing any tests with the production enclave. Will this never be the case?

Yawning:
So I had a thought about this, that might lead to more scope creep.  Given that we are doing #965, should allowing debug enclaves be consensus backed instead?

 1. We have a network wide consensus parameter that indicates if we allow debug enclaves or not.
 2. Each runtime's registry descriptor (`go/registry/runtime.go:Runtime`) gets a field that indicates if it is a debug enclave or not (Maybe move the exiting `TEEHardware` field into a struct or something?)
 3. On runtime registration (currently only in the genesis block, but the code is generic), if the newly registered runtime is a debug enclave, check to see if the consensus parameter (1) to allow such things is set, if not, reject the registration.
 4. Nodes that indicate runtime support MUST have the debug status match what is in the runtime descriptor (2).

For now `1` can be faked with a flag.  The other stuff can be implemented in a way that once we have consensus backed params, the changes should be minor (especially since the IAS proxy can query a node or interpret a genesis document).

kostko:
> should allowing debug enclaves be consensus backed instead?

I think this (and the proposed design) is a good idea.

So the runtimes would still need to respect the compile-time value of the parameter when validating AVRs as they currently can't trust consensus (that would require compiling-in the genesis block anyway which seems equivalent to just reading the compile-time env variable).

Yawning:
> So the runtimes would still need to respect the compile-time value of the parameter when validating AVRs as they currently can't trust consensus (that would require compiling-in the genesis block anyway which seems equivalent to just reading the compile-time env variable).

Yeah, this is just to prevent misconfigured nodes from doing the wrong thing, by forcing the policy to be uniform (MRENCLAVE will mismatch if the enclave is built with the wrong flags, so we're ok there.).

go/common/sgx/ias/quote.go
matevz: Fixed.

matevz:
I opened a separate issue https://github.com/oasislabs/oasis-core/issues/2285 for having on-chain debug enclaves flag. This PR addresses the CLI part only.

go/common/node/node.go
kostko: This is actually not needed as `avrBundle.Open` already calls `validate` which calls `Verify` on the quote. So this was already being verified in `Open` and can be removed.

go/common/sgx/ias/ias.go
```diff
@@ -101,6 +101,11 @@ func (e *httpEndpoint) VerifyEvidence(ctx context.Context, quoteBinary, pseManif
 		return nil, nil, nil, fmt.Errorf("ias: invalid nonce length")
 	}
 
+	// Drop enclaves compiled in debug mode.
```
kostko: This comment seems out of place? This verifies the quote.

go/common/sgx/ias/quote.go
```diff
@@ -146,6 +174,19 @@ func (q *Quote) Verify() error {
 	if mrSignerBlacklist[q.Report.MRSIGNER] {
 		return fmt.Errorf("ias/quote: blacklisted MRSIGNER")
 	}
+
+	if !unsafeAllowDebugEnclaves {
+		// Drop debug enclaves, if we are in production mode.
```
kostko: Maybe instead of "drop" use the term "disallow" (everywhere, also in error message below).

go/common/sgx/ias/quote_test.go
```diff
@@ -10,6 +10,10 @@ import (
 )
 
 func TestQuote(t *testing.T) {
+	// TODO: Generate another testdata AVR with debug bit disabled.
```
kostko: Can we get one? cc @pro-wh

go/ias/ias.go
```diff
@@ -209,18 +210,24 @@ func New(identity *identity.Identity) (*IAS, error) {
...
 	return s, nil
 }
 
 func init() {
-	Flags.String(cfgProxyAddress, "", "IAS proxy address")
-	Flags.String(cfgTLSCertFile, "", "IAS proxy TLS certificate")
-	Flags.Bool(cfgDebugSkipVerify, false, "skip IAS AVR signature verification (UNSAFE)")
+	Flags.String(CfgProxyAddress, "", "IAS proxy address")
+	Flags.String(CfgTLSCertFile, "", "IAS proxy TLS certificate")
+	Flags.Bool(CfgDebugSkipVerify, false, "skip IAS AVR signature verification (UNSAFE)")
```
kostko: I would hide the unsafe flags (e.g., using `Flags.MarkHidden`).

go/registry/api/api.go
```diff
@@ -80,6 +80,10 @@ var (
 	// register a runtime with bad Capabilities.TEE.Hardware.
 	ErrBadCapabilitiesTEEHardware = errors.New("registry: bad capabilities.TEE.Hardware")
 
+	// ErrDebugEnclave is the error returned when a node tries to register
+	// runtimes with enclaves running in debug mode.
+	ErrDebugEnclave = errors.New("registry: enclave running in debug mode not allowed")
```
kostko: Is the same error used if the node tries to register a production enclave but the network is running in debug mode? In this case the error description (and doc comment) should probably be updated.

runtime/src/common/sgx/avr.rs
```diff
@@ -316,14 +298,36 @@ pub fn verify(avr: &AVR) -> Fallible<AuthenticatedAVR> {
         _ => return Err(AVRError::MalformedQuote.into()),
     };
 
+    if option_env!("OASIS_UNSAFE_ALLOW_DEBUG_ENCLAVES").is_none() {
+        // Drop debug enclaves, if we are in production mode.
```
kostko: Same as above -- disallow instead of drop.

runtime/src/common/sgx/avr.rs
```diff
@@ -316,14 +298,36 @@ pub fn verify(avr: &AVR) -> Fallible<AuthenticatedAVR> {
...
+        if quote_body
```
kostko: This could be simplified by something like:
```rust
let is_debug = quote_body.report_body...contains(AttributeFlags::DEBUG);
let allow_debug = option_env!("...").is_none();
if is_debug != allow_debug { /* error */ }
```

go/common/sgx/ias/quote_test.go
matevz: Does this mean AVR needs to be signed with production key? (i.e. can you turn off debug bit and sign it with testing key?)

matevz: We cannot obtain AVR signed with the production key at the moment. Also confirmed that you cannot generate AVR with debug bit disabled without the production key. I am resolving this for now.

PR #2268, opened 2019-10-16
go/tendermint: ABCI interface cleanup

kostko:
Simplifies ABCI timers and transaction execution.


kostko:
Rebased on master (with timer changes from #2244) so we can merge this early.

Issue #2176, opened 2019-09-26
go/registry: Bad things happen on a dump/restore


Yawning:
Right now the registry dump/restore process will attempt to migrate the current validator set, by dumping and restoring all of the signed validator node descriptors (among other things).

As the epoch will get reset following the restoration, while each node descriptor's expiration is expressed in terms of an absolute epoch, the restored node descriptors will have an expiration time in the far flung future.

Since it is not possible to fixup the expiration in each node's descriptor, an alternative solution needs to be considered.

A "simple" solution could be to preserve the epoch time across migrations (so internally the epoch will be calculated by `baseEpoch + (blockHeight / interval)`, where `baseEpoch` is a parameter that is passed in the genesis block).  There is one place in the scheduler that also will need to be changed slightly (to prevent elections in the `0th` epoch), but other that that, this shouldn't be overly complicated.

PR #2271, opened 2019-10-17
go/epochtime: Add a notion of a base epoch

Yawning:
The `tendermint_mock` backend does not support this, because no one in
their right minds should be using it in situations where epoch
persistence across a dump/restore cycle matters.

Fixes: #2176


Yawning:
This probably isn't breaking, because the default base epoch if left unspecified is zero.

PR #2276, opened 2019-10-17
go oasis node: await first block before tests that need it

pro-wh:
this was a helper I used to run some node tests selectively


pro-wh:
oh ok yeah

PR #2280, opened 2019-10-18
go/tendermint: Remove Tendermint query indirection

### LINKED ISSUE #2001: Make sure all backend getters take a block height argument
Some backend getters currently don't take a block height (e.g., they always assume latest height) argument while they should.

### LINKED ISSUE #2173: miniscule nit: weird inefficiencies w/ Unmarshal and immediate re-Marshal
In this code:

https://github.com/oasislabs/ekiden/blob/4e9f4a306ea0e76a3daf459e040dd517b801e985/go/tendermint/apps/keymanager/keymanager.go#L161

the returned `status` -- an `api.Status` object -- is given to `cbor.Marshal` in 

https://github.com/oasislabs/ekiden/blob/4e9f4a306ea0e76a3daf459e040dd517b801e985/go/tendermint/apps/keymanager/keymanager.go#L165

if there was no error.  Inside `GetStatus`, 

https://github.com/oasislabs/ekiden/blob/4e9f4a306ea0e76a3daf459e040dd517b801e985/go/tendermint/apps/keymanager/state.go#L62

the data in `raw` was just `cbor.Unmarshal` to get the return value.  While it might make sense to do a data sanity check to ensure that `raw` can be successfully unmarshalled, we could just return `raw` so that the caller doesn't have to invoke `cbor.Marshal` and avoid that overhead.  Of course, doing the cancellation makes the semantics of `MutableState` and `ImmutableState` different (store byte slices instead of status values) which might be undesirable, so feel free to mark this as Will Not Fix.

### LINKED ISSUE #2273: Remove Tendermint query indirection
The Tendermint ABCI protocol has its own notion of queries which requires serialization round-trips for no good reason (also see #2173). As we are running everything in-process we should just stop using this and expose a proper query interface with methods instead of all this indirection.

kostko:
Fixes #2273 
Fixes #2173 
Fixes #2001


go/registry/tendermint/tendermint.go
```diff
@@ -72,35 +73,21 @@ func (r *tendermintBackend) DeregisterEntity(ctx context.Context, sigTimestamp *
 }
 
 func (r *tendermintBackend) GetEntity(ctx context.Context, id signature.PublicKey) (*entity.Entity, error) {
```
matevz: Rename `r` to `t` or `tb` for `tendermintBackend`.

go/keymanager/tendermint/tendermint.go
```diff
@@ -27,43 +27,28 @@ const BackendName = tmapi.BackendName
 type tendermintBackend struct {
 	logger *logging.Logger
 
-	service  service.TendermintService
+	service service.TendermintService
+	querier *app.QueryFactory
```
matevz: Is this the right term? Another weird term is `inquirer`, but it also seems wrong :P

go/staking/tendermint/tendermint.go
```diff
@@ -47,94 +48,57 @@ func (b *tendermintBackend) Symbol() string {
 }
 
 func (b *tendermintBackend) TotalSupply(ctx context.Context) (*api.Quantity, error) {
```
matevz: `b` -> `tb`?

go/scheduler/tendermint/tendermint.go
```diff
@@ -28,21 +28,22 @@ var (
...
+	service service.TendermintService
+	querier *app.QueryFactory
+
 	notifier *pubsub.Broker
 }
 
 func (s *tendermintScheduler) Cleanup() {
 }
 
 func (s *tendermintScheduler) GetCommittees(ctx context.Context, id signature.PublicKey, height int64) ([]*api.Committee, error) {
```
matevz: `s` -> `ts`?

go/roothash/tendermint/tendermint.go
```diff
@@ -96,48 +96,35 @@ func (r *tendermintBackend) GetGenesisBlock(ctx context.Context, id signature.Pu
...
 	r.Lock()
-	r.genesisBlocks[id.ToMapKey()] = &blk
+	r.genesisBlocks[id.ToMapKey()] = blk
 	r.Unlock()
 
-	return &blk, nil
+	return blk, nil
 }
 
 func (r *tendermintBackend) GetLatestBlock(ctx context.Context, id signature.PublicKey) (*block.Block, error) {
```
matevz: `r` -> `tb`?

go/keymanager/tendermint/tendermint.go
Yawning: Both "inquirer" and "querier" are valid.

kostko:
All done, also included fix for #2001.

go/client/client.go
```diff
@@ -244,7 +244,7 @@ func (c *Client) WatchBlocks(ctx context.Context, runtimeID signature.PublicKey)
 // Pass RoundLatest to get the latest block.
 func (c *Client) GetBlock(ctx context.Context, runtimeID signature.PublicKey, round uint64) (*block.Block, error) {
 	if round == RoundLatest {
-		return c.common.roothash.GetLatestBlock(ctx, runtimeID)
+		return c.common.roothash.GetLatestBlock(ctx, runtimeID, 0)
```
Yawning: Hmm, random thought.  There's a few places where we do this.  I wonder if people will confuse RoundLatest with 0.  Depending on the API, certain places expect one or the other...

go/grpc/registry/entity.proto
```diff
@@ -36,13 +36,15 @@ message DeregisterResponse {
 
 message EntityRequest {
     bytes id = 1;
+    int64 height = 2;
```
Yawning: General question applicable to everywhere that has a height param.  Should it consistently be `height = 1` so that when requests grow more fields, it still is consistent?  Or does it not matter?

go/client/client.go
kostko: So one is latest roothash round the other is latest consensus height.

It can be confusing yeah, not sure how to improve on that?

go/grpc/registry/entity.proto
kostko: Aren't we getting rid of protobufs anyway?

go/client/client.go
Yawning: Fair enough.

go/grpc/registry/entity.proto
Yawning: At some point, it's been on our todo list forever...

PR #2284, opened 2019-10-18
.buildkite/rust-coverage: tarpaulin revert to 0.8.7

ptrus:
- 0.9 seems to have increased the number of segfaults in tests.
- don't fail build if coverage fails


ptrus:
seems like oasis-core still fails a lot with this version, but at least oasis-runtime works fine

kostko:
If it randomly breaks too many builds let's just remove it for now? Having proper coverage numbers is less important than having a working CI.

ptrus:
> If it randomly breaks too many builds let's just remove it for now? Having proper coverage numbers is less important than having a working CI.

yeah i just made it soft-fail so failures won't fail the whole build

PR #2286, opened 2019-10-21
go/common/node: fix address IsRoutable

ptrus:
~In hindsight the split into IPv4 and IPv6 ranges was not necessary so i'm ok with reverting that part.~ reverted

However the `"::ffff:0:0/96",      // RFC 4291: IPv4-mapped Address` had to be removed as go converts IP4-mapped IPv6 addresses to IPv4 addresses automatically. So that range basically matched all IPv4 addresses.

### DIFF
--- a/go/common/node/address.go
+++ b/go/common/node/address.go
@@ -144,7 +144,6 @@ func init() {
 		"255.255.255.255/32", // RFC 919: Limited Broadcast
 		"::1/128",            // RFC 4291: Loopback Address
 		"::/128",             // RFC 4291: Unspecified Address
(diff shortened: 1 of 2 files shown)


go/common/node/address.go
```diff
@@ -60,11 +61,23 @@ func (a *Address) FromIP(ip net.IP, port uint16) error {
...
+				return false
+			}
 		}
+	} else if ipv6 := a.IP.To16(); ipv6 != nil {
+		for _, v := range unroutableIPv6Networks {
+			if v.Contains(a.IP) {
+				return false
+			}
+		}
+
```
kostko: Remove empty line?

ptrus:
done :+1:

Issue #2265, opened 2019-10-16
Reward distribution mechanism


kostko:
We need a reward distribution mechanism.

*Estimated cost: ~2 sprints*

kostko:
What else needs to be done here?

pro-wh:
probably only refinement left. let's open separate issues for those

kostko:
What about commission, do we need to support that? cc @peterjgilbert

kostko:
I guess this is done now.

Issue #1898, opened 2019-07-08
tendermint: Replace iavl with our own mkvs


Yawning:
We use the tendermint iavl library to store state for everything, but there's no reason that we need to do so, and it may be nice to re-use the MKVS implementation that we have for our own storage at some point.

Required:
 * [x] Stop hashing keys (#1743)
 * [x] Garbage collection (for state pruning).
 * [x] Support for iterating over keys (prefix scan).

*Estimated cost: 1 sprint*

kostko:
Another benefit is that our implementation is not sensitive to the order of updates so no sorting of keys is required. But we also need to evaluate performance first (would be beneficial for other purposes anyway).

kostko:
Yet another benefit of this is that it would make it easier to write light consensus clients (they could just use our storage interface which already supports remote queries with proofs) -- in Go or in Rust.

PR #2288, opened 2019-10-22
Distribute staking rewards

pro-wh:
It was stated in the incentive design that we would reward accounts for having escrowed stake.

```
hook onEpochChange():
  for a in accounts:
    a.escrowed += amountEligibleStake(a) * rewardRate
  butLessOverTime()
```

blocks #2265 
depends #2303

### DIFF
--- a/go/staking/api/api.go
+++ b/go/staking/api/api.go
@@ -3,7 +3,8 @@ package api
 
 import (
 	"context"
-	"errors"
+
+	"github.com/pkg/errors"
 
 	"github.com/oasislabs/oasis-core/go/common/cbor"
 	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
--- a/go/staking/api/api.go
+++ b/go/staking/api/api.go
@@ -3,8 +3,7 @@ package api
 
 import (
 	"context"
-
-	"github.com/pkg/errors"
+	"errors"
 
 	"github.com/oasislabs/oasis-core/go/common/cbor"
 	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
--- a/go/staking/api/api.go
+++ b/go/staking/api/api.go
@@ -501,6 +501,14 @@ type DebondingDelegation struct {
 	DebondEndTime epochtime.EpochTime `json:"debond_end"`
 }
 
+// RewardStep is one of the time periods in the reward schedule.
+type RewardStep struct {
+	Until       epochtime.EpochTime `json:"until"`
+	Interval    epochtime.EpochTime `json:"interval"`
+	Numerator   Quantity            `json:"numerator"`
+	Denominator Quantity            `json:"denominator"`
+}
+
 // Genesis is the initial ledger balances at genesis for use in the genesis
 // block and test cases.
 type Genesis struct {
@@ -519,6 +527,7 @@ type Genesis struct {
 type ConsensusParameters struct {
 	Thresholds              map[ThresholdKind]Quantity `json:"thresholds,omitempty"`
 	DebondingInterval       epochtime.EpochTime        `json:"debonding_interval,omitempty"`
+	RewardSchedule          []RewardStep               `json:"reward_schedule,omitempty"`
 	AcceptableTransferPeers map[signature.MapKey]bool  `json:"acceptable_transfer_peers,omitempty"`
 	Slashing                map[SlashReason]Slash      `json:"slashing,omitempty"`
 }
--- a/go/staking/api/api.go
+++ b/go/staking/api/api.go
@@ -503,10 +503,8 @@ type DebondingDelegation struct {
 
 // RewardStep is one of the time periods in the reward schedule.
 type RewardStep struct {
-	Until       epochtime.EpochTime `json:"until"`
-	Interval    epochtime.EpochTime `json:"interval"`
-	Numerator   Quantity            `json:"numerator"`
-	Denominator Quantity            `json:"denominator"`
+	Until epochtime.EpochTime `json:"until"`
+	Scale Quantity            `json:"scale"`
 }
 
 // Genesis is the initial ledger balances at genesis for use in the genesis
--- a/go/staking/api/rewards.go
+++ b/go/staking/api/rewards.go
@@ -1,0 +1,16 @@
+package api
+
+import (
+	"math/big"
+)
+
+var RewardAmountDenominator *Quantity
+
+func init() {
+	// Denominated in 1000th of a percent.
+	RewardAmountDenominator = NewQuantity()
+	err := RewardAmountDenominator.FromBigInt(big.NewInt(100_000))
+	if err != nil {
+		panic(err)
+	}
+}
--- a/go/staking/api/rewards.go
+++ b/go/staking/api/rewards.go
@@ -4,6 +4,7 @@ import (
 	"math/big"
 )
 
+// RewardAmountDenominator is the denominator for the reward rate.
 var RewardAmountDenominator *Quantity
 
 func init() {
--- a/go/tendermint/apps/scheduler/scheduler.go
+++ b/go/tendermint/apps/scheduler/scheduler.go
@@ -1,10 +1,12 @@
 package scheduler
 
 import (
+	"bytes"
 	"context"
 	"crypto"
 	"fmt"
 	"math/rand"
+	"sort"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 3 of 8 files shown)


pro-wh:
In this version:

- reward tokens come out of thin air
- no events emitted
- we modify the state tree values while we iterate over a range--is that legal?

pro-wh:
I changed this to pay rewards from the common pool instead of from thin air

go/staking/api/api.go
```diff
@@ -3,7 +3,8 @@ package api
 
 import (
 	"context"
-	"errors"
+
+	"github.com/pkg/errors"
```
kostko: Why this change?

go/tendermint/apps/staking/state/state.go
```diff
@@ -62,10 +62,14 @@ var (
 	//
 	// Value is empty.
 	debondingQueueKeyFmt = keyformat.New(0x57, uint64(0), &signature.MapKey{}, &signature.MapKey{}, uint64(0))
+	// rewardScheduleKeyFmt is the key format used for the reward schedule.
+	//
+	// Value is a CBOR-serialized slice of staking.RewardStep
```
kostko:
```suggestion
	// Value is a CBOR-serialized slice of staking.RewardStep.
```

go/tendermint/apps/staking/state/state.go
```diff
@@ -525,6 +547,82 @@ func (s *MutableState) TransferFromCommon(ctx *abci.Context, toID signature.Publ
...
+	if stepRelativeTime%activeStep.Interval != 0 {
+		return nil
+	}
+
+	commonPool, err := s.CommonPool()
+	if err != nil {
+		return errors.Wrap(err, "loading common pool")
+	}
+
+	s.Snapshot.IterateRange(accountKeyFmt.Encode(), nil, true, func(key, value []byte) bool {
```
kostko: This seems to go over all accounts? Shouldn't we only reward accounts that belong to entities running validator nodes? Otherwise people can decide to delegate to escrow accounts that don't run any nodes and thus get all the rewards without taking on any risk (as those accounts will never be slashed).

go/tendermint/apps/staking/state/state.go
```diff
@@ -525,6 +547,82 @@ func (s *MutableState) TransferFromCommon(ctx *abci.Context, toID signature.Publ
...
+			err = errors.Wrap(err1, "dividing by reward step denominator")
+			return true
+		}
+
+		if q.IsZero() {
+			return false
+		}
+
+		if err1 := staking.Move(&ent.Escrow.Active.Balance, commonPool, q); err1 != nil {
+			err = errors.Wrap(err1, "transferring to active escrow balance from common pool")
```
kostko: So if there is not enough funds to pay out the rewards, this will cause a consensus halt?

go/tendermint/apps/staking/state/state.go
kostko: Maybe we should record all nodes that were active for larger part of an epoch and only distribute rewards to accounts belonging to their entities.

go/staking/api/api.go
pro-wh: oh I was going to add some context to some errors, but then I refactored that stuff into other files. I'll revert this for now

go/tendermint/apps/staking/state/state.go
pro-wh: ok I'll work on that

go/tendermint/apps/staking/state/state.go
pro-wh: gonna include this in the next push

pro-wh:
in this version:
- rewards distributed as part of election
  - it's nice to have it all there so that you can't evade elections while getting rewards
  - if a node of yours was considered eligible for any committee or validator, then you're eligible
- no more separate reward interval
- reward step now has a scaling factor that affects rewards distributed in that interval
- moved denominator to a shared variable 🤷

go/tendermint/apps/staking/state/state.go
pro-wh: went with distributing reward when an entity's node is considered for election

go/staking/api/rewards.go
```diff
@@ -0,0 +1,16 @@
+package api
+
+import (
+	"math/big"
+)
+
+var RewardAmountDenominator *Quantity
```
kostko: Add documentation block.

go/tendermint/apps/scheduler/scheduler.go
```diff
@@ -40,6 +41,8 @@ var (
 	rngContextMerge                = []byte("EkS-ABCI-Merge")
 	rngContextValidators           = []byte("EkS-ABCI-Validators")
 
+	RewardFactorEpochElectionAny *staking.Quantity
```
kostko: Add documentation block. Is this something that should be part of the API or is it specific to the tendermint implementation?

go/tendermint/apps/staking/state/state.go
```diff
@@ -543,6 +565,64 @@ func (s *MutableState) TransferFromCommon(ctx *abci.Context, toID signature.Publ
...
+
+		if q.IsZero() {
+			continue
+		}
+
+		if err := staking.Move(&ent.Escrow.Active.Balance, commonPool, q); err != nil {
+			return errors.Wrap(err, "transferring to active escrow balance from common pool")
+		}
+
+		s.SetAccount(id, ent)
```
kostko: We should do all updates at the end when we are sure that there are no errors (or use a snapshot?). Otherwise this can cause stuff to still be committed (e.g., there is no implicit rollback in case of errors).

Note that in the current implementation this is probably not a problem as any error here will cause a consensus failure and nothing will be committed. But in case this is ever used outside of `BeginBlock` it may be useful to mitigate.

Another reason for deferring updates is non-determinism -- IAVL is sensitive to order of updates, so we (currently) need to sort the updates before committing them. _(An alternative is to have an intermediate layer that caches and sorts all IAVL updates before they are committed or to use our own data structure, e.g. #1898, which doesn't have this problem. But this is out of scope for this PR.)_

go/tendermint/apps/staking/state/state.go
```diff
@@ -543,6 +565,64 @@ func (s *MutableState) TransferFromCommon(ctx *abci.Context, toID signature.Publ
...
+		// We're past the end of the schedule.
+		return nil
+	}
+
+	commonPool, err := s.CommonPool()
+	if err != nil {
+		return errors.Wrap(err, "loading common pool")
+	}
+
+	for mk := range accounts {
```
kostko: Iteration over Go maps is non-deterministic and may cause consensus failures (different validators will produce different hashes due to IAVL being sensitive to order of updates). One idea is to add updates to a list and sort them before committing -- this would also solve the issue described below about possibly committing partial updates.

go/tendermint/apps/staking/state/state.go
pro-wh: is there a place outside of BeginBlock that would disrespect an error condition and proceed to act on the same tree? I'd consider continuing in that condition to be the problem.

I'll treat the nondeterministic map iteration as a separate issue and discuss that in your other comment.

go/tendermint/apps/staking/state/state.go
pro-wh: I'll sort these

go/tendermint/apps/scheduler/scheduler.go
pro-wh: ah yeah this would be better in API land where we've been keeping the denominators

go/staking/api/rewards.go
pro-wh: ok

go/tendermint/apps/scheduler/scheduler.go
pro-wh: or, more consistent, at least

go/tendermint/apps/staking/state/state.go
kostko: > is there a place outside of BeginBlock that would disrespect an error condition and proceed to act on the same tree? I'd consider continuing in that condition to be the problem.

Potentially in DeliverTx. An error there aborts the transaction but should not commit any state changes.

We could enforce that at the mux layer, but then we need a way to still have some state side effects even for failing transactions (because maybe sometimes you would want some even then).

pro-wh: > aborts the transaction but should not commit any state changes

aw ew

probably better to snapshot, so that we have slightly less "have all accounts in memory" situations, in case we ever have to get rid of them

kostko: Yeah makes sense. So for now I see that you just have a comment that says that accounts may be corrupted in case of errors. Maybe just add something like "if this is an issue, the caller should snapshot and rollback in case of errors".

pro-wh: added

pro-wh:
Gonna merge. This will not set a reward schedule in any testing or production genesis document, so it will not actually distribute any rewards on its own.

PR #2290, opened 2019-10-22
go/tendermint: Add validator slashing

### LINKED ISSUE #1901: Detect and punish validator misbehaviour
Validators should detect double signing (Tendermint/Cosmos should have existing provisions for this) and our ABCI mux needs to be updated to support handling incoming double-signing evidence for punishing validators.

Including E2E tests that invoke Byzantine behavior (e.g., using the Byzantine node) and verify that slashing and its interaction with staking/delegation works correctly.

*Estimated cost: ~1 sprint*

### LINKED ISSUE #2272: Persistent blacklist of known bad keys
We should consider having a persistent blacklist of keys that were found to be provably misbehaving (e.g. nodes that were double signing things they shouldn't be).

Those keys should be stored in the consensus layer and be prevented from registering and/or participating in the system.

kostko:
Based on #2301 
Fixes #1901 
Fixes #2272 

TODO
* [x] Add registry transaction to unfreeze node (if freeze period has passed).
* [x] Test double signing.
* [x] Bump protocol version number.
* [x] Trigger early validator election.

### DIFF
--- a/go/registry/api/api.go
+++ b/go/registry/api/api.go
@@ -156,6 +156,9 @@ type Backend interface {
 	// GetNode gets a node by ID.
 	GetNode(context.Context, signature.PublicKey, int64) (*node.Node, error)
 
+	// GetNodeStatus returns a node's status.
+	GetNodeStatus(context.Context, signature.PublicKey, int64) (*NodeStatus, error)
+
 	// GetNodes gets a list of all registered nodes.
 	GetNodes(context.Context, int64) ([]*node.Node, error)
 
--- a/go/registry/api/api.go
+++ b/go/registry/api/api.go
@@ -54,6 +54,10 @@ var (
 	// runtime registation in the genesis document.
 	RegisterGenesisRuntimeSignatureContext = []byte("EkRunGen")
 
+	// RegisterUnfreezeNodeSignatureContext is the context used for
+	// unfreezing nodes.
+	RegisterUnfreezeNodeSignatureContext = []byte("EkUzNReg")
+
 	// ErrInvalidArgument is the error returned on malformed argument(s).
 	ErrInvalidArgument = errors.New("registry: invalid argument")
 
@@ -100,6 +104,10 @@ var (
 	// ErrNodeExpired is the error returned when a node is expired.
 	ErrNodeExpired = errors.New("registry: node expired")
 
+	// ErrNodeCannotBeUnfrozen is the error returned when a node cannot yet be
+	// unfrozen due to the freeze period not being over yet.
+	ErrNodeCannotBeUnfrozen = errors.New("registry: node cannot be unfrozen yet")
+
 	// ErrForbidden is the error returned when an operation is forbiden by
 	// policy.
 	ErrForbidden = errors.New("registry: forbidden by policy")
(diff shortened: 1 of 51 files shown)


go/registry/api/status.go
```diff
@@ -0,0 +1,21 @@
...
+	epochtime "github.com/oasislabs/oasis-core/go/epochtime/api"
+)
+
+// FreezeForever is an epoch that can be used to freeze a node for
+// all (practical) time.
+const FreezeForever epochtime.EpochTime = 0xffffffffffffffff
+
+// NodeStatus is live status of a node.
+type NodeStatus struct {
+	// FreezeEndTime is the epoch when a frozen node becomes unfrozen.
```
Yawning: Slightly misleading, given the check.

kostko: Yes once this passes the node also needs to explicitly unfreeze itself to be unfrozen (at which point this will be reset). I joined the two under a single value, could add a separate flag for being frozen.

Yawning: I think them being merged is fine, maybe clarify the doc comment.

go/registry/api/api.go
```diff
@@ -54,6 +54,10 @@ var (
 	// runtime registation in the genesis document.
 	RegisterGenesisRuntimeSignatureContext = []byte("EkRunGen")
 
+	// RegisterUnfreezeNodeSignatureContext is the context used for
+	// unfreezing nodes.
+	RegisterUnfreezeNodeSignatureContext = []byte("EkUzNReg")
```
Yawning: I wonder if there is a way to ensure that we don't reuse contexts...  Alternatively I would be ok with changing the definition of a context to be between 1 to 255 (inclusive) octets.

go/registry/tendermint/tendermint.go
```diff
@@ -310,15 +333,6 @@ func (tb *tendermintBackend) onABCIEvents(ctx context.Context, events []abcitype
 					Entity:         &dereg.Entity,
 					IsRegistration: false,
 				})
-
-				// Node deregistrations.
```
Yawning: Do we no longer support this?

go/tendermint/apps/staking/slashing.go
```diff
@@ -0,0 +1,109 @@
...
+		)
+		return err
+	}
+
+	app.logger.Warn("slashed validator for double signing",
+		"node_id", node.ID,
+		"entity_id", node.EntityID,
+	)
+
+	// TODO: Trigger early validator re-election?
```
Yawning: This is "easy" to do, so I would be in favor of it.

go/tendermint/tendermint.go
```diff
@@ -350,6 +350,14 @@ func (t *tendermintService) broadcastTx(ctx context.Context, tag byte, tx interf
 	}
 }
 
+func (t *tendermintService) BroadcastEvidence(ctx context.Context, evidence tmtypes.Evidence) error {
+	if _, err := t.client.BroadcastEvidence(evidence); err != nil {
```
Yawning:
```
_, err := t.client.BroadcastEvidence(evidence)
return err
```

go/registry/api/api.go
kostko: Yeah we should go for longer contexts, the current length is arbitrary. Also if possible it would be nice to enforce non-conflicting contexts, not sure how to do it -- associate context with a specific type and keep a map?

go/registry/tendermint/tendermint.go
kostko: No, entity registration is not possible when there are any remaining nodes (and those are only removed after they expire + a debonding period elapses). Otherwise a node operator could simply deregister before the debonding period expires and there would be no way to know who to slash.

go/registry/api/api.go
kostko: #2311

pro-wh:
when you rebase, feel free to use a common denominator for the slashing fraction

kostko:
Rebased and added support for rescheduling.

go/tendermint/apps/staking/slashing.go
kostko: Done.

Issue #1879, opened 2019-07-01
Reorganize backends to be per consensus backend


kostko:
Since #1878 the backends for various services (staking, beacon, scheduler, roothash, etc.) have been made per consensus backend, so setting the consensus backend configures all of the appropriate backends.

We should also reorganize the Tendermint-based backends to live under `go/tendermint` and change the `common.ConsensusBackend` interface to include a method that returns all the (initialized) backends appropriate for that consensus backend.

kostko:
While we are doing this, we should also:
* Create `go/consensus` and move `go/common/consensus` stuff there.
* Move `go/tendermint` under `go/consensus/tendermint`.

Issue #1633, opened 2019-04-22
Write a comprehensive set of auditing tools


Yawning:
We should have as much automated tooling to monitor the state of our blockchain as possible to detect misbehavior.

* [ ] Audit log for entity, node, runtime registration, and staking.
* [ ] Automated anomaly detection (eg: Something should flip out if invariants are violated, one entity ends up with all the stake, someone is mounting a sybil attack, etc).
* [ ] A sanity checking tool for the genesis block (this should be reusable so that it can be called from both each backend's InitChain *and* various stand-alone tools).

More ideas welcome.  This probably will require spinning up a node that keeps all history, and tracks state changes.  How to best present this information is still an open question (should we have an at-a-glance overview of blockchain health as a webpage a la the [Tor Consensus Health](https://consensus-health.torproject.org/)?)

*Estimated cost: ~2 sprints*

PR #2294, opened 2019-10-23
Move various config flags to the genesis document

### LINKED ISSUE #965: Specific node parameters should be backed by consensus

abukosek:
Closes #965.

TODO:
- [x] Move lots of cmdline config flags to the genesis document
- [x] Rebase to latest master
- [x] Re-generate the existing `genesis.json` used in tests
- [x] Fix tests
- [x] Fix byzantine e2e tests


go/registry/api/api.go
```diff
@@ -792,19 +792,22 @@ type Genesis struct {
 	// KeyManagerOperator is the ID of the entity that is allowed to operate
 	// key manager nodes.
 	KeyManagerOperator signature.PublicKey `json:"km_operator"`
+
+	// Configuration is also stored as part of the genesis state.
+	Config Config `json:"config"`
```
kostko: As elsewhere, probably no need for additional indirection.

go/beacon/api/api.go
```diff
@@ -19,10 +19,18 @@ type Backend interface {
...
 	// beacon for latest finalized block.
 	GetBeacon(context.Context, int64) ([]byte, error)
+
+	// ToGenesis returns the genesis state at specified block height.
+	ToGenesis(context.Context, int64) (*Genesis, error)
+}
+
+// Genesis is the beacon genesis state
+type Genesis struct {
+	Config Config `json:"config"`
```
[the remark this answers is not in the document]
abukosek: The `Config` struct already existed, so I just included it here.  I'll move the flags directly into the `Genesis` struct.

go/oasis-node/cmd/genesis/genesis.go
```diff
@@ -41,6 +47,34 @@ const (
...
+
+	// EpochTime config flags.
+	cfgEpochTimeBackend            = "epochtime.backend"
+	cfgEpochTimeTendermintInterval = "epochtime.tendermint.interval"
+
+	// Tendermint config flags.
+	cfgConsensusTimeoutCommit      = "tendermint.consensus.timeout_commit"
+	cfgConsensusSkipTimeoutCommit  = "tendermint.consensus.skip_timeout_commit"
+	cfgConsensusEmptyBlockInterval = "tendermint.consensus.empty_block_interval"
+	cfgABCIPruneStrategy           = "tendermint.abci.prune.strategy"
```
Yawning: This doesn't need to be globally consistent.

go/oasis-node/cmd/genesis/genesis.go
```diff
@@ -41,6 +47,34 @@ const (
...
+	cfgABCIPruneNumKept            = "tendermint.abci.prune.num_kept"
```
Yawning: Neither does this.

go/oasis-node/cmd/genesis/genesis.go
```diff
@@ -41,6 +47,34 @@ const (
...
+	cfgRegistryDebugBypassStake              = "registry.debug.bypass_stake" // nolint: gosec
+
+	// Roothash config flags.
+	cfgRoundTimeout = "roothash.round_timeout"
+
+	// Scheduler config flags.
+	cfgSchedulerDebugBypassStake      = "scheduler.debug.bypass_stake" // nolint: gosec
+	cfgSchedulerDebugStaticValidators = "scheduler.debug.static_validators"
+	cfgSchedulerAlgorithm             = "worker.txnscheduler.algorithm"
+	cfgSchedulerBatchFlushTimeout     = "worker.txnscheduler.flush_timeout"
```
ptrus: probably also `txnscheduler.batching.max_batch_size`

go/oasis-node/cmd/genesis/genesis.go
```diff
@@ -41,6 +47,34 @@ const (
...
+
+	// Beacon config flags.
+	cfgBeaconDebugDeterministic = "beacon.debug.deterministic"
+
+	// EpochTime config flags.
+	cfgEpochTimeBackend            = "epochtime.backend"
```
ptrus: possibly also `consensus.backend` (not that we support any other than tendermint atm :P)

go/oasis-node/cmd/genesis/genesis.go
```diff
@@ -41,6 +47,34 @@ const (
...
 	cfgChainID            = "chain.id"
 
+	// Registry config flags.
+	cfgRegistryDebugAllowUnroutableAddresses = "registry.debug.allow_unroutable_addresses"
+	cfgRegistryDebugAllowRuntimeRegistration = "registry.debug.allow_runtime_registration"
+	cfgRegistryDebugBypassStake              = "registry.debug.bypass_stake" // nolint: gosec
+
+	// Roothash config flags.
+	cfgRoundTimeout = "roothash.round_timeout"
+
```
ptrus: probably also `debug.allow_test_keys` flag?

go/oasis-node/cmd/genesis/genesis.go
abukosek: This will probably need some refactoring to make it work -- we read the genesis document as part of the tendermint initialization, so we'd have to move that to an earlier point.  I'd rather leave this for a separate PR/issue :)

kostko: We do have our own genesis provider, it is not tied to Tendermint. See `go/genesis`. I think that one is initialized in the oasis-node binary so you could in theory look at it before initializing Tendermint.

go/epochtime/api/api.go
```diff
@@ -50,4 +50,10 @@ type SetableBackend interface {
 type Genesis struct {
 	// Base is the starting epoch.
 	Base EpochTime `json:"base"`
+
+	// Interval is the epoch interval (in blocks).
+	Interval int64 `json:"interval"`
+
+	// Backend is the chosen epochtime backend.
+	Backend string `json:"backend"`
```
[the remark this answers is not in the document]
abukosek: I've decided to leave this as-is for now, since this PR is already quite big.  It should be easy enough to change later in a separate PR.

go/oasis-node/cmd/genesis/genesis.go
abukosek: I talked to Yawning about this and this flag is also used in tools where the genesis document isn't available, so this will have to stay as it is :)

go/beacon/init.go
```diff
@@ -6,43 +6,19 @@ import (
...
-	CfgDebugDeterministic = "beacon.debug.deterministic"
-)
-
-// Flags has the configuration flags.
-var Flags = flag.NewFlagSet("", flag.ContinueOnError)
-
 // New constructs a new Backend based on the configuration flags.
 func New(ctx context.Context, timeSource epochtime.Backend, tmService service.TendermintService) (api.Backend, error) {
-	backend := commonFlags.ConsensusBackend()
+	backend := tmService.GetGenesis().ConsensusBackend
```
kostko: This looks funny as you are querying the tendermint service to give you the name of the consensus backend :grin: But this should all be fixed once we do #1879 as then the consensus backend will create these. Maybe just add an `XXX` comment that this will be improved in oasis-core#1879.

go/epochtime/api/api.go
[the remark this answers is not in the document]
kostko: Ok, maybe just add some `TODO` comments (here and in `epochtime.New`) that this should be changed, probably as part of #1879 which will reorganize backends to make these things more obvious.

go/genesis/api/api.go
```diff
@@ -25,6 +28,8 @@ type Document struct {
 	Time time.Time `json:"genesis_time"`
 	// ChainID is the ID of the chain.
 	ChainID string `json:"chain_id"`
+	// ConsensusBackend is the consensus backend.
+	ConsensusBackend string `json:"consensus_backend"`
```
kostko: Why not have this in `Consensus.Backend` since we have `Consensus`?

go/genesis/api/api.go
```diff
@@ -35,6 +40,12 @@ type Document struct {
 	Staking staking.Genesis `json:"staking"`
 	// KeyManager is the key manager genesis state.
 	KeyManager keymanager.Genesis `json:"keymanager"`
+	// Scheduler is the scheduler genesis state.
+	Scheduler scheduler.Genesis `json:"scheduler"`
+	// Beacon is the beacon genesis state.
+	Beacon beacon.Genesis `json:"beacon"`
+	// Consensus is the consensus genesis state.
+	Consensus tendermint.Genesis `json:"consensus"`
```
kostko: This structure should probably not be defined in `tendermint` but in `go/common/consensus` (to make it consensus-backend agnostic).

go/epochtime/api/api.go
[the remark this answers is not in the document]
ptrus: i'll change this in https://github.com/oasislabs/oasis-core/pull/2309 (as im moving it under the consensus backend flag as well)

go/tendermint/apps/beacon/state/state.go
```diff
@@ -15,6 +16,8 @@ var (
 	//
 	// Value is raw random beacon.
 	beaconKeyFmt = keyformat.New(0x40)
+	// genesisKeyFmt is the key format used for genesis.
+	genesisKeyFmt = keyformat.New(0x41)
```
kostko: Why do we need to store this in ABCI state, isn't the genesis document always available anyway?

go/genesis/grpc.go
```diff
@@ -72,18 +74,26 @@ func (s *grpcServer) ToGenesis(ctx context.Context, req *pb.GenesisRequest) (*pb
...
 		// https://github.com/tendermint/tendermint/issues/2543
-		Height:     0,
-		ChainID:    genesisDoc.ChainID,
-		Time:       time.Now(),
-		EpochTime:  *epochtimeGenesis,
-		Registry:   *registryGenesis,
-		RootHash:   *roothashGenesis,
-		Staking:    *stakingGenesis,
-		KeyManager: *keymanagerGenesis,
+		Height:           0,
```
Yawning: Unrelated to your PR, we should record the height at which the dump was taken and just log/ignore it when restoring.

go/genesis/grpc.go
```diff
@@ -72,18 +74,26 @@ func (s *grpcServer) ToGenesis(ctx context.Context, req *pb.GenesisRequest) (*pb
...
+		ChainID:          genesisDoc.ChainID,
+		ConsensusBackend: genesisDoc.ConsensusBackend,
+		Time:             time.Now(),
```
Yawning: Likewise unrelated to your PR, I would want dumps taken on different hosts to be byte for byte identical, this should probably use the block timestamp.

go/tendermint/apps/roothash/state/state.go
```diff
@@ -83,6 +89,17 @@ func (s *ImmutableState) Runtimes() []*RuntimeState {
 	return runtimes
 }
 
+func (s *ImmutableState) RoundTimeout() (time.Duration, error) {
+	_, raw := s.Snapshot.Get(roundTimeoutKeyFmt.Encode())
+	if raw == nil {
```
Yawning: Hmmm, should this just bail if it is missing? (likewise for the other methods...)

We really need sanity checking of the genesis state at some point. :(

go/tendermint/genesis_single.go
```diff
@@ -31,7 +36,30 @@ func (p *singleNodeGenesisProvider) GetTendermintGenesisDocument() (*tmtypes.Gen
 // running a single node "network", primarily for testing.
 func NewSingleNodeGenesisProvider(identity *identity.Identity) (genesis.Provider, error) {
 	doc := &genesis.Document{
-		Time: time.Now(),
+		Time:             time.Now(),
+		ConsensusBackend: "tendermint",
```
Yawning: All of these should be constants if possible.

go/tendermint/genesis_single.go
```diff
@@ -31,7 +36,30 @@ func (p *singleNodeGenesisProvider) GetTendermintGenesisDocument() (*tmtypes.Gen
...
+		EpochTime: epochtime.Genesis{
+			Backend: "tendermint_mock",
```
Yawning: Whowa, really?  Yes, this is used by the unit tests.  It also was used when the node was being brought up stand-alone.  If we no longer care to support that use case, we should make this code significantly harder to trigger.

go/oasis-node/node_test.go
```diff
@@ -50,19 +50,10 @@ var (
...
-		{"consensus.backend", "tendermint"},
 		{"consensus.validator", true},
-		{"registry.debug.allow_unroutable_addresses", true},
-		{"registry.debug.allow_runtime_registration", true},
-		{"registry.debug.bypass_stake", true},
 		{"roothash.tendermint.index_blocks", true},
-		{"scheduler.debug.bypass_stake", true},
-		{"scheduler.debug.static_validators", true},
 		{"storage.backend", "leveldb"},
 		{"staking.debug.genesis_state", stakingTests.DebugGenesisState},
```
Yawning: This flag should be folded into `NewSingleNodeGenesisProvider` and gotten rid of somehow.

go/tendermint/tendermint.go
```diff
@@ -645,11 +640,6 @@ func (t *tendermintService) getTendermintGenesis() (*tmtypes.GenesisDoc, error)
 		return nil, errors.Wrap(err, "tendermint: failed to create genesis doc")
 	}
 
-	// HACK: Certain test cases use TimeoutCommit < 1 sec, and care about the
```
Yawning: You did move these somewhere right?  (sorry github's review-view makes it hard to find stuff).

go/tendermint/apps/roothash/state/state.go
kostko: Yeah the improved sanity checks and genesis tooling will be done next in #1633.

go/oasis-node/node_test.go
kostko: Do we even need a separate staking genesis state? This can probably be removed now as long as there is a (test?) genesis document provider which supplies the right genesis values.

Yawning: Not anymore, though it shouldn't be in `NewSingleNodeGenesisProvider` if we want to continue to support running single node networks.

kostko: AFAIK the only single node network now is the one in unit tests (now that there is the network runner). So maybe just rename it to `NewTestNodeGenesisProvider` or something and just drop support for running nodes in a single-node-without-genesis-file configuration?

Yawning: I would be ok with that.

go/tendermint/tendermint.go
abukosek: Yeah, sorry, I moved this back now.

go/tendermint/genesis_single.go
abukosek: I'll implement the suggestion from comment https://github.com/oasislabs/oasis-core/pull/2294#discussion_r340552075.

go/tendermint/apps/beacon/state/state.go
abukosek: I thought we wanted it to be consensus-backed, so that malicious nodes can't modify it on dump/restore.  The beacon's genesis data is basically just a single bool, so it shouldn't take much space.

abukosek:
Thanks to everyone for your comments, this PR is now ready for re-review :)

go/oasis-node/node_test.go
Yawning: Weren't we going to remove that flag and just fold the test genesis state into the provider somehow?

abukosek: I'd rather leave this refactor for a separate PR, as I'll be away for the next few days and will have to go through merge-conflict hell when I return if this PR isn't merged soon :)
I did rename the genesis provider and made it available only in unit tests, so it should at least be hard to trigger by accident.

abukosek: Nevermind, I mixed things up, moving the debug staking stuff into the test genesis doc is actually not that much extra work, I'll do it now :)

abukosek: OK, I implemented this in a separate commit, please re-review.

go/tendermint/genesis_testnode.go
```diff
@@ -0,0 +1,95 @@
+package tendermint
```
kostko: Could we move this under a `tests` submodule to make it more clear?

go/tendermint/apps/roothash/state/state.go
kostko: But otherwise yeah, I think these should still panic instead of returning a default (and the default should be set in `InitChain` instead).

go/common/consensus/consensus.go
```diff
@@ -27,3 +29,11 @@ type Backend interface {
 	// while they are running.
 	RegisterGenesisHook(func())
 }
+
+// Genesis contains various consensus config flags that should be part of the genesis state.
+type Genesis struct {
+	Backend                     string        `json:"consensus_backend"`
+	ConsensusTimeoutCommit      time.Duration `json:"consensus_timeout_commit"`
```
kostko: Remove `Consensus` and `consensus_` prefixes?

go/oasis-node/cmd/node/node.go
```diff
@@ -359,7 +369,9 @@ func (n *Node) initGenesis() error {
 //
 // WARNING: This will misbehave iff cmd != RootCommand().  This is exposed
 // for the benefit of tests and the interface is not guaranteed to be stable.
-func NewNode() (*Node, error) {
+//
+// testNode is true when the node is used in unit tests.
+func NewNode(testNode bool) (*Node, error) {
```
kostko: I think it would be nicer to have these as:
* `newNode(testNode bool)` (internal, not exported)
* `NewNode()`
* `NewTestNode()`

go/worker/txnscheduler/algorithm/batching/batching.go
```diff
@@ -17,9 +17,6 @@ const (
 	// Name of the scheduling algorithm.
 	Name = "batching"
 
-	// CfgMaxBatchSize configures the max batch size.
-	CfgMaxBatchSize = "worker.txnscheduler.batching.max_batch_size"
-
 	cfgMaxQueueSize      = "worker.txnscheduler.batching.max_queue_size"
```
ptrus: `cfgMaxQueueSize` &  `cfgMaxBatchSizeBytes` should probably also be consistent (sorry missed them last time since we don't every actually modify these two at the moment)

kostko: Probably not `cfgMaxQueueSize` but yeah `cfgMaxBatchSizeBytes` should be.

go/scheduler/api/api.go
```diff
@@ -163,17 +164,29 @@ type Backend interface {
...
 	// the staking related checks and operations.
-	DebugBypassStake bool
+	DebugBypassStake bool `json:"debug_bypass_stake"`
 
 	// DebugStaticValidators is true iff the scheduler should use
 	// a static validator set instead of electing anything.
-	DebugStaticValidators bool
+	DebugStaticValidators bool `json:"debug_static_validators"`
+
+	// SchedulingAlgorithm is the transaction scheduling algorithm.
```
kostko: Oh, this belongs in the "transaction scheduler" part while this is the _committee_ scheduler genesis part. So maybe we should just put this under roothash as that has the most to do with batches.

go/oasis-node/cmd/genesis/genesis.go
```diff
@@ -14,10 +14,14 @@ import (
 	flag "github.com/spf13/pflag"
```
ptrus: Might make sense to start versioning the genesis file and including a version field in the doc? (specially now as it might be changing a bit more often)

PR #2296, opened 2019-10-23
Add delayed node shutdown command

### LINKED ISSUE #1811: Add a delayed shutdown mechanism for nodes
Nodes currently commit to being available for the while at re-registration time.  There needs to be a way to:
 * Halt re-registration.
 * Sit around handling requests as normal till the last registration interval expires.
 * Gracefully terminate.

Just killing nodes does bad things to the system, and should be discouraged/disincentivised (Alternatively, a way to revoke a registration to trigger rescheduling may also be acceptable).

*Estimated cost: <1 sprint*

### LINKED ISSUE #2180: Wrap Tendermint subscriptions into buffered channels
Currently the Tendermint consensus backend directly exposes unbuffered subscriptions. We need to use those as the implementation of buffered subscriptions can cause automatic unsubscriptions in case these are not processed in a timely manner.

The problem with Tendermint's unbuffered subscriptions is that they are dangerous (e.g., see #2178) and can result in deadlocks.

We should wrap all subscriptions in our own buffered channel and an internal goroutine which reads from Tendermint's unbuffered subscription and buffers things internally.

jberci:
Closes #1811 
Closes #2180

### DIFF
--- a/go/common/persistent/persistent.go
+++ b/go/common/persistent/persistent.go
@@ -1,0 +1,112 @@
+// Package persistent provides a wrapper around a key-value database for use as
+// general node-wide storage.
+package persistent
+
+import (
+	"errors"
+	"path/filepath"
+
+	bolt "github.com/etcd-io/bbolt"
+
+	"github.com/oasislabs/oasis-core/go/common/cbor"
+)
+
+const dbName = "persistent-store.db"
+
+var (
+	// ErrNotFound is returned when the requested key could not be found in the database.
+	ErrNotFound = errors.New("persistent: key not found in database")
+)
+
+type StoreTxFunc func(*bolt.Tx, *bolt.Bucket) error
+
+// CommonStore is the interface to the common storage for the node.
+type CommonStore struct {
+	db *bolt.DB
+}
+
+// Close closes the database handle.
+func (cs *CommonStore) Close() {
+	cs.db.Close()
+}
+
+// GetServiceStore returns a handle to a per-service bucket for the given service.
+func (cs *CommonStore) GetServiceStore(name string) (*ServiceStore, error) {
+	byteName := []byte(name)
+	err := cs.db.Update(func(tx *bolt.Tx) error {
+		_, berr := tx.CreateBucketIfNotExists(byteName)
+		return berr
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	ss := &ServiceStore{
+		store:    cs,
+		name:     name,
+		byteName: byteName,
+	}
+	return ss, nil
+}
+
+// NewCommonStore opens the default common node storage and returns a handle.
+func NewCommonStore(dataDir string) (*CommonStore, error) {
+	db, err := bolt.Open(filepath.Join(dataDir, dbName), 0600, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	cs := &CommonStore{
+		db: db,
+	}
+
+	return cs, nil
+}
+
+// ServiceStore is a storage wrapper that automatically calls view callbacks with appropriate buckets.
+type ServiceStore struct {
+	store *CommonStore
+
+	name     string
+	byteName []byte
+}
+
+// Close invalidates the per-service database handle.
+func (ss *ServiceStore) Close() {
+	ss.store = nil
+}
+
+// View executes the given callback within a read-only transaction.
+func (ss *ServiceStore) View(cb StoreTxFunc) error {
+	return ss.store.db.View(func(tx *bolt.Tx) error {
+		bkt := tx.Bucket(ss.byteName)
+		return cb(tx, bkt)
+	})
+}
+
+// Update executes the given callback within a read-write transaction.
+func (ss *ServiceStore) Update(cb StoreTxFunc) error {
+	return ss.store.db.Update(func(tx *bolt.Tx) error {
+		bkt := tx.Bucket(ss.byteName)
+		return cb(tx, bkt)
+	})
+}
+
+// GetCBOR is a helper for retrieving CBOR-serialized values.
+func (ss *ServiceStore) GetCBOR(key []byte, value interface{}) error {
+	return ss.View(func(tx *bolt.Tx, bkt *bolt.Bucket) error {
+		bytes := bkt.Get(key)
+		if bytes != nil {
+			return cbor.Unmarshal(bytes, value)
+		}
+		return ErrNotFound
+	})
+}
+
+// PutCBOR is a helper for storing CBOR-serialized values.
+func (ss *ServiceStore) PutCBOR(key []byte, value interface{}) error {
+	return ss.Update(func(tx *bolt.Tx, bkt *bolt.Bucket) error {
+		bytes := cbor.Marshal(value)
+		return bkt.Put(key, bytes)
+	})
+}
--- a/go/control/grpc.go
+++ b/go/control/grpc.go
@@ -1,0 +1,42 @@
+package control
+
+import (
+	"context"
+
+	"github.com/oasislabs/oasis-core/go/client"
+	"github.com/oasislabs/oasis-core/go/common/grpc"
+	pb "github.com/oasislabs/oasis-core/go/grpc/control"
+)
+
+var (
+	_ pb.ControlServer = (*grpcServer)(nil)
+)
+
+type grpcServer struct {
+	nodeClient *client.Client
+}
+
+func (s *grpcServer) WaitSync(ctx context.Context, req *pb.WaitSyncRequest) (*pb.WaitSyncResponse, error) {
+	err := s.nodeClient.WaitSync(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return &pb.WaitSyncResponse{}, nil
+}
+
+func (s *grpcServer) IsSynced(ctx context.Context, req *pb.IsSyncedRequest) (*pb.IsSyncedResponse, error) {
+	synced, err := s.nodeClient.IsSynced(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return &pb.IsSyncedResponse{
+		Synced: synced,
+	}, nil
+}
+
+func NewGRPCServer(grpc *grpc.Server, nodeClient *client.Client) {
+	s := &grpcServer{
+		nodeClient: nodeClient,
+	}
+	pb.RegisterControlServer(grpc.Server(), s)
+}
--- a/go/control/grpc.go
+++ b/go/control/grpc.go
@@ -12,8 +12,23 @@ var (
 	_ pb.ControlServer = (*grpcServer)(nil)
 )
 
+// Shutdownable is an interface the node presents for shutting itself down.
+type Shutdownable interface {
+	// RequestShutdown is the method called by the control server to trigger node shutdown.
+	RequestShutdown() <-chan struct{}
+}
+
 type grpcServer struct {
+	node       Shutdownable
 	nodeClient *client.Client
+}
+
+func (s *grpcServer) RequestShutdown(ctx context.Context, req *pb.ShutdownRequest) (*pb.ShutdownResponse, error) {
+	ch := s.node.RequestShutdown()
+	if req.GetWait() {
+		<-ch
+	}
+	return &pb.ShutdownResponse{}, nil
 }
 
 func (s *grpcServer) WaitSync(ctx context.Context, req *pb.WaitSyncRequest) (*pb.WaitSyncResponse, error) {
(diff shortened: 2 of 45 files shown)


go/oasis-node/cmd/debug/control/control.go
```diff
@@ -1,5 +1,5 @@
-// Package client implements the client debug sub-commands.
-package client
+// Package control implements the control debug sub-commands.
```
kostko: Probably the control stuff should be moved outside debug?

go/oasis-node/cmd/control/control.go
```diff
@@ -1,5 +1,5 @@
-// Package client implements the client debug sub-commands.
-package client
+// Package control implements the control debug sub-commands.
```
kostko:
```suggestion
// Package control implements the control sub-commands.
```

go/oasis-node/cmd/control/control.go
```diff
@@ -91,11 +99,35 @@ func doWaitSync(cmd *cobra.Command, args []string) {
...
+		os.Exit(1)
+	}
+}
+
 // Register registers the client sub-command and all of it's children.
 func Register(parentCmd *cobra.Command) {
-	clientCmd.PersistentFlags().AddFlagSet(cmdGrpc.ClientFlags)
+	controlCmd.PersistentFlags().AddFlagSet(cmdGrpc.ClientFlags)
+
+	controlShutdownCmd.Flags().BoolVarP(&shutdownWait, "wait", "", false, "wait for the node to finish shutdown")
```
kostko: Shouldn't this set a short flag to use? Otherwise why the `P` version?

go/worker/registration/worker.go
```diff
@@ -34,12 +35,16 @@ const (
 // Flags has the configuration flags.
 var Flags = flag.NewFlagSet("", flag.ContinueOnError)
 
-// Registration is a service handling worker node registration.
-type Registration struct {
+type ShutdownCallback func()
```
kostko: Add documentation block.

go/worker/registration/worker.go
```diff
@@ -109,19 +115,29 @@ func (r *Registration) doNodeRegistration() {
...
-		case <-r.stopCh:
+		case <-w.stopCh:
 			return
 		case epoch := <-ch:
+			// Check first if a clean halt was requested.
+			select {
+			case <-w.stopReqCh:
+				w.logger.Info("node registration halt requested")
+				w.Stop()
+				w.shutdownCallback()
```
kostko: Check if callback is nil and only call it if non-nil.

go/oasis-node/cmd/node/node.go
```diff
@@ -222,6 +223,9 @@ func (n *Node) initAndStartWorkers(logger *logging.Logger) error {
 		n.svcTmnt,
 		n.P2P,
 		&workerCommonCfg,
+		func() {
+			n.Stop()
```
kostko: Maybe add a comment on when this is called (and maybe log something).

go/worker/registration/worker.go
```diff
@@ -109,19 +117,31 @@ func (r *Registration) doNodeRegistration() {
...
 		select {
-		case <-r.stopCh:
+		case <-w.stopCh:
 			return
 		case epoch := <-ch:
+			// Check first if a clean halt was requested.
+			select {
+			case <-w.stopReqCh:
+				w.logger.Info("node registration halt requested")
+				w.Stop()
```
kostko: Actually this needs to be changed to work as follows:
* When stop is requested the node should no longer update node registrations.
* But it must not shut down immediately after an epoch passes. Instead it must wait until the registry tells it that it has deregistered because a node can be registered until epoch X and if it is currently epoch X-10 then it needs to wait 10 epochs for deregistration before it is safe to shut down.

go/worker/registration/worker.go
```diff
@@ -130,108 +148,149 @@ func (r *Registration) doNodeRegistration() {
...
+	// Loop broken; shutdown requested.
+	publicKey := w.identity.NodeSigner.Public()
+
+	regCh, sub := w.registry.WatchNodes()
+	defer sub.Close()
+
+	// Check if the node is already deregistered.
+	node, err := w.registry.GetNode(w.ctx, publicKey, 0)
+	if err != nil {
+		w.logger.Error("can't get this node from the registry during shutdown wait",
```
kostko: Hm, when a node expires, it will be removed from the registry, so this is normal (if error is `registry.ErrNoSuchNode`)?

go/worker/registration/worker.go
```diff
@@ -328,16 +384,25 @@ func New(
...
 	logger := logging.GetLogger("worker/registration")
 
 	entityID, registrationSigner, err := GetRegistrationSigner(logger, dataDir, identity)
 	if err != nil {
 		return nil, err
 	}
 
-	r := &Registration{
+	shutdown := func() {
+		logger.Info("node deregistered, shutting down")
```
ptrus: maybe something like: `logger.Info("node deregistered, dsipatching shutdown callback")`

go/control/grpc.go
```diff
@@ -0,0 +1,53 @@
...
+	_ pb.ControlServer = (*grpcServer)(nil)
+)
+
+type grpcServer struct {
+	registration *registration.Worker
+	nodeClient   *client.Client
+}
+
+func (s *grpcServer) RequestShutdown(ctx context.Context, req *pb.ShutdownRequest) (*pb.ShutdownResponse, error) {
+	s.registration.ScheduleShutdown()
```
ptrus: A quick thought (not sure if this was already considered):

I think it would make a bit more sense if `ScheduleShutdown` was a method of `cmd/node/node.go`. 
That method would stop the registration worker and wait that node is un-registered and then also stop itself.

I think that would also avoid the need for passing in the `ShutdownCallback` to the registration worker (which "feels" wrong) and make this more extendable (in case in future we'd like to do more things on shutdown request).

go/worker/registration/worker.go
```diff
@@ -319,6 +370,11 @@ func GetRegistrationSigner(logger *logging.Logger, dataDir string, identity *ide
 	return entity.ID, entitySigner, nil
 }
 
+// ScheduleShutdown requests that the node not register itself in the next epoch.
+func (w *Worker) ScheduleShutdown() {
```
ptrus: Should the notion of a "scheduled shutdown" be persisted to disk?

If i'm not wrong, the way it is done currently: if a node get's restarted while a shutdown was scheduled it would start normally and forget about the scheduled shutdown. I think it would make sense to handle this case.

go/control/grpc.go
kostko: Yeah it would be a bit strange to have all this logic in the node as it needs to interact with the registry which is something that the registration worker does, so I think it belongs there.

Instead of passing an arbitrary callback you could just pass a stoppable interface (one with a `Stop()` method) so you could pass the node instance directly and avoid importing the node command. That would be slightly cleaner IMO.

go/tendermint/pubsub.go
```diff
@@ -0,0 +1,68 @@
...
+
+	tmpubsub "github.com/tendermint/tendermint/libs/pubsub"
+	tmtypes "github.com/tendermint/tendermint/types"
+)
+
+var (
+	_ tmtypes.Subscription = (*tendermintPubsubBuffer)(nil)
+)
+
+type tendermintPubsubBuffer struct {
```
kostko: Even though these are internal things, can you add some documentation comments around the struct that explains why this is needed and what it does (wrapping to avoid deadlocks in tendermint pubsub).

go/tendermint/tendermint.go
```diff
@@ -357,7 +357,11 @@ func (t *tendermintService) Subscribe(subscriber string, query tmpubsub.Query) (
 	// force-unsubscribe the channel if processing takes too long.
 
 	subFn := func() (tmtypes.Subscription, error) {
-		return t.node.EventBus().SubscribeUnbuffered(t.ctx, subscriber, query)
+		sub, err := t.node.EventBus().SubscribeUnbuffered(t.ctx, subscriber, query)
+		if err != nil {
+			return nil, err
+		}
+		return newTendermintPubsubBuffer(sub), nil
```
kostko: Don't forget to update `broadcastTx` to remove the "drain hack" that shouldn't be needed anymore.

go/control/grpc.go
ptrus: > Yeah it would be a bit strange to have all this logic in the node as it needs to interact with the registry which is something that the registration worker does, so I think it belongs there.

That logic could stay in the registration worker, but would be under the registration worker `Stop()`  method. I don't think it is correct to tie the "stopping of registration worker" with shutting down the Node. For example we might want to support stopping the registration worker without shutting down the node in future (while the other way: shutting down a node gracefully, should always wait for the registration worker to stop and de-register), that's why i think the `ScheduleShutdown` should be a `Node` method, that calls `registrationWorker.Stop()` (which would contain the logic to also wait that node is deregistered) and then shuts itself down.

go/oasis-node/cmd/node/node.go
```diff
@@ -107,12 +111,13 @@ type Node struct {
 	TransactionSchedulerWorker *txnscheduler.Worker
 	MergeWorker                *merge.Worker
 	P2P                        *p2p.P2P
-	WorkerRegistration         *registration.Registration
+	WorkerRegistration         *registration.Worker
```
ptrus: minor: `WorkerRegistration` -> `RegistrationWorker`

go/oasis-test-runner/oasis/compute.go
```diff
@@ -29,11 +27,18 @@ type Compute struct {
...
 type ComputeCfg struct {
+	NodeCfg
+
 	Entity *Entity
 
 	RuntimeBackend string
 }
 
+// SocketPath returns the path to the storage node's gRPC socket.
+func (worker *Compute) SocketPath() string {
```
ptrus: `SocketPath()` can probably be moved to the common `Node` struct?

go/worker/merge/worker.go
```diff
@@ -36,7 +36,7 @@ type Worker struct {
 	cfg     Config
 
 	commonWorker *workerCommon.Worker
-	registration *registration.Registration
+	registration *registration.Worker
```
ptrus: `registration` -> `registrationWorker`

go/worker/merge/worker.go
```diff
@@ -189,7 +189,7 @@ func (w *Worker) registerRuntime(id signature.PublicKey) error {
 func newWorker(
 	enabled bool,
 	commonWorker *workerCommon.Worker,
-	registration *registration.Registration,
+	registration *registration.Worker,
```
ptrus: `registration` -> `registrationWorker`

go/common/persistent/persistent.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// Close invalidates the per-service database handle.
+func (ss *ServiceStore) Close() {
+	ss.common = nil
+}
+
+// DB returns the underlying database handle.
+func (ss *ServiceStore) DB() *bolt.DB {
+	return ss.common.DB()
```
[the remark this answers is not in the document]
jberci: I made this as a compromise between providing all the functionality in the wrapper and letting users do their own thing if we're not hiding it's a BoltDB in the background. There aren't any current users tho, so I could remove it?

go/worker/registration/worker.go
jberci: I believe this is done now.

go/control/grpc.go
kostko: > that's why i think the ScheduleShutdown should be a Node method, that calls registrationWorker.Stop() (which would contain the logic to also wait that node is deregistered) and then shuts itself down.

Ah, yeah that makes sense. So the change would be minor -- that instead of calling a callback it would only signal that it has stopped. And then an external caller (e.g. the Node) would wait on that event and shut itself down.

I like this better and it should be a simple change.

go/oasis-test-runner/scenario/e2e/halt_restore.go
```diff
@@ -141,9 +141,16 @@ func (sc *haltRestoreImpl) Run(childEnv *env.Env) error {
 		return fmt.Errorf("scenario/e2e/halt_restore: failed waiting for halt epoch: %w", err)
 	}
 
-	// Wait additional few seconds for genesis docs to be dumped.
-	// XXX: fixed in #2296 where we wait for nodes to actually shutdown.
-	time.Sleep(3 * time.Second)
+	// Wait for validators to exit so that genesis docs are dumped.
```
ptrus: thx for doing this part! :+1: 

just wondering, why didn't we need to set validators to be `Restartable`?

jberci: I've made the validators restartable too now. The errors are ignored because the validators are shutdown/crash in the time between network runs, when there's nobody watching the error channel (and on the second network run, a new channel is created).

PR #2300, opened 2019-10-24
Initial sentry node support

tjanez:


go/oasis-test-runner/oasis/args.go
```diff
@@ -82,6 +83,46 @@ func (args *argBuilder) tendermintConsensusTimeoutCommit(d time.Duration) *argBu
...
+
+func (args *argBuilder) tendermintPrivatePeers(peers string) *argBuilder {
+	args.vec = append(args.vec, []string{
+		"--" + tendermint.CfgP2PPrivatePeers, peers,
+	}...)
+	return args
+}
+
+func (args *argBuilder) addSentriesAsPersistentPeers(sentries []*Sentry) *argBuilder {
+	peers := []string{}
```
Yawning: `var peers []string` is sufficient.

go/oasis-test-runner/oasis/args.go
```diff
@@ -82,6 +83,46 @@ func (args *argBuilder) tendermintConsensusTimeoutCommit(d time.Duration) *argBu
...
+	for _, sentry := range sentries {
+		peers = append(peers, sentry.tmAddress+"@127.0.0.1:"+strconv.Itoa(int(sentry.consensusPort)))
+	}
+	args = args.tendermintPersistentPeers(strings.Join(peers[:], ","))
```
Yawning: `strings.Join(peers, ",")`

go/common/node/address.go
```diff
@@ -137,6 +138,45 @@ func toProtoAddress(addr Address) *pbCommon.Address {
 	return pbAddr
 }
 
+// ConsensusAddress contains the ID and address for connecting to a consensus
+// member.
+// NOTE: This could be the P2P ID and IP of a sentry node.
+type ConsensusAddress struct {
+	ID      signature.PublicKey `json:"id"`
+	Address Address             `json:"address"`
```
ravenac95: General question...

Unless I'm mistake, the tendermint protocol sentry nodes as used for cosmos are meant to be a dynamic set of machines. These sentry node machines could in theory continuously disappear and reappear with different IP addresses and identities. The gossip nature of the tendermint protocol would allow this to work. Is there a reason that we need to specify more than the `ID` when registering a node for consensus?

cc @Yawning @kostko

kostko: Yeah this would probably work as well, but would rely exclusively on the seed nodes to establish peers via PEX.

Also tendermint AFAIK requires you to specify validator addresses in the genesis block?

kostko: Also "sentry nodes" (aka proxies) for other things (gRPC and possibly libp2p) will definitely require addresses and we should keep that in mind while designing any proxy stuff.

Yawning: > Also tendermint AFAIK requires you to specify validator addresses in the genesis block?

Strictly speaking, no: https://godoc.org/github.com/tendermint/tendermint/types#GenesisValidator

Our tendermint seed node implementation currently uses the genesis document to populate the initial address book.  This can be omitted, assuming "enough" nodes use the seed node.  Time to a functional network after redeploy will be longer.

From a practical perspective, we use more than tendermint, and I don't see us implementing a gossip protocol anytime soon.  So I'm not sure what this saves if anything.

ravenac95: As long as it's possible for the tendermint sentry node to disappear entirely, I would say whatever gets is alright. 

> Also "sentry nodes" (aka proxies) for other things (gRPC and possibly libp2p) will definitely require addresses and we should keep that in mind while designing any proxy stuff.

I would say registering an address for libp2p/grpc proxy would be fine. Would it be possible to register a DNS name and cert for those proxies so that IPs can float and be scaled out?

kostko: > Would it be possible to register a DNS name and cert for those proxies so that IPs can float and be scaled out?

Yes.

go/oasis-test-runner/oasis/fixture.go
```diff
@@ -239,6 +250,18 @@ func (f *ComputeWorkerFixture) Create(net *Network) (*Compute, error) {
...
 
+// SentryFixture is a sentry node fixture.
+type SentryFixture struct {
+	Validators []int `json:"validators"`
+}
+
+// Create instantiates the client node described by the fixture.
+func (f *SentryFixture) Create(net *Network) (*Sentry, error) {
+	return net.NewSentry(&SentryCfg{
+		ValidatorIndices: f.Validators,
```
kostko: Any reason why we don't resolve the indices here as we do with other things?

go/oasis-test-runner/oasis/fixture.go
```diff
@@ -118,7 +128,8 @@ func (f *ValidatorFixture) Create(net *Network) (*Validator, error) {
 	}
 
 	return net.NewValidator(&ValidatorCfg{
-		Entity: entity,
+		Entity:        entity,
+		SentryIndices: f.Sentries,
```
kostko: Any reason why we don't resolve the indices here as we do with other things?

go/oasis-test-runner/oasis/validator.go
```diff
@@ -107,25 +123,45 @@ func (net *Network) NewValidator(cfg *ValidatorCfg) (*Validator, error) {
...
-		ID: valIdentity.NodeSigner.Public(),
-	}
-	if err = valConsensusAddr.Address.FromIP(netPkg.ParseIP("127.0.0.1"), val.consensusPort); err != nil {
-		return nil, fmt.Errorf("oasis/validator: failed to parse IP: %w", err)
+	var consensusAddress node.ConsensusAddress
+	localhost := netPkg.ParseIP("127.0.0.1")
+	if len(val.sentryIndices) > 0 {
+		// Use the first sentry node's public address for initial validator
+		// registration.
+		// NOTE: This only applies to the Genesis document and the first epoch.
```
ptrus: Any reason why we don't update the `registry node init` to allow specifying multiple consensus addresses?

go/tendermint/tendermint.go
```diff
@@ -515,15 +519,28 @@ func (t *tendermintService) lazyInit() error {
...
+	// lowercasing the whole string is ok.
+	tenderConfig.P2P.PersistentPeers = strings.ToLower(viper.GetString(CfgP2PPersistentPeers))
+	tenderConfig.P2P.PexReactor = !viper.GetBool(CfgP2PDisablePeerExchange)
 	tenderConfig.P2P.SeedMode = viper.GetBool(CfgP2PSeedMode)
 	// Seed Ids need to be Lowecase as p2p/transport.go:MultiplexTransport.upgrade()
-	// uses a case sensitive string comparision to validate public keys
+	// uses a case sensitive string comparision to validate public keys.
 	// Since Seeds is expected to be in comma-delimited id@host:port format,
 	// lowercasing the whole string is ok.
 	tenderConfig.P2P.Seeds = strings.ToLower(viper.GetString(CfgP2PSeeds))
```
ptrus: I there there were some plans in making `CfgP2PSeeds` a string slice (on our end) instead of comma-delimited list, maybe a good time to change it

go/tendermint/tendermint.go
```diff
@@ -515,15 +519,28 @@ func (t *tendermintService) lazyInit() error {
 	tenderConfig.P2P.ListenAddress = viper.GetString(CfgCoreListenAddress)
 	tenderConfig.P2P.ExternalAddress = viper.GetString(cfgCoreExternalAddress)
 	tenderConfig.P2P.AllowDuplicateIP = true // HACK: e2e tests need this.
+	tenderConfig.P2P.PrivatePeerIDs = viper.GetString(CfgP2PPrivatePeers)
+	// Persistent peers need to be lowecase as p2p/transport.go:MultiplexTransport.upgrade()
+	// uses a case sensitive string comparision to validate public keys.
+	// Since persistent peers is expected to be in comma-delimited id@host:port format,
```
ptrus: similarly as with `CfgP2PSeeds` we maybe want to define `CfgP2PPrivatePeers` as a s string slice argument

go/oasis-test-runner/oasis/fixture.go
tjanez: Yes, makes sense. I've moved the sentry node resolution to `fixture.go`.

go/oasis-test-runner/oasis/fixture.go
tjanez: This one won't work because sentries are created before validators and hence, it is not possible to resolve them at this time.

We could move the creation of validators before the creation of sentry nodes, but we would then be unable to resolve sentry nodes when creating new validators.

To sum up, one of two can resolve indices upon creation, the other needs to have indices.

go/oasis-test-runner/oasis/validator.go
tjanez: Great suggestion, I've done that.

tjanez:
Ready for review.

go/oasis-test-runner/oasis/validator.go
```diff
@@ -101,19 +116,36 @@ func (net *Network) NewValidator(cfg *ValidatorCfg) (*Validator, error) {
...
+		var consensusAddrs []string
+		for _, sentry := range val.sentries {
+			var consensusAddr node.ConsensusAddress
+			consensusAddr.ID = sentry.publicKey
+			if err = consensusAddr.Address.FromIP(localhost, sentry.consensusPort); err != nil {
+				return nil, fmt.Errorf("oasis/validator: failed to parse IP address: %w", err)
+			}
+			consensusAddrs = append(consensusAddrs, consensusAddr.String())
+		}
+		consensusAddrStr = strings.Join(consensusAddrs, ",")
```
ptrus: might be nicer to pass multiple addresses be putting the `--node.consensus_address` flag multiple times? (and then using `viper.GetStringSlice` to get the slice of addresses)

go/worker/registration/registration.go
```diff
@@ -228,9 +230,45 @@ func (r *Registration) registerNode(epoch epochtime.EpochTime) error {
...
+		// Query sentry nodes for their consensus address(es).
+		for i, sentryAddr := range sentryAddrs {
+			var client *sentryClient.Client
+			client, err = sentryClient.New(r.ctx, &sentryAddr, sentryCerts[i], r.identity)
+			if err != nil {
+				r.logger.Error("failed to create client to a sentry node",
+					"err", err,
+					"sentry_address", sentryAddr,
+				)
+				return err
```
ptrus: returning on errors here means all sentry nodes need to be reachable for this to succeed right? Should instead just `continue` on errors and make sure at least on connection worked?

go/oasis-test-runner/scenario/e2e/sentry.go
```diff
@@ -0,0 +1,94 @@
...
+
+	f.Network.LogWatcherHandlers = append(
+		f.Network.LogWatcherHandlers,
+		// NOTE: This currently works because logs from all nodes are checked
+		// by the same log watcher handler.
+		// It needs to be properly implemented after:
+		// - https://github.com/oasislabs/oasis-core/issues/2355
+		// - https://github.com/oasislabs/oasis-core/issues/2356
+		// are implemented.
+		oasis.LogAssertPeerExchangeDisabled(),
```
ptrus: i might be miss-remembering but didn't this PR used to include multiple checks (such as asserting no validator IPs in logs?)

go/worker/registration/registration.go
tjanez: Yes, agreed. I meant to do that, but forgot to change `return`s to `continue`s.

tjanez: Done.

go/oasis-test-runner/scenario/e2e/sentry.go
tjanez: Yes, it had a check like that, but it couldn't be implemented easily due to #2355 and #2356.

I opened #2363 to add this test as a follow-up work to initial sentry node support.

go/oasis-test-runner/oasis/validator.go
tjanez: Indeed, the code already uses `viper.GetStringSlice` and I've made validator initialization pass the `--node.consensus_address` flag multiple times (if needed).

go/sentry/client/client.go
```diff
@@ -0,0 +1,148 @@
...
+	addresses, err := node.FromProtoConsensusAddresses(resp.Addresses)
+	if err != nil {
+		return nil, err
+	}
+
+	return addresses, nil
+}
+
+func (c *Client) createConnection() {
+	c.logger.Debug("creating connection to sentry node")
```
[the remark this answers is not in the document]
tjanez: Yes, agreed.

go/sentry/client/client.go
```diff
@@ -0,0 +1,148 @@
...
+	// connection before populating addresses.
+	// Dialing is deferred until use, which can't happen.
+	manualResolver, address, cleanupFn := manual.NewManualResolver()
+	conn, err := grpc.Dial(address, opts, grpc.WithBalancerName(roundrobin.Name)) //nolint: staticcheck
+	if err != nil {
+		cleanupFn()
+		c.logger.Error("failed to dial the sentry node",
+			"err", err,
+		)
+		return
```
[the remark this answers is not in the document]
tjanez: Good idea, done.

go/sentry/client/client.go
```diff
@@ -0,0 +1,148 @@
...
+func (c *Client) Close() {
+	if c.state != nil {
+		c.state.kill()
+		c.state = nil
+	}
+}
+
+func (c *Client) GetConsensusAddresses() ([]node.ConsensusAddress, error) {
+
+	if c.state == nil {
```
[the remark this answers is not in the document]
tjanez: Yes.

go/sentry/client/client.go
```diff
@@ -0,0 +1,148 @@
...
+
+	sentryAddress *node.Address
+	sentryCert    *x509.Certificate
+
+	nodeIdentity *identity.Identity
+
+	state *clientState
+}
+
+type clientState struct {
```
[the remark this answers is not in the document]
tjanez: Agreed, I've removed `clientState` and simplified things.

go/oasis-node/cmd/node/node.go
```diff
@@ -85,6 +88,7 @@ func Run(cmd *cobra.Command, args []string) {
 type Node struct {
 	svcMgr       *background.ServiceManager
 	grpcInternal *grpc.Server
+	grpcExternal *grpc.Server
```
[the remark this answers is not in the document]
tjanez: Agreed. I've removed it and created sentry worker in `go/sentry/worker`.

go/worker/sentry/worker.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"github.com/oasislabs/oasis-core/go/common/grpc"
+	"github.com/oasislabs/oasis-core/go/common/identity"
+	"github.com/oasislabs/oasis-core/go/common/logging"
+	"github.com/oasislabs/oasis-core/go/sentry"
+	"github.com/oasislabs/oasis-core/go/sentry/api"
+)
+
+const (
+	CfgEnabled     = "worker.sentry.enabled"
```
kostko: Add documentation comments.

go/worker/sentry/worker.go
```diff
@@ -0,0 +1,123 @@
...
+
+// Flags has the configuration flags.
+var Flags = flag.NewFlagSet("", flag.ContinueOnError)
+
+// Enabled returns true if Sentry worker is enabled.
+func Enabled() bool {
+	return viper.GetBool(CfgEnabled)
+}
+
+type Worker struct {
```
kostko: Add documentation comment.

go/worker/sentry/worker.go
```diff
@@ -0,0 +1,123 @@
...
+func New(backend api.Backend, identity *identity.Identity) (*Worker, error) {
+	w := &Worker{
+		enabled: Enabled(),
+		backend: backend,
+		quitCh:  make(chan struct{}),
+		initCh:  make(chan struct{}),
+		logger:  logging.GetLogger("worker/sentry"),
+	}
+
+	if w.enabled {
```
kostko: We should probably have a check so that you can't specify both `worker.sentry.enabled` & co. (sentry node options) and `worker.sentry.address` & co. (sentry consumers/clients aka. validator nodes) at the same time.

go/worker/sentry/worker.go
```diff
@@ -0,0 +1,123 @@
...
+		w.grpcServer = grpcServer
+		// Initialize and register the sentry gRPC service.
+		sentry.NewGRPCServer(w.grpcServer.Server(), backend)
+	}
+
+	return w, nil
+}
+
+func init() {
+	Flags.Bool(CfgEnabled, false, "Enable Sentry worker")
```
kostko: Maybe update help descriptions to make it really obvious which options are for sentry nodes and which are for sentry node consumers.

go/worker/sentry/worker.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// Stop halts the service.
+func (w *Worker) Stop() {
+	if !w.enabled {
+		close(w.quitCh)
+		return
+	}
+
+	// TODO: How to close the gRPC server?
```
kostko: Call `w.grpcServer.Stop()` and then during cleanup call `w.grpcServer.Cleanup()`.

go/worker/sentry/worker.go
```diff
@@ -0,0 +1,123 @@
...
+// Quit returns a channel that will be closed when the service terminates.
+func (w *Worker) Quit() <-chan struct{} {
+	return w.quitCh
+}
+
+// Cleanup performs the service specific post-termination cleanup.
+func (w *Worker) Cleanup() {
+}
+
+func New(backend api.Backend, identity *identity.Identity) (*Worker, error) {
```
kostko: Add documentation comment.

go/sentry/tendermint/tendermint.go
```diff
@@ -0,0 +1,49 @@
+// Package tendermint implements the tendermint backed sentry backend.
```
kostko: I think we don't need a "sentry backend" for each consensus backend. As long as `GetAddresses()` is exposed by the general `consensus.Backend` interface, this service can be totally consensus backend-agnostic.

I would just move this implementation to `go/sentry` directly (no `tendermint` directory needed) and change it so that instead of passing a `TendermintService` you pass a `consensus.Backend`.

tjanez: Agreed. Done.

go/worker/sentry/worker.go
tjanez: Thanks, done.

go/sentry/sentry.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/oasislabs/oasis-core/go/consensus"
+	"github.com/oasislabs/oasis-core/go/sentry/api"
+)
+
+// BackendName is the name of this implementation.
+const BackendName = "sentry"
+
+var _ api.Backend = (*Backend)(nil)
+
+type Backend struct {
```
kostko: This can be private as you only expose `api.Backend`.

go/sentry/sentry.go
```diff
@@ -0,0 +1,51 @@
...
+	"fmt"
+
+	"github.com/oasislabs/oasis-core/go/common/logging"
+	"github.com/oasislabs/oasis-core/go/common/node"
+	"github.com/oasislabs/oasis-core/go/consensus"
+	"github.com/oasislabs/oasis-core/go/sentry/api"
+)
+
+// BackendName is the name of this implementation.
+const BackendName = "sentry"
```
kostko: This is probably not needed anymore.

go/worker/sentry/worker.go
tjanez: I tried :slightly_smiling_face:. Please, take a look.

kostko: :+1:

go/sentry/client/client.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	if c.conn != nil {
+		c.conn.Close()
+		c.conn = nil
+	}
+}
+
+// GetConsensusAddresses returns the list of sentry node's consensus addresses.
+func (c *Client) GetConsensusAddresses() ([]node.ConsensusAddress, error) {
+
```
kostko: Remove empty line.

go/sentry/client/client.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+var (
+	// ErrSentryNotAvailable is the error returned when the sentry node is not
+	// available.
+	ErrSentryNotAvailable = errors.New("sentry/client: sentry node not available")
+)
+
+type Client struct {
+	ctx context.Context
```
kostko: It would be better if `GetConsensusAddresses` took a `context.Context` argument instead of storing a context here.

go/sentry/client/client.go
```diff
@@ -0,0 +1,137 @@
...
+	_ api.Backend = (*Client)(nil)
+)
+
+var (
+	// ErrSentryNotAvailable is the error returned when the sentry node is not
+	// available.
+	ErrSentryNotAvailable = errors.New("sentry/client: sentry node not available")
+)
+
+type Client struct {
```
kostko: Add documentation block.

go/sentry/grpc.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+var _ pb.SentryServer = (*grpcServer)(nil)
+
+type grpcServer struct {
+	backend api.Backend
+}
+
+func (s *grpcServer) GetConsensusAddresses(ctx context.Context, req *pb.GetConsensusAddressesRequest) (*pb.GetConsensusAddressesResponse, error) {
+	addresses, err := s.backend.GetConsensusAddresses()
```
kostko: Once the interface is changed as per my other comment, this should pass `ctx`.

go/tendermint/tendermint.go
tjanez: Done.

go/tendermint/tendermint.go
tjanez: Done.

go/tendermint/tendermint.go
tjanez: I've also decided to make a breaking configuration change here and renamed `tendermint.seeds` configuraiton flag to `tendermint.seed` to be consistent with newly added `tendermint.*peer` options.

@ptrus, please take a look.

go/consensus/tendermint/tendermint.go
```diff
@@ -70,6 +70,10 @@ const (
 	cfgABCIPruneStrategy = "tendermint.abci.prune.strategy"
 	cfgABCIPruneNumKept  = "tendermint.abci.prune.num_kept"
 
+	CfgP2PPrivatePeers        = "tendermint.private_peers"
```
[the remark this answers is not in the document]
tjanez: Done.

go/sentry/sentry.go
```diff
@@ -0,0 +1,46 @@
...
+	b.logger.Debug("successfully obtained consensus addresses",
+		"addresses", addrs,
+	)
+
+	return addrs, nil
+}
+
+// New constructs a new sentry Backend instance.
+func New(
+	consensusBackend consensus.Backend,
```
kostko: This can now be all in the same line.

PR #2303, opened 2019-10-24
split escrow between normal tokens and tokens undergoing debonding

pro-wh:
We'll need this for rewards, so that we can increase the value of the escrow shares that aren't undergoing debonding.

blocks #2288

cc #2290, which also changes slashing to take a fraction. That PR uses a single shared denominator, which is fine.

### DIFF
--- a/go/staking/api/api.go
+++ b/go/staking/api/api.go
@@ -371,6 +371,8 @@ func (p *SharePool) sharesForTokens(amount *Quantity) (*Quantity, error) {
 	return q, nil
 }
 
+// Deposit moves tokens into the combined balance, raising the shares.
+// If an error occurs, the pool and affected accounts are left in an invalid state.
 func (p *SharePool) Deposit(shareDst, tokenSrc, tokenAmount *Quantity) error {
 	shares, err := p.sharesForTokens(tokenAmount)
 	if err != nil {
@@ -418,6 +420,8 @@ func (p *SharePool) tokensForShares(amount *Quantity) (*Quantity, error) {
 	return q, nil
 }
 
+// Withdraw moves tokens out of the combined balance, reducing the shares.
+// If an error occurs, the pool and affected accounts are left in an invalid state.
 func (p *SharePool) Withdraw(tokenDst, shareSrc, shareAmount *Quantity) error {
 	tokens, err := p.tokensForShares(shareAmount)
 	if err != nil {
--- a/go/staking/api/api.go
+++ b/go/staking/api/api.go
@@ -332,48 +332,69 @@ func MoveUpTo(dst, src, n *Quantity) (*Quantity, error) {
 	return amount, nil
 }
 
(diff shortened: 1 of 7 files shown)


go/staking/api/api.go
```diff
@@ -332,48 +332,69 @@ func MoveUpTo(dst, src, n *Quantity) (*Quantity, error) {
...
 
-	return issuedShares, nil
+	return q, nil
 }
 
-// TokensForShares computes the amount of tokens to be received for shares in the
-// given escrow account.
-func TokensForShares(acc *EscrowAccount, amount *Quantity) (*Quantity, error) {
-	if amount.IsZero() || acc.Balance.IsZero() || acc.TotalShares.IsZero() {
+func (p *SharePool) Deposit(shareDst, tokenSrc, tokenAmount *Quantity) error {
```
kostko: Documentation comment? Mention that in case of errors the pool can still be modified.

go/staking/api/api.go
```diff
@@ -384,19 +405,40 @@ func TokensForShares(acc *EscrowAccount, amount *Quantity) (*Quantity, error) {
...
 	//       lost due to loss of precision.
-	if err := q.Quo(&acc.TotalShares); err != nil {
+	if err := q.Quo(&p.TotalShares); err != nil {
 		return nil, err
 	}
 
 	return q, nil
 }
 
+func (p *SharePool) Withdraw(tokenDst, shareSrc, shareAmount *Quantity) error {
```
kostko: Documentation comment?

go/tendermint/apps/staking/staking.go
```diff
@@ -112,48 +112,26 @@ func (app *stakingApplication) onEpochChange(ctx *abci.Context, epoch epochtime.
 			escrow = state.Account(e.EscrowID)
 		}
 
-		// Compute amount of debonded tokens.
-		tokens, err := staking.TokensForShares(&escrow.Escrow, &deb.Shares)
-		if err != nil {
-			app.logger.Error("failed to compute amount of tokens from shares",
+		var tokens staking.Quantity
+		if err := escrow.Escrow.Debonding.Withdraw(&tokens, &deb.Shares, &deb.Shares); err != nil {
```
kostko: Is it ok if `Withdraw` is called with `srcShares` and `shareAmount` being the same pointer? Based on that function code it seems like there could be some unintended corruption.

pro-wh: ah shoot no

go/staking/api/api.go

go/staking/api/api.go

go/tendermint/apps/staking/staking.go
pro-wh: added Clone ):

pro-wh: oops, had some other instances. fixed those too

pro-wh: added tests in #2288

PR #2305, opened 2019-10-25
Use the optimized ed25519 batch verify where applicable

### LINKED ISSUE #1351: Implement a Ed25519 batch verify, use it in the RootHash
With Ed25519 it's possible to do batch verification which optimizes verifying a set of signatures vs a set of public keys all at once.  As our committee sizes grow (and we need to start validating SGX commits), it's probably worth thinking about utilizing this technique to reduce the time spent validating commits.

`golang.org/x/crypto/ed25519` last I checked is still agl's Go port of djb's `ref10`, but he left out the batch verification implementation.  Porting the rest of it over should be a straight forward drop-in process.

If signature verification/signing ends up being a perf bottleneck, it's possible to implement all of these much faster than, `ref10` as well, though doing it without massive pain will require Go 1.12 (for `math/bits` integer intrinsics).

Yawning:
Not to be content with the ~2x perf increase we already have over using the runtime library's verify, we can do better since our ed25519 implementation has a Bos-Coster based batch verify.

 * [x] Add the required API(s).
 * [x] Use the new API(s) for better performance.
   * [x] Storage receipts.
   * [x] Merge commitments.

Fixes #1351.

### DIFF
--- a/go/common/crypto/signature/signature.go
+++ b/go/common/crypto/signature/signature.go
@@ -3,6 +3,7 @@ package signature
 
 import (
 	"bytes"
+	"crypto/rand"
 	"encoding"
 	"encoding/base64"
 	"encoding/hex"
@@ -64,6 +65,8 @@ var (
 
 	testPublicKeys        sync.Map
 	blacklistedPublicKeys sync.Map
+
+	defaultOptions = &ed25519.Options{}
 )
 
 // MapKey is a PublicKey as a fixed sized byte array for use as a map key.
@@ -118,7 +121,7 @@ func (k PublicKey) Verify(context, message, sig []byte) bool {
 	if len(sig) != SignatureSize {
 		return false
 	}
-	if _, isBlacklisted := blacklistedPublicKeys.Load(k.ToMapKey()); isBlacklisted {
+	if k.isBlacklisted() {
 		return false
 	}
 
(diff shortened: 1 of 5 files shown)


go/common/crypto/signature/signature.go
```diff
@@ -506,6 +514,77 @@ func (s *SignedPublicKey) Open(context []byte, pub *PublicKey) error { // nolint
 	return s.Signed.Open(context, pub)
 }
 
+// VerifyManyToOne verifies multiple signatures against a single context and
+// message, returning true iff every signature is valid.
+func VerifyManyToOne(context []byte, message []byte, sigs []Signature) bool {
+	// Our batch verify supports doing Ed25519ph/Ed25519ctx in bulk,
+	// but we're stuck with this stupidity.
```
kostko: :grin:

go/tendermint/abci/mux.go
```diff
@@ -331,8 +331,12 @@ func (mux *abciMux) InitChain(req types.RequestInitChain) types.ResponseInitChai
 
 	// HACK: Can't emit tags from InitChain, stash the genesis state
 	// so that processing can happen in BeginBlock.
+	//
+	// This is also stashed in the CheckTx tree (even though it will get)
```
kostko:
```suggestion
	// This is also stashed in the CheckTx tree (even though it will get
```

go/tendermint/abci/mux.go
```diff
@@ -331,8 +331,12 @@ func (mux *abciMux) InitChain(req types.RequestInitChain) types.ResponseInitChai
...
+	// rolled back, so that it is usable from the genesisHooks.
```
kostko:
```suggestion
	// rolled back), so that it is usable from the genesisHooks.
```

Issue #2057, opened 2019-09-04
go: Convert to go 1.13 error semantics


Yawning:
* `fmt.Errorf("... %w ...", err)` instead of `pkg/errors.Wrap`
 * `if errors.Is(err, os.ErrExist)` instead of `if err = os.ErrExist` (use package method for error comparisons).

This should let us get rid of the `github.com/pkg/errors` import.

Yawning:
Fixed by #2912

PR #2309, opened 2019-10-28
consensus halt support

ptrus:
Fixes: https://github.com/oasislabs/oasis-core/issues/1860

Adds halt support to the tm mux app. Halt is implemented as an epoch in which all transactions are rejected and begin/end block are no-ops. After the halt epoch, the tm is stopped by triggering a panic.
Timesource is also removed from all backends and is now part of mux ApplicationState.

Why epoch height instead of block height:
- adds some buffer time so all nodes can catch up (since panicking will terminate the nodes)

Why hard panic (not only "read-only" mode):
- if there would be no hard panic, exported genesis would differ depending on when would it be exported (e.g. rounds and therefore epochs would keep increasing)

HaltEpoch height can currently only be specified via the genesis. We should provide a debug command on the node to override this height. This would enable operators
to coordinate and change HaltEpoch in case the halt should be postponed/forwarded (this is needed until we support updates of genesis parameters).

TODO:
- [x] dump genesis when reaching halt epoch
- [x] ~consider doing  #1879 in this PR~
- [x] ~debug command to override halt epoch~ (probably a separate PR makes more sense, https://github.com/oasislabs/oasis-core/issues/2327)
- [x] fix byzantine tests
- [x] cleanup the actual halt
- [x] add tests


go/tendermint/tendermint.go
```diff
@@ -408,17 +412,45 @@ func (t *tendermintService) GetGenesis() *genesis.Document {
...
+	}
+	t.timeSource = timeSource
+
+	// XXX: with https://github.com/oasislabs/oasis-core/issues/1879 initialize
+	// other backends here as well.
+
+	return timeSource, nil
+}
+
+func (t *tendermintService) ForceInitialize() error {
```
Yawning: If you cave the part that actually does the lazy init into a separate function, you can get rid of the kludges around locking.  Eg:
```
  doLazyInit := func() error {
    t.Lock()
    defer t.Unlock()

    // Do stuff.
  }
  if err := doLazyInit(); err != nil {
    return err
  }

  // Start consensus backends.
}
```

go/tendermint/abci/mux.go
```diff
@@ -363,13 +363,47 @@ func (mux *abciMux) InitChain(req types.RequestInitChain) types.ResponseInitChai
...
 	)
 
+	if mux.state.InHaltEpoch() {
+		mux.logger.Info("BeginBlock: in halt epoch, emitting empty block.")
+		return types.ResponseBeginBlock{}
+	}
+
+	if mux.state.AfterHaltEpoch() {
+		mux.logger.Info("BeginBlock: halt epoch over, trigginring a halt")
+		// XXX: Ideally there would be a nicer way to do this in tendermint.
```
Yawning: Shouldn't this trigger a graceful shutdown somehow?  We have code that brings the node down cleanly on SIGTERM/SIGINT...

go/tendermint/abci/mux.go
```diff
@@ -494,6 +535,13 @@ func (mux *abciMux) CheckTx(req types.RequestCheckTx) types.ResponseCheckTx {
 }
 
 func (mux *abciMux) DeliverTx(req types.RequestDeliverTx) types.ResponseDeliverTx {
+	if mux.state.InHaltEpoch() {
+		mux.logger.Info("CheckTx: in halt epoch, rejecting all transactions")
```
Yawning: `DeliverTx:`

go/tendermint/abci/mux.go
```diff
@@ -363,13 +363,47 @@ func (mux *abciMux) InitChain(req types.RequestInitChain) types.ResponseInitChai
...
+
 func (mux *abciMux) BeginBlock(req types.RequestBeginBlock) types.ResponseBeginBlock {
 	blockHeight := mux.state.BlockHeight()
 	mux.logger.Debug("BeginBlock",
 		"req", req,
 		"block_height", blockHeight,
 	)
 
+	if mux.state.InHaltEpoch() {
+		mux.logger.Info("BeginBlock: in halt epoch, emitting empty block.")
```
Yawning: There can be lots of blocks in the halt epoch, this will be ridiculously spammy and likely not useful.  It should log once and only once, on the transition.

I would like to see the block number and epoch logged as well, if we are doing this.

go/tendermint/tendermint.go
ptrus: :+1:

ptrus:
Thx for the early review! 

> On the transition into the halt epoch, each node should probably dump state to a known location with a known filename dump-<chainid>-<epoch>-<blockheight>.json, just so there is always a checkpoint available since you are opting to halt the node in a state where it will be impossible to generate a dump.

Yeah that makes sense and something like this was planned (but i missed it in the todo's)

go/tendermint/abci/mux.go
ptrus: makes sense :+1: will switch to only logging it once

go/genesis/api/api.go
```diff
@@ -50,6 +56,25 @@ func (d *Document) UnmarshalCBOR(data []byte) error {
...
+		return err
+	}
+
+	w, err := os.Create(path)
+	if err != nil {
+		return errors.Wrap(err, "WriteFileJSON: failed to get writer for genesis file")
+	}
+	defer w.Close()
+
+	if _, err = w.Write(docJSON); err != nil {
```
Yawning: https://golang.org/pkg/io/ioutil/#WriteFile

ptrus:
I think this should be good for another review @kostko @Yawning in case there are any design changes needed before i finalize the remaining TODOs

go/common/consensus/consensus.go
```diff
@@ -3,12 +3,21 @@
...
+	epochtime "github.com/oasislabs/oasis-core/go/epochtime/api"
 )
 
 // Backend is an interface that a consensus backend must provide.
 type Backend interface {
+	// Initialize initializes the consensus backends
+	// XXX: as of now, this only initializes the epochtime backend, with
+	// https://github.com/oasislabs/oasis-core/issues/1879 this should
+	// initialize all backends.
+	InitializeBackends() (epochtime.Backend, error)
```
kostko: Is there a reason why this wouldn't happen as part of the consensus backend constructor?

The only thing that you need to expose then is a way to get the various backends, e.g.:
```golang
EpochTime() epochtime.Backend
```

go/epochtime/api/api.go
```diff
@@ -36,6 +36,9 @@ type Backend interface {
 
 	// ToGenesis returns the genesis state at the specified block height.
 	ToGenesis(ctx context.Context, height int64) (*Genesis, error)
+
+	// Start starts the epochtime backend.
+	Start(ctx context.Context) error
```
kostko: Why is this needed?

go/epochtime/init.go
```diff
@@ -26,14 +25,14 @@ const (
...
 
 // New constructs a new Backend based on the configuration flags.
-func New(ctx context.Context, tmService service.TendermintService) (api.Backend, error) {
+func New(tmService service.TendermintService) (api.Backend, error) {
 	backend := viper.GetString(CfgBackend)
 	switch strings.ToLower(backend) {
 	case tendermint.BackendName:
 		interval := viper.GetInt64(CfgTendermintInterval)
-		return tendermint.New(ctx, tmService, interval)
+		return tendermint.New(tmService, interval)
```
kostko: This seems awkward if you also provide a way for the consensus backend to initialize epochtime. Maybe move initialization of an epochtime backend to consensus in advance of #1879?

go/epochtime/tendermint/tendermint.go
```diff
@@ -131,16 +129,29 @@ func New(ctx context.Context, service service.TendermintService, interval int64)
...
-	r.notifier = pubsub.NewBrokerEx(func(ch *channels.InfiniteChannel) {
-		r.RLock()
-		defer r.RUnlock()
 
-		if r.lastNotified == r.epoch {
-			ch.In() <- r.epoch
+	return r, nil
+}
+
+func (t *tendermintBackend) Start(ctx context.Context) error {
```
kostko: This seems very awkward, why not initialize in `New` (see my other comment about moving tendermint-specific epochtime backend under `go/tendermint` in advance of #1879)?

go/genesis/api/api.go
```diff
@@ -50,6 +58,19 @@ func (d *Document) UnmarshalCBOR(data []byte) error {
...
 
+// WriteFileJSON writes the genesis document into a JSON file.
+func (d *Document) WriteFileJSON(filename string) error {
+	docJSON, err := json.Marshal(d)
+	if err != nil {
+		return err
+	}
+
+	if err = ioutil.WriteFile(filename, docJSON, filePerm); err != nil {
+		return errors.Wrap(err, "WriteFileJSON: failed to write genesis file")
```
kostko: For new code no need to use the external errors package (see #2057):
```suggestion
		return fmt.Errorf("WriteFileJSON: failed to write genesis file: %w", err)
```

go/genesis/file.go
```diff
@@ -19,8 +22,9 @@ func (p *fileProvider) GetGenesisDocument() (*api.Document, error) {
 }
 
 // NewFileProvider creates a new local file genesis provider.
-func NewFileProvider(filename string) (api.Provider, error) {
-	logger := logging.GetLogger("genesis/file")
+func NewFileProvider() (api.Provider, error) {
+	filename := flags.GenesisFile()
+	logger := logging.GetLogger("genesis/file").With("filename", filename)
```
kostko: You set `filename` here, but also again below?

go/genesis/tendermint.go
```diff
@@ -0,0 +1,146 @@
+package genesis
```
kostko: Uh, this doesn't seem nice, can we move this Tendermint-specific thing somewhere in `go/tendermint`? Also it seems strange that this is a genesis provider -- as providers are meant for loading genesis documents but this seems like it is used for dumping one?

Maybe instead, as I mentioned in my earlier comment, expose a method in `consensus.Backend` (e.g., `GetGenesis(int64)` and implement it in `go/tendermint` instead.

go/genesis/grpc.go
```diff
@@ -34,15 +32,9 @@ type grpcServer struct {
...
-	if height <= 0 {
-		var err error
-		if height, err = s.tendermintService.GetHeight(); err != nil {
-			return nil, err
-		}
-	}
 
-	// Get genesis doc.
-	genesisProvider, err := New()
+	// Get tendermint state as a genesis doc.
```
kostko: Since we are changing this, can we make it consensus backend-agnostic? E.g., make `consensus.Backend` provide some method that generates the genesis document and call it here instead of hardcoding "tendermint"?

go/tendermint/abci/mux.go
```diff
@@ -731,17 +818,31 @@ func (s *ApplicationState) CheckTxTree() *iavl.MutableTree {
...
-func (s *ApplicationState) EpochChanged(timeSource epochtime.Backend) (bool, epochtime.EpochTime) {
+func (s *ApplicationState) EpochChanged() (bool, epochtime.EpochTime) {
 	blockHeight := s.BlockHeight()
 	if blockHeight == 0 {
 		return false, epochtime.EpochInvalid
 	} else if blockHeight == 1 {
 		// There is no block before the first block. For historic reasons, this is defined as not
 		// having had a transition.
-		currentEpoch, err := timeSource.GetEpoch(s.ctx, blockHeight)
+		currentEpoch, err := s.timeSource.GetEpoch(s.ctx, blockHeight)
```
kostko: Aw, this will conflict with #2290 :-)

ptrus: :grin:

go/epochtime/tendermint/tendermint.go
ptrus: I needed these since currently `New()` needed a reference to tendermint service, and the tendermint service needed a reference to the timesource.

Agreed it will be nicer if i move this into `go/tendermint` and modify it a bit to avoid having this additional method  :+1:

go/genesis/grpc.go
ptrus: will do :+1:

go/genesis/tendermint.go
ptrus: makes sense :+1:

go/common/consensus/consensus.go
kostko: Yeah I think it would be better to initialize all backends in `New` and then the provided method would be to just return the initialized backend(s).

go/genesis/grpc.go
ptrus: I moved it into tendermint/genesis/genesis.go for now and not yet part of the `consensus.Backend`. Added a comment on moving it into `consensus.Backend` interface with #1879

go/oasis-node/cmd/debug/byzantine/tendermint.go
```diff
@@ -26,69 +23,12 @@ import (
 	"github.com/oasislabs/oasis-core/go/tendermint/service"
 )
 
-var _ epochtime.Backend = (*fakeTimeBackend)(nil)
```
ptrus: I'm not entirely sure why this was needed before and if it was indeed save to remove now with the refactor, but the byzantine tests seem to pass.

ptrus:
ready for a re-review. Still a couple of TODOs left (see description) and will also wait that https://github.com/oasislabs/oasis-core/pull/2294 get's merged (as will enable some more cleanup)

go/tendermint/abci/mux.go
```diff
@@ -200,19 +194,29 @@ func (a *ApplicationServer) RegisterGenesisHook(hook func()) {
...
+	a.mux.registerHaltHook(hook)
+}
+
 // Pruner returns the ABCI state pruner.
 func (a *ApplicationServer) Pruner() StatePruner {
 	return a.mux.state.statePruner
 }
 
+func (a *ApplicationServer) SetEpochtime(epochTime epochtime.Backend) {
+	a.mux.state.timeSource = epochTime
```
kostko: Maybe add a comment somewhere that this must be set before the mux can be used? Also add a documentation comment for this method.

go/oasis-node/cmd/node/node.go
```diff
@@ -172,6 +173,19 @@ func (n *Node) initBackends() error {
 	return nil
 }
 
+func (n *Node) dumpGenesis(ctx context.Context, blockHeight int64) error {
+	doc, err := tmgenesis.ToGenesis(ctx, blockHeight, n.svcTmnt, n.Epochtime, n.KeyManager, n.Registry, n.RootHash, n.Staking)
+	if err != nil {
+		return err
+	}
+	dumpPath := filepath.Join(cmdCommon.DataDir(), fmt.Sprintf("genesis-dump-at-%d.json", blockHeight))
```
kostko: Maybe include chain ID and epoch besides block height?

go/tendermint/abci/mux.go
```diff
@@ -379,6 +423,34 @@ func (mux *abciMux) BeginBlock(req types.RequestBeginBlock) types.ResponseBeginB
...
+	}
+
+	if mux.state.AfterHaltEpoch(ctx) {
+		mux.logger.Info("BeginBlock: after halt epoch, halting",
+			"block_height", blockHeight,
+		)
+		// TODO: rather gracefully shutdown the node here, e.g.:
+		//  - add haltHook that does: 'cmd/node/node.go.Stop()'
+		//  - rename the existing halt hooks (to 'pre-halt' hooks)
+		panic("halt")
```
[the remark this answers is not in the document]
ptrus: :+1: will leave it as a panic and a comment

go/oasis-node/cmd/debug/byzantine/tendermint.go
pro-wh: this is a typecheck that fakeTimeBackend implements epochtime.Backend.

ptrus: > this is a typecheck that *fakeTimeBackend implements epochtime.Backend.

:grin: i removed the whole `fakeTimeBackend` and was wondering what's the need for it (not the line specifically)

go/common/consensus/genesis/genesis.go
```diff
@@ -0,0 +1,12 @@
+// Package consensus provides consensus config flags that should be part of the genesis state.
+package consensus
```
kostko: So this has been moved to `consensus/genesis` so it should be `genesis` now probably? Is the reason for the move some issue with circular imports?

go/genesis/grpc.go
```diff
@@ -35,72 +35,19 @@ type grpcServer struct {
...
-	if err != nil {
-		return nil, err
-	}
-	registryGenesis, err := s.registryBackend.ToGenesis(ctx, height)
-	if err != nil {
-		return nil, err
-	}
-	roothashGenesis, err := s.roothashBackend.ToGenesis(ctx, height)
+	// Get tendermint state as a genesis doc.
+	genesisDoc, err := s.tendermintService.ToGenesis(ctx, height, s.keymanagerBackend, s.registryBackend, s.roothashBackend, s.stakingBackend, s.schedulerBackend)
```
kostko: Once #1879 is done we should also change the grpc server to take `consensus.Backend` instead of `service.TendermintService` specifically? Or maybe we can do this already?

go/oasis-node/cmd/debug/byzantine/tendermint.go
```diff
@@ -9,12 +9,10 @@ import (
 	tmtypes "github.com/tendermint/tendermint/types"
 
 	"github.com/oasislabs/oasis-core/go/common/identity"
-	"github.com/oasislabs/oasis-core/go/common/pubsub"
-	epochtime "github.com/oasislabs/oasis-core/go/epochtime/api"
-	"github.com/oasislabs/oasis-core/go/genesis"
+
```
kostko: Remove empty line.

go/oasis-node/cmd/debug/byzantine/tendermint.go
```diff
@@ -23,123 +21,51 @@ import (
...
 	if err != nil {
 		return errors.Wrap(err, "tendermint New")
 	}
 
-	if err = ht.service.ForceInitialize(); err != nil {
-		return errors.Wrap(err, "honest Tendermint service ForceInitialize")
-	}
-
 	// Register honest mux apps.
 	// This isn't very flexible. It's configured to match what we use in end-to-end tests.
```
kostko: Should this comment be updated now that options come from genesis?

go/oasis-node/cmd/debug/byzantine/tendermint.go
```diff
@@ -148,7 +74,7 @@ func (ht *honestTendermint) start(id *identity.Identity, dataDir string, useMock
 	}
 	ht.schedulerQuery = schedApp.QueryFactory().(*schedulerapp.QueryFactory)
 	// storage has no registration
-	roothashApp := roothashapp.New(timeSource, nil, 10*time.Second)
+	roothashApp := roothashapp.New(nil, 10*time.Second)
```
kostko: Hm, the round timeout is not backed by consensus? I know this is not part of this PR but can you check and remove, I remember there being a `RoundTimeout` in the roothash genesis doc, but it seems it doesn't do anything?

go/oasis-test-runner/scenario/e2e/halt_restore.go
```diff
@@ -0,0 +1,162 @@
...
+		return err
+	}
+
+	logger.Info("triggering epoch transition to halt epoch")
+	if err = sc.net.Controller().SetEpoch(ctx, 2); err != nil {
+		return fmt.Errorf("scenario/e2e/halt_restore: setting halt epoch: %w", err)
+	}
+	logger.Info("halt epoch transition done")
+
+	// TODO: add a 'wait-epoch' command, switch to epochtime backend, and wait for 3rd epoch here.
```
kostko: Either a command or a method on the `sc.net.Controller()`?

But wait, if this is the mock epochtime backend, there will be no 3rd epoch unless you call `SetEpoch` (which already waits for the epoch transition to happen)?

go/oasis-test-runner/scenario/e2e/halt_restore.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	if len(files) == 0 {
+		return fmt.Errorf("scenario/e2e/halt_restore: genesis file not found in: %s", dumpGlobPath)
+	}
+	if len(files) > 1 {
+		return fmt.Errorf("scenario/e2e/halt_restore: more than one genesis file found in: %s", dumpGlobPath)
+	}
+	dumpPath := files[0]
+
+	// XXX: Assert all exported genesis files match
```
kostko: Will this be done in this PR?

go/tendermint/abci/mux.go
```diff
@@ -200,19 +194,31 @@ func (a *ApplicationServer) RegisterGenesisHook(hook func()) {
...
+	a.mux.registerHaltHook(hook)
+}
+
 // Pruner returns the ABCI state pruner.
 func (a *ApplicationServer) Pruner() StatePruner {
 	return a.mux.state.statePruner
 }
 
+// SetEpochtime sets the mux epochtime.
+// XXX: epochtime needs to be set before mux can be used. // TODO: add check for this
```
kostko: Will this be done in this PR?

go/tendermint/abci/mux.go
```diff
@@ -379,6 +423,36 @@ func (mux *abciMux) BeginBlock(req types.RequestBeginBlock) types.ResponseBeginB
...
+
+			mux.logger.Debug("Dispatching halt hooks")
+			for _, hook := range mux.haltHooks {
+				hook(mux.state.ctx, blockHeight, mux.state.haltEpochHeight)
+			}
+			mux.logger.Debug("Halt hook dispatch complete")
+		}
+	}
+
+	if mux.state.haltMode {
```
kostko: Maybe use something like:
```golang
switch mux.state.haltMode {
case false:
    if !mux.state.inHaltEpoch(ctx) {
        break
    }
    // ...
case true:
    if !mux.state.afterHaltEpoch(ctx) {
        break
    }
    // ...
}
```

go/tendermint/abci/mux.go
```diff
@@ -379,6 +423,36 @@ func (mux *abciMux) BeginBlock(req types.RequestBeginBlock) types.ResponseBeginB
...
+		if mux.state.afterHaltEpoch(ctx) {
+			mux.logger.Info("BeginBlock: after halt epoch, halting",
+				"block_height", blockHeight,
+			)
+			// XXX: there is no way to stop tendermint consensus other than triggering
+			// a panic atm. Once possible, we should stop the consensus layer here and
+			// gracefully shutdown the node.
+			panic("haltting")
```
kostko:
```suggestion
			panic("tendermint: after halt epoch, halting")
```

go/tendermint/abci/mux.go
```diff
@@ -481,6 +555,13 @@ func (mux *abciMux) executeTx(ctx *Context, tx []byte) error {
 func (mux *abciMux) CheckTx(req types.RequestCheckTx) types.ResponseCheckTx {
 	ctx := NewContext(ContextCheckTx, mux.currentTime, mux.state)
 
+	if mux.state.haltMode {
```
kostko: This could be done in `executeTx` as that is called for both `CheckTx` and `DeliverTx`.

go/tendermint/abci/mux.go
```diff
@@ -516,6 +604,11 @@ func (mux *abciMux) EndBlock(req types.RequestEndBlock) types.ResponseEndBlock {
 		"block_height", mux.state.BlockHeight(),
 	)
 
+	if mux.state.haltMode {
+		mux.logger.Debug("EndBlock: in halt, emitting empty block.")
```
kostko:
```suggestion
		mux.logger.Debug("EndBlock: in halt, emitting empty block")
```

go/tendermint/abci/mux.go
ptrus: :+1:

go/genesis/grpc.go
ptrus: good idea, can already do it yes

go/common/consensus/genesis/genesis.go
ptrus: yes and yes :)

go/oasis-node/cmd/debug/byzantine/tendermint.go
ptrus: yeah can probably just remove the whole comment? (keeping the `// Register honest mux apps.`)

go/oasis-node/cmd/debug/byzantine/tendermint.go
ptrus: oh yeah it's wrong at the moment, some places use `RoundTimeout` from genesis, but the mux app still uses the passed in timeout :grin: will fix it

go/oasis-node/cmd/debug/byzantine/tendermint.go
kostko: :+1:

go/oasis-node/cmd/debug/byzantine/tendermint.go
ptrus: fixed in separate pr: https://github.com/oasislabs/oasis-core/pull/2333 (it wasn't actually wrong, since in all cases (but here) the param form genesis was taken, just an unnecessary param)

go/oasis-test-runner/scenario/e2e/halt_restore.go
ptrus: yeah with a method on the `sc.net.Controller()` i'd switch to the non-mock backend

go/oasis-node/cmd/node/node.go
```diff
@@ -507,14 +530,32 @@ func newNode(testNode bool) (*Node, error) {
 			return nil, err
 		}
 		node.svcMgr.Register(node.svcTmnt)
-
+		node.Epochtime = node.svcTmnt.EpochTime()
```
kostko: Add new line after this line.

go/oasis-test-runner/scenario/e2e/halt_restore.go
kostko: But why not just keep the mock backend and call `SetEpoch`? It should behave the same as if you wait for the epoch transition in a non-mock backend.

The only problem currently is that this will cause the nodes to terminate, so you need to make sure this is allowed in this test (also see #2296 which adds support for allowing node shutdown in tests).

ptrus: once we're in the "halt" epoch we aren't able to switch epochs anymore with the mock backend (since consensus is in read-only mode). However with the non mock backend the epoch will change since blocks continue to happen.

kostko: Oh, oops, yeah that's true :-) So you'll add the RPC method in this PR? Could be part of the client service (`go/client`).

ptrus: yeah i'll do it :+1:

ptrus: Updated the test, using the normal epochtime now. 

Cannot actually wait for the epoch after the halt epoch though, as the block in which the consensus halts is never produced (which is the correct behavior) - so I currently ignore the the error in the test and it actually works more like a "wait for shutdown" command. See the comment in the code.

Also the panic in the consensus doesn't actually shutdown the node at the moment (i think this is fixed in https://github.com/oasislabs/oasis-core/pull/2296) so updating the test to allow node shutdowns should be done after either this or #2296 PR is merged.

ptrus: yeah should just be changed to wait for node termination (similarly as it's done in #2296) when these branches are merged

go/oasis-test-runner/scenario/e2e/halt_restore.go
```diff
@@ -0,0 +1,205 @@
...
+	ctx := context.Background()
+	sc.logger.Info("waiting for halt epoch")
+	// Wait for halt epoch.
+	err = sc.net.Controller().WaitEpoch(ctx, haltEpoch)
+	if err != nil {
+		return fmt.Errorf("scenario/e2e/halt_restore: failed waiting for halt epoch: %w", err)
+	}
+
+	// Wait additional few seconds for genesis docs to be dumped.
+	// XXX: fixed in #2296 where we wait for nodes to actually shutdown.
```
ptrus: Note: added a temporary sleep workaround since waiting for the halt epoch was a bit flaky on the CI (one validator would panic, and the others were stuck before reaching the last epoch due to consensus timeout - and if the Controller was the stuck validator, the test would timeout). Will fix it in #2296 by actually waiting on node shutdown.

PR #2313, opened 2019-10-30
Miscellaneous improvements

tjanez:
I think it will be easier to review these separately, so I extracted them from my #2300 PR.

### DIFF
--- a/go/oasis-test-runner/oasis/validator.go
+++ b/go/oasis-test-runner/oasis/validator.go
@@ -66,7 +66,8 @@ func (val *Validator) startNode() error {
 		tendermintCoreListenAddress(val.consensusPort).
 		grpcDebugPort(val.grpcDebugPort).
 		storageBackend("client").
-		appendNetwork(val.net)
+		appendNetwork(val.net).
+		appendEntity(val.entity)
 
 	if _, err := val.net.startOasisNode(val.dir, nil, args, "validator", false, false); err != nil {
 		return errors.Wrap(err, "oasis/validator: failed to launch node")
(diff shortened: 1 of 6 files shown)


go/worker/common/worker.go
```diff
@@ -339,7 +346,7 @@ func New(
 ) (*Worker, error) {
 	cfg, err := newConfig()
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "worker/common: failed to initialize config")
```
kostko: For new code, no need to use the external errors package (see #2057):
```suggestion
		return nil, fmt.Errorf("worker/common: failed to initialize config: %w", err)
```

go/oasis-test-runner/oasis/validator.go
```diff
@@ -99,7 +101,7 @@ func (net *Network) NewValidator(cfg *ValidatorCfg) (*Validator, error) {
 		"registry", "node", "init",
 		"--datadir", val.dir.String(),
 		"--node.consensus_address", fmt.Sprintf("127.0.0.1:%d", val.consensusPort),
-		"--node.expiration", "1000000",
+		"--node.expiration", "1",
```
Yawning: At one point, this would break horribly if it wasn't at least "2", but that might have been fixed.

tjanez: Yes, @kostko, told me that it should be `1` or `2`.

I tried with `1` and the tests passed :slightly_smiling_face:.

kostko: Validators expiring seems really dangerous though. I think the scheduler will currently even allow having only a single validator which could then do arbitrary stuff?

Yawning: > I think the scheduler will currently even allow having only a single validator which could then do arbitrary stuff?

Correct.  However:
 * We have not established the minimum number of validators that we want (and it is a policy decision that I am not qualified to make).  The correct thing to do is for someone to come up with validator committee minimum parameters, that we enforce.
 * Even with such parameters, the only safe behavior in the event that we do not have a sufficient node pool to elect a validator committee is to halt.

kostko: Agreed, https://github.com/oasislabs/oasis-core/issues/2314.

PR #2320, opened 2019-10-31
go/registry: Add stricter node registration checks

kostko:


go/common/crypto/signature/signature.go
```diff
@@ -215,6 +215,11 @@ func (k PublicKey) String() string {
 	return hexKey
 }
 
+// IsValid checks whether the public key is well-formed.
+func (k PublicKey) IsValid() bool {
+	return len(k) == PublicKeySize
```
Yawning: Probably can also check for small order points here, and reject them, though eventually both the go runtime library and our optimized version will do so.

https://github.com/jedisct1/libsodium/blob/master/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10.c#L1019

Yawning: Though as I think about this, we need to stop using `signature.PublicKey` as shorthand for a 32 byte opaque identifier first.  There are certain things that will never be valid Ed25519 public keys, that are valid identifiers after all.

kostko:
So the added `IsValid` check is used for things that actually are public keys (consensus and P2P public keys), not for runtime IDs.

Otherwise I agree that we should change the runtime ID's type to something like `common.Namespace`, see https://github.com/oasislabs/oasis-core/issues/1693#issuecomment-506937420.

go/common/crypto/signature/signature.go
kostko: Added blacklist check and blacklisted the small order points from the above link.

PR #2323, opened 2019-11-01
(no merge) checking control unit test coverage

pro-wh:
to compare against #2321


pro-wh:
all good here :ok_hand:

Issue #2022, opened 2019-08-21
Enforce "sensible" limits on adversary controlled descriptor fields


Yawning:
We currently allow unlimited quantities of certain things, that are attacker controlled.  To prevent an explosion in the BFT state (and increased Tx processing time), all of these things need to be clamped.

Eg:
 * `Entity.Nodes`
 * `Node.CommitteeInfo.Addresses` (and the other address fields).
 * `Node.CommitteeInfo.Certificate` (Ok, because the parsing will fail, but listed as an example of a potential hazard.)

Exempt:
 * The runtime descriptor (not adversary controlled).

*Estimated cost: <1 sprint*

Yawning:
I think a sensible general purpose band-aid solution for this for now is to enforce maximum sizes on the various descriptors.

kostko:
So #2326 clamped this at the consensus transaction level, we should probably also have explicit limits in our higher-level checks?

> The runtime descriptor (not adversary controlled).

It may be in the future when runtime registrations are allowed for anyone with enough stake.

PR #2326, opened 2019-11-04
go/tendermint/abci: Enforce a maximum transaction size limit

Yawning:
To avoid trivial DoS attacks against the consensus layer, enforce an
upper bound on the maximum transaction size (default: 32 KiB).  At a
future date, this limit can either be made more fine-grained, or the
limit can be adjusted as required.

Inspired by #2022.

### DIFF
--- a/go/oasis-node/cmd/genesis/genesis.go
+++ b/go/oasis-node/cmd/genesis/genesis.go
@@ -74,6 +74,7 @@ const (
 	cfgConsensusTimeoutCommit      = "consensus.tendermint.timeout_commit"
 	cfgConsensusSkipTimeoutCommit  = "consensus.tendermint.skip_timeout_commit"
 	cfgConsensusEmptyBlockInterval = "consensus.tendermint.empty_block_interval"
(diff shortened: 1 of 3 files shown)


go/oasis-node/cmd/genesis/genesis.go
```diff
@@ -195,6 +196,7 @@ func doInitGenesis(cmd *cobra.Command, args []string) {
 		TimeoutCommit:      viper.GetDuration(cfgConsensusTimeoutCommit),
 		SkipTimeoutCommit:  viper.GetBool(cfgConsensusSkipTimeoutCommit),
 		EmptyBlockInterval: viper.GetDuration(cfgConsensusEmptyBlockInterval),
+		MaxTxSize:          viper.GetInt(cfgConsensusMaxTxSize),
```
kostko: Maybe use `viper.GetSizeInBytes`?

Yawning: Fixed.

Yawning:
Not sure what I was thinking when I named the branch that way.  Oh well.

PR #2331, opened 2019-11-05
[WIP] Ledger device support for CLI

david-yan:
*** DO NOT MERGE ***

Added support for signing transactions using a Ledger device. The header values will initially be those for the Ethereum app, until the external developer is done with the BOLOS app, and we can test on it.


go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &Signer{device}, nil
+}
+
+func (s *Signer) Public() signature.PublicKey {
+	derivationPath := accounts.DefaultRootDerivationPath
+	path := make([]byte, 1+4*len(derivationPath))
+	path[0] = byte(len(derivationPath))
```
eauge: `len` returns an int, should we check if `0 <= len(derivationPath) < 128`

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,85 @@
...
+	for i, component := range derivationPath {
+		binary.BigEndian.PutUint32(path[1+4*i:], component)
+	}
+
+	header := []byte{ETH_CLA, ETH_INS_DERIVE, 0, 0, byte(len(path))}
+	message := append(header, path...)
+
+	response, err := s.device.Exchange(message)
+	if err != nil {
+		return nil
```
eauge: return err

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	pkLength := int(response[0])
```
eauge: should we verify that response has the expected length?

go/common/crypto/signature/signers/ledger/ledger_signer.go
eauge: or actually if this cannot return an error, we should panic, one of the two. But we should not swallow errors

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,85 @@
...
+	message := append(header, payload...)
+	response, err := s.device.Exchange(message)
+	if err != nil {
+		return nil, err
+	}
+	return response, nil
+}
+
+func (s *Signer) ContextSign(context signature.Context, message []byte) ([]byte, error) {
+	return nil, nil
```
eauge: does this need to be implemented?

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (s *Signer) String() string {
+	return "Ledger Signer"
+}
+
+func (s *Signer) Reset() {
+	s.device.Close()
+}
+
+func (s *Signer) UnsafeBytes() []byte {
```
eauge: what is this suppose to return?

go/common/crypto/signature/signers/ledger/ledger_signer.go
Yawning: If the signer is intended to actually be useful, yes.

go/go.mod
```diff
@@ -21,9 +23,10 @@ require (
 	github.com/cznic/strutil v0.0.0-20181122101858-275e90344537 // indirect
 	github.com/dgraph-io/badger v2.0.0-rc.2.0.20190625224416-e0016993b790+incompatible
 	github.com/eapache/channels v1.1.0
-	github.com/eapache/queue v1.1.0 // indirect
 	github.com/edsrzf/mmap-go v1.0.0 // indirect
+	github.com/elastic/gosigar v0.10.5 // indirect
 	github.com/etcd-io/bbolt v1.3.3
+	github.com/ethereum/go-ethereum v1.9.6
```
Yawning: Ugh, that repo is huge.

go/common/crypto/signature/signers/ledger/ledger_signer.go
Yawning: The raw Ed25519 private key.  It's used for the per-node tendermint p2p link key.

Yawning: I'll get rid of the call as part of #2332.

go/go.mod
Yawning: Since this is temporary, instead of inflicting the pain of pulling in the entire repo on all of us, just use:
`var defaultRootDerivationPath = []uint32{0x80000000 + 44, 0x80000000 + 60, 0x80000000 + 0, 0}`

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,85 @@
...
+func NewSigner() (*Signer, error) {
+	device, err := ledger_go.FindLedger()
+	if err != nil {
+		return nil, err
+	}
+	return &Signer{device}, nil
+}
+
+func (s *Signer) Public() signature.PublicKey {
+	derivationPath := accounts.DefaultRootDerivationPath
```
Yawning: The derivation code is duplicated, refactor out into a function.

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,85 @@
+package ledger
+
+import (
+	"encoding/binary"
+
+	"github.com/ethereum/go-ethereum/accounts"
```
kostko: Why do we need to use anything from ethereum? Is the only reason for importing this the `DefaultRootDerivationPath`? Can't we use our own key derivation path? If we really need to use that, just copy the value here instead of importing the huge pile of geth.

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/oasislabs/ledger-go"
```
kostko: Please group imports as mentioned [in the style guide](https://github.com/oasislabs/oasis-core/blob/master/CONTRIBUTING.md#go-style-guide).

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
+)
+
+const (
+	ETH_CLA        byte = 0xE0
```
kostko: Add some comments what these magic constants are. Is this another Ethereum thing (based on `ETH_` prefix)?

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,85 @@
...
+	ETH_INS_CONFIG byte = 0x06
+	ETH_INS_DERIVE byte = 0x02
+	ETH_INS_SIGN   byte = 0x04
+)
+
+type Signer struct {
```
kostko: Add documentation comments to public types and functions.

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,85 @@
...
+	path := make([]byte, 1+4*len(derivationPath))
+	path[0] = byte(len(derivationPath))
+	for i, component := range derivationPath {
+		binary.BigEndian.PutUint32(path[1+4*i:], component)
+	}
+
+	header := []byte{ETH_CLA, ETH_INS_DERIVE, 0, 0, byte(len(path))}
+	message := append(header, path...)
+
+	response, err := s.device.Exchange(message)
```
kostko: Is the public key expected to change between invocations? Otherwise the signer should cache this instead of talking to the device each time a public key is requested.

go/common/crypto/signature/signers/ledger/ledger_signer.go
kostko: Oh I see this is temporary, ok.

go/common/crypto/signature/signers/ledger/ledger_signer.go
kostko: All our signatures use contexts, so this is required see the existing file-based signer implementations.

kostko:
Next time use draft PRs for things that should not be merged ;-)

go/common/crypto/signature/signers/ledger/ledger_signer.go
Yawning: This no longer exists in the base `Signer` interface in master.

go/common/crypto/signature/signers/ledger/ledger_signer.go
david-yan: Yes, this is something that I'll need to communicate to the external developer for our Ledger app. Currently this is left empty because there is no such instruction in the Ethereum Ledger app.

go/common/crypto/signature/signers/ledger/ledger_signer.go
david-yan: Great! That's good to know.

go/go.mod
david-yan: Will do. This wasn't intended to be merged into master, so no one else should have had to build the entire repo.

david-yan:
@kostko I didn't realize that draft PRs were a thing, but that would have definitely been appropriate here 😀. Thanks!

I'm going to close this PR for now, and replace it with a draft PR.

PR #2333, opened 2019-11-05
go/roothash: use RoundTimeout from genesis

ptrus:
https://github.com/oasislabs/oasis-core/pull/2309#discussion_r342196981

use RoundTimeout from genesis on all places

### DIFF
--- a/go/roothash/api/api.go
+++ b/go/roothash/api/api.go
@@ -83,6 +83,9 @@ type Backend interface {
 	// ComputeCommit submits a batch of compute commitments for slashing.
 	ComputeCommit(context.Context, signature.PublicKey, []commitment.ComputeCommitment) error
 
+	// ConsensusParameters returns consensus parameters at specified block height.
+	ConsensusParameters(context.Context, int64) (*ConsensusParameters, error)
+
 	// ToGenesis returns the genesis state at specified block height.
 	ToGenesis(context.Context, int64) (*Genesis, error)
 
--- a/go/roothash/api/api.go
+++ b/go/roothash/api/api.go
@@ -48,19 +48,8 @@ var (
 	_ cbor.Unmarshaler = (*MergeDiscrepancyDetectedEvent)(nil)
 )
 
-// Info contains information about a root hash backend.
-type Info struct {
-	// ComputeRoundTimeout is the compute round timeout.
-	ComputeRoundTimeout time.Duration
-	// MergeRoundTimeout is the merge round timeout.
-	MergeRoundTimeout time.Duration
-}
-
 // Backend is a root hash implementation.
 type Backend interface {
-	// Info returns information about a root hash backend.
-	Info() Info
-
 	// GetGenesisBlock returns the genesis block.
 	GetGenesisBlock(context.Context, signature.PublicKey, int64) (*block.Block, error)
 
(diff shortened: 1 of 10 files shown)


go/roothash/api/api.go
```diff
@@ -48,19 +48,8 @@ var (
 	_ cbor.Unmarshaler = (*MergeDiscrepancyDetectedEvent)(nil)
 )
 
-// Info contains information about a root hash backend.
-type Info struct {
```
ptrus: Removed this since this was only used in a single place https://github.com/oasislabs/oasis-core/pull/2333/files#diff-2b8a5495d9729944400cef9dece417edR361

and we can just query genesis there. Can revert this if we prefer it that way. (but in that case we might also want to move `commonWorker.GenesisDoc.RootHash.Parameters.TransactionScheduler` into the `Info` struct for consistency?)

go/worker/merge/committee/node.go
```diff
@@ -358,7 +358,7 @@ func (n *Node) tryFinalizeResultsLocked(pool *commitment.Pool, didTimeout bool)
 	// NOTE: The roothash backend will start counting its timeout on its own based on
 	//       any received commits so in the worst case the actual timeout will be
 	//       2*roundTimeout.
-	roundTimeout := n.commonNode.Roothash.Info().ComputeRoundTimeout
+	roundTimeout := n.commonNode.GenesisDoc.RootHash.Parameters.RoundTimeout
```
kostko: This should probably query the current value of the parameters at the current height instead of assuming they will stay the same since genesis.

E.g., maybe add a `ConsensusParameters(context.Context, int64)` to the roothash API (instead of the removed `Info`) and make that query the app? We should probably do the same in other places if we are using the genesis doc directly as the parameters may be mutable eventually.

ptrus: :+1: yeah that was my thinking as well (but then i noticed we just take the value from the doc directly in the txnscheduler case :grin:) will update

go/roothash/tendermint/tendermint.go
```diff
@@ -242,6 +232,16 @@ func (tb *tendermintBackend) ComputeCommit(ctx context.Context, id signature.Pub
 	return nil
 }
 
+func (tb *tendermintBackend) ConsensusParameters(ctx context.Context, height int64) (*api.ConsensusParameters, error) {
+	/// XXX: could cache per height.
```
ptrus: Since this wont be changing at the moment, should we just cache the first result?

kostko: Yeah we can cache it for now, just add a comment when this would need to change.

go/worker/merge/committee/node.go
```diff
@@ -358,7 +358,14 @@ func (n *Node) tryFinalizeResultsLocked(pool *commitment.Pool, didTimeout bool)
 	// NOTE: The roothash backend will start counting its timeout on its own based on
 	//       any received commits so in the worst case the actual timeout will be
 	//       2*roundTimeout.
-	roundTimeout := n.commonNode.Roothash.Info().ComputeRoundTimeout
+	params, err := n.commonNode.Roothash.ConsensusParameters(ctx, 0)
```
kostko: We should probably query at epoch height (e.g., `epoch.GetGroupVersion()` where `epoch` can be passed by the caller).

ptrus: moved it into Group `epoch` & `epochsnapshot`, and is now queried on epoch transition.

PR #2336, opened 2019-11-05
Add consensus gas fees

### LINKED ISSUE #2266: Gas fees for consensus txs
To mitigate DoS by transaction spam we need to introduce gas fees for consensus transactions that can be performed by non-staked entities (e.g., staking transfers, escrow, reclaim escrow).

All operations which can only be performed by staked entities are exempt.

*Estimated cost: ~1 sprint*

kostko:
Fixes #2266 

TODO
* [x] Tests for non-zero gas costs with different fees and out-of-gas conditions.
* [x] Add configuration of minimum gas price for validator nodes and reject transactions with lower implied gas price.
* [x] Distribute gas fees to signers of the block.


go/oasis-test-runner/scenario/e2e/gas_fees.go
```diff
@@ -0,0 +1,416 @@
...
+	}
+	_ = newTotalEntityBalance.Add(commonPool)
+
+	sc.logger.Info("making sure that fees have been disbursed",
+		"total_fees", totalFees,
+		"disbursed_fees", newTotalEntityBalance,
+		"common_pool", commonPool,
+	)
+
+	if newTotalEntityBalance.Cmp(&totalFees) != 0 {
```
Yawning: This should probably assert that the non-common pool disbursment is non-zero, on the off chance that something is horribly wrong and only the common pool is getting paid.

go/tendermint/apps/staking/state/gas.go
```diff
@@ -0,0 +1,92 @@
...
+// feeAccumulator is the per-block fee accumulator that gets all fees paid
+// in a block.
+type feeAccumulator struct {
+	balance quantity.Quantity
+}
+
+// AuthenticateAndPayFees authenticates the message signer and makes sure that
+// any gas fees are paid.
+//
+// This method modifies the common pool balance to transfer collected gas fees.
```
Yawning: common pool -> general pool?

go/tendermint/apps/staking/state/gas.go
```diff
@@ -0,0 +1,92 @@
...
+	if account.General.Nonce != nonce {
+		logger.Error("invalid account nonce",
+			"account_id", id,
+			"account_nonce", account.General.Nonce,
+			"nonce", nonce,
+		)
+		return nil, staking.ErrInvalidNonce
+	}
+
+	if ctx.IsCheckOnly() {
```
Yawning: Should the balance check be executed unconditionally?

go/tendermint/apps/staking/state/gas.go
```diff
@@ -0,0 +1,92 @@
...
+	state.SetAccount(id, account)
+
+	// Configure gas accountant on the context.
+	ctx.SetGasAccountant(abci.NewGasAccountant(fee.Gas))
+
+	return account, nil
+}
+
+// PersistBlockFees persists the accumulated fee balance for the current block.
+func PersistBlockFees(ctx *abci.Context) error {
```
Yawning: If this can't fail, should it return an error?

go/tendermint/apps/staking/state/gas.go
kostko: Well for the non-check case the balance check happens in `Move`. I will add a comment that this is the case.

go/tendermint/apps/staking/state/gas.go
kostko: Oops, this comment is no longer correct, I'll fix it.

go/oasis-test-runner/scenario/e2e/gas_fees.go
kostko: Done.

Issue #2330, opened 2019-11-04
Separate committee size parameters for compute/merge/txnscheduler


kostko:
The committee sizes for compute/merge/txnscheduler committees are currently the same. We should split these in the runtime descriptor to enable independent tuning.

Issue #2329, opened 2019-11-04
Roothash round timeout should be per-runtime


kostko:
Currently the roothash round timeout is a global consensus parameter.

We should probably make it per-runtime (e.g., part of the runtime descriptor) as different runtimes could have different timeouts. Also there should probably be separate timeouts for compute and merge committees.

PR #2337, opened 2019-11-05
registry: add runtime-specific RoundTimeout and committee sizes

matevz:
This is PR for #2329, #2330, #2354.

It does the following:
* Introduce `ComputeParams`, `MergeParams`, `TxnSchedulerParams`, and `StorageParams` structs for runtime descriptor, oasis-test-runner config and fixtures.
* Split `RoundTimeout` to `ComputeRoundTimeout` and `MergeRoundTimeout` and move them from worker to per-runtime setting in genesis.
* Split `ReplicaGroupSize` to `ComputeGroupSize`, `MergeGroupSize`, `TxnSchedulerGroupSize` and move them from genesis to per-runtime setting in genesis.
* Split `ReplicaAllowedStrugglers` to `ComputeAllowedStrugglers`, and `MergeAllowedStrugglers`.
* Join `TransactionSchedulerParameters` with other parameters in `TxnSchedulerParameters`, and move it from genesis to per-runtime setting in genesis.
* Cosmetic changes to worker, registry, runtime cmd arguments for consistency (e.g. `transaction_scheduler` or `txscheduler` -> `txn_scheduler`).
* Use Cfg* constants instead of duplicate config strings in `test_node.go`.


go/oasis-node/cmd/genesis/genesis.go
```diff
@@ -379,16 +372,8 @@ func AppendRegistryState(doc *genesis.Document, entities, runtimes, nodes []stri
...
-			RoundTimeout: viper.GetDuration(cfgRoundTimeout),
-			TransactionScheduler: roothash.TransactionSchedulerParameters{
-				Algorithm:         viper.GetString(cfgSchedulerAlgorithm),
-				BatchFlushTimeout: viper.GetDuration(cfgSchedulerBatchFlushTimeout),
-				MaxBatchSize:      viper.GetUint64(cfgSchedulerMaxBatchSize),
-				MaxBatchSizeBytes: uint64(viper.GetSizeInBytes(cfgSchedulerMaxBatchSizeBytes)),
-			},
-		},
-		Blocks: make(map[signature.MapKey]*block.Block),
+		Parameters: roothash.ConsensusParameters{},
```
kostko: Not needed as this is already the default.

go/oasis-test-runner/oasis/fixture.go
```diff
@@ -132,9 +131,14 @@ type RuntimeFixture struct {
...
 	GenesisState string `json:"genesis_state"`
 
-	ReplicaGroupSize       int `json:"replica_group_size"`
-	ReplicaGroupBackupSize int `json:"replica_group_backup_size"`
-	StorageGroupSize       int `json:"storage_group_size"`
+	ComputeGroupSize       int `json:"compute_group_size"`
+	ComputeGroupBackupSize int `json:"compute_group_backup_size"`
+	MergeGroupSize         int `json:"merge_group_size"`
+	MergeGroupBackupSize   int `json:"merge_group_backup_size"`
+	// TODO: Add Storage, TransactionScheduler GroupSizes? -Matevz
```
kostko: You can just add them now?

go/registry/api/runtime.go
```diff
@@ -107,6 +92,50 @@ type Runtime struct {
 
 	// KeyManager is the key manager runtime ID for this runtime.
 	KeyManager signature.PublicKey `json:"key_manager"`
+
+	// ReplicaAllowedStragglers is the number of allowed stragglers.
+	ReplicaAllowedStragglers uint64 `json:"replica_allowed_stragglers"`
```
kostko: So other things were moved into either compute/merge/txn sched, should this be as well?

go/registry/api/runtime.go
```diff
@@ -262,3 +301,7 @@ type RuntimeGenesis struct {
 	// invalid iff the State is non-empty or StateRoot is an empty hash.
 	StorageReceipt signature.Signature `json:"storage_receipt"`
 }
+
+// TransactionSchedulerParameters is the transaction scheduler parameters.
+type TransactionSchedulerParameters struct {
```
kostko: Since there is now a bunch of parameters I think it would be better to have multiple structs:
* compute parameters,
* merge parameters and
* txn scheduler parameters.

These would contain the group size/backup group size/etc. parameters instead of having everything directly inside the runtime descriptor.

go/worker/merge/committee/node.go
```diff
@@ -372,9 +371,10 @@ func (n *Node) tryFinalizeResultsLocked(pool *commitment.Pool, didTimeout bool)
 	// to consensus and not change the internal Discrepancy flag.
 	cid := pool.GetCommitteeID()
 	consensusTimeout := state.consensusTimeout[cid]
+	runtimeTimeout := n.commonNode.Runtime.MergeRoundTimeout
```
kostko: This should be the compute round timeout as `pool` is the per-compute-committee pool.

go/worker/txnscheduler/committee/node.go
```diff
@@ -360,8 +365,29 @@ func (n *Node) worker() {
 
 	n.logger.Info("starting committee node")
 
+	// Initialize transaction scheduler's algorithm.
+	txnAlgorithm, err := txnSchedulerAlgorithm.New(
+		n.commonNode.Runtime.TxnSchedulerAlgorithm,
```
kostko: Could put a `runtime := n.commonNode.Runtime` before this block to avoid repeating the field path multiple times.

go/common/version/version.go
```diff
@@ -54,18 +54,18 @@ var (
 	// the runtime.
 	//
 	// NOTE: This version must be synced with runtime/src/common/version.rs.
-	RuntimeProtocol = Version{Major: 0, Minor: 8, Patch: 0}
```
kostko: This doesn't need to change?

go/oasis-test-runner/oasis/fixture.go
```diff
@@ -137,9 +136,10 @@ type RuntimeFixture struct {
 	Binary       string `json:"binary"`
 	GenesisState string `json:"genesis_state"`
 
-	ReplicaGroupSize       int `json:"replica_group_size"`
-	ReplicaGroupBackupSize int `json:"replica_group_backup_size"`
-	StorageGroupSize       int `json:"storage_group_size"`
+	Compute      registry.ComputeParameters
```
kostko: Add `json` annotations for all of these.

go/registry/api/runtime.go
```diff
@@ -262,3 +327,7 @@ type RuntimeGenesis struct {
...
```
kostko: Why is this struct still here? You added `TxnSchedulerParameters` so I you can remove this.

kostko:
It seems tests are failing.

go/common/version/version.go
matevz: Yeah, I wasn't sure about which structures are in which version notions. So `RuntimeProtocol` isn't changed, but `CommitteeProtocol` and `BackendProtocol` should be bumped when changing runtime descriptor?

go/registry/api/runtime.go
kostko: I still see the struct?

go/roothash/api/api.go
```diff
@@ -170,55 +170,16 @@ type Genesis struct {
...
-	}
-
-	if g.Parameters.TransactionScheduler.MaxBatchSize < 1 {
-		return fmt.Errorf("roothash: sanity check failed: max batch size must be >= 1")
-	}
-
-	if g.Parameters.TransactionScheduler.MaxBatchSizeBytes < 1 {
-		return fmt.Errorf("roothash: sanity check failed: max batch size in bytes must be >= 1")
-	}
+	// TODO: Should we make per-runtime parameters sanity check here as well? -Matevz
```
kostko: No, that needs to be done in the registry. Did you just remove the sanity checks instead of moving them?

go/roothash/api/api.go
```diff
@@ -170,55 +170,16 @@ type Genesis struct {
...
-type ConsensusParameters struct {
-	// RoundTimeout is the round timeout.
-	RoundTimeout time.Duration `json:"round_timeout"`
-
-	// TransactionScheduler is the transaction scheduler configuration.
-	TransactionScheduler TransactionSchedulerParameters `json:"txn_scheduler"`
-}
-
 // TransactionSchedulerAlgorithmBatching is the name of the batching algorithm.
 const TransactionSchedulerAlgorithmBatching = "batching"
```
kostko: This also needs to move to the registry now.

go/common/version/version.go
matevz: lol I also bumped the `BackendProtocol` version, but someone also bumped it in master so after rebase it was gone :P It's there now :+1:

go/roothash/api/api.go
matevz: Yeah, I moved sanity checks away temporarily. I re-added them now to registry.

go/consensus/tendermint/tests/genesis_testnode.go
```diff
@@ -58,15 +57,7 @@ func NewTestNodeGenesisProvider(identity *identity.Identity) (genesis.Provider,
...
-			Parameters: roothash.ConsensusParameters{
-				RoundTimeout: 1 * time.Second,
-				TransactionScheduler: roothash.TransactionSchedulerParameters{
-					Algorithm:         batching.Name,
-					BatchFlushTimeout: 1 * time.Second,
-					MaxBatchSize:      10,
-					MaxBatchSizeBytes: 16 * 1024 * 1024,
-				},
-			},
+			Parameters: roothash.ConsensusParameters{},
```
kostko: Not needed if default, can probably remove the whole `RootHash` section.

go/oasis-test-runner/oasis/runtime.go
```diff
@@ -1,17 +1,16 @@
...
 	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
 	"github.com/oasislabs/oasis-core/go/common/node"
 	"github.com/oasislabs/oasis-core/go/common/sgx"
+	regRtCmd "github.com/oasislabs/oasis-core/go/oasis-node/cmd/registry/runtime"
 	"github.com/oasislabs/oasis-core/go/oasis-test-runner/env"
 	registry "github.com/oasislabs/oasis-core/go/registry/api"
+	"github.com/pkg/errors"
+	"os"
+	"path/filepath"
+	"strconv"
```
kostko: Move these groups of imports back to how they were before.

go/registry/api/runtime.go
```diff
@@ -69,6 +73,63 @@ func (k *RuntimeKind) FromString(str string) error {
...
+	GroupSize uint64 `json:"group_size"`
+
+	// GroupBackupSize is the size of the discrepancy resolution group.
+	GroupBackupSize uint64 `json:"group_backup_size"`
+
+	// AllowedStragglers is the number of allowed stragglers.
+	AllowedStragglers uint64 `json:"allowed_stragglers"`
+
+	// RoundTimeout is the round timeout of the nodes in the group.
+	RoundTimeout time.Duration `json:"round_timeout"`
```
kostko: I would change these timeout values to be `uint64` and always denominated in seconds. Less than second granularities could cause problems so let's just avoid them.

go/worker/merge/committee/node.go
```diff
@@ -356,11 +356,10 @@ func (n *Node) tryFinalizeResultsLocked(pool *commitment.Pool, didTimeout bool)
 	}()
 
 	epoch := n.commonNode.Group.GetEpochSnapshot()
-	// NOTE: The roothash backend will start counting its timeout on its own based on
-	//       any received commits so in the worst case the actual timeout will be
-	//       2*roundTimeout.
-	params := epoch.GetRoothashParams()
```
kostko: Do we still need `GetRoothashParams` after this change? Probably not, so it is best to just remove it (and anything related that is not needed anymore).

go/registry/api/runtime.go
matevz: I'm not sure that's a good idea since we use `time.Duration` all over the code for storing the duration (all bunch of timeouts and intervals). It's also handy for string parsing (e.g. "10s", "1m5s"). If so, I would rather check whether the granularity passed is a second or more, if that's ok?

kostko: Yeah adding sanity checks would also be fine.

PR #2339, opened 2019-11-05
Ledger device support for CLI

david-yan:
Added support for signing transactions using a Ledger device. Adds two new flags to cli for `oasis-node stake account gen_*`
Example:
```
oasis/oasis-core% ./go/oasis-node/oasis-node stake account gen_burn --help
Generate a burn transaction

Usage:
  oasis-node stake account gen_burn [flags]

Flags:
  -g, --genesis.file string           path to genesis file (default "genesis.json")
  -h, --help                          help for gen_burn
  -s, --signer string                 Type of signer to use. Options: [file, ledger]
      --signer.file.dir string        Path to file containing private key
```


go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,85 @@
...
+
+	message := append(header, payload...)
+	response, err := s.device.Exchange(message)
+	if err != nil {
+		return nil, err
+	}
+	return response, nil
+}
+
+func (s *Signer) ContextSign(context signature.Context, message []byte) ([]byte, error) {
```
Yawning: Note.  `Sign` will be deprecated via #2342, and `ContextSign` will be the only signing method available.

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return nil
+	}
+	pkLength := int(response[0])
+	addrLength := int(response[pkLength+1])
+	return response[pkLength+2 : pkLength+2+addrLength]
+}
+
+func (s *Signer) Sign(tx []byte) ([]byte, error) {
+	derivationPath := accounts.DefaultRootDerivationPath
```
Yawning: So every single signer created with the same hardware token will get the same keypair?

Our signers are role-segregated for a reason, and are meant to be instantiated via the `signature.SignerFactory` interface.  On the off chance a user wants to regenerate/rotate their key, this appears to not support that.

david-yan: So are you saying that the design should be that there are multiple devices connected to the machine, and each are assigned a role? Each device either generates or loads a 24 word seed to derive its keys, and each device only stores a single master seed at a time. So to answer your first question, yes two devices with the same master seed will get the same keypair. If we want to have role-segregated signers and still work for a single device, we can have the roles map to different derivation paths.

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,43 @@
...
+	app, err := ledger.FindLedgerOasisApp()
+	if err != nil {
+		return nil, err
+	}
+	return &Signer{app}, nil
+}
+
+func (s *Signer) Public() signature.PublicKey {
+	var pubKey signature.PublicKey
+	retrieved, _ := s.app.GetPublicKeyEd25519(defaultDerivationPath)
```
david-yan: The error is thrown out here. I don't think we want this, but the process of changing the `Signer` interface is pretty tedious, so I won't do it unless we can't think of a better solution.

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -19,10 +19,12 @@ const (
...
-	cfgVerbose = "verbose"
-	cfgForce   = "force"
-	cfgRetries = "retries"
-	cfgEntity  = "entity"
+	cfgVerbose       = "verbose"
+	cfgForce         = "force"
+	cfgRetries       = "retries"
+	cfgEntity        = "entity"
+	cfgSigner        = "signer"
+	cfgFileSignerDir = "signer.file.dir"
```
david-yan: Unsure as to whether I should include this flag, or just reuse the `entity` flag.

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,43 @@
...
+// FindLedger. This will connect to the first Ledger device found.
+func NewSigner() (signature.Signer, error) {
+	app, err := ledger.FindLedgerOasisApp()
+	if err != nil {
+		return nil, err
+	}
+	return &Signer{app}, nil
+}
+
+func (s *Signer) Public() signature.PublicKey {
```
kostko: Documentation blocks for all public methods on public structs.

go/oasis-node/cmd/common/common.go
```diff
@@ -200,3 +202,15 @@ func LoadEntity(dataDir string) (*entity.Entity, signature.Signer, error) {
 	entitySignerFactory := fileSigner.NewFactory(dataDir, signature.SignerEntity)
 	return entity.Load(dataDir, entitySignerFactory)
 }
+
+func LoadSigner(signer string) (signature.Signer, error) {
```
kostko: Add documentation block.

go/oasis-node/cmd/common/common.go
```diff
@@ -200,3 +202,15 @@ func LoadEntity(dataDir string) (*entity.Entity, signature.Signer, error) {
...
+	switch signer {
+	case "ledger":
```
kostko: These should be constants in each signer package, e.g. `ledgerSigner.SignerName`.

go/oasis-node/cmd/common/common.go
```diff
@@ -200,3 +202,15 @@ func LoadEntity(dataDir string) (*entity.Entity, signature.Signer, error) {
...
+		return ledgerSigner.NewSigner()
+	case "file":
+		_, signer, err := LoadEntity(flags.FileSignerDir())
+		return signer, err
+	default:
+		return nil, fmt.Errorf("signature format %s is not supported", signer)
```
kostko: This is not a signature format but a signer, so the error should say something like `"signer %s is not supported"`.

go/oasis-node/cmd/common/common.go
kostko:
```suggestion
func LoadSigner(signerName string) (signature.Signer, error) {
```

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -100,6 +112,9 @@ func init() {
 
 	EntityFlags.StringP(cfgEntity, "e", "", "Path to directory containing entity private key and descriptor")
 
+	SignerFlags.StringP(cfgSigner, "s", "", "Type of signer to use. Options: [file, ledger]")
```
kostko: There should be a default (file) as otherwise this breaks existing use cases.

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -100,6 +112,9 @@ func init() {
...
+	SignerFlags.String(cfgFileSignerDir, "", "Path to file containing private key")
```
kostko: The description is incorrect (it should point to the directory containing the private keys). Also it should default to the data directory as otherwise this breaks existing use cases.

go/oasis-node/cmd/common/common.go
```diff
@@ -200,3 +202,15 @@ func LoadEntity(dataDir string) (*entity.Entity, signature.Signer, error) {
 	entitySignerFactory := fileSigner.NewFactory(dataDir, signature.SignerEntity)
```
kostko: Actually, why did you introduce a separate `LoadSigner` function? Based on the comment above, the correct way to do this is to add the signer selection part into `LoadEntity`.

And the ledger signer should expose a factory (a `SignerFactory`, see how the file-based signer does it) which only supports the `signature.SignerEntity` role.

go/oasis-node/cmd/common/common.go
```diff
@@ -200,3 +202,15 @@ func LoadEntity(dataDir string) (*entity.Entity, signature.Signer, error) {
...
 	return entity.Load(dataDir, entitySignerFactory)
 }
+
+func LoadSigner(signer string) (signature.Signer, error) {
+	switch signer {
+	case "ledger":
+		return ledgerSigner.NewSigner()
+	case "file":
+		_, signer, err := LoadEntity(flags.FileSignerDir())
```
kostko: This is incorrect as if `signer.file.dir` is different from the data directory where the entity descriptor is kept, entity loading will fail.

When you move this to `LoadEntity`, you should differentiate between the data directory (should be used as the `entity.Load` argument) and the signer directory (should be used as the `fileSigner.NewFactory` argument).

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,43 @@
...
+	defaultDerivationPath = []uint32{44, 118, 0, 0, 0}
+)
+
+type Signer struct {
+	app *ledger.LedgerOasis
+}
+
+// NewSigner will attempt to initialize a Ledger Signer by running
+// FindLedger. This will connect to the first Ledger device found.
+func NewSigner() (signature.Signer, error) {
```
kostko: Should this take a ledger device number and derivation path as an argument? The factory (see other comments) can then populate them with `0` (for now if the upstream module does not support multiple devices) and the derivation path for the `EntitySigner` role.

The derivation paths could be mapped similar to how the file signer maps filenames:
```go
	rolePEMFiles = map[signature.SignerRole]string{
		signature.SignerEntity:    FileEntityKey,
		signature.SignerNode:      FileIdentityKey,
		signature.SignerP2P:       FileP2PKey,
		signature.SignerConsensus: FileConsensusKey,
	}
```

Except that you would only support the `signature.SignerEntity` role for now (with the above derivation path).

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,43 @@
+package ledger
+
+import (
+	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
+	ledger "github.com/zondax/ledger-oasis-go"
+)
+
+var (
+	defaultDerivationPath = []uint32{44, 118, 0, 0, 0}
```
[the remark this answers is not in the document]
david-yan: I took this path from a test. I'm not sure if this is the default path we want, or if we want a default path at all.

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,126 @@
...
+	if fac.address == "" {
+		device, err := ledger.FindLedgerOasisApp()
+		if err != nil {
+			return nil, err
+		}
+
+		return &Signer{device, path}, nil
+	}
+
+	device, err := ledger.ConnectLedgerOasisApp(fac.address, path)
```
david-yan: Build will fail until https://github.com/ZondaX/ledger-oasis-go/pull/4, which enables ledger devices to be connected by address, is merged. We can't rely on a device index, since the index is not guaranteed to be the same between calls.

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+var (
+	_ signature.SignerFactoryCtor = NewFactory
+	_ signature.SignerFactory     = (*Factory)(nil)
+	_ signature.Signer            = (*Signer)(nil)
+
+	// SignerDerivationPath is the derivation path used for generating
+	// the signature key on the Ledger device.
+	SignerDerivationPath = []uint32{44, 118, 0, 0, 0}
```
david-yan: Not sure if we want a default path. If not, then the factory can also take a derivation path as an argument, and `roleDerivationPath` would be unnecessary.

go/oasis-node/cmd/common/common.go
```diff
@@ -191,12 +193,21 @@ func GetInputReader(cmd *cobra.Command, cfg string) (io.ReadCloser, bool, error)
 }
 
 // LoadEntity loads the entity and it's signer.
-func LoadEntity(dataDir string) (*entity.Entity, signature.Signer, error) {
+func LoadEntity(backing string, factoryArgs string) (*entity.Entity, signature.Signer, error) {
```
david-yan: @kostko is this what you had in mind?

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -98,7 +104,8 @@ func init() {
 	DebugTestEntityFlags.Bool(CfgDebugTestEntity, false, "use the test entity (UNSAFE)")
 	_ = DebugTestEntityFlags.MarkHidden(CfgDebugTestEntity)
 
-	EntityFlags.StringP(cfgEntity, "e", "", "Path to directory containing entity private key and descriptor")
+	EntityFlags.StringP(cfgEntity, "e", "", "The source of the entity, based on its backing. For file backing, this is the path to the directory containing the entity private key and descriptor. For ledger backing, this is the expected address stored on the device.")
```
david-yan: A bit of a long description. I was wondering if I should split this argument? Or if I just need a more succinct description.

go/common/crypto/signature/signers/file/file_signer.go
```diff
@@ -18,6 +18,9 @@ const (
 	privateKeyPemType = "ED25519 PRIVATE KEY"
 
 	filePerm = 0600
+
+	// SignerName is the name used to identify this package.
```
kostko:
```suggestion
	// SignerName is the name used to identify the file backed signer.
```

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,43 @@
+package ledger
+
+import (
+	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
+	ledger "github.com/zondax/ledger-oasis-go"
```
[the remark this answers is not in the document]
kostko: External imports should be above internal imports, not below.

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,126 @@
...
+	"errors"
+	"io"
+
+	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
+
+	ledger "github.com/zondax/ledger-oasis-go"
+)
+
+const (
+	// SignerName is the name used to identify this package.
```
kostko:
```suggestion
	// SignerName is the name used to identify the Ledger backed signer.
```

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,126 @@
+package ledger
```
kostko: Add something like:
```
// Package ledger provides a Ledger backed signer.
```

go/common/crypto/signature/signers/ledger/ledger_signer.go
kostko: No, role-based derivation path would still be necessary if you wanted support for multiple roles (currently not, but maybe some day). But each role could just be a suffix to the passed derivation path.

If the derivation path (prefix) is configurable, it should be passed in via configuration arguments.

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,126 @@
...
+	// SignerDerivationPath is the derivation path used for generating
+	// the signature key on the Ledger device.
+	SignerDerivationPath = []uint32{44, 118, 0, 0, 0}
+
+	roleDerivationPaths = map[signature.SignerRole][]uint32{
+		signature.SignerEntity: SignerDerivationPath,
+	}
+)
+
+// Factory is a Ledger backed SignerFactory
```
kostko:
```suggestion
// Factory is a Ledger backed SignerFactory.
```

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,126 @@
...
+type Factory struct {
+	roles   []signature.SignerRole
+	address string
+}
+
+// NewFactory creates a new factory with the specified roles
```
kostko:
```suggestion
// NewFactory creates a new factory with the specified roles.
```

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// Load will create a Signer backed by a Ledger device by searching for
+// a device with the expected address. If no address is provided, this will
+// created a Signer with the first Ledger device it finds.
+// The only role allowed is SignerEntity.
+func (fac *Factory) Load(role signature.SignerRole) (signature.Signer, error) {
+	path, ok := roleDerivationPaths[role]
+	if !ok {
+		return nil, errors.New("ledger signer is not allowed to have this role")
```
kostko:
```suggestion
		return nil, fmt.Errorf("role %s is not supported when using the Ledger backed signer", role)
```

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	if fac.address == "" {
```
kostko: Isn't this something that should be handled by the upstream library (e.g., the `ConnectLedgerOasisApp` function)?

go/common/crypto/signature/signers/ledger/ledger_signer.go
kostko: What kind of address is this? This should be something that is independent of the derivation path and only dependent on the actual device (like a device ID).

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// Signer is a Ledger backed Signer.
+type Signer struct {
+	device *ledger.LedgerOasis
+	path   []uint32
+}
+
+// Public retrieves the public key from the Ledger device.
+// It requires a user confirmation on the device to ensure that the
```
kostko: Wait what, fetching the public key requires user confirmation? This seems wrong and annoying? What security benefit is there of confirming fetching of a public key? Callers of `Public()` will not display the public key to the user so there is no way for the user to check anything.

If this is really the case, it should be changed IMO.

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,126 @@
...
+// correct public key is displayed.
+func (s *Signer) Public() signature.PublicKey {
+	var pubKey signature.PublicKey
+	retrieved, _, err := s.device.GetAddressPubKeyEd25519(s.path)
+	if err != nil {
+		panic("failed to retrieve public key from device")
```
kostko: Propagate the actual error message:
```suggestion
		panic(fmt.Errorf("failed to retrieve public key from device: %w", err))
```

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,126 @@
...
+	return pubKey
+}
+
+// ContextSign generates a signature with the private key over the context and
+// message.
+func (s *Signer) ContextSign(context signature.Context, message []byte) ([]byte, error) {
+	return s.device.SignEd25519(s.path, []byte(context), message)
+}
+
+// String returns the address of the account on the Ledger device.
```
kostko: Can this be more like `"[ledger signer: <pubkey>]"`? Also call `s.Public()` instead of duplicating code.

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,126 @@
...
+	device *ledger.LedgerOasis
+	path   []uint32
+}
+
+// Public retrieves the public key from the Ledger device.
+// It requires a user confirmation on the device to ensure that the
+// correct public key is displayed.
+func (s *Signer) Public() signature.PublicKey {
+	var pubKey signature.PublicKey
+	retrieved, _, err := s.device.GetAddressPubKeyEd25519(s.path)
```
kostko: Cache the public key as it should never change during operation and querying the device each time seems excessive?

go/go.sum
```diff
@@ -55,6 +55,8 @@ github.com/coreos/go-semver v0.3.0 h1:wkHLiw0WNATZnSG7epLsujiMCgPAc9xhjJ4tgnAxmf
 github.com/coreos/go-semver v0.3.0/go.mod h1:nnelYz7RCh+5ahJtPPxZlU+153eP4D4r3EedlOD2RNk=
 github.com/coreos/go-systemd v0.0.0-20190321100706-95778dfbb74e/go.mod h1:F5haX7vjVVG0kc13fIWeqUViNPyEJxv/OmvnBo0Yme4=
 github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f/go.mod h1:E3G3o1h8I7cfcXa63jLwjI0eiQQMgzzUDFVpN/nH/eA=
+github.com/cosmos/ledger-go v0.9.2 h1:Nnao/dLwaVTk1Q5U9THldpUMMXU94BOTWPddSmVB6pI=
```
kostko: Why do we need a cosmos-specific dependency?

go/oasis-node/cmd/common/common.go
```diff
@@ -191,12 +193,21 @@ func GetInputReader(cmd *cobra.Command, cfg string) (io.ReadCloser, bool, error)
 }
 
 // LoadEntity loads the entity and it's signer.
-func LoadEntity(dataDir string) (*entity.Entity, signature.Signer, error) {
+func LoadEntity(backing string) (*entity.Entity, signature.Signer, error) {
```
kostko:
```suggestion
func LoadEntity(backend string) (*entity.Entity, signature.Signer, error) {
```

go/oasis-node/cmd/common/common.go
```diff
@@ -191,12 +193,21 @@ func GetInputReader(cmd *cobra.Command, cfg string) (io.ReadCloser, bool, error)
...
 	if flags.DebugTestEntity() {
 		return entity.TestEntity()
 	}
 
 	// TODO/hsm: Configure factory dynamically.
```
kostko: Remove the TODO now that the factory is configured dynamically?

go/oasis-node/cmd/common/common.go
```diff
@@ -191,12 +193,21 @@ func GetInputReader(cmd *cobra.Command, cfg string) (io.ReadCloser, bool, error)
...
-	entitySignerFactory := fileSigner.NewFactory(dataDir, signature.SignerEntity)
-	return entity.Load(dataDir, entitySignerFactory)
+	switch backing {
+	case ledgerSigner.SignerName:
+		factory := ledgerSigner.NewFactory(flags.EntityFileDir(), signature.SignerEntity)
+		signer, err := factory.Load(signature.SignerEntity)
+		return nil, signer, err
+	case fileSigner.SignerName:
+		factory := factorySigner.NewFactory(flags.EntityLedgerAddress(), signature.SignerEntity)
+		return entity.Load(dataDir, factory)
```
kostko: Where does the `dataDir` come from? You should call `DataDir()` or `DataDirOrPwd()` here.

go/oasis-node/cmd/common/common.go
```diff
@@ -191,12 +193,21 @@ func GetInputReader(cmd *cobra.Command, cfg string) (io.ReadCloser, bool, error)
...
+	default:
+		return nil, nil, fmt.Errorf("signer with %s backing is not supported", backing)
```
kostko:
```suggestion
		return nil, nil, fmt.Errorf("unsupported signer backend: %s", backing)
```

go/oasis-node/cmd/common/common.go
```diff
@@ -191,12 +193,21 @@ func GetInputReader(cmd *cobra.Command, cfg string) (io.ReadCloser, bool, error)
...
 	}
 
 	// TODO/hsm: Configure factory dynamically.
-	entitySignerFactory := fileSigner.NewFactory(dataDir, signature.SignerEntity)
-	return entity.Load(dataDir, entitySignerFactory)
+	switch backing {
+	case ledgerSigner.SignerName:
+		factory := ledgerSigner.NewFactory(flags.EntityFileDir(), signature.SignerEntity)
+		signer, err := factory.Load(signature.SignerEntity)
+		return nil, signer, err
```
kostko: Please explain why the entity is not loaded and nil is returned here? It should be exactly the same invocation as for the file based signer (`entity.Load(DataDir(), factory)`)?

go/oasis-node/cmd/common/common.go
```diff
@@ -191,12 +193,21 @@ func GetInputReader(cmd *cobra.Command, cfg string) (io.ReadCloser, bool, error)
...
 		return entity.TestEntity()
 	}
 
 	// TODO/hsm: Configure factory dynamically.
-	entitySignerFactory := fileSigner.NewFactory(dataDir, signature.SignerEntity)
-	return entity.Load(dataDir, entitySignerFactory)
+	switch backing {
+	case ledgerSigner.SignerName:
+		factory := ledgerSigner.NewFactory(flags.EntityFileDir(), signature.SignerEntity)
+		signer, err := factory.Load(signature.SignerEntity)
```
kostko: Use `entity.Load` as below and this will not be needed.

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -19,10 +19,12 @@ const (
...
 
-	cfgVerbose = "verbose"
-	cfgForce   = "force"
-	cfgRetries = "retries"
-	cfgEntity  = "entity"
+	cfgVerbose       = "verbose"
+	cfgForce         = "force"
+	cfgRetries       = "retries"
+	cfgEntity        = "entity"
+	cfgFileDir       = "entity.file.dir"
```
kostko: The prefix should either be `signer.` or `entity.signer.` as `entity.file.dir` sounds really confusing? Also below.

go/oasis-node/cmd/stake/account.go
```diff
@@ -87,6 +88,12 @@ var (
 		Short: "Generate an amend_commission_schedule transaction",
 		Run:   doAccountAmendCommissionSchedule,
 	}
+
+	accountLedgerListCmd = &cobra.Command{
```
kostko: Why is this command in the staking subcommand? This seems like something that is signer-specific.

The ledger based signer will not only be used for staking, but also for other kinds of transactions and signatures (e.g., entity registration). I would put this under a new subcommand, e.g. `oasis-node signer ledger list_devices` or something.

go/oasis-node/cmd/common/common.go
david-yan: Maybe a misunderstanding on my part, but I didn't think there would necessarily be an Entity associated with the Ledger device. If there is always an associated Entity, then I guess the dataDir is always required.

go/go.sum
david-yan: We don't, this was leftover from a previous build.

go/common/crypto/signature/signers/ledger/ledger_signer.go
david-yan: Hmm I suppose.

go/common/crypto/signature/signers/ledger/ledger_signer.go
david-yan: So from talking to Juan, there doesn't seem to be a unique ID to identify each device, so the solution is to use the address derived from each device's mnemonic.

go/oasis-node/cmd/stake/account.go
david-yan: Agreed. I wasn't sure where to put this, so I just added this here.

go/common/crypto/signature/signers/ledger/ledger_signer.go
kostko: Should then this be based on a specific (static) derivation path, separate from the one used for the actual key? As otherwise if you have multiple roles (currently not, but we must think ahead), you will need to configure the address for each role which seems annoying?

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// Public retrieves the public key from the Ledger device.
+func (s *Signer) Public() signature.PublicKey {
+	if s.publicKey != nil {
+		return *s.publicKey
+	}
+
+	var pubKey signature.PublicKey
+	retrieved, err := s.device.GetPublicKeyEd25519(s.path)
```
kostko: I see that you added caching and removed the comment about user confirmation, but I want to be sure that this operation will never require user confirmation?

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,118 @@
...
+
+// ContextSign generates a signature with the private key over the context and
+// message.
+func (s *Signer) ContextSign(context signature.Context, message []byte) ([]byte, error) {
+	return s.device.SignEd25519(s.path, []byte(context), message)
+}
+
+// String returns the address of the account on the Ledger device.
+func (s *Signer) String() string {
+	return fmt.Sprintf("[ledger signer: %x]", s.Public())
```
kostko: This should probably use `%s` as the `signature.PublicKey` type has its own string representation.

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -19,10 +19,12 @@ const (
...
 
-	cfgVerbose = "verbose"
-	cfgForce   = "force"
-	cfgRetries = "retries"
-	cfgEntity  = "entity"
+	cfgVerbose       = "verbose"
+	cfgForce         = "force"
+	cfgRetries       = "retries"
+	cfgEntity        = "entity"
+	cfgFileDir       = "signer.file.dir"
```
kostko:
```suggestion
	cfgSignerFileDir       = "signer.file.dir"
```

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -19,10 +19,12 @@ const (
...
+	cfgLedgerAddress = "signer.ledger.address"
```
kostko:
```suggestion
	cfgSignerLedgerAddress = "signer.ledger.address"
```

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -98,7 +110,9 @@ func init() {
 	DebugTestEntityFlags.Bool(CfgDebugTestEntity, false, "use the test entity (UNSAFE)")
 	_ = DebugTestEntityFlags.MarkHidden(CfgDebugTestEntity)
 
-	EntityFlags.StringP(cfgEntity, "e", "", "Path to directory containing entity private key and descriptor")
+	EntityFlags.StringP(cfgEntity, "e", "file", "Backing for of the entity. Options: [file, ledger]")
```
kostko: Hm this seems like a confusing choice for the signer backend? Shouldn't it be something like `cfgSigner` and be named `signer`? If not needed, the old `entity` option should be removed instead of being repurposed.

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -71,11 +73,21 @@ func DebugTestEntity() bool {
 	return DebugDontBlameOasis() && viper.GetBool(CfgDebugTestEntity)
 }
 
-// Entity returns the set entity directory.
+// Entity returns the entity backing.
 func Entity() string {
 	return viper.GetString(cfgEntity)
 }
 
+// EntityFileDir returns the set entity directory.
```
kostko: Isn't this the signer file dir? It could be used for non-entity things? As the other comment says I think these should be the `Signer`, `SignerFileDir` and `SignerLedgerAddress` respectively.

go/oasis-node/cmd/common/flags/flags.go
kostko: Also `EntityFlags` should now probably be `SignerFlags`?

go/oasis-node/cmd/stake/account.go
```diff
@@ -295,6 +296,14 @@ func doAccountAmendCommissionSchedule(cmd *cobra.Command, args []string) {
 	cmdConsensus.SignAndSaveTx(tx)
 }
 
+func doLedgerList(cmd *cobra.Command, args []string) {
```
kostko: Hm leftover? This is not needed now that you added a signer subcommand.

go/oasis-node/cmd/signer/signer.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/oasislabs/oasis-core/go/common/logging"
+)
+
+var (
+	signerCmd = &cobra.Command{
+		Use:   "signer",
+		Short: "signer backend utilities",
+	}
+
+	logger = logging.GetLogger("cmd/stake")
```
kostko: This seems unused (and the logger name is wrong), you can remove it for now.

go/oasis-node/cmd/signer/ledger/ledger.go
```diff
@@ -0,0 +1,38 @@
...
+		Run:   doLedgerList,
+	}
+)
+
+func doLedgerList(cmd *cobra.Command, args []string) {
+	if err := cmdCommon.Init(); err != nil {
+		cmdCommon.EarlyLogAndExit(err)
+	}
+
+	ledger.ListOasisDevices(ledgerSigner.SignerDerivationPath)
```
kostko: Should this wrapper (that uses a specific derivation path) live in the `go/common/crypto/signature/signer/ledger` instead of here? Just call it `ListDevices` and make it take no arguments.

kostko:
Note: this will require updates to the node operator docs due to CLI changes.

go/common/crypto/signature/signers/ledger/ledger_signer.go
david-yan: Correct, as per the documentation in ledger-oasis-go, this call does not require user confirmation.

go/oasis-node/cmd/signer/ledger/ledger.go
david-yan: Ah good idea!

go/common/crypto/signature/signers/ledger/ledger_signer.go
david-yan: Right, so there are a few options that I see for resolving this issue, and would like to address it now. We could:
1. Leave it as is, the ListDevices command will always display the devices based on some static path, and the user will always used that path when connecting to the device, but will use a different path based on role.
2. We make the path configurable, adding a flag, and removing the role to path mapping. Or keep it as default paths. In which case, the path (or role) will be required for all calls involving Ledger. 

I'd prefer the first option since it would make the process for the users easier, although it does not allow for paths to be configurable.

kostko: There is also option 3:
* Use a static path for listing and "pinning" devices.
* Use a (separate from the above) static path prefix based on role (e.g., specific role has specific prefix).
* Allow the user to configure the path suffix (and use a default if not specified).

go/common/crypto/signature/signers/ledger/ledger_signer.go
jleni: We careful with `118`. That number belongs to Cosmos. I know the SLIP-0044 registration is still missing, but maybe it is a good idea to unify this `118` into a single place to avoid future mix ups.

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -98,7 +110,9 @@ func init() {
 	DebugTestEntityFlags.Bool(CfgDebugTestEntity, false, "use the test entity (UNSAFE)")
 	_ = DebugTestEntityFlags.MarkHidden(CfgDebugTestEntity)
 
-	EntityFlags.StringP(CfgEntity, "e", "", "Path to directory containing entity private key and descriptor")
+	SignerFlags.StringP(cfgSigner, "s", "file", "Backing for of the entity. Options: [file, ledger]")
```
kostko:
```suggestion
	SignerFlags.StringP(cfgSigner, "s", "file", "Signer backend [file, ledger]")
```

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -98,7 +110,9 @@ func init() {
...
+	SignerFlags.String(cfgSignerFileDir, "", "Path to directory containing entity private key and descriptor")
```
kostko:
```suggestion
	SignerFlags.String(cfgSignerFileDir, "", "File signer: path to directory containing private key")
```

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -98,7 +110,9 @@ func init() {
...
+	SignerFlags.String(cfgSignerLedgerAddress, "", "Select Ledger device based on this specified address. If blank, any available Ledger device will be connected to.")
```
kostko:
```suggestion
	SignerFlags.String(cfgSignerLedgerAddress, "", "Ledger signer: select Ledger device based on this specified address. If blank, any available Ledger device will be connected to.")
```

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -71,9 +73,19 @@ func DebugTestEntity() bool {
 	return DebugDontBlameOasis() && viper.GetBool(CfgDebugTestEntity)
 }
 
-// Entity returns the set entity directory.
-func Entity() string {
-	return viper.GetString(CfgEntity)
+// Entity returns the entity backing.
```
kostko:
```suggestion
// Signer returns the configured signer backend name.
```

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -71,9 +73,19 @@ func DebugTestEntity() bool {
...
+func Signer() string {
+	return viper.GetString(cfgSigner)
+}
+
+// SignerFileDir returns the directory with the signer keys.
```
kostko:
```suggestion
// SignerFileDir returns the directory with the signer keys (for file-based signer).
```

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -71,9 +73,19 @@ func DebugTestEntity() bool {
...
+func SignerFileDir() string {
+	return viper.GetString(cfgSignerFileDir)
+}
+
+// SignerLedgerAddress returns the address to search for.
```
kostko:
```suggestion
// SignerLedgerAddress returns the address to search for (for Ledger-based signer).
```

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -34,8 +36,8 @@ var (
 	RetriesFlags = flag.NewFlagSet("", flag.ContinueOnError)
 	// DebugTestEntityFlags has the test entity enable flag.
 	DebugTestEntityFlags = flag.NewFlagSet("", flag.ContinueOnError)
-	// EntityFlags has the entity flag.
-	EntityFlags = flag.NewFlagSet("", flag.ContinueOnError)
+	// SignerFlags has the entity flag.
```
kostko:
```suggestion
	// SignerFlags has the signer-related flags.
```

go/oasis-node/cmd/common/flags/flags.go
```diff
@@ -19,10 +19,12 @@ const (
...
 	CfgConsensusValidator = "consensus.validator"
 
-	cfgVerbose = "verbose"
-	cfgForce   = "force"
-	cfgRetries = "retries"
-	CfgEntity  = "entity"
+	cfgVerbose             = "verbose"
+	cfgForce               = "force"
+	cfgRetries             = "retries"
+	cfgSigner              = "signer"
```
kostko: Group signer flags into its own group (e.g., add new line above this line).

go/oasis-node/cmd/common/common.go
```diff
@@ -191,12 +193,19 @@ func GetInputReader(cmd *cobra.Command, cfg string) (io.ReadCloser, bool, error)
...
 		return entity.TestEntity()
 	}
 
-	// TODO/hsm: Configure factory dynamically.
-	entitySignerFactory := fileSigner.NewFactory(dataDir, signature.SignerEntity)
-	return entity.Load(dataDir, entitySignerFactory)
+	var factory signature.SignerFactory
+	switch backing {
+	case ledgerSigner.SignerName:
+		factory = ledgerSigner.NewFactory(flags.EntityFileDir(), signature.SignerEntity)
```
kostko:
```suggestion
		factory = ledgerSigner.NewFactory(flags.SignerFileDir(), signature.SignerEntity)
```

go/oasis-node/cmd/common/common.go
```diff
@@ -191,12 +193,19 @@ func GetInputReader(cmd *cobra.Command, cfg string) (io.ReadCloser, bool, error)
...
+	case fileSigner.SignerName:
+		factory = factorySigner.NewFactory(flags.EntityLedgerAddress(), signature.SignerEntity)
```
kostko:
```suggestion
		factory = factorySigner.NewFactory(flags.SignerLedgerAddress(), signature.SignerEntity)
```

go/oasis-node/cmd/common/common.go
```diff
@@ -191,12 +193,19 @@ func GetInputReader(cmd *cobra.Command, cfg string) (io.ReadCloser, bool, error)
 }
 
 // LoadEntity loads the entity and it's signer.
-func LoadEntity(dataDir string) (*entity.Entity, signature.Signer, error) {
+func LoadEntity(backend string) (*entity.Entity, signature.Signer, error) {
```
kostko:
```suggestion
func LoadEntity(signerBackend string) (*entity.Entity, signature.Signer, error) {
```

go/common/crypto/signature/signers/ledger/ledger_signer.go
kostko: Yeah any kind of default derivation path prefix should be clearly annotated and not use magic numbers. E.g., there should be a constant that defines 44 and 128 (or something else) with comments, so that it is more clear where these come from (e.g., from SLIP-44 registration).

go/common/crypto/signature/signers/ledger/ledger_signer.go
david-yan: Ok that makes sense. I'll implement option 3 then.

go/common/ledger/ledger.go
```diff
@@ -0,0 +1,35 @@
...
+
+var (
+	// ListingDerivationPath is the path used to list and connect to devices by address.
+	ListingDerivationPath = []uint32{ListingPathPurpose, ListingPathCoinType, ListingPathAccount, ListingPathChange, ListingPathIndex}
+)
+
+type LedgerDevice = ledger.LedgerOasis
+
+func ListDevices() {
+	ledger.ListOasisDevices(ListingDerivationPath)
```
david-yan: This command requires user confirmation for all devices listed, which is pretty annoying and unnecessary from a user's perspective. @jleni is there a chance that we can add a function to `ledger-oasis-go` that gets the address and does not require user confirmation?

jleni: Sorry, that should have been exposed there! It is actually part of the API and it is already exposed in JS. Adding that right now to the go library.

go/go.sum
```diff
@@ -509,6 +513,16 @@ github.com/willf/bitset v1.1.10 h1:NotGKqX0KwQ72NUzqrjZq5ipPNDQex9lo3WpaS8L2sc=
 github.com/willf/bitset v1.1.10/go.mod h1:RjeCKbqT1RxIR/KWY6phxZiaY1IyutSBfGjNPySAYV4=
 github.com/xiang90/probing v0.0.0-20190116061207-43a291ad63a2/go.mod h1:UETIi67q53MR2AWcXfiuqkDkRtnGDLqkBTpCHuJHxtU=
 github.com/xordataexchange/crypt v0.0.3-0.20170626215501-b2862e3d0a77/go.mod h1:aYKd//L2LvnjZzWKhF00oedf4jCCReLcmhLdhm1A27Q=
+github.com/zondax/hid v0.9.0 h1:eiT3P6vNxAEVxXMw66eZUAAnU2zD33JBkfG/EnfAKl8=
+github.com/zondax/hid v0.9.0/go.mod h1:l5wttcP0jwtdLjqjMMWFVEE7d1zO0jvSPA9OPZxWpEM=
+github.com/zondax/ledger-go v0.11.0 h1:EEqUh6eaZucWAaGo87G7sJiqRNJpzBZr+I9PpGgjjPg=
+github.com/zondax/ledger-go v0.11.0/go.mod h1:NI6JDs8VWwgh+9Bf1vPZMm9Xufp2Q7Iwm2IzxJWzmus=
+github.com/zondax/ledger-oasis-go v0.0.0-20191203183031-855587dca777 h1:iTs0RNdmbEmqcGVCnUglZ+s8eUru2guedYETLBOsf2c=
```
kostko: There is a bunch of leftover versions here. Please clean and regenerate `go.sum`.

go/oasis-node/cmd/registry/entity/entity.go
```diff
@@ -358,8 +359,17 @@ func loadOrGenerateEntity(dataDir string, generate bool) (*entity.Entity, signat
 		return nil, nil, fmt.Errorf("loadOrGenerateEntity: sanity check failed: one or more unsafe debug flags set")
 	}
 
-	// TODO/hsm: Configure factory dynamically.
-	entitySignerFactory := fileSigner.NewFactory(dataDir, signature.SignerEntity)
+	var entitySignerFactory signature.SignerFactory
```
kostko: This seems like duplicated code from `LoadEntity` in `cmd/common`. Instead of duplicating the code move it into a separate function in `cmd/common/common` and use that function in both places.

go/common/crypto/signature/signer.go
```diff
@@ -201,7 +202,11 @@ func PrepareSignerContext(context Context) ([]byte, error) {
 		if chainContext == "" {
 			return nil, errNoChainContext
 		}
-		context = context + chainContextSeparator + chainContext
+		decodedChainContext, err := hex.DecodeString(string(chainContext))
```
kostko: What are you doing here? There is no requirement that "chainContext" is a hex string. It can be an arbitrary string of 64 bytes.

For Ledger I believe we just restricted to a subset of ASCII, we could enforce that for contexts, but in this case such a check should be done at context registration (so it is caught immediately) not when things are being signed.

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -0,0 +1,142 @@
...
+	ledgerCommon "github.com/oasislabs/oasis-core/go/common/ledger"
+)
+
+const (
+	// SignerName is the name used to identify the Ledger backed signer.
+	SignerName = "ledger"
+
+	// SignerPathPurpose is set to 44 to indicate use of the BIP-0044 specification.
+	SignerPathPurpose uint32 = 44
+	// SignerPathCoinType is set to 118, the number owned by Cosmos via SLIP-0044 registration.
```
kostko: Why don't we set this to a number that is not owned by anyone at the moment?

go/oasis-node/cmd/common/common.go
```diff
@@ -57,6 +59,21 @@ func DataDirOrPwd() (string, error) {
 	return dataDir, nil
 }
 
+func EntityFactory(signerBackend string, entityDir string) (signature.SignerFactory, error) {
```
kostko: The method is strangely named and is missing a documentation block. It should be something like `SignerFactory` instead as it has nothing to do with an entity.

Also why do you use `signerBackend` and `entityDir` (which should be `fileSignerDir`) and not just use the `SignerDirOrPwd` and other flags instead?

go/common/crypto/signature/signer.go
david-yan: Sorry I was just trying to reduce the length of the resulting context to test on the device. I'll revert this change now.

go/oasis-node/cmd/common/common.go
david-yan: Ah it should be `SignerFactory` and `signerDir`. I renamed `fileSignerDir` to `signerDir` because it seems that signers will require a directory no matter what, not just for signers with a file backend.

go/oasis-test-runner/oasis/entity.go
```diff
@@ -65,7 +65,7 @@ func (ent *Entity) DescriptorPath() string {
 
 func (ent *Entity) toGenesisArgs() []string {
 	if ent.dir != nil {
-		return []string{"--entity", ent.dir.String()}
+		return []string{"--signer", "file", "--signer.dir", ent.dir.String()}
```
kostko: Import proper constant instead of hardcoding `"file"`.

go/oasis-test-runner/oasis/entity.go
```diff
@@ -90,7 +90,8 @@ func (ent *Entity) update() error {
 
 	args := []string{
 		"registry", "entity", "update",
-		"--" + common.CfgDataDir, ent.dir.String(),
+		"--" + flags.CfgSigner, "file",
```
kostko: Import proper constant instead of hardcoding `"file"`.

go/oasis-test-runner/oasis/entity.go
```diff
@@ -142,7 +143,8 @@ func (net *Network) NewEntity(cfg *EntityCfg) (*Entity, error) {
 		if !cfg.Restore {
 			args := []string{
 				"registry", "entity", "init",
-				"--" + common.CfgDataDir, entityDir.String(),
+				"--" + flags.CfgSigner, "file",
```
kostko: Import proper constant instead of hardcoding `"file"`.

go/oasis-test-runner/scenario/e2e/registry_cli.go
```diff
@@ -256,7 +256,8 @@ func (r *registryCLIImpl) initEntity(childEnv *env.Env, entDir string) (*entity.
 
 	args := []string{
 		"registry", "entity", "init",
-		"--" + common.CfgDataDir, entDir,
+		"--" + flags.CfgSigner, "file",
```
kostko: Import proper constant instead of hardcoding `"file"`.

go/oasis-test-runner/scenario/e2e/registry_cli.go
```diff
@@ -277,7 +278,8 @@ func (r *registryCLIImpl) updateEntity(childEnv *env.Env, nodes []*node.Node, no
 
 	args := []string{
 		"registry", "entity", "update",
-		"--" + common.CfgDataDir, entDir,
+		"--" + flags.CfgSigner, "file",
```
kostko: Import proper constant instead of hardcoding `"file"`.

go/oasis-test-runner/scenario/e2e/registry_cli.go
```diff
@@ -411,7 +413,8 @@ func (r *registryCLIImpl) initNode(childEnv *env.Env, ent *entity.Entity, entDir
 			"--" + cmdRegNode.CfgP2PAddress, strings.Join(testAddressesStr, ","),
 			"--" + cmdRegNode.CfgRole, testNode.Roles.String(),
 			"--" + cmdRegNode.CfgNodeRuntimeID, testNode.Runtimes[0].ID.String(),
-			"--" + flags.CfgEntity, entDir,
+			"--" + flags.CfgSigner, "file",
```
kostko: Import proper constant instead of hardcoding `"file"`.

go/oasis-test-runner/scenario/e2e/registry_cli.go
```diff
@@ -503,7 +506,8 @@ func (r *registryCLIImpl) genRegisterEntityTx(childEnv *env.Env, nonce int, txPa
 		"--" + consensus.CfgTxFeeGas, strconv.Itoa(feeGas),
 		"--" + flags.CfgDebugDontBlameOasis,
 		"--" + common.CfgDebugAllowTestKeys,
-		"--" + flags.CfgEntity, entDir,
+		"--" + flags.CfgSigner, "file",
```
kostko: Import proper constant instead of hardcoding `"file"`.

go/oasis-test-runner/scenario/e2e/registry_cli.go
```diff
@@ -525,7 +529,8 @@ func (r *registryCLIImpl) genDeregisterEntityTx(childEnv *env.Env, nonce int, tx
...
```
kostko: Import proper constant instead of hardcoding `"file"`.

Issue #2277, opened 2019-10-18
Invariant checking during longer-term runs


kostko:
We should add tests that check if any invariants are being violated during longer runs. The test(s) would run long-term scenarios (injecting various kinds of transactions).

There needs to be:
* A generic way to describe system invariants (e.g., total amount of tokens as accounted from all the places that hold tokens should stay the same) such that it is easy to check them at any point in time. This may be possible to reuse in #1633.
* A generic way to inject transactions (consensus and runtime, with priority on consensus first) during the long-term scenarios. Probably using some kind of random distribution, based on an initial random seed.

The tests can run using our existing test runner, and they should be run daily in CI on the current master.

*Estimated cost: ~1 sprint*

PR #2340, opened 2019-11-06
a node that follows and checks

pro-wh:
This is proposed as the beginning of auditing tools. There's no fast sync, so you would join the network as a node, and while you catch up, you do additional processing to check invariants. Our ABCI mux system allows us to put these checks in a mux app.

cc #2277


kostko:
Probably also related to #1633. cc @abukosek

pro-wh:
new in this push:
- checking not on every block. checking on some interval. and it's implemented in this way (which nobody asked for) where it randomly chooses one height in each interval
- starting to port some sanity checks from genesis-time

pro-wh:
from Friday and additional notes:
- mostly I've adapted genesis-time sanity checks by extracting functions that operate on abstract types. at genesis time, we call them with parts of the genesis document. at follow time, we load from the mux apps' state modules. there's some code sharing, but we'll have to be careful in the future to make sure we call these sanity checks from both places when we add new ones.
- the mux apps' state modules tend to expose methods that load entire datasets into memory, e.g. all delegation records. that's consistent with the design of the genesis-time sanity checks, which operate on an all-in-memory genesis document. for now I'm going ahead and doing things that way

pro-wh:
new in this push and from Monday:
- fixes to increase unit test fidelity so that they pass followtool's sanity checks
- changes to sanity checks tests to make the combined system not self-contradictory
- specifying a key manager in a runtime record is now optional. key-manager-type runtimes now don't specify a (meta?) key manager. previously they would specify that they are their own key manager.
- transfers of staking tokens to/from a runtime through roothash messages now adjust the staking token's total supply so that the staking mux app's state will still add up right after the transfer. this is intended to be an interim consistent behavior before we decommission runtime transfers for #2377.

kostko:
> specifying a key manager in a runtime record is now optional. key-manager-type runtimes now don't specify a (meta?) key manager. previously they would specify that they are their own key manager.

So this also fixes #2459?

go/genesis/api/api_test.go
```diff
@@ -136,9 +136,9 @@ func TestGenesisSanityCheck(t *testing.T) {
 
 	testRuntimeID := hex2pk("0000000000000000000000000000000000000000000000000000000000000001")
 	testRuntime := &registry.Runtime{
-		ID:         testRuntimeID,
-		Kind:       registry.KindCompute,
-		KeyManager: testKMRuntime.ID,
+		ID:            testRuntimeID,
+		Kind:          registry.KindCompute,
+		KeyManagerOpt: &testKMRuntime.ID,
```
kostko: Why the rename to `KeyManagerOpt`? Whether something is optional or required shouldn't be in the name.

pro-wh:
> So this also fixes #2459 or it just makes the key manager optional?

just makes the key manager optional

go/genesis/api/api_test.go
pro-wh: oh I did that to make sure I would notice all usages. I'll change it back

pro-wh:
in the latest push:
- added more corrections to the node and to the checks to make sure we always pass the checks

ready for review

go/keymanager/api/api.go
```diff
@@ -136,9 +136,8 @@ type Genesis struct {
 	Statuses []*Status `json:"statuses,omitempty"`
 }
 
-// SanityCheck does basic sanity checking on the genesis state.
-func (g *Genesis) SanityCheck() error {
-	for _, status := range g.Statuses {
+func SanityCheckStatuses(statuses []*Status) error {
```
kostko: Add documentation block (also all other split `Sanity*` methods).

go/consensus/tendermint/apps/followtool/api.go
```diff
@@ -0,0 +1,6 @@
+package followtool
+
+const (
+	AppID   uint8  = 0x98
```
kostko: Add documentation blocks.

go/keymanager/api/api.go
pro-wh: ok

go/consensus/tendermint/apps/followtool/api.go
pro-wh: ok

pro-wh:
renamed

pro-wh:
ok all lights greenish. gonna merge

PR #2343, opened 2019-11-06
tests: Make E2E test coverage computation optional

tjanez:
Enable it on Buildkite.

### DIFF
--- a/.buildkite/scripts/test_e2e.sh
+++ b/.buildkite/scripts/test_e2e.sh
@@ -28,11 +28,18 @@ if [[ "${BUILDKITE:-""}" != "" ]]; then
     mkdir -p ${WORKDIR}/e2e
 fi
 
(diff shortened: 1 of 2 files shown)


.buildkite/scripts/test_e2e.sh
```diff
@@ -28,11 +28,18 @@ if [[ "${BUILDKITE:-""}" != "" ]]; then
     mkdir -p ${WORKDIR}/e2e
 fi
 
+# Use integrationrunner-wrapper.sh as node binary if we need to compute E2E
+# tests' coverage.
+node_binary=""
```
kostko: Hm, default is no binary? That won't work?

tjanez: Yea, submitted the PR too fast, fixed it :slightly_smiling_face:

PR #2345, opened 2019-11-06
go/oasis-node/cmd/registry/node: Make consensus address' ID@ optional

### LINKED ISSUE #2341: Registry node init command should not require consensus ID
Currently provisioning a new consensus node (without having a pre-provisioned identity) fails with: `failed to parse node's consensus address" err="node: consensus address doesn't have ID@ part`. 

The command should probably work in a way that allows overriding the `ID`, but by default uses the newly generated one.

tjanez:
Fixes #2341.

This should enable provisioning consensus nodes as before (i.e. without having a pre-provisioned identity).

### DIFF
--- a/go/oasis-node/cmd/registry/node/node.go
+++ b/go/oasis-node/cmd/registry/node/node.go
@@ -232,11 +232,14 @@ func doInit(cmd *cobra.Command, args []string) {
 
 		var consensusAddr node.ConsensusAddress
 		if err = consensusAddr.UnmarshalText([]byte(addrStr)); err != nil {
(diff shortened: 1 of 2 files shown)


tjanez:
> I think this workaround is not needed anymore either

Yes, that's true. But I need it for #2300, so I would rather just leave it in.

tjanez:
> Yes, that's true. But I need it for #2300, so I would rather just leave it in.

It turns out that pre-provisioning the validator's identity is not required for #2300 (@ptrus, thanks for persuading me :slightly_smiling_face:!).

I've updated the PR, please take another look.

Issue #2344, opened 2019-11-06
e2e coverage broke make e2e-test and stdout of CLI


matevz:
PR https://github.com/oasislabs/oasis-core/pull/2229 broke two things:
* [x] running `make test-e2e` out of the box fails with entity ID not given error. It seems you need to run `cd go; make integrationrunner` first. Please either update the README or put this dependency in the Makefile.
* [x] getting the node binary with `net.Config().NodeBinary` in e2e tests returns the path to the `integrationrunner-wrapper.sh` now which also appends a human-readable coverage report to stdout. I am parsing stdout in my new `stake-cli` e2e tests and expect a strictly formatted output. Please either remove any output from the wrapper or provide a way to run `oasis-node` directly without the wrapper.
* [x] remove generated `coverage-e2e-XXXXXXXXXX.txt` files from the source directory. Putting them in `/tmp/oasis-test-runnerXXXXXXXXX` seems a better idea.

pro-wh:
#2343 fixes the default invocation

pro-wh:
the location of coverage output files is consistent with unit tests with coverage. both put them in the checkout. let's not change that as a one-off.

pro-wh:
should run without the extra output now

PR #2347, opened 2019-11-06
go oasis-node: close stdout before test framework prints junk

pro-wh:
blocks #2344

there's no option to disable messages from the test framework

### DIFF
--- a/go/oasis-node/integrationrunner/integration_test.go
+++ b/go/oasis-node/integrationrunner/integration_test.go
@@ -4,6 +4,8 @@ import (
 	"flag"
 	"os"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
 var run = flag.Bool("integration.run", false, "Run the node")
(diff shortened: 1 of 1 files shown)


go/oasis-node/integrationrunner/integration_test.go
```diff
@@ -16,4 +17,7 @@ func TestIntegration(t *testing.T) {
 	}
 
 	launch()
+
+	// Suppress coverage report on stdout.
+	_ = os.Stdout.Close()
```
Yawning: Do we only care about stdout?

Also, `os.Stdout = ioutil.Discard` might be slightly better behaved, though file descriptor re-use is unlikely.

pro-wh: I wanted that too, but the types don't match.

pro-wh: and yeah the issue that this fixes is only about stdout.

Yawning: Oh, it's an `os.File`, so something like:
```
  f, err := os.OpenFile(os.DevNull, os.O_RW, 0)
  if err != nil {
    t.Fatalf("Failed to open /dev/null") // Use testify if you want, whatever.
  }
  os.Stdout = f
```

pro-wh: ok

PR #2353, opened 2019-11-08
Staking commission

pro-wh:
it was said that rewards should have some portion split off as "commission" and paid to the node operator, i.e., the entity that owns the node. that's the account where we keep the escrow pools.

it was said that one should only be able to change the commission rate once per some interval. that interval is modeled in the consensus parameter CommissionRateChangeInterval.

it was said that one should have to post a future bound on the commission rate. this distance into the future is modeled in the consensus parameter CommissionRateBoundLead.

each entity's schedule of commission rates and bounds are represented in a CommissionSchedule structure in the account's new `.Escrow.CommissionSchedule` field.

there's a new transaction where an entity can amend its own commission rate and bound schedule. in this iteration, we charge a fixed gas fee for it, no matter the complexity of the schedule. when we process the amendment, we also prune old schedule information that is no longer in effect.

### DIFF
--- a/go/consensus/tendermint/apps/staking/transactions.go
+++ b/go/consensus/tendermint/apps/staking/transactions.go
@@ -362,13 +362,6 @@ func (app *stakingApplication) amendCommissionSchedule(ctx *abci.Context, state 
 		return err
 	}
 
-	commit := true
-	defer func() {
-		if commit {
-			state.SetAccount(id, from)
-		}
-	}()
-
 	epoch, err := app.state.GetEpoch(ctx.Ctx(), ctx.BlockHeight()+1)
 	if err != nil {
 		return err
@@ -379,8 +372,10 @@ func (app *stakingApplication) amendCommissionSchedule(ctx *abci.Context, state 
 			"err", err,
 			"from", id,
 		)
-		commit = false
+		return err
 	}
+
+	state.SetAccount(id, from)
 
 	return nil
 }
--- a/go/consensus/tendermint/apps/staking/transactions.go
+++ b/go/consensus/tendermint/apps/staking/transactions.go
@@ -332,3 +332,55 @@ func (app *stakingApplication) reclaimEscrow(ctx *abci.Context, state *stakingSt
 
 	return nil
 }
+
+func (app *stakingApplication) amendCommissionSchedule(ctx *abci.Context, state *stakingState.MutableState, signedAmendCommissionSchedule *staking.SignedAmendCommissionSchedule) error {
+	var amendCommissionSchedule staking.AmendCommissionSchedule
+	if err := signedAmendCommissionSchedule.Open(staking.AmendCommissionScheduleSignatureContext, &amendCommissionSchedule); err != nil {
+		app.logger.Error("ReclaimEscrow: invalid signature",
+			"signed_amend_commission_schedule", signedAmendCommissionSchedule,
+		)
+		return staking.ErrInvalidSignature
+	}
+
+	// Authenticate sender and make sure fees are paid.
+	id := signedAmendCommissionSchedule.Signature.PublicKey
+	from, err := stakingState.AuthenticateAndPayFees(ctx, state, id, amendCommissionSchedule.Nonce, &amendCommissionSchedule.Fee)
+	if err != nil {
+		return err
+	}
+
+	if ctx.IsCheckOnly() {
+		return nil
+	}
+
+	// Charge gas for this transaction.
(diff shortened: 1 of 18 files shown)


pro-wh:
open for review

go/consensus/tendermint/apps/staking/transactions.go
```diff
@@ -332,3 +332,55 @@ func (app *stakingApplication) reclaimEscrow(ctx *abci.Context, state *stakingSt
...
+	if err != nil {
+		return err
+	}
+
+	if err = from.Escrow.CommissionSchedule.AmendAndPruneAndValidate(&amendCommissionSchedule.Amendment, epoch, params.CommissionRateChangeInterval, params.CommissionRateBoundLead); err != nil {
+		app.logger.Error("AmendCommissionSchedule: amendment not acceptable",
+			"err", err,
+			"from", id,
+		)
+		commit = false
```
kostko: This should probably `return err`.

go/staking/api/api.go
```diff
@@ -451,8 +489,9 @@ type GeneralAccount struct {
 // EscrowAccount is an escrow account the balance of which is subject to
 // special delegation provisions and a debonding period.
 type EscrowAccount struct {
-	Active    SharePool `json:"active"`
-	Debonding SharePool `json:"debonding"`
+	Active             SharePool          `json:"active"`
+	Debonding          SharePool          `json:"debonding"`
+	CommissionSchedule CommissionSchedule `json:"commission_schedule"`
```
kostko: Update `SanityCheck` below to verify all commission schedules in genesis.

go/staking/api/commission.go
```diff
@@ -0,0 +1,269 @@
...
+	RateMax quantity.Quantity   `json:"rate_max"`
+}
+
+type CommissionSchedule struct {
+	Rates  []CommissionRateStep      `json:"rates"`
+	Bounds []CommissionRateBoundStep `json:"bounds"`
+}
+
+// validateNondegenerate detects degenerate steps.
+func (cs *CommissionSchedule) validateNondegenerate(commissionRateChangeInterval epochtime.EpochTime) error {
```
kostko: We should probably limit the maximum number of steps. An implicit limit is the transaction size, but maybe we should choose a sensible higher-level limit as well.

go/consensus/tendermint/apps/staking/transactions.go
pro-wh: does it keep the transaction fee update if I do?

kostko: Yes.

pro-wh: ok

go/staking/api/api.go

go/staking/api/commission.go
pro-wh: ok I'll add that

pro-wh:
what does this break

kostko:
> what does this break

It introduces a new transaction type and new consensus parameters. A node running this version would therefore process the new transaction differently from nodes that are running a previous version potentially leading to different state hashes.

pro-wh:
k I'll bump it

go/staking/api/api.go
```diff
@@ -39,6 +39,9 @@ var (
 	// ReclaimEscrowSignatureContext is the context used for escrow reclimation.
 	ReclaimEscrowSignatureContext = signature.NewContext("oasis-core/staking: reclaim escrow", signature.WithChainSeparation())
 
+	// AmendCommissionScheduleSignatureContext is the context used for escrow reclimation.
+	AmendCommissionScheduleSignatureContext = signature.NewContext("oasis-core/staking: amend commission schedule")
```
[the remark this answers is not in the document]
pro-wh: ok added

pro-wh:
new e2e test for the transaction generator and submission!

go/staking/api/api.go
```diff
@@ -241,6 +247,24 @@ func (r *ReclaimEscrow) UnmarshalCBOR(data []byte) error {
...
 
+// AmendCommissionSchedule is an amendment to a commission schedule.
+type AmendCommissionSchedule struct {
+	Nonce uint64  `json:"nonce"`
+	Fee   gas.Fee `json:"fee"`
+
+	Amendment CommissionSchedule `json:"amendment"`
+}
+
+// MarshalCBOR serializes the type into a CBOR byte vector.
```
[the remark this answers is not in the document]

PR #2360, opened 2019-11-08
reorganize consensus backends

ptrus:
Closes: https://github.com/oasislabs/oasis-core/issues/1879

TODO:
- [x] fix byzantine tests setup


ptrus:
ready for review (probably some more things could be cleaned up a bit)

go/oasis-node/cmd/node/node.go
```diff
@@ -570,8 +542,18 @@ func newNode(testNode bool) (*Node, error) {
 		}
 		node.svcMgr.Register(node.svcTmnt)
 		node.Epochtime = node.svcTmnt.EpochTime()
-
-		// Initialize the various node backends.
+		node.Beacon = node.svcTmnt.Beacon()
+		node.KeyManager = node.svcTmnt.KeyManager()
+		node.Registry = node.svcTmnt.Registry()
+		node.svcMgr.RegisterCleanupOnly(node.Registry, "registry backend")
```
kostko: This doesn't seem right? The tendermint service should dispatch cleanup to its own backends?

go/consensus/tendermint/tendermint.go
```diff
@@ -33,21 +34,29 @@ import (
...
-	staking "github.com/oasislabs/oasis-core/go/staking/api"
-	"github.com/oasislabs/oasis-core/go/tendermint/abci"
-	"github.com/oasislabs/oasis-core/go/tendermint/api"
-	"github.com/oasislabs/oasis-core/go/tendermint/crypto"
-	"github.com/oasislabs/oasis-core/go/tendermint/db"
-	"github.com/oasislabs/oasis-core/go/tendermint/epochtime"
-	"github.com/oasislabs/oasis-core/go/tendermint/epochtime_mock"
-	"github.com/oasislabs/oasis-core/go/tendermint/service"
+	keymanagerAPI "github.com/oasislabs/oasis-core/go/keymanager/api"
+	"github.com/oasislabs/oasis-core/go/registry"
```
kostko: So some are named `<backend>API` and then the tendermint version is imported as `<backend>` but some are named `<backend>` and then the tendermint version is imported as `tm<backend>`. I think it should be consistent.

ptrus: I changed to: api packages are all `<backend>API`, tendermint backends are all `tm<backend>`

go/oasis-node/cmd/node/node.go
ptrus: added a service manager in the tendermint service and this is done there now. Changes are in a separate commit for easier review (will fixup the "cr fixes" commit before merging)

PR #2361, opened 2019-11-09
go: Bump the ed25519 and tendermint dependencies

Yawning:
I made a really stupid mistake in the ed25519 code that required a minor
fix.  This pulls in the new ed25519 and tendermint versions.


Yawning:
Draft because the tendermint tag doesn't exist yet, and so I can't commit the modified go.sum yet.

Issue #1693, opened 2019-05-10
Properly allocate runtime IDs


Yawning:
This is more of a policy decision than an actual coding one.  For the sake of sanity, we should start properly allocating runtime IDs.  Currently absolutely everything uses `0000000000000000000000000000000000000000000000000000000000000000` which is sub-optimal to the extreme.

I'm tempted to explicitly reject the default runtime ID value since that's what the uninitialized value will be on the Go side, but that might be overly disruptive, and it will be pointless if people just start using runtime ID `1` instead.

Things that need runtime IDs:
 * runtime-ethereum
 * simple-keyvalue
 * keymanager

Note: It's not inconceivable that the same runtime will have multiple runtime IDs, since it doubles as a deployment identifier.  Test runtimes that should never see production deployment should probably live in a dedicated part of the runtime ID space as well.

kostko:
I think we should segment runtime identifiers as follows:
* The first byte of the runtime ID is reserved and can only be set to one of the predefined values. This is enforced by the consensus layer and other applications relying on runtime kinds.
* The first byte is interpreted as a bit vector, each bit specifying a flag. Bit 0 is the most significant bit.
* Bit 0 specifies whether the runtime is a test runtime (0 = production, 1 = test). Test runtimes are only allowed if a registry consensus parameter `DebugAllowTestRuntimes` is set. We could additionally only allow debug enclaves to be used with test runtimes.
* Bit 1 specifies whether the runtime is a key manager runtime (0 = not key manager, 1 = key manager). Only key manager runtime IDs are allowed to be used for the key manager runtime, enforced by the registry.
* Bit 2 specifies whether the runtime is a stateless runtime (0 = stateful, 1 = stateless).
* Bits 3-7 are reserved and must be set to zero.

Does this give us enough room for future extensions or do we need to reserve more than the first byte?

Yawning:
Does the rest of the runtime ID need to have high entropy?  Otherwise I'd say we can reserve more bits (first 32 bits), because we do have 256 bits to work with.

kostko:
No, runtime IDs don't need high entropy IMO, so yeah I agree that reserving more makes sense here.

Yawning:
Fixed by https://github.com/oasislabs/oasis-core/pull/2530

PR #2365, opened 2019-11-11
Add sanity checks for the genesis state

abukosek:
Part of #1633.

### DIFF
--- a/go/genesis/api/api.go
+++ b/go/genesis/api/api.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"strings"
 	"time"
 
 	beacon "github.com/oasislabs/oasis-core/go/beacon/api"
@@ -77,6 +78,53 @@ func (d *Document) WriteFileJSON(filename string) error {
 	if err = ioutil.WriteFile(filename, docJSON, filePerm); err != nil {
 		return fmt.Errorf("WriteFileJSON: failed to write genesis file: %w", err)
 	}
+	return nil
+}
+
+// SanityCheck does basic sanity checking on the contents of the genesis document.
+func (d *Document) SanityCheck() error {
+	if d.Height < 0 {
+		return fmt.Errorf("genesis: sanity check failed: height must be >= 0")
+	}
+
+	if d.Time.After(time.Now()) {
+		return fmt.Errorf("genesis: sanity check failed: time of genesis document is in the future")
+	}
+
+	if strings.TrimSpace(d.ChainID) == "" {
+		return fmt.Errorf("genesis: sanity check failed: chain ID must not be empty")
+	}
+
+	var err error
+	if err = d.EpochTime.SanityCheck(); err != nil {
+		return err
+	}
+	if err = d.Registry.SanityCheck(); err != nil {
+		return err
+	}
+	if err = d.RootHash.SanityCheck(); err != nil {
+		return err
+	}
+	if err = d.Staking.SanityCheck(); err != nil {
+		return err
+	}
+	if err = d.KeyManager.SanityCheck(); err != nil {
+		return err
+	}
+	if err = d.Scheduler.SanityCheck(); err != nil {
+		return err
+	}
+	if err = d.Beacon.SanityCheck(); err != nil {
+		return err
+	}
+	if err = d.Consensus.SanityCheck(); err != nil {
+		return err
+	}
+
+	if d.HaltEpoch < d.EpochTime.Base {
+		return fmt.Errorf("genesis: sanity check failed: halt epoch is in the past")
+	}
+
 	return nil
 }
 
--- a/go/keymanager/api/api.go
+++ b/go/keymanager/api/api.go
@@ -4,6 +4,7 @@ package api
 import (
 	"context"
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/oasislabs/oasis-core/go/common/cbor"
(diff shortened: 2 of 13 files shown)


go/common/consensus/genesis/genesis.go
```diff
@@ -11,3 +14,16 @@ type Genesis struct {
 	EmptyBlockInterval time.Duration `json:"empty_block_interval"`
 	MaxTxSize          uint          `json:"max_tx_size"`
 }
+
+// SanityCheck does basic sanity checking on the genesis state.
+func (g *Genesis) SanityCheck() error {
+	if g.Backend != "tendermint" {
```
Yawning: constify "tendermint"?

go/common/crypto/signature/signature.go
```diff
@@ -609,33 +609,33 @@ func BuildPublicKeyBlacklist(allowTestKeys bool) {
...
-		// p (=0, order 4).
-		"edffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7f",
-		// p+1 (=1, order 1).
-		"eeffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7f",
-	} {
-		var pk PublicKey
-		if err := pk.UnmarshalHex(v); err != nil {
-			panic(err)
-		}
+		// Explicitly forbid other keys here.
```
Yawning: The small order points should unconditionally be rejected.

I assume you did this because you want to use `IsValid`.  There are certain things (namedly runtime IDs) that are currently `signature.PublicKey`, that should not be.  The solution here is not to accept invalid public keys in the public key verification routine, but to fix the types to stop expressing things as public keys when they are not (See #1693).

go/genesis/api/api.go
```diff
@@ -80,6 +81,54 @@ func (d *Document) WriteFileJSON(filename string) error {
 	return nil
 }
 
+// SanityCheck does basic sanity checking on the contents of the genesis document.
+func (d *Document) SanityCheck() error {
+	if d.Height < 0 {
+		return fmt.Errorf("genesis: sanity check failed: height must be >= 0")
+	}
+
+	// TODO: Maybe add some epsilon here to account for network latency and time drift?
```
Yawning: Don't think this is required.  The genesis document will be generated a good amount of time prior to the sanity check being applied.

go/keymanager/api/api.go
```diff
@@ -130,6 +131,39 @@ type Genesis struct {
...
+
+		// Verify currently active key manager node IDs.
+		for _, node := range status.Nodes {
+			if !node.IsValid() {
+				return fmt.Errorf("keymanager: sanity check failed: key manager node ID %s is invalid", node.String())
+			}
+		}
+
+		// Verify SGX policy signatures if the policy exists.
+		if status.Policy != nil {
```
Yawning: Hm, a key manager without a policy is largely nonsensical, but we can't tell if enforcement is disabled easily. :(

go/registry/api/api.go
```diff
@@ -908,3 +908,110 @@ type ConsensusParameters struct {
...
+
+		if !rt.ID.IsValid() {
+			return fmt.Errorf("registry: sanity check failed: runtime ID %s is invalid", rt.ID.String())
+		}
+
+		if rt.RegistrationTime > uint64(time.Now().Unix()+61*60) {
+			return fmt.Errorf("registry: sanity check failed: runtime ID %s registration time is more than 1h1m in the future", rt.ID.String())
+		}
+
+		if !rt.KeyManager.IsValid() {
```
Yawning: Should probably also cross-check that the keymanager is actually a valid key manager runtime.

go/registry/api/api.go
```diff
@@ -908,3 +908,110 @@ type ConsensusParameters struct {
...
+
+		if !seenEntities[n.EntityID.ToMapKey()] {
+			return fmt.Errorf("registry: sanity check failed: node ID %s has unknown controlling entity", n.ID.String())
+		}
+
+		if n.RegistrationTime > uint64(time.Now().Unix()+61*60) {
+			return fmt.Errorf("registry: sanity check failed: node ID %s registration time is more than 1h1m in the future", n.ID.String())
+		}
+
+		for _, rt := range n.Runtimes {
```
Yawning: Certain node roles should not have any runtimes, certain roles should/must have runtimes.

go/roothash/api/api.go
```diff
@@ -193,3 +194,74 @@ type TransactionSchedulerParameters struct {
...
 	MaxBatchSizeBytes uint64 `json:"max_batch_size_bytes"`
 }
+
+// SanityCheck does basic sanity checking on the genesis state.
+func (g *Genesis) SanityCheck() error {
+	if g.Parameters.RoundTimeout < 1*time.Second {
+		return fmt.Errorf("roothash: sanity check failed: round timeout must be >= 1 second")
+	}
+
+	if g.Parameters.TransactionScheduler.Algorithm != "batching" {
```
Yawning: constify "batching"

abukosek: I couldn't do this nicely because of import cycles, but I added a comment to point to the actual constant that should be used here.

go/common/crypto/signature/signature.go
abukosek: Yeah, `IsValid` was failing on all-zeros runtime IDs.  I've removed the `IsValid` checks for runtime IDs and reverted the change in this file -- once #1693 is done and all runtime IDs are changed to be of type `common.Namespace`, those checks wouldn't have made any sense anyways :)

abukosek:
Comments addressed, ready for re-review :)

go/staking/api/api.go
```diff
@@ -520,6 +520,95 @@ func (p *ConsensusParameters) SanityCheck() error {
 	return nil
 }
 
+// SanityCheck does basic sanity checking on the genesis state.
+func (g *Genesis) SanityCheck() error {
```
Yawning: Should this enforce the new lower limit added in #2366?

abukosek: It looks like that is already sufficiently enforced when adding the escrow, so not sure if it would be beneficial to duplicate the check here.  Also, if the parameter changes, the genesis document might become unloadable on a dump/restore if it were enforced on this level.

Yawning: Ok.

go/genesis/api/api.go
```diff
@@ -80,6 +81,53 @@ func (d *Document) WriteFileJSON(filename string) error {
 	return nil
 }
 
+// SanityCheck does basic sanity checking on the contents of the genesis document.
+func (d *Document) SanityCheck() error {
+	if d.Height < 0 {
```
kostko: Shouldn't we just make the height an `uint64`?

go/roothash/api/api.go
kostko: Maybe define the valid algorithm names here (e.g., in `go/roothash/api`) instead and use the constant in the worker which needs to import it anyway?

go/genesis/api/api.go
abukosek: I think it's set as `int64` for compatibility with Tendermint -- having it as `uint64` might just overflow silently.

abukosek:
Thanks to everyone for the comments!  As agreed, the tests for sanity checking will be part of a separate PR to make them easier to review.

PR #2370, opened 2019-11-12
go/tendermint/apps/scheduler: Change validator elections again

Yawning:
Meetings happened when I was asleep, so change the algorithm to a
sensible version of what was decided on.

Validators are now picked, 1 per entity (at random if multiple candiates
ran by a single entity are available), in order of decending stake, up
to the maximum number of validators.

Unlike the previous version of this where eligibility was also
restricted to the top-N entities by stake, this election algorithm will
continue to go down the entity list if required in an attempt to fill
out the validator set.

Incentives are now "any entity that is running a validator at the time
of the election, that meets the minimum stake threshold", rather than
being restricted to "top-N" as well.

Note: The "1 per entity" limit is configurable.  If more are allowed,
then each entity (in descending order) will have the maximum number of
nodes that they have the stake to support be elected.  This isn't great
but the limit being higher should only ever happen in test environments
anyway.

### DIFF
--- a/go/tendermint/apps/scheduler/scheduler.go
+++ b/go/tendermint/apps/scheduler/scheduler.go
@@ -560,7 +560,7 @@ func (app *schedulerApplication) electAllCommittees(ctx *abci.Context, request t
 func (app *schedulerApplication) electValidators(ctx *abci.Context, beacon []byte, entityStake *stakeAccumulator, entitiesEligibleForReward map[signature.MapKey]bool, nodes []*node.Node, params *scheduler.ConsensusParameters) error {
 	// Filter the node list based on eligibility and minimum required
 	// entity stake.
-	var preFilteredNodeList []*node.Node
+	var nodeList []*node.Node
 	entMap := make(map[signature.MapKey]bool)
 	for _, n := range nodes {
 		if !n.HasRoles(node.RoleValidator) {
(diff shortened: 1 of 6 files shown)


Yawning:
This will break existing deployments, since we use a single entity, unless we're ok with our test deployments only having one validator elected.  Set the self explanatory configuration flag to something sensible.

go/tendermint/apps/scheduler/scheduler.go
```diff
@@ -569,37 +569,26 @@ func (app *schedulerApplication) electValidators(ctx *abci.Context, beacon []byt
...
 			continue
 		}
-		preFilteredNodeList = append(preFilteredNodeList, n)
+		nodeList = append(nodeList, n)
 		entMap[n.EntityID.ToMapKey()] = true
+
+		// Everyone that "runs" a validator node is eligible for rewards.
+		//
+		// Yes, this is stupid in that, there's no proof that the node is
+		// actually running.
```
kostko: Yeah so even if the node is actually scheduled and it is not running it will not be penalized in any way (except not getting fees) as we are not slashing for liveness. So this is just distributing rewards to people that have registered "nodes" and have some stake bonded.

Yawning: I stand by "this is stupid".

ptrus:
cc @ravenac95 if we plan to run more than one validator in the public testnet, each should have it's own entity

Yawning:
Well, you can run more than one validator, it's just that only one will get elected.

PR #2371, opened 2019-11-12
go/staking: Slash by fixed amount

kostko:


Yawning:
Ok.  Do we want to have the flexibility to specify, "slash absolutely everything"?  (I guess that can happen when we need it...)

PR #2374, opened 2019-11-13
Write signed registry txs to disk

ravenac95:
I modeled this as closely as possible to the `stake` subcommand but I hacked it together quickly cause I found this was a pain point for joining the testnet. See the section here: https://github.com/oasislabs/docs/blob/9428add3df027ba1ab3d7a96013eed895a20fa8d/src/operators/joining-the-testnet.md#registering-you-entity

### DIFF
--- a/go/oasis-node/cmd/registry/entity/entity.go
+++ b/go/oasis-node/cmd/registry/entity/entity.go
@@ -28,11 +28,10 @@ import (
 )
 
 const (
-	cmdRegister = "register"
-
 	cfgAllowEntitySignedNodes = "entity.debug.allow_entity_signed_nodes"
 	cfgNodeID                 = "entity.node.id"
 	cfgNodeDescriptor         = "entity.node.descriptor"
+	cfgTxFile                 = "entity.transaction.file"
 
 	entityGenesisFilename = "entity_genesis.json"
 )
@@ -42,6 +41,8 @@ var (
 	initFlags                 = flag.NewFlagSet("", flag.ContinueOnError)
 	updateFlags               = flag.NewFlagSet("", flag.ContinueOnError)
 	registerOrDeregisterFlags = flag.NewFlagSet("", flag.ContinueOnError)
+	submitFlags               = flag.NewFlagSet("", flag.ContinueOnError)
+	txFileFlags               = flag.NewFlagSet("", flag.ContinueOnError)
 
 	entityCmd = &cobra.Command{
 		Use:   "entity",
@@ -61,15 +62,15 @@ var (
 	}
 
 	registerCmd = &cobra.Command{
-		Use:   cmdRegister,
-		Short: "register an entity",
-		Run:   doRegisterOrDeregister,
+		Use:   "gen_register",
+		Short: "generate a register entity transaction",
+		Run:   doGenRegister,
 	}
 
 	deregisterCmd = &cobra.Command{
-		Use:   "deregister",
-		Short: "deregister an entity",
-		Run:   doRegisterOrDeregister,
+		Use:   "gen_deregister",
+		Short: "generate a deregister entity transaction",
+		Run:   doGenDeregister,
 	}
 
 	listCmd = &cobra.Command{
@@ -78,8 +79,21 @@ var (
 		Run:   doList,
 	}
 
+	submitCmd = &cobra.Command{
+		Use:   "submit",
+		Short: "submit a pre-generated transaction",
+		Run:   doSubmit,
+	}
+
 	logger = logging.GetLogger("cmd/registry/entity")
 )
+
+type serializedTx struct {
+	Register   *entity.SignedEntity `json:"register,omitempty"`
+	Deregister *signature.Signed    `json:"deregister,omitempty"`
+}
+
+type generateEntitySignedTxFunc func(txFilePath string, ent *entity.Entity, signer signature.Signer) (*serializedTx, error)
 
 func doConnect(cmd *cobra.Command) (*grpc.ClientConn, grpcRegistry.EntityRegistryClient) {
 	conn, err := cmdGrpc.NewClient(cmd)
@@ -268,9 +282,35 @@ func signAndWriteEntityGenesis(dataDir string, signer signature.Signer, ent *ent
 	return nil
 }
 
-func doRegisterOrDeregister(cmd *cobra.Command, args []string) {
+func saveSignedTx(tx *serializedTx, txFilePath string) error {
+	b, _ := json.Marshal(tx)
+	if err := ioutil.WriteFile(txFilePath, b, 0600); err != nil {
+		return err
+	}
+	return nil
+}
+
+func loadSignedTx(txFilePath string) (*serializedTx, error) {
+	var tx serializedTx
+	b, err := ioutil.ReadFile(txFilePath)
+	if err != nil {
+		return nil, err
+	}
+	if err = json.Unmarshal(b, &tx); err != nil {
+		return nil, err
+	}
+	return &tx, nil
+}
+
+func doGenerateEntitySignedTx(signedTxFunc generateEntitySignedTxFunc, cmd *cobra.Command, args []string) {
 	if err := cmdCommon.Init(); err != nil {
 		cmdCommon.EarlyLogAndExit(err)
+	}
+
+	txFilePath := viper.GetString(cfgTxFile)
+	if txFilePath == "" {
+		logger.Error("must specify an --" + cfgTxFile)
+		os.Exit(1)
 	}
 
 	dataDir, err := cmdCommon.DataDirOrPwd()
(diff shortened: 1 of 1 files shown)


go/go.sum
```diff
@@ -351,6 +351,7 @@ github.com/oasislabs/iavl v0.12.0-ekiden2 h1:kzdp6OTHYrn8TXlAo+2NsL84Qi2gqydFvjh
 github.com/oasislabs/iavl v0.12.0-ekiden2/go.mod h1:B/tMpl5cg7n42n3xYQTCckJzQezoI75jedkc8FOiOF0=
 github.com/oasislabs/iavl v0.12.0-ekiden3 h1:8544fXJb57urhAEpTlIwDBdTJukgpPS/FCS/yj14I8E=
 github.com/oasislabs/iavl v0.12.0-ekiden3/go.mod h1:B/tMpl5cg7n42n3xYQTCckJzQezoI75jedkc8FOiOF0=
+github.com/oasislabs/oasis-core v0.0.0-20191112165703-4d36eb49f175 h1:7JTlbjFLTDZ/i6eTgg2AHNwXiGrSUZ7FTocd0HY1M0w=
```
Yawning: Why?  (revert this change)

ravenac95:
> I guess? I'm not sure how useful this will be in practice, registry transactions have a definite shelf life due to the timestamp.

Ya, the timestamp was something I was wondering about. That being said, the issue is mostly around keeping the entity's private key offline. For simplicity's sake, I wanted to keep the instructions for joining the testnet restricted to using the node that the operator is deploying. So the options for registering an entity after the genesis file has been set: 

1. Have the node operator do the registration _on_ the deployed node (didn't want to suggest that)
2. Allow entity registration over the internet via something we provide or (like we do currently) ssh tunneling to the deployed node. 
3. Allow for the signing of a registry request (this PR)

Yawning:
The better long term solution is probably a combination of:

 * Allow entity registration over the internet via something we provide
 * Delaying the node registration till the node is synced (should be done already), AND the registry contains the node's owning entity (with the node listed in the entity descriptor), while allowing the node to submit txes.

But I guess this is ok for now.

Yawning:
Before I approve this, should using a saved transaction file (the 2 step process) be mandatory like how it is in the staking client?

ravenac95:
> Before I approve this, should using a saved transaction file (the 2 step process) be mandatory like how it is in the staking client?

Hrm. I personally don't have an opinion on the matter. I would only say it might make sense to make it mandatory for consistency's sake.

kostko:
I would be for mandatory.

ravenac95:
> I would be for mandatory.

Sounds good

go/oasis-node/cmd/registry/entity/entity.go
```diff
@@ -33,6 +33,7 @@ const (
 	cfgAllowEntitySignedNodes = "entity.debug.allow_entity_signed_nodes"
 	cfgNodeID                 = "entity.node.id"
 	cfgNodeDescriptor         = "entity.node.descriptor"
+	cfgSignedFile             = "entity.signed_tx_file"
```
kostko: Consistency with staking CLI:
```suggestion
	cfgTxFile             = "entity.transaction.file"
```

go/oasis-node/cmd/registry/entity/entity.go
```diff
@@ -42,6 +43,8 @@ var (
 	initFlags                 = flag.NewFlagSet("", flag.ContinueOnError)
 	updateFlags               = flag.NewFlagSet("", flag.ContinueOnError)
 	registerOrDeregisterFlags = flag.NewFlagSet("", flag.ContinueOnError)
+	submitFlags               = flag.NewFlagSet("", flag.ContinueOnError)
+	signedTxFlags             = flag.NewFlagSet("", flag.ContinueOnError)
```
kostko: Consistency with staking CLI:
```suggestion
	txFileFlags             = flag.NewFlagSet("", flag.ContinueOnError)
```

go/oasis-node/cmd/registry/entity/entity.go
```diff
@@ -320,6 +354,48 @@ func doRegisterOrDeregister(cmd *cobra.Command, args []string) {
 	os.Exit(1)
 }
 
+func doSubmit(cmd *cobra.Command, args []string) {
+	signedTxPath := viper.GetString(cfgSignedFile)
+	if signedTxPath == "" {
+		logger.Error("must specify an --entity.signed_tx_file")
```
kostko:
```suggestion
		logger.Error("must specify an --" + cfgTxFile)
```

Yawning:
Don't forget to update `registerCmd.Short` and `deregisterCmd.Short` to reflect the new behavior (See how the staking app describes the operations).

go/oasis-node/cmd/registry/entity/entity.go
```diff
@@ -61,15 +64,17 @@ var (
 	}
 
 	registerCmd = &cobra.Command{
-		Use:   cmdRegister,
-		Short: "register an entity",
-		Run:   doRegisterOrDeregister,
+		Use:     cmdRegister,
+		Aliases: []string{"register"},
```
kostko: Why aliases?

go/oasis-node/cmd/registry/entity/entity.go
```diff
@@ -61,15 +64,17 @@ var (
...
+		Short:   "generate a register entity transaction",
+		Run:     doRegisterOrDeregister,
 	}
 
 	deregisterCmd = &cobra.Command{
-		Use:   "deregister",
-		Short: "deregister an entity",
-		Run:   doRegisterOrDeregister,
+		Use:     "gen_deregister",
+		Aliases: []string{"deregister"},
```
kostko: Why aliases?

go/oasis-node/cmd/registry/entity/entity.go
```diff
@@ -61,15 +64,17 @@ var (
...
+		Aliases: []string{"register"},
+		Short:   "generate a register entity transaction",
+		Run:     doRegisterOrDeregister,
 	}
 
 	deregisterCmd = &cobra.Command{
-		Use:   "deregister",
-		Short: "deregister an entity",
-		Run:   doRegisterOrDeregister,
+		Use:     "gen_deregister",
```
kostko: I know this isn't your fault, but it seems inconsistent, `cmdRegister` is a constant but this one is not? Should be either IMO.

go/oasis-node/cmd/registry/entity/entity.go
```diff
@@ -78,9 +83,20 @@ var (
 		Run:   doList,
 	}
 
+	submitCmd = &cobra.Command{
+		Use:   "submit",
+		Short: "submit a signed transaction",
```
kostko:
```suggestion
		Short: "submit a signed registry transaction",
```

go/oasis-node/cmd/registry/entity/entity.go
```diff
@@ -78,9 +83,20 @@ var (
...
+		Run:   doSubmit,
+	}
+
 	logger = logging.GetLogger("cmd/registry/entity")
 )
 
+type serializedTx struct {
+	Register   *entity.SignedEntity `json:"register"`
```
kostko:
```suggestion
	Register   *entity.SignedEntity `json:"register,omitempty"`
```

go/oasis-node/cmd/registry/entity/entity.go
```diff
@@ -78,9 +83,20 @@ var (
...
+	Deregister *signature.Signed    `json:"deregister"`
```
kostko:
```suggestion
	Deregister *signature.Signed    `json:"deregister,omitempty"`
```

go/oasis-node/cmd/registry/entity/entity.go
```diff
@@ -289,18 +333,47 @@ func doRegisterOrDeregister(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
+	var tx *serializedTx
+	switch cmd.Use == cmdRegister {
+	case true:
+		tx, err = doSignRegister(txFilePath, ent, privKey)
```
kostko: Rename `privKey` to `signer` to make it more clear.

go/oasis-node/cmd/registry/entity/entity.go
```diff
@@ -314,61 +387,69 @@ func doRegisterOrDeregister(cmd *cobra.Command, args []string) {
...
 		logger.Error("failed to register entity",
 			"err", err,
 		)
 		return err
 	}
 
-	logger.Info("registered entity",
-		"entity", signer.Public(),
-	)
+	logger.Info("registered entity")
```
kostko: No need to remove the `entity` argument, just replace the value with `signed.Signature.PublicKey`. (Same below.)

go/oasis-node/cmd/registry/entity/entity.go
ravenac95: I actually realize the aliasing was not a great idea because it's a breaking change to the CLI anyhow. So I can remove. The original idea was that I wanted to keep the sub-commands on the CLI consistent so people wouldn't be confused but this actually is more confusing.

go/oasis-node/cmd/registry/entity/entity.go
ravenac95: I'll make them both constants.

ravenac95: Actually I decided to use literals. Thoughts on the last commit @kostko?

ravenac95:
@kostko I tested the changes on the initial commit but I haven't tested it recently. I'll do that and report the results here.

ravenac95:
@kostko tested. Works!

go/oasis-node/cmd/registry/entity/entity.go
```diff
@@ -268,11 +282,39 @@ func signAndWriteEntityGenesis(dataDir string, signer signature.Signer, ent *ent
...
+	}
+	return &tx, nil
+}
+
+func doGenerateEntitySignedTx(signedTxFunc generateEntitySignedTxFunc, cmd *cobra.Command, args []string) {
 	if err := cmdCommon.Init(); err != nil {
 		cmdCommon.EarlyLogAndExit(err)
 	}
 
+	// If we specify a signed file then create a register transaction in a
```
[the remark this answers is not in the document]
ravenac95: Oh oops... that's relic. I will remove.

PR #2375, opened 2019-11-13
Enable domain separation by ChainID for specific contexts

### LINKED ISSUE #2364: Include ChainID as part of signature context
To make sure signed statements are properly isolated between different networks, we should include ChainID as part of the signature context. It may not make sense to include it everywhere (e.g., for signed entity/node/runtime descriptors it would make it impossible to dump-restore to a different ChainID).

kostko:
Fixes #2364 

Note: This is a BREAKING change as it changes the signing contexts. It also makes some contexts dependent on the value of ChainID as specified in the genesis file so specific statements can no longer be mixed between different chains.

Note: This changes the invocations for generating staking transactions as the genesis file must now be passed as an argument (see E2E tests) in order to extract the ChainID.

### DIFF
--- a/go/common/crypto/signature/signer_test.go
+++ b/go/common/crypto/signature/signer_test.go
@@ -1,0 +1,78 @@
+package signature
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestContext(t *testing.T) {
+	require := require.New(t)
+
+	// Make sure a context can be registered.
+	var ctx, ctx2 Context
+	require.NotPanics(func() { ctx = NewContext("test: dummy context 1") })
+	require.NotPanics(func() { ctx2 = NewContext("test: dummy context 2") })
+
+	// Make sure we panic if the same context is registered twice.
+	require.Panics(func() { NewContext("test: dummy context 1") })
+	// Even with different options.
+	require.Panics(func() { NewContext("test: dummy context 1", WithChainSeparation()) })
+
+	// Make sure we panic if context is too long.
+	require.Panics(func() { NewContext(strings.Repeat("a", 500)) })
+
+	// Make sure we panic if context includes the chain context separator.
+	require.Panics(func() { NewContext("test: dummy context 1 for chain blah") })
+
+	// Make sure a context with chain separation can be registered.
+	var chainCtx Context
+	require.NotPanics(func() { chainCtx = NewContext("test: dummy context 3", WithChainSeparation()) })
+
+	// Make sure we cannot use an unregistered context.
+	unregCtx := Context("test: unregistered")
+	_, err := PrepareSignerMessage(unregCtx, []byte("message"))
+	require.Error(err, "PrepareSignerMessage should fail with unregistered context")
+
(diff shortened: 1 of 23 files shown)


go/common/crypto/signature/signer_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	// Make sure we can set a chain context.
+	require.NotPanics(func() { SetChainContext("test: oasis-core tests") })
+	// Make sure we can set the same chain context again.
+	require.NotPanics(func() { SetChainContext("test: oasis-core tests") })
+	// Make sure we can't modify a set chain context.
+	require.Panics(func() { SetChainContext("test: context change") })
+
+	_, err = PrepareSignerMessage(chainCtx, []byte("message"))
+	require.NoError(err, "PrepareSignerMessage should work with chain context")
```
Yawning: Should probably test that the signer messages are different with and without contexts.

go/registry/api/api.go
```diff
@@ -66,7 +66,7 @@ var (
 
 	// RegisterUnfreezeNodeSignatureContext is the context used for
 	// unfreezing nodes.
-	RegisterUnfreezeNodeSignatureContext = signature.NewContext("oasis-core/registry: unfreeze node")
+	RegisterUnfreezeNodeSignatureContext = signature.NewContext("oasis-core/registry: unfreeze node", signature.WithChainSeparation())
```
Yawning: We don't have a method for actually generating these via CLI, but this would break it if such a thing exists.

go/registry/api/api.go
```diff
@@ -40,7 +40,7 @@ var (
 
 	// DeregisterEntitySignatureContext is the context used for entity
 	// deregistration.
-	DeregisterEntitySignatureContext = signature.NewContext("oasis-core/registry: deregister entity")
+	DeregisterEntitySignatureContext = signature.NewContext("oasis-core/registry: deregister entity", signature.WithChainSeparation())
```
Yawning: This breaks the registry cli.

kostko: I should probably wait for #2374 then.

Do we have tests for the registry CLI? We should have something similar to how we have E2E tests for the staking CLI, I'll open an issue.

go/common/crypto/signature/signer_test.go
kostko: Good idea, added.

go/registry/api/api.go
Yawning: I assume entity deregistrations don't interact with the debonding logic at all (we don't need to persist signed deregistrations across a dump/restore boundary)...

kostko: Yeah we don't need to persist deregistrations which is why we can easily add a chain separation context.

Ideally we could do the same for node/entity registrations, but then dump/restore wouldn't work without everyone re-signing.

go/registry/api/api.go
kostko: I've changed this back for now.

go/registry/api/api.go
kostko: I've changed this back for now.

PR #2380, opened 2019-11-13
go/common/crypto/signature: Fail when unmarshaling nil keys

### LINKED ISSUE #2379: Nil public keys can crash consensus
Nil public keys in strategic places can crash consensus due to `ToMapKey` panicking.

kostko:
Fixes #2379 

Note: This is a BREAKING change as it makes some existing serialized state invalid.


Yawning:
Should `ToMapKey` have a special case as well?

kostko:
> Should `ToMapKey` have a special case as well?

Good idea.

kostko:
Please take another look.

PR #2390, opened 2019-11-15
staking: Basic docs and message test vectors

### LINKED ISSUE #2389: Documentation and test vectors for staking messages

kostko:
Fixes #2389


kostko:
This is incomplete but I think it is a good first start. We should plan to add some more documentation, first about the consensus layer and then also about the runtimes, committees, scheduling, etc.

kostko:
Rebased on master and updated initial documentation after #2399.

PR #2395, opened 2019-11-18
oasis-test-runner: Log watcher fixes and improvements

### LINKED ISSUE #2355: oasis-test-runner: Ensure each node gets its own set of log watcher handlers
Currently, all created log watchers get the same copy of the configured log watcher handlers:
https://github.com/oasislabs/oasis-core/blob/f30484fa34ce0b7d333579540be1277be4fdae83/go/oasis-test-runner/oasis/oasis.go#L418-L422

### LINKED ISSUE #2356: oasis-test-runner: Support specifying custom/extra log watcher handlers for each node
This should enable verifying that only certain nodes' logs contain a logging event or text.

tjanez:
This PR:

- Implements log watcher handler factories which makes it easy for each node to get its own set log watcher handlers.
- Enables specifying different log watcher handler factories for each node.

TODO:
- [x] Fix [End-to-End tests that fail](https://buildkite.com/oasislabs/oasis-core-ci/builds/672) due to this change.
- [x] Implement #2356 to enable fixing some failing End-to-End tests.

Closes #2355.
Closes #2356.

### DIFF
--- a/go/oasis-test-runner/log/handlers.go
+++ b/go/oasis-test-runner/log/handlers.go
@@ -6,6 +6,15 @@ import (
 	"strings"
 )
 
+var (
+	_ WatcherHandler        = (*assertBaseHandler)(nil)
+	_ WatcherHandler        = (*assertContainsHandler)(nil)
+	_ WatcherHandler        = (*assertJSONContainsHandler)(nil)
+	_ WatcherHandlerFactory = (*assertBaseFactory)(nil)
+	_ WatcherHandlerFactory = (*assertContainsFactory)(nil)
+	_ WatcherHandlerFactory = (*assertJSONContainsFactory)(nil)
+)
+
 type assertBase struct {
 	message string
 }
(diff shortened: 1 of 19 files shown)


go/oasis-test-runner/log/log.go
```diff
@@ -6,6 +6,12 @@ import (
 	"github.com/pkg/errors"
 )
 
+// WatcherHandlerFactory is a factory interface for log file watcher handlers.
+type WatcherHandlerFactory interface {
+	// Generate will generate and return a WatcherHandler ready for use.
+	Generate() (WatcherHandler, error)
```
kostko: Maybe call it `NewHandler`, `New` or `Create` instead of `Generate`?

go/oasis-test-runner/log/handlers.go
```diff
@@ -14,53 +23,94 @@ func (a *assertBase) fail() error {
...
-		assertBase: assertBase{message},
-		text:       text,
-		expected:   true,
+type assertContainsFactory struct {
+	assertContains
+}
+
+func (fac *assertContainsFactory) Generate() (WatcherHandler, error) {
+	return &assertContainsHandler{
+		assertContains: assertContains{
```
kostko: Just passing `fac.assertContains` should be enough to create a copy (as composition is done by value not by pointer so value assignment should copy), so something like:
```suggestion
		assertContains: fac.assertContains,
```
Also elsewhere to simplify code.

go/oasis-test-runner/oasis/fixture.go
```diff
@@ -118,6 +119,8 @@ type ValidatorFixture struct {
 
 	Entity int `json:"entity"`
 
+	LogWatcherHandlerFactories []log.WatcherHandlerFactory `json:"log_fac"`
```
kostko: Fixtures are supposed to be serializable but factories currently are not. Either set `json:"-"` or make the factories serializable (probably something like having a registry of factory names and using the name on serialization and recreating from the registry on deserialization).

Also in other fixture structs.

go/oasis-test-runner/oasis/oasis.go
```diff
@@ -55,6 +57,8 @@ type Node struct {
 
 	restartable bool
 	doStartNode func() error
+
+	LogWatcherHandlerFactories []log.WatcherHandlerFactory
```
kostko: Group public fields together? Also do these need to be publicly mutable (probably not?) or can they be private and just have public accessor methods?

go/oasis-test-runner/scenario/e2e/basic.go
```diff
@@ -121,21 +121,51 @@ func (sc *basicImpl) Fixture() (*oasis.NetworkFixture, error) {
 			},
 		},
 		Validators: []oasis.ValidatorFixture{
-			oasis.ValidatorFixture{Entity: 1},
-			oasis.ValidatorFixture{Entity: 1},
-			oasis.ValidatorFixture{Entity: 1},
+			oasis.ValidatorFixture{
+				Entity:                     1,
+				LogWatcherHandlerFactories: DefaultBasicLogWatcherHandlerFactories,
```
kostko: So network-wide defaults (e.g., for all nodes) are no longer supported? This seems annoying in case all the factories are the same.

go/oasis-test-runner/log/handlers.go
tjanez: Thanks for the tip!

Done.

go/oasis-test-runner/oasis/fixture.go
tjanez: I've set `json:"-"`.

go/oasis-test-runner/oasis/oasis.go
tjanez: Yes, they don't need to be publicly mutable. I've renamed them to `logWatcherHandlerFactories`.

go/oasis-test-runner/scenario/e2e/basic.go
tjanez: Good suggestion. I've reworked the last commit to retain network-wide default log watcher handler factories and added the ability to enable/disable them in each node fixture (with sensible defaults).

go/oasis-test-runner/oasis/fixture.go
```diff
@@ -296,6 +316,9 @@ type ByzantineFixture struct {
 	Script       string `json:"script"`
 	IdentitySeed string `json:"identity_seed"`
 	Entity       int    `json:"entity"`
+
+	EnableDefaultLogWatcherHandlerFactories bool                        `json:"enable_default_log_fac"`
```
kostko: Why does the byzantine node fixture have this while others do not?

go/oasis-test-runner/oasis/fixture.go
```diff
@@ -306,6 +329,10 @@ func (f *ByzantineFixture) Create(net *Network) (*Byzantine, error) {
 	}
 
 	return net.NewByzantine(&ByzantineCfg{
+		NodeCfg: NodeCfg{
+			DisableDefaultLogWatcherHandlerFactories: !f.EnableDefaultLogWatcherHandlerFactories,
```
kostko: Is the reason for inverting it so that the default is "disabled"? Could we make it consistent in both structures?

tjanez: My reasoning was to enable sensible defaults for different node types.

For example, Byzantine nodes should not have network's default log watcher handler factories enabled because they are byzantine and don't (always) report (their) byzantine behavior in their logs.
[Here](https://buildkite.com/oasislabs/oasis-core-ci/builds/847#f37f7791-1cfe-438c-b897-9479a58dd77c) is an example where honest nodes reported merge discrepancy but `byzantine-0` node didn't.

go/oasis-test-runner/oasis/fixture.go
tjanez: Other fixtures could have `DisableDefaultLogWatcherHandlerFactories bool` but I didn't include it everywhere to prevent "spam" the code too much since none of the current tests need that functionality.

It is trivial to add such a field when a node type would need it.

kostko: Ok.

go/oasis-test-runner/oasis/fixture.go
kostko: Ok.

PR #2396, opened 2019-11-18
ci: put coverage aside

pro-wh:
don't fail when coverage data is corrupted.

also install gocovmerge in the container

### DIFF
--- a/docker/testing/Dockerfile
+++ b/docker/testing/Dockerfile
@@ -37,6 +37,3 @@ RUN wget -O codecov https://codecov.io/bash && \
 # Install tarpaulin.
 RUN RUSTFLAGS="--cfg procmacro2_semver_exempt" \
     cargo install --version 0.9.1 cargo-tarpaulin
-
-# Install gocovmerge for e2e coverage.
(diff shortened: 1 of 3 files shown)


peterjgilbert:
So we can get rid of https://github.com/oasislabs/oasis-core/blob/d01874fa2d8a211b5c93cbf9a1c5be88ff1e2a71/scripts/build-run-e2e-cov.sh#L8 ?

pro-wh:
eh I guess

docker/testing/Dockerfile
```diff
@@ -37,3 +37,6 @@ RUN wget -O codecov https://codecov.io/bash && \
 # Install tarpaulin.
 RUN RUSTFLAGS="--cfg procmacro2_semver_exempt" \
     cargo install --version 0.9.1 cargo-tarpaulin
+
+# Install gocovmerge for e2e coverage.
+RUN GO111MODULE=on go get github.com/wadey/gocovmerge@b5bfa59ec0adc420475f97f89b58045c721d761c
```
kostko: Maybe move the commit into an argument similar to other versions. Maybe it would also be useful if this was part of the development container?

pro-wh: sure I'll do that

PR #2397, opened 2019-11-19
go/common/cbor: Bump fxmacker/cbor to 1.2.0

kostko:
This requires removing the conflicting Marshaler/Unmarshaler interfaces
as using them causes infinite recursion.


Yawning:
Ok.  It removes a bunch of boilerplate which is nice.  If we ever need this sort of thing again (unlikely), we can add encode/decode interfaces or something I suppose.

kostko:
> we can add encode/decode interfaces or something I suppose.

Yeah I considered that (renaming to Encode/Decode*) but just using the `cbor` package explicitly seems nicer (no boilerplate) and is not any longer to type in call sites.

Issue #2378, opened 2019-11-13
Registry gas costs


kostko:
Some of the registry operations (e.g., entity registration/deregistration, unfreezing a node) should probably be changed to cost gas.

Yawning:
Should the gas cost for an entity (re-)registration scale based on how many nodes the entity claims to own?  Alternatively, should the initial node registration (not re-registrations) cost gas (might be hard to do if we want to allow people to update without it costing anything, since the node descriptor goes away)?

kostko:
> Should the gas cost for an entity (re-)registration scale based on how many nodes the entity claims to own?

Yeah that's a good idea. But then there needs to be separate charges for entity-signed node registrations.

> Should the node registrations cost gas?
> Maybe just the initial node registration?

That would enable the node key to drain the entity's account?

> Maybe a combination of "entity registrations cost more gas if it includes more nodes", and "self-signed (vs entity signed) node registrations are gas-less"?

Yeah this makes the most sense -- so entity registrations cost more gas based on the length of the `nodes` field and entity-signed nodes must pay for gas.

PR #2399, opened 2019-11-19
go/consensus: Use common transaction format

kostko:
Part of #2378 

:boom: **This is a BREAKING change as it changes the consensus transaction format. It also completely changes the staking/registry CLI for generating and submitting transactions so any documentation will need to be updated.** :boom: 

You have been warned.


go/grpc/consensus/consensus.pb.go
```diff
@@ -0,0 +1,189 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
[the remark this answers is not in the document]
kostko: Ooops, forgot to update `.gitignore` :-)

kostko:
> do we have a way to dump the contents of a on-disk tx?

Oh that's a good idea, I'll add a tool to show the content of the signed transaction.

go/consensus/api/transaction/transaction.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
+)
+
+// SignatureContext is the context used for signing transactions.
+var SignatureContext = signature.NewContext("oasis-core/consensus: tx", signature.WithChainSeparation())
+
+// Transaction is an unsigned consensus transaction.
+type Transaction struct {
+	// Nonce is an optional nonce to prevent replay.
+	Nonce uint64 `json:"nonce,omitempty"`
```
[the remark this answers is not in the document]
kostko: Nope, I'll update the comment and annotation.

kostko:
Rebased on master and added the following (in separate commits):
* Correctly propagate errors from consensus applications.
* Add submission manager to manager transaction nonces. It currently supports batching multiple method calls into one transaction to decrease latency with multiple concurrent transactions being submitted by a node, but our transaction format does not support it at the moment.

Maybe we should at least change the format to support multiple calls but only accept a single call for now (e.g., enforce length of one in consensus)?

Yawning:
> Maybe we should at least change the format to support multiple calls but only accept a single call for now (e.g., enforce length of one in consensus)?

As long as there's no plans to ever do anything other than treating txes in a batch independently, I think this (and increasing it later) are both fine.

kostko:
> As long as there's no plans to ever do anything other than treating txes in a batch independently

Yeah I would be in favor of treating them independently, the same as if they were submitted in separate transactions. E.g., there should be no atomicity guarantees for batched calls.

go/common/errors/errors.go
```diff
@@ -0,0 +1,83 @@
...
+	msg    string
+}
+
+func (e *codedError) Error() string {
+	return e.msg
+}
+
+// New creates a new error.
+//
+// Module and code must be unique. If they are not, this method will
```
ptrus: Maybe something like `Module and code pair must be unique` might be more correct. (something that states that the pair of module & code needs to be unique, not each on it's own)

go/consensus/api/api.go
```diff
@@ -0,0 +1,153 @@
...
+	staking "github.com/oasislabs/oasis-core/go/staking/api"
+)
+
+// ClientBackend is a limited consensus interface used by clients that
+// connect to the local node.
+type ClientBackend interface {
+	// SubmitTx submits a signed consensus transaction.
+	SubmitTx(ctx context.Context, tx *transaction.SignedTransaction) error
+
+	// TODO: Add things like following consensus blocks.
```
ptrus: will this be done in this PR? (otherwise open an issue)

kostko: No it will not be done as it has nothing to do with the transaction format :-) I'll open an issue, this is something that the blockexploder people will need.

go/consensus/api/transaction/transaction.go
```diff
@@ -0,0 +1,91 @@
...
+	if len(m) == 0 {
+		return fmt.Errorf("transaction: empty method")
+	}
+
+	return nil
+}
+
+// NewMethodName creates a new method name.
+func NewMethodName(backendName, method string) MethodName {
+	// TODO: Check for duplicate method names.
```
ptrus: will this be done in this PR?

go/consensus/tendermint/abci/mux.go
```diff
@@ -221,11 +232,23 @@ func (a *ApplicationServer) Pruner() StatePruner {
...
 
 // SetEpochtime sets the mux epochtime.
-// XXX: epochtime needs to be set before mux can be used.
+//
+// Epochtime must be set before the multiplexer can be used.
 func (a *ApplicationServer) SetEpochtime(epochTime epochtime.Backend) {
 	a.mux.state.timeSource = epochTime
 }
 
+// SetFeeHandler configures the transaction fee handler for the
```
ptrus: `SetTransactionFeeHandler`

go/consensus/tendermint/service/service.go
```diff
@@ -29,6 +28,10 @@ type TendermintService interface {
 	// registered.
 	RegisterApplication(abci.Application) error
 
+	// SetFeeHandler configures the transaction fee handler for the
```
ptrus: `SetTransactionFeeHandler`

go/consensus/tendermint/abci/mux.go
```diff
@@ -221,11 +232,23 @@ func (a *ApplicationServer) Pruner() StatePruner {
...
+// ABCI multiplexer.
+func (a *ApplicationServer) SetTransactionFeeHandler(handler TransactionFeeHandler) error {
+	if a.mux.state.txFeeHandler != nil {
+		return fmt.Errorf("mux: transaction fee handler already configured")
```
ptrus: similar check might make sense above for `SetEpochtime` as well :)

go/consensus/tendermint/abci/mux.go
```diff
@@ -288,23 +311,7 @@ func (mux *abciMux) Info(req types.RequestInfo) types.ResponseInfo {
 }
 
 func (mux *abciMux) SetOption(req types.RequestSetOption) types.ResponseSetOption {
```
ptrus: Is this still needed? (if we skipped it, the embedded impl. from `types.BaseApplication` would be used right?)

go/consensus/tendermint/abci/mux.go
```diff
@@ -80,9 +95,11 @@ type Application interface {
 	// and Query calls and accessing genesis state.
```
ptrus: comment above still mentions `for de-multiplexing SetOption` which is not relevant anymore

go/consensus/tendermint/apps/epochtime_mock/epochtime_mock.go
```diff
@@ -85,24 +88,23 @@ func (app *epochTimeMockApplication) BeginBlock(ctx *abci.Context, request types
...
+	switch tx.Method {
+	case MethodSetEpoch:
+		var epoch epochtime.EpochTime
+		if err := cbor.Unmarshal(tx.Body, &epoch); err != nil {
+			return err
+		}
+
+		return app.setEpoch(ctx, state, epoch)
+	default:
+		return errors.New("epochtime_mock: invalid argument")
```
ptrus: - use `fmt.Errorf`
- maybe also include `tx.Method` in the error

go/oasis-node/cmd/common/grpc/grpc.go
```diff
@@ -74,7 +74,11 @@ func NewServerLocal(installWrapper bool) (*cmnGrpc.Server, error) {
 func NewClient(cmd *cobra.Command) (*grpc.ClientConn, error) {
 	addr, _ := cmd.Flags().GetString(CfgAddress)
 
-	conn, err := grpc.Dial(addr, grpc.WithInsecure()) // TODO: TLS?
+	conn, err := grpc.Dial(
```
ptrus: Should the `defaultAddress` be switched to the internal socket at some point?

kostko: This is impossible as you don't know the path. Could just be `internal.sock` if that's what you mean?

kostko: I'll just set it to `"unix:"+localSocketFilename`.

go/consensus/api/transaction/transaction.go
kostko: Done.

go/consensus/tendermint/abci/mux.go
kostko: Done.

go/oasis-node/cmd/common/grpc/grpc.go
kostko: Done.

ptrus: oh right, you would be missing the `--datadir`. I'm fine with either then, maybe no default would actually be the best in that case?

go/consensus/tendermint/abci/mux.go
```diff
@@ -320,12 +368,12 @@ func (mux *abciMux) Query(req types.RequestQuery) types.ResponseQuery {
...
 		)
 		return types.ResponseQuery{
-			Code: api.CodeOK.ToInt(),
+			Code: types.CodeTypeOK,
 		}
 	}
 
 	return types.ResponseQuery{
-		Code: api.CodeInvalidQuery.ToInt(),
+		Code: types.CodeTypeOK,
```
ptrus: not sure if it matters really, but the code here used to be `api.CodeInvalidQuery`

kostko: Yeah this is on purpose as we don't use (and will not use) the Tendermint query interface (as we have our own that can avoid all the re-serialization/deserialization). Could also add a random fail code instead.

kostko: I'll add a comment to make this more clear.

kostko: Maybe we should just remove `Query`.

kostko:
Added a pretty printer for transactions, it outputs stuff like:
```
Signer: 53405b2ddcce0f4302a017ef1301010d5406b4527271202d19d3e16eb87b1235
        (signature: dynkEmLoXrntyYROPMB83pmVOL7sES6xEN1+/4A+6N7k+hHkw4XbtBBxiMprSUaeyOx9Ny7ynWAMQIDTwkIxBA==)
Nonce:  0
Fee:    0 (gas limit: 0, gas price: 0)
Method: registry.RegisterEntity
Body:
  {
    "untrusted_raw_value": {
      "id": "U0BbLdzOD0MCoBfvEwEBDVQGtFJycSAtGdPhbrh7EjU=",
      "nodes": null,
      "allow_entity_signed_nodes": true
    },
    "signature": {
      "public_key": "U0BbLdzOD0MCoBfvEwEBDVQGtFJycSAtGdPhbrh7EjU=",
      "signature": "etkQFvcRzR7U7QN6asBCvM1KjU3Whktqe/V80GQ2AcgDQ5Y1GjGFULbTgDuDKrhlfbT7P8RtNtB4/J0zNC7/Aw=="
    }
  }
```

```
Signer: 53405b2ddcce0f4302a017ef1301010d5406b4527271202d19d3e16eb87b1235
        (signature: 57993E0eANEfhnKVjj8ooR7cdg5kkrOb6VXPUdOa7PLvIJgi4RYcq8J5u1GdTeNh5RTn7EHZd4Trk+JhIGgvBg==)
Nonce:  0
Fee:    99900 (gas limit: 1000, gas price: 99)
Method: staking.Transfer
Body:
  {
    "xfer_to": "U0BbLdzOD0MCoBfvEwEBDVQGtFJycSAtGdPhbrh7EjU=",
    "xfer_tokens": "12345"
  }
```
etc.

go/consensus/tendermint/abci/mux.go
kostko: Removed.

go/consensus/tendermint/tendermint.go
```diff
@@ -587,68 +626,70 @@ func (t *tendermintService) Scheduler() schedulerAPI.Backend {
...
+	}
 
-		if t.scheduler, err = tmscheduler.New(t.ctx, t); err != nil {
-			t.Logger.Error("scheduler: failed to initialize scheduler backend",
-				"err", err,
-			)
-			return err
-		}
-		t.svcMgr.RegisterCleanupOnly(t.scheduler, "scheduler backend")
+	if t.registry, err = tmregistry.New(t.ctx, t); err != nil {
```
ptrus: No more registry metrics?

kostko: The only metrics were based on actual registry method invocations (and the registry backend now no longer has these methods as all transactions go through `SubmitTx` on the consensus backend). We should probably move these metrics to the consensus part as otherwise you are only counting local invocations.

ptrus: it did actually also periodically check `updatePeriodicMetrics` (which posted number of registered nodes and entities). But i do agree that that actually didn't even belong in the metrics wrapper and should be done somewhere else :) (would be nice to keep those metrics though)

kostko: Oh I didn't see that, I'll add it back :-)

go/common/crypto/signature/signature.go
```diff
@@ -430,6 +437,35 @@ func (s *Signed) ToProto() *common.Signed {
...
+//
+// It should only be used for pretty printing.
+type PrettySigned struct {
+	Body      interface{} `json:"untrusted_raw_value"`
+	Signature Signature   `json:"signature"`
+}
+
+// PrettyPrint returns a pretty-printed representation of the structure.
+func (p PrettySigned) PrettyPrint(prefix string, v interface{}) string {
+	pp := v.(*PrettySigned)
```
ptrus: Pretty sure that i'm missing something, but wouldn't the same be achieved with:
```
func (pp *PrettySigned) PrettyPrint(prefix string) string {
```

kostko: Yeah it is not, but I think this can be changed, I was experimenting with the interface but now I see that it can be cleaned up, thanks!

go/consensus/tendermint/tendermint.go
kostko: Added back (just the periodic updater), PTAL.

ptrus: looks good :+1:

go/oasis-test-runner/scenario/e2e/stake_cli.go
```diff
@@ -414,6 +429,19 @@ func (s *stakeCLIImpl) checkEscrowBalance(childEnv *env.Env, src signature.Publi
...
 }
 
+func (s *stakeCLIImpl) showTx(childEnv *env.Env, txPath string) error {
+	s.logger.Info("pretty printing generated transaction")
+
+	args := []string{
+		"consensus", "show_tx",
+		"--" + consensus.CfgTxFile, txPath,
+	}
+	if err := runSubCommand(childEnv, "show_tx", s.basicImpl.net.Config().NodeBinary, args); err != nil {
```
pro-wh: why is subcommand the preferred way to do this

go/oasis-node/cmd/consensus/consensus.go
```diff
@@ -78,24 +81,98 @@ func doSubmitTx(cmd *cobra.Command, args []string) {
...
+		v := reflect.New(reflect.TypeOf(bodyType)).Interface()
+		if err := cbor.Unmarshal(tx.Body, v); err != nil {
+			fmt.Printf("  <error: %s>\n", err)
+			fmt.Printf("  <malformed: %s>\n", base64.StdEncoding.EncodeToString(tx.Body))
+			break
+		}
+
+		// If the body type supports pretty printing, use that.
+		if pp, ok := v.(interface {
+			PrettyPrint(prefix string) string
```
pro-wh: maybe we should export this interface

go/oasis-node/cmd/consensus/consensus.go
```diff
@@ -78,24 +81,98 @@ func doSubmitTx(cmd *cobra.Command, args []string) {
...
+	if err := cbor.Unmarshal(sigTx.Blob, &tx); err != nil {
+		logger.Error("failed to unmarshal unsigned transaction",
+			"err", err,
+		)
+		os.Exit(1)
+	}
+
+	// Pretty print the transaction.
+	fmt.Printf("Signer: %s\n", sigTx.Signature.PublicKey)
+	fmt.Printf("        (signature: %s)\n", sigTx.Signature.Signature)
```
pro-wh: SignedTransaction, unlike other SignedXxx structs, doesn't have a PrettyPrint method, and instead there's special treatment in doShowTx.

go/oasis-node/cmd/consensus/consensus.go
```diff
@@ -78,24 +81,98 @@ func doSubmitTx(cmd *cobra.Command, args []string) {
...
 
+func doShowTx(cmd *cobra.Command, args []string) {
+	if err := cmdCommon.Init(); err != nil {
+		cmdCommon.EarlyLogAndExit(err)
+	}
+
+	sigTx := loadTx()
+
+	var tx transaction.Transaction
+	if err := cbor.Unmarshal(sigTx.Blob, &tx); err != nil {
```
pro-wh: :information_source: show_tx doesn't verify any signatures

go/Makefile
```diff
@@ -45,7 +45,7 @@ lint:
 
 # Test.
 test: generate
-	@env -u GOPATH $(OASIS_GO) test -race -v ./...
+	@env -u GOPATH $(OASIS_GO) test -timeout 5m -race -v ./...
```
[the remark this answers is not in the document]
kostko: Not in the current version, but during development there were hangs. And I think waiting the default 10m is too much so I lowered it to 5m as I think that unit tests shouldn't be taking that long if everything is correct.

go/consensus/api/transaction/transaction.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/oasislabs/oasis-core/go/common/cbor"
+	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
+)
+
+var (
+	// SignatureContext is the context used for signing transactions.
+	SignatureContext = signature.NewContext("oasis-core/consensus: tx", signature.WithChainSeparation())
+
+	registeredMethods sync.Map
```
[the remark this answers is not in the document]
kostko: No, but this is a convention that was already established for registering domain separation contexts so I followed it. While most mutations are in initializers which are not executed concurrently, making sure that this doesn't cause accidental corruption when invoked from a random point is the prudent thing to do.

go/consensus/tendermint/apps/epochtime_mock/epochtime_mock.go
```diff
@@ -85,24 +89,23 @@ func (app *epochTimeMockApplication) BeginBlock(ctx *abci.Context, request types
...
 	state := newMutableState(ctx.State())
 
-	if tx.TxSetEpoch != nil {
-		return app.setEpoch(ctx, state, tx.TxSetEpoch.Epoch)
+	switch tx.Method {
+	case MethodSetEpoch:
+		var epoch epochtime.EpochTime
+		if err := cbor.Unmarshal(tx.Body, &epoch); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
kostko: > to get some arguably hacky savings, transaction request bodies don't all uniformly have a dedicated request structure type. here, for example, it's a bare EpochTime

This was already like this before (e.g., the method took an epochtime directly), so nothing has changed.

go/consensus/tendermint/apps/registry/registry.go
```diff
@@ -237,24 +253,8 @@ func (app *registryApplication) registerEntity(
...
-	ctx *abci.Context,
-	state *registryState.MutableState,
-	sigTimestamp *signature.Signed,
-) error {
-	id, timestamp, err := registry.VerifyDeregisterEntityArgs(app.logger, sigTimestamp)
-	if err != nil {
-		return err
-	}
-
-	err = registry.VerifyTimestamp(timestamp, uint64(ctx.Now().Unix()))
```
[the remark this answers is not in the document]
kostko: Yes, as noted in the commit comment (removing timestamps is a separate commit), timestamps are not needed anymore as nonces are now used everywhere.

go/consensus/tendermint/service/service.go
```diff
@@ -1,16 +1,15 @@
 // Package service provides the tendermint service interface.
+// TODO: Move this under tendermint/internal.
```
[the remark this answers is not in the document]
kostko: I was considering it but left it out for now. I'll remove the comment as having access to the Tendermint implementation seems useful in places for now.

go/consensus/tendermint/tendermint.go
```diff
@@ -587,68 +589,72 @@ func (t *tendermintService) Scheduler() schedulerAPI.Backend {
...
-		if t.keymanager, err = tmkeymanager.New(t.ctx, t); err != nil {
-			t.Logger.Error("initialize: failed to initialize keymanager backend",
-				"err", err,
-			)
-			return err
-		}
+	if err := t.initEpochtime(); err != nil {
+		return err
+	}
+	if err := t.mux.SetEpochtime(t.epochtime); err != nil {
```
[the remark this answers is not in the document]
kostko: Yes as requested during review, see comment from @ptrus.

go/epochtime/tests/mock_tester.go
```diff
@@ -57,11 +57,14 @@ func EpochtimeSetableImplementationTest(t *testing.T, backend api.Backend) {
 func MustAdvanceEpoch(t *testing.T, backend api.SetableBackend, increment uint64) api.EpochTime {
 	require := require.New(t)
 
-	epoch, err := backend.GetEpoch(context.Background(), 0)
+	ctx, cancel := context.WithTimeout(context.Background(), recvTimeout)
```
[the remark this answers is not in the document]
kostko: In case of bugs there can be timeouts here, yes. So I added this to fail early instead of blocking for 10m. Of course the timeouts should not happen now, only if something is messed up as it was during development.

go/oasis-node/cmd/debug/byzantine/scheduler.go
```diff
@@ -21,11 +21,7 @@ func schedulerNextElectionHeight(svc service.TendermintService, kind scheduler.C
 	if err != nil {
 		return 0, fmt.Errorf("Tendermint Subscribe error: %w", err)
 	}
-	defer func() {
-		if err := tendermintUnsubscribeDrain(svc, "script", schedulerapp.QueryApp, sub); err != nil {
-			panic(fmt.Sprintf("Tendermint unsubscribe drain: %+v", err))
-		}
-	}()
+	defer svc.Unsubscribe("script", schedulerapp.QueryApp) // nolint: errcheck
```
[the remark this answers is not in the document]
kostko: Yes, this has been true since @jberci wrapped all tendermint subscriptions, see #2180.

go/oasis-node/cmd/debug/byzantine/registry.go
```diff
@@ -53,15 +53,8 @@ func registryRegisterNode(svc service.TendermintService, id *identity.Identity,
...
-		TxRegisterNode: &registryapp.TxRegisterNode{
-			Node: *signedNode,
-		},
-	}); err != nil {
-		return errors.Wrap(err, "Tendermint BroadcastTx commit")
-	}
-
-	return nil
+	tx := registry.NewRegisterNodeTx(0, nil, signedNode)
+	return consensus.SignAndSubmitTx(context.Background(), svc, id.NodeSigner, tx)
```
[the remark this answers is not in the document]
kostko: Ok.

go/roothash/tests/tester.go
```diff
@@ -57,10 +56,10 @@ func RootHashImplementationTests(t *testing.T, backend api.Backend, epochtime ep
 		if len(rtStates) > 0 {
 			// This is entity deregistration based, and all of the
 			// runtimes used in this test share the entity.
-			rtStates[0].rt.Cleanup(t, registry, epochtime)
+			rtStates[0].rt.Cleanup(t, consensus.Registry(), consensus)
```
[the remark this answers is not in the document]
kostko: > (wait, does it?)

So the use case would be testing a registry client implementation, but still using the non-client consensus backend. Which is a bit hacky I agree and could be improved, but we can't do everything in this PR.

go/oasis-test-runner/scenario/e2e/stake_cli.go
kostko: Adding the subcommand was response to a review suggestion from @Yawning. I think it is useful to have a subcommand for displaying the content of transactions.

go/oasis-node/cmd/consensus/consensus.go
kostko: Good idea, I'll move it to common.

go/oasis-node/cmd/consensus/consensus.go
kostko: That's a good idea.

go/oasis-node/cmd/consensus/consensus.go
kostko: Correct, because otherwise you would need to pass a genesis file as well so that signatures could be verified. I'll change it so that it also displays whether the signature is valid or not if you pass the genesis file.

go/consensus/api/api.go
```diff
@@ -0,0 +1,157 @@
...
+
+import (
+	"context"
+
+	beacon "github.com/oasislabs/oasis-core/go/beacon/api"
+	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
+	"github.com/oasislabs/oasis-core/go/common/node"
+	"github.com/oasislabs/oasis-core/go/consensus/api/transaction"
+	epochtime "github.com/oasislabs/oasis-core/go/epochtime/api"
+	genesisAPI "github.com/oasislabs/oasis-core/go/genesis/api"
```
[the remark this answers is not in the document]
kostko: This is not something that I changed in this PR, but sure I can fix it.

go/scheduler/api/api.go
```diff
@@ -58,7 +49,7 @@ func (r Role) String() string {
 	case Leader:
 		return "leader"
 	default:
-		return fmt.Sprintf("unknown role: %d", r)
+		return fmt.Sprintf("[unknown role: %d]", r)
```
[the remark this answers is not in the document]
kostko: Convention for when returned things are malformed. No strong preference.

go/consensus/api/submission.go
```diff
@@ -0,0 +1,103 @@
...
+
+type submissionManager struct {
+	backend Backend
+
+	logger *logging.Logger
+}
+
+func (m *submissionManager) signAndSubmitTx(ctx context.Context, signer signature.Signer, tx *transaction.Transaction) error {
+	// Update transaction nonce.
+	acct, err := m.backend.Staking().AccountInfo(ctx, signer.Public(), 0)
```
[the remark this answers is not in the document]
kostko: That's a good idea, I'll decouple it.

go/consensus/api/submission.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		m.logger.Error("failed to fetch nonce",
+			"account_id", signer.Public(),
+			"err", err,
+		)
+		return backoff.Permanent(err)
```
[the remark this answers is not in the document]
kostko: Good point.

go/consensus/tendermint/apps/registry/registry.go
```diff
@@ -207,15 +207,8 @@ func (app *registryApplication) registerEntity(
...
-		err = registry.VerifyTimestamp(ent.RegistrationTime, uint64(ctx.Now().Unix()))
-		if err != nil {
-			app.logger.Error("RegisterEntity: invalid timestamp",
-				"entity_timestamp", ent.RegistrationTime,
-				"now", uint64(ctx.Now().Unix()),
-			)
-			return err
-		}
+	if ctx.IsCheckOnly() {
+		return nil
```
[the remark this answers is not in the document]
kostko: Because CheckTx should really be minimal to avoid expensive operations if gas is not paid.

go/roothash/api/api.go
```diff
@@ -72,19 +81,41 @@ type Backend interface {
...
 }
 
+// ComputeCommit is the argument set for the ComputeCommit method.
+type ComputeCommit struct {
+	ID      signature.PublicKey            `json:"id"`
+	Commits []commitment.ComputeCommitment `json:"commits"`
+}
+
+// NewComputeCommitTx creates a new compute commit transaction.
+func NewComputeCommitTx(runtimeID signature.PublicKey, commits []commitment.ComputeCommitment) *transaction.Transaction {
```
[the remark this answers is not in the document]
kostko: Initial thought was that callers of these transactions will not directly pay for gas. But yeah let's make it consistent anyway.

go/staking/tests/tendermint.go
```diff
@@ -12,9 +12,9 @@ import (
 	tmtypes "github.com/tendermint/tendermint/types"
 
 	"github.com/oasislabs/oasis-core/go/common/identity"
+	"github.com/oasislabs/oasis-core/go/consensus/api"
```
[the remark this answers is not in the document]
kostko: Moved.

kostko:
> we now have 3 IDs associated with each component: ID, backend name, and app name

The "app name" is just for ordering, so maybe we should change it to `AppOrder`. But not in this PR.

kostko:
> "backend name" now refers to the component's name (e.g. 'registry') rather than identifying which implementation of the backend interface (e.g. 'tendermint')

Yeah somewhere I use "module name" maybe that or "component name" would be better than backend. So roothash, registry, etc. are components and not backends. Then "tendermint" is a consensus backend providing concrete component implementations.

kostko:
> this PR needs a proper description. requesting changes. please include ~19 items to describe what comes with the first commit of this branch to reviewers, developers working on this project, and team members managing communications to node operators.

I'll update the commit message of the first commit to explain what changes it has. Other changes that are in separate commits are explained there already.

kostko:
For now I've renamed the newly introduced "BackendName" to "ModuleName".

I've opened #2409 to track terminology cleanup. Let's not waste time on it in this PR.

PR #2404, opened 2019-11-20
abci: move calling InitChain on apps from BeginBlock to mux InitChain

matevz:
PR for https://github.com/oasislabs/oasis-core/issues/2122:
* Moves `InitChain()` calls to applications from `BeginBlock()` of mux to `InitChain()` where it should be.
* Temporarily stores the events emitted during `InitChain()` and prepends them to `BeginBlock` events during the first `BeginBlock()` call in mux.
* Replaces immutable state calls with `NewMutableState(abciCtx.State())` in `QueryAt()` functions when called from `InitChain`.

### DIFF
--- a/go/consensus/tendermint/abci/mux.go
+++ b/go/consensus/tendermint/abci/mux.go
@@ -36,8 +36,9 @@ import (
 )
 
 const (
-	stateKeyGenesisDigest  = "OasisGenesisDigest"
-	stateKeyGenesisRequest = "OasisGenesisRequest"
+	stateKeyGenesisDigest   = "OasisGenesisDigest"
+	stateKeyGenesisRequest  = "OasisGenesisRequest"
+	stateKeyInitChainEvents = "OasisInitChainEvents"
 
 	metricsUpdateInterval = 10 * time.Second
 )
@@ -391,18 +392,9 @@ func (mux *abciMux) InitChain(req types.RequestInitChain) types.ResponseInitChai
 	genesisDigest := sha512.Sum512_256(tmp.Bytes())
 	mux.state.deliverTxTree.Set([]byte(stateKeyGenesisDigest), genesisDigest[:])
 
-	// HACK: Can't emit tags from InitChain, stash the genesis state
-	// so that processing can happen in BeginBlock.
-	//
-	// This is also stashed in the CheckTx tree (even though it will get
-	// rolled back), so that it is usable from the genesisHooks.
-	b, _ = req.Marshal()
-	mux.state.deliverTxTree.Set([]byte(stateKeyGenesisRequest), b)
-	mux.state.checkTxTree.Set([]byte(stateKeyGenesisRequest), b)
-
 	resp := mux.BaseApplication.InitChain(req)
 
-	// HACK: The state is only updated iff validators or consensus paremeters
+	// HACK: The state is only updated iff validators or consensus parameters
 	// are returned.
 	//
 	// See: tendermint/consensus/replay.go (Handshaker.ReplayBlocks)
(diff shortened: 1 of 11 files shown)


go/consensus/tendermint/abci/mux.go
```diff
@@ -398,6 +389,37 @@ func (mux *abciMux) InitChain(req types.RequestInitChain) types.ResponseInitChai
...
+			)
+			panic("mux: InitChain: fatal error in application: '" + app.Name() + "': " + err.Error())
+		}
+	}
+
+	mux.logger.Debug("InitChain: initializing of applications complete", "num_collected_events", len(ctx.GetEvents()))
+
+	// Since returning emitted events doesn't work for InitChain() response yet,
+	// we store those and return them in BeginBlock().
+	mux.initChainEvents = ctx.GetEvents()
```
kostko: This doesn't seem right, the events should be stored in persistent state (`mux.state.*TxTree`) otherwise a node crash will just erase events and cause divergence.

go/consensus/tendermint/abci/mux.go
```diff
@@ -412,7 +404,6 @@ func (mux *abciMux) InitChain(req types.RequestInitChain) types.ResponseInitChai
 
 	// Dispatch registered genesis hooks.
 	mux.RLock()
-	defer mux.RUnlock()
```
kostko: If `hook()` below panics, this will now lead to the mutex being held forever. You can wrap this part in a function and keep using defer, e.g.:
```golang
func() {
    mux.RLock()
    defer mux.RUnlock()

    mux.logger.Debug("Dispatching genesis hooks")
    // ...
}()
```

go/consensus/tendermint/abci/mux.go
```diff
@@ -422,6 +413,38 @@ func (mux *abciMux) InitChain(req types.RequestInitChain) types.ResponseInitChai
 
 	mux.logger.Debug("Genesis hook dispatch complete")
 
+	mux.RUnlock()
+
+	// TODO Matevz: It seems stateKeyGenesisRequest is still used in roothash's ForeignExecuteTx for generating new runtimes?
```
kostko: Looked at that and it seems wrong to me as in `ForeignExecuteTx` it is using the roothash genesis block to populate the genesis block of a runtime that was just registered.

The runtime structure already contains a `Genesis` field that has either a `StateRoot` or `State`. I think we should:
* Remove `Blocks` from roothash genesis.
* Add `Round` to `RuntimeGenesis` and use that when initializing a runtime genesis block.
* Change `onNewRuntime` and `ForeignExecuteTx` to only use the runtime's `Genesis` structure.
* Update runtime-ethereum's genesis utilities to generate runtime descriptors instead of roothash state.

However this is independent of `InitChain` and so should be done in a separate issue/PR. I've filed #2426, maybe just reference it in this comment, e.g. `(see oasis-core#2426)`.

go/consensus/tendermint/abci/mux.go
```diff
@@ -568,7 +544,23 @@ func (mux *abciMux) BeginBlock(req types.RequestBeginBlock) types.ResponseBeginB
...
-	response.Events = ctx.GetEvents()
+
+	// Check any events left from InitChain.
+	_, evJSON := mux.state.deliverTxTree.Get([]byte(stateKeyInitChainEvents))
+	if evJSON == nil {
+		// For the second and the following blocks, only return events from the
+		// application's BeginBlock calls.
+		response.Events = ctx.GetEvents()
+	} else {
+		// During the first block, also emit application events generated during
```
kostko: Can we also enforce that this only happens in the first block and a non-nil value in `stateKeyInitChainEvents` is an error (panic) otherwise?

go/consensus/tendermint/abci/mux.go
```diff
@@ -568,7 +545,26 @@ func (mux *abciMux) BeginBlock(req types.RequestBeginBlock) types.ResponseBeginB
 	}
 
 	response := mux.BaseApplication.BeginBlock(req)
-	response.Events = ctx.GetEvents()
+
+	// Check any events left from InitChain.
+	_, evJSON := mux.state.deliverTxTree.Get([]byte(stateKeyInitChainEvents))
```
kostko: Actually on second thought this will cause needless lookups on each block. Can we just look this up on on the first block?

matevz: Fixed.

go/consensus/tendermint/abci/mux.go
```diff
@@ -411,16 +403,48 @@ func (mux *abciMux) InitChain(req types.RequestInitChain) types.ResponseInitChai
...
+			panic("mux: InitChain: fatal error in application: '" + app.Name() + "': " + err.Error())
+		}
 	}
 
-	mux.logger.Debug("Genesis hook dispatch complete")
+	mux.logger.Debug("InitChain: initializing of applications complete", "num_collected_events", len(ctx.GetEvents()))
+
+	// Since returning emitted events doesn't work for InitChain() response yet,
+	// we store those and return them in BeginBlock().
+	evJSON, _ := json.Marshal(ctx.GetEvents())
```
kostko: Oh didn't notice this before, but why do you use JSON to serialize this? Replace it with CBOR serialization.

matevz: No reason actually.. I used it for debugging and then left it there. :/

PR #2417, opened 2019-11-23
Add development docker to Makefile

jleni:
I find myself frequently going to the README.md to find the right parameters to run the docker container used for development.

I thought it was worth adding this to the Makefile and sharing the change.

Note: There is an old existing issue with the container. Files created inside the container are using the root user id, so later they are not easy to access/delete. This is something that could be fixed by creating the right user ids inside the docker container. This is anyway unrelated to the this and can be fixed later.


kostko:
Thanks for your contribution, looks good!

PR #2423, opened 2019-11-25
Add tests for genesis sanity checks

### LINKED ISSUE #2358: go/oasis-node/cmd/genesis: Add a self-delegation for the test entity
When creating a genesis file with the debug test entity enabled, the genesis provisioning tool injects a staking ledger entity for said debug test entity with one hundred billion general balance, and one hundred billion escrow balance.

This was done prior to delegation support, and a self delegation needs to be added as well.  While we are here, a component somewhere should enforce that the delegations and escrow balance matches, because this should trigger a hard failure due to invariant violation, instead of working.

abukosek:
Part of #1633.
Closes #2358.

This PR adds tests for the functionality implemented in PR #2365, as well as a few new sanity checks.
It also adds self-delegation for the test entity.


abukosek:
Thanks! :)

PR #2424, opened 2019-11-25
registry validation consistency

ptrus:
Fixes: https://github.com/oasislabs/oasis-core/issues/2400

Changes: 
- `committee.Certificate` & `p2p.ID` are now both required fields in registry (note: this will likely break Genesis SanityCheck on a dump in existing deployments where validators have an invalid P2P ID by default - to mitigate this: should probably allow invalid P2P keys for the next few versions in Genesis)
- optional node addresses are now validated as well in the registry (before addresses were only validated if required)
- registration worker no longer registers with committee addresses unless role requires it, and provides all required IDs

### DIFF
--- a/go/registry/api/api.go
+++ b/go/registry/api/api.go
@@ -138,6 +138,21 @@ var (
 		MethodUnfreezeNode,
 		MethodRegisterRuntime,
 	}
+
+	// ConsensusAddressRequiredRoles are the Node roles that require Consensus Address.
+	ConsensusAddressRequiredRoles = node.RoleValidator
+
+	// CommitteeAddressRequiredRoles are the Node roles that require Committee Address.
+	CommitteeAddressRequiredRoles = (node.RoleComputeWorker |
+		node.RoleStorageWorker |
+		node.RoleTransactionScheduler |
+		node.RoleKeyManager |
+		node.RoleMergeWorker)
+
+	// P2PAddressRequiredRoles are the Node roles that require P2P Address.
+	P2PAddressRequiredRoles = (node.RoleComputeWorker |
+		node.RoleTransactionScheduler |
+		node.RoleMergeWorker)
 )
 
 // Backend is a registry implementation.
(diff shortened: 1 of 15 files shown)


go/registry/api/api.go
```diff
@@ -1056,6 +1063,12 @@ func (g *Genesis) SanityCheck() error { // nolint: gocyclo
 		if !n.Consensus.ID.IsValid() {
 			return fmt.Errorf("registry: sanity check failed: node ID %s has an invalid consensus ID", n.ID.String())
 		}
+
+		if !n.P2P.ID.IsValid() {
+			return fmt.Errorf("registry: sanity check failed: node ID %s has an invalid P2P ID", n.ID.String())
+		}
+
+		// XXX: Validate addresses as well?
```
kostko: Yes?

go/worker/registration/worker.go
```diff
@@ -354,19 +380,14 @@ func (w *Worker) consensusValidatorHook(n *node.Node) error {
 		}
 	}
 
+	// XXX: Could use registry.VerifyAddress() here to filter out invalid addresses.
```
kostko: Good idea to prevent submitting invalid registrations.

go/registry/api/api.go
ptrus: Did you see the comment above about this breaking Genesis SanityCheck on a dump in existing deployments where validators have an invalid P2P ID / addresses? Should we allow invalid P2P keys/addresses for the next few versions in Genesis?

kostko: Oh I see, ok in this case file an issue so that we don't forget when this can be fixed (probably on next redeploy after all nodes have either expired or updated their registrations?). Also reference the issue in the comment and say why there is currently no such check.

ptrus: done: https://github.com/oasislabs/oasis-core/issues/2428

go/worker/registration/worker.go
ptrus: done

PR #2429, opened 2019-11-26
oasis-node/cmd: append git HEAD commit to oasis-node version

matevz:
When talking to node operators I noticed some of them used an older version of `oasis-node` binary. This PR runs `git rev-parse HEAD` during compilation and appends this information to the oasis node's version. Running `oasis-node --version` will now produce:
```
oa@matevz-oa:~/oasis-core/go$ oasis-node/oasis-node --version
Software version: 0.2.0-alpha+d8c21ad1
Runtime protocol version: 0.9.0
Consensus protocol version: 0.18.0
Committee protocol version: 0.6.0
Tendermint core version: 0.32.7
ABCI library version: 0.16.1
```

### DIFF
--- a/go/Makefile
+++ b/go/Makefile
@@ -11,6 +11,14 @@ OASIS_GO ?= go
 #   migrate to a version of Go that has the fix, it can be removed.
 GOFLAGS ?= -trimpath -ldflags=-buildid= -v
 
+# Append git HEAD commit to the oasis node version, if git command exists and
+# this folder is in some git repository. Use a truncated commit hash and append
+# +, if the branch is dirty.
+VERSION_BUILD = $(shell git describe --always --match "" --dirty=+ 2>/dev/null)
+ifneq ($(VERSION_BUILD),)
+	GOFLAGS_VERSION_EXTRA = -ldflags "-X github.com/oasislabs/oasis-core/go/common/version.Build=+$(VERSION_BUILD)"
+endif
+
 all: generate build
 
 # Generate required files.
@@ -24,7 +32,7 @@ generate:
 # Build all the things.
 build: generate
 	@echo "Building Oasis node"
-	@env -u GOPATH $(OASIS_GO) build $(GOFLAGS) -o ./oasis-node/oasis-node ./oasis-node
+	@env -u GOPATH $(OASIS_GO) build $(GOFLAGS) $(GOFLAGS_VERSION_EXTRA) -o ./oasis-node/oasis-node ./oasis-node
 	@echo "Building Oasis test harness"
 	@env -u GOPATH $(OASIS_GO) build $(GOFLAGS) -o ./oasis-test-runner/oasis-test-runner ./oasis-test-runner
 	@echo "Building Oasis local network runner"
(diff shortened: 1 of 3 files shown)


go/oasis-node/cmd/root.go
```diff
@@ -21,10 +21,13 @@ import (
 )
 
 var (
+	// versionBuild stores current git HEAD and is set by the linker.
+	versionBuild = ""
```
kostko: Should this live in `go/common/version/version.go` instead as we keep all the other versions there?

Yawning:
Maybe use the short commit hash since collisions are unlikely (and don't really matter in this context).
Maybe also show if the branch is dirty or not.

matevz:
> Maybe use the short commit hash since collisions are unlikely (and don't really matter in this context).
> Maybe also show if the branch is dirty or not.

Ok, I truncate to 7 characters as what github usually does. Also append `+` at the end of the commit, if the branch was dirty. At least that's how `git` cmd shows that a submodule is dirty.

kostko:
Yeah you can just do something like `git describe --always --dirty=+`?

matevz:
> Yeah you can just do something like `git describe --always --dirty=+`?

Um.. git describe gives me `0.1.0-3486-g5b126889`. This doesn't smell like the current HEAD :thinking:

kostko:
Ah yes you need to add something like `--match thistagwillneverexist` which seems hacky :/

matevz:
> Ah yes you need to add something like `--match thistagwillneverexist` which seems hacky :/

Actually `git describe --always --match "" --dirty=+` works just fine.

go/Makefile
```diff
@@ -11,6 +11,15 @@ OASIS_GO ?= go
 #   migrate to a version of Go that has the fix, it can be removed.
 GOFLAGS ?= -trimpath -ldflags=-buildid= -v
 
+# Append git HEAD commit to the oasis node version, if git command exists and
+# this folder is in some git repository.
+GIT_REVISION = $(shell git rev-parse HEAD 2>/dev/null)
+ifneq ($(GIT_REVISION),)
+	# Truncate the commit hash and append +, if the branch was dirty.
+	VERSION_BUILD = $(shell git describe --always --match "" --dirty=+)
```
kostko: Hm, why execute `git rev-parse HEAD` first and then `git describe ...`? Could you just do the describe or is there some reason why both are needed?

matevz: I thought we would need full `$GIT_REVISION` also somewhere else. But I guess `$VERSION_BUILD` with potential dirty suffix + at the end is perfectly fine. Fixed.

go/common/version/version.go
```diff
@@ -50,6 +50,9 @@ func (v Version) MajorMinor() Version {
 }
 
 var (
+	// VersionBuild stores current git HEAD and is set by the linker.
+	VersionBuild = ""
```
kostko: Since this is used as `version.VersionBuild` it would be better if it was just `Build` so used as `version.Build`. Probably the linter will also complain.

PR #2433, opened 2019-11-26
go/cmd/identity: add command for generating node identity keys

ptrus:
Fixes: https://github.com/oasislabs/oasis-core/issues/2413

Usage:
```
oasis-node/oasis-node identity init --datadir /tmp/test
Generated identity files in: /tmp/test
```


ptrus:
> can you add an E2E test for this new subcommand

done, please check again!

PR #2437, opened 2019-11-27
Use genesis document hash instead of ChainID for chain domain separation

### LINKED ISSUE #2410: Use genesis document hash for domain separation instead of just ChainID
Currently only ChainID is used for domain separation between different networks. A more reliable way (in light of "forgetting to change the chain ID") would be to just use the hash of the encoded genesis document.

We already pass the genesis file everywhere we require ChainID-based domain separation so this should be a trivial change.

kostko:
Fixes #2410 

This is a BREAKING change as it invalidates all signatures that require chain domain separation (e.g., transactions) and also changes the consensus ChainID.

### DIFF
--- a/go/consensus/tendermint/tendermint.go
+++ b/go/consensus/tendermint/tendermint.go
@@ -934,7 +934,7 @@ func genesisToTendermint(d *genesisAPI.Document) (*tmtypes.GenesisDoc, error) {
 	}
 
 	doc := tmtypes.GenesisDoc{
-		ChainID:     d.ChainID,
+		ChainID:     d.ChainContext()[:tmtypes.MaxChainIDLen],
 		GenesisTime: d.Time,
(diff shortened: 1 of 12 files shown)


go/consensus/tendermint/tendermint.go
```diff
@@ -934,7 +934,7 @@ func genesisToTendermint(d *genesisAPI.Document) (*tmtypes.GenesisDoc, error) {
 	}
 
 	doc := tmtypes.GenesisDoc{
-		ChainID:     d.ChainID,
+		ChainID:     d.ChainContext()[:tmtypes.MaxChainIDLen],
```
Yawning: So we always truncate here?

kostko: Yeah, unfortunately Tendermint has `MaxChainIDLen` at `50`. We have a fork so we could change it :-)

Issue #2376, opened 2019-11-13
Tests for registry CLI


kostko:
Similar to #2275 we should also have E2E tests for the registry CLI to make sure that we don't break it.

PR #2442, opened 2019-11-29
Matevz/feature/registry cli tests

matevz:
PR for #2376:
* adds `registry-cli` e2e test,
* adds `debug.dont_blame_oasis` requirement for `debug.test_entity` flag,
* replaces bunch of `--someflag` -> `--`+CfgSomeFlag,
* improves error msg for wrong entity folder permission.

### DIFF
--- a/go/oasis-node/cmd/registry/entity/entity.go
+++ b/go/oasis-node/cmd/registry/entity/entity.go
@@ -29,8 +29,8 @@ import (
 
 const (
 	cfgAllowEntitySignedNodes = "entity.debug.allow_entity_signed_nodes"
-	cfgNodeID                 = "entity.node.id"
-	cfgNodeDescriptor         = "entity.node.descriptor"
+	CfgNodeID                 = "entity.node.id"
+	CfgNodeDescriptor         = "entity.node.descriptor"
 
 	entityGenesisFilename = "entity_genesis.json"
 )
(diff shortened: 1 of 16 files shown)


go/oasis-node/cmd/registry/entity/entity.go
```diff
@@ -392,14 +392,15 @@ func Register(parentCmd *cobra.Command) {
 
 func init() {
 	entityFlags.Bool(cfgAllowEntitySignedNodes, false, "Entity signing key may be used for node registration (UNSAFE)")
+	_ = entityFlags.MarkHidden(cfgAllowEntitySignedNodes)
```
kostko: Nice catch, should we also feature-gate it behind the `debug.dont_blame_oasis` flag?

go/oasis-test-runner/scenario/e2e/registry_cli.go
```diff
@@ -0,0 +1,699 @@
...
+	}
+	if ent == nil {
+		return nil, fmt.Errorf("failed to load entity: no entity")
+	}
+	if signer.String() == "" {
+		return nil, fmt.Errorf("failed to load entity: empty signer")
+	}
+	if ent.ID.String() == "" {
+		return nil, fmt.Errorf("failed to load entity: empty entity")
+	}
```
kostko: I don't think you need to test these as you are basically testing file signer for correctness here (and this should be done in the signer's unit test instead).

go/oasis-node/cmd/registry/entity/entity.go
matevz: Done. Not sure how many tests will be broken by this. Let's see...

Issue #2434, opened 2019-11-26
make operators have to sign blocks to get the staking reward


pro-wh:
##### SUMMARY

we want a feature where we require liveness to be eligible for the staking reward. for that, the entity needs to have a node that signed at least some fraction of the blocks in an epoch.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

go/consensus/tendermint/apps/staking

kostko:
This was done.

PR #2443, opened 2019-12-03
go staking: reward for signing blocks

pro-wh:
blocks #2434

Give entities a reward on their staked amount when they sign 3/4 of the blocks of an epoch.

Relies on the property that an entity can have at most one node on the validator committee.

In this implementation, the staking mux app counts up how many blocks are created and how many each entity is seen signing. At the end of an epoch, we check which entities meet the 3/4 threshold and add rewards and reset the count.

### DIFF
--- a/go/consensus/tendermint/apps/staking/signing_rewards.go
+++ b/go/consensus/tendermint/apps/staking/signing_rewards.go
@@ -1,0 +1,84 @@
+package staking
+
+import (
+	"bytes"
+	"fmt"
+	"math"
+	"sort"
+
+	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
+	"github.com/oasislabs/oasis-core/go/consensus/tendermint/abci"
+	stakingState "github.com/oasislabs/oasis-core/go/consensus/tendermint/apps/staking/state"
+	epochtime "github.com/oasislabs/oasis-core/go/epochtime/api"
+	staking "github.com/oasislabs/oasis-core/go/staking/api"
+)
+
+const (
+	SigningThresholdNumerator   = 3
+	SigningThresholdDenominator = 4
+)
+
+func (app *stakingApplication) updateEpochSigning(ctx *abci.Context, signingEntities []signature.PublicKey) error {
+	stakeState := stakingState.NewMutableState(ctx.State())
+
+	epochSigning, err := stakeState.EpochSigning()
+	if err != nil {
+		return fmt.Errorf("loading epoch signing info: %w", err)
+	}
+
+	oldTotal := epochSigning.Total
+	epochSigning.Total = oldTotal + 1
+	if epochSigning.Total <= oldTotal {
+		return fmt.Errorf("incrementing total blocks count: overflow, old_total=%d", oldTotal)
+	}
+
+	for _, entityID := range signingEntities {
+		oldCount := epochSigning.ByEntity[entityID]
+		epochSigning.ByEntity[entityID] = oldCount + 1
+		if epochSigning.ByEntity[entityID] <= oldCount {
+			return fmt.Errorf("incrementing count for entity %s: overflow, old_count=%d", entityID, oldCount)
+		}
+	}
+
+	stakeState.SetEpochSigning(epochSigning)
+
+	return nil
+}
+
+func (app *stakingApplication) rewardEpochSigning(ctx *abci.Context, time epochtime.EpochTime) error {
+	stakeState := stakingState.NewMutableState(ctx.State())
+
+	epochSigning, err := stakeState.EpochSigning()
+	if err != nil {
+		return fmt.Errorf("loading epoch signing info: %w", err)
+	}
+
+	stakeState.ClearEpochSigning()
+
+	if epochSigning.Total == 0 {
+		return nil
+	}
+
+	var eligibleEntities []signature.PublicKey
+	if epochSigning.Total > math.MaxUint64/SigningThresholdNumerator {
+		return fmt.Errorf("determining eligibility: overflow in total blocks, total=%d", epochSigning.Total)
+	}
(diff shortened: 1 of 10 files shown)


go/consensus/tendermint/apps/staking/signing_rewards.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
+	"github.com/oasislabs/oasis-core/go/consensus/tendermint/abci"
+	stakingState "github.com/oasislabs/oasis-core/go/consensus/tendermint/apps/staking/state"
+	epochtime "github.com/oasislabs/oasis-core/go/epochtime/api"
+	staking "github.com/oasislabs/oasis-core/go/staking/api"
+)
+
+const (
+	SigningThresholdNumerator   = 3
```
kostko: These should probably be consensus parameters.

go/staking/api/api.go
```diff
@@ -29,6 +29,10 @@ const (
 )
 
 var (
+	// RewardFactorEpochSigned is the factor for a reward
+	// distributed per epoch to entities that have signed 3/4 of the blocks.
+	RewardFactorEpochSigned *quantity.Quantity
```
kostko: When would you change this to something different from 1? Couldn't the same be achieved by lowering the reward amount?

go/consensus/tendermint/apps/staking/signing_rewards.go
```diff
@@ -0,0 +1,84 @@
...
+
+	var eligibleEntities []signature.PublicKey
+	if epochSigning.Total > math.MaxUint64/SigningThresholdNumerator {
+		return fmt.Errorf("")
+	}
+	for entityID, count := range epochSigning.ByEntity {
+		if count > math.MaxUint64/SigningThresholdDenominator {
+			return fmt.Errorf("determining eligibility for entity %s: overflow in threshold comparison, count=%d", entityID, count)
+		}
+		if count*SigningThresholdDenominator < epochSigning.Total*SigningThresholdNumerator {
```
kostko: `epochSigning.Total*SigningThresholdNumerator` could be computed just once outside the loop?

go/staking/api/api.go
pro-wh: to control the relative sizes of rewards. currently this one is 1 and the one for election eligibility is 0. this "what you did" factor is multiplied by the reward schedule factor. and then later by the amount staked

go/consensus/tendermint/apps/staking/signing_rewards.go
pro-wh: yeah, although I leave that optimization to the compiler

pro-wh: https://godbolt.org/z/CXMYMy
aw it doesn't

go/consensus/tendermint/apps/staking/signing_rewards.go
```diff
@@ -0,0 +1,84 @@
...
+
+	stakeState.ClearEpochSigning()
+
+	if epochSigning.Total == 0 {
+		return nil
+	}
+
+	var eligibleEntities []signature.PublicKey
+	if epochSigning.Total > math.MaxUint64/SigningThresholdNumerator {
+		return fmt.Errorf("")
```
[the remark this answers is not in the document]
pro-wh: oh yeah

go/consensus/tendermint/apps/staking/state/state.go
```diff
@@ -378,6 +382,28 @@ func (s *ImmutableState) LastBlockFees() (*quantity.Quantity, error) {
 	return &q, nil
 }
 
+type EpochSigning struct {
+	Total    uint64
+	ByEntity map[signature.PublicKey]uint64
+}
+
+func (s *ImmutableState) EpochSigning() (EpochSigning, error) {
```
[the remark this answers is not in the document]
pro-wh: ok

go/consensus/tendermint/apps/staking/signing_rewards.go
pro-wh: aw

go/consensus/tendermint/apps/staking/signing_rewards.go
pro-wh: eh I might as well, since you went through the trouble of commenting about it

go/staking/api/api.go
kostko: Oh I see we didn't remove the scheduler-based rewards. Any reason why we should keep both?

pro-wh: - so we can support the scheduler one later, if we want to.

.......... or anyone using this codebase. if anyone's looking, this is not a statement about governance :speak_no_evil:

pro-wh: oops, gotta update that code comment too now that it's not hardcoded to 3/4

pro-wh:
ok so this causes some performance regressions to the point that CI is timing out. SGX E2E tests step 5 runs byzantine/compute-straggler in about 4 minutes, sentry in about 4 minutes, then stake-cli runs for 2 minutes and times out

pro-wh:
from a build on master https://buildkite.com/oasislabs/oasis-core-ci/builds/868#73e7522e-f955-4adc-87db-f85aed7667cc, compute-straggler normally takes 1 minute, sentry takes 2 minutes, and stake-cli takes 1 minute

pro-wh:
hm, yet steps 1-4 haven't gotten slower

pro-wh:
hm, the build prior to a comment change ran well within time too https://buildkite.com/oasislabs/oasis-core-ci/builds/882#2c142c10-4437-4ae9-8f52-2ece299e3bbf

Issue #2428, opened 2019-11-26
Genesis SanityCheck Addresses once existing deployments have cleaned up registry


ptrus:
Since existing deployments currently include invalid P2P.IDs and addresses we cannot yet sanity check those fields in genesis as it would prevent doing a import/export deployment. After next deployment, check if registry state is ok and can be sanity checked.

Validate P2P.ID and Consensus/Committee/P2P addresses after existing deployments have cleared up registry.

kostko:
While here, we should also deduplicate the node descriptor checks, possibly by having the registry doc's `SanityCheck` call `VerifyRegisterNodeArgs`.

PR #2448, opened 2019-12-05
go/registry: Reject node registrations with duplicate keys

### LINKED ISSUE #2421: Reject node registrations with duplicate keys
The registry should reject registrations of nodes with duplicate keys (e.g., consensus, p2p, etc. keys).

abukosek:
Closes #2421.

### DIFF
--- a/go/consensus/tendermint/apps/registry/transactions.go
+++ b/go/consensus/tendermint/apps/registry/transactions.go
@@ -162,7 +162,15 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
 		)
 		return err
 	}
-	newNode, err := registry.VerifyRegisterNodeArgs(params, app.logger, sigNode, untrustedEntity, ctx.Now(), ctx.IsInitChain(), regRuntimes)
+	regNodes, err := state.Nodes()
+	if err != nil {
+		app.logger.Error("RegisterNode: failed to obtain registry nodes",
+			"err", err,
+			"signed_node", sigNode,
+		)
+		return err
+	}
+	newNode, err := registry.VerifyRegisterNodeArgs(params, app.logger, sigNode, untrustedEntity, ctx.Now(), ctx.IsInitChain(), regRuntimes, regNodes)
 	if err != nil {
 		return err
 	}
--- a/go/registry/api/api.go
+++ b/go/registry/api/api.go
@@ -314,6 +314,7 @@ func VerifyRegisterNodeArgs( // nolint: gocyclo
 	now time.Time,
 	isGenesis bool,
 	regRuntimes []*Runtime,
+	regNodes []*node.Node,
 ) (*node.Node, error) {
 	var n node.Node
 	if sigNode == nil {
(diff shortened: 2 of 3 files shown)


go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -162,7 +162,15 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
 		)
 		return err
 	}
-	newNode, err := registry.VerifyRegisterNodeArgs(params, app.logger, sigNode, untrustedEntity, ctx.Now(), ctx.IsInitChain(), regRuntimes)
+	regNodes, err := state.Nodes()
```
kostko: Instead of fetching all nodes I think we should have indices by keys so we would need only a single lookup per registration per key. There can be a lot of nodes so this can otherwise get expensive.

go/registry/api/api.go
```diff
@@ -505,6 +506,46 @@ func VerifyRegisterNodeArgs( // nolint: gocyclo
...
+
+		if rn.Consensus.ID.Equal(n.Consensus.ID) {
+			logger.Error("RegisterNode: duplicate node consensus ID",
+				"node", n,
+				"existing_node", rn,
+			)
+			return nil, ErrInvalidArgument
+		}
+
+		if rn.P2P.ID.Equal(n.P2P.ID) {
```
ptrus: note that currently if multiple validators would use the same sentry node, they would register with same p2p ids

go/consensus/tendermint/apps/registry/transactions.go
abukosek: That's a good idea, but this can be optimized later (I can do it in next sprint if you want :).  We'd need multiple orders of magnitude more nodes on our testnet than we have now for this to become a bottleneck.

go/registry/api/api.go
abukosek: That will no longer work after this PR is merged.  I think the idea of issue #2421 was to discourage sharing of keys.

ptrus: no keys are shared, the nodes just use the same sentry node as a proxy (and register with sentries id and address, the way it's currently implemented). should probably explicitly enforce one validator per sentry if we plan to not support it anymore.

go/registry/tests/tester.go
```diff
@@ -598,6 +625,9 @@ func (ent *TestEntity) NewTestNodes(nCompute int, nStorage int, runtimes []*node
 
 		// Add a registration with no P2P addresses.
 		invalid1 := *nod.Node
+		invalid1.P2P.ID = randomPK(rng)
```
ptrus: Aren't all these cases actually registration of a node with same `node.ID`, so it shouldn't be needed to set random IDs for each case?

go/registry/api/api.go
ptrus: nevermind, it will be fine actually, as the `node.Consensus.Address.ID` will be duplicated, not the `node.Consensus.ID`, so it should be fine :+1:

go/registry/tests/tester.go
abukosek: The P2P and Consensus IDs must differ in any case, so this is actually required.

ptrus: but aren't they already unique? (lines 614 and 617)

abukosek: Ah, good point :D  I've removed the unnecessary lines now.

go/consensus/tendermint/apps/registry/transactions.go
kostko: Yeah, can you file an issue and assign yourself?

abukosek: Done! #2454 :)

ptrus:
Should probably also add sanity check of duplicates to the registry sanity checker: https://github.com/oasislabs/oasis-core/blob/ca872151c44cacaea46eb61cbf0cf9556efd8f14/go/registry/api/api.go#L1037

abukosek:
I left that out on purpose, as I assumed it will cause issues with existing node configs and would need to be done at a later time (similar to issue #2428).

PR #2453, opened 2019-12-06
Add fuzzing for transaction methods

### LINKED ISSUE #2245: Fuzz consensus interface
Consensus interface should be fuzzed, the following mux apps currently accept transactions:
* Staking.
* Registry.
* Roothash.
* Mock epochtime (only used for testing).

*Estimated cost: ~1 sprint*

### LINKED ISSUE #2246: Fuzz storage interface
Storage interface should be fuzzed.

jberci:
Closes #2245 
Closes #2246

### DIFF
--- a/go/consensus/tendermint/fuzz/fuzz.go
+++ b/go/consensus/tendermint/fuzz/fuzz.go
@@ -1,0 +1,101 @@
+// +build gofuzz
+
+package fuzz
+
+import (
+	"context"
+
+	"github.com/tendermint/tendermint/abci/types"
+
+	"github.com/oasislabs/oasis-core/go/common/cbor"
+	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
+	"github.com/oasislabs/oasis-core/go/common/crypto/signature/signers/memory"
+	"github.com/oasislabs/oasis-core/go/common/fuzz"
+	"github.com/oasislabs/oasis-core/go/consensus/api/transaction"
+	"github.com/oasislabs/oasis-core/go/consensus/tendermint/abci"
+	"github.com/oasislabs/oasis-core/go/consensus/tendermint/apps/epochtime_mock"
+	registryApp "github.com/oasislabs/oasis-core/go/consensus/tendermint/apps/registry"
+	stakingApp "github.com/oasislabs/oasis-core/go/consensus/tendermint/apps/staking"
+)
+
+var (
+	txSigner signature.Signer = memory.NewTestSigner("consensus-fuzz")
+
+	// FuzzableApps is a list of ABCI apps the fuzzer can fuzz.
+	FuzzableApps []abci.Application = []abci.Application{
+		epochtimemock.New(),
+		registryApp.New(),
+		stakingApp.New(),
+	}
+
+	// FuzzableMethods is a list of all the apps' transaction methods that are fuzzable.
+	FuzzableMethods []transaction.MethodName
+)
+
+func init() {
+	for _, app := range FuzzableApps {
+		FuzzableMethods = append(FuzzableMethods, app.Methods()...)
+	}
+}
+
+func Fuzz(data []byte) int {
+	var err error
+
+	ctx := context.Background()
+
+	var pruneCfg abci.PruneConfig
+
+	appConfig := &abci.ApplicationConfig{
+		DataDir:         "/tmp/oasis-node-fuzz-consensus",
+		Pruning:         pruneCfg,
+		HaltEpochHeight: 1000000,
+		MinGasPrice:     1,
+	}
+
+	// The muxer will start with the previous state, if it exists (the state database isn't cleared).
+	muxer, _ := abci.NewMockMux(ctx, appConfig)
+	defer muxer.MockClose()
+
+	for _, app := range FuzzableApps {
+		muxer.MockRegisterApp(app)
+	}
+
+	if len(data) < 2 {
+		return -1
+	}
+	meth := int(data[0])
+	if meth >= len(FuzzableMethods) {
+		return -1
+	}
+
+	methodName := FuzzableMethods[meth]
+
+	blob, _ := fuzz.NewFilledInstance(data, methodName.BodyType())
+
+	tx := &transaction.Transaction{
+		Method: methodName,
+		Body: cbor.Marshal(blob),
+	}
+
+	signedTx, err := transaction.Sign(txSigner, tx)
+	if err != nil {
+		panic("error signing transaction")
+	}
+
+	txBlob := cbor.Marshal(&signedTx)
+
+	// Check the transaction.
+	checkReq := types.RequestCheckTx{
+		Tx: txBlob,
+		Type: types.CheckTxType_New,
+	}
+	muxer.CheckTx(checkReq)
+
+	// And try executing it too.
+	deliverReq := types.RequestDeliverTx{
+		Tx: txBlob,
+	}
+	muxer.DeliverTx(deliverReq)
+
+	return 0
+}
--- a/go/storage/fuzz/fuzz.go
+++ b/go/storage/fuzz/fuzz.go
@@ -1,0 +1,62 @@
+// +build gofuzz
+
+package fuzz
+
+import (
+	"context"
+	"io/ioutil"
+
+	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
+	"github.com/oasislabs/oasis-core/go/common/crypto/signature/signers/file"
+	commonFuzz "github.com/oasislabs/oasis-core/go/common/fuzz"
+	"github.com/oasislabs/oasis-core/go/common/identity"
+	"github.com/oasislabs/oasis-core/go/storage"
+	"github.com/oasislabs/oasis-core/go/storage/api"
+)
+
+const (
+	dataDir string = "/tmp/oasis-node-fuzz-storage"
+	identityDir string = dataDir + "/identity"
+)
+
+var (
+	storageBackend api.Backend
+
+	fuzzer *commonFuzz.InterfaceFuzzer
+)
+
+func init() {
+	signerFactory := file.NewFactory(identityDir, signature.SignerNode, signature.SignerP2P, signature.SignerConsensus)
+	identity, err := identity.Load(identityDir, signerFactory)
+	if err != nil {
+		panic(err)
+	}
+
+	// Every Fuzz invocation should get its own database,
+	// otherwise the database handles would clash.
+	localDB, err := ioutil.TempDir(dataDir, "worker")
+	if err != nil {
+		panic(err)
+	}
+
+	// Create the storage backend service.
+	storageBackend, err = storage.New(context.Background(), localDB, identity, nil, nil)
+	if err != nil {
+		panic(err)
+	}
+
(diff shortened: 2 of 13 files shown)


go/consensus/tendermint/apps/fuzz/fuzz.go
```diff
@@ -0,0 +1,223 @@
+// +build gofuzz
```
kostko: Any reason why this is under `tendermint/apps/fuzz`? Wouldn't it make more sense to be under `tendermint/fuzz` directly?

go/consensus/tendermint/apps/fuzz/fuzz.go
```diff
@@ -0,0 +1,223 @@
...
+
+	"github.com/oasislabs/oasis-core/go/common/cbor"
+	"github.com/oasislabs/oasis-core/go/common/entity"
+	"github.com/oasislabs/oasis-core/go/common/fuzz"
+	"github.com/oasislabs/oasis-core/go/common/node"
+	"github.com/oasislabs/oasis-core/go/consensus/api/transaction"
+	"github.com/oasislabs/oasis-core/go/consensus/tendermint/abci"
+	"github.com/oasislabs/oasis-core/go/consensus/tendermint/apps/epochtime_mock"
+	registryApp "github.com/oasislabs/oasis-core/go/consensus/tendermint/apps/registry"
+	//roothashApp "github.com/oasislabs/oasis-core/go/consensus/tendermint/apps/roothash"
```
kostko: Remove commented out imports.

go/consensus/tendermint/apps/fuzz/fuzz.go
```diff
@@ -0,0 +1,223 @@
...
+					Name: "RegisterRuntime",
+					TxFunc: func(fuzzer *gofuzz.Fuzzer) interface{} {
+						var runtime registry.SignedRuntime
+						fuzzer.Fuzz(&runtime)
+						return &runtime
+					},
+				},
+			},
+		},
+		/*App{  // roothash
```
kostko: Remove commented-out apps and add a comment on why they are not there.

go/consensus/tendermint/apps/fuzz/fuzz.go
```diff
@@ -0,0 +1,223 @@
...
+)
+
+var (
+	// Fuzzy is the list of all apps using the consensus interfaces and transactions.
+	Fuzzy []App = []App{
+		App{ // epochtime_mock
+			Instance: epochtimemock.New(),
+			Methods: []Method{
+				Method{
+					Name: "SetEpoch",
```
kostko: You should reference the method name constants here.

go/consensus/tendermint/apps/fuzz/fuzz.go
jberci: I thought it makes more sense in `apps/` because the code is very specific to the transaction system we're using and it's directly related to our apps. It might make sense in the `abci/` directory alongside the muxer, but I'm not sure putting it in `tendermint/` would make sense.

kostko: I think of this in the same way as test helpers (which live under `tendermint/tests`). The `tendermint/apps` directory contains the apps themselves.

go/consensus/tendermint/apps/fuzz/fuzz.go
```diff
@@ -0,0 +1,135 @@
...
+	appConfig := &abci.ApplicationConfig{
+		DataDir:         "/tmp/fuzzytest",
+		Pruning:         pruneCfg,
+		HaltEpochHeight: 1000000,
+		MinGasPrice:     1,
+	}
+
+	muxer, closer, _ := abci.NewTestMuxer(ctx, appConfig)
+	defer closer()
+	abciCtx := abci.NewContext(abci.ContextCheckTx, time.Now(), muxer.GetApplicationState())
```
kostko: Using `CheckTx` will avoid most of the actual code paths, you should test both check and deliver.

go/consensus/tendermint/abci/mux.go
```diff
@@ -566,6 +566,12 @@ func (mux *abciMux) BeginBlock(req types.RequestBeginBlock) types.ResponseBeginB
 	return response
 }
 
+func (mux *abciMux) TestExecuteTx(tx []byte) error {
```
kostko: Where is this used?

go/consensus/tendermint/apps/fuzz/fuzz.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	methodName := Fuzzy[app].Methods[meth]
+
+	tx := &transaction.Transaction{
+		Method: methodName,
+		Body: cbor.Marshal(TxFunc(app, meth, fuzzer)),
+	}
+
+	err := Fuzzy[app].Instance.ExecuteTx(abciCtx, tx)
```
kostko: Hm so this doesn't mock a transaction signer for the transactions (see actual `executeTx` on the multiplexer which verifies the signature and sets the transaction signer)? Some transactions expect it being set to a valid value.

Also bypassing the multiplexer transaction processing doesn't invoke the configured `txAuthHandler`.

go/consensus/tendermint/apps/fuzz/fuzz.go
```diff
@@ -0,0 +1,135 @@
...
+	var pruneCfg abci.PruneConfig
+
+	appConfig := &abci.ApplicationConfig{
+		DataDir:         "/tmp/fuzzytest",
+		Pruning:         pruneCfg,
+		HaltEpochHeight: 1000000,
+		MinGasPrice:     1,
+	}
+
+	muxer, closer, _ := abci.NewTestMuxer(ctx, appConfig)
```
kostko: Hm so this always starts with empty application state?

go/consensus/tendermint/abci/mux.go
```diff
@@ -1135,6 +1156,10 @@ func (s *ApplicationState) metricsWorker() {
 	}
 }
 
+func NewApplicationStateTesting(ctx context.Context, cfg *ApplicationConfig) (*ApplicationState, error) {
```
kostko: Add documentation block.

go/consensus/tendermint/abci/mux.go
```diff
@@ -859,6 +869,17 @@ func (mux *abciMux) checkDependencies() error {
 	return nil
 }
 
+func (mux *abciMux) GetApplicationState() *ApplicationState {
+	return mux.state
+}
+
+func NewTestMuxer(ctx context.Context, cfg *ApplicationConfig) (*abciMux, func(), error) { // nolint
```
kostko: Hm so this exported function returns a private type (`abciMux`), how does this work?

go/consensus/tendermint/abci/mux.go
jberci: It used to be used in a few previous incarnations but dead now. Thanks.

go/consensus/tendermint/abci/mux.go
```diff
@@ -302,6 +302,16 @@ type abciMux struct {
 	invalidatedTxs sync.Map
 }
 
+type MockABCIMuxApp interface {
```
kostko: Add documentation comment to explain what this is and when it should be used.

go/consensus/tendermint/fuzz/fuzz.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/oasislabs/oasis-core/go/consensus/tendermint/abci"
+	"github.com/oasislabs/oasis-core/go/consensus/tendermint/apps/epochtime_mock"
+	registryApp "github.com/oasislabs/oasis-core/go/consensus/tendermint/apps/registry"
+	stakingApp "github.com/oasislabs/oasis-core/go/consensus/tendermint/apps/staking"
+)
+
+var (
+	txSigner signature.Signer = memory.NewTestSigner("consensus-fuzz")
+
+	FuzzableApps []abci.Application = []abci.Application{
```
kostko: Add doc comment (also below).

go/consensus/tendermint/apps/fuzz/fuzz.go
jberci: The database isn't cleared. Generally, all these fuzzers leave stuff behind so you can always restart.

go/storage/fuzz/fuzz.go
```diff
@@ -0,0 +1,53 @@
...
+
+func Fuzz(data []byte) int {
+	signerFactory := file.NewFactory(identityDir, signature.SignerNode, signature.SignerP2P, signature.SignerConsensus)
+	identity, err := identity.Load(identityDir, signerFactory)
+	if err != nil {
+		panic(err)
+	}
+
+	// Every fuzz worker should get its own database, otherwise we can't work.
+	localDB, err := ioutil.TempDir(dataDir, "worker")
```
kostko: How is this shared between invocations? Won't each `Fuzz` invocation start with its own database?

go/storage/fuzz/fuzz.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		panic(err)
+	}
+	defer os.RemoveAll(localDB)
+
+	// Create the storage backend service.
+	storage, err := storage.New(context.Background(), localDB, identity, nil, nil)
```
kostko: Would it be possible to only create the database once and have it shared between invocations? Or am I mistaken in how `Fuzz` is used?

go/storage/fuzz/fuzz.go
jberci: Yep, comment was mildly wrong.

go/storage/fuzz/fuzz.go
jberci: The fuzzer doesn't really provide any per-worker startup/shutdown hooks. I could use `init` for startup, but I can't see any solution for cleanup, which we'd need to properly close the database. When I tried running with just one database (and a single worker process), the library decided to stop working with a `resource temporarily unavailable`.

go/consensus/tendermint/abci/mux.go
```diff
@@ -799,6 +799,10 @@ func (mux *abciMux) doCleanup() {
 	}
 }
 
+func (mux *abciMux) MockRegisterApp(app Application) error {
```
ptrus: this should probably be behind the debug flag?

go/consensus/tendermint/abci/mux.go
```diff
@@ -858,6 +862,10 @@ func (mux *abciMux) checkDependencies() error {
 	return nil
 }
 
+func (mux *abciMux) MockClose() {
+	mux.doCleanup()
```
ptrus: this should probably be behind the debug flag?

ptrus: actually can these methods be removed from `abciMux` struct, and a `mockAbciMux` struct is created in the `mux_mock.go` that embeeds `abciMux` and adds these two methods?

kostko:
> should we setup this to be run daily/weekly/continiously on buildkite?

Yeah, but not as part of this PR. @jberci can you file an issue for running fuzzers on CI every once in a while.

Issue #2426, opened 2019-11-26
Remove Blocks field from roothash genesis


kostko:
The roothash genesis document includes a `Blocks` field that is used to populate the genesis block of a runtime that was just registered. This seems incorrect as the runtime structure already contains a `Genesis` field that has either a `StateRoot` or `State`.

I think we should:
* [x] Remove `Blocks` from roothash genesis.
* [x] Add `Round` to `RuntimeGenesis` and use that when initializing a runtime genesis block.
* [ ] ~Change `onNewRuntime` and `ForeignExecuteTx` to only use the runtime's `Genesis` structure.~
* [x] Update runtime-ethereum's genesis utilities to generate ~runtime descriptors instead of roothash state~ RuntimeGenesis map instead of Blocks.

kostko:
This was done in #2462.

matevz:
Note that oasis-runtime fix https://github.com/oasislabs/oasis-runtime/pull/958 for playback-genesis is blocked by https://github.com/oasislabs/oasis-runtime/issues/962

Issue #1686, opened 2019-05-08
Verify storage receipts for non-empty runtime genesis state roots


kostko:
After #1679 the registry should verify the provided storage receipts when a runtime is registered with a non-empty genesis state root. Currently such registrations are rejected.

kostko:
The question is -- which storage committee should sign these receipts as before there is a runtime, there are no committees for that runtime, so no storage nodes :-)

PR #2462, opened 2019-12-10
roothash/api: rename Genesis.Blocks -> RuntimeStates

matevz:
PR for #2426:
- replaces `roothash.Genesis.Blocks` with map storing `api.RuntimeGenesis` objects,
- adds `RuntimeGenesis.Round` and `--runtime.genesis.round` flag,
- on `InitChain()` generates runtime genesis block using `RuntimeGenesis` instead of block,
- replaces `RuntimeGenesis.StorageReceipt` with `StorageReceipts` i.e. one receipt per storage node,
- adds sanity checks for `RuntimeGenesis` including `StorageSignatures` verification of `StateRoot` and tests for it,
- adds `RuntimeGenesis` struct to rust,
- adds base64 serialization/deserialization of `Bytes` in rust.

### DIFF
--- a/go/consensus/tendermint/apps/roothash/genesis.go
+++ b/go/consensus/tendermint/apps/roothash/genesis.go
@@ -8,12 +8,13 @@ import (
 	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
 	"github.com/oasislabs/oasis-core/go/consensus/tendermint/abci"
 	registryState "github.com/oasislabs/oasis-core/go/consensus/tendermint/apps/registry/state"
-	genesisApi "github.com/oasislabs/oasis-core/go/genesis/api"
-	roothash "github.com/oasislabs/oasis-core/go/roothash/api"
-	"github.com/oasislabs/oasis-core/go/roothash/api/block"
+	genesisAPI "github.com/oasislabs/oasis-core/go/genesis/api"
+	"github.com/oasislabs/oasis-core/go/registry/api"
+	roothashAPI "github.com/oasislabs/oasis-core/go/roothash/api"
+	storageAPI "github.com/oasislabs/oasis-core/go/storage/api"
 )
 
-func (app *rootHashApplication) InitChain(ctx *abci.Context, request types.RequestInitChain, doc *genesisApi.Document) error {
+func (app *rootHashApplication) InitChain(ctx *abci.Context, request types.RequestInitChain, doc *genesisAPI.Document) error {
 	st := doc.RootHash
 
 	// The per-runtime roothash state is done primarily via DeliverTx, but
@@ -35,26 +36,25 @@ func (app *rootHashApplication) InitChain(ctx *abci.Context, request types.Reque
 	return nil
 }
 
-func (rq *rootHashQuerier) Genesis(ctx context.Context) (*roothash.Genesis, error) {
+func (rq *rootHashQuerier) Genesis(ctx context.Context) (*roothashAPI.Genesis, error) {
 	runtimes := rq.state.Runtimes()
 
-	// Get per-runtime blocks.
-	blocks := make(map[signature.PublicKey]*block.Block)
+	// Get per-runtime states.
+	rtStates := make(map[signature.PublicKey]*api.RuntimeGenesis)
 	for _, rt := range runtimes {
-		blk := *rt.CurrentBlock
-		// Header should be a normal header for genesis.
-		blk.Header.HeaderType = block.Normal
-		// There should be no previous hash.
-		blk.Header.PreviousHash.Empty()
-		// No messages.
-		blk.Header.Messages = nil
-		// No storage signatures.
-		blk.Header.StorageSignatures = []signature.Signature{}
-		blocks[rt.Runtime.ID] = &blk
+		rtState := api.RuntimeGenesis{
+			StateRoot: rt.CurrentBlock.Header.StateRoot,
+			// State is always empty in Genesis regardless of StateRoot.
+			State:           storageAPI.WriteLog{},
+			StorageReceipts: []signature.Signature{},
+			Round:           rt.CurrentBlock.Header.Round,
+		}
+
+		rtStates[rt.Runtime.ID] = &rtState
 	}
 
-	genesis := &roothash.Genesis{
-		Blocks: blocks,
+	genesis := &roothashAPI.Genesis{
+		RuntimeStates: rtStates,
 	}
 	return genesis, nil
 }
--- a/go/consensus/tendermint/apps/roothash/roothash.go
+++ b/go/consensus/tendermint/apps/roothash/roothash.go
@@ -382,12 +382,18 @@ func (app *rootHashApplication) onNewRuntime(ctx *abci.Context, runtime *registr
 	}
 
 	// Create genesis block.
-	genesisBlock := genesis.Blocks[runtime.ID]
-	if genesisBlock == nil {
-		now := ctx.Now().Unix()
-		genesisBlock = block.NewGenesisBlock(runtime.ID, uint64(now))
-		if !runtime.Genesis.StateRoot.IsEmpty() {
-			genesisBlock.Header.StateRoot = runtime.Genesis.StateRoot
+	now := ctx.Now().Unix()
+	genesisBlock := block.NewGenesisBlock(runtime.ID, uint64(now))
+	// Fill the Header fields with Genesis runtime states, if this was called during InitChain().
+	genesisBlock.Header.Round = runtime.Genesis.Round
+	genesisBlock.Header.StateRoot = runtime.Genesis.StateRoot
+	genesisBlock.Header.StorageSignatures = runtime.Genesis.StorageReceipts
+	if ctx.IsInitChain() {
+		genesisRts := genesis.RuntimeStates[runtime.ID]
+		if genesisRts != nil {
+			genesisBlock.Header.Round = genesisRts.Round
+			genesisBlock.Header.StateRoot = genesisRts.StateRoot
+			genesisBlock.Header.StorageSignatures = runtime.Genesis.StorageReceipts
 		}
 	}
 
--- a/go/oasis-node/cmd/debug/storage/export.go
+++ b/go/oasis-node/cmd/debug/storage/export.go
@@ -80,16 +80,20 @@ func doExport(cmd *cobra.Command, args []string) {
 	defer storageBackend.Cleanup()
 
 	// For each storage root.
-	for runtimeID, blk := range genesisDoc.RootHash.Blocks {
+	for runtimeID, rtg := range genesisDoc.RootHash.RuntimeStates {
 		logger.Info("fetching checkpoint write log",
 			"runtime_id", runtimeID,
 		)
 
+		// Use RuntimeID for the Roothash namespace.
+		var ns common.Namespace
+		_ = ns.UnmarshalBinary(runtimeID[:])
+
 		// Get the checkpoint iterator.
 		root := storageAPI.Root{
-			Namespace: blk.Header.Namespace,
-			Round:     blk.Header.Round,
-			Hash:      blk.Header.StateRoot,
+			Namespace: ns,
+			Round:     rtg.Round,
+			Hash:      rtg.StateRoot,
 		}
 		it, err := storageBackend.GetCheckpoint(context.Background(),
 			&storageAPI.GetCheckpointRequest{
@@ -108,7 +112,7 @@ func doExport(cmd *cobra.Command, args []string) {
 
 		fn := fmt.Sprintf("storage-dump-%v-%d.json",
 			runtimeID.String(),
-			blk.Header.Round,
+			rtg.Round,
 		)
 		fn = filepath.Join(destDir, fn)
 		if err = exportIterator(fn, &root, it); err != nil {
--- a/go/registry/api/api.go
+++ b/go/registry/api/api.go
@@ -889,8 +889,11 @@ func VerifyRegisterRuntimeArgs(logger *logging.Logger, sigRt *SignedRuntime, isG
 	}
 
 	if !isGenesis && !rt.Genesis.StateRoot.IsEmpty() {
-		// TODO: Verify storage receipt for the state root, reject such registrations for now.
+		// TODO: Verify storage receipt for the state root, reject such registrations for now. See oasis-core#1686.
 		return nil, ErrInvalidArgument
+	}
+	if err := rt.Genesis.SanityCheck(isGenesis); err != nil {
+		return nil, err
 	}
 
 	// Ensure there is at least one member of the compute group.
--- a/go/registry/api/runtime.go
+++ b/go/registry/api/runtime.go
@@ -231,10 +231,44 @@ type RuntimeGenesis struct {
 	StateRoot hash.Hash `json:"state_root"`
 
 	// State is the state identified by the StateRoot. It may be empty iff
-	// the StorageReceipt is not invalid or StateRoot is an empty hash.
+	// all StorageReceipts are valid or StateRoot is an empty hash or if used
+	// in network genesis (e.g. during consensus chain init).
 	State storage.WriteLog `json:"state"`
 
-	// StorageReceipt is the storage receipt for the state root. It may be
-	// invalid iff the State is non-empty or StateRoot is an empty hash.
-	StorageReceipt signature.Signature `json:"storage_receipt"`
+	// StorageReceipts are the storage receipts for the state root. The list
+	// may be empty or a signature in the list invalid iff the State is non-
+	// empty or StateRoot is an empty hash or if used in network genesis
+	// (e.g. during consensus chain init).
+	StorageReceipts []signature.Signature `json:"storage_receipts"`
+
+	// Round is the runtime round in the genesis.
+	Round uint64 `json:"round"`
+}
+
+// SanityCheck does basic sanity checking of RuntimeGenesis.
+// isGenesis is true, if it is called during consensus chain init.
+func (rtg *RuntimeGenesis) SanityCheck(isGenesis bool) error {
+	if isGenesis {
+		return nil
+	}
+
+	// Require that either State is non-empty or Storage receipt being valid or StateRoot being non-empty.
+	if len(rtg.State) == 0 && !rtg.StateRoot.IsEmpty() {
+		// If State is empty and StateRoot is not, then all StorageReceipts must correctly verify StorageRoot.
+		if len(rtg.StorageReceipts) == 0 {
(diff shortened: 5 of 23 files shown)


go/consensus/tendermint/apps/roothash/roothash.go
```diff
@@ -383,13 +383,20 @@ func (app *rootHashApplication) onNewRuntime(ctx *abci.Context, runtime *registr
...
-	genesisBlock := genesis.Blocks[runtime.ID]
-	if genesisBlock == nil {
-		now := ctx.Now().Unix()
-		genesisBlock = block.NewGenesisBlock(runtime.ID, uint64(now))
-		if !runtime.Genesis.StateRoot.IsEmpty() {
-			genesisBlock.Header.StateRoot = runtime.Genesis.StateRoot
+	now := ctx.Now().Unix()
+	genesisBlock := block.NewGenesisBlock(runtime.ID, uint64(now))
+	// Fill the Header fields with Genesis runtime states, if this was called during InitChain().
+	if ctx.IsInitChain() {
```
kostko: So this is not entirely correct as it can happen that this is in InitChain but there is no `genesis.RuntimesStates` defined. In that case the genesis state supplied in the runtime descriptor should still be used (if any).

go/oasis-node/cmd/debug/byzantine/compute.go
```diff
@@ -173,7 +173,7 @@ func (cbc *computeBatchContext) createCommitment(id *identity.Identity, rak sign
 		InputStorageSigs:  cbc.bd.StorageSignatures,
 	}
 	if rak != nil {
-		rakSig, err := signature.Sign(rak, commitment.ComputeResultsHeaderSignatureContext, cbor.Marshal(header))
+		rakSig, err := signature.Sign(rak, signature.ComputeResultsHeaderSignatureContext, cbor.Marshal(header))
```
kostko: Wtf?

go/common/crypto/signature/signer.go
```diff
@@ -17,6 +17,14 @@ const (
 )
 
 var (
+	// ComputeSignatureContext is the signature context used to sign compute
```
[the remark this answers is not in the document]
matevz: I verify the `StorageReceipt` signature in `RuntimeGenesis.SanityCheck()` and I need `ComputeSignatureContext` for the context. To avoid circular dependencies I moved `ComputeSignatureContext` from `roothash/api/commitment` to `common/crypto/signature`.

go/oasis-node/cmd/debug/byzantine/compute.go
matevz: Yeah, see my first comment. ;)

go/consensus/tendermint/apps/roothash/genesis.go
```diff
@@ -35,26 +36,29 @@ func (app *rootHashApplication) InitChain(ctx *abci.Context, request types.Reque
...
-		blk.Header.StorageSignatures = []signature.Signature{}
-		blocks[rt.Runtime.ID] = &blk
+		rtState := api.RuntimeGenesis{
+			StateRoot: rt.CurrentBlock.Header.StateRoot,
+			// State is always empty in Genesis regardless of StateRoot.
+			State: storageAPI.WriteLog{},
+			Round: rt.CurrentBlock.Header.Round,
+		}
+
+		// TODO Matevz: I'm simply taking the first StorageSignature, if it exists. Is this ok?
```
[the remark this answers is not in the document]
matevz: Um.. but isn't ReceiptStorage just signed StateRoot?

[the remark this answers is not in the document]
kostko: There can be multiple storage nodes. Each receipt is from one storage node.

go/common/crypto/signature/signer.go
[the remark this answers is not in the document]
Yawning: Oh hell no.  If you absolutely need to do something like this, carve out another sub-package under `roothash/api` for it.  Don't move context declarations into `common/crypto/signature`.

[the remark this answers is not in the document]
matevz: I will not check the signatures becaues I also need public keys from the storage nodes which I don't have during the sanity check. I will move these contexts back and clean it up. For signature verification, there is a separate issue https://github.com/oasislabs/oasis-core/issues/1686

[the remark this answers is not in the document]
matevz: Done.

go/consensus/tendermint/apps/roothash/genesis.go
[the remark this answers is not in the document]
matevz: Done.

.changelog/2426.feature.md
```diff
@@ -0,0 +1,7 @@
+Refactoring of roothash genesis block for runtime.
```
kostko: This seems like a breaking change so it should be `.breaking.md`.

go/oasis-node/cmd/debug/storage/export.go
```diff
@@ -80,16 +81,23 @@ func doExport(cmd *cobra.Command, args []string) {
 	defer storageBackend.Cleanup()
 
 	// For each storage root.
-	for runtimeID, blk := range genesisDoc.RootHash.Blocks {
+	for runtimeID, rtg := range genesisDoc.RootHash.RuntimeStates {
 		logger.Info("fetching checkpoint write log",
 			"runtime_id", runtimeID,
 		)
 
+		// Roothash has empty namespace.
```
kostko: This doesn't seem correct, the namespace should be the runtime ID.

go/registry/api/runtime.go
```diff
@@ -231,10 +231,43 @@ type RuntimeGenesis struct {
...
+	// empty or StateRoot is an empty hash or if used in network genesis
+	// (e.g. during Tendermint InitChain).
+	StorageReceipts []signature.Signature `json:"storage_receipts"`
+
+	// Round is the runtime round in the genesis.
+	Round uint64 `json:"round"`
+}
+
+// SanityCheck does basic sanity checking of RuntimeGenesis.
+// isGenesis is true, if it is called during Tendermint's InitChain.
```
kostko: The comment shouldn't involve Tendermint, say "during consensus chain init" or "network genesis" or something (unify with docs in field descriptions above).

go/registry/api/api.go
```diff
@@ -888,9 +888,8 @@ func VerifyRegisterRuntimeArgs(logger *logging.Logger, sigRt *SignedRuntime, isG
 		return nil, ErrInvalidArgument
 	}
 
-	if !isGenesis && !rt.Genesis.StateRoot.IsEmpty() {
```
kostko: Are you sure this same check is covered in `SanityCheck`? So if in non-genesis, registrations which use a non-empty `StateRoot` must be rejected until we have a resolution to #1686.

go/registry/api/runtime.go
```diff
@@ -231,10 +231,43 @@ type RuntimeGenesis struct {
...
+			return fmt.Errorf("runtimegenesis: sanity check failed: when State is empty either StorageReceipts must be populated or StateRoot being non-empty")
+		}
+		for _, sr := range rtg.StorageReceipts {
+			if !sr.PublicKey.IsValid() {
+				return fmt.Errorf("runtimegenesis: sanity check failed: when State is empty either all StorageReceipts must be valid or StateRoot being non-empty. storage_receipt_public_key %s", sr.PublicKey)
+			}
+
+			if !sr.Verify(storage.ReceiptSignatureContext, rtg.StateRoot[:]) {
+				return fmt.Errorf("runtimegenesis: sanity check failed: StorageReceipt verification on StateRoot failed. storage_receipt_public_key %s", sr.PublicKey)
+			}
```
kostko: Even if `Verify` succeeds, runtime registration should still be rejected until #1686 is solved. Add a comment that this is temporary and reference issue #1686.

go/registry/api/runtime.go
```diff
@@ -231,10 +231,43 @@ type RuntimeGenesis struct {
...
+func (rtg *RuntimeGenesis) SanityCheck(isGenesis bool) error {
+	if isGenesis {
+		return nil
+	}
+
+	// Require that either State is non-empty or Storage receipt being valid or StateRoot being non-empty.
+	if len(rtg.State) == 0 && !rtg.StateRoot.IsEmpty() {
+		// If State is empty and StateRoot is not, then all StorageReceipts must correctly verify StorageRoot.
+		if len(rtg.StorageReceipts) == 0 {
+			return fmt.Errorf("runtimegenesis: sanity check failed: when State is empty either StorageReceipts must be populated or StateRoot being non-empty")
```
kostko: This message doesn't seem correct, shouldn't this be something like "... either StorageReceipts must be valid or StateRoot must be an empty hash"?

go/registry/api/runtime.go
```diff
@@ -231,10 +231,43 @@ type RuntimeGenesis struct {
...
+		}
+		for _, sr := range rtg.StorageReceipts {
+			if !sr.PublicKey.IsValid() {
+				return fmt.Errorf("runtimegenesis: sanity check failed: when State is empty either all StorageReceipts must be valid or StateRoot being non-empty. storage_receipt_public_key %s", sr.PublicKey)
```
kostko: Fix message as above and include the public key in parenthesis, e.g. ` (public key: %s)`.

go/registry/api/runtime.go
```diff
@@ -231,10 +231,43 @@ type RuntimeGenesis struct {
...
+			}
+
+			if !sr.Verify(storage.ReceiptSignatureContext, rtg.StateRoot[:]) {
+				return fmt.Errorf("runtimegenesis: sanity check failed: StorageReceipt verification on StateRoot failed. storage_receipt_public_key %s", sr.PublicKey)
```
kostko: Same as above.

go/roothash/api/api.go
```diff
@@ -169,42 +170,14 @@ func SanityCheckBlocks(blocks map[signature.PublicKey]*block.Block) error {
...
-			return fmt.Errorf("roothash: sanity check failed: non-empty roothash messages")
-		}
-	}
-	return nil
-}
-
 // SanityCheck does basic sanity checking on the genesis state.
-func (g *Genesis) SanityCheck() error {
+// isGenesis is true, if it is called during the Tendermint's InitChain.
+func (g *Genesis) SanityCheck(isGenesis bool) error {
```
kostko: When would `SanityCheck` on a genesis document be called with `isGenesis` set to false? I think this argument should be removed and the below `SanityCheck` always called with `true`.

runtime/src/common/bytes.rs
```diff
@@ -171,37 +175,90 @@ macro_rules! impl_bytes {
...
+    // Use hash of an empty string as a test key.
+    const TEST_KEY_BYTES: [u8; 32] = [
+        0xc6, 0x72, 0xb8, 0xd1, 0xef, 0x56, 0xed, 0x28, 0xab, 0x87, 0xc3, 0x62, 0x2c, 0x51, 0x14,
+        0x06, 0x9b, 0xdd, 0x3a, 0xd7, 0xb8, 0xf9, 0x73, 0x74, 0x98, 0xd0, 0xc0, 0x1e, 0xce, 0xf0,
+        0x96, 0x7a,
+    ];
+
+    #[test]
+    fn test_serde_base64() {
+        // Serialize
```
kostko:
```suggestion
        // Serialize.
```
(also in comments below)

runtime/src/common/registry.rs
```diff
@@ -0,0 +1,35 @@
...
+//!
+//! This **MUST** be kept in sync with go/registry/api.
+//!
+use super::{
+    super::storage::mkvs::WriteLog,
+    crypto::{hash, signature::SignatureBundle},
+};
+use serde_derive::{Deserialize, Serialize};
+
+// RuntimeGenesis is the runtime genesis information that is used to
```
kostko: Should use Rust documentation comments (`///`) and not include the `<name> is` prefix. Also for fields below.

go/registry/api/runtime.go
```diff
@@ -231,10 +231,43 @@ type RuntimeGenesis struct {
 	StateRoot hash.Hash `json:"state_root"`
 
 	// State is the state identified by the StateRoot. It may be empty iff
-	// the StorageReceipt is not invalid or StateRoot is an empty hash.
+	// all StorageReceipts are valid or StateRoot is an empty hash or if used
+	// in network genesis (e.g. during Tendermint InitChain).
```
kostko: Remove the tendermint thing.

go/registry/api/runtime.go
```diff
@@ -231,10 +231,43 @@ type RuntimeGenesis struct {
...
 	State storage.WriteLog `json:"state"`
 
-	// StorageReceipt is the storage receipt for the state root. It may be
-	// invalid iff the State is non-empty or StateRoot is an empty hash.
-	StorageReceipt signature.Signature `json:"storage_receipt"`
+	// StorageReceipts are the storage receipts for the state root. The list
+	// may be empty or a signature in the list invalid iff the State is non-
+	// empty or StateRoot is an empty hash or if used in network genesis
+	// (e.g. during Tendermint InitChain).
```
kostko: Remove the tendermint thing.

runtime/src/common/bytes.rs
```diff
@@ -171,37 +175,90 @@ macro_rules! impl_bytes {
...
                     where
-                        A: ::serde::de::SeqAccess<'de>,
+                        E: ::serde::de::Error,
                     {
                         let mut array = [0; $size];
-                        for i in 0..$size {
-                            array[i] = seq
-                                .next_element()?
-                                .ok_or_else(|| ::serde::de::Error::invalid_length(i, &self))?;
+                        let bytes = base64::decode(data).unwrap();
```
kostko: This should not unwrap, otherwise a serialization error will cause the program to panic instead of returning a serialization error. Handle the error and return a proper error type instead.

go/registry/api/api.go
matevz: Yeah, `SanityCheck()` only covers the cases specified in `RuntimeGenesis` struct. I put the `!isGenesis && !rt.Genesis.StateRoot.IsEmpty()` back in now and a reference to #1686 .

go/registry/api/runtime.go
matevz: Yeah. I rewrote this code so many times with all the double negations that I missed it.

go/roothash/api/api.go
matevz: Ok, I left the internal RuntimeGenesis `isGenesis` parameter alone and only remove it from the roothash's one.

go/registry/api/runtime.go
```diff
@@ -231,10 +231,44 @@ type RuntimeGenesis struct {
...
+
+	// Require that either State is non-empty or Storage receipt being valid or StateRoot being non-empty.
+	if len(rtg.State) == 0 && !rtg.StateRoot.IsEmpty() {
+		// If State is empty and StateRoot is not, then all StorageReceipts must correctly verify StorageRoot.
+		if len(rtg.StorageReceipts) == 0 {
+			return fmt.Errorf("runtimegenesis: sanity check failed: when State is empty either StorageReceipts must be populated or StateRoot must be empty")
+		}
+		for _, sr := range rtg.StorageReceipts {
+			if !sr.PublicKey.IsValid() {
+				return fmt.Errorf("runtimegenesis: sanity check failed: when State is empty either all StorageReceipts must be valid or StateRoot must be empty. (public_key %s)", sr.PublicKey)
```
kostko:
```suggestion
				return fmt.Errorf("runtimegenesis: sanity check failed: when State is empty either all StorageReceipts must be valid or StateRoot must be empty (public key: %s)", sr.PublicKey)
```
(also below)

go/roothash/api/api.go
```diff
@@ -169,42 +170,14 @@ func SanityCheckBlocks(blocks map[signature.PublicKey]*block.Block) error {
...
-
-		if len(hdr.Messages) != 0 {
-			return fmt.Errorf("roothash: sanity check failed: non-empty roothash messages")
-		}
-	}
-	return nil
-}
-
 // SanityCheck does basic sanity checking on the genesis state.
+// isGenesis is true, if it is called during the Tendermint's InitChain.
```
kostko: This no longer exists so the comment should be removed.

runtime/src/common/bytes.rs
```diff
@@ -171,37 +175,97 @@ macro_rules! impl_bytes {
...
                     }
 
-                    fn visit_bytes<E>(self, data: &[u8]) -> Result<$name, E>
+                   fn visit_bytes<E>(self, data: &[u8]) -> Result<$name, E>
                     where
                         E: ::serde::de::Error,
                     {
-                        let mut array = [0; $size];
-                        if data.len() != $size {
+                        if data.len()!=$size {
```
kostko:
```suggestion
                        if data.len() != $size {
```

runtime/src/common/bytes.rs
```diff
@@ -171,37 +175,97 @@ macro_rules! impl_bytes {
...
+        // Serialize.
+        impl_bytes!(TestKey, 32, "test key");
+        let test_key = TestKey(TEST_KEY_BYTES);
+        let test_key_str = serde_json::to_string(&test_key).unwrap();
+        assert_eq!(
+            test_key_str,
+            "\"xnK40e9W7Sirh8NiLFEUBpvdOte4+XN0mNDAHs7wlno=\""
+        );
+
+        // Deserialize
```
kostko:
```suggestion
        // Deserialize.
```

runtime/src/common/bytes.rs
```diff
@@ -171,37 +175,97 @@ macro_rules! impl_bytes {
...
+        let new_test_key: TestKey = serde_json::from_str(&test_key_str).unwrap();
+        assert_eq!(new_test_key, test_key);
+    }
+
+    #[test]
+    fn test_serde_cbor() {
+        // Serialize
```
kostko:
```suggestion
        // Serialize.
```

runtime/src/common/bytes.rs
```diff
@@ -171,37 +175,97 @@ macro_rules! impl_bytes {
...
+
+        let test_key = TestKey(TEST_KEY_BYTES);
+        let test_key_vec = serde_cbor::to_vec(&test_key).unwrap();
+
+        // CBOR prepends "X " to the binary value.
+        let mut expected_test_key_vec = vec![88, 32];
+        expected_test_key_vec.extend_from_slice(&TEST_KEY_BYTES);
+        assert_eq!(test_key_vec, expected_test_key_vec);
+
+        // Deserialize
```
kostko:
```suggestion
        // Deserialize.
```

PR #2463, opened 2019-12-10
go/runtime: Per-runtime block history and prune policy

### LINKED ISSUE #1920: Independent roothash pruning policy
Currently the Ekiden runtime pruning policy is coupled with the Tendermint ABCI state pruning policy. This makes it hard to control how runtime history is kept as there is no simple correlation between Tendermint blocks and roothash rounds.

This should be changed so that the Ekiden runtime pruning policy becomes independent. In order to support this even while Tendermint ABCI state may be pruned, we need to make sure to keep a copy of all non-pruned Ekiden runtime blocks. This can be achieved either by:
* Keeping them in a separate index (e.g., the existing roothash block index which can be moved from the tendermint roothash backend and repurposed).
* Turning the pruning policy around so that any Tendermint blocks which are still required by the Ekiden runtime pruning policy are not pruned (e.g., the new Ekiden runtime pruning policy directs what is pruned).

The benefit of the first approach is that we can independently prune consensus and the downside is that we would duplicate some amount of state (e.g., Ekiden runtime blocks that would exist in both ABCI state until pruned and in this new index). Note that the separate index is needed even in the second case in order to map round numbers to consensus heights (this is the current implementation).

The refactored pruning policy support must also take into account that storage needs to use it in order to know what to sync and where checkpoints (e.g., `GetCheckpoint`) can be used to speed up syncing.

*Estimated cost: 1 sprint*

### LINKED ISSUE #1922: Common runtimes flag
Currently indexer and worker node runtimes use separate flags, `client.indexer.runtimes` & `worker.runtime.id`. Might make sense refactoring that to a single common `--runtimes` flag.

kostko:
Based on #2436 
Fixes #1920 
Fixes #1922 

This breaks the runtime-related configuration flags:

- Renamed --client.indexer.backend to
  --runtime.history.tag_indexer.backend. The default has changed to an
  empty string (disabled) and must now be explicitly set to "bleve" to
  enable.

- Removed --worker.runtime.id and --client.indexer.runtimes and replaced
  them with a single --runtime.supported flag.

- Changed value format of --worker.runtime.binary which now takes a map
  of runtime ID to binary path in the form of `<hex-runtime-ID>:<path>`.

This adds the following new configuration flags:

  --runtime.history.pruner.strategy
  --runtime.history.pruner.interval
  --runtime.history.pruner.num_kept

TODO
* [x] Roothash block history interface.
* [x] Common runtime configuration endpoint (`go/runtime/registry`) with unified runtime configuration.
* [x] Change workers to use the common runtime configuration endpoint.
* [x] Change runtime client to use the common runtime configuration endpoint.
* [x] Port block indexer.
* [x] Add tests for the new interfaces.
* [x] Implement prune policy.
* [x] Move `go/runtime/client/indexer` to `go/runtime/tagindexer`.
* [x] Use per-runtime indexer database.


go/oasis-test-runner/scenario/e2e/basic.go
```diff
@@ -117,7 +117,7 @@ func (sc *basicImpl) Fixture() (*oasis.NetworkFixture, error) {
 					GroupSize:         1,
 					MaxBatchSize:      1,
 					MaxBatchSizeBytes: 1000,
-					BatchFlushTimeout: 10 * time.Second,
+					BatchFlushTimeout: 1 * time.Second,
```
Yawning: Buildkite can be kind of slow is this ok?

go/worker/common/committee/node.go
```diff
@@ -249,61 +248,22 @@ func (n *Node) handleNewEventLocked(ev *roothash.Event) {
...
 func (n *Node) worker() {
 	n.logger.Info("starting committee node")
 
 	defer close(n.quitCh)
 	defer (n.cancelCtx)()
 
 	// Wait for the runtime.
-	if err := n.waitForRuntime(); err != nil {
+	if _, err := n.Runtime.RegistryDescriptor(n.ctx); err != nil {
+		n.logger.Error("failed to wait fo registry descriptor",
```
Yawning: `fo`->`for`

go/oasis-test-runner/scenario/e2e/basic.go
kostko: Yeah because this is not a timeout in the failing sense but it is how long the scheduler will wait before flushing a batch. A longer interval will just make tests wait longer for no good reason.

kostko:
> Don't forget to remove bbolt from the go.mod, nothing should be using it after this commit.

Done.

PR #2464, opened 2019-12-11
go/registry: Optimize check for duplicate keys

### LINKED ISSUE #2454: go/registry: Optimize check for duplicate keys on node registration
On new node registration, we currently go through all registered nodes to check that the new node's consensus and P2P keys are unique, which is not very optimal.
It would be better to have a map of keys to node IDs, so that we can check if a key already exists with a simple lookup.

abukosek:
Closes #2454.

### DIFF
--- a/go/consensus/tendermint/apps/registry/state/state.go
+++ b/go/consensus/tendermint/apps/registry/state/state.go
@@ -6,6 +6,7 @@ import (
 	"github.com/tendermint/iavl"
 
 	"github.com/oasislabs/oasis-core/go/common/cbor"
+	"github.com/oasislabs/oasis-core/go/common/crypto/hash"
 	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
 	"github.com/oasislabs/oasis-core/go/common/entity"
 	"github.com/oasislabs/oasis-core/go/common/keyformat"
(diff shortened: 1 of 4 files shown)


go/consensus/tendermint/apps/registry/state/state.go
```diff
@@ -50,6 +51,16 @@ var (
...
+	// keyMapKeyFmt is the key format used for key-to-node-id map.
+	// This stores the consensus and P2P to Node ID mappings.
+	//
+	// Value is binary signature.PublicKey (node ID).
+	keyMapKeyFmt = keyformat.New(0x17, &signature.PublicKey{})
+	// certificateMapKeyFmt is the key format used for certificate-to-node-id map.
+	// This stores the hash-of-certificate to Node ID mappings.
+	//
+	// Value is binary signature.PublicKey (node ID).
+	certificateMapKeyFmt = keyformat.New(0x18, []byte{})
```
kostko: Why `[]byte{}` instead of `&hash.Hash{}`?

abukosek: Oops, fixed! :)

go/consensus/tendermint/apps/registry/state/state.go
```diff
@@ -390,6 +401,40 @@ func (s *ImmutableState) ConsensusParameters() (*registry.ConsensusParameters, e
 	return &params, err
 }
 
+func (s *ImmutableState) NodeByConsensusOrP2PKey(key signature.PublicKey) (*node.Node, error) {
```
kostko: Add type assertion at the top that `ImmutableState` implements `NodeLookup`.

abukosek:
Thanks!

PR #2468, opened 2019-12-12
Temporarily disable p2p id checks (for public testnet upgrade)

peterjgilbert:
Related to https://github.com/oasislabs/oasis-core/issues/2428. For compatibility with dumps from networks running pre-https://github.com/oasislabs/oasis-core/pull/2424 versions of oasis-node, we need to accept 0-valued p2p ids.

### DIFF
--- a/go/registry/api/api.go
+++ b/go/registry/api/api.go
@@ -501,10 +501,12 @@ func VerifyRegisterNodeArgs( // nolint: gocyclo
 
 	// Validate P2PInfo.
 	if !n.P2P.ID.IsValid() {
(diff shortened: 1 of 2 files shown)


Yawning:
I'm moderately opposed to this out of principle, but if migration is required then I suppose there's no other option.  Can you make an issue to revert the removal of the checks so that we do not forget?

peterjgilbert:
> I guess this is fine. Make sure to file an issue to revert this before merging please.

No need to merge this one. We will do the testnet redeploy from branch `oasis-testnet-p2-2019-12-17`.

PR #2470, opened 2019-12-13
Reorganize Makefiles

tjanez:
This will make integrating release tooling easier since it makes information such as version info and linker flags available from all Makefiles.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -1,0 +1,47 @@
+SHELL := /bin/bash
+
+# Check if we're running in an interactive terminal.
+ISATTY := $(shell [ -t 0 ] && echo 1)
+
+ifdef ISATTY
+	# Running in interactive terminal, OK to use colors!
+	MAGENTA := \e[35;1m
+	CYAN := \e[36;1m
+	OFF := \e[0m
+
+	# Built-in echo doesn't support '-e'.
+	ECHO = /bin/echo -e
+else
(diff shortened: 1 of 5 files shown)


common.mk
```diff
@@ -0,0 +1,47 @@
...
+	MAGENTA := ""
+	CYAN := ""
+	OFF := ""
+
+	# OK to use built-in echo.
+	ECHO := echo
+endif
+
+# Try to determine Oasis Core's git revision and if the working directory is
+# dirty
```
kostko:
```suggestion
# dirty.
```

go/Makefile
```diff
@@ -1,69 +1,75 @@
...
-
-# Append git HEAD commit to the oasis node version, if git command exists and
-# this folder is in some git repository. Use a truncated commit hash and append
-# +, if the branch is dirty.
-VERSION_BUILD = $(shell git describe --always --match "" --dirty=+ 2>/dev/null)
-ifneq ($(VERSION_BUILD),)
-	GOFLAGS_VERSION_EXTRA = -ldflags "-X github.com/oasislabs/oasis-core/go/common/version.Build=+$(VERSION_BUILD)"
+include ../common.mk
+
+ifneq ($(GOLDFLAGS),)
```
kostko: Explain where `GOLDFLAGS` comes from.

PR #2472, opened 2019-12-16
Deprecate old/unused db backends

### LINKED ISSUE #2050: Clean up persistent storage backends
We currently support the following persistent storage backends:
* LevelDB
* Badger

Both share a lot of common code and maintaining duplicate implementations seems to be overkill. One possibility is to create an abstract "Key-Value DB" interface (e.g., key/value operations, iterators, etc.) and then implement the NodeDB in terms of it (this is what Tendermint does for example).

Doing so, however, may prevent the backends from taking advantage of specific database features which makes supporting multiple KVDB backends not that useful (and piles further abstractions on top of storage). So the question is if something like this is needed or is it enough if we just remove one of the backends going forward?

I would be in favor of just removing one of the backends.

Yawning:
Fixes #2050 

To reduce maintenance burden, we should prune the various database types that we support.
 * [x] Remove our tendermint boltdb backend, default to badger.
 * [x] Remove the leveldb storage backend, default to badger.

While I am here, the following changes could/should be made:
 * [x] Bump badger to a newer release (WARNING: Import `github.com/dgraph-io/badger/v2`).
 * [x] Use badger's native compression support.  There's two options, zstd, and snappy.  We opt for the latter.
 * [x] Move `common/persistent` from boltdb -> badgerdb.
 * [x] Move `worker/common/host/local_storage.go` from boltdb -> badgerdb.
 * Move the indexer to badgerdb (done in #2463)


kostko:
> Move the indexer to badgerdb (done in #2436)

You probably mean #2463.

Issue #1829, opened 2019-06-19
Support access control policy filtering on GRPC Proxies


ravenac95:
We need to expose the access control policy for GRPC proxies to use to filter traffic ingress

## Details

We would like to support an architecture that would look something like this for grpc and libp2p proxies:

![LibP2P_GRPC Proxy Architecture](https://user-images.githubusercontent.com/862590/59773486-8eb22280-9262-11e9-817d-653237468e17.png)

General Ideas:
* The proxy layer should terminate TLS and validate the Client TLS certificate is from a valid node identity in the access control policy. 

## Acceptance Criteria
* An interface is designed to properly expose the required access control policy to support a GRPC proxy
* The Ekiden-Node can be configured to accept proxied requests over GRPC.

Issue #1930, opened 2019-07-17
go/worker: Don't always start the registration worker


Yawning:
Right now the registration worker is always spun up, if workers are.  This is somewhat silly and it should only happen if the node needs to actually register the node descriptor with the registry (the client spins up workers, so it also pulls this in, but it doesn't have an entity etc).

PR #2475, opened 2019-12-16
go/grpc: sentry nodes

ptrus:
Closes: #1829

TODO: 
- [x] some more TODO in comments (some extracting/generalizing to be done)


go/storage/api/api.go
```diff
@@ -322,6 +324,14 @@ type Backend interface {
 	Initialized() <-chan struct{}
 }
 
+// TODO: extract this out of storage API, into a more general 'PolicyBackend' interface.
```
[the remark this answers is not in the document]
ptrus: Thanks for the great suggestion and i think i now implemented most of this (so feel free to check again). I still have quite some TODOs in the code so feel free to ignore those, mostly interested in any high level comments. 

A recap of changes since last time:
- added `grpc/proxy` which implements a message agnostic proxy to forward the requests. It's used in `worker/sentry/grpc` so that it forwards all requsts to the upstream node
- added `grpc/auth` which implements server interceptors for authentication. Still need to ckean the existing Polcy implmentation a bit as mentioned in the TODOs

plan to address all the remaining TODOs tomorrow, but that shouldn't be blocking review :)

go/oasis-node/cmd/registry/node/node.go
```diff
@@ -352,7 +359,7 @@ func Register(parentCmd *cobra.Command) {
 func init() {
 	flags.String(CfgEntityID, "", "Entity ID that controls this node")
 	flags.Uint64(CfgExpiration, 0, "Epoch that the node registration should expire")
-	flags.StringSlice(CfgCommitteeAddress, nil, "Address(es) the node can be reached as a committee member")
+	flags.StringSlice(CfgCommitteeAddress, nil, "Address(es) the node can be reached as a committee member of the form [Certificate@]ip:port (where Certificate@ part is optional and represents base64 encoded node certificate for TLS)")
```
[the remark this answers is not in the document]
ptrus: yeah indeed. So for this exact command the annoyance can be a bit avoided by instead accepting the certificate path here. But yeah the general problem remains, and as you mentioned, we can only avoid the need for storing the certificates in the registry by rewriting the TLS authentication. Can open an issue for this in case you think we might want do to it.

.changelog/1829.feature.md
```diff
@@ -0,0 +1,3 @@
+Add gRPC Sentry Nodes support.
+
+TODO: description
```
ptrus: actually this should probably be `1829.breaking.md` due to the changes to the committe address in registry

go/common/grpc/policy/policy_test.go
```diff
@@ -151,20 +160,27 @@ func TestAccessPolicy(t *testing.T) {
...
+	grpcServer, err := cmnGrpc.NewServer(serverConfig)
 	require.NoErrorf(err, "Failed to create a new gRPC server: %v", err)
 
 	// Create a new pingServer with a new RuntimePolicyChecker.
-	policyChecker := NewDynamicRuntimePolicyChecker()
+	watcher := NewPolicyWatcher()
+	// Register the PolicyWatcherService.
+	api.RegisterService(grpcServer.Server(), watcher)
+
+	policyChecker := NewDynamicRuntimePolicyChecker("ping", watcher)
```
kostko: Shouldn't `"ping"` be `"PingService"`? Maybe change it to use `serviceDesc.ServiceName`?

go/common/grpc/policy/policy_test.go
```diff
@@ -202,21 +218,42 @@ func TestAccessPolicy(t *testing.T) {
...
+		require.Equal(api.ServicePolicies{Service: "ping", AccessPolicies: expectedPolicy}, p, "initial policy")
+	case <-time.After(recvTimeout):
+		t.Fatalf("Failed to receive value, initial policy")
+	}
 
 	_, err = client.Ping(ctx, &PingQuery{testNs})
 	require.EqualError(
 		err,
-		"rpc error: code = PermissionDenied desc = grpc: calling Ping method for runtime 06956b25ae9fabb8295ce6879f0995c7ad02f8b2a1b22cbade17960a70d765ea not allowed for client CN=oasis-node",
+		"rpc error: code = PermissionDenied desc = grpc: calling /PingServer/Ping method for runtime 06956b25ae9fabb8295ce6879f0995c7ad02f8b2a1b22cbade17960a70d765ea not allowed for client CN=oasis-node",
```
kostko: Hm, where does `PingServer` come from, shouldn't it be `PingService`?

go/common/grpc/proxy/proxy.go
```diff
@@ -0,0 +1,229 @@
...
+// Client -> Upstream
+func (p *proxy) proxyUpstream(downstream grpc.ServerStream, upstream grpc.ClientStream) <-chan error {
+	errCh := make(chan error, 1)
+
+	go func() {
+		for {
+			// XXX: since we are using CBOR we are able to unmarshal messages
+			// without knowing the schema. This wouldn't work with protobuf, and
+			// a raw binary codec should be used instead.
+			var m interface{}
```
kostko: You can probably use `cbor.RawMessage` instead of `interface{}` to avoid any deserialization from taking place.

go/common/grpc/proxy/proxy.go
```diff
@@ -0,0 +1,229 @@
...
+func (p *proxy) proxyDownstream(upstream grpc.ClientStream, downstream grpc.ServerStream) <-chan error {
+	errCh := make(chan error, 1)
+	var headerSent bool
+	go func() {
+		for {
+			// Wait for stream msg (from server).
+			// XXX: since we are using CBOR we are able to unmarshal messages
+			// without knowing the schema. This wouldn't work with protobuf, and
+			// a raw binary codec should be used instead.
+			var m interface{}
```
kostko: Same as above, `cbor.RawMessage` should work.

go/common/grpc/testing/ping.go
```diff
@@ -0,0 +1,207 @@
...
+// PingQuery is the PingServer query.
+type PingQuery struct {
+}
+
+// PingResponse is the response of the PingServer.
+type PingResponse struct {
+}
+
+// PingServer is a testing ping server interface.
+type PingServer interface {
```
kostko: Hm what's the relationship between this server/service/etc. and the one defined in the policy test? Could these be unified?

go/oasis-node/cmd/node/node.go
```diff
@@ -129,6 +130,7 @@ type Node struct {
 	TransactionSchedulerWorker *txnscheduler.Worker
 	MergeWorker                *merge.Worker
 	SentryWorker               *workerSentry.Worker
+	SentryGrpcWorker           *workerGrpcSentry.Worker
```
kostko: Hm could this be part of the regular sentry worker instead of having a separate one for gRPC? And then maybe have configuration flags like `sentry.consensus.enabled` and `sentry.grpc.enabled` for specific things?

go/oasis-node/cmd/registry/node/node.go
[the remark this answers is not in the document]
kostko: > Can open an issue for this in case you think we might want do to it sooner or later.

Yeah I think this could be useful to have, so file an issue and we'll see when/if we can do it.

go/worker/sentry/grpc/init.go
```diff
@@ -0,0 +1,152 @@
...
+	CfgUpstreamAddress = "worker.sentry.grpc.upstream.address"
+
+	// CfgUpstreamCert is the path to the certificate files of the upstream node.
+	CfgUpstreamCert = "worker.sentry.grpc.upstream.cert"
+)
+
+// Flags has the configuration flags.
+var Flags = flag.NewFlagSet("", flag.ContinueOnError)
+
+// TOOD: duplicated with: worker/registration/worker.go
```
kostko: Candidate for common gRPC?

go/worker/sentry/grpc/worker.go
```diff
@@ -0,0 +1,199 @@
...
+	identity *identity.Identity
+}
+
+type upstreamConn struct {
+	conn              *grpc.ClientConn
+	resolverCleanupCb func()
+}
+
+// TODO: this whole authFunction is currently hardcoded to only supports 'storage' backend,
+//       so that i was able to do test. Sentry GRPC worker should be upstream agnostic,
```
kostko: Yeah that would be ideal. What exactly needs to change here? So the most problematic thing seems to be extracting "namespace" from request? How could this be handled -- each service method would need to define how to extract the namespace.

* Either the upstream provides a "key path" in the CBOR document for each method (so the proxy would unmarshal into an `interface{}` and extract the path, then parse it as a namespace), or
* the proxy should be expected to know all services (registry of all services by name?) and have "namespace extractor" methods registered.

Maybe the second one is simpler as we can assume that the proxy also knows all services (as it is just another oasis-node). The first one would allow for simpler proxies though (they would only need to know how to decode and traverse CBOR).

go/worker/sentry/grpc/worker.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	defer pSub.Close()
+
+	// Watch Policies.
+	for {
+		select {
+		case p, ok := <-pCh:
+			if !ok {
+				g.logger.Error("WatchPolicies stream failure")
+				// TODO: Panic? Retry?
```
kostko: When can this happen?

go/worker/storage/committee/policy.go
```diff
@@ -10,29 +11,38 @@ import (
...
 	// sync-up.
 	storageNodesPolicy = &committee.AccessPolicy{
 		Actions: []accessctl.Action{
-			"GetDiff",
-			"GetCheckpoint",
+			accessctl.Action(api.MethodGetDiff.Full()),
+			accessctl.Action(api.MethodGetCheckpoint.Full()),
+		},
+	}
+	sentryNodesPolicy = &committee.AccessPolicy{
```
kostko: Is there a case where sentry nodes would not be allowed to proxy all methods?

.changelog/1829.feature.md
kostko: :+1:

go/worker/storage/committee/policy.go
ptrus: no, but this is exactly what this specifies (these are all the methods that use access policies)

go/common/grpc/testing/ping.go
ptrus: done

go/common/grpc/auth/auth_test.go
```diff
@@ -0,0 +1,143 @@
+package auth_test
```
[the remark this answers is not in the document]
ptrus: need a separate package to avoid circular dependency: this imports `common/grpc/testing` ->`common/grpc`->`common/grpc/auth`. Can move it in a test package under `common/grpc/auth/test` if that's nicer.

go/worker/sentry/grpc/worker.go
ptrus: done something in this direction:
> the proxy should be expected to know all services (registry of all services by name?) and have "namespace extractor" methods registered.

by creating a registry of method types and namespace extractors. please take another look

go/worker/sentry/grpc/worker.go
ptrus: i think connection errors with the upstream (for whatever reason) would make this fail

go/oasis-node/cmd/node/node.go
ptrus: yeah makes sense, will do

go/common/grpc/service.go
```diff
@@ -1,38 +1,132 @@
...
 
-import "fmt"
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"sync"
+
+	"github.com/fxamacker/cbor"
+	"github.com/oasislabs/oasis-core/go/common"
```
kostko: Separate group for internal imports.

go/common/grpc/service.go
```diff
@@ -1,38 +1,132 @@
...
+	if _, isRegistered := registeredMethods.Load(md.FullName()); isRegistered {
+		panic(fmt.Errorf("service: method already registered: %s", name))
+	}
+	registeredMethods.Store(md.FullName(), md)
+
+	return md
+}
+
+// NewMethodDesc creates a new method name for the given service.
+func (sn ServiceName) NewMethodDesc(name string, requestType interface{}) *MethodDesc {
```
kostko: Yeah maybe just call it `NewMethod` after you remove `NewMethodName`.

go/common/grpc/service.go
```diff
@@ -1,38 +1,132 @@
...
 // ServiceName is a gRPC service name.
 type ServiceName string
 
+// NamespaceExtractor extracts namespce from a method request.
+type NamespaceExtractor func(req interface{}) (common.Namespace, error)
+
+// ServiceNameFromMethod extract service name from method name.
+func ServiceNameFromMethod(methodName string) ServiceName {
+	substrs := strings.Split(methodName, "/")
+	// TODO: enforce no '/' in Service or Method names.
```
kostko: Good idea, this should be checked in constructors.

go/common/grpc/service.go
```diff
@@ -1,38 +1,132 @@
...
+	return md
+}
+
+// WithNamespacePolicies tells weather the enpoint does have per namespace
+// polices defined.
+func (m *MethodDesc) WithNamespacePolicies(f NamespaceExtractor) *MethodDesc {
+	m.accessControl = true
+	m.namespaceExtractor = f
+
+	registeredMethods.Store(m.short, m)
```
kostko: Hm this doesn't seem right? Shouldn't this be `m.full`?

Also are you sure you need to actually update the stored value? Since you are storing pointers, just updating `m` should be enough, assuming it is already registered?

go/common/grpc/service.go
```diff
@@ -1,38 +1,132 @@
...
 }
 
-// Full returns the full method name.
-func (m *MethodName) Full() string {
+// FullName returns the full method name.
+func (m *MethodDesc) FullName() string {
 	return m.full
 }
+
+// IsAccessControlled retruns if method is accessc controlled.
```
kostko:
```suggestion
// IsAccessControlled retruns if method is access controlled.
```

go/common/grpc/service.go
```diff
@@ -1,38 +1,132 @@
...
+
+	if _, isRegistered := registeredMethods.Load(md.FullName()); isRegistered {
+		panic(fmt.Errorf("service: method already registered: %s", name))
+	}
+	registeredMethods.Store(md.FullName(), md)
+
+	return md
+}
+
+// WithNamespacePolicies tells weather the enpoint does have per namespace
```
kostko:
```suggestion
// WithNamespacePolicies tells weather the endpoint does have per namespace
```

go/common/grpc/service.go
```diff
@@ -1,38 +1,132 @@
...
+	if err := cbor.Unmarshal(*req, v); err != nil {
+		return nil, fmt.Errorf("unmarshal error: %w", err)
+	}
+	return v, nil
+}
+
+// ExtractNamespace extracs the from the method request.
+func (m *MethodDesc) ExtractNamespace(req interface{}) (common.Namespace, error) {
+	if m.namespaceExtractor == nil {
+		return *new(common.Namespace), fmt.Errorf("method not namespaced")
```
kostko:
```suggestion
		return common.Namespace{}, fmt.Errorf("method not namespaced")
```

go/storage/api/grpc.go
```diff
@@ -3,80 +3,130 @@ package api
...
+	// MethodSyncGetPrefixes is the name of the SyncGetPrefixes method.
+	MethodSyncGetPrefixes = ServiceName.NewMethodDesc("SyncGetPrefixes", GetPrefixesRequest{})
+	// MethodSyncIterate is the name of the SyncIterate method.
+	MethodSyncIterate = ServiceName.NewMethodDesc("SyncIterate", IterateRequest{})
+	// MethodApply is the name of the Apply method.
+	MethodApply = ServiceName.NewMethodDesc("Apply", ApplyRequest{}).
+			WithNamespacePolicies(func(req interface{}) (common.Namespace, error) {
+			r, ok := req.(*ApplyRequest)
+			if !ok {
+				return *new(common.Namespace), errInvalidRequestType
```
kostko:
```suggestion
				return common.Namespace{}, errInvalidRequestType
```
(also elsewhere)

go/worker/sentry/grpc/worker.go
kostko: Cool, yeah this looks like a nice approach.

go/common/grpc/service.go
ptrus: yeah you are right for both :+1:

ptrus:
@kostko This should be ready for review, i do plan to also address these two things soon:
- cleanup commits into few logical commits
- update all the gRPC service definitions to also provide the type (and then remove the old `NewMethodName` and rename the new one)

go/oasis-node/cmd/registry/node/node.go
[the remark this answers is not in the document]
kostko: #2556

go/worker/storage/service_external.go
```diff
@@ -17,14 +26,20 @@ type storageService struct {
 	debugRejectUpdates bool
 }
 
-func (s *storageService) checkUpdateAllowed(ctx context.Context, method string, ns common.Namespace) error {
-	if s.debugRejectUpdates {
-		return errors.New("storage: rejecting update operations")
+func (s *storageService) AuthFunc(ctx context.Context, fullMethodName string, req interface{}) error {
```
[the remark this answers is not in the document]
ptrus: yeah the only specific thing is `grpcPolicy` policy checker. I think we could have a generic `AuthFunc` that accepts the policy checker as an arg (or the policy checker struct could define a generic auth func). Will check how that would look

go/worker/storage/service_external.go
```diff
@@ -58,59 +73,57 @@ func (s *storageService) SyncIterate(ctx context.Context, request *api.IterateRe
...
 
 func (s *storageService) Apply(ctx context.Context, request *api.ApplyRequest) ([]*api.Receipt, error) {
-	if err := s.checkUpdateAllowed(ctx, "Apply", request.Namespace); err != nil {
-		return nil, err
-	}
 	if err := s.ensureInitialized(ctx); err != nil {
 		return nil, err
 	}
+	if s.debugRejectUpdates {
+		return nil, errors.New("storage: rejecting update operations")
```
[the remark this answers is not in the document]
ptrus: :+1:

go/worker/sentry/grpc/worker.go
```diff
@@ -0,0 +1,234 @@
...
+func (g *Worker) worker() {
+	go func() {
+		defer close(g.quitCh)
+		defer (g.cancelCtx)()
+		<-g.grpc.Quit()
+	}()
+
+	g.policyWatcher = policyAPI.NewPolicyWatcherClient(g.conn)
+
+	// XXX: this could also be closed after getting the initial policy.
```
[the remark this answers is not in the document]
ptrus: i don't think anything waits for this worker initialization so i don't think ti matter, i'll just remove the comment

go/worker/sentry/grpc/worker.go
```diff
@@ -0,0 +1,234 @@
...
+		panic("failed to watch policies")
+	}
+	defer pSub.Close()
+
+	// Watch Policies.
+	for {
+		select {
+		case p, ok := <-pCh:
+			if !ok {
+				g.logger.Error("WatchPolicies stream failure")
```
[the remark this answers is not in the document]
ptrus: probably not here, but there should be some other log message indicating the reason (i assume the error returned by the `grpc/policy/api/handlerWatchPolicies` is logged somewhere by grpc)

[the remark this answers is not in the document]
ptrus: ok i tested it out, so this indeed happens when the connection to the upstream node fails. This can be seen in other logs, so probably no additional logging necessary here.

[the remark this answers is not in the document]
kostko: Ok then maybe just a comment saying that this happens when the connection to the upstream node fails. Also, should it automatically retry instead of just failing?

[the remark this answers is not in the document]
ptrus: Yeah i was thinking about this. Since in probably all cases the upstream node will be on a different host, it's not unreasonable to expect this failing from time to time. Currently what would happen is the sentry would shutdown (and probably in any reasonable deployment get automatically restarted), but we could also handle retrying here i guess.

go/worker/registration/worker.go
```diff
@@ -310,17 +310,98 @@ func (w *Worker) registerNode(epoch epochtime.EpochTime) error {
 		}
 	}
 
+	var err error
+	sentryAddrs := w.workerCommonCfg.SentryAddresses
+	// TODO: explicit flag for this.
```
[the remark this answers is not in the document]
ptrus: i was thinking about having a separate explicit config flag, but i think it's unnecessary

go/runtime/enclaverpc/api/api.go
```diff
@@ -3,8 +3,16 @@ package api
...
+	"fmt"
 
 	"github.com/oasislabs/oasis-core/go/common"
+	"github.com/oasislabs/oasis-core/go/common/cbor"
+)
+
+const (
+	// Make sure this always matches the appropriate method in
+	// `keymanager-runtime/src/methods.rs`.
+	GetPublicKeyRequestMethod = "get_public_key"
```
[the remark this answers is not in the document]
ptrus: any recommendations on how to handle this? are you fine with moving this (and the other methods) back to keymanager package, but still using it in `enclaverpc` service authentication definition?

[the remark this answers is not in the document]
kostko: No, the check is specific to the key manager enclave RPC endpoint, so it should not be part of generic EnclaveRPC. In this case the access control function should be defined here:
https://github.com/oasislabs/oasis-core/blob/ptrus/feature/grpc-sentry-support/go/worker/keymanager/service.go

What is the reason that the access control function needs to be defined in the generic gRPC service definition?

go/common/grpc/service.go
```diff
@@ -1,38 +1,132 @@
...
+// extractor defined.
+func (m *MethodDesc) WithNamespaceExtractor(f NamespaceExtractor) *MethodDesc {
+	m.namespaceExtractor = f
+	return m
 }
 
-// Short returns the short method name.
-func (m *MethodName) Short() string {
+// WithAccessControll tells weather the endpoint does have access controll.
+func (m *MethodDesc) WithAccessControll(f func(req interface{}) bool) *MethodDesc {
```
kostko:
```suggestion
func (m *MethodDesc) WithAccessControl(f func(req interface{}) bool) *MethodDesc {
```

go/runtime/enclaverpc/api/api.go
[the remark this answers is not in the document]
kostko: So another question -- how will this work in this example:
* If the worker is a key manager, it exports an EnclaveRPC service connecting to the key manager enclave on the external gRPC server. It uses access control policy A.
* If the worker is a foobar worker, it exports an EnclaveRPC service connecting to the foobar enclave on the external gRPC server. It uses access control policy B.

Each instance should have its own access control policy.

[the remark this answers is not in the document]
ptrus: > What is the reason that the access control function needs to be defined in the generic gRPC service definition?

so that the proxy (`sentry/grpc`) could apply the correct access control. Without it, we cannot really know from the request itself if policies should be applied or not.

> So another question -- how will this work in this example:

maybe we shouldn't have a "generic" enclaberpc service?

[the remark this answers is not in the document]
ptrus: i guess we could extend the `WatchPolicies` thing that the sentry uses, to also support sharing custom (non "grpc/policy")  policies. But i think this "other" type of policies should first be a bit generalized, so it can be reused between different services.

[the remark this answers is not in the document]
kostko: > maybe we shouldn't have a "generic" enclaverpc service?

So the idea is that you can call arbitrary methods via EnclaveRPC regardless of endpoints, e.g., that the API is the same so that you can use the same client.

I would be ok with having multiple server-side instances (even with different gRPC service names and different access control policies), but the methods should be exactly the same and should map to the `Transport` interface.

So the service name could be `EnclaveRPC.KeyManager` and the km-specific bits would be defined in the km package. It would be nice if we could avoid duplicating all the definitions as other than the access control policy they would be exactly the same.

Maybe in this case it would make sense to have a factory of EnclaveRPC gRPC services? So basically you could just wrap `serviceName`, `method*`, `serviceDesc` into a structure and have `handler*` be methods. Then you could have a constructor that creates a new service and you could pass access control policies?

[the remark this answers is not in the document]
ptrus: yeah, this approach makes sense, thanks :+1:

go/worker/storage/service_external.go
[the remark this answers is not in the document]
ptrus: done (see `grpc/policy.go#GRPCAuthenticationFunction`) it's now used by both keymanager and storage grpc services

go/runtime/enclaverpc/api/api.go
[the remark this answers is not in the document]
ptrus: done something like this ^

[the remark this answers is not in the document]
ptrus: done something like this ^

go/worker/common/committee/accessctl.go
```diff
@@ -21,6 +23,13 @@ func (ap AccessPolicy) AddRulesForCommittee(policy *accessctl.Policy, committee
 	}
 }
 
+func (ap AccessPolicy) AddCertPolicy(policy *accessctl.Policy, cert *x509.Certificate) {
```
kostko: Add documentation comment for public method.

go/runtime/enclaverpc/api/grpc.go
```diff
@@ -2,34 +2,33 @@ package api
...
-			{
-				MethodName: methodCallEnclave.Short(),
-				Handler:    handlerCallEnclave,
-			},
-		},
-		Streams: []grpc.StreamDesc{},
-	}
+	errInvalidRequestType = fmt.Errorf("invalid request type")
+
+	serviceNameBase = "EnclaveRPC"
```
kostko: This and the next one could be consts.

go/runtime/enclaverpc/api/grpc.go
```diff
@@ -2,34 +2,33 @@ package api
...
+
+	methodCallEnclaveName = "CallEnclave"
 )
 
-func handlerCallEnclave( // nolint: golint
+// EnclaveRPC is the enclave rpc gRPC service.
+type EnclaveRPC struct {
+	// ServiceName is the EnclaveRPC service name.
+	ServiceName cmnGrpc.ServiceName
+	// MethodCallEnclave is the EnclaveRPC CallEnclave method description.
```
kostko:
```suggestion
	// MethodCallEnclave is the EnclaveRPC CallEnclave method descriptor.
```

go/runtime/enclaverpc/api/grpc.go
```diff
@@ -2,34 +2,33 @@ package api
...
+	errInvalidRequestType = fmt.Errorf("invalid request type")
+
+	serviceNameBase = "EnclaveRPC"
+
+	methodCallEnclaveName = "CallEnclave"
 )
 
-func handlerCallEnclave( // nolint: golint
+// EnclaveRPC is the enclave rpc gRPC service.
+type EnclaveRPC struct {
```
kostko: Maybe call it just `Service` or `GRPCService`?

go/runtime/enclaverpc/api/grpc.go
```diff
@@ -53,23 +52,59 @@ func handlerCallEnclave( // nolint: golint
...
 
 // RegisterService registers a new EnclaveRPC transport service with the given gRPC server.
-func RegisterService(server *grpc.Server, service Transport) {
-	server.RegisterService(&serviceDesc, service)
+func (e *EnclaveRPC) RegisterService(server *grpc.Server, service Transport) {
+	server.RegisterService(&e.ServiceDesc, service)
+}
+
+// EnclaveRPC creates a new EnclaveRPC service.
+func New(serviceNamePrefix string, accessControl func(req interface{}) bool) *EnclaveRPC {
```
kostko: Maybe call it `NewService` or `NewGRPCService` as `New` seems to generic (there are other structs in this package).

go/common/grpc/proxy/proxy.go
```diff
@@ -0,0 +1,207 @@
...
+	"google.golang.org/grpc/status"
+
+	"github.com/oasislabs/oasis-core/go/common/cbor"
+	policy "github.com/oasislabs/oasis-core/go/common/grpc/policy/api"
+	"github.com/oasislabs/oasis-core/go/common/logging"
+)
+
+// Handler returns a grpc StreamHandler than can be used
+// to proxy requests to provided client.
+// XXX: potentially the connection should be established in this package,
```
kostko: Plan to file issues around these?

go/common/grpc/service.go
```diff
@@ -1,38 +1,132 @@
...
 
-// Short returns the short method name.
-func (m *MethodName) Short() string {
+// WithAccessControl tells weather the endpoint does have access controll.
+func (m *MethodDesc) WithAccessControl(f func(req interface{}) bool) *MethodDesc {
+	m.accessControl = f
+	return m
+}
+
+// MethodDesc is a gRPC method name.
```
kostko:
```suggestion
// MethodDesc is a gRPC method descriptor.
```

go/common/node/address.go
```diff
@@ -118,6 +125,55 @@ func (ca *ConsensusAddress) String() string {
 	return fmt.Sprintf("%s@%s", ca.ID, ca.Address)
 }
 
+// CommitteeAddress represents an Oasis committee address that includes a
+// server certificate and a TCP address.
+// NOTE: The address certificate can be different from the actual node
+// certificate to allow using a sentry node's addresses.
+// TODO: tests
```
kostko: I see this has been added.

go/keymanager/api/grpc.go
```diff
@@ -0,0 +1,63 @@
...
+
+		return skipPolicyCheck
+	}
+
+	// Service is the Keymanager enclave gRPC service.
+	Service = enclaverpc.New(ModuleName, requestSkipPolicyCheck)
+)
+
+// This is the Go analog of the Rust RPC Frame from runtime/src/rpc/types.go.
+type frame struct {
```
kostko: It may be useful to expose this publicly in `api.go` (e.g., for other consumers that may want to peek into frames).

go/common/grpc/proxy/proxy.go
ptrus: yeah i'll file a single issue about potential proxy benchmarks and possible improvements, since i feel like for most of these some benchmarks should be first done (e.g. compare perf of accessing a storage node directly or via a sentry)

go/common/grpc/policy/api/grpc.go
```diff
@@ -0,0 +1,115 @@
...
+
+func handlerWatchPolicies(srv interface{}, stream grpc.ServerStream) error {
+	if err := stream.RecvMsg(nil); err != nil {
+		return err
+	}
+
+	ctx := stream.Context()
+	grpcSrv, ok := srv.(PolicyWatcherClient)
+	if !ok {
+		return fmt.Errorf("Invalid backend")
```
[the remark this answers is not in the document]
ptrus: yeah it can't, removed the check

.changelog/1829.breaking.md
```diff
@@ -0,0 +1,18 @@
...
+
+Added flags:
+
+- `--worker.sentry.grpc.enabled` - enables the gRPC proxy (requires 
+`--worker.sentry.enabled` flag)
+- `--worker.sentry.grpc.client.port` - port on which gRPC proxy is accessible.
+- `--worker.sentry.grpc.client.address` - addresses on which gRPC proxy is
+accessible (needed so protected nodes can query sentries for its addresses)
+- `-worker.sentry.grpc.upstream.address` - address of the protected node
+- `worker.sentry.grpc.upstream.cert` - public ceritficate of the upstream grpc
```
kostko:
```suggestion
- `worker.sentry.grpc.upstream.cert` - public certificate of the upstream gRPC
```

.changelog/1829.breaking.md
```diff
@@ -0,0 +1,18 @@
+Add gRPC Sentry Nodes support.
+
+This adds gRPC proxying and policy enforcment support to existing sentry nodes, 
+which enables protecting upstream nodes gRPC endpoints.
+
+Added flags:
```
kostko:
```suggestion
Added/changed flags:
```

.changelog/1829.breaking.md
```diff
@@ -0,0 +1,18 @@
+Add gRPC Sentry Nodes support.
+
+This adds gRPC proxying and policy enforcment support to existing sentry nodes, 
+which enables protecting upstream nodes gRPC endpoints.
```
kostko:
```suggestion
which enables protecting upstream nodes' gRPC endpoints.
```

go/keymanager/api/grpc.go
```diff
@@ -0,0 +1,64 @@
...
+		return skipPolicyCheck
+	}
+
+	// Service is the Keymanager enclave gRPC service.
+	Service = enclaverpc.NewService(ModuleName, requestSkipPolicyCheck)
+)
+
+// Frame is the Go analog of the Rust RPC Frame defined in
+// client/src/rpc/client.rs.
+type Frame struct {
```
kostko: Move this to `api.go` as it is not gRPC-specific.

go/oasis-test-runner/oasis/args.go
```diff
@@ -158,24 +159,47 @@ func (args *argBuilder) workerClientPort(port uint16) *argBuilder {
 	return args
 }
 
-func (args *argBuilder) workerRegistrySentryAddresses(addrs []string) *argBuilder {
+func (args *argBuilder) disableRegistrationWorker() *argBuilder {
+	args.vec = append(args.vec, []string{
+		"--" + registration.CfgRegistrationWorkerEnabled + "=false",
```
kostko: Is this a new flag? It is not mentioned in the changelog fragment. Also, is this required because of #1930 (e.g., it could be handled automatically)?

go/worker/common/config.go
```diff
@@ -17,7 +19,14 @@ var (
 	// CfgClientPort configures the worker client port.
 	CfgClientPort = "worker.client.port"
 
-	cfgClientAddresses = "worker.client.addresses"
+	CfgClientAddresses = "worker.client.addresses"
```
kostko: Add documentation comment.

go/oasis-test-runner/oasis/args.go
ptrus: Oh forgot to add it to the changelog fragment. So it's not technically needed actually, i just wanted to disable the registration worker on sentry nodes as it's not needed (in case it's enabled the registrations are failing anyway, so no actual difference). So i'm fine with removing the flag and postponing this to the #1930  - i don't see a reason it couldn't be done automatically (maybe by checking node roles?).

ptrus: i removed it for now as it's not technically needed, and it can be properly fixed with #1930

ptrus:
thanks for all the reviews on the huge PR @kostko  :-)

PR #2484, opened 2019-12-17
Expose consensus blocks and scheduler API via internal gRPC

### LINKED ISSUE #2482: Add consensus block and transaction metadata accessors
In order to enable people to build "network explorers" we should expose some additional methods via the consensus API endpoint, specifically:
* Consensus block metadata.
* Access to raw consensus transactions within a block.
* Stream of consensus blocks as they are finalized.

kostko:
Fixes #2482 

TODO
* [x] See if we need to add any other accessors.
* [x] Add tests.

### DIFF
--- a/go/common/persistent/persistent.go
+++ b/go/common/persistent/persistent.go
@@ -11,7 +11,7 @@ import (
 	"github.com/dgraph-io/badger/v2"
 	"github.com/dgraph-io/badger/v2/options"
 
-	ekbadger "github.com/oasislabs/oasis-core/go/common/badger"
+	cmnBadger "github.com/oasislabs/oasis-core/go/common/badger"
 	"github.com/oasislabs/oasis-core/go/common/cbor"
 	"github.com/oasislabs/oasis-core/go/common/logging"
 )
@@ -24,7 +24,7 @@ var ErrNotFound = errors.New("persistent: key not found in database")
 // CommonStore is the interface to the common storage for the node.
 type CommonStore struct {
 	db *badger.DB
-	gc *ekbadger.GCWorker
+	gc *cmnBadger.GCWorker
 }
 
 // Close closes the database handle.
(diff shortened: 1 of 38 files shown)


go/common/persistent/persistent.go
```diff
@@ -50,6 +50,8 @@ func NewCommonStore(dataDir string) (*CommonStore, error) {
 	opts = opts.WithLogger(ekbadger.NewLogAdapter(logger))
```
Yawning: Not changing the `ekbadger` import here?

go/scheduler/tests/tester.go
```diff
@@ -23,8 +23,8 @@ const recvTimeout = 5 * time.Second
 
 // SchedulerImplementationTests exercises the basic functionality of a
 // scheduler backend.
-func SchedulerImplementationTests(t *testing.T, backend api.Backend, consensus consensusAPI.Backend) {
-	seed := []byte("SchedulerImplementationTests")
+func SchedulerImplementationTests(t *testing.T, name string, backend api.Backend, consensus consensusAPI.Backend) {
```
Yawning: The name argument appears to always be the empty string, unless I'm mistaken.

kostko: You are mistaken, it can also be the string `"client"` for the client backend (see `node_test.go`). It is required as there is no way to deregister runtimes so each backend test needs its own seed.

go/common/persistent/persistent.go
kostko: Oops, missed that somehow. Fixed now, thanks!

go/consensus/tests/tester.go
```diff
@@ -0,0 +1,47 @@
+// Package tests is a collection of scheduler implementation test cases.
```
ptrus: `scheduler` -> `consensus`

go/consensus/tests/tester.go
```diff
@@ -0,0 +1,47 @@
...
+	require.True(blk.Height > 0, "block height should be greater than zero")
+
+	_, err = backend.GetTransactions(ctx, consensus.HeightLatest)
+	require.NoError(err, "GetTransactions")
+
+	blockCh, blockSub, err := backend.WatchBlocks(ctx)
+	require.NoError(err, "WatchBlocks")
+	defer blockSub.Close()
+
+	// Wait for a few blocks.
```
ptrus: maybe `watch/test few blocks`, and move the `3` into a constant :-)

PR #2489, opened 2019-12-18
build: Make `go generate optional

Yawning:
Fixes 2473.

 * [x] Build/CI related changes to make the `go generate` step optional.
 * [x] Include the compiled Intel aesmd protobuf file in git.
 * [x] Update the documentation.


Yawning:
> Could you also make the protobuf-compiler package in Fedora/Ubuntu package lists in the README optional?

We still have protobuf dependencies in the rust side, not sure if we need the compiler, don't want to uninstall it to check.

Issue #2435, opened 2019-11-27
oasis-node cmd for wiping the state


matevz:
Current upgrade process for the node operator requires downloading a new genesis file, wiping the state (but keeping the entity) and run the new oasis node. I suggest we add `oasis-node clean` command or `--rebuild` flag or something similar which effectively does (taken from operator docs):
```bash
$ rm -rf /serverdir/node/tendermint
$ rm -rf /serverdir/node/bleve-tag-index.bleve.db
$ rm /serverdir/node/abci-mux-state.bolt.db
$ rm /serverdir/node/persistent-store.db
```

And the next time you run `oasis-node` it re-downloads and verifies all blocks.

It also seems convenient for node operator support to just say to them "try running oasis-node clean" instead of pointing them to all the rm commands in docs.

PR #2490, opened 2019-12-18
go/oasis-node: Add the `unsafe-reset` sub-command

Yawning:
Fixes: #2435

As it probably can be deduced from the name, this sub-command will
reset the node back to a freshly provisioned state (preserving any
key material if it exists).

The `-n` (`--dry_run`) flag will go through the process without actually
removing state, so that it is possible to preview the results of
executing the command.

Usage: `oasis-node unsafe-reset --datadir=/path/to/datadir`

### DIFF
--- a/go/consensus/tendermint/tendermint.go
+++ b/go/consensus/tendermint/tendermint.go
@@ -94,6 +94,10 @@ const (
 
 	// CfgConsensusMinGasPrice configures the minimum gas price for this validator.
 	CfgConsensusMinGasPrice = "consensus.tendermint.min_gas_price"
+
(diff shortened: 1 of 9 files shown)


ptrus:
Nice, this will be really useful :+1: 

we could use (and therefore at least partially test) the command in the dump-restore e2e test: https://github.com/oasislabs/oasis-core/blob/master/go/oasis-test-runner/scenario/e2e/dump_restore.go#L71

Yawning:
> we could use and therefore (at least partially) test the command in the dump-restore e2e test

It would need options to preserve certain things it currently will obliterate (worker local storage for the KM, the mkvs store for storage).

PR #2494, opened 2019-12-20
Make storage per-runtime

### LINKED ISSUE #2474: Make storage database per-runtime
Currently a single storage database stores data for multiple runtimes which requires namespacing all nodes. After #2463 which introduces a per-runtime data directory we should consider making storage per-runtime as well and remove namespacing within the database.

kostko:
Fixes #2474 

* First commit removes the global node storage backend and instead replaces it with per-runtime storage backends. To support answering queries for multiple runtimes via gRPC a runtime registry-based storage router is added.

This PR changes database file organization within the node data directory and storage database internals and as such it is BREAKING WITHOUT MIGRATION.

TODO
* [x] Remove namespacing within a single storage database, add a namespace argument to storage backends for cross-checks.
* [x] Client storage backend should automatically track the correct runtime. Also consider removing support for tracking multiple runtimes which increases complexity and should now no longer be needed (as there is a storage instance per runtime).
* [x] Make untrusted local storage per runtime.


go/oasis-node/cmd/node/unsafe_reset.go
```diff
@@ -34,15 +34,17 @@ var (
...
 	}
 
+	runtimesGlob = filepath.Join(runtimeRegistry.RuntimesDir, "*")
+
 	nodeStateGlobs = []string{
 		"abci-mux-state.*.db",
 		"persistent-store.*.db",
 		tendermint.StateDir,
-		runtimeRegistry.RuntimesDir,
+		filepath.Join(runtimesGlob, "history.db"),
```
ptrus: maybe should reference `runtime/history/.DbFilename` (currently it's not exported though), instead of hardcoding `"history.db"`

kostko: Good idea, done.

go/roothash/tests/tester.go
```diff
@@ -215,17 +219,28 @@ func (s *runtimeState) testEpochTransitionBlock(t *testing.T, scheduler schedule
 	}
 }
 
-func testSucessfulRound(t *testing.T, backend api.Backend, consensus consensusAPI.Backend, storage storage.Backend, states []*runtimeState) {
+func testSucessfulRound(t *testing.T, backend api.Backend, consensus consensusAPI.Backend, identity *identity.Identity, states []*runtimeState) {
 	for _, state := range states {
-		state.testSuccessfulRound(t, backend, consensus, storage)
+		state.testSuccessfulRound(t, backend, consensus, identity)
```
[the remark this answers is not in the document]
kostko: Omg :grin:

kostko:
> but I want to know the background in #2474

As mentioned on Slack (and in the changelog fragment), the main point for doing this is to simplify the storage database schema, so per-key namespacing is no longer required. It also makes it simpler to remove or copy data for a specific runtime.

Yawning:
Do we want to reduce the storage cache sizes?

kostko:
Good idea, I'll make it configurable.

PR #2496, opened 2019-12-20
Add CHANGELOG.md and use towncrier for automatic assembly

### LINKED ISSUE #759: Automatically generate the changelog
We should consider automatically generating the changelog.

tjanez:
Closes #759.

### DIFF
--- a/.changelog/README.md
+++ b/.changelog/README.md
@@ -1,0 +1,93 @@
+# Change Log fragments
+
+This directory collects Change Log fragments:
+short files that each contain a snippet of MarkDown formatted text that will be
+assembled using [towncrier] to form the [Change Log] section for the next
+release.
+
+## Description
+
+A Change Log fragment should be a description of aspects of the change (if any)
+that are relevant to users.
+
+_NOTE: This could be very different from the commit message and pull request
+description, which are a description of the change as relevant to the people
+working on the code itself._
+
+The description could use one of the following two formats:
+
+- One line change summary followed by an empty line and a more detailed
+  explanation in the body.
+
+  For example:
+
+  ```text
+  Remove staking-related roothash messages.
+
+  There is no longer a plan to support direct manipulation of the staking accounts
+  from the runtimes in order to isolate the runtimes from corrupting the
+  consensus layer.
+
+  To reduce complexity, the staking-related roothash messages were removed. The
+  general roothash message mechanism stayed as-is since it may be useful in the
+  future, but any commits with non-empty messages are rejected for now.
+  ```
+
+- Shorter multi-line change description.
+
+  For example:
+
+  ```text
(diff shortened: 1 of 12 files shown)


.changelog/README.md
```diff
@@ -0,0 +1,93 @@
...
+```bash
+pip3 install https://github.com/oasislabs/towncrier/archive/oasis-master.tar.gz
+```
+
+_NOTE: [towncrier] requires Python 3.5+._
+
+And then run:
+
+```bash
+towncrier build --version <NEXT-VERSION> --draft
```
kostko: What about when you want to do an actual release?

CHANGELOG.md
```diff
@@ -0,0 +1,64 @@
...
+- Reduce Badger in-memory cache sizes.
+  ([#2484](https://github.com/oasislabs/oasis-core/issues/2484))
+
+  The default is 1 GiB per badger instance and we use a few instances so this
+  resulted in some nice memory usage.
+
+
+## 19.0 (2019-12-18)
+
+- Initial release.
```
pro-wh: can we have this not say 19.0 (2019-12-18) is the initial release? it's not the initial release. if we don't want to go back and retroactively describe what changed between a9a532e6be67512b3e4e581b5a57a97dff5715ca and 19.0, it's less confusing not to have this section at all

.changelog/README.md
tjanez: You would leave out the `--draft` argument and towncrier would stage the changes, i.e. remove the Change Log fragments and add a new section to the `CHANGELOG.md`.
After verifying that everything looks ok, you would just commit the changes.

I plan to create `changelog` Make target that would call towncrier underneath to make things even simpler.

CHANGELOG.md
tjanez: Agreed, this is not ideal. I would still like to have this section so that readers will know what the Change Log section for the next release, i.e. 19.1, refers to.

I've changed the contents to actually describe a process change, i.e. that the project started using a new versioning and release process and adopted a [CalVer](http://calver.org) (calendar versioning) scheme. The latter also explains why this release has version 19.0 as opposed to 1.0 or similar.

Please, take another look.

tjanez:
> The lint on GitHub Actions seems to be failing though, seems like some issue with the remote branch not being fully fetched? Probably an explicit `git fetch origin <BASE-BRANCH>` is needed because the linter tries to inspect what changed from the base branch.

Yes, indeed the [GitHub Checkout action](https://github.com/actions/checkout/) only fetches the pull request branch by default. Added a command that manually fetches the `master` branch.

CHANGELOG.md
pro-wh: thx for the changes

PR #2498, opened 2019-12-21
go staking: undisable transfers for some senders

pro-wh:
ostensibly for faucet purposes while we run the rest of the network with transfers disabled, this lets us identify a whitelist of accounts from which we allow transfers when otherwise transfers are disabled.

cc #2481


pro-wh:
gonna add changelog and merge. tests coming in a separate PR

PR #2500, opened 2019-12-23
per entity signature count tool

ptrus:
Implements a cli tool to query per entity signature counts.

Could potentially extract this into it's own repo.


.changelog/2500.feature.md
```diff
@@ -0,0 +1 @@
+Entity block signatures count tool.
```
kostko: Maybe add a short paragraph explaining what the tool does.

go/extra/stats/.gitignore
```diff
@@ -0,0 +1 @@
+extra/stats
```
kostko: Change parent makefile to use `basename` on the finale binary filename instead of creating additional subdirectories?

go/extra/stats/cmd/stats.go
```diff
@@ -0,0 +1,343 @@
...
+			"err", err,
+		)
+		os.Exit(1)
+	}
+
+	client := consensusAPI.NewConsensusClient(conn)
+	return conn, client
+}
+
+func registryClient(cmd *cobra.Command) (*grpc.ClientConn, registryAPI.Backend) {
```
kostko: Both service clients can share the same gRPC connection.

kostko:
> Could potentially extract this into it's own repo.

I think it is useful having this here for now.

ravenac95:
This looks good to me! My approval is meaningless here tho :).

Issue #2459, opened 2019-12-09
Compute runtime's key manager should be checked


kostko:
Currently no stateful checks are performed on the registered compute runtime's key manager ID (the only check is that the runtime itself is not also the key manager). Checks are however performed for the genesis document state.

Checks should be added to the `RegisterRuntime` handler that in case a key manager runtime ID is specified, it must exist and be a valid key manager runtime.

~We should also allow explicitly specifying a NULL key manager (e.g., for non-confidential runtimes).~ (Already part of #2340.)

PR #2501, opened 2019-12-23
add key manager ID check

matevz:
PR for #2459:
- adds a check whether key manager ID of the newly registered runtime points to a valid key manager in the registry.

### DIFF
--- a/go/consensus/tendermint/apps/registry/transactions.go
+++ b/go/consensus/tendermint/apps/registry/transactions.go
@@ -402,6 +402,12 @@ func (app *registryApplication) registerRuntime(
 		return err
 	}
 
+	if rt.Kind == registry.KindCompute {
+		if err = registry.VerifyRegisterComputeRuntimeArgs(app.logger, rt, state); err != nil {
+			return err
+		}
+	}
+
 	if ctx.IsCheckOnly() {
 		return nil
 	}
--- a/go/registry/api/api.go
+++ b/go/registry/api/api.go
@@ -281,6 +281,14 @@ type NodeLookup interface {
 	NodeByCertificate(cert []byte) (*node.Node, error)
 }
 
+// RuntimeLookup interface implements various ways for the verification
+// functions to look-up runtimes in the registry's state.
+type RuntimeLookup interface {
+	// Runtime looks up a runtime by its identifier.
+	// It returns either the runtime with given id or error.
+	Runtime(id common.Namespace) (*Runtime, error)
+}
+
 // VerifyRegisterEntityArgs verifies arguments for RegisterEntity.
 func VerifyRegisterEntityArgs(logger *logging.Logger, sigEnt *entity.SignedEntity, isGenesis bool) (*entity.Entity, error) {
 	var ent entity.Entity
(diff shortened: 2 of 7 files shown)


.changelog/2459.feature.md
```diff
@@ -0,0 +1,2 @@
+When registering a new runtime, check whether the given key manager ID points
```
kostko: This is actually `breaking` as the same transaction can be processed differently by different versions.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -402,6 +402,23 @@ func (app *registryApplication) registerRuntime(
...
 	}
 
+	if rt.Kind == registry.KindCompute {
+		// Find key manager runtime, if it is set.
+		var km *registry.Runtime
+		if rt.KeyManager != nil {
+			km, err = state.ImmutableState.Runtime(*rt.KeyManager)
+		}
+
+		if err != nil {
```
kostko: This check should be in the upper if, e.g.:
```go
if km, err = state...; err != nil { ...
```

go/registry/api/api.go
```diff
@@ -930,6 +930,34 @@ func VerifyRegisterRuntimeArgs(logger *logging.Logger, sigRt *SignedRuntime, isG
 	return &rt, nil
 }
 
+// VerifyRegisterComputeRuntimeArgs verifies ComputeRuntime-specific arguments for RegisterRuntime.
+func VerifyRegisterComputeRuntimeArgs(logger *logging.Logger, rt *Runtime, km *Runtime) error {
+	// Check runtime's key manager, if key manager ID is set.
+	if rt.KeyManager != nil {
+		// key manager runtime must exist.
```
kostko:
```suggestion
		// Key manager runtime must exist.
```

go/registry/api/api.go
```diff
@@ -930,6 +930,34 @@ func VerifyRegisterRuntimeArgs(logger *logging.Logger, sigRt *SignedRuntime, isG
...
+		if km == nil {
```
kostko: This seems awkward, as mentioned in my previous review, it would be better to handle this via a `RuntimeLookup` interface, similar to the [existing `NodeLookup` interface](https://github.com/oasislabs/oasis-core/blob/master/go/registry/api/api.go#L274-L282). I will also need this interface in #2504.

So instead of passing `km`, pass the `RuntimeLookup` interface and call `runtimeLookup.Runtime(*rt.KeyManager)` here.

matevz: I added `RuntimeLookup` with only single method `Runtime()` now. This method was already implemented before by `ImmutableState`.

go/registry/api/api.go
```diff
@@ -930,6 +937,43 @@ func VerifyRegisterRuntimeArgs(logger *logging.Logger, sigRt *SignedRuntime, isG
...
+		if err != nil {
+			logger.Error("RegisterRuntime: error when fetching the runtime's key manager from registry",
+				"runtime", rt.ID,
+				"key_manager", rt.KeyManager,
+			)
+			return err
+		}
+
+		// Key manager runtime must exist.
+		if km == nil {
```
kostko: I think that `Runtime()` should return `ErrNoSuchRuntime` if the runtime doesn't exist, so this additional check is not needed?

go/scheduler/tests/tester.go
```diff
@@ -31,7 +31,7 @@ func SchedulerImplementationTests(t *testing.T, name string, backend api.Backend
 	rt, err := registryTests.NewTestRuntime(seed, nil)
 	require.NoError(err, "NewTestRuntime")
 
-	// Populate the registry with an entity and nodes.
+	// Populate the registryAPI with an entity and nodes.
```
kostko: Too much search and replace?

go/scheduler/tests/tester.go
```diff
@@ -126,7 +126,7 @@ func SchedulerImplementationTests(t *testing.T, name string, backend api.Backend
 
 	ensureValidCommittees(3, 1, int(rt.Runtime.TxnScheduler.GroupSize))
 
-	// Cleanup the registry.
+	// Cleanup the registryAPI.
```
kostko: Too much search and replace?

go/registry/api/api.go
```diff
@@ -930,6 +938,34 @@ func VerifyRegisterRuntimeArgs(logger *logging.Logger, sigRt *SignedRuntime, isG
 	return &rt, nil
 }
 
+// VerifyRegisterComputeRuntimeArgs verifies ComputeRuntime-specific arguments for RegisterRuntime.
```
kostko:
```suggestion
// VerifyRegisterComputeRuntimeArgs verifies compute runtime-specific arguments for RegisterRuntime.
```

go/scheduler/tests/tester.go
```diff
@@ -14,7 +14,7 @@ import (
 	consensusAPI "github.com/oasislabs/oasis-core/go/consensus/api"
 	epochtime "github.com/oasislabs/oasis-core/go/epochtime/api"
 	epochtimeTests "github.com/oasislabs/oasis-core/go/epochtime/tests"
-	registry "github.com/oasislabs/oasis-core/go/registry/api"
+	registryAPI "github.com/oasislabs/oasis-core/go/registry/api"
```
kostko: Why this change? There doesn't seem to be any other changes in this file?

PR #2502, opened 2019-12-23
Use CheckTx in transaction scheduler

### LINKED ISSUE #1963: Transaction scheduler should use CheckTx
As of #1555 the runtime exposes a CheckTx API for verifying incoming transactions but this is currently unused. The transaction scheduler should use the API to reject incoming transactions early instead of submitting them through the processing pipeline.

The runtime must ensure that any such checks are cheap enough that they can be performed quickly and not become a bottleneck of the whole pipeline.

*Estimated cost: 1 sprint*

abukosek:
Closes #1963.

TODO:
- [X] Make txn sched workers run runtimes.
- [X] Run CheckTx on transactions before queuing them.
- [X] Add a cmdline flag to toggle this check.
- [X] Figure out how to interpret CheckTx response.
- [X] Figure out why it hangs in e2e tests that use encrypted storage.
- [X] Figure out why it hangs in SGX e2e tests.
- [X] Write changelog fragment.

### DIFF
--- a/go/worker/txnscheduler/committee/node.go
+++ b/go/worker/txnscheduler/committee/node.go
@@ -20,7 +20,9 @@ import (
 	"github.com/oasislabs/oasis-core/go/roothash/api/block"
 	"github.com/oasislabs/oasis-core/go/runtime/transaction"
 	storage "github.com/oasislabs/oasis-core/go/storage/api"
+	commonWorker "github.com/oasislabs/oasis-core/go/worker/common"
 	"github.com/oasislabs/oasis-core/go/worker/common/committee"
+	"github.com/oasislabs/oasis-core/go/worker/common/host"
 	"github.com/oasislabs/oasis-core/go/worker/common/p2p"
 	computeCommittee "github.com/oasislabs/oasis-core/go/worker/compute/committee"
 	txnSchedulerAlgorithm "github.com/oasislabs/oasis-core/go/worker/txnscheduler/algorithm"
@@ -50,6 +52,8 @@ var (
 
 // Node is a committee node.
 type Node struct {
+	*commonWorker.RuntimeHostNode
+
 	commonNode  *committee.Node
 	computeNode *computeCommittee.Node
 
@@ -385,6 +389,15 @@ func (n *Node) worker() {
 
 	n.logger.Info("starting committee node")
 
+	// Initialize worker host for the new runtime.
+	if err := n.InitializeRuntimeWorkerHost(n.ctx); err != nil {
+		n.logger.Error("failed to initialize worker host",
+			"err", err,
+		)
+		return
+	}
+	defer n.StopRuntimeWorkerHost()
+
 	// Initialize transaction scheduler's algorithm.
 	runtime, err := n.commonNode.Runtime.RegistryDescriptor(n.ctx)
 	if err != nil {
@@ -437,6 +450,7 @@ func (n *Node) worker() {
 func NewNode(
 	commonNode *committee.Node,
 	computeNode *computeCommittee.Node,
+	workerHostFactory host.Factory,
 ) (*Node, error) {
 	metricsOnce.Do(func() {
 		prometheus.MustRegister(nodeCollectors...)
@@ -445,6 +459,7 @@ func NewNode(
 	ctx, cancel := context.WithCancel(context.Background())
 
 	n := &Node{
+		RuntimeHostNode:  commonWorker.NewRuntimeHostNode(commonNode, workerHostFactory),
 		commonNode:       commonNode,
 		computeNode:      computeNode,
 		ctx:              ctx,
--- a/go/worker/txnscheduler/committee/node.go
+++ b/go/worker/txnscheduler/committee/node.go
@@ -10,6 +10,7 @@ import (
 	"github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
 
+	"github.com/oasislabs/oasis-core/go/common/cbor"
 	"github.com/oasislabs/oasis-core/go/common/crash"
 	"github.com/oasislabs/oasis-core/go/common/crypto/hash"
 	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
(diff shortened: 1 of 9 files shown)


go/worker/txnscheduler/worker.go
```diff
@@ -159,36 +173,107 @@ func newWorker(
...
+					}
+
+					workerHost := workerRT.GetWorkerHost()
+					if workerHost == nil {
+						w.logger.Debug("runtime has shut down",
+							"runtime", rt.ID,
+						)
+						continue
+					}
+					if rt.Capabilities.TEE, grr = workerHost.WaitForCapabilityTEE(w.ctx); grr != nil {
```
kostko: I found the reason for why SGX tests are not working in this PR. So all batches are being rejected due to the following error:
> roothash/commitment: batch RAK signature invalid

This is because here you actually overwrite the RAK that is registered for the compute runtime (by the compute worker's `RegisterRole` handler). And since each enclave has its own RAK this makes all signatures for the compute enclave invalid.

The solution is either to:
* Support multiple enclaves to be registered for the same runtime in the node descriptor (e.g., multiple RAKs), for different purposes (e.g., one for compute and another one for txnscheduler).
* Remove this for loop so that only the enclave for compute is registered. There is no need to register the txnscheduler's RAK in the registry as this runtime is only used for local checks and it doesn't actually sign any batches.

I would be in favor of option two as it is much simpler and the txnscheduler runtime is only used for CheckTx, which currently doesn't sign the results using RAK. If we later desired to verify that the scheduler actually performed CheckTx, we could do add this later.

abukosek: Ah, that makes sense!  I'll fix it.  Thanks for the help!

go/worker/txnscheduler/service.go
```diff
@@ -15,6 +19,55 @@ func (w *Worker) SubmitTx(ctx context.Context, rq *api.SubmitTxRequest) (*api.Su
 		return nil, api.ErrUnknownRuntime
 	}
 
+	if w.checkTxEnabled {
```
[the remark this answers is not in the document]
abukosek: Done, filed as #2548.

go/worker/txnscheduler/service.go
```diff
@@ -15,6 +19,55 @@ func (w *Worker) SubmitTx(ctx context.Context, rq *api.SubmitTxRequest) (*api.Su
...
+		cbor.MustUnmarshal(resultRaw, &result)
+		if result.Error != nil {
+			w.logger.Error("worker CheckTx failed with error",
+				"err", result.Error,
+			)
+			return nil, fmt.Errorf("%w: %s", api.ErrCheckTxFailed, *result.Error)
+		}
+
+		w.logger.Debug("worker CheckTx successful, queuing transaction")
+	}
```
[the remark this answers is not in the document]
abukosek: Done :)

go/worker/common/runtime_host.go
```diff
@@ -268,6 +269,14 @@ func (n *RuntimeHostNode) GetWorkerHostLocked() host.Host {
 	return n.workerHost
 }
 
+// GetCurrentBlock returns the current roothash block from the underlying common node.
+func (n *RuntimeHostNode) GetCurrentBlock() roothash.Block {
+	n.commonNode.CrossNode.Lock()
+	defer n.commonNode.CrossNode.Unlock()
+
+	return *n.commonNode.CurrentBlock
```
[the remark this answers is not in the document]
abukosek: The way it was used previously, this couldn't happen, I think, but I've now removed this code, since it's no longer needed, as CheckTx is done in the node code.

go/worker/txnscheduler/worker.go
```diff
@@ -159,36 +173,62 @@ func newWorker(
...
-		initCh:       make(chan struct{}),
-		logger:       logging.GetLogger("worker/txnscheduler"),
+		enabled:        enabled,
+		checkTxEnabled: checkTxEnabled,
+		commonWorker:   commonWorker,
+		registration:   registration,
+		compute:        compute,
+		runtimes:       make(map[common.Namespace]*committee.Node),
+		ctx:            ctx,
+		cancelCtx:      cancelCtx,
```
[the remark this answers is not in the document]
abukosek: Oops, leftover from before, I've removed it.

go/worker/txnscheduler/committee/node.go
```diff
@@ -128,6 +136,59 @@ func (n *Node) HandlePeerMessage(ctx context.Context, message *p2p.Message) (boo
 	return false, nil
 }
 
+// CheckTx checks the given call in the node's runtime.
+func (n *Node) CheckTx(ctx context.Context, call []byte) error {
+	if n.commonNode.CurrentBlock == nil {
```
kostko: Accessing the `CurrentBlock` requires holding the CrossNode lock if called outside of methods which assume the caller does so (e.g., the ones ending in `Locked`).

go/worker/txnscheduler/committee/node.go
```diff
@@ -128,6 +136,59 @@ func (n *Node) HandlePeerMessage(ctx context.Context, message *p2p.Message) (boo
...
+		n.logger.Error("worker host CheckTx response doesn't contain exactly one result",
+			"num_results", len(resp.WorkerCheckTxBatchResponse.Results),
+		)
+		return api.ErrCheckTxFailed
+	}
+
+	// Interpret CheckTx result.
+	resultRaw := resp.WorkerCheckTxBatchResponse.Results[0]
+	var result transaction.TxnOutput
+	cbor.MustUnmarshal(resultRaw, &result)
```
kostko: Why panic on unmarshal? It should log the error and return `ErrCheckTxFailed`.

go/worker/txnscheduler/committee/node.go
```diff
@@ -136,6 +197,15 @@ func (n *Node) QueueCall(ctx context.Context, call []byte) error {
 		return api.ErrNotLeader
 	}
 
+	if n.checkTxEnabled {
+		// Check transaction before queuing it.
+		err := n.CheckTx(ctx, call)
+		if err != nil {
```
kostko: Could be `if err := n.CheckTx(...); err != nil { ...`.

abukosek:
Fixed!  Ready for re-review :)

Issue #2406, opened 2019-11-21
Enable runtime registrations


kostko:
Currently runtime registrations are disabled, we should make sure that anyone can register a runtime, subject to certain provisions like paying gas fees and maybe also depositing stake in escrow.

## Linked Issues
* [x] [Properly allocate runtime IDs](https://github.com/oasislabs/oasis-core/issues/1693)
* [x] [Compute runtime's key manager should be checked](https://github.com/oasislabs/oasis-core/issues/2459)
* [ ] [Make sure we can scale to lots of runtimes](https://github.com/oasislabs/oasis-core/issues/2349)
* [x] Consensus fee model for runtimes
* [x] Allow anyone to register a runtime
* [x] Suspend runtimes which cannot pay housekeeping fees
* [x] [Compute nodes should set fees for consensus transactions](https://github.com/oasislabs/oasis-core/issues/2521)
* [x] [Add key manager policy document submission](https://github.com/oasislabs/oasis-core/issues/2516)
* [x] [Allow anyone to register a key manager](https://github.com/oasislabs/oasis-core/issues/2445)
* [x] [Flexible key manager policy signers](https://github.com/oasislabs/oasis-core/issues/2444)
* [x] [Remove granular role registration](https://github.com/oasislabs/oasis-core/issues/2107)
* [x] [Runtime node admission policies](https://github.com/oasislabs/oasis-core/issues/2513)
* [x] [runtime: Stop hardcoding runtime IDs into enclave binaries](https://github.com/oasislabs/oasis-core/issues/2519)
* [x] [Registering a runtime should require a deposit](https://github.com/oasislabs/oasis-core/issues/2638)
* [x] Enable runtime registrations by default (invert/rename the consensus parameter).

*Estimated cost: 3 sprints*

PR #2504, opened 2019-12-30
Roothash processing fees

kostko:
Part of #2406.

TODO
* [x] Charge per-epoch maintenance fee.
* [x] Suspend runtimes which did not pay the maintenance fee.
* [x] Restore suspended runtimes from genesis dump.
* [x] Resume runtimes which paid the maintenance fee.
* [x] Emit runtime registered event when the runtime is resumed.
* [x] Add missing existence checks to `registerRuntime`.
* [x] E2E tests.


go/consensus/tendermint/apps/registry/genesis.go
```diff
@@ -75,6 +75,25 @@ func (app *registryApplication) InitChain(ctx *abci.Context, request types.Reque
...
+			"runtime_owner", v.Signature.PublicKey,
+		)
+		if err := app.registerRuntime(ctx, state, v); err != nil {
+			app.logger.Error("InitChain: failed to register runtime",
+				"err", err,
+				"runtime", v,
+			)
+			return errors.Wrap(err, "registry: genesis suspended runtime registration failure")
+		}
+		var rt registry.Runtime
```
Yawning: Instead of doing this, just make `app.registerRuntime` return `(*registry.Runtime, error)`?

Otherwise, the `cbor.Unmarshal` call failing is a sign that something when absolutely horrifically wrong since the runtime registration does an Unmarshal.

go/consensus/tendermint/apps/registry/state/state.go
```diff
@@ -238,82 +242,157 @@ func (s *ImmutableState) SignedNodes() ([]*node.SignedNode, error) {
...
+}
+
+func (s *ImmutableState) getRuntime(keyFmt *keyformat.KeyFormat, id common.Namespace) (*registry.Runtime, error) {
+	signedRuntime, err := s.getSignedRuntime(keyFmt, id)
+	if err != nil {
 		return nil, err
 	}
 	var runtime registry.Runtime
-	if err = cbor.Unmarshal(signedRuntime.Blob, &runtime); err != nil {
+	if err := cbor.Unmarshal(signedRuntime.Blob, &runtime); err != nil {
```
Yawning: Why the change here? `err` is already declared and the correct type...

go/consensus/tendermint/apps/registry/state/state.go
```diff
@@ -238,82 +242,157 @@ func (s *ImmutableState) SignedNodes() ([]*node.SignedNode, error) {
...
+		}
+		runtimes = append(runtimes, &rt)
+	})
+
+	return runtimes, nil
+}
+
+// AllRuntimes returns a list of all registered runtimes.
+//
+// This excludes any suspended runtimes.
```
Yawning: The comment does not match what the function does.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -225,6 +228,56 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
...
+			"new_node", newNode,
+			"existing_node", existingNode,
+			"entity", newNode.EntityID,
+		)
+		return registry.ErrInvalidArgument
+	}
+
+	// For each runtime the node registers for, require it to pay a maintenance fee for
+	// each epoch the node is registered in.
+	additionalEpochs := newNode.Expiration - uint64(epoch)
```
Yawning: It is possible to create a new node, that expires in the current epoch (additionalEpochs = 0), without paying gas.  I don't think the node will actually do anything, but that doesn't seem great.

go/oasis-test-runner/scenario/e2e/identity_cli.go
```diff
@@ -54,7 +55,7 @@ func (i *identityCLIImpl) Run(childEnv *env.Env) error {
 		"identity", "init",
 		"--" + common.CfgDataDir, i.dataDir,
 	}
-	if err := runSubCommand(childEnv, "identity-init", i.nodeBinary, args); err != nil {
+	if err := cli.RunSubCommand(childEnv, i.logger, "identity-init", i.nodeBinary, args); err != nil {
```
Yawning: Not your fault but the receiver `i` makes me want to change it to `ident` or something.

kostko: Yeah, good idea, done.

go/consensus/tendermint/apps/registry/transactions.go
kostko: I changed the expiry check above to explicitly disallow registering nodes which expire in the current epoch.

PR #2505, opened 2019-12-30
txsource: short runner

pro-wh:
This adds short versions of the txsource e2e tests for use in regular CI tests. Adding these to the regular CI tests is supposed to keep the system from rotting.

The run-e2e-txsource.sh script is removed. Use .buildkite/scripts/test_e2e.sh with `-t txsource-transfer` instead.


pro-wh:
hm, this hasn't added txsource to the test coverage

PR #2510, opened 2020-01-03
go: Use common.Namespace for runtime IDs

Yawning:
Prerequisite for implementing the new runtime ID allocation scheme, I would like to merge this prior to doing the other changes just to save having to continually rebase this change set.

### DIFF
--- a/go/keymanager/api/api.go
+++ b/go/keymanager/api/api.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/oasislabs/oasis-core/go/common"
 	"github.com/oasislabs/oasis-core/go/common/cbor"
 	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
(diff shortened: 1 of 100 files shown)


go/keymanager/api/api.go
```diff
@@ -142,10 +143,7 @@ type Genesis struct {
 // SanityCheckStatuses examines the statuses table.
 func SanityCheckStatuses(statuses []*Status) error {
 	for _, status := range statuses {
-		// Verify key manager runtime ID.
-		if !status.ID.IsValid() {
-			return fmt.Errorf("keymanager: sanity check failed: key manager runtime ID %s is invalid", status.ID.String())
-		}
+		// TODO: Verify key manager runtime ID.
```
kostko: I assume this will be checked when the allocation scheme is implemented?

Yawning: Yeah.

PR #2514, opened 2020-01-03
go node: unite compute, merge, and transaction scheduler roles

### LINKED ISSUE #2107: Remove granular role registration
_Re: Core Platform Meeting (2019/09/17)_

The registry currently allows granular registration of roles within the compute committee but this should be changed so that the per-runtime compute execution roles (txnscheduler/compute/merge nodes) are now a single "compute" role from the viewpoint of the registry.

The scheduler should still schedule individual roles as-is and would pick from all valid nodes registered with the new "compute" role.

pro-wh:
We're removing the separation among registering nodes for the compute, merge, and transaction scheduler roles. You now have to register for and enable all or none of these roles, under a new, broadened, and confusing--you're welcome--term "compute."

non-changelog data:

- the `.enabled` flag is moved to its own package, `computeenable`. this is because there are dependencies leading toward the merge worker package, but that would be a weird place to put the compute enable flag.
- a backwards incompatible change is made to the on-chain representation of registered nodes, and no migration is provided. you're welcome.

closes #2107

### DIFF
--- a/.changelog/2107.breaking.md
+++ b/.changelog/2107.breaking.md
@@ -1,0 +1,5 @@
+go node: Unite compute, merge, and transaction scheduler roles.
+
+We're removing the separation among registering nodes for the compute, merge, and transaction scheduler roles.
+You now have to register for and enable all or none of these roles, under a new, broadened, and confusing--you're
+welcome--term "compute."
--- a/go/consensus/tendermint/apps/beacon/beacon.go
+++ b/go/consensus/tendermint/apps/beacon/beacon.go
@@ -122,7 +122,12 @@ func (app *beaconApplication) onBeaconEpochChange(ctx *abci.Context, epoch epoch
 	case true:
 		// UNSAFE/DEBUG - Deterministic beacon.
 		entropyCtx = DebugEntropyCtx
-		entropy = []byte("If you change this, you will fuck up the byzantine tests!!!")
+		// We're setting this random seed so that we have suitable committee schedules for Byzantine E2E scenarios,
+		// where we want nodes to be scheduled for only one committee. The permutations derived from this on the first
+		// epoch need to have (i) an index that's compute worker only and (ii) an index that's merge worker only. See
+		// /go/oasis-test-runner/scenario/e2e/byzantine.go for the permutations generated from this seed. These
+		// permutations are generated independently of the deterministic node IDs.
+		entropy = []byte("If you change this, you will fuck up the byzantine tests!!")
 	}
 
 	b := GetBeacon(epoch, entropyCtx, entropy)
--- a/go/consensus/tendermint/apps/beacon/beacon.go
+++ b/go/consensus/tendermint/apps/beacon/beacon.go
@@ -22,7 +22,7 @@ var (
 	errUnexpectedTimer       = errors.New("beacon: unexpected timer")
 
 	prodEntropyCtx  = []byte("EkB-tmnt")
-	debugEntropyCtx = []byte("Ekb-Dumm")
+	DebugEntropyCtx = []byte("Ekb-Dumm")
 
 	_ abci.Application = (*beaconApplication)(nil)
 )
@@ -121,11 +121,11 @@ func (app *beaconApplication) onBeaconEpochChange(ctx *abci.Context, epoch epoch
 		}
 	case true:
 		// UNSAFE/DEBUG - Deterministic beacon.
-		entropyCtx = debugEntropyCtx
+		entropyCtx = DebugEntropyCtx
 		entropy = []byte("If you change this, you will fuck up the byzantine tests!!!")
 	}
 
-	b := getBeacon(epoch, entropyCtx, entropy)
+	b := GetBeacon(epoch, entropyCtx, entropy)
 
 	app.logger.Debug("onBeaconEpochChange: generated beacon",
 		"epoch", epoch,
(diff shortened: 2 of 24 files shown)


pro-wh:
aw this is going to break the byzantine tests isn't it

go/consensus/tendermint/apps/scheduler/scheduler.go
```diff
@@ -393,7 +393,7 @@ func (app *schedulerApplication) isSuitableTransactionScheduler(n *node.Node, rt
 }
 
 func (app *schedulerApplication) isSuitableMergeWorker(n *node.Node, rt *registry.Runtime, ts time.Time) bool {
-	return n.HasRoles(node.RoleMergeWorker)
+	return n.HasRoles(node.RoleComputeWorker)
```
kostko: This (and storage) needs to be made per-runtime (as mentioned in https://github.com/oasislabs/oasis-core/issues/2107#issuecomment-570572948), similar to txnscheduler/compute.

.changelog/2107.breaking.md
```diff
@@ -0,0 +1,5 @@
+go node: Unite compute, merge, and transaction scheduler roles.
+
+We're removing the separation among registering nodes for the compute, merge, and transaction scheduler roles.
+You now have to register for and enable all or none of these roles, under a new, broadened, and confusing--you're
+welcome--term "compute."
```
kostko: Yeah this is quite confusing, maybe we should make it so:
* The overall term for all three is "compute".
* The old "compute" is renamed to "executor" (or something, no strong preference here).

Thoughts?

go/worker/computeenable/init.go
```diff
@@ -0,0 +1,25 @@
+package computeenable
```
kostko: Ugh, this seems ugly. In light of my previous comment around naming we could also do something like this:
* Rename existing `compute` to `executor` (or some similar name as mentioned above).
* Restructure workers to have `go/worker/compute/{txnscheduler,executor,merge}`.
* Have this "computeenable" thing in `go/worker/compute` potentially with some common code that initializes all three workers.

go/consensus/tendermint/apps/scheduler/scheduler.go
pro-wh: ok I added the runtime snippet to these

.changelog/2107.breaking.md
pro-wh: ya ok

pro-wh: #2525

go/worker/computeenable/init.go
pro-wh: #2525 for the first 2 bullet points

pro-wh:
remaining weirdness:
- registry node runtimes are shared across multiple roles
- some tests might not be up to date on per-runtime merge and storage change

go/worker/computeenable/init.go
pro-wh: moved computeenable to compute. initialization stuff is mostly in executor, with that being the one that needs a registration callback (to wait for TEE capability)

go/consensus/tendermint/apps/beacon/beacon.go
```diff
@@ -122,6 +122,7 @@ func (app *beaconApplication) onBeaconEpochChange(ctx *abci.Context, epoch epoch
 	case true:
 		// UNSAFE/DEBUG - Deterministic beacon.
 		entropyCtx = debugEntropyCtx
+		entropy = []byte("luck=1")
```
kostko: Why is this required? Isn't it the same if you just tweak the identity seeds (with different values ofc)? If it is not, add a comment why this is required.

pro-wh: I'll add a comment

pro-wh:
alright, I'd better take inventory of what went down in coverage

1. StateProcessingBatch.cancel
2. Namespace.UnmarshalBinary ErrMalformedNamespace
3. Namespace.UnmarshalText error from base64 DecodeString
4. Namespace.UnmarshalHex error from hex.DecodeString
5. Namespace.Equal `cmp == nil`
6. cmd debug storage check-roots Args error from nrFn
7. ibid, error from ValidateRuntimeIDStr
8. cmd debug storage force-finalize Args
9. cmd debug storage ValidateRuntimeIDStr error from hex.DecodeString
10. cmd debug storage checkDiff error from storageClient.GetDiff
11. ibid, error from write log iterator Next
12. ibid, error from write log iterator Value
13. cmd debug storage doCheckRoots error from storageClient.New
14. ibid, error from client.GetBlock
15. ibid, error from storageWorkerClient.GetLastSyncedRound
16. ibid, retry when storage worker is not synced
17. ibid, error from client.GetBlock
18. cmd debug storage doForceFinalize `failed` branch

I'm stopping here. Observations:
- covered -> not covered transitions are almost all error handling branches, which I suspect are ones that we don't explicitly exercise in tests
- item 18 https://codecov.io/gh/oasislabs/oasis-core/pull/2514/changes#L282 ought to be impossible, because you would have to go through line 247 (never covered) before you could get to line 282 (previously covered, now not covered)
- what is force-finalize? do we exercise it?

pro-wh:
now the runtime-dynamic test doesn't work, because you wouldn't be able to register per-runtime storage nodes until the runtimes are registered

pro-wh:
alright, adjusted as the comment there described

PR #2518, opened 2020-01-06
Add script to initialize an oasis-node client

david-yan:
- Pulls and compares the genesis file from a provided URL. If a new genesis file is pulled, wipe state.
- Starts `oasis-node` client from provided config file.
- Adds script to deployment Docker image.


peterjgilbert:
I'd consider prefixing the script name with `unsafe` or `debug` to make it clear that this script will fetch arbitrary genesis data from the network and may wipe state.

peterjgilbert:
> Please explain why this needs to be in oasis-core?

Good point. What is the workflow that will use this script? Can that script/job implement this logic independently?

peterjgilbert:
> > Please explain why this needs to be in oasis-core?
> 
> Good point. What is the workflow that will use this script? Can that script/job implement this logic independently?

^ @david-yan, @ravenac95 ?

david-yan:
@kostko you're right, there's no reason for this script to exist in oasis-core. I've moved this to a new repo. The new PR can be found here https://github.com/oasislabs/dockerfiles/pull/1

PR #2529, opened 2020-01-09
host/protocol: add RuntimeID to WorkerInfoRequest and use it

matevz:
PR for https://github.com/oasislabs/oasis-core/issues/2519 and https://github.com/oasislabs/oasis-core/issues/1062:
- adds `RuntimeID` field to `WorkerInfoRequest`,
- adds `runtime_id` to runtime `Protocol`
- adds check to simple-keyvalue client that remote `runtime_id` equals the CLI passed one,
- adds new `multiple-runtimes` e2e test which spins up 2 runtimes on each compute node with shared runtime binary,
- adds `make force-test` rule to go/ for running *all* tests without taking caching into account,
- adds sorting scenarios alphabetically when calling `oasis-test-runner list`.

### DIFF
--- a/go/worker/common/host/protocol/types.go
+++ b/go/worker/common/host/protocol/types.go
@@ -1,6 +1,7 @@
 package protocol
 
 import (
+	"github.com/oasislabs/oasis-core/go/common"
 	"github.com/oasislabs/oasis-core/go/common/crypto/hash"
 	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
 	"github.com/oasislabs/oasis-core/go/common/sgx/ias"
@@ -53,7 +54,7 @@ type Body struct {
 	Error *Error `json:",omitempty"`
 
 	// Worker interface.
-	WorkerInfoRequest                    *Empty                                `json:",omitempty"`
+	WorkerInfoRequest                    *WorkerInfoRequest                    `json:",omitempty"`
 	WorkerInfoResponse                   *WorkerInfoResponse                   `json:",omitempty"`
 	WorkerPingRequest                    *Empty                                `json:",omitempty"`
 	WorkerShutdownRequest                *Empty                                `json:",omitempty"`
@@ -94,6 +95,12 @@ type Empty struct {
 // Error is a message body representing an error.
 type Error struct {
 	Message string `json:"message"`
+}
+
+// WorkerInfoRequest is a worker info request message body.
+type WorkerInfoRequest struct {
+	// RuntimeID is the assigned runtime ID of the loaded runtime.
+	RuntimeID common.Namespace `json:"runtime_id"`
 }
 
 // WorkerInfoResponse is a worker info response message body.
--- a/go/worker/common/host/sandboxed.go
+++ b/go/worker/common/host/sandboxed.go
@@ -540,7 +540,9 @@ func (h *sandboxedHost) checkInfo(worker *process) error {
 
 	// Request information about the running runtime and abort if the protocol
 	// version is incompatible.
-	rsp, err := worker.protocol.Call(ctx, &protocol.Body{WorkerInfoRequest: &protocol.Empty{}})
+	rsp, err := worker.protocol.Call(ctx, &protocol.Body{WorkerInfoRequest: &protocol.WorkerInfoRequest{
+		RuntimeID: h.cfg.ID,
+	}})
 	if err != nil || rsp.WorkerInfoResponse == nil {
 		return errors.Wrap(err, "error while requesting runtime info")
 	}
--- a/runtime/src/protocol.rs
+++ b/runtime/src/protocol.rs
@@ -15,7 +15,7 @@ use io_context::Context;
 use slog::Logger;
 
 use crate::{
-    common::{cbor, logger::get_logger, version::Version},
+    common::{cbor, logger::get_logger, runtime::RuntimeId, version::Version},
     dispatcher::Dispatcher,
     rak::RAK,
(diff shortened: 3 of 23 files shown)


go/worker/common/host/protocol/types.go
```diff
@@ -96,6 +96,12 @@ type Error struct {
 	Message string `json:"message"`
 }
 
+// WorkerInfoRequest is a worker info request message body.
+type WorkerInfoRequest struct {
+	// RuntimeID is the assigned runtime ID of the loaded runtime.
+	RuntimeID []byte `json:"runtime_id"`
```
kostko: Use the `common.Namespace` type here?

runtime/src/types.rs
```diff
@@ -56,7 +56,10 @@ pub enum Body {
     },
 
     // Runtime worker interface.
-    WorkerInfoRequest {},
+    WorkerInfoRequest {
+        #[serde(with = "serde_bytes")]
+        runtime_id: Vec<u8>,
```
kostko: Why not use the `RuntimeId` type?

go/worker/common/host/sandboxed.go
```diff
@@ -82,6 +82,7 @@ type process struct {
 
 	logger *logging.Logger
 
+	runtimeID      *common.Namespace
```
kostko: Why is this optional?

runtime/src/types.rs
matevz: Will it import/export itself nicely with `serde_cbor`, if I use `RuntimeId` type?

kostko: Both should work.

runtime/src/protocol.rs
```diff
@@ -62,6 +64,8 @@ pub struct Protocol {
     last_request_id: AtomicUsize,
     /// Pending outgoing requests.
     pending_out_requests: Mutex<HashMap<u64, channel::Sender<Body>>>,
+    /// Runtime ID
+    pub runtime_id: Mutex<Option<RuntimeId>>,
```
kostko: Don't make this public, instead add an accessor that returns a `RuntimeId` (or panics if one is not available -- with corresponding note of this in the documentation block).

runtime/src/protocol.rs
```diff
@@ -62,6 +64,8 @@ pub struct Protocol {
     last_request_id: AtomicUsize,
     /// Pending outgoing requests.
     pending_out_requests: Mutex<HashMap<u64, channel::Sender<Body>>>,
+    /// Runtime ID
```
kostko:
```suggestion
    /// Runtime identifier.
```

tests/runtimes/simple-keyvalue/api/src/api.rs
```diff
@@ -9,6 +9,9 @@ pub struct KeyValue {
 }
 
 runtime_api! {
+    //  Gets runtime ID of the runtime.
+    pub fn get_remote_runtime_id(()) -> Option<String>;
```
kostko: Why "remote"? Everything is remote :-)

tests/runtimes/simple-keyvalue/src/main.rs
```diff
@@ -243,21 +251,18 @@ fn main() {
...
-        // HACK: Tests always use this runtime ID.
-        let runtime_id =
-            RuntimeId::from_str("8000000000000000000000000000000000000000000000000000000000000000")
-                .unwrap();
-
         // Create the key manager client.
+        let rt_id = protocol.get_runtime_id();
         let km_client = Arc::new(oasis_core_keymanager_client::RemoteClient::new_runtime(
-            runtime_id,
+            rt_id, // TODO Matevz: Should we pass simple-keyvalue runtime ID or the one of the key manager?
```
kostko: The simple-keyvalue runtime ID.

runtime/src/protocol.rs
```diff
@@ -84,12 +88,21 @@ impl Protocol {
             stream,
             last_request_id: AtomicUsize::new(0),
             pending_out_requests: Mutex::new(HashMap::new()),
+            runtime_id: Mutex::new(None),
             runtime_version: runtime_version,
         }
     }
 
+    /// Return the runtime identifier for this worker. Panics, if the runtime identifier is not set.
+    pub fn get_runtime_id(self: &Arc<Protocol>) -> RuntimeId {
```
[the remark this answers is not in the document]
kostko: Also for the documentation block, put the notes about panicking into a separate section, e.g.:
```
# Panics

The method will panic if the runtime identifier is not set.
```

tests/runtimes/simple-keyvalue/api/src/api.rs
matevz: I was mixing the `get_runtime_id()` of the protocol and this one, so I deliberately name it _remote. But it's better to stay consistent, yeah.

runtime/src/protocol.rs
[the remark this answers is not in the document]
matevz: Wow, I didn't know rustdoc has special blocks `# Errors` and `# Panics` for this.

.changelog/2529.feature.md
```diff
@@ -0,0 +1,3 @@
+`RuntimeID` is not hardcoded anymore in the enclave, but is passed when
```
kostko: This is a breaking change as it breaks the runtime protocol. Change the type and mention in a paragraph that it breaks the runtime protocol.

This also requires a runtime protocol version bump (in both Go and Rust).

PR #2530, opened 2020-01-09
go: Use the new runtime ID allocation scheme

Yawning:
* [x] Use the new runtime ID allocation scheme.
 * [x] Document the new runtime ID allocation scheme.
 * [x] Write one of those stupid changelog fragments.


go/oasis-node/cmd/registry/runtime/runtime.go
```diff
@@ -260,8 +260,20 @@ func runtimeFromFlags() (*registry.Runtime, signature.Signer, error) {
 			}
 			kmID = &tmpKmID
 		}
+		if id.IsKeyManager() {
+			logger.Error("runtime ID does has the key manager flag set",
```
kostko:
```suggestion
			logger.Error("runtime ID has the key manager flag set",
```

go/registry/api/api.go
```diff
@@ -1005,6 +1031,10 @@ type ConsensusParameters struct {
 	// allowed outside of the genesis block.
 	DebugAllowRuntimeRegistration bool `json:"debug_allow_runtime_registration"`
 
+	// DebugAllowTestRuntimes is true iff test runtimes should be allowed to
+	// be registered.
+	DebugAllowTestRuntimes bool `json:"debug_allow_test_runtimes"`
```
kostko: Require "don't blame oasis"?

Yawning: I can see external developers using this flag, unlike a lot of the other ones that are feature gated that way.  If we are serious about having third parties writing runtimes, it is probably best not to feature gate it that way?

kostko: Yeah, makes sense.

Issue #2527, opened 2020-01-08
Assign enum values greater than 0


matevz:
Currently, `enum` values in both go and rust code mostly start counting from `0` (e.g. `KindCompute = 0`, `KindKeyManager = 1`). This caused troubles in the past, for example when object you were looking for in Tendermint state or a tree didn't exist, the object was initialized with default values and random stuff was starting to appear in the logs.

For sanity, always start counting enum values from `1` and reserve `0` for `Invalid`.

kostko:
> For sanity, always start counting enum values from 1 and reserve 0 for Undefined.

So `0` shouldn't be undefined it should be invalid and explicitly rejected.

PR #2533, opened 2020-01-10
go registry: for runtimes to whitelist node operators

### LINKED ISSUE #2513: Runtime node admission policies
Each runtime should be able to define its node admission policy.

Currently only two policies should be supported:
* Entity whitelist (only nodes belonging to whitelisted entities can register to host a runtime).
* Anyone with enough stake (currently the only supported policy).

The second one (anyone with enough stake) can introduce liveness issues as long as there is no slashing for compute node liveness (see #2078).

pro-wh:
closes #2513

no additional info

### DIFF
--- a/.changelog/2513.feature.md
+++ b/.changelog/2513.feature.md
@@ -1,0 +1,9 @@
+Runtime node admission policies.
+
+With this, each runtime can define its node admission policy.
+
+Currently only two policies should be supported:
+* Entity whitelist (only nodes belonging to whitelisted entities can register to host a runtime).
+* Anyone with enough stake (currently the only supported policy).
+
+The second one (anyone with enough stake) can introduce liveness issues as long as there is no slashing for compute node liveness (see #2078).
--- a/go/registry/api/runtime.go
+++ b/go/registry/api/runtime.go
@@ -135,6 +135,20 @@ type StorageParameters struct {
 	GroupSize uint64 `json:"group_size"`
 }
 
+// AnyNodeRuntimeAdmissionPolicy allows any node to register.
+type AnyNodeRuntimeAdmissionPolicy struct{}
+
+// EntityWhitelistRuntimeAdmissionPolicy allows only whitelisted entities' nodes to register.
+type EntityWhitelistRuntimeAdmissionPolicy struct {
+	Entities map[signature.PublicKey]bool `json:"entities"`
+}
+
+// RuntimeAdmissionPolicy is a specification of which nodes are allowed to register for a runtime.
+type RuntimeAdmissionPolicy struct {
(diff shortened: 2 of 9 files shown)


go/registry/api/runtime.go
```diff
@@ -160,6 +160,13 @@ type Runtime struct {
 
 	// Storage stores parameters of the storage committee.
 	Storage StorageParameters `json:"storage,omitempty"`
+
+	// NodeEntityWhitelist is a map of entity IDs that may register nodes for this runtime to the true.
+	// This list applies to all roles.
+	NodeEntityWhitelist map[signature.PublicKey]bool `json:"node_entity_whitelist"`
+
+	// AllowAnyNode--if this is set, then nodes from entities other than those in NodeEntityWhitelist are allowed too.
+	AllowAnyNode bool `json:"allow_any_node"`
```
kostko: Instead of this I would have an `AdmissionPolicyKind` enum with `AdmissionPolicyEntityWhitelist` and `AdmissionPolicyAnyone` kinds. (Default kind should be invalid and rejected, as per #2527).

So maybe it would make sense to make a separate `AdmissionPolicy` structure (and corresponding field in `Runtime`) with `Kind` and `EntityWhitelist` fields to group things nicely and make it less messier to extend in the future.

pro-wh:
🤦 this is gonna take a lot more boilerplate

go/registry/api/runtime.go
pro-wh: our other tagged union hack was to have a struct with a bunch of pointer fields

pro-wh: hm although that was probably for rust interop

kostko: Thought about that as well, I would be ok with that too.

pro-wh: oh, one of the variants is a unit type, so there will be no rust interop today

pro-wh: I went with pointer fields \:

pro-wh:
@peterjgilbert want to review? there's been lots of new commits since the last approval

.changelog/2513.feature.md
```diff
@@ -0,0 +1,9 @@
+Runtime node admission policies.
+
+Each runtime should be able to define its node admission policy.
```
kostko: Update the text to reflect that this is what is being added not that it should be added :-)

pro-wh: ok

pro-wh:
oh, correction: TestAdmissionPolicy was unfinished and abandoned. I'm removing it. There's only the added tests to node_test.

PR #2534, opened 2020-01-10
runtime: Persist (sealed) RAK in untrusted local storage

kostko:
Previously the RAK was totally ephemeral and got regenerated on each runtime
enclave restart.

This caused problems in case an enclave got restarted in the middle of an epoch
as the RAK only got updated during a node's re-registration (which is once per
epoch). Until the re-registration and an epoch transition happened, all
RAK-signed messages were rejected.

Also changes some of the `Protocol` methods to take a `self: &Arc<Self>` which will be useful for work done in #2529 to defer dispatcher startup after the runtime ID has been received.

### DIFF
--- a/.changelog/2534.feature.md
+++ b/.changelog/2534.feature.md
@@ -1,0 +1,9 @@
+runtime: Persist (sealed) RAK in untrusted local storage.
+
+Previously the RAK was totally ephemeral and got regenerated on each runtime
+enclave restart.
+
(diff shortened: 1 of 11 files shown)


kostko:
After some more thought, let's not do this. A different solution would be to allow RAK updates mid-epoch. I'll extract the common sealing parts into a separate PR.

PR #2542, opened 2020-01-12
Fix simple-keyvalue-client crash

### LINKED ISSUE #2531: simple-keyvalue-client crashes non-deterministally
#### SUMMARY
`simple-keyvalue-client` (source: tests/clients/simple-keyvalue/src/main.rs) is a simple binary that exercises the default network fixture erected by `oasis-net-runner`. It does so by invoking a series of hard-coded RPCs on the `simple-keyvalue` runtime. It's very much like an e2e test. Sometimes (in my testing, 2 out of 4 times), it crashes.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
tests/clients/simple-keyvalue

##### OASIS NODE VERSION
```paste below
Software version: 19.0-git7c3d5f3c
Runtime protocol version: 0.10.0
Consensus protocol version: 0.22.0
Committee protocol version: 0.7.0
Tendermint core version: 0.32.8
ABCI library version: 0.16.1
```

##### OS / ENVIRONMENT
$ go version
go version go1.13.5 linux/amd64
$ rustc --version
rustc 1.39.0-nightly (521d78407 2019-08-25)

##### STEPS TO REPRODUCE
Follow steps from oasis-core README.md:
```
./go/oasis-net-runner/oasis-net-runner \
  --net.node.binary go/oasis-node/oasis-node \
  --net.runtime.binary target/debug/simple-keyvalue \
  --net.runtime.loader target/debug/oasis-core-runtime-loader \
  --net.keymanager.binary target/debug/oasis-core-keymanager-runtime

./target/debug/simple-keyvalue-client \
  --runtime-id 0000000000000000000000000000000000000000000000000000000000000000 \
  --node-address unix:/tmp/oasis-net-runner<REPLACE_THIS_WITH_YOUR_VALUE>/net-runner/network/client-0/internal.sock
```

(Note: It's not OK to have an existing instance of the network fixture already running, or a different crash will occur. Start oasis-net-runner from scratch.)

##### ACTUAL RESULTS
```paste below
[...]
Querying transaction tags (kv_op=insert)...                                                                                                                                                                
Found transaction: index=0 input=TxnCall { method: "insert", args: Map({Text("key"): Text("hello_key"), Text("value"): Text("hello_value")}) } output=Success(Null)                                        
Querying transaction tags (kv_op=insert)...                                                                                                                                                                
thread 'main' panicked at 'query transactions: CallFailed("Codec(ErrorImpl { code: Message(\"invalid type: null, expected a sequence\"), offset: 0 })")                                                    
```

When running with backtrace, line 209 from tests/clients/simple-keyvalue/src/main.rs is the last identifiable problatic line:
```
199     println!("Querying transaction tags (kv_op=insert)...");
200     let query = Query {
201         round_min: 0,
202         round_max: 3,
203         conditions: vec![QueryCondition {
204             key: b"kv_op".to_vec(),
205             values: vec![b"insert".to_vec().into()],
206         }],
207         limit: 0,
208     };
209     let txns = rt
210         .block_on(kv_client.txn_client().query_txs(query))
211         .expect("query transactions");
212     println!("Found transactions:");
```

##### EXPECTED RESULTS
Client should finish without crashing. It did, 2/4 times.

kostko:
Fixes #2531 

The crash reported in the above issue is actually result of two separate problems, both of which should be fixed by this PR:
* The runtime client endpoint should return empty sequences instead of nil as serde doesn't know how to decode a `NULL` when the expected type is a sequence.
* The simple-keyvalue client used a hardcoded round number in the query (`3`) which can easily be incorrect due to the time at which the client is called.

### DIFF
--- a/.changelog/2542.bugfix.md
+++ b/.changelog/2542.bugfix.md
@@ -1,0 +1,4 @@
+go/runtime/client: Return empty sequences instead of nil.
+
+The runtime client endpoint should return empty sequences instead of `nil` as serde doesn't know how
+to decode a `NULL` when the expected type is a sequence.
--- a/go/runtime/client/client.go
+++ b/go/runtime/client/client.go
@@ -326,7 +326,7 @@ func (c *runtimeClient) GetTxs(ctx context.Context, request *api.GetTxsRequest) 
 		return nil, err
 	}
 
-	var inputs [][]byte
+	inputs := [][]byte{}
 	for _, tx := range txs {
 		inputs = append(inputs, tx.Input)
 	}
@@ -391,7 +391,7 @@ func (c *runtimeClient) QueryTxs(ctx context.Context, request *api.QueryTxsReque
 		return nil, err
 	}
 
-	var output []*api.TxResult
+	output := []*api.TxResult{}
 	for round, txResults := range results {
 		// Fetch block for the given round.
 		var blk *block.Block
--- a/tests/clients/simple-keyvalue/src/main.rs
+++ b/tests/clients/simple-keyvalue/src/main.rs
@@ -150,12 +150,9 @@ fn main() {
 
     // Test wait_block_indexed call.
     println!("Waiting for block to be indexed...");
-    rt.block_on(
-        kv_client
-            .txn_client()
-            .wait_block_indexed(latest_snapshot.block.header.round),
-    )
-    .expect("wait block indexed");
+    let latest_round = latest_snapshot.block.header.round;
+    rt.block_on(kv_client.txn_client().wait_block_indexed(latest_round))
+        .expect("wait block indexed");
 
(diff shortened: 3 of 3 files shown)


kostko:
> Sucks that golang didn't take a clue from Java about how painful it is to maintain nil-safety.

In this case the returned nil is not unsafe. The problem is that in Golang both a `nil` and an empty slice can be used interchangeably (e.g., in range constructs and append etc. builtins). But when serialized into CBOR a `nil` is encoded as CBOR `NULL` while an empty slice is encoded as an empty sequence.

However the Rust serialization library (serde) doesn't know how to deserialize a sequence (e.g. `Vec<T>`) from a `NULL` by default and decoding fails.

Issue #1794, opened 2019-06-07
Support for node re-registration 


ptrus:
From: https://github.com/oasislabs/ekiden/issues/1762#issuecomment-497312500

> Yawning commented 8 days ago
As a companion to this, there should be support for forcing re-registration (eg: the workerhost should trigger re-registration when recovering from the enclave crashing). The current re-registration logic is once per epoch transition with no retries, which will cause hilarity when the RAK gets regenerated.

kostko:
Node re-registration is implemented with #2545. The only remaining part is on the consumer side, the committee bookkeeping component (group) needs to subscribe to registry node updates for any nodes that are part of the committee so it receives any updates.

PR #2545, opened 2020-01-13
go/worker/common/host: Emit events on (re)starts/failures

kostko:
See #1794 

This replaces the previous mechanism which was implemented twice, first
to receive the CapabilityTEE and second to receive the runtime version.

The new mechanism is based on events so that the caller is able to get
notified when changes occur (e.g., on restarts).

### DIFF
--- a/go/worker/registration/worker.go
+++ b/go/worker/registration/worker.go
@@ -11,6 +11,7 @@ import (
 	flag "github.com/spf13/pflag"
 	"github.com/spf13/viper"
 
+	"github.com/oasislabs/oasis-core/go/common"
 	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
 	fileSigner "github.com/oasislabs/oasis-core/go/common/crypto/signature/signers/file"
 	"github.com/oasislabs/oasis-core/go/common/entity"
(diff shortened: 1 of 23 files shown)


kostko:
> Since this adds WaitForStart it might be ok for now, but the moment we add another event, I think this will break, so thought should be given now...

Yeah so currently the broker will replay the last event, so as long as there are only started and failed to start events this is fine. However if we decide to add more, this could cause problems (e.g., missed start event).

I guess we could just get rid of the `Wait*` path and just use events. We will need to do this for triggering registrations anyway, so maybe we should invert the logic now.

kostko:
The second commit now changes the registration worker to be event-based (see commit message for details) and removes (the need for) `WaitForStart`. PTAL.

go/worker/compute/committee/node.go
```diff
@@ -890,6 +910,25 @@ func (n *Node) worker() {
...
+					}
+					rt.Capabilities.TEE = ev.Started.CapabilityTEE
+					nd.Runtimes = append(nd.Runtimes, rt)
+					return nil
+				})
+			case ev.FailedToStart != nil:
+				// Worker failed to start -- we can no longer service requests.
+				n.roleProvider.SetUnavailable()
+			default:
+				// Unknown event.
```
Yawning: Probably want to at least log here?

go/worker/registration/worker.go
```diff
@@ -48,15 +49,57 @@ var (
...
+// RegisterNodeHook is a function that is used to update the node descriptor.
+type RegisterNodeHook func(*node.Node) error
+
 // Delegate is the interface for objects that wish to know about the worker's events.
 type Delegate interface {
 	// RegistrationStopped is called by the worker when the registration loop exits cleanly.
 	RegistrationStopped()
 }
 
+// RoleProvider is a the node descriptor role provider interface.
```
Yawning: "is a the"

go/worker/compute/committee/node.go
kostko: Good idea, added.

PR #2554, opened 2020-01-15
"unfatalize"

pro-wh:
for record keeping: we put these bandaids on the seed node to skip over the unroutable nodes in the genesis document

### DIFF
--- a/.changelog/2554.bugfix.md
+++ b/.changelog/2554.bugfix.md
@@ -1,0 +1,6 @@
+go tendermint: Unfatalize seed populating nodes from genesis
+
+In some cases we'd prefer to include some nodes in the genesis document
+even when they're registered with an invalid address.
+This makes the seed node ignore those entries and carry on, while
+keeping those entries available for the rest of the system.
--- a/go/consensus/tendermint/seed.go
+++ b/go/consensus/tendermint/seed.go
@@ -21,6 +21,8 @@ import (
 	registry "github.com/oasislabs/oasis-core/go/registry/api"
 )
 
+var logger = logging.GetLogger("consensus/tendermint/seed")
+
 // SeedService is a Tendermint seed service.
 type SeedService struct {
 	addr      *p2p.NetAddress
@@ -169,7 +171,7 @@ func populateAddrBookFromGenesis(addrBook p2p.AddrBook, doc *genesis.Document, o
 		var tmvAddr *p2p.NetAddress
 		tmvAddr, err := api.NodeToP2PAddr(&openedNode)
 		if err != nil {
-			logging.GetLogger("seed").Error("tendermint/seed: failed to reformat genesis validator address",
+			logger.Error("failed to reformat genesis validator address",
 				"err", err,
 			)
 			continue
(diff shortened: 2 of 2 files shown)


go/consensus/tendermint/seed.go
```diff
@@ -173,7 +174,10 @@ func populateAddrBookFromGenesis(addrBook p2p.AddrBook, doc *genesis.Document, o
 		var tmvAddr *p2p.NetAddress
 		tmvAddr, err := api.NodeToP2PAddr(&openedNode)
 		if err != nil {
-			return errors.Wrap(err, "tendermint/seed: failed to reformat genesis validator address")
+			logging.GetLogger("seed").Error("tendermint/seed: failed to reformat genesis validator address",
```
kostko: Create a (global?) logger instance and use it here and below instead of creating a new instance each time.

pro-wh: ya good idea

.changelog/2554.bugfix.md
```diff
@@ -0,0 +1,6 @@
+go tendermint: Unfatalize seed populating nodes from genesis
+
+In some cases we'd prefer to include some nodes in the genesis document
+even when they're registered with an invalid address.
+This makes the seed nodeignore those entries and carry on, while
```
kostko:
```suggestion
This makes the seed node ignore those entries and carry on, while
```

go/consensus/tendermint/seed.go
```diff
@@ -168,7 +171,10 @@ func populateAddrBookFromGenesis(addrBook p2p.AddrBook, doc *genesis.Document, o
 		var tmvAddr *p2p.NetAddress
 		tmvAddr, err := api.NodeToP2PAddr(&openedNode)
 		if err != nil {
-			return errors.Wrap(err, "tendermint/seed: failed to reformat genesis validator address")
+			logger.Error("tendermint/seed: failed to reformat genesis validator address",
```
kostko: No need to include the `tendermint/seed` prefix, instead use `consensus/tendermint/seed` for the logger module. Also below.

go/consensus/tendermint/seed.go
```diff
@@ -13,13 +13,16 @@ import (
...
 	"github.com/oasislabs/oasis-core/go/common/identity"
+	"github.com/oasislabs/oasis-core/go/common/logging"
 	"github.com/oasislabs/oasis-core/go/common/node"
 	"github.com/oasislabs/oasis-core/go/consensus/tendermint/api"
 	"github.com/oasislabs/oasis-core/go/consensus/tendermint/crypto"
 	genesis "github.com/oasislabs/oasis-core/go/genesis/api"
 	registry "github.com/oasislabs/oasis-core/go/registry/api"
 )
 
+var logger = logging.GetLogger("seed")
```
kostko:
```suggestion
var logger = logging.GetLogger("consensus/tendermint/seed")
```

go/consensus/tendermint/seed.go
pro-wh: ok

PR #2558, opened 2020-01-16
Don't allow duplicate P2P connections from the same IP by default

peterjgilbert:


go/consensus/tendermint/tendermint.go
```diff
@@ -82,6 +82,8 @@ const (
 	CfgP2PPersistentPeer = "tendermint.persistent_peer"
 	// CfgP2PDisablePeerExchange disables tendermint's peer-exchange (Pex) reactor.
 	CfgP2PDisablePeerExchange = "tendermint.disable_peer_exchange"
+	// CfgP2PAllowDuplicateIP allows multiple connections from the same IP.
+	CfgP2PAllowDuplicateIP = "tendermint.allow_duplicate_ip"
```
Yawning: This should be `tendermint.debug.allow_duplicate_ip` for consistency with other options.

Issue #2362, opened 2019-11-11
Simplify sentry node setup


tjanez:
After [initial Sentry node support is implemented](https://github.com/oasislabs/oasis-core/pull/2300), simplify the sentry node setup:
- Instead of manually and statically specifying sentry nodes as validator's persistent peers, support automatically configuring them upon querying sentry nodes for their consensus addresses and automatically restarting the Tenderming service (if necessary).
- Instead of manually disabling Tendermint's Peer exchange reactor when using sentry nodes, do that automatically (if sentry nodes are configured).

ptrus:
> Instead of manually and statically specifying sentry nodes as validator's persistent peers, support automatically configuring them upon querying sentry nodes for their consensus addresses and automatically restarting the Tenderming service (if necessary).

looking at the tendermint code, this would require restarting the tedermint service. That together with the fact that validators don't technically need to specify sentries as persistent peers probably makes implementing this not really worth considering the effort vs benefits.

> Instead of manually disabling Tendermint's Peer exchange reactor when using sentry nodes, do that automatically (if sentry nodes are configured).

Since `disable_pex` flag is useful by itself (e.g. someone might want to connect to known trusted nodes via persistent peers and disable pex) i think having to manually disable pex reactor is fine.


However i did try to address some other "simplifications" in #2560

kostko:
> Since disable_pex flag is useful by itself (e.g. someone might want to connect to known trusted nodes via persistent peers and disable pex) i think having to manually disable pex reactor is fine.

Yeah having an option to disable PEX manually seems reasonable, but is there ever a need for having PEX enabled on the upstream node when using sentry nodes? If not, you can do both -- keep the manual "disable PEX" option and also automatically disable PEX on the upstream node when using sentry nodes.

PR #2560, opened 2020-01-16
go/sentry: simplify tm sentry flags

ptrus:
Fixes: #2362 

This changes the sentry configuration to:
- on sentry node:
  - single flag:  `--tendermint.sentry.upstream_address` 
- on upstream node:
  - `--worker.registration.sentry.address`, `--worker.registration.sentry.cert_fiile` (for querying addresses at registration time)
  -  `--tendermint.disable_peer_exchange` 

One regression to the previous sentry setup is that the upstream node no longer sets "tendermint.persistent_peers" to sentry nodes. I think this is fine, since in any real sentry node setup, the sentry nodes are the only nodes connecting to the upstream node, so not sure if setting persistent peers is/was really needed/useful. Let me know what others think.

Also updated the E2E tests so that nodes behind sentries do not connect to the seed node, since currently the sentry setup was not really properly tested as all nodes got addresses of upstream nodes through the seed node. In future should probably enhance the sentry E2E test further, by blocking all connections to the upstream nodes not originating from sentry nodes.

### DIFF
--- a/go/consensus/tendermint/tendermint.go
+++ b/go/consensus/tendermint/tendermint.go
@@ -76,8 +76,9 @@ const (
 	cfgABCIPruneStrategy = "tendermint.abci.prune.strategy"
 	cfgABCIPruneNumKept  = "tendermint.abci.prune.num_kept"
 
-	// CfgP2PPrivatePeerID configures tendermint's private peer ID(s).
-	CfgP2PPrivatePeerID = "tendermint.private_peer_id"
+	// CfgSentryUpstreamAddress defines nodes for which we act as a sentry for.
+	CfgSentryUpstreamAddress = "tendermint.sentry.upstream_address"
+
 	// CfgP2PPersistentPeer configures tendermint's persistent peer(s).
 	CfgP2PPersistentPeer = "tendermint.persistent_peer"
 	// CfgP2PDisablePeerExchange disables tendermint's peer-exchange (Pex) reactor.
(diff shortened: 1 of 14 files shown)


go/consensus/tendermint/tendermint.go
```diff
@@ -76,10 +76,9 @@ const (
 	cfgABCIPruneStrategy = "tendermint.abci.prune.strategy"
 	cfgABCIPruneNumKept  = "tendermint.abci.prune.num_kept"
 
-	// CfgP2PPrivatePeerID configures tendermint's private peer ID(s).
-	CfgP2PPrivatePeerID = "tendermint.private_peer_id"
-	// CfgP2PPersistentPeer configures tendermint's persistent peer(s).
-	CfgP2PPersistentPeer = "tendermint.persistent_peer"
```
kostko: Hm, actually having an option to configure persistent peers may still be useful (e.g., to establish connections to known good peers).

kostko:
> I think this is fine, since in any real sentry node setup, the sentry nodes are the only nodes connecting to the upstream node, so not sure if setting persistent peers is/was really needed/useful. Let me know what others think.

So one place where persistent peers would IMO be useful is on the sentry nodes in order to ensure connections to some "known-good" nodes are established (if available).

ptrus:
> So one place where persistent peers would IMO be useful is on the sentry nodes in order to ensure connections to some "known-good" nodes are established (if available).

yeah agreed, i think this is actually used a lot in in cosmos/tendermint (e.g. "relay nodes"). I will leave it in, and make the added sentry flag just append to the persistent peers.

go/consensus/tendermint/tendermint.go
ptrus: yeah added it back, so now i'm also thinking if there's even any benefit in setting `private_peer_ids` on the sentry node in our case. In tendermint's case it's useful since it prevents others from knowing which nodes are behind a specific sentry. In our case this isn't really true, since this will be visible from the registry. 

This would mean that there's no benefit of the (just added) `tendermint.sentry.upstream_address` flag, since it would do the same as the `persistent_peers` flag. So in that case, the sentry node configuration would be:

- on sentry node:
  - set sentry nodes as private_peers
- on upstream node:
  - `worker.registration.sentry.address`, `worker.registration.sentry.cert_fiile` (for querying addresses at registration time) (just renames of the existing flags)
  - `tendermint.disable_peer_exchange`

^ which seems simple enough

go/consensus/tendermint/tendermint.go
```diff
@@ -898,16 +899,12 @@ func (t *tendermintService) lazyInit() error {
 	tenderConfig.TxIndex.Indexer = "null"
 	tenderConfig.P2P.ListenAddress = viper.GetString(CfgCoreListenAddress)
 	tenderConfig.P2P.ExternalAddress = viper.GetString(cfgCoreExternalAddress)
-	// Convert persistent peer IDs to lowercase (like other IDs) since
-	// Tendermint stores them in a map and uses a case sensitive string
-	// comparison to check ID equality.
-	tenderConfig.P2P.PrivatePeerIDs = strings.ToLower(strings.Join(viper.GetStringSlice(CfgP2PPrivatePeerID), ","))
+	tenderConfig.P2P.PexReactor = !viper.GetBool(CfgP2PDisablePeerExchange)
```
kostko: Should configuring a node to use a sentry also disable the PEX reactor automatically or is there a reason why you would want to have it enabled?

ptrus: i think that we currently don't have (or need) an explicit `tendermint.sentry.enabled` flag. So it would just be a duplicated flag of the `CfgP2PDisablePeerExchange`.

We could do it based on the `worker.registration.sentry.address` flag, but that assumes that the registration worker is enabled (which currently always is). But i image there being a case where someone would handle registration differently, and wouldn't even need to set the sentry flags.

But yeah if the `worker.registration.sentry.address` is indeed set, then we could automatically disable pex, but just feels a bit to magic and maybe explicitly disabling it might be better.

kostko: Oh I see, ok then it is fine as-is.

ptrus:
thanks for review!

PR #2575, opened 2020-01-21
go/registry: deduplicate registry sanity checks

ptrus:
Fixes: #2428

Deduplicate the node sanity checks by reusing `VerifyNodeRegistrationArgs` function.

TODO:
- [x] do similar for enitites and runtimes
- [x] move registry sanity checks into a separate file
- [x] move other sanity checks


ptrus:
> Yeah this would probably be a good idea for any other modules that currently have everything in api.go.

If it's just moving it similarly as i did for the `registry/api.go` (without any additional changes), then i can do it as part of this PR. Otherwise i'll rather open an issue for it and do it in a separate PR.

kostko:
> If it's just moving it similarly as i did for the registry/api.go (without any additional changes), then i can do it as part of this PR. Otherwise i'll rather open an issue for it and do it in a separate PR.

Yeah it should be just that.

go/registry/api/sanity_check.go
```diff
@@ -0,0 +1,220 @@
...
+	return nil
+}
+
+// Runtimes lookup used in sanity checks.
+type sanityCheckRuntimeLookup struct {
+	runtimes          map[common.Namespace]*Runtime
+	suspendedRuntimes map[common.Namespace]*Runtime
+}
+
+func NewRuntimeLookup(runtimes []*Runtime, suspendedRuntimes []*Runtime) RuntimeLookup {
```
kostko: Does this need to be public? Also maybe rename to something like `newSanityCheckRuntimeLookup`.

ptrus:
one regression caused by the change is that now sanity check failures don't contain the detailed message, since in `VerifyRegister*` methods return generic `ErrInvalidArgument` errors. The reason for failure can still be seen from surrounding logs though, but maybe we consider updating the `VerifyRegister*` methods to return error reason in returned errors.

kostko:
> The reason for failure can still be seen from surrounding logs though, but we should consider updating the VerifyRegister* methods to return failure reason in returned errors.

As long as the "invalid argument" error is wrapped, you can add arbitrary formatting (the error propagation code should automatically unwrap errors). So something like:
```golang
fmt.Errorf("%w: my nice message", ErrInvalidArgument)
```

ptrus:
> As long as the "invalid argument" error is wrapped, you can add arbitrary formatting (the error propagation code should automatically unwrap errors). So something like:

yeah i'm asking if i should do it :-) I guess no reason not to.

PR #2584, opened 2020-01-22
go/worker/common: Refresh node descriptors mid-epoch

### LINKED ISSUE #1794: Support for node re-registration

### LINKED ISSUE #2193: key manager: client mixes together all nodes' certs
so there's no authentication between different nodes at the TLS layer. not sure if that's desired though

## Details

encountered in go/keymanager/client/client.go

from internal key manager audit

## Acceptance Criteria
* if we want authentication, we change this

kostko:
Fixes #1794.
Fixes #2193 

Previously node descriptors were only refreshed on an epoch transition which
meant that any later updates were ignored. This caused stale RAKs to stay in
effect when runtime restarts happened.

Enabling mid-epoch refresh also makes having more ephemeral keys easier.

TODO
* [x] Watch node descriptors of committee members.
* [x] Remove `NodeInfo` from commitment pool.
* [x] Add a hook to get notified of node descriptor updates for things like gRPC policy updates.
* [x] Add general node/committee watcher interfaces.
* [x] Finish gRPC committee client.
* [x] Replace custom watchers in storage/runtime clients with generalized interfaces.
* [x] Replace custom watchers in keymanager clients with generalized interfaces.
* [ ] Tests for the new general interface.


go/runtime/committee/client.go
```diff
@@ -0,0 +1,406 @@
...
+	sync.Mutex
+
+	nodes []signature.PublicKey
+	index int
+}
+
+func (rr *roundRobinNodeSelectionPolicy) UpdateNodes(nodes []signature.PublicKey) {
+	rr.Lock()
+	rr.nodes = nodes
+	rr.index = 0
```
Yawning: Won't this cause every client to hit the same node?

kostko: Good idea, I added a random shuffle.

PR #2593, opened 2020-01-24
client/src/transaction: add storage Apply gRPC method

ptrus:
Needed by `genesis` in https://github.com/oasislabs/oasis-runtime

### DIFF
--- a/client/src/transaction/api/storage.rs
+++ b/client/src/transaction/api/storage.rs
@@ -1,5 +1,6 @@
 //! Client for service defined in go/storage/api.
 use grpcio::{CallOption, Channel, Client, Result};
+use serde_cbor::value::Value;
 use serde_derive::{Deserialize, Serialize};
 
 use oasis_core_runtime::{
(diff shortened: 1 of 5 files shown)


client/src/transaction/api/storage.rs
```diff
@@ -17,6 +18,12 @@ pub struct ApplyRequest {
     pub writelog: WriteLog,
 }
 
+grpc_method!(
+    METHOD_APPLY,
+    "/oasis-core.Storage/Apply",
+    ApplyRequest,
+    Value
```
kostko: So this being `Value` is not very useful. Can you add a receipt type, even if it is just a value internally, e.g.:
```rust
// XXX: This should be an actual receipt once somebody needs it from Rust.
/// A storage receipt.
type Receipt = Value;
```
And then the result type should be `Vec<Receipt>`.

client/src/transaction/api/storage.rs
```diff
@@ -50,6 +57,10 @@ impl StorageClient {
         }
     }
 
+    pub fn apply(&self, request: &ApplyRequest, opt: CallOption) -> Result<Value> {
```
kostko: Add documentation comment.

PR #2600, opened 2020-01-27
registry: bug fixes

### LINKED ISSUE #2058: go/common/identity: Use Ed25519 instead of P-256 for TLS certs
We use P-256 for the TLS cert signing key.  Since Go 1.13 supports it, we could switch that to Ed25519.  That said, I don't have anything against P-256 at this moment, and it would only be for consistency.

### LINKED ISSUE #2598: Genesis import fails for expired nodes
Genesis import fails when the registry contains expired nodes:
```
level=error module=tendermint/registry caller=genesis.go:83 ts=2020-01-25T22:24:31.352771Z msg="InitChain: failed to register node" err="registry: node expired" node="unsupported value type"
level=error module=abci-mux caller=mux.go:433 ts=2020-01-25T22:24:31.352813Z msg="InitChain: fatal error in application" err="registry: genesis node registration failure: registry: node expired" app=200_registry
```

This is an issue because nodes are kept in the registry after expiring for the debonding period.

### LINKED ISSUE #2599: registry: Node descriptors should be signed by all public keys in the node
The node descriptor isn't actually signed by the consensus, p2p (and in some cases the node descriptor's own) signing keys.  This means that there is no proof of possession of the keys, and with duplicated keys being rejected by the consensus layer, it is possible to craft descriptors that cause valid registrations to be rejected.

Yawning:
Fixes #2058
Fixes #2598
Fixes #2599

### DIFF
--- a/go/consensus/tendermint/apps/registry/transactions.go
+++ b/go/consensus/tendermint/apps/registry/transactions.go
@@ -125,7 +125,7 @@ func (app *registryApplication) deregisterEntity(ctx *abci.Context, state *regis
 func (app *registryApplication) registerNode( // nolint: gocyclo
 	ctx *abci.Context,
 	state *registryState.MutableState,
-	sigNode *node.SignedNode,
+	sigNode *node.MultiSignedNode,
 ) error {
 	if ctx.IsCheckOnly() {
 		return nil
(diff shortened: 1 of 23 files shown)


go/registry/api/api.go
```diff
@@ -426,14 +426,14 @@ func VerifyRegisterNodeArgs( // nolint: gocyclo
...
 
-	// Checking the expiration only makes sense if this routine is
-	// validating the genesis document, since node descriptors being
-	// expired is presumably ok at runtime(?).
-	if isGenesis && n.Expiration <= uint64(epoch) {
-		logger.Error("RegisterNode: expired node in genesis",
+	// Registrations that are expired are allowed in the genesis document
+	// as expired node registrations are preserved through the debonding
+	// period of an entity.
+	if !isGenesis && n.Expiration <= uint64(epoch) {
```
ptrus: mind also deduplicating this with: https://github.com/oasislabs/oasis-core/blob/master/go/consensus/tendermint/apps/registry/transactions.go#L242-L257 (more ore less removing the check there, as that code path already calls `VerifyRegisterNodeArgs`)

ptrus: (or at minimum the check there needs to be fixed as well to only check for non-genesis)

go/registry/api/api.go
```diff
@@ -402,37 +403,37 @@ func VerifyRegisterNodeArgs( // nolint: gocyclo
...
+		logger.Error("RegisterNode: registration is expired",
 			"node", n,
 		)
-		return nil, nil, fmt.Errorf("%w: expired node in genesis", ErrInvalidArgument)
+		return nil, nil, fmt.Errorf("%w: registration is expired", ErrInvalidArgument)
```
ptrus: i know this was wrong before, but it should probably be `ErrNodeExpired`

Yawning: Eh, nothing else there does fine grained error reporting.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -236,17 +244,15 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
...
+	// that are being persisted via the debonding period.
+	//
+	// Note: VerifyRegisterNodeArgs has checks around this to make sure
+	// that genesis is the only time this is allowed.
+	var additionalEpochs uint64
+	if newNode.Expiration > uint64(epoch) {
+		additionalEpochs = newNode.Expiration - uint64(epoch)
 	}
-	additionalEpochs := newNode.Expiration - uint64(epoch)
 	if params.MaxNodeExpiration > 0 && additionalEpochs > params.MaxNodeExpiration {
```
ptrus: this one is also already checked in `VerifyRegisterNodeArgs`

go/registry/api/api.go
Yawning: No, I can't deduplicate this, because the supplementary sanity checker calls it, and it will force crash the node when it encounters descriptors in the debonding period.

PR #2607, opened 2020-01-28
Upgrade mechanism

### LINKED ISSUE #2465: Node controller interface to schedule upgrade
As per #2446 we need a node controller interface to schedule upgrades. This includes specifying the format of upgrade descriptors.

### LINKED ISSUE #2466: Binary halts after scheduled upgrade
Depends on #2465.

As per #2446 the oasis-node binary should properly halt once the scheduled upgrade epoch ends. If restarted, the binary should immediately halt before attempting to process any consensus blocks.

### LINKED ISSUE #2467: Migration handlers for upgrading consensus state
Depends on #2466.

As per #2446 we should support writing upgrade-specific migration handlers (based on the unique upgrade name specified in the upgrade descriptor). The upgrade handlers must run in the new binary before any consensus processing starts.

Ideally the old state should be backed up before the migration to support rollback in case of errors.

### LINKED ISSUE #2547: Requesting shutdown twice results in a panic
Calling `oasis-node control shutdown` twice in a row results in a panic due to an attempt to close an already closed channel in `RequestDeregistration`.
https://github.com/oasislabs/oasis-core/blob/de279f77856d7721148e7a672cdf975b0ecc58d3/go/worker/registration/worker.go#L429-L439

jberci:
Fixes #2547 
Closes #2465 
Closes #2466 
Closes #2467

### DIFF
--- a/go/consensus/tendermint/abci/mux.go
+++ b/go/consensus/tendermint/abci/mux.go
@@ -27,6 +27,7 @@ import (
 	"github.com/oasislabs/oasis-core/go/consensus/api/transaction"
 	epochtime "github.com/oasislabs/oasis-core/go/epochtime/api"
 	genesis "github.com/oasislabs/oasis-core/go/genesis/api"
+	upgrade "github.com/oasislabs/oasis-core/go/upgrade/api"
 )
 
 const (
@@ -264,12 +265,12 @@ func (a *ApplicationServer) EstimateGas(caller signature.PublicKey, tx *transact
 
 // NewApplicationServer returns a new ApplicationServer, using the provided
 // directory to persist state.
-func NewApplicationServer(ctx context.Context, cfg *ApplicationConfig) (*ApplicationServer, error) {
+func NewApplicationServer(ctx context.Context, upgrader upgrade.Backend, cfg *ApplicationConfig) (*ApplicationServer, error) {
 	metricsOnce.Do(func() {
 		prometheus.MustRegister(abciCollectors...)
 	})
 
-	mux, err := newABCIMux(ctx, cfg)
+	mux, err := newABCIMux(ctx, upgrader, cfg)
 	if err != nil {
 		return nil, err
 	}
@@ -284,8 +285,9 @@ type abciMux struct {
 	sync.RWMutex
 	types.BaseApplication
 
-	logger *logging.Logger
-	state  *applicationState
+	logger   *logging.Logger
+	upgrader upgrade.Backend
+	state    *applicationState
 
 	appsByName     map[string]Application
 	appsByMethod   map[transaction.MethodName]Application
@@ -457,6 +459,7 @@ func (mux *abciMux) InitChain(req types.RequestInitChain) types.ResponseInitChai
 
 func (mux *abciMux) BeginBlock(req types.RequestBeginBlock) types.ResponseBeginBlock {
 	blockHeight := mux.state.BlockHeight()
+
 	mux.logger.Debug("BeginBlock",
 		"req", req,
 		"block_height", blockHeight,
@@ -479,6 +482,21 @@ func (mux *abciMux) BeginBlock(req types.RequestBeginBlock) types.ResponseBeginB
 	// Create BeginBlock context.
 	ctx := mux.state.NewContext(ContextBeginBlock, mux.currentTime)
 	defer ctx.Close()
+
+	currentEpoch, err := mux.state.GetCurrentEpoch(ctx.Ctx())
+	if err != nil {
+		panic("mux: can't get current epoch in BeginBlock")
+	}
+
+	// Check if there are any upgrades pending or if we need to halt for an upgrade.
+	switch err = mux.upgrader.ConsensusUpgrade(ctx, currentEpoch, blockHeight); err {
+	case nil:
+		// Everything ok.
+	case upgrade.ErrStopForUpgrade:
+		panic("mux: reached upgrade epoch")
+	default:
+		panic(fmt.Sprintf("mux: error while trying to perform consensus upgrade: %v", err))
+	}
 
 	switch mux.state.haltMode {
 	case false:
@@ -867,7 +885,7 @@ func (mux *abciMux) checkDependencies() error {
 	return nil
 }
 
-func newABCIMux(ctx context.Context, cfg *ApplicationConfig) (*abciMux, error) {
+func newABCIMux(ctx context.Context, upgrader upgrade.Backend, cfg *ApplicationConfig) (*abciMux, error) {
 	state, err := newApplicationState(ctx, cfg)
 	if err != nil {
 		return nil, err
@@ -875,6 +893,7 @@ func newABCIMux(ctx context.Context, cfg *ApplicationConfig) (*abciMux, error) {
 
 	mux := &abciMux{
 		logger:         logging.GetLogger("abci-mux"),
+		upgrader:       upgrader,
 		state:          state,
 		appsByName:     make(map[string]Application),
 		appsByMethod:   make(map[transaction.MethodName]Application),
--- a/go/control/control.go
+++ b/go/control/control.go
@@ -5,12 +5,14 @@ import (
 	"context"
 
 	consensus "github.com/oasislabs/oasis-core/go/consensus/api"
-	"github.com/oasislabs/oasis-core/go/control/api"
+	control "github.com/oasislabs/oasis-core/go/control/api"
+	upgrade "github.com/oasislabs/oasis-core/go/upgrade/api"
 )
 
 type nodeController struct {
-	node      api.Shutdownable
+	node      control.Shutdownable
 	consensus consensus.Backend
+	upgrader  upgrade.Backend
 }
 
 func (c *nodeController) RequestShutdown(ctx context.Context, wait bool) error {
@@ -49,10 +51,19 @@ func (c *nodeController) IsSynced(ctx context.Context) (bool, error) {
 	}
 }
 
+func (c *nodeController) UpgradeBinary(ctx context.Context, descriptor *upgrade.Descriptor) error {
+	return c.upgrader.SubmitDescriptor(ctx, descriptor)
+}
+
+func (c *nodeController) CancelUpgrade(ctx context.Context) error {
+	return c.upgrader.CancelUpgrade(ctx)
+}
+
 // New creates a new oasis-node controller.
-func New(node api.Shutdownable, consensus consensus.Backend) api.NodeController {
+func New(node control.Shutdownable, consensus consensus.Backend, upgrader upgrade.Backend) control.NodeController {
 	return &nodeController{
 		node:      node,
 		consensus: consensus,
+		upgrader:  upgrader,
 	}
 }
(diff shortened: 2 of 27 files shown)


go/upgrade/api/api.go
```diff
@@ -0,0 +1,92 @@
...
+
+const (
+	UpgradeStageStartup UpgradeStage = iota
+	UpgradeStageConsensus
+
+	upgradeStageAll = (UpgradeStage(1) << iota) - 1
+)
+
+var (
+	DescriptorSchema = []byte(fmt.Sprintf(
```
kostko: Hm why are we using JSON Schema to validate descriptors here, we are not using it anywhere else? I would suggest the API just takes a `*Descriptor` structure and we can validate it in Go (e.g., add `IsValid`)?

go/upgrade/api/api.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+const (
+	UpgradeMethInternal = "internal"
+)
+
+type UpgradeStage uint64
+
+const (
+	UpgradeStageStartup UpgradeStage = iota
```
kostko: I would avoid iota and add explicit numbers.

go/upgrade/migrations/migrations.go
```diff
@@ -0,0 +1,67 @@
...
+	StartupUpgrade(*Context) error
+	ConsensusUpgrade(*Context) error
+}
+
+type Context struct {
+	Descriptor *upgradeApi.Descriptor
+	Logger     *logging.Logger
+}
+
+type testMigrationHandler struct {
```
kostko: Move the test migration into a separate file.

go/upgrade/upgrade.go
```diff
@@ -0,0 +1,252 @@
+// Package upgrade implements the consensus upgrade backend.
```
kostko:
```suggestion
// Package upgrade implements the node upgrade backend.
```

go/consensus/tendermint/abci/mux.go
```diff
@@ -480,6 +483,20 @@ func (mux *abciMux) BeginBlock(req types.RequestBeginBlock) types.ResponseBeginB
 	ctx := mux.state.NewContext(ContextBeginBlock, mux.currentTime)
 	defer ctx.Close()
 
+	currentEpoch, err := mux.state.currentEpoch(ctx)
+	if err != nil {
+		panic("mux: can't get current epoch in BeginBlock")
+	}
+
+	// Check if there are any upgrades pending or if we need to halt for an upgrade.
+	if err := mux.upgrader.ConsensusUpgrade(ctx, currentEpoch); err != nil {
```
kostko: Could use a switch here instead of nested ifs.

go/go.mod
```diff
@@ -46,6 +46,7 @@ require (
 	github.com/pelletier/go-toml v1.4.0 // indirect
 	github.com/pkg/errors v0.8.1
 	github.com/prometheus/client_golang v0.9.4
+	github.com/qri-io/jsonschema v0.1.1
```
kostko: Is this (and `go.sum` update) still needed?

go/control/control.go
```diff
@@ -46,10 +52,16 @@ func (c *nodeController) IsSynced(ctx context.Context) (bool, error) {
...
 }
 
+func (c *nodeController) UpgradeBinary(ctx context.Context, descriptor *upgrade.Descriptor) error {
+	return c.upgrader.SubmitDescriptor(ctx, descriptor)
+}
+
 // New creates a new oasis-node controller.
-func New(node api.Shutdownable, consensus consensus.Backend) api.NodeController {
+func New(node control.Shutdownable, consensus consensus.Backend, upgrader upgrade.Backend) control.NodeController {
+	fmt.Println("node controller interface created")
```
kostko: Don't forget to remove this.

go/oasis-node/cmd/debug/byzantine/tendermint.go
```diff
@@ -37,7 +37,7 @@ func (ht *honestTendermint) start(id *identity.Identity, dataDir string) error {
 	}
 	genesisDoc.SetChainContext()
 
-	ht.service, err = tendermint.New(context.Background(), dataDir, id, genesis)
+	ht.service, err = tendermint.New(context.Background(), dataDir, id, nil, genesis) // XXX
```
kostko: What's with the XXX? I guess it should be fine if the byzantine node doesn't support upgrades ;-)

go/upgrade/api/api.go
```diff
@@ -0,0 +1,114 @@
...
+	// where the node binary itself has the migration code.
+	UpgradeMethInternal = "internal"
+)
+
+// UpgradeStage is used in the upgrade descriptor to store completed stages.
+type UpgradeStage uint64
+
+const (
+	// UpgradeStageStartup is the startup upgrade stage, executed at the beginning of node startup.
+	UpgradeStageStartup UpgradeStage = 0
```
kostko: Let's make zero an invalid upgrade stage as it is the default value. The first valid stage should start with one.

go/upgrade/api/api.go
```diff
@@ -0,0 +1,114 @@
...
+	// the consensus layer has reached the scheduled shutdown epoch and should be interrupted.
+	ErrStopForUpgrade = errors.New(ModuleName, 1, "upgrade: reached upgrade epoch")
+
+	// ErrUpgradePending is the error returned when there is a pending upgrade and the node detects that it is
+	// not the one performing it.
+	ErrUpgradePending = errors.New(ModuleName, 2, "upgrade: this binary is scheduled to be replaced")
+
+	// ErrUnknownUpgradeMethod is the error returned when the submitted upgrade descriptor specifies an unknown
+	// upgrade method.
+	//ErrUnknownUpgradeMethod = errors.New(ModuleName, 3, "upgrade: upgrade method specified in descriptor is not supported")
```
kostko: Why is this commented out?

go/upgrade/api/api.go
```diff
@@ -0,0 +1,114 @@
...
+	Name string `json:"name"`
+	// Method is the upgrade method that should be used for this upgrade.
+	Method string `json:"method"`
+	// Identifier is a hash of the binary to be used for upgrading.
+	Identifier string `json:"identifier"`
+	// Epoch is the epoch at which the upgrade should happen.
+	Epoch epochtime.EpochTime `json:"epoch"`
+}
+
+func (d Descriptor) IsValid() bool {
```
kostko: Documentation comment here and below.

go/upgrade/migrations/migrations.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+)
+
+// Handler is the interface used by migration handlers.
+type Handler interface {
+	// StartupUpgrade is called by the upgrade manager to perform
+	// the node startup portion of the upgrade.
+	StartupUpgrade(*Context) error
+
+	// ConsensusUpgrade is called by the upgrade manager to perform
```
kostko: Describe what the interface contains (consensus backend-specific structure).

go/consensus/tendermint/abci/state.go
```diff
@@ -278,6 +278,11 @@ func (s *applicationState) OwnTxSigner() signature.PublicKey {
 	return s.ownTxSigner
 }
 
+func (s *applicationState) currentEpoch(ctx *Context) (epochtime.EpochTime, error) {
```
kostko: Looks like a good idea to have this exposed (also see #2625 which adds this too under a different name).

go/oasis-test-runner/oasis/log.go
```diff
@@ -66,3 +67,15 @@ func LogAssertNoMergeDiscrepancyDetected() log.WatcherHandlerFactory {
 func LogAssertPeerExchangeDisabled() log.WatcherHandlerFactory {
 	return LogAssertEvent(tendermint.LogEventPeerExchangeDisabled, "peer exchange not disabled")
 }
+
+// LogAssertUpgradeStartup returns a handler which checks whether a startup migration
+// handler was run based on JSON log output.
+func LogAssertUpgradeStartup() log.WatcherHandlerFactory {
+	return LogAssertEvent(upgrade.LogEventStartupUpgrade, "startup upgrade ran")
```
kostko: The error message is shown in case of an error so it should be `expected startup upgrade did not run`. Similar below.

go/upgrade/upgrade.go
```diff
@@ -0,0 +1,253 @@
...
+	handler migrations.Handler
+
+	logger *logging.Logger
+}
+
+func (u *upgradeManager) SubmitDescriptor(ctx context.Context, descriptor *api.Descriptor) error {
+	u.lock.Lock()
+	defer u.lock.Unlock()
+
+	u.pending = &api.PendingUpgrade{
```
kostko: So submitting a new descriptor just overwrites the old one? Should we perhaps have an explicit "cancel pending upgrade" action?

go/upgrade/upgrade.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	// Otherwise, the upgrade should proceed right now. Check that we're the right binary.
+	thisHash, err := hashSelf()
+	if err != nil {
+		return err
+	}
+
+	var upgraderHash hash.Hash
+	if err = upgraderHash.UnmarshalHex(u.pending.Descriptor.Identifier); err != nil {
```
kostko: Is there a reason why the `Identifier` field is not of type `hash.Hash`?

go/upgrade/upgrade.go
```diff
@@ -0,0 +1,253 @@
...
+		u.logger.Warn("performing consensus upgrade",
+			"name", u.pending.Descriptor.Name,
+			"submitted_by", u.pending.SubmittingVersion,
+			"version", u.pending.RunningVersion,
+			logging.LogEvent, api.LogEventConsensusUpgrade,
+		)
+		err := u.handler.ConsensusUpgrade(u.ctx, privateCtx)
+		if err == nil {
+			// Save the updated descriptor state.
+			err = u.flushDescriptor()
```
kostko: There is one tiny problem -- assume the upgrade succeeded, and flush returned without errors. Now if the node crashes before the upgrade block is complete, the block will not be committed but the "upgrade has been done" will be committed and so will not run on restart, causing a state corruption.

I think a safer approach would be to consider the current height as an indicator of whether the upgrade has been performed or not. The consensus upgrade must be executed exactly once at the block height after the epoch transition. If we are past that height, then the upgrade has been completed, but we may re-enter that same height multiple times in case we crash at an unlucky time.

go/upgrade/upgrade.go
```diff
@@ -0,0 +1,288 @@
...
+			logging.LogEvent, api.LogEventConsensusUpgrade,
+		)
+		return u.handler.ConsensusUpgrade(u.ctx, privateCtx)
+	}
+	return nil
+}
+
+func (u *upgradeManager) Close() {
+	err := u.flushDescriptor()
+	u.logger.Info(fmt.Sprintf("on upgrade manager close, pending ugprade is %v", u.pending), "err", err)
```
kostko: Log that `u.pending` as a structured parameter?

PR #2608, opened 2020-01-28
go/consensus/api: add SignedTransaction Hash method

peterjgilbert:


go/consensus/api/transaction/transaction.go
```diff
@@ -110,9 +111,19 @@ type SignedTransaction struct {
 	signature.Signed
 }
 
+// TxHash returns the cryptographic hash of the encoded transaction.
+func (s *SignedTransaction) TxHash() hash.Hash {
+	data := cbor.Marshal(s)
+	var txHash hash.Hash
+	txHash.FromBytes(data)
```
kostko: Just use `txHash.From(s)` as it is equivalent to the above.

PR #2610, opened 2020-01-28
cmd/signer: Add extract command, add entity signer pk check

david-yan:
Resolves https://github.com/oasislabs/oasis-core/issues/2609

- Adds `oasis-node signer extract` command
- Adds check to entity and signer public keys when running `entity.Load`


go/oasis-node/cmd/signer/signer.go
```diff
@@ -12,14 +17,44 @@ var (
 		Use:   "signer",
 		Short: "signer backend utilities",
 	}
+
+	extractCmd = &cobra.Command{
+		Use:   "extract",
```
ravenac95: can we make this `export-public`?

go/oasis-node/cmd/common/common.go
```diff
@@ -278,3 +278,17 @@ func LoadEntity(signerBackend string, entityDir string) (*entity.Entity, signatu
 
 	return entity.Load(entityDir, factory)
 }
+
+func ExtractEntity(signerBackend string, entityDir string) error {
```
kostko: Add documentation comment, consider renaming (also below you call this exporting) as this just creates an empty entity from a public key?

go/common/entity/entity.go
```diff
@@ -77,6 +77,11 @@ func Load(baseDir string, signerFactory signature.SignerFactory) (*Entity, signa
 		return nil, nil, err
 	}
 
+	if ent.ID != signer.Public() {
+		signer.Reset()
+		return nil, nil, fmt.Errorf("signer, entity public key mismatch")
```
kostko: Show the actual keys, e.g., `(entity: %s signer: %s)`.

.changelog/2609.feature.md
```diff
@@ -0,0 +1,5 @@
+Export Ledger entities
```
kostko: This doesn't seem ledger specific? I would rather says "Export signer public keys as empty entities" or something like that.

go/oasis-node/cmd/signer/signer.go
```diff
@@ -12,14 +17,44 @@ var (
 		Use:   "signer",
 		Short: "signer backend utilities",
 	}
+
+	exportCmd = &cobra.Command{
+		Use:   "export",
+		Short: "export the public key from signer to an entity",
```
kostko:
```suggestion
		Short: "export the public key from signer as an empty entity",
```

go/common/entity/entity.go
```diff
@@ -77,6 +77,11 @@ func Load(baseDir string, signerFactory signature.SignerFactory) (*Entity, signa
 		return nil, nil, err
 	}
 
+	if ent.ID != signer.Public() {
+		signer.Reset()
+		return nil, nil, fmt.Errorf("public key mismatch\nsigner: %s\nentity: %s", signer.Public(), ent.ID)
```
kostko: Don't use newlines in error messages, do something like:
```suggestion
		return nil, nil, fmt.Errorf("public key mismatch (signer: %s entity: %s)", signer.Public(), ent.ID)
```

go/common/entity/entity.go
```diff
@@ -77,6 +77,11 @@ func Load(baseDir string, signerFactory signature.SignerFactory) (*Entity, signa
 		return nil, nil, err
 	}
 
+	if ent.ID != signer.Public() {
```
kostko: While this works as the current implementation is a fixed-size array, please use `!ent.ID.Equal(signer.Public())` to be consistent with other parts.

go/oasis-node/cmd/signer/signer.go
```diff
@@ -12,14 +17,44 @@ var (
...
+	}
+	entityDir, err := cmdFlags.SignerDirOrPwd()
+	if err != nil {
+		logger.Error("failed to retrieve signer dir",
+			"err", err,
+		)
+		os.Exit(1)
+	}
+	if err := cmdCommon.ExportEntity(cmdFlags.Signer(), entityDir); err != nil {
+		logger.Error("failed to extract entity",
```
kostko:
```suggestion
		logger.Error("failed to export entity",
```

david-yan:
@kostko should all be done. Sorry for not squashing the commit message properly in my previous PR.

go/common/entity/entity.go
kostko: Hm, this still uses newlines?

david-yan: Oh. How embarrassing. I thought I committed this suggestion.

PR #2614, opened 2020-01-29
go/registry: Handle the old and busted node descriptor envelope

Yawning:
> Crawling in my skin
> These wounds, they will not heal

### DIFF
--- a/go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
+++ b/go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
@@ -1,0 +1,219 @@
+// Package fixgenesis implements the fix-genesis command.
+package fixgenesis
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"time"
+
+	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
+	"github.com/spf13/viper"
+
+	beacon "github.com/oasislabs/oasis-core/go/beacon/api"
+	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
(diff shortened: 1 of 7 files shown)


go/registry/api/sanity_check.go
```diff
@@ -109,10 +116,18 @@ func SanityCheckNodes(
 	}
 
 	for _, sn := range nodes {
+		isOldNode := isOldNode(sn, isGenesis)
+
 		// Open the node to get the referenced entity.
 		var n node.Node
-		if err := sn.Open(RegisterGenesisNodeSignatureContext, &n); err != nil {
-			return fmt.Errorf("registry: sanity check failed: unable to open signed node")
+		if isOldNode {
```
kostko: Should this be inverted?

Yawning: yes

go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
```diff
@@ -0,0 +1,219 @@
...
+	// The node descriptor signature envelope format in the registry has
+	// changed.  Convert to the new envelope.
+	//
+	// Note: Actually using the genesis document requires that some
+	// signature checks be disabled.
+	for _, osn := range oldReg.Nodes {
+		var nsn node.MultiSignedNode
+		nsn.MultiSigned.Signatures = []signature.Signature{osn.Signed.Signature}
+
+		// TODO: Someone that understands the issue can also fix the node
```
kostko: We just need to clear all flags and set the (new) validator flag (other nodes are not persisted anyway AFAIK).

PR #2618, opened 2020-01-30
go/worker/storage: Add configurable limits for storage operations

### LINKED ISSUE #1914: Enforce limits for storage operations
Currently the client can submit arbitrarily large batches of storage operations which can cause significant load on the storage node. We should review all operations and make sure that they have sensible limits in place.

* [ ] Limit the depth of forked roots to prevent increasing the number of passes that need to be done in `Finalize`. The depth should currently be limited to 2 as this is the maximum depth that is currently generated during normal operation.
* [ ] Limit the size of storage batches in `Apply`.
* [ ] Limit the amount of `Apply` operations in a round based on identity and role of caller (e.g., a txn scheduler node may only commit up to the number of compute committees I/O roots, a compute node may only commit 1 I/O root and 1 state root, a merge node may only merge).

*Estimated cost: 1 sprint*

kostko:
Fixes #1914


Yawning:
Errr, don't these need to be consensus backed?

kostko:
> Errr, don't these need to be consensus backed?

Hm, good point, once we have storage availability probes they should probably take these into account. I'll move them over to the runtime's storage config.

kostko:
Changed to consensus parameters.

PR #2619, opened 2020-01-30
go/cmd: add explicit --wait flag, prepend unix: if user forgets it

matevz:
PR for https://github.com/oasislabs/oasis-core/issues/2438:
* introduces new `--wait` flag which explicitly waits for gRPC address to become available,
* automatically prepends a `unix:` prefix, if the address is a local file name and prints a warning message.

### DIFF
--- a/go/oasis-node/cmd/common/grpc/grpc.go
+++ b/go/oasis-node/cmd/common/grpc/grpc.go
@@ -4,6 +4,8 @@ package grpc
 import (
 	"crypto/tls"
 	"errors"
+	"fmt"
+	"os"
 	"path/filepath"
 
 	"github.com/spf13/cobra"
@@ -12,6 +14,7 @@ import (
 	"google.golang.org/grpc"
 
 	cmnGrpc "github.com/oasislabs/oasis-core/go/common/grpc"
+	"github.com/oasislabs/oasis-core/go/common/logging"
 	"github.com/oasislabs/oasis-core/go/oasis-node/cmd/common"
 )
 
@@ -20,6 +23,8 @@ const (
 	CfgServerPort = "grpc.port"
 	// CfgDebugPort configures the remote address.
 	CfgAddress = "address"
+	// CfgWait waits for the remote address to become available.
+	CfgWait = "wait"
 
 	defaultAddress      = "unix:" + localSocketFilename
 	localSocketFilename = "internal.sock"
@@ -32,6 +37,8 @@ var (
 	ServerLocalFlags = flag.NewFlagSet("", flag.ContinueOnError)
 	// ClientFlags has the flags for a gRPC client.
 	ClientFlags = flag.NewFlagSet("", flag.ContinueOnError)
+
+	logger = logging.GetLogger("cmd/grpc")
 )
 
 // NewServerTCP constructs a new gRPC server service listening on
@@ -73,10 +80,19 @@ func NewServerLocal(installWrapper bool) (*cmnGrpc.Server, error) {
 func NewClient(cmd *cobra.Command) (*grpc.ClientConn, error) {
 	addr, _ := cmd.Flags().GetString(CfgAddress)
 
+	if _, err := os.Stat(addr); err == nil {
+		logger.Warn(fmt.Sprintf("'%s' is a file name. Assuming 'unix:%s'.", addr, addr))
+		addr = "unix:" + addr
+	}
+
+	opts := []grpc.DialOption{grpc.WithInsecure()}
+	if viper.GetBool(CfgWait) {
+		opts = append(opts, grpc.WithDefaultCallOptions(grpc.WaitForReady(true)))
+	}
+
 	conn, err := cmnGrpc.Dial(
 		addr,
-		grpc.WithInsecure(),
-		grpc.WithDefaultCallOptions(grpc.WaitForReady(true)),
+		opts...,
 	)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


go/common/grpc/grpc.go
```diff
@@ -494,6 +494,10 @@ func NewServer(config *ServerConfig) (*Server, error) {
 
 // Dial creates a client connection to the given target.
 func Dial(target string, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
+	if len(target) > 0 && target[0] == '/' {
```
kostko: This shouldn't be in the common gRPC code, rather it should be part of the gRPC wrapper for CLI tools, e.g. `go/oasis-node/cmd/common/grpc`.

go/common/grpc/grpc.go
```diff
@@ -494,6 +494,10 @@ func NewServer(config *ServerConfig) (*Server, error) {
...
+		logging.GetLogger("grpc").Warn(fmt.Sprintf("'%s' is not a valid socket name. Did you forget to prepend 'unix:'?", target))
```
kostko: Don't create a new logger each time, create/use a global instance instead.

go/oasis-node/cmd/common/grpc/grpc.go
```diff
@@ -73,6 +77,10 @@ func NewServerLocal(installWrapper bool) (*cmnGrpc.Server, error) {
 func NewClient(cmd *cobra.Command) (*grpc.ClientConn, error) {
 	addr, _ := cmd.Flags().GetString(CfgAddress)
 
+	if len(addr) > 0 && addr[0] == '/' {
```
Yawning: Why not `len(addr) > 0 && !strings.HasPrefix(addr, "unix:")`.  Though at this point, we could automagically try to fix-up the address by prepending the prefix for them if they forget.

matevz: Just `len(addr) > 0 && !strings.HasPrefix(addr, "unix:")` is not enough, because you can also pass a valid IP. Yeah, I will simply replace the addr, if it smells like a filename.

matevz: Changed, please review.

ptrus:
maybe we should just update the code that when using the CLI the `WaitReady` is set to False (i think always when you use the command you already expect the socket to be up - there could still be a flag to override the behavior in case there is a use case for the other )

matevz:
> maybe we should just update the code that when using the CLI the `WaitReady` is set to False (i think always when you use the command you already expect the socket to be up - there could still be a flag to override the behavior in case there is a use case for the other )

Currently I implemented that address is prepended with `unix:` automatically, if it smells like a filename. Do you think we should still drop out by default, if the socket is not available? I think @kostko suggested the wait behavior by default.

ptrus:
> suggested the wait behavior by default.

I think `WaitForReady` makes sense by default for non-CLI usages (e.g. test and other clients in the code - i think that's also what he ment). When using it via a CLI command - you usually want to interact with an already running node, and i think it makes more sense to default to `False`.

kostko:
> I think `WaitForReady` makes sense by default for non-CLI usages (e.g. test and other clients in the code - i think that's also what he ment)

Yeah this was mostly meant for tests. For manual CLI invocations removing `WaitForReady` is fine. With the new E2E test harness `WaitForReady` on CLI tools should no longer be needed, but please test it out first.

go/oasis-node/cmd/common/grpc/grpc.go
```diff
@@ -73,6 +77,11 @@ func NewServerLocal(installWrapper bool) (*cmnGrpc.Server, error) {
...
```
ptrus: this also doesn't cover the case where if path is relative (i still think changing the `WaitForReady` in cli cases is a more correct fix)

matevz: I included both now. If `addr` is the name of a local file, it prepends `unix:`. In either case, it never enables `WaitForReady` option. Let's see how tests react...

matevz:
> > I think `WaitForReady` makes sense by default for non-CLI usages (e.g. test and other clients in the code - i think that's also what he ment)
> 
> Yeah this was mostly meant for tests. For manual CLI invocations removing `WaitForReady` is fine. With the new E2E test harness `WaitForReady` on CLI tools should no longer be needed, but please test it out first.

I assume you meant go unit tests? Because e2e tests execute everything from CLI, so the order of actions (and waiting) is important there.

go/oasis-node/cmd/common/grpc/grpc.go
```diff
@@ -73,10 +78,15 @@ func NewServerLocal(installWrapper bool) (*cmnGrpc.Server, error) {
 func NewClient(cmd *cobra.Command) (*grpc.ClientConn, error) {
 	addr, _ := cmd.Flags().GetString(CfgAddress)
 
+	_, err := os.Stat(addr)
+	if !os.IsNotExist(err) {
```
ptrus: Does this now assume that the argument here is a socket and not an url? is this ok?

matevz:
If I remove `WaitForReady` flag, `txsource-transfer-short` and `runtime-dynamic`, e2e tests need to be fixed as well.

go/oasis-node/cmd/common/grpc/grpc.go
matevz: This was cleaned now. Please recheck.

kostko:
> If I remove `WaitForReady` flag, `txsource-transfer-short` and `runtime-dynamic` e2e tests need to be fixed as well.

You could easily add a new flag, e.g., `--wait`, which would be false by default but always used in tests.

matevz:
Both tests are now fixed. Please review.

go/oasis-test-runner/oasis/ias.go
```diff
@@ -30,7 +30,8 @@ func (ias *iasProxy) startNode() error {
 		debugDontBlameOasis().
 		debugAllowTestKeys().
 		grpcServerPort(ias.grpcPort).
-		iasDebugMock().
+		grpcWait()
```
kostko: This looks like it is missing a dot?

PR #2621, opened 2020-01-30
go: Re-enable signature verification disabled for migration

### LINKED ISSUE #2615: go/registry: Revert most of #2614
#2614 does a lot of things that are somewhere between sub-optimal and utterly horrible for security.  Revert all of the bad, once it is possible to do so.

Yawning:
Now that the migration has hopefully been done, re-enable all of the
signature verification that was disabled for the sake of allowing for
migration.

Fixes #2615


Yawning:
This should back out the horrific evil done in #2614, but it can't be merged till a release with said evil is made.

PR #2623, opened 2020-01-30
Flexible key manager policy signers

### LINKED ISSUE #2444: Flexible key manager policy signers
Currently the set of key manager policy signers is hard-coded. While the set must be compiled into the key manager and key manager client binaries for security reasons, we should make it so that people can easily build key managers with a different set. This is required so that people can develop their own runtimes which use their own key managers.

One way to do it is to have all key manager functionality exposed as a library crate and then the actual key manager implementation only needs to provide the set of key manager policy signers.

abukosek:
Closes #2444.


keymanager-api-common/src/api.rs
```diff
@@ -0,0 +1,247 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct TrustedPolicySigners {
+    /// Set of trusted signers.
+    pub signers: HashSet<OasisPublicKey>,
+    /// Threshold for determining if enough valid signatures are present.
+    pub threshold: usize,
+}
+
+impl TrustedPolicySigners {
+    pub fn new() -> Self {
```
kostko: Should this be a `Default::default()` impl instead?

keymanager-runtime/src/main.rs
```diff
@@ -47,6 +37,9 @@ fn main() {
                 _rpc_demux: &mut RpcDemux,
                 rpc: &mut RpcDispatcher,
                 _txn: &mut TxnDispatcher| {
+        // Initialize the set of trusted policy signers.
+        init_trusted_policy_signers();
+
         // Register RPC methods exposed via EnclaveRPC to remote clients.
```
kostko: Below (and `crate::methods::*`) seems like a lot of boilerplate that each key manager runtime will need anyway if it only wants to change the set of policy signers? Consider moving them to the common crate.

keymanager-runtime/api/src/api.rs
```diff
@@ -1,232 +1,4 @@
...
-
-/// Signed key manager access control policy.
-#[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct SignedPolicySGX {
-    pub policy: PolicySGX,
-    pub signatures: Vec<SignatureBundle>,
-}
+use oasis_core_runtime::runtime_api;
 
 runtime_api! {
```
kostko: Can this API also be moved to the common API crate?

keymanager-runtime/src/main.rs
abukosek: I did initially consider moving this to the api-common crate, but decided against it, since it's more flexible this way.  New key managers might want to run other code during initialization and/or add new API endpoints.

I got the impression that this is something we'd eventually want to support, so there's no point in moving this into the common crate now, only to move it back a few weeks later :)

kostko: Ok, let's leave it like this for now.

abukosek:
Thanks!  :)

PR #2625, opened 2020-01-30
Special rewards for block proposer

pro-wh:
- a larger portion of the fees
- an additional reward

### DIFF
--- a/.changelog/2625.breaking.md
+++ b/.changelog/2625.breaking.md
@@ -1,0 +1,4 @@
+Special rewards for block proposer
+
+- a larger portion of the fees
+- an additional reward
--- a/go/consensus/tendermint/abci/state.go
+++ b/go/consensus/tendermint/abci/state.go
@@ -61,8 +61,12 @@ type ApplicationState interface {
 	// GetBaseEpoch returns the base epoch.
 	GetBaseEpoch() (epochtime.EpochTime, error)
 
-	// GetEpoch returns current epoch at block height.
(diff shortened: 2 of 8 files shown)


go/consensus/tendermint/abci/mux.go
```diff
@@ -1000,11 +1000,31 @@ func (s *ApplicationState) GetBaseEpoch() (epochtime.EpochTime, error) {
...
 }
 
+// GetCurrentEpoch returns the epoch at the current block height,
+// applying all the weird legacy off-by-one rules we have.
+func (s *ApplicationState) GetCurrentEpoch(ctx context.Context) (epochtime.EpochTime, error) {
+	blockHeight := s.BlockHeight()
+	if blockHeight == 0 {
+		return epochtime.EpochInvalid, nil
+	} else if blockHeight == 1 {
+		currentEpoch, err := s.timeSource.GetEpoch(ctx, blockHeight)
```
kostko: Would something break if we make this consistent? Was this because of the deferred InitChain that we used to have?

pro-wh: https://github.com/oasislabs/oasis-core/commit/6b4965e1489ca37574ec999f7e3a8243ffe8aee5#diff-fd9bee7572e091f358b540058316ab3eR761 the `+1` stuff was added to fix something about queries

pro-wh: testing in #2634

PR #2626, opened 2020-01-31
Rewards and fees consensus parameters

### LINKED ISSUE #2624: Put some staking reward and fee distribution parameters in consensus parameters
##### SUMMARY

Currently they're hardcoded, things like reward "factors" and fee distribution "weights." But we have pretty good support for managing consensus parameters, so we ought to move them there.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

go/consensus/tendermint/apps/staking et al.

##### ADDITIONAL INFORMATION

 this section is complete with no entry

pro-wh:
closes #2624

no additional info

### DIFF
--- a/.changelog/2624.breaking.md
+++ b/.changelog/2624.breaking.md
@@ -1,0 +1,4 @@
+Rewards and fees consensus parameters
+
+Currently they're hardcoded, things like reward "factors" and fee distribution "weights."
+But we have pretty good support for managing consensus parameters, so we ought to move them there.
--- a/.changelog/2624.breaking.md
+++ b/.changelog/2624.breaking.md
@@ -1,4 +1,4 @@
 Rewards and fees consensus parameters
 
-Currently they're hardcoded, things like reward "factors" and fee distribution "weights."
+Previously things like reward "factors" and fee distribution "weights" were hardcoded.
 But we have pretty good support for managing consensus parameters, so we ought to move them there.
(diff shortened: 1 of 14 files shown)


.changelog/2624.breaking.md
```diff
@@ -0,0 +1,4 @@
+Rewards and fees consensus parameters
+
+Currently they're hardcoded, things like reward "factors" and fee distribution "weights."
```
kostko:
```suggestion
Previously things like reward "factors" and fee distribution "weights" were hardcoded.
```

go/staking/api/api.go
```diff
@@ -415,6 +408,18 @@ type ConsensusParameters struct {
 	DisableTransfers       bool                         `json:"disable_transfers,omitempty"`
 	DisableDelegation      bool                         `json:"disable_delegation,omitempty"`
 	UndisableTransfersFrom map[signature.PublicKey]bool `json:"undisable_transfers_from,omitempty"`
+
+	// The proportion of fees disbursed to entities of the nodes that voted for a block.
+	FeeWeightVote int64 `json:"fee_weight_vote,omitempty"`
+	// The proportion of fees disbursed to the entity of the node that proposed a block.
+	FeeWeightPropose int64 `json:"fee_weight_propose,omitempty"`
```
kostko: Add sanity checks that the sum of both weights is greater than zero (as otherwise the check in `disburseFees` will always fail).

pro-wh: I gotta think about this

pro-wh: yeah, that. and that FeeWeightVote should be nonnegative

PR #2632, opened 2020-01-31
oasis-node/cmd/stake: Make info subcommand tolerate invalid thresholds

tjanez:


go/oasis-test-runner/scenario/e2e/stake_cli.go
```diff
@@ -333,6 +338,28 @@ func (s *stakeCLIImpl) testAmendCommissionSchedule(childEnv *env.Env, cli *cli.H
...
+		return fmt.Errorf("scenario/e2e/stake: failed to query common token info: %s error: %w", b.String(), err)
+	}
+	// Check that subcommand reported warnings for invalid staking threshold kinds.
+	subCmdOutput := b.String()
+	thresholdKinds := []string{"compute", "storage"}
+	for _, kind := range thresholdKinds {
+		if !strings.Contains(subCmdOutput, fmt.Sprintf("invalid staking threshold kind: %s", kind)) {
+			return fmt.Errorf("scenario/e2e/stake: querying common token info should warn about invalid staking threshold for kind: %s", kind)
+		}
+	}
```
tjanez: I didn't know if this check is something we would want or is it testing a detail that is not important enough to be covered?

PR #2633, opened 2020-01-31
go/staking/api: Check if thresholds for all kinds are defined in genesis

tjanez:
Follow up to #2632.


.changelog/2633.feature.md
```diff
@@ -0,0 +1 @@
+go/staking/api: Check if thresholds for all kinds are defined in genesis.
```
kostko: Shouldn't this be bugfix?

tjanez: Yes, agreed.

PR #2634, opened 2020-02-01
go tendermint: mux offset block height consistently

pro-wh:
from https://github.com/oasislabs/oasis-core/pull/2625#discussion_r373370431

We've been using `blockHeight+1` for getting the epoch time except for on blockHeight=1. The hypothesis in this PR is that we don't need that special case.


pro-wh:
- one could reasonably guess that the cause of the non-plus-one was by mistake
- this hasn't affected anything that CI tests check
- at least one other member has agreed to see where this takes us

due to these factors, I'm merging this

PR #2642, opened 2020-02-04
Optionally require a deposit for registering a runtime

### LINKED ISSUE #2638: Registering a runtime should require a deposit
Registering a runtime should require a deposit. Withdrawing the deposit should result in the runtime being suspended in the next epoch.

kostko:
Fixes #2638 

Also changes the runtime descriptor to include an explicit `EntityID` and adds a stateful stake accumulator.


kostko:
Now also with dynamic runtime E2E tests.

kostko:
Yeah, I think adding a stateful stake accumulator would be best to solve this in the general case.

kostko:
Added stateful stake accumulator, PTAL.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -306,18 +312,37 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
...
+	if !params.DebugBypassStake {
+		stakeAcc, err = stakingState.NewStakeAccumulatorCache(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to create stake accumulator cache: %w", err)
 		}
 
+		claim := stakeClaimForNode(newNode.ID)
+		var thresholds []staking.ThresholdKind
+		if newNode.HasRoles(node.RoleComputeWorker | node.RoleKeyManager) {
+			thresholds = append(thresholds, staking.KindCompute)
```
Yawning: The key manager doesn't get it's own threshold?  Shouldn't it?

go/oasis-node/cmd/genesis/genesis.go
```diff
@@ -417,6 +421,8 @@ func AppendRootHashState(doc *genesis.Document, exports []string, l *logging.Log
 		RuntimeStates: make(map[common.Namespace]*registry.RuntimeGenesis),
 
 		Parameters: roothash.ConsensusParameters{
+			DebugDoNotSuspendRuntimes: viper.GetBool(cfgRoothashDebugDoNotSuspendRuntimes),
```
Yawning: Should these parameters require DontBlameOasis to set?

kostko: They do, genesis sanity check will fail otherwise.

go/consensus/tendermint/apps/registry/transactions.go
kostko: Yeah, good idea, I'll add it.

kostko: Done.

PR #2646, opened 2020-02-05
go/consensus/tendermint: Expose more P2P config params

### LINKED ISSUE #2597: Expose more consensus P2P configuration parameters
Currently we use a lot of [Tendermint's default P2P configuration parameters](https://github.com/tendermint/tendermint/blob/master/config/config.go#L560). We should consider changing these defaults and expose some of them in our configuration as well (e.g., max inbound/outbound peers, send/recv rates).

peterjgilbert:
Fixes #2597

TODO: Update sentry node config doc: https://docs.oasis.dev/operators/sentry-node.html#configuring-the-oasis-sentry-node

### DIFF
--- a/go/consensus/tendermint/tendermint.go
+++ b/go/consensus/tendermint/tendermint.go
@@ -80,13 +80,21 @@ const (
 	CfgSentryUpstreamAddress = "tendermint.sentry.upstream_address"
 
 	// CfgP2PPersistentPeer configures tendermint's persistent peer(s).
-	CfgP2PPersistentPeer = "tendermint.persistent_peer"
+	CfgP2PPersistentPeer = "tendermint.p2p.persistent_peer"
 	// CfgP2PDisablePeerExchange disables tendermint's peer-exchange (Pex) reactor.
-	CfgP2PDisablePeerExchange = "tendermint.disable_peer_exchange"
+	CfgP2PDisablePeerExchange = "tendermint.p2p.disable_peer_exchange"
 	// CfgP2PSeeds configures tendermint's seed node(s).
-	CfgP2PSeed = "tendermint.seed"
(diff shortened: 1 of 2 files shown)


go/consensus/tendermint/tendermint.go
```diff
@@ -1338,6 +1350,10 @@ func init() {
 	Flags.StringSlice(CfgP2PPersistentPeer, []string{}, "Tendermint persistent peer(s) of the form ID@ip:port")
 	Flags.Bool(CfgP2PDisablePeerExchange, false, "Disable Tendermint's peer-exchange reactor")
 	Flags.Bool(CfgP2PSeedMode, false, "run the tendermint node in seed mode")
+	Flags.Int(CfgP2PMaxNumInboundPeers, 40, "Max number of inbound peers")
+	Flags.Int(CfgP2PMaxNumOutboundPeers, 10, "Max number of outbound peers (excluding persistent peers)")
```
kostko: Based on feedback we should increase this default as 10 seems low?

go/consensus/tendermint/tendermint.go
```diff
@@ -87,6 +87,14 @@ const (
 	CfgP2PSeed = "tendermint.seed"
 	// CfgP2PSeedMode enables the tendermint seed mode.
 	CfgP2PSeedMode = "tendermint.seed_mode"
+	// CfgP2PMaxNumInboundPeers configures the max number of inbound peers.
+	CfgP2PMaxNumInboundPeers = "tendermint.max_num_inbound_peers"
```
kostko: Namespace all of these under `tendermint.p2p.*`.

go/consensus/tendermint/tendermint.go
peterjgilbert: Increased default to 20

PR #2648, opened 2020-02-06
go/registry: Enable non-genesis runtime registrations by default

### LINKED ISSUE #2445: Allow anyone to register a key manager
Related to #2406.

Currently only the key manager operator entity can register a key manager. We should allow anyone to register a key manager subject to certain provisions like paying gas fees and maybe also depositing stake in escrow.

kostko:
See #2406 
Fixes #2445 (together with #2642)

### DIFF
--- a/go/consensus/tendermint/apps/registry/genesis.go
+++ b/go/consensus/tendermint/apps/registry/genesis.go
@@ -29,18 +29,6 @@ func (app *registryApplication) InitChain(ctx *abci.Context, request types.Reque
 	state := registryState.NewMutableState(ctx.State())
 	state.SetConsensusParameters(&st.Parameters)
 
-	if st.Parameters.KeyManagerOperator.IsValid() {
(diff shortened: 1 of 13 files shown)


go/consensus/tendermint/apps/registry/genesis.go
```diff
@@ -29,18 +29,6 @@ func (app *registryApplication) InitChain(ctx *abci.Context, request types.Reque
 	state := registryState.NewMutableState(ctx.State())
 	state.SetConsensusParameters(&st.Parameters)
 
-	if st.Parameters.KeyManagerOperator.IsValid() {
```
Yawning: I'm still totally against this.

kostko: Added a separate consensus parameter to allow explicitly disabling key manager runtime registrations outside genesis. Other than that #2642 also introduces a way to specify a staking threshold for key manager runtime registrations.

Issue #1062, opened 2018-10-10
E2E tests with multiple runtimes


kostko:
We should run E2E tests with multiple runtimes to ensure that all components work correctly in the presence of multiple runtimes.

matevz:
`multiple-runtimes` e2e test was added for this in #2529 .

PR #2650, opened 2020-02-06
send expected epoch along when calling SubmitTx

matevz:
PR for #1062: 
- runtime/client: add expectedEpochNumber parameter to SubmitTx call,
- txnscheduler/committee/node: check whether expectedEpochNumber matches the local one,
- executor/node, merge/node: abort, if state change detected,
- e2e/multiple-runtimes:
  - reduce all group sizes to 1 with no backups,
  - use epochTime mock,
  - add new numComputeWorkers parameter,
- cmd/node/node.go, common/commmittee/node.go, common/worker.go, runtime/client/client.go, worker/common/committee/group.go: access `RootHash`, `Scheduler`, `Beacon`, `KeyManager`, `EpochTime`, and `Registry` fields indirectly using `consensus` member.

### DIFF
--- a/.changelog/2650.breaking.md
+++ b/.changelog/2650.breaking.md
@@ -1,0 +1,8 @@
+Send and check expected epoch number during transaction execution
+
+Stress tests revealed some race conditions during transaction execution when
+there is an epoch transition. Runtime client now sends `expectedEpochNumber`
+parameter in `SubmitTx` call. The transaction scheduler checks whether the
+expected epoch matches its local one. Additionally, if state transition occurs
+during transaction execution, Executor and Merge committee correctly abort the
+transaction.
--- a/.changelog/2650.internal.1.md
+++ b/.changelog/2650.internal.1.md
@@ -1,0 +1,12 @@
+Replace redundant fields with `Consensus` accessors
+
+`Backend` in `go/consensus/api` contains among others accessors for
+`Beacon`, `EpochTime`, `Registry`, `RootHash`, `Scheduler`, and
+`KeyManager`. Use those instead of direct references. The following
+structs were affected:
+
+- `Node` in `go/cmd/node`,
+- `Node` in `go/common/commmittee`,
+- `Worker` in `go/common`,
+- `clientCommon` in `go/runtime/client`,
+- `Group` in `go/worker/common/committee`.
--- a/.changelog/2650.internal.2.md
+++ b/.changelog/2650.internal.2.md
@@ -1,0 +1,1 @@
+e2e/multiple-runtimes: Enable `EpochtimeMock`, add `numComputeWorkers`
--- a/go/common/version/version.go
+++ b/go/common/version/version.go
@@ -62,7 +62,7 @@ var (
 
 	// CommitteeProtocol versions the P2P protocol used by the
 	// committee members.
-	CommitteeProtocol = Version{Major: 0, Minor: 7, Patch: 0}
+	CommitteeProtocol = Version{Major: 0, Minor: 8, Patch: 0}
 
 	// ConsensusProtocol versions all data structures and processing used by
 	// the epochtime, beacon, registry, roothash, etc. modules that are
--- a/go/oasis-node/cmd/node/node.go
+++ b/go/oasis-node/cmd/node/node.go
@@ -12,7 +12,6 @@ import (
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
 
-	beacon "github.com/oasislabs/oasis-core/go/beacon/api"
 	"github.com/oasislabs/oasis-core/go/common"
 	"github.com/oasislabs/oasis-core/go/common/crash"
 	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
@@ -44,7 +43,6 @@ import (
 	"github.com/oasislabs/oasis-core/go/oasis-node/cmd/common/tracing"
 	"github.com/oasislabs/oasis-core/go/oasis-node/cmd/debug/supplementarysanity"
 	registryAPI "github.com/oasislabs/oasis-core/go/registry/api"
-	roothash "github.com/oasislabs/oasis-core/go/roothash/api"
 	runtimeClient "github.com/oasislabs/oasis-core/go/runtime/client"
 	runtimeClientAPI "github.com/oasislabs/oasis-core/go/runtime/client/api"
 	runtimeRegistry "github.com/oasislabs/oasis-core/go/runtime/registry"
@@ -109,18 +107,11 @@ type Node struct {
 
 	Consensus consensusAPI.Backend
 
-	Upgrader   upgradeAPI.Backend
-	Genesis    genesisAPI.Provider
-	Identity   *identity.Identity
-	Beacon     beacon.Backend
-	Epochtime  epochtime.Backend
-	Registry   registryAPI.Backend
-	RootHash   roothash.Backend
-	Scheduler  scheduler.Backend
-	Sentry     sentryAPI.Backend
-	Staking    stakingAPI.Backend
-	IAS        iasAPI.Endpoint
-	KeyManager keymanagerAPI.Backend
+	Upgrader upgradeAPI.Backend
+	Genesis  genesisAPI.Provider
+	Identity *identity.Identity
+	Sentry   sentryAPI.Backend
+	IAS      iasAPI.Endpoint
 
 	RuntimeRegistry runtimeRegistry.Registry
 	RuntimeClient   runtimeClientAPI.RuntimeClient
@@ -191,9 +182,9 @@ func (n *Node) initBackends() error {
 
 	// Initialize and register the internal gRPC services.
 	grpcSrv := n.grpcInternal.Server()
-	scheduler.RegisterService(grpcSrv, n.Scheduler)
-	registryAPI.RegisterService(grpcSrv, n.Registry)
-	stakingAPI.RegisterService(grpcSrv, n.Staking)
+	scheduler.RegisterService(grpcSrv, n.Consensus.Scheduler())
+	registryAPI.RegisterService(grpcSrv, n.Consensus.Registry())
+	stakingAPI.RegisterService(grpcSrv, n.Consensus.Staking())
 	consensusAPI.RegisterService(grpcSrv, n.Consensus)
 
 	cmdCommon.Logger().Debug("backends initialized")
@@ -235,13 +226,10 @@ func (n *Node) initWorkers(logger *logging.Logger) error {
 		dataDir,
 		compute.Enabled() || workerStorage.Enabled() || workerKeymanager.Enabled(),
 		n.Identity,
-		n.RootHash,
-		n.Registry,
-		n.Scheduler,
 		n.svcTmnt,
 		n.P2P,
 		n.IAS,
-		n.KeyManager,
+		n.Consensus.KeyManager(),
 		n.RuntimeRegistry,
 		genesisDoc,
 	)
@@ -259,8 +247,8 @@ func (n *Node) initWorkers(logger *logging.Logger) error {
 	// Initialize the registration worker.
 	n.RegistrationWorker, err = registration.New(
 		dataDir,
-		n.Epochtime,
-		n.Registry,
+		n.Consensus.EpochTime(),
+		n.Consensus.Registry(),
 		n.Identity,
 		n.svcTmnt,
 		n.P2P,
@@ -287,7 +275,7 @@ func (n *Node) initWorkers(logger *logging.Logger) error {
 		n.CommonWorker,
 		n.IAS,
 		n.RegistrationWorker,
-		n.KeyManager,
+		n.Consensus.KeyManager(),
 	)
 	if err != nil {
 		return err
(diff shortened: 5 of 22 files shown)


go/worker/compute/txnscheduler/service.go
```diff
@@ -15,6 +17,18 @@ func (w *Worker) SubmitTx(ctx context.Context, rq *api.SubmitTxRequest) (*api.Su
 		return nil, api.ErrUnknownRuntime
 	}
 
+	// TODO: Is this the right way to fetch current epoch?
+	epoch, err := w.commonWorker.Consensus.EpochTime().GetEpoch(ctx, consensus.HeightLatest)
```
kostko: No this is not the right way as the consensus may have already processed the epoch transition but the worker has not.

First, move the check into the per-runtime `QueueCall` as this will give you access to the correct `committee.Node` instance (these are per-runtime as different runtime workers may have processed different epochs).

Then use the `n.commonNode.Group.GetEpochSnapshot()` to get the current epoch. Note that the epoch number is currently not exposed, but adding it should be trivial.

go/runtime/client/client.go
```diff
@@ -93,16 +93,34 @@ func (c *runtimeClient) doSubmitTxToLeader(
 	defer close(submitCtx.closeCh)
 
 	op := func() error {
-		_, err := client.SubmitTx(submitCtx.ctx, req)
+		// Populate request with the current epoch.
+		// TODO: Is this the right way to fetch current epoch?
+		epochTime, err := c.common.consensus.EpochTime().GetEpoch(context.Background(), consensus.HeightLatest)
```
kostko: This is also not the right way to fetch the client's current epoch as you need to check what epoch the client block watcher has actually processed. This information should be made available in the `blockWatcher` (updated in `watch()` when a block kind is `block.EpochTransition`). The watcher instance is available in `SubmitTx` which calls `doSubmitTxToLoader`.

go/worker/compute/txnscheduler/committee/node.go
```diff
@@ -204,13 +205,18 @@ func (n *Node) CheckTx(ctx context.Context, call []byte) error {
...
-func (n *Node) QueueCall(ctx context.Context, call []byte) error {
+func (n *Node) QueueCall(ctx context.Context, expectedEpoch epochtime.EpochTime, call []byte) error {
 	// Check if we are a leader. Note that we may be in the middle of a
 	// transition, but this shouldn't matter as the client will retry.
 	if !n.commonNode.Group.GetEpochSnapshot().IsTransactionSchedulerLeader() {
 		return api.ErrNotLeader
 	}
 
+	// Check if expected client's epoch matches the current worker's one.
+	if n.commonNode.Group.GetEpochSnapshot().EpochTime != expectedEpoch {
```
kostko: No need to create two snapshots for the two ifs, reuse the same one?

go/worker/common/committee/group.go
```diff
@@ -445,6 +455,7 @@ func (g *Group) GetEpochSnapshot() *EpochSnapshot {
 	}
 
 	s := &EpochSnapshot{
+		EpochTime:    g.activeEpoch.epochTime,
```
kostko: Make this private and add accessor.

go/worker/common/committee/group.go
```diff
@@ -78,6 +82,9 @@ type EpochSnapshot struct {
 
 	executorCommitteeID hash.Hash
 
+	// EpochTime is the epoch time.
+	EpochTime api.EpochTime
```
kostko: Maybe call it `epochNumber`.

go/runtime/client/client.go
```diff
@@ -138,6 +144,12 @@ func (c *runtimeClient) SubmitTx(ctx context.Context, request *api.SubmitTxReque
 	}
 	c.Unlock()
 
+	req := &txnscheduler.SubmitTxRequest{
+		RuntimeID:     request.RuntimeID,
+		ExpectedEpoch: watcher.EpochLatest,
```
kostko: This is racy, watcher runs in a separate goroutine. Make `EpochLatest` private and add accessor that performs synchronization.

go/runtime/client/watcher.go
```diff
@@ -127,8 +129,19 @@ func (w *blockWatcher) watch() {
 			w.watched[*newWatch.id] = newWatch
 			if conn := w.committeeClient.GetConnection(); conn != nil {
 				client := txnscheduler.NewTransactionSchedulerClient(conn)
+				var epochNumber api.EpochTime
+				epochNumber, err = w.common.consensus.EpochTime().GetEpoch(w.common.ctx, height)
```
kostko: In this case `height` will always be zero, so this will actually be incorrect.

You should create a function-local variable called `currentEpoch` (outside the for loop) and update it only below after processing an epoch transition block. Here, you just pass `currentEpoch` instead of doing any other queries.

go/runtime/client/watcher.go
```diff
@@ -166,8 +179,18 @@ func (w *blockWatcher) watch() {
 
 				// Tell every client to resubmit as nothing further can be finalized by this committee.
 				for key, watch := range w.watched {
+					var epochNumber api.EpochTime
+					epochNumber, err = w.common.consensus.EpochTime().GetEpoch(w.common.ctx, height)
```
kostko: You should create a function-local variable called `currentEpoch` and update it once above (e.g., immediately after entering the epoch transition block `if`).

go/runtime/client/client.go
```diff
@@ -179,6 +191,7 @@ func (c *runtimeClient) SubmitTx(ctx context.Context, request *api.SubmitTxReque
 			continue
 
 		case resp, ok = <-respCh:
+			req.ExpectedEpochNumber = resp.epochNumber
```
kostko: This seems problematic, you are creating a shared reference `req`, mutating it here and then passing the same reference to potentially different goroutines below (like 217, `go c.doSubmitTxToLeader`). This can cause problems, so instead only create a new `req` each time just before calling `go c.doSubmitTxToLoader`.

go/runtime/client/watcher.go
```diff
@@ -149,14 +155,23 @@ func (w *blockWatcher) watch() {
 
 		// Find a new committee leader.
 		if current.Header.HeaderType == block.EpochTransition || !gotFirstBlock {
-			if err := w.committeeWatcher.EpochTransition(w.common.ctx, height); err != nil {
+			currentEpochNumber, err = w.common.consensus.EpochTime().GetEpoch(w.common.ctx, height)
```
kostko: A minor nit - in case of an error (should never happen, but hey) this will mutate the `currentEpochNumber` with a potentially invalid epoch. It should instead not change the `currentEpochNumber` when an error occurs.

.changelog/2650.breaking.md
```diff
@@ -0,0 +1,8 @@
+send and check expected epoch number during transaction execution
```
tjanez: Start the summary with the title case, i.e. `send` -> `Send`.

.changelog/2650.internal.1.md
```diff
@@ -0,0 +1,9 @@
+replace Roothash, Scheduler and Registry fields with Consensus accessors
```
tjanez: Same here.

.changelog/2650.internal.1.md
```diff
@@ -0,0 +1,9 @@
...
+
+ConsensusAPI.Backend contains accessors for RootHash, Scheduler, and Registry.
```
tjanez: Since Change Log is Markdown, use `` ` ``s when referencing types, variables or file paths.

.changelog/2650.internal.1.md
```diff
@@ -0,0 +1,9 @@
...
+Use those instead of direct references. The following structs were affected:
+- Node in cmd/node/node.go,
```
tjanez: Perhaps even better than referencing file paths would be to reference Go packages here.

.changelog/2650.internal.2.md
```diff
@@ -0,0 +1 @@
+e2e/multiple-runtimes: enable `EpochtimeMock`, add `numComputeWorkers`
```
tjanez: Same as above, use `Enable ...`.

go/common/version/version.go
```diff
@@ -70,7 +70,7 @@ var (
 	//
 	// NOTE: Any change in the major or minor versions are considered
 	//       breaking changes for the protocol.
-	ConsensusProtocol = Version{Major: 0, Minor: 23, Patch: 0}
+	ConsensusProtocol = Version{Major: 0, Minor: 24, Patch: 0}
```
kostko: Why does this break the consensus protocol, I don't see any changes here? It does change the committee protocol though.

go/oasis-node/cmd/node/node.go
```diff
@@ -111,9 +110,6 @@ type Node struct {
 	Identity   *identity.Identity
 	Beacon     beacon.Backend
 	Epochtime  epochtime.Backend
-	Registry   registryAPI.Backend
```
kostko: If you're at it, why only some of the backends? What about beacon, epochtime, staking and key manager?

go/oasis-node/cmd/node/node.go
```diff
@@ -723,7 +710,7 @@ func newNode(testNode bool) (*Node, error) {
 	controlAPI.RegisterService(node.grpcInternal.Server(), node.NodeController)
 	if flags.DebugDontBlameOasis() {
 		// Initialize and start the debug controller if we are in debug mode.
-		node.DebugController = control.NewDebug(node.Epochtime, node.Registry)
+		node.DebugController = control.NewDebug(node.Epochtime, node.Consensus.Registry())
```
kostko: Could just pass consensus and the controller could get both from there.

go/runtime/client/client.go
```diff
@@ -100,6 +95,9 @@ func (c *runtimeClient) doSubmitTxToLeader(
 		if err == txnscheduler.ErrNotLeader || status.Code(err) == codes.Unavailable {
 			return err
 		}
+		if err == txnscheduler.ErrEpochNumberMismatch {
```
kostko: Use `errors.Is(err, txnscheduler.ErrEpochNumberMismatch)` instead (also above).

go/worker/compute/executor/committee/node.go
```diff
@@ -943,6 +945,11 @@ func (n *Node) worker() {
 				func() {
 					n.commonNode.CrossNode.Lock()
 					defer n.commonNode.CrossNode.Unlock()
+
+					if state, ok := n.state.(StateProcessingBatch); !ok || state.done != processingDoneCh {
+						n.logger.Warn("state snapshot changed unexpectedly, aborting")
```
kostko: No need to warn here, maybe log at debug level or just remove as this is expected to happen.

go/worker/compute/executor/committee/node.go
```diff
@@ -953,6 +960,11 @@ func (n *Node) worker() {
 			func() {
 				n.commonNode.CrossNode.Lock()
 				defer n.commonNode.CrossNode.Unlock()
+
+				if state, ok := n.state.(StateProcessingBatch); !ok || state.done != processingDoneCh {
+					n.logger.Warn("state snapshot changed unexpectedly, aborting")
```
kostko: No need to warn here, maybe log at debug level or just remove as this is expected to happen.

go/worker/compute/executor/committee/node.go
```diff
@@ -943,6 +945,11 @@ func (n *Node) worker() {
 				func() {
 					n.commonNode.CrossNode.Lock()
 					defer n.commonNode.CrossNode.Unlock()
+
+					if state, ok := n.state.(StateProcessingBatch); !ok || state.done != processingDoneCh {
```
kostko: Add comment why this check is needed (to avoid stale events). Also elsewhere.

go/worker/compute/merge/committee/node.go
```diff
@@ -733,7 +734,8 @@ func (n *Node) worker() {
 				defer n.commonNode.CrossNode.Unlock()
 
 				state, ok := n.state.(StateWaitingForResults)
-				if !ok {
+				if !ok || state.timer.C != timerCh {
+					n.logger.Warn("state snapshot changed unexpectedly, aborting")
```
kostko: No need to warn here, maybe log at debug level or just remove as this is expected to happen.

go/worker/compute/merge/committee/node.go
```diff
@@ -746,6 +748,11 @@ func (n *Node) worker() {
 				n.commonNode.CrossNode.Lock()
 				defer n.commonNode.CrossNode.Unlock()
 
+				if state, ok := n.state.(StateProcessingMerge); !ok || state.doneCh != mergeDoneCh {
+					n.logger.Warn("state snapshot changed unexpectedly, aborting")
```
kostko: No need to warn here, maybe log at debug level or just remove as this is expected to happen.

go/worker/compute/txnscheduler/committee/node.go
```diff
@@ -2,12 +2,12 @@ package committee
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"sync"
 	"time"
 
 	"github.com/opentracing/opentracing-go"
+	"github.com/pkg/errors"
```
kostko: Don't introduce a new dependency on a package we are trying to remove. Put the other errors back.

matevz: Ooops, some IDE magic again :thinking:

.changelog/2650.internal.1.md
matevz: Good idea. I was just thinking about this...

PR #2654, opened 2020-02-06
additional genesis signature check disablement

pro-wh:
The Tendermint genesis one is encountered in a dry run.

The seed one is speculative.


Yawning:
I mean, I guess?  No guarantees that the system is going to be secure in any way with the number of checks that are being removed.

PR #2662, opened 2020-02-10
Add markdownlint

### LINKED ISSUE #2572: Add Markdown lint CI step
##### SUMMARY
Add Markdown lint CI step to ensure our documentation in `docs/` and Change Log/Change Log fragments are in good shape.

##### ISSUE TYPE
- Ops

tjanez:
Closes #2572.

TODO:
- [x] Use `gitlint` to lint the Change Log fragments.

### DIFF
--- a/.changelog/README.md
+++ b/.changelog/README.md
@@ -1,14 +1,14 @@
 # Change Log fragments
 
 This directory collects Change Log fragments:
-short files that each contain a snippet of MarkDown formatted text that will be
-assembled using [towncrier] to form the [Change Log] section for the next
+short files that each contain a snippet of MarkDown formatted text that will
+be assembled using [towncrier] to form the [Change Log] section for the next
 release.
 
 ## Description
 
-A Change Log fragment should be a description of aspects of the change (if any)
-that are relevant to users.
+A Change Log fragment should be a description of aspects of the change (if
+any) that are relevant to users.
 
 _NOTE: This could be very different from the commit message and pull request
 description, which are a description of the change as relevant to the people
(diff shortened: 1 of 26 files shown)


.github/workflows/ci-lint.yml
```diff
@@ -39,3 +43,7 @@ jobs:
         env:
           BASE_BRANCH: ${{ github.base_ref }}
         if: github.event_name == 'pull_request'
+      - name: Lint Markdown files
+        run: |
+          make lint-md
+          make changelog
```
kostko: Should this be `make lint-changelog` instead?

.markdownlint.yml
```diff
@@ -0,0 +1,18 @@
...
+# For more information, see:
+# https://github.com/DavidAnson/markdownlint#optionsconfig.
+
+# Enable all rules.
+default: true
+
+line-length:
+  # Line length checking is not strict by default.
+  strict: true
+  line_length: 80
```
kostko: Any specific reason for 80?

.changelog/README.md
```diff
@@ -24,13 +24,13 @@ The description could use one of the following two formats:
   ```text
   Remove staking-related roothash messages.
```
kostko: Below you removed dots from the generated changelog, but both examples here have them (which is also the reason why I always add a dot).

tjanez: Good catch :slightly_smiling_face:!

While assembling Change Logs for the latest releases I've seen most developers did not follow this style (intentionally or unintentionally), so I've decided to suggest to adopt the Git commit message style for Change Log fragments.
It should be familiar to developers and prevent confusion between requiring a Change Log fragment's subject line to end with period and requiring a Git commit message's subject line **not** to end with a period.

Also, the new style is perhaps nicer in the assembled Change Log since subject line are always amended with links to issues/pull requests in parentheses.

I've added the relevant commit describing this change.

.markdownlint.yml
tjanez: Nothing in particular, 80 is something I'm most used to.

PR #2664, opened 2020-02-11
go tendermint badger: don't scan whole keyspace

pro-wh:
when we run off the end of a range iteration

but now I need to try rewriting the whole badger iterator to make sure I understand it

### DIFF
--- a/go/consensus/tendermint/db/badger/badger.go
+++ b/go/consensus/tendermint/db/badger/badger.go
@@ -338,22 +338,22 @@ func (it *badgerDBIterator) Next() {
 		panic("Next with invalid iterator")
 	}
 
-	for {
-		it.iter.Next()
-		if !it.iter.Valid() {
-			break
-		}
-
-		item := it.iter.Item()
-		k := fromDBKeyNoCopy(item.KeyCopy(nil))
-		if dbm.IsKeyInDomain(k, it.start, it.end) {
-			it.current.item = item
-			it.current.key = k
-			it.current.value = nil // Copy done on access.
-			return
-		}
+	it.iter.Next()
+	if !it.iter.Valid() {
+		it.Close()
+		return
 	}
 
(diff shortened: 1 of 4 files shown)


go/consensus/tendermint/db/badger/badger.go
```diff
@@ -338,22 +338,22 @@ func (it *badgerDBIterator) Next() {
 		panic("Next with invalid iterator")
 	}
 
-	for {
```
kostko: Wow, why did we keep scanning before? This looks like it kept scanning until reaching the end of the database because once the key was no longer inside the domain, we just kept going until the underlying iterator was valid.

Looking at the old bolt implementation it seems like the same was done there. Where did we get those semantics from, looking at other TM-DB backends, this doesn't seem to be the case? cc @Yawning

Yawning: I don't remember, sorry.

Yawning:
Why do you want to merge this into `stable/20.1.x` instead of `master`?

kostko:
> Why do you want to merge this into stable/20.1.x instead of master?

Just wanted to say the same, so this should go into master first. We can backport it later.

go/consensus/tendermint/db/badger/badger.go
pro-wh: key suspect is this https://github.com/oasislabs/oasis-core/blob/v20.1.2/go/consensus/tendermint/db/badger/badger.go#L302

I don't yet understand the case where we would need that, and I want to try to write an implementation that doesn't rely on that. But if it's true that we need to scan into the beginning of the range, that's why we'd need this loop.

pro-wh:
I'll rebase on master

pro-wh:
Ah, coverage of the changed routines is pretty good too

PR #2665, opened 2020-02-11
Add sanity checks for stake accumulator state integrity

kostko:


go/registry/api/api.go
```diff
@@ -1285,3 +1286,51 @@ var DefaultGasCosts = transaction.Costs{
...
+}
+
+// StakeThresholdsForRuntime returns the staking thresholds for the given runtime.
+func StakeThresholdsForRuntime(rt *Runtime) (thresholds []staking.ThresholdKind) {
+	switch rt.Kind {
+	case KindCompute:
+		thresholds = append(thresholds, staking.KindRuntimeCompute)
+	case KindKeyManager:
+		thresholds = append(thresholds, staking.KindRuntimeKeyManager)
+	default:
```
Yawning: I know registrations that don't have the expected kind are rejected, but, not handling the default case seems sub-optimal.

kostko: Yeah, should never happen at this point so I added a panic.

PR #2667, opened 2020-02-11
go tendermint db: restructure badger iterator

pro-wh:
Here's an alternate implementation.

- Responsibility for correct seeking is moved down to BadgerDB.
- Responsibility for caching keys and values is moved up or down.
- Bounds checking is moved to Valid.
- Methods now don't call Close internally. Remember to call it yourself and don't abuse Valid.


pro-wh:
alright, gonna merge into #2664

PR #2673, opened 2020-02-14
Makefile: build-runtimes explicitly set target-dir

ptrus:
This is a workaround needed to avoid recompilations in newer cargo nightly
builds.

See https://github.com/oasislabs/oasis-core/issues/2579 for more details.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,12 +18,15 @@ build-tools:
 	@# Suppress "binary already exists" error by redirecting stderr and stdout to /dev/null.
 	@cargo install --path tools >/dev/null 2>&1 || true
 
+# NOTE: We epxplictly set target-dir as a workaround to avoid recompilations in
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -22,8 +22,8 @@ build-runtimes:
 	@for e in $(RUNTIMES); do \
 		$(ECHO) "$(MAGENTA)*** Building runtime: $$e...$(OFF)"; \
 		(cd $$e && \
-			cargo build --target x86_64-fortanix-unknown-sgx && \
-			cargo build && \
+			cargo build --target-dir target/x86_64-fortanix-unknown-sgx --target x86_64-fortanix-unknown-sgx && \
```
kostko: Add comment in the makefile why this is needed and reference the relevant issue.

PR #2675, opened 2020-02-14
Make node TLS certificates totally ephemeral

### LINKED ISSUE #2098: Make node TLS certificates totally ephemeral
Node TLS certificates used for authenticating gRPC connections should be made totally ephemeral (e.g., regenerated periodically while the node is running) and validity of such certificates should be shortened (e.g., regenerate every X epochs, valid for ~X+1 epochs).

Care needs to be taken to not make the node unreachable until the next epoch transition as currently the other nodes will only update their local cache on an epoch transition.

The transition could look like this:
* Certificate K is in use and registered.
* Epoch X-1: Certificate K' is generated but not yet in use (K is still in use). Node registers `{K, K'}` as valid certificates.
* Epoch X: Certificate K' is now in use, K is erased. Node registers `{K'}` as a valid certificate.

This would give enough time for propagating the updated certificate. Care must be taken to properly propagate these changes when using Sentry gRPC nodes.

For dynamically supplying certificates to the gRPC server we can use the [`GetCertificate` callback in the `tls.Config` structure](https://golang.org/pkg/crypto/tls/#Config).

abukosek:
Closes #2098.

TODO:
- [x] Add a flag to set node TLS certificate rotation interval.
- [x] Figure out why the storage client unit tests fail with `storage client: failed to write to any storage node` (I used `GetCertificate` where I should have used `GetClientCertificate` :man_facepalming: ).
- [x] Fix sentry nodes (these need to have persistent TLS certs, so no rotation for you).
- [x] Fix e2e tests (need to take NextCertificate into account when comparing nodes in the registry_cli test).
- [x] Enable cert rotation in e2e tests.
- [x] Fix e2e tests again (tests using the IAS proxy should also not do rotation).
- [x] Put the TLSSigner stuff back in (I temporarily removed this before to make debugging easier).
- [x] Write changelog fragment.
- [x] Fix the gRPC proxy so it dials connections itself rather than relying on an externally-established connection.
- [x] Upstream node should send its TLS certs to the sentry node on init and on every rotation.
- [x] Enable TLS cert rotation in e2e tests that use sentry nodes.
- [x] Make the IAS proxy not panic when connection is dropped, it should just reconnect.
- [x] Address review comments.
- [x] Add check to gRPC proxy to make sure that upstream connection is in a good state before forwarding calls (and re-dial upstream if the connection has gone bad).
- [x] Add separate long-term TLS certificates for the sentry node's control connection client.
- [x] Address review comments.

### DIFF
--- a/.buildkite/code.pipeline.yml
+++ b/.buildkite/code.pipeline.yml
@@ -203,7 +203,7 @@ steps:
   ###############
   - label: E2E tests
     parallelism: 7
-    timeout_in_minutes: 11
+    timeout_in_minutes: 12
     command:
       - .buildkite/scripts/download_e2e_test_artifacts.sh
       - .buildkite/scripts/test_e2e.sh
--- a/.buildkite/code.pipeline.yml
+++ b/.buildkite/code.pipeline.yml
@@ -203,7 +203,7 @@ steps:
   ###############
   - label: E2E tests
     parallelism: 7
-    timeout_in_minutes: 12
+    timeout_in_minutes: 9
     command:
       - .buildkite/scripts/download_e2e_test_artifacts.sh
       - .buildkite/scripts/test_e2e.sh
--- a/.buildkite/code.pipeline.yml
+++ b/.buildkite/code.pipeline.yml
@@ -203,7 +203,7 @@ steps:
   ###############
   - label: E2E tests
     parallelism: 7
-    timeout_in_minutes: 9
+    timeout_in_minutes: 11
     command:
       - .buildkite/scripts/download_e2e_test_artifacts.sh
       - .buildkite/scripts/test_e2e.sh
--- a/.changelog/2098.breaking.md
+++ b/.changelog/2098.breaking.md
@@ -1,0 +1,5 @@
+Sentry nodes no longer require TLS certificate file of the upstream node
+
+The `worker.sentry.grpc.upstream.cert` option has been removed.
+Instead, use `worker.sentry.grpc.upstream.id` to specify the
+upstream node's ID.
--- a/.changelog/2098.feature.md
+++ b/.changelog/2098.feature.md
@@ -1,0 +1,6 @@
+node: Add automatic TLS certificate rotation support
+
+It is now possible to automatically rotate the node's TLS
+certificates every N epochs by passing the command-line flag
+`worker.registration.rotate_certs`.
+Do not use this if you use sentry nodes or IAS proxies.
--- a/.changelog/2098.feature.md
+++ b/.changelog/2098.feature.md
@@ -3,4 +3,4 @@ node: Add automatic TLS certificate rotation support
 It is now possible to automatically rotate the node's TLS
 certificates every N epochs by passing the command-line flag
 `worker.registration.rotate_certs`.
-Do not use this if you use sentry nodes or IAS proxies.
+Do not use this option on sentry nodes or IAS proxies.
--- a/go/common/grpc/grpc.go
+++ b/go/common/grpc/grpc.go
@@ -23,6 +23,7 @@ import (
 	"google.golang.org/grpc/keepalive"
 
 	"github.com/oasislabs/oasis-core/go/common/grpc/auth"
+	"github.com/oasislabs/oasis-core/go/common/identity"
 	"github.com/oasislabs/oasis-core/go/common/logging"
 	"github.com/oasislabs/oasis-core/go/common/service"
 )
@@ -298,8 +299,8 @@ type ServerConfig struct { // nolint: maligned
 	Port uint16
 	// Path is the path for the local server. Leave nil to create a TCP server.
 	Path string
-	// Certificate is the certificate used by the server. Should be nil for local servers.
-	Certificate *tls.Certificate
+	// Identity is the identity of the worker that's running the server.
+	Identity *identity.Identity
 	// InstallWrapper specifies whether intercepting facilities should be enabled on this server,
 	// to enable intercepting RPC calls with a wrapper.
 	InstallWrapper bool
@@ -473,11 +474,14 @@ func NewServer(config *ServerConfig) (*Server, error) {
 		grpc.KeepaliveParams(serverKeepAliveParams),
 		grpc.CustomCodec(&CBORCodec{}),
 	}
-	if config.Certificate != nil {
+	if config.Identity != nil && config.Identity.GetTLSCertificate() != nil {
 		tlsConfig := &tls.Config{
-			Certificates: []tls.Certificate{*config.Certificate},
-			ClientAuth:   clientAuthType,
+			ClientAuth: clientAuthType,
+			GetCertificate: func(ch *tls.ClientHelloInfo) (*tls.Certificate, error) {
+				return config.Identity.GetTLSCertificate(), nil
+			},
 		}
+
 		sOpts = append(sOpts, grpc.Creds(credentials.NewTLS(tlsConfig)))
 	}
 	sOpts = append(sOpts, config.CustomOptions...)
--- a/go/common/grpc/proxy/proxy.go
+++ b/go/common/grpc/proxy/proxy.go
@@ -9,6 +9,7 @@ import (
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 
@@ -73,6 +74,12 @@ func (p *proxy) handler(srv interface{}, stream grpc.ServerStream) error {
 	}
 	// Pass subject header upstream.
 	upstreamCtx = metadata.AppendToOutgoingContext(upstreamCtx, policy.ForwardedSubjectMD, sub)
+
+	// Check if upstream connection was disconnected.
+	if p.upstreamConn != nil && p.upstreamConn.GetState() == connectivity.Shutdown {
+		// We need to redial if the connection was shut down.
+		p.upstreamConn = nil
+	}
 
 	// Dial upstream if necessary.
 	if p.upstreamConn == nil {
--- a/go/common/grpc/proxy/proxy.go
+++ b/go/common/grpc/proxy/proxy.go
@@ -17,21 +17,29 @@ import (
 	"github.com/oasislabs/oasis-core/go/common/logging"
 )
 
-// Handler returns a grpc StreamHandler than can be used
-// to proxy requests to provided client.
-// XXX: potentially the connection should be established in this package,
-// with some sensible defaults e.g. KeepAlive set.
-// We might also want to establish a pool of connections to the upstream.
-func Handler(conn *grpc.ClientConn) grpc.StreamHandler {
+// Dialer should return a gRPC ClientConn that will be used
+// to forward calls to.
+type Dialer func(ctx context.Context) (*grpc.ClientConn, error)
+
+// Handler returns a gRPC StreamHandler than can be used
+// to proxy requests to the client returned by the proxy dialer.
+func Handler(dialer Dialer) grpc.StreamHandler {
 	proxy := &proxy{
 		logger:       logging.GetLogger("grpc/proxy"),
-		upstreamConn: conn,
+		dialer:       dialer,
+		upstreamConn: nil, // Will be dialed on-demand.
 	}
 
 	return grpc.StreamHandler(proxy.handler)
 }
 
 type proxy struct {
+	// This is the dialer callback we use to make new connections to the
+	// upstream server if the connection drops, etc.
+	dialer Dialer
+
+	// This is a cached client connection to the upstream server, so we
+	// don't have to re-dial it on every call.
 	upstreamConn *grpc.ClientConn
 
 	logger *logging.Logger
@@ -66,6 +74,15 @@ func (p *proxy) handler(srv interface{}, stream grpc.ServerStream) error {
 	// Pass subject header upstream.
 	upstreamCtx = metadata.AppendToOutgoingContext(upstreamCtx, policy.ForwardedSubjectMD, sub)
 
+	// Dial upstream if necessary.
+	if p.upstreamConn == nil {
+		var grr error
+		p.upstreamConn, grr = p.dialer(stream.Context())
+		if grr != nil {
+			return grr
+		}
+	}
+
 	upstreamStream, err := grpc.NewClientStream(
 		upstreamCtx,
 		desc,
@@ -92,7 +109,7 @@ func (p *proxy) handler(srv interface{}, stream grpc.ServerStream) error {
 				// can still be in progress.
 				p.logger.Debug("downstream EOF")
 				if err = upstreamStream.CloseSend(); err != nil {
-					p.logger.Error("failrue closing upstream stream",
+					p.logger.Error("failure closing upstream stream",
 						"err", err,
 					)
 				}
@@ -154,6 +171,7 @@ func (p *proxy) proxyUpstream(downstream grpc.ServerStream, upstream grpc.Client
 func (p *proxy) proxyDownstream(upstream grpc.ClientStream, downstream grpc.ServerStream) <-chan error {
 	errCh := make(chan error, 1)
 	var headerSent bool
+
 	go func() {
 		for {
 			// Wait for stream msg (from upstream).
--- a/go/common/identity/identity.go
+++ b/go/common/identity/identity.go
@@ -30,6 +30,10 @@ const (
 
 	tlsKeyFilename  = "tls_identity.pem"
 	tlsCertFilename = "tls_identity_cert.pem"
+
+	// These are used for the sentry client connection to the sentry node and are never rotated.
+	tlsSentryClientKeyFilename  = "sentry_client_tls_identity.pem"
+	tlsSentryClientCertFilename = "sentry_client_tls_identity_cert.pem"
 )
 
 // ErrCertificateRotationForbidden is returned by RotateCertificates if
@@ -48,14 +52,17 @@ type Identity struct {
 	P2PSigner signature.Signer
 	// ConsensusSigner is a node consensus key signer.
 	ConsensusSigner signature.Signer
+	// TLSSentryClientCertificate is the client certificate used for
+	// connecting to the sentry node's control connection.  It is never rotated.
+	TLSSentryClientCertificate *tls.Certificate
+	// DoNotRotateTLS flag is true if we mustn't rotate the TLS certificates below.
+	DoNotRotateTLS bool
 	// tlsSigner is a node TLS certificate signer.
 	tlsSigner signature.Signer
 	// tlsCertificate is a certificate that can be used for TLS.
 	tlsCertificate *tls.Certificate
 	// nextTLSCertificate is a certificate that can be used for TLS in the next rotation.
 	nextTLSCertificate *tls.Certificate
-	// DoNotRotateTLS flag is true if we mustn't rotate the TLS certificates.
-	DoNotRotateTLS bool
 }
 
 // RotateCertificates rotates the identity's TLS certificates.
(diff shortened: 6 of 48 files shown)


go/common/grpc/grpc.go
```diff
@@ -473,11 +474,14 @@ func NewServer(config *ServerConfig) (*Server, error) {
...
 	sOpts = append(sOpts, config.CustomOptions...)
 
-	if config.Certificate != nil {
+	if config.Identity != nil {
 		tlsConfig := &tls.Config{
-			Certificates: []tls.Certificate{*config.Certificate},
-			ClientAuth:   clientAuthType,
+			ClientAuth: clientAuthType,
+			GetCertificate: func(ch *tls.ClientHelloInfo) (*tls.Certificate, error) {
+				return config.Identity.TLSCertificate, nil
```
kostko: This seems racy. Can you change it so that getting the `TLSCertificate` from identity is synchronized with rotation (e.g., add accessor that uses a mutex)?

go/common/identity/identity.go
```diff
@@ -42,6 +39,28 @@ type Identity struct {
 	TLSSigner signature.Signer
 	// TLSCertificate is a certificate that can be used for TLS.
 	TLSCertificate *tls.Certificate
+	// NextTLSCertificate is a certificate that can be used for TLS in the next epoch.
+	NextTLSCertificate *tls.Certificate
+}
+
+// RotateCertificates rotates the TLS certificates.
+// This is called on each epoch change.
+func (i *Identity) RotateCertificates() error {
```
kostko: Synchronize with tls certificate accessor.

go/common/identity/identity.go
```diff
@@ -86,40 +105,22 @@ func doLoadOrGenerate(dataDir string, signerFactory signature.SignerFactory, sho
...
-	if shouldGenerate {
-		cert, err = tlsCert.LoadOrGenerate(tlsCertPath, tlsKeyPath, CommonName)
-	} else {
-		cert, err = tlsCert.Load(tlsCertPath, tlsKeyPath)
+	// TLS certificate is always freshly generated.
+	cert, err := tlsCert.Generate(CommonName)
+	if err != nil {
+		return nil, err
 	}
+	nextCert, err := tlsCert.Generate(CommonName)
```
kostko: Also make sure to reduce certificate expiration once this works.

go/runtime/committee/client.go
```diff
@@ -266,7 +266,9 @@ func (cc *committeeClient) updateConnectionLocked(n *node.Node) error {
 	}
 	if cc.clientIdentity != nil {
 		// Configure TLS client authentication if required.
-		tlsCfg.Certificates = []tls.Certificate{*cc.clientIdentity.TLSCertificate}
+		tlsCfg.GetCertificate = func(ch *tls.ClientHelloInfo) (*tls.Certificate, error) {
+			return cc.clientIdentity.TLSCertificate, nil
```
kostko: Synchronization.

go/worker/registration/worker.go
```diff
@@ -531,7 +555,6 @@ func (w *Worker) registerNode(epoch epochtime.EpochTime, hook RegisterNodeHook)
 		w.registrationSigner,
 		w.identity.P2PSigner,
 		w.identity.ConsensusSigner,
-		w.identity.TLSSigner,
```
kostko: Why has this been removed? It should probably be signed with both signers when `NextCertificate` is non-nil.

go/registry/api/api.go
```diff
@@ -572,18 +572,16 @@ func VerifyRegisterNodeArgs( // nolint: gocyclo
...
 		return nil, nil, err
 	}
-	if !sigNode.MultiSigned.IsSignedBy(certPub) {
-		logger.Error("RegisterNode: not signed by TLS certificate key",
-			"signed_node", sigNode,
-			"node", n,
-		)
-		return nil, nil, fmt.Errorf("%w: registration not signed by TLS certificate key", ErrInvalidArgument)
+	if n.Committee.NextCertificate != nil {
+		_, grr := verifyNodeNextCertificate(logger, &n)
```
kostko: Why a separate function, it looks similar?

go/registry/api/api.go
```diff
@@ -572,18 +572,16 @@ func VerifyRegisterNodeArgs( // nolint: gocyclo
 		)
 		return nil, nil, err
 	}
-	certPub, err := verifyNodeCertificate(logger, &n)
+	_, err := verifyNodeCertificate(logger, &n)
 	if err != nil {
 		return nil, nil, err
 	}
-	if !sigNode.MultiSigned.IsSignedBy(certPub) {
```
kostko: Why remove this check, we should sign with and verify both signatures.

go/common/identity/identity.go
abukosek: Sentry nodes and IAS proxies require non-rotating certificates, so I've left this as-is for now.

go/worker/registration/worker.go
abukosek: That was just to make debugging easier before, I've added it back in now :)

go/registry/api/api.go
abukosek: Fixed, thanks!

go/registry/api/api.go
abukosek: This was also removed for debugging and is now back.

.changelog/2098.feature.md
```diff
@@ -0,0 +1,6 @@
+node: Add automatic TLS certificate rotation support
+
+It is now possible to automatically rotate the node's TLS
+certificates every N epochs by passing the command-line flag
+`worker.registration.rotate_certs`.
+Do not use this if you use sentry nodes or IAS proxies.
```
Yawning: Why does this break those two things?

go/common/grpc/grpc.go
```diff
@@ -298,8 +299,8 @@ type ServerConfig struct { // nolint: maligned
 	Port uint16
 	// Path is the path for the local server. Leave nil to create a TCP server.
 	Path string
-	// Certificate is the certificate used by the server. Should be nil for local servers.
-	Certificate *tls.Certificate
+	// Identity is the identity of the worker that's running the server.
+	Identity *identity.Identity
```
Yawning: This makes it kind of hard to reuse this code for other things (#2686).

go/common/identity/identity.go
```diff
@@ -30,31 +32,124 @@ const (
...
-	// TLSCertificate is a certificate that can be used for TLS.
-	TLSCertificate *tls.Certificate
+	// tlsSigner is a node TLS certificate signer.
+	tlsSigner signature.Signer
+	// tlsCertificate is a certificate that can be used for TLS.
+	tlsCertificate *tls.Certificate
+	// nextTLSCertificate is a certificate that can be used for TLS in the next rotation.
+	nextTLSCertificate *tls.Certificate
+	// DoNotRotate flag is true if we mustn't rotate the TLS certificates.
+	DoNotRotate bool
```
Yawning: Is the intention that this flag will also eventually encompass other signing keys?  If not, the name could be made more descriptive.

go/worker/registration/worker.go
```diff
@@ -430,9 +465,18 @@ func (w *Worker) gatherCommitteeAddresses(sentryCommitteeAddrs []node.CommitteeA
 		}
 		for _, addr := range addrs {
 			committeeAddresses = append(committeeAddresses, node.CommitteeAddress{
-				Certificate: w.identity.TLSCertificate.Certificate[0],
+				Certificate: w.identity.GetTLSCertificate().Certificate[0],
 				Address:     addr,
 			})
+			// Make sure to also include the certificate that will be valid
+			// in the next epoch, so that the node remains reachable.
+			nextCert := w.identity.GetNextTLSCertificate()
```
Yawning: could merge these two lines: `if nextCert := w.identity.GetNextTLSCertificate(); nextCert != nil {`

.changelog/2098.feature.md
abukosek: Sentry nodes need the original node's certificate to be able to connect to it over GRPC and the validator node also needs the sentry node's certificate in order to query the sentry nodes on each re-registration.
I've talked to Tadej and Peter Us, who initially wrote the sentry node, and our conclusion was that doing automatic rotations when sentry nodes are involved is not possible the way the sentry node currently works.

go/common/grpc/grpc.go
abukosek: You can fill-in only the certificate field of the Identity struct and it will work the same way as before :)

go/common/identity/identity.go
abukosek: Good point, I'll fix this.

go/sentry/api/api.go
```diff
@@ -13,9 +14,22 @@ type SentryAddresses struct {
...
+
 // Backend is a sentry backend implementation.
 type Backend interface {
 	// Get addresses returns the list of consensus and committee addresses of
 	// the sentry node.
 	GetAddresses(context.Context) (*SentryAddresses, error)
+
+	// SetUpstreamTLSCertificates notifies the sentry node of the new
+	// TLS certificates used by its upstream node.
+	SetUpstreamTLSCertificates(context.Context, *tls.Certificate, *tls.Certificate) error
```
kostko: Use `*UpstreamTLSCertificates` as type here (or use the alternative `[][]byte` mentioned above).

go/sentry/api/api.go
```diff
@@ -13,9 +14,22 @@ type SentryAddresses struct {
 	Committee []node.CommitteeAddress
 }
 
+// UpstreamTLSCertificates contains the upstream TLS certificates.
+type UpstreamTLSCertificates struct {
+	Certificate     *tls.Certificate
```
kostko: This is serializable so add explicit field annotations. Also it should probaby not use `tls.Certificate` directly but only be a `[]byte`.

Alternative: just use a single argument of type `[][]byte` to include all certificates.

go/sentry/api/api.go
```diff
@@ -13,9 +14,22 @@ type SentryAddresses struct {
...
 	// Get addresses returns the list of consensus and committee addresses of
 	// the sentry node.
 	GetAddresses(context.Context) (*SentryAddresses, error)
+
+	// SetUpstreamTLSCertificates notifies the sentry node of the new
+	// TLS certificates used by its upstream node.
+	SetUpstreamTLSCertificates(context.Context, *tls.Certificate, *tls.Certificate) error
+
+	// GetUpstreamTLSCertificates returns the TLS certificates of the sentry node's upstream node.
+	GetUpstreamTLSCertificates(context.Context) (*UpstreamTLSCertificates, error)
```
kostko: Same here, should probably return `[][]byte` instead.

go/sentry/sentry.go
```diff
@@ -16,10 +18,15 @@ import (
...
 
 type backend struct {
+	sync.RWMutex
+
 	logger *logging.Logger
 
 	consensus consensus.Backend
 	identity  *identity.Identity
+
+	upstreamTLSCertificate     *tls.Certificate
```
kostko: I don't think you actually need to separate between the two here as you add them to a pool later. You could just simplify into `upstreamTLSCertificates [][]byte`.

go/sentry/client/client.go
```diff
@@ -45,9 +45,11 @@ func (c *Client) createConnection() error {
 	certPool := x509.NewCertPool()
 	certPool.AddCert(c.sentryCert)
 	creds := credentials.NewTLS(&tls.Config{
-		Certificates: []tls.Certificate{*c.nodeIdentity.TLSCertificate},
-		RootCAs:      certPool,
-		ServerName:   identity.CommonName,
+		RootCAs:    certPool,
+		ServerName: identity.CommonName,
+		GetClientCertificate: func(cri *tls.CertificateRequestInfo) (*tls.Certificate, error) {
+			return c.nodeIdentity.GetTLSCertificate(), nil
```
kostko: This will probably not work as this certificate can be rotated, the control connection should use its own long-term certs?

abukosek: Yes, I'll add that as part of this PR :)

go/common/grpc/proxy/proxy.go
```diff
@@ -66,6 +74,15 @@ func (p *proxy) handler(srv interface{}, stream grpc.ServerStream) error {
 	// Pass subject header upstream.
 	upstreamCtx = metadata.AppendToOutgoingContext(upstreamCtx, policy.ForwardedSubjectMD, sub)
 
+	// Dial upstream if necessary.
+	if p.upstreamConn == nil {
```
kostko: Hm so if I see this correctly, after the initial dialing `p.upstreamConn` is never set to `nil`. So the `p.dialer` will only ever be invoked once and so the upstream certificates will never be refreshed?

Maybe this should additionally check if the connection is in a bad state and re-dial in that case? Of course the optimal thing would be to instead have a dynamic `RootCAs` but I don't think this is possible?

go/oasis-node/cmd/identity/identity.go
```diff
@@ -50,7 +50,7 @@ func doNodeInit(cmd *cobra.Command, args []string) {
 		)
 		os.Exit(1)
 	}
-	if _, err = identity.LoadOrGenerate(dataDir, nodeSignerFactory); err != nil {
+	if _, err = identity.LoadOrGenerate(dataDir, nodeSignerFactory, true); err != nil {
```
Yawning: Does this mean that nodes provisioned this way won't have ephemeral TLS certs unless they obliterate the persisted cert?

abukosek: Yes.  As rotation is off by default for now, I thought this would make most sense.
Once we enable rotation by default, we can also disable persistent TLS certs here or add a flag to control this behavior.

go/common/grpc/proxy/proxy.go
abukosek: Ah, good idea, I'll add a check.  Yes, as far as I've checked, it's not possible to dynamically change the certs once the connection has been established.

go/sentry/api/api.go
abukosek: Thanks for the suggestion, I've switched to using `[][]byte` for all these cases.

.buildkite/code.pipeline.yml
```diff
@@ -203,7 +203,7 @@ steps:
   ###############
   - label: E2E tests
     parallelism: 7
-    timeout_in_minutes: 9
+    timeout_in_minutes: 12
```
ptrus: Can probably change this back before merging.

.changelog/2098.feature.md
```diff
@@ -0,0 +1,6 @@
+node: Add automatic TLS certificate rotation support
```
ptrus: Probably open an issue in https://github.com/oasislabs/docs once this is merged, so we wont forget adding/updating docs

go/common/identity/identity.go
```diff
@@ -30,31 +32,124 @@ const (
 	tlsCertFilename = "tls_identity_cert.pem"
 )
 
+// ErrCertificateRotationForbidden is returned by RotateCertificates if
+// TLS certificate rotation is forbidden.  This happens when rotation is
```
ptrus:
```suggestion
// TLS certificate rotation is forbidden. This happens when rotation is
```

go/oasis-node/cmd/ias/auth_registry.go
```diff
@@ -53,15 +53,40 @@ func (auth *registryAuthenticator) VerifyEvidence(ctx context.Context, evidence
...
+func (auth *registryAuthenticator) worker(ctx context.Context) {
 	waitRuntimes := viper.GetInt(cfgWaitRuntimes)
 	if waitRuntimes <= 0 {
 		close(auth.initCh)
 	}
 
-	ch, sub, err := auth.client.WatchRuntimes(ctx)
+	var redialAttempts uint
+
+Redial:
```
ptrus: Could use the [cenkalti/backoff](https://github.com/cenkalti/backoff) package we use in other similar cases?

go/worker/sentry/grpc/init.go
```diff
@@ -29,9 +32,8 @@ const (
 
 	// CfgUpstreamAddress is the grpc address of the upstream node.
 	CfgUpstreamAddress = "worker.sentry.grpc.upstream.address"
-
-	// CfgUpstreamCert is the path to the certificate files of the upstream node.
-	CfgUpstreamCert = "worker.sentry.grpc.upstream.cert"
+	// CfgUpstreamID is the node ID of the upstream node.
+	CfgUpstreamID = "worker.sentry.grpc.upstream.id"
```
ptrus: Label the PR with `breaking-cfg` and include the change in the changelog fragment.

go/worker/sentry/worker.go
```diff
@@ -133,16 +133,21 @@ func New(backend api.Backend, identity *identity.Identity) (*Worker, error) {
...
 	}
 	w.grpcWorker = sentryGrpcWorker
 
-	// Stop in case of grpc/worker quitting.
 	go func() {
-		<-w.grpcWorker.Quit()
-		w.Stop()
+		for { // nolint: S1000
+			// Stop in case of grpc/worker quitting.
+			select {
```
ptrus: Does this change anything compared to current behaviour?

.buildkite/code.pipeline.yml
abukosek: Yes, this is still a WIP :)

.changelog/2098.feature.md
abukosek: Will do.

go/common/identity/identity.go
abukosek: This is not a typo, I prefer using double-spaces at the end of sentences in monospace fonts, because it makes sentence boundaries clearer and easier to read.

go/oasis-node/cmd/ias/auth_registry.go
abukosek: I think that would be an overkill for this simple use case -- the only time this connection is dropped is if the upstream node rotates its TLS certificates, a constant 2s retry is good enough.

go/worker/sentry/grpc/init.go
abukosek: Will do.

go/worker/sentry/worker.go
abukosek: That is a leftover from some testing, I'll clean it up before this is done.

abukosek:
Thanks to everyone for your previous review comments.  This PR is now ready for re-review :)

go/common/identity/identity.go
```diff
@@ -28,33 +30,132 @@ const (
...
 	// P2PSigner is a node P2P link key signer.
 	P2PSigner signature.Signer
 	// ConsensusSigner is a node consensus key signer.
 	ConsensusSigner signature.Signer
-	// TLSSigner is a node TLS certificate signer.
-	TLSSigner signature.Signer
-	// TLSCertificate is a certificate that can be used for TLS.
-	TLSCertificate *tls.Certificate
+	// tlsSigner is a node TLS certificate signer.
+	tlsSigner signature.Signer
```
Yawning, line 56: As a matter of style I prefer private struct members to be separated from public ones (and not interleaved like it is here).

go/worker/sentry/grpc/worker.go
```diff
@@ -152,6 +165,32 @@ func (g *Worker) worker() {
 	defer close(g.quitCh)
 	defer (g.cancelCtx)()
 
+	if g.upstreamConn == nil {
+		var numRetries uint
+
+		dialUpstream := func() error {
+			_, err := g.upstreamDialer(g.ctx)
+			if err != nil {
+				if numRetries < 60 {
```
ptrus, line 174: Could instead wrap the existing backoff with [`WithMaxRetries`](https://godoc.org/github.com/cenkalti/backoff#WithMaxRetries)

go/worker/sentry/grpc/init.go
```diff
@@ -123,28 +153,34 @@ func New(identity *identity.Identity) (*Worker, error) {
 	if g.enabled {
 		logger.Info("Initializing gRPC sentry worker")
 
-		upstreamConn, err := initConnection(identity)
-		if err != nil {
-			return nil, fmt.Errorf("gRPC sentry worker initializing upstream connection failure: %w", err)
+		g.upstreamDialer = func(ctx context.Context) (*grpc.ClientConn, error) {
```
ptrus, line 156: Could the mutex be removed if instead the dialer just did `initConnection` and returned it (without storing it in the sentry worker struct)? 

It might be also a bit cleaner, since in that case there will always be two connections, one the proxy will use and one the policy watcher? (and each would keep track of it's own connection like both already do, and if needed, initialize more connections). I think the only difference to current impl. would be that currently the connection is ?sometimes? shared (I think currently it depends if `proxy` or `sentry` first invoke the dialer, if the connection is shared or not?).

---

On the other hand if we want to keep using one connection i think it could be done cleaner if `upstreamDialer` is it's own struct (and not part of sentry worker), that maintains a connection (or a pool of connections).

.changelog/2098.breaking.md
```diff
@@ -0,0 +1,5 @@
+Sentry nodes no longer require TLS certificate file of the upstream node
```
ptrus, line 1: Nice :+1:

go/worker/sentry/grpc/init.go
abukosek: The idea is to replace the policy watcher with a new sentry command (e.g. `UpdatePolicies` or something) that is called by the upstream node in a similar way to `SetUpstreamTLSCertificates` added in this PR.  This will allow further simplifications of the sentry code, but that should be done in a separate PR.
I'll make an issue for this after I merge the PR.

abukosek:
Thanks!

PR #2676, opened 2020-02-14
Makefile: set correct target-dir for elf2sgxs

ptrus:
Also using env var setting for consistency, as `cargo elf2sgxs` doesn't support the `--target-dir` flag


ptrus:
thanks @jberci for noticing, let me know if this works now :-)

jberci:
This seems to fix the Rust compilation issues for me, yeah. Thanks!

PR #2679, opened 2020-02-15
ci: skip some steps for non-code changes

### LINKED ISSUE #2573: Skip some CI steps based on pull request's changes
##### SUMMARY
When one makes a pull request that e.g. only adds documentation or assembles the Change Log from fragments, all the *heavy* Buildkite pipeline steps (e.g. Go/Rust building, Go tests, E2E tests) should be skipped.

##### ISSUE TYPE
- Ops

pro-wh:
in this PR, we skip some heavy Go and Rust builds and tests when changes are only in docs

blocks #2573 

test: https://github.com/oasislabs/oasis-core/pull/2682

### DIFF
--- a/.buildkite/scripts/skip_if_only_docs_changes.sh
+++ b/.buildkite/scripts/skip_if_only_docs_changes.sh
@@ -4,12 +4,15 @@ only_docs_changes() {
   # https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec
   test -n "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" &&
     git diff --quiet "refs/remotes/origin/$BUILDKITE_PULL_REQUEST_BASE_BRANCH.." -- \
-      ':(exclude).changelog' \
-      ':(exclude)CHANGELOG.md' \
-      ':(exclude)CONTRIBUTING.md' \
-      ':(exclude)LICENSE' \
-      ':(exclude)README.md' \
-      ':(exclude)SECURITY.md'
+      ':(exclude)*.md' \
+      ':(exclude).changelog/' \
+      ':(exclude).github/CODEOWNERS' \
+      ':(exclude).github/ISSUE_TEMPLATE/' \
+      ':(exclude).gitlint' \
+      ':(exclude).markdownlint.yml' \
+      ':(exclude).punch_config.py' \
+      ':(exclude)docs/' \
+      ':(exclude)towncrier.toml'
 }
 
 if only_docs_changes; then
--- a/.buildkite/scripts/skip_if_only_docs_changes.sh
+++ b/.buildkite/scripts/skip_if_only_docs_changes.sh
@@ -2,12 +2,14 @@
 
 only_docs_changes() {
   # https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec
+  # Some non-docs things related to workflows are also in .github/, but this script currently
+  # is specific to buildkite, where we wouldn't care about changes there anyway. If we later
+  # use this in other CI places, we can narrow it down.
   test -n "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" &&
     git diff --quiet "refs/remotes/origin/$BUILDKITE_PULL_REQUEST_BASE_BRANCH.." -- \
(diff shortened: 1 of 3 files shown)


.buildkite/scripts/skip_if_only_docs_changes.sh
```diff
@@ -0,0 +1,17 @@
+# Source this file rather than fork-exec-ing it.
+
+only_docs_changes() {
+  test -n "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" &&
+    git diff --quiet "refs/remotes/origin/$BUILDKITE_PULL_REQUEST_BASE_BRANCH.." -- \
```
tjanez: I would include a reference to [git's `pathspec` documentation](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec) since this is quite a new thing.

.buildkite/scripts/skip_if_only_docs_changes.sh
```diff
@@ -0,0 +1,17 @@
...
+      ':(exclude).changelog' \
+      ':(exclude)CHANGELOG.md' \
```
tjanez: I would replace all these with `:(exclude)*.md` to cover existing and new Markdown files anywhere (we have many README.md files in various subtrees).

.buildkite/scripts/skip_if_only_docs_changes.sh
```diff
@@ -0,0 +1,17 @@
+# Source this file rather than fork-exec-ing it.
+
+only_docs_changes() {
```
tjanez: I would generalize and rename this to `only_non_code_changes()` (also the `.sh` file) and include other pathspec patterns:
- `:(exclude).github/`
- `:(exclude)docker/`
- `:(exclude)docs/` (in case there will be non-Markdown files in the `docs/` directory, e.g. configuration files, images, etc.
- `:(exclude).codecov.yml`
- `:(exclude).dockerignore`
- `:(exclude).gitignore`
- `:(exclude).gitlint`
- `:(exclude).gitmodules`
- `:(exclude).goreleaser.yml`
- `:(exclude).markdownlint.yml`
- `:(exclude).punch_config.py`
- `:(exclude)towncrier.toml`

.buildkite/scripts/skip_if_only_docs_changes.sh
pro-wh: ok

.buildkite/scripts/skip_if_only_docs_changes.sh
pro-wh: ok

.buildkite/scripts/skip_if_only_docs_changes.sh
pro-wh: ok

pro-wh: I'll keep `docker/` as code for now

pro-wh: hm some of these other ones we'll still want to run CI on:

- codecov, so that we generate coverage for it
- .github/workflows, because ugh this is philosophically part of CI, and we're not clear that we'll never do code builds on github actions, or that this helper is specific to buildkite
- .gitmodules

pro-wh: okay overall I'm not ready to extend this to those other non-documentation non-code files

pro-wh: I'll scope down this PR

tjanez: > I'll keep `docker/` as code for now

Ok

> codecov, so that we generate coverage for it

Ok

> .github/workflows, because ugh this is philosophically part of CI, and we're not clear that we'll never do code builds on github actions, or that this helper is specific to buildkite

I would keep excluding this one since the helper being in ` .buildkite/scripts/skip_if_only_docs_changes.sh` makes it Buildkite-specific (for now).

> .gitmodules

We should just delete it since it is empty.

> I'll scope down this PR

Ok, this is a good start and we can extend it later.

pro-wh: ok

Issue #2689, opened 2020-02-18
Bump minimum golang version to 1.13.8


kostko:
Recent versions contain security-related fixes in the Go runtime.

tjanez:
I would like to note that the [`release` GitHub Actions workflow](https://github.com/oasislabs/oasis-core/actions?query=workflow:release) currently automatically uses the latest Go 1.13.x release:

https://github.com/oasislabs/oasis-core/blob/ab684b462b77869080c8ac6cf4f4c2c0abb3e25c/.github/workflows/release.yml#L21-L24

PR #2700, opened 2020-02-20
docker: bump go version

pro-wh:
blocks #2689 

This updates the development and testing Docker images.


pro-wh:
ya good idea

pro-wh:
this is approved, so I'll clobber the 0.3.0 images with my new one

PR #2702, opened 2020-02-21
buildkite: Add dynamic pipeline generator for conditional skipping

### LINKED ISSUE #2573: Skip some CI steps based on pull request's changes

tjanez:
Use Buildkite's dynamic pipelines feature to dynamically choose between _Code_/_Code-skip_ pipelines depending on whether the build was triggered for a pull request and the list of changes files.

Closes #2573.


tjanez:
> Should we also remove `buildkite/scripts/skip_if_only_docs_changes.sh` (and its use) now?

Yes, good catch. Done.

.buildkite/pipeline.sh
```diff
@@ -0,0 +1,45 @@
...
+# https://buildkite.com/docs/pipelines/defining-steps#dynamic-pipelines.
+#
+
+set -eux
+
+# Helper that ensures the build is triggered for a pull request and that there
+# are no code-related changes compared to the pull request's base branch.
+pr_and_no_code_related_changes() {
+  # Check if the build was triggered for a pull request.
+  if [[ -z $BUILDKITE_PULL_REQUEST_BASE_BRANCH ]]; then
```
pro-wh: can't skip in this case:
- we should build on master for PR merge commits, because there are steps that are set to run on master only
- you'd normally push a branch first and then open a PR for it. that first push wouldn't be built under this rule, because a PR doesn't exist yet

tjanez:
> can't skip in this case:
> - we should build on master for PR merge commits, because there are steps that are set to run on master only

I'm afraid I don't understand. So, for non-pull-request builds, the full code pipeline will always be run. For an example, see the [recent builds for the `master` branch](https://buildkite.com/oasislabs/oasis-core-ci/builds?branch=master).

> - you'd normally push a branch first and then open a PR for it. that first push wouldn't be built under this rule, because a PR doesn't exist yet

Yes, as I announced internally, I needed to make a slight Buildkite triggering change:
- Builds won't be triggered for all branches automatically, just the `master` and `stable/*` branches.
- For other branches, one will need to make a pull request.

This is needed since the dynamic Buildkite pipeline needs to know the base branch and changes from it to decide whether to run the full code pipeline or skip it.

pro-wh:
oh ok then thanks

PR #2706, opened 2020-02-21
readme: avoid tricky parsing of hard wrapped link

### LINKED ISSUE #2707: gitlint/git lint commits selection working wrong
##### SUMMARY

https://github.com/oasislabs/oasis-core/pull/2706/checks?check_run_id=461288544#step:7:184
that commit shouldn't be included. when I do `gitlint --debug --commits origin/master...HEAD` locally, it doesn't. something's wrong on CI where it is including it

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

github workflows

##### OASIS NODE VERSION

```paste below

```
(n/a)

##### OS / ENVIRONMENT

(n/a)

##### STEPS TO REPRODUCE

push certain PRs. possibly related to merge commits

##### ACTUAL RESULTS

gitlint checks all historical commits

```paste below

```
(too verbose. see action output in link)

##### EXPECTED RESULTS

should run on b05113a7f4862a31dbe6d62fc334d66817738cde, 269279fb77415d0cba4e8537fab944d6b91cba6e, 39f0a8432ac6314b971c4680d8e12f9329d912b5, 1e39860c6953fec12393f36e1535855f93f19dc3, and a069f99d0e64c603dfe079d596d4eebd643dfc37

pro-wh:
test PR for git lint problem

fixes #2707

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ lint-go:
 lint-git: fetch-git
 	@COMMIT_SHA=`git rev-parse $(OASIS_CORE_GIT_ORIGIN_REMOTE)/$(RELEASE_BRANCH)` && \
 	echo "Running gitlint for commits from $(OASIS_CORE_GIT_ORIGIN_REMOTE)/$(RELEASE_BRANCH) ($${COMMIT_SHA:0:7})..."; \
-	gitlint --commits $(OASIS_CORE_GIT_ORIGIN_REMOTE)/$(RELEASE_BRANCH)...HEAD
+	gitlint --commits $(OASIS_CORE_GIT_ORIGIN_REMOTE)/$(RELEASE_BRANCH)..HEAD
 
 lint-md:
 	@npx markdownlint-cli '**/*.md' --ignore .changelog/
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,8 @@ lint-targets := lint-go lint-git lint-md lint-changelog
 lint-go:
 	@$(MAKE) -C go lint
 
+# NOTE: gitlint internally uses git rev-list, where A..B is asymmetric difference, which is kind of the opposite of
+# how git diff interprets A..B vs A...B.
 lint-git: fetch-git
 	@COMMIT_SHA=`git rev-parse $(OASIS_CORE_GIT_ORIGIN_REMOTE)/$(RELEASE_BRANCH)` && \
 	echo "Running gitlint for commits from $(OASIS_CORE_GIT_ORIGIN_REMOTE)/$(RELEASE_BRANCH) ($${COMMIT_SHA:0:7})..."; \
--- a/README.md
+++ b/README.md
@@ -50,8 +50,8 @@ Prerequisites:
 
 * Linux (if you are not on Linux, you will need to either set up a VM with the
   proper environment or, if Docker is available for your platform, use the
-  provided Docker image which does this for you, [see below](
-  #using-the-development-docker-image)).
+  provided Docker image which does this for you,
+  [see below](#using-the-development-docker-image)).
 
 * System packages:
   * [Bubblewrap](https://github.com/projectatomic/bubblewrap) (at least version
--- a/.changelog/2706.trivial.md
+++ b/.changelog/2706.trivial.md
@@ -1,0 +1,1 @@
+readme: Avoid tricky parsing of hard wrapped link


.github/workflows/ci-lint.yml
```diff
@@ -42,6 +42,11 @@ jobs:
           python -m pip install \
             https://github.com/oasislabs/towncrier/archive/oasis-master.tar.gz \
             gitlint
+      - name: Lint git commits
```
tjanez: Hmm... changing the order will have unintended consequences, i.e. the next step will not be run when _Lint git commits_ step fails...

README.md
```diff
@@ -50,8 +50,8 @@ Prerequisites:
 
 * Linux (if you are not on Linux, you will need to either set up a VM with the
   proper environment or, if Docker is available for your platform, use the
-  provided Docker image which does this for you, [see below](
-  #using-the-development-docker-image)).
+  provided Docker image which does this for you,
+  [see below](#using-the-development-docker-image)).
```
tjanez: I'm curious if this is just visual/human issue or does any tool have problems with that?

Makefile
```diff
@@ -69,7 +69,7 @@ lint-go:
 lint-git: fetch-git
 	@COMMIT_SHA=`git rev-parse $(OASIS_CORE_GIT_ORIGIN_REMOTE)/$(RELEASE_BRANCH)` && \
 	echo "Running gitlint for commits from $(OASIS_CORE_GIT_ORIGIN_REMOTE)/$(RELEASE_BRANCH) ($${COMMIT_SHA:0:7})..."; \
-	gitlint --commits $(OASIS_CORE_GIT_ORIGIN_REMOTE)/$(RELEASE_BRANCH)...HEAD
+	gitlint --commits $(OASIS_CORE_GIT_ORIGIN_REMOTE)/$(RELEASE_BRANCH)..HEAD
```
tjanez: I'm still curious how/why https://github.com/oasislabs/oasis-core/actions/runs/42614734 worked and https://github.com/oasislabs/oasis-core/actions/runs/43145438 didn't work?

It would be nice if you explained this in git commit's message.

tjanez: > I'm still curious how/why https://github.com/oasislabs/oasis-core/actions/runs/42614734 worked and https://github.com/oasislabs/oasis-core/actions/runs/43145438 didn't work?

Sorry, I see the explanation now: https://github.com/oasislabs/oasis-core/issues/2707#issuecomment-589895264.

> It would be nice if you explained this in git commit's message.

Actually, we should make a note in `.github/workflows/ci-lint.yml` that the Change log fragment check step modifies git history.

tjanez: > Actually, we should make a note in .github/workflows/ci-lint.yml that the Change log fragment check step modifies git history.

I think it's even better if the Change log fragment check step doesn't mess with git's history. I've implemented this in #2712.

.github/workflows/ci-lint.yml
pro-wh: ^ I'm willing to make that compromise for the correctness of this step

README.md
pro-wh: my IDE's shabby markdown mode ):

pro-wh: image

tjanez: Aha, I see :slightly_smiling_face:.

No worries, we can wrap it like you suggested.

Makefile
pro-wh: thanks. I'll back out that part of this PR

tjanez: From `git-diff` manual:
```
git diff [<options>] <commit> <commit> [--] [<path>...]
   This is to view the changes between two arbitrary <commit>.

git diff [<options>] <commit>..<commit> [--] [<path>...]
   This is synonymous to the previous form. If <commit> on one side is
   omitted, it will have the same effect as using HEAD instead.

git diff [<options>] <commit>...<commit> [--] [<path>...]
   This form is to view the changes on the branch containing and up to
   the second <commit>, starting at a common ancestor of both
   <commit>. "git diff A...B" is equivalent to "git diff $(git
   merge-base A B) B". You can omit any one of <commit>, which has the
   same effect as using HEAD instead.
```

In case a branch is rebased on top of the current `${RELEASE_BRANCH}` (e.g. `master` branch), using `<commit>..<commit` or `<commit>...<commit>` doesn't make a difference.

However, in case a branch is not up-to-date, then using the `<commit>...<commit>` form will compare the current branch's `HEAD` with the common ancestor of the `${RELEASE_BRANCH}` and the current branch, i.e. the commit from which the current branch was branched from.
In my opinion, that makes more sense since one wants to see `gitlint`'s output for the commits he made in his current branch.

.github/workflows/ci-lint.yml
pro-wh: nvm, backing this out due to fixes elsewhere that make its unnecessary

tjanez: This should no longer be necessary since a different fix landed in #2712.

Makefile
pro-wh: Thanks for looking into this. `gitlint` internally uses `git rev-list` rather than `git diff`, with `git rev-list`'s documentation stating:

```
       Thus, the following command:

                   $ git rev-list foo bar ^baz

       means "list all the commits which are reachable from foo or bar, but
       not from baz".

       A special notation "<commit1>..<commit2>" can be used as a short-hand
       for "^'<commit1>' <commit2>". For example, either of the following may
       be used interchangeably:

                   $ git rev-list origin..HEAD
                   $ git rev-list HEAD ^origin

       Another special notation is "<commit1>...<commit2>" which is useful
       for merges. The resulting set of commits is the symmetric difference
       between the two operands. The following two commands are equivalent:

                   $ git rev-list A B --not $(git merge-base --all A B)
                   $ git rev-list A...B
```

tjanez: Oh, I see, thanks for looking in how gitlint implements this.

In this case, the correct way is to use `<commit>..<commit>`.

Can you add a comment above the `lint-git` target noting that gitlint uses `git rev-list` internally?

pro-wh: yeah

pro-wh:
alright codeowners please review

Issue #2506, opened 2019-12-31
Add more txsource workloads


kostko:
We should add more [txsource workloads](https://github.com/oasislabs/oasis-core/tree/master/go/oasis-node/cmd/debug/txsource/workload) to test out the system during longer term runs, specifically:

* [x] Delegation workloads
* [x] changing the commission schedule workload
* [x] Registry workloads.
* [x] Runtime workloads.

We should also consider using a genesis block which uses gas fees and rewards to test out those parts of the code.

kostko:
We should also add transactions that add randomized additional fields as payloads, e.g. see #2708.

kostko:
Also we should have stress tests that send lots of transactions in parallel every once in a while, so we test with large blocks.

kostko:
Based on a bug discovered in #2708, we should test with oversized transactions and also restart (some) validator nodes while the long-term txsource test is running.

kostko:
* [ ] We should also have some workloads that perform various historic queries using our exposed APIs and make sure the responses make sense.

ptrus:
done in all the linked PRs

PR #2709, opened 2020-02-23
go/oasis-test-runner: Improve txsource E2E test

### LINKED ISSUE #2708: Non-determinism due to incorrect parameter caching
Reporting an attack as a part of the ongoing "The Quest" challenge, and I'm guessing this is the place to do it.

##### SUMMARY
I wrote a [tx spammer](https://github.com/SebastianJ/oasis-spammer) in Go that spams the network with bogus transactions (with each tx having a payload as close as possible to the max tx size of 32768 bytes).

Initially this forced roughly 30% of the nodes on the network to crash and stop syncing, ending up with the error:

```
{"caller":"indexer_service.go:69","height":152154,"level":"info","module":"tendermint:txindex","msg":"Indexed block","ts":"2020-02-22T08:36:14.948753672Z"}
panic: Failed to process committed block (152155:2A579845B98EA9BF9D699409BDD19D4240A48C1ED0E90268FF8AB13547E0D02B): Wrong Block.Header.AppHash. Expected F9A183178993CE3F8D10F83BB1CBD5F41650ECE387324E94BAE7116E1C54F4F3, got E3D8BA16B0321B3E1B3A706A6AAF0835803A30F7F842C2054A24324097E52A94
```

(Multiple people on Slack whose nodes got terminated also confirmed that they received the same error message).

The only way to recover from this error was to remove all *.db and *.wal files, restart the node and effectively resync from block 0. This didn't work every time though - sometimes it did work, sometimes it didn't - so frustratedly enough you'd have to repeat this a couple of times in order to get the node to sync the chain properly again.

Here's how the top 25 leaderboard (using SmartStake's dashboard) looked after the initial wave of the attack:
The Quest Top 25 After the Attack

Outside of the top 25 plenty of other nodes were also affected. A lot of the nodes have since recovered (after performing full resyncs from block 0) but some are still offline since the start of the attack.

##### ISSUE TYPE
- Bug Report / Attack Report

##### COMPONENT NAME
`go/consensus/tendermint`

##### OASIS NODE VERSION
```
$ ./oasis-node --version
Software version: 20.3.1
Runtime protocol version: 0.11.0
Consensus protocol version: 0.23.0
Committee protocol version: 0.7.0
Tendermint core version: 0.32.8
ABCI library version: 0.16.1
```

##### OS / ENVIRONMENT
```
$ go version
go version go1.13.7 linux/amd64
```

##### STEPS TO REPRODUCE
```
mkdir -p oasis-spammer && cd oasis-spammer
bash <(curl -s -S -L https://raw.githubusercontent.com/SebastianJ/oasis-spammer/master/scripts/install.sh)
cd data && curl -LOs https://gist.githubusercontent.com/SebastianJ/c0844c22e582bf6c20787cc9744b6964/raw/f3730bd54c03c584d96511e05e581a1d8de2e309/bubzbase64.txt && mv bubzbase64.txt data.txt && cd ..
./oasis-spammer --genesis-file PATH/TO/genesis.json --entity-path PATH/TO/entity/folder --socket unix:PATH/TO/internal.sock --count 100000 --pool-size 100
```

The steps above will download the statically compiled go binary (for Linux), download the list of receiver addresses (to randomly send txs to) and also download the tx payload, a Base64 encoded picture of the one and only - **_Mr Bubz_**:
Bubz

##### ACTUAL RESULTS
Several nodes in "The Quest" network were terminated and couldn't get back to syncing/joining consensus unless they removed all databases (*.db & *.wal), restarted their nodes and resynced from block 0.

The following error message (or similar alterations of it) was displayed:
```
{"caller":"node.go:735","err":"tendermint: failed to create node: error during handshake: error on replay: Wrong Block.Header.AppHash. Expected E8774EFDB10979C780E1C16BEAFD2B67AC170295E4DAC7CB978AF25D0E35C199, got DF92822F6870181DF2D62B0DFB33D7436A70B6B1C90A5033E3971F143D454108","level":"error","module":"oasis-node","msg":"failed to start tendermint service","ts":"2020-02-22T07:29:38.5682535Z"}
```

##### EXPECTED RESULTS
Nodes should be able to withstand an attack like this. They should't get terminated or have to resync from block 0 to be able to participate in the consensus process again.

kostko:
See #2506.
Fixes #2708.

Includes a test that successfully reproduces #2708.

TODO
* [X] Now that we have a failing test, add the trivial fix for #2708.


go/oasis-node/cmd/debug/txsource/workload/oversized.go
```diff
@@ -0,0 +1,100 @@
...
+	// Fetch genesis consensus parameters.
+	// TODO: Don't dump everything, instead add a method to query just the parameters.
+	genesisDoc, err := cnsc.StateToGenesis(ctx, 1)
+	if err != nil {
+		return fmt.Errorf("failed to query state at genesis: %w", err)
+	}
+
+	var nonce uint64
+	fee := transaction.Fee{
+		Gas: 1000,
```
ptrus: make this a named constant

go/oasis-node/cmd/debug/txsource/workload/oversized.go
```diff
@@ -0,0 +1,100 @@
...
+		err = cnsc.SubmitTx(submitCtx, signedTx)
+		switch err {
+		case nil:
+			// This should never happen.
+			cancel()
+			return fmt.Errorf("successfully submitted an oversized transaction")
+		case consensus.ErrOversizedTx:
+			// Submitting an oversized transaction is an error, so we expect this to fail.
+			oversizedLogger.Info("transaction rejected due to ErrOversizedTx")
+		default:
```
ptrus: Could we ensure here that indeed timeout happened (and not some other unmatched error), probably another case matching on `context.Canceled`?

go/consensus/tendermint/tendermint.go
```diff
@@ -1357,6 +1360,7 @@ func init() {
 	Flags.Uint64(CfgConsensusMinGasPrice, 0, "minimum gas price")
 	Flags.Uint64(CfgConsensusSubmissionGasPrice, 0, "gas price used when submitting consensus transactions")
 	Flags.Uint64(CfgConsensusSubmissionMaxFee, 0, "maximum transaction fee when submitting consensus transactions")
+	Flags.Bool(CfgConsensusDisableCheckTx, false, "do not perform CheckTx on incoming transactions")
```
Yawning: This 100% should be hidden and behind the "this is unsafe" feature gate.

go/oasis-node/cmd/debug/txsource/workload/oversized.go
kostko: The problem is that gRPC wraps this in its own errors so it's not actually `context.Canceled`. In any case, we shouldn't fail here in case of errors (the only failure is if the transaction is accepted), so the result would be the same.

go/consensus/tendermint/tendermint.go
kostko: Yup, forgot about it. Should be changed now.

go/oasis-node/cmd/debug/txsource/workload/oversized.go
kostko: Done.

go/oasis-node/cmd/debug/txsource/workload/oversized.go
ptrus: failing on other errors would ensure that this actually testing oversised tx's (not any other potential errors). However i do see that the `mux.maxTxSize` check is one of the first checks that we do, so it should probably be fine :+1:

PR #2715, opened 2020-02-24
go/consensus/tendermint: Store consensus parameters in ABCI state

### LINKED ISSUE #2710: Store consensus parameters for consensus backend itself
We currently only store consensus parameters for the consensus backend itself (`.Consensus.Parameters`) in the genesis document. This doesn't allow for the option to change the consensus parameters later like we do for other mux apps where these parameters are stored under their own keys in ABCI state.

Consensus parameters should be moved under its own key in ABCI state to allow for future updates.

kostko:
Fixes #2710.

This allows for updating the consensus parameters later.

### DIFF
--- a/.changelog/2710.breaking.md
+++ b/.changelog/2710.breaking.md
@@ -1,0 +1,3 @@
+go/consensus/tendermint: Store consensus parameters in ABCI state
+
+This allows for updating the consensus parameters later.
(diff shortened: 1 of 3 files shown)


Yawning:
I'm not convinced this is correct.  This behaves quite differently from everything else (that stages alterations to state in the checktx tree, then applies them to the delivertx tree).

Additionally, while I am in favor of "performance", I'm not convinced that caching this is worth the hassle from a safety point of view.

kostko:
Right, so "set" must not be used outside InitChain/BeginBlock/EndBlock. We can enforce this nicely (also for all other apps where we support setting consensus parameters) in #2674 so let's move it there.

PR #2718, opened 2020-02-24
registration workload

ptrus:
- Adds registration txsource workloads (part of: https://github.com/oasislabs/oasis-core/issues/2506)

Currently runtime and entities only register at start and the workload then continuously registers nodes. Could update to also register new entities & runtimes during the run, if reviewers find it useful.

TODO:
- [x] cleanup todo's 
- [x] randomized registration descriptors (randomized roles, expiration epochs, could do more)

### DIFF
--- a/go/consensus/api/api.go
+++ b/go/consensus/api/api.go
@@ -44,6 +44,8 @@ var (
 // ClientBackend is a limited consensus interface used by clients that
 // connect to the local node.
 type ClientBackend interface {
+	TransactionAuthHandler
+
 	// SubmitTx submits a signed consensus transaction.
 	SubmitTx(ctx context.Context, tx *transaction.SignedTransaction) error
 
@@ -117,9 +119,6 @@ type Backend interface {
 	// SubmitEvidence submits evidence of misbehavior.
 	SubmitEvidence(ctx context.Context, evidence Evidence) error
 
-	// TransactionAuthHandler returns the transaction authentication handler.
-	TransactionAuthHandler() TransactionAuthHandler
-
 	// SubmissionManager returns the transaction submission manager.
 	SubmissionManager() SubmissionManager
 
@@ -150,7 +149,7 @@ type Backend interface {
 type TransactionAuthHandler interface {
 	// GetSignerNonce returns the nonce that should be used by the given
 	// signer for transmitting the next transaction.
-	GetSignerNonce(ctx context.Context, id signature.PublicKey, height int64) (uint64, error)
+	GetSignerNonce(ctx context.Context, req *GetSignerNonceRequest) (uint64, error)
 }
 
 // EvidenceKind is kind of evindence of a node misbehaving.
@@ -207,4 +206,10 @@ func NewConsensusEvidence(inner interface{}) ConsensusEvidence {
 type EstimateGasRequest struct {
 	Caller      signature.PublicKey      `json:"caller"`
 	Transaction *transaction.Transaction `json:"transaction"`
+}
+
+// GetSignerNonceRequest is a GetSignerNonce request.
+type GetSignerNonceRequest struct {
+	ID     signature.PublicKey `json:"id"`
+	Height int64               `json:"height"`
 }
--- a/go/consensus/api/api.go
+++ b/go/consensus/api/api.go
@@ -50,6 +50,9 @@ type ClientBackend interface {
 	// StateToGenesis returns the genesis state at the specified block height.
 	StateToGenesis(ctx context.Context, height int64) (*genesis.Document, error)
 
+	// EstimateGas calculates the amount of gas required to execute the given transaction.
+	EstimateGas(ctx context.Context, req *EstimateGasRequest) (transaction.Gas, error)
+
 	// WaitEpoch waits for consensus to reach an epoch.
 	//
 	// Note that an epoch is considered reached even if any epoch greater than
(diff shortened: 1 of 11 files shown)


go/consensus/api/api.go
```diff
@@ -57,6 +57,9 @@ type ClientBackend interface {
 	// in the future).
 	WaitEpoch(ctx context.Context, epoch epochtime.EpochTime) error
 
+	// GetEpoch returns the current epoch.
+	GetEpoch(ctx context.Context, height int64) (epochtime.EpochTime, error)
```
kostko: Add test for the new method (see `go/consensus/tests`).

go/oasis-node/cmd/debug/txsource/workload/registration.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	// Fetch fees.
+	// TODO: Don't dump everything, instead add a method to query just the parameters.
+	genesisDoc, err := cnsc.StateToGenesis(ctx, 1)
+	if err != nil {
+		return fmt.Errorf("failed to query state at genesis: %w", err)
+	}
+	// Entity registration fees.
+	entityRegisterFee := transaction.Fee{
```
kostko: Would be cool if this used `EstimateGas` (currently not exposed in the consensus `ClientBackend`, but should be trivial to do), related to #2723 (just the API part) instead of manually calculating the gas cost.

go/oasis-node/cmd/debug/txsource/workload/registration.go
```diff
@@ -0,0 +1,343 @@
...
+		registryLogger.Error("failed to sign node descriptor",
+			"err", err,
+		)
+		return nil, err
+	}
+
+	return sigNode, nil
+}
+
+func getAccountNonce(ctx context.Context, conn *grpc.ClientConn, pubKey signature.PublicKey) (uint64, error) {
```
kostko: Maybe we should expose `TransactionAuthHandler` in `consensus.ClientBackend` instead (it has a `GetSignerNonce` method)?

go/oasis-node/cmd/debug/txsource/workload/registration.go
```diff
@@ -0,0 +1,343 @@
...
+func (registration) Run(gracefulExit context.Context, rng *rand.Rand, conn *grpc.ClientConn, cnsc consensus.ClientBackend, rtc runtimeClient.RuntimeClient) error {
+	ctx := context.Background()
+	var err error
+
+	if err = ns.UnmarshalHex("0000000000000000000000000000000000000000000000000000000000000002"); err != nil {
+		panic(err)
+	}
+
+	baseDir := viper.GetString(cmdCommon.CfgDataDir)
+	nodeIdentitiesDir := filepath.Join(baseDir, "node-identities")
```
kostko: These directories should be per-workload, e.g., include `NameRegistration` in the path. Maybe we should have better support for this in the txsource runner.

go/oasis-node/cmd/debug/txsource/workload/registration.go
ptrus: cool! done and works :+1:

go/consensus/api/api.go
ptrus: done

go/consensus/api/api.go
```diff
@@ -50,13 +50,19 @@ type ClientBackend interface {
 	// StateToGenesis returns the genesis state at the specified block height.
 	StateToGenesis(ctx context.Context, height int64) (*genesis.Document, error)
 
+	// EstimateGas calculates the amount of gas required to execute the given transaction.
+	EstimateGas(ctx context.Context, req *EstimateGasRequest) (transaction.Gas, error)
```
kostko: Now that there's `EstimateGas` also add it to the consensus client API test.

go/oasis-node/cmd/debug/txsource/workload/registration.go
kostko: Nice!

go/oasis-node/cmd/debug/txsource/workload/registration.go
ptrus: yeah maybe? Probably would actually need to expose the `GetSignerNonce` method directly so it can work with the grpc backend? Could just embed the `TransactionAuthHandler` in `ClientBackend`?

go/oasis-node/cmd/debug/txsource/workload/registration.go
ptrus: yeah, fixed this by passing the correct base directory from the e2e/txsource runner (base dir is now the workload's directory)

go/oasis-node/cmd/debug/txsource/workload/registration.go
ptrus: done

go/oasis-node/cmd/debug/txsource/workload/registration.go
```diff
@@ -0,0 +1,339 @@
...
+	NameRegistration = "registration"
+
+	registryNumEntities        = 10
+	registryNumNodesPerEntity  = 5
+	registryNodeMaxEpochUpdate = 5
+)
+
+var registryLogger = logging.GetLogger("cmd/txsource/workload/registry")
+
+var ns common.Namespace
```
kostko: Maybe put this as state in the registration structure instead of making it global? This avoids potential conflicts with other workloads.

ptrus: good idea, thanks!

PR #2721, opened 2020-02-25
buildkite: use stable agents for daily longtests

ptrus:


ptrus:
@kostko check again please, i added another commit after your review

.buildkite/longtests.pipeline.yml
```diff
@@ -19,14 +19,14 @@ docker_plugin_default_config: &docker_plugin_default_config
     # Intel SGX Application Enclave Services Manager (AESM) daemon running on
     # the Buildkite host.
     - /var/run/aesmd/aesm.socket:/var/run/aesmd/aesm.socket
-    # Propapage the tmpfs mount.
-    - /tmp:/tmp
+    - /var/tmp/longtests:/var/tmp/longtests
```
kostko: Is there enough space in this directory? Now that tests will run for 6h this can accumulate a lot of data. Also maybe add a comment why this special directory (to preserve data in case the run fails).

ptrus: yeah will be enough for a run (it's over 200gb)

PR #2726, opened 2020-02-25
Add rust-toolchain

nhynes:
This makes it easier to not incorrectly `make`.


kostko:
What exactly does this do? Is it a hint for some tooling? Anyway, please squash commits.

nhynes:
> What exactly does this do?

The [rust-toolchain](https://github.com/rust-lang/rustup#the-toolchain-file) file overrides rustup's default toolchain when invoking `cargo`, `rustc`, and friends. This is helpful when working on projects that use multiple rust versions.

> please squash commits
👍 will do

PR #2727, opened 2020-02-26
go staking: disable all transactions for locked accounts

pro-wh:
The account field `general.transfers_not_before` is replaced with `general.not_before`. Accounts can't send any transaction before the epoch in that field.

### DIFF
--- a/go/common/errors/errors.go
+++ b/go/common/errors/errors.go
@@ -51,8 +51,8 @@ func New(module string, code uint32, msg string) error {
 	}
 
 	key := errorKey(module, code)
-	if _, isRegistered := registeredErrors.Load(key); isRegistered {
-		panic(fmt.Errorf("error: already registered: %s", key))
+	if prev, isRegistered := registeredErrors.Load(key); isRegistered {
+		panic(fmt.Errorf("error: already registered: %s %s", key, prev))
 	}
 	registeredErrors.Store(key, e)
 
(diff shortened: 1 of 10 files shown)


go/common/errors/errors.go
```diff
@@ -51,8 +51,8 @@ func New(module string, code uint32, msg string) error {
 	}
 
 	key := errorKey(module, code)
-	if _, isRegistered := registeredErrors.Load(key); isRegistered {
-		panic(fmt.Errorf("error: already registered: %s", key))
+	if prev, isRegistered := registeredErrors.Load(key); isRegistered {
+		panic(fmt.Errorf("error: already registered: %s %s", key, prev))
```
kostko:
```suggestion
		panic(fmt.Errorf("error: already registered: %s (existing: %s)", key, prev))
```

go/consensus/tendermint/abci/context.go
```diff
@@ -77,12 +77,14 @@ type Context struct {
 }
 
 // NewMockContext creates a new mock context for use in tests.
-func NewMockContext(mode ContextMode, now time.Time) *Context {
+func NewMockContext(mode ContextMode, now time.Time, state *iavl.MutableTree) *Context {
```
kostko: We should remove this now that we have mock application state (I actually removed it in #2674). Can you use `NewMockApplicationState` in tests instead as it already handles creating a new in-memory tree for you?

pro-wh:
could we ditch this and program in the locked funds in terms of debonding entries?

pro-wh:
^ #2747

pro-wh:
we'll go without this feature as implemented. the current idea is to put locked tokens in debonding delegations at genesis. we added a test for this use case in #2747

PR #2741, opened 2020-02-27
Increase scheduler max batch size

nhynes:
The value of 16M was taken from the [the CLI flags](https://github.com/oasislabs/oasis-core/blob/master/go/oasis-node/cmd/registry/runtime/runtime.go#L525)


pro-wh:
@matevz, any argument for the current value of 1000?

PR #2742, opened 2020-02-27
Increase leniency of semver parsing

nhynes:
Not all sources of semvers provide three fields. For instance, [runtime.Version()](https://golang.org/pkg/runtime/#Version) will return only major and minor when it's part of a release build ([usage](https://golang.org/pkg/runtime/#Version)). This fixes oasis-node startup issues.

### DIFF
--- a/go/common/version/version.go
+++ b/go/common/version/version.go
@@ -101,18 +101,18 @@ var Versions = struct {
 }
 
 func parseSemVerStr(s string) Version {
-	split := strings.Split(s, ".")
-	if len(split) != 3 {
-		panic("version: failed to split SemVer")
-	}
+	split := strings.SplitN(s, ".", 4)
 
-	var semVers []uint16
-	for _, v := range split {
-		i, err := strconv.ParseUint(v, 10, 16)
+	var semVers []uint16 = []uint16{0, 0, 0}
+	for i, v := range split {
+		if i >= 3 {
+			break
+		}
+		ver, err := strconv.ParseUint(v, 10, 16)
 		if err != nil {
 			panic("version: failed to parse SemVer: " + err.Error())
 		}
-		semVers = append(semVers, uint16(i))
+		semVers[i] = uint16(ver)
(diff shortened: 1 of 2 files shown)


go/common/version/version.go
```diff
@@ -102,17 +102,17 @@ var Versions = struct {
...
 func parseSemVerStr(s string) Version {
 	split := strings.Split(s, ".")
-	if len(split) != 3 {
-		panic("version: failed to split SemVer")
-	}
 
-	var semVers []uint16
-	for _, v := range split {
-		i, err := strconv.ParseUint(v, 10, 16)
+	var semVers []uint16 = []uint16{0, 0, 0}
```
nhynes: or maybe `nil`?

nhynes:
mdlint is giving spurious errors
`.changelog/2742.bugfix.md:4:6 MD034/no-bare-urls Bare URL used [Context: "https://golang.org/pkg/runtime..."`

pro-wh:
Does Go claim to use semver?

go/common/version/version.go
pro-wh: I like that this preallocates

go/common/version/version.go
```diff
@@ -102,17 +102,17 @@ var Versions = struct {
...
+	for i, v := range split {
+		if i >= 3 {
```
pro-wh: might also be good to use https://golang.org/pkg/strings/#SplitN

pro-wh:
if I search "semver," I find this https://semver.org/#spec-item-2:

> A normal version number MUST take the form X.Y.Z

and then they don't explain what "normal" refers to

go/common/version/version.go
nhynes: minor performance enhancement

go/common/version/version.go
nhynes: no, I mean like `Option<u16>`

PR #2745, opened 2020-02-28
Smaller docs improvements

tjanez:
Mostly a clarification on how to set up Rust and take advantage of the newly added `rust-toolchain` file which rustup uses to download and install the correct version of the Rust toolchain automatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,14 +54,13 @@ Prerequisites:
   [see below](#using-the-development-docker-image)).
 
 * System packages:
-  * [Bubblewrap](https://github.com/projectatomic/bubblewrap) (at least version
-    0.3.3).
-  * [GCC](http://gcc.gnu.org/) (including C++ subpackage).
-  * [Protobuf](https://github.com/protocolbuffers/protobuf) compiler.
-  * [GNU Make](https://www.gnu.org/software/make/).
-  * [CMake](https://cmake.org/).
-  * [OpenSSL](https://www.openssl.org/) development package.
-  * [libseccomp](https://github.com/seccomp/libseccomp) development package.
+  * [Bubblewrap] (at least version 0.3.3).
+  * [GCC] (including C++ subpackage).
+  * [Protobuf] compiler.
+  * [GNU Make].
+  * [CMake].
+  * [OpenSSL] development package.
+  * [libseccomp] development package.
 
   On Fedora 29+, you can install all the above with:
 
@@ -78,7 +77,7 @@ Prerequisites:
   sudo apt install bubblewrap gcc g++ protobuf-compiler make cmake libssl-dev libseccomp-dev
   ```
 
-* [Go](https://golang.org) (at least version 1.13.8).
+* [Go] (at least version 1.13.8).
 
   If your distribution provides a new-enough version of Go, just use that.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -101,36 +100,56 @@ Prerequisites:
...
-  Once you have [`rustup` installed](https://www.rust-lang.org/tools/install),
-  install the nightly with:
+  We follow [Rust upstream's recommendation][rust-upstream-rustup] on using
+  [rustup] to install and manage Rust versions.
 
-  ```
-  rustup install nightly
-  ```
+  _NOTE: [rustup] cannot be installed alongside a distribution packaged [Rust]
+  version. You will need to remove it before you can start using [rustup]._
```
pro-wh: Doesn't it get weird to link every single use of a term?

README.md
```diff
@@ -101,36 +100,56 @@ Prerequisites:
...
 
   ```
-  rustup target add x86_64-fortanix-unknown-sgx
+  cargo install fortanix-sgx-tools sgxs-tools
   ```
 
-  Install its utilities by running:
+  _NOTE: These utilities must be compiled with a nightly version of the [Rust]
+  toolchain since they use the `#![feature]` macro.
+  Make sure you run the above command from the Oasis Core git checkout or by
```
pro-wh: is this true of the farther above rustup install too?

README.md
```diff
@@ -101,36 +100,56 @@ Prerequisites:
...
+
+  Make sure you are in Oasis Core git checkout directory so that [rustup] will
+  add the Fortanix SGX [Rust] target to the [Rust] toolchain specified in
+  Oasis Core's [rust-toolchain] file.
+
+  Then install it by running:
 
   ```
-  cargo install fortanix-sgx-tools sgxs-tools
+  rustup target add x86_64-fortanix-unknown-sgx
```
pro-wh: why are the `rustup target add` and `cargo install` reversed from before?

README.md
nhynes: wikipedia says yes

README.md
```diff
@@ -101,36 +100,56 @@ Prerequisites:
...
-  ```
+  _NOTE: [rustup] cannot be installed alongside a distribution packaged [Rust]
+  version. You will need to remove it before you can start using [rustup]._
 
-  Then make it the default version with:
+  Install it by running:
 
   ```
-  rustup default nightly
+  curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
```
nhynes:
```suggestion
  curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain nightly
```
if you're feeling terse

README.md
```diff
@@ -101,36 +100,56 @@ Prerequisites:
...
+  The version of the [Rust] toolchain we use in Oasis Core is specified in the
+  [rust-toolchain] file and will be downloaded and installed automatically.
+
+* [Fortanix Rust EDP] utilities.
+
+  Install them by running:
 
   ```
-  rustup target add x86_64-fortanix-unknown-sgx
+  cargo install fortanix-sgx-tools sgxs-tools
```
nhynes:
```suggestion
  cargo +nightly install fortanix-sgx-tools sgxs-tools
```
will avoid most version issues

README.md
nhynes: you only need the target for compiling to sgx. the sgx tools are binaries that compile to native. You'd be best off using `rustup target add --toolchain nightly-2020-02-16 x86_64-fortanix-unknown-sgx` if you don't set the default toolchain during the initial install.

README.md
tjanez: Yes, it is weird. I've fixed it.

README.md
tjanez: No. I've refactored the instructions and added an explicit step _Oasis Core's Rust toolchain version with Fortanix SGX target._ which is the only step where one needs to be in the Oasis Core git checkout directory.

README.md
tjanez: Thanks for the suggestions. I decided to leave it as is as I would prefer not to set the nightly as a default toolchain for all user's projects.

README.md
tjanez: Thanks, I added that.

README.md
tjanez: As @nhynes explained, the Fortanix SGX target is not needed for compiling the sgx tools, however, sgx tools are required for compiling to sgx.

I would like to avoid hard-coding a particular nightly version in the README because experience suggests we will forget to update it.
Hence, I tried to use such instructions that never need to specify a nightly version and only rely on the `rust-toolchain` file.

PR #2748, opened 2020-03-02
Add sanity check for registered entities' stake

tjanez:
Extract and generalize registry's staking sanity checks in `go/consensus/tendermint/apps/supplementarysanity.checkStakeClaims()` and move them to `go/registry/api.SanityCheckStake()` function.
    
Augment the checks to check if an entity has enough stake for all stake claims in the Genesis document to prevent panics at oasis-node start-up due to entities not having enough stake in the escrow to satisfy all their stake claims.

Example panic when an entity doesn't have enough stake to be registered:
```
panic: mux: InitChain: fatal error in application: '200_registry': registry: genesis entity registration failure: staking: insufficient stake

goroutine 1 [running]:
github.com/oasislabs/oasis-core/go/consensus/tendermint/abci.(*abciMux).InitChain(0xc0001f64b0, 0x0, 0xed5d4d490, 0x0, 0xc000fd3580, 0x32, 0xc00005ae40, 0xc000128fc0, 0x1, 0x1, ...)
	github.com/oasislabs/oasis-core/go@/consensus/tendermint/abci/mux.go:442 +0xedd
github.com/tendermint/tendermint/abci/client.(*localClient).InitChainSync(0xc00117ade0, 0x0, 0xed5d4d490, 0x0, 0xc000fd3580, 0x32, 0xc00005ae40, 0xc000128fc0, 0x1, 0x1, ...)
	github.com/tendermint/tendermint@v0.32.8/abci/client/local_client.go:223 +0x101
github.com/tendermint/tendermint/proxy.(*appConnConsensus).InitChainSync(0xc007dbca90, 0x0, 0xed5d4d490, 0x0, 0xc000fd3580, 0x32, 0xc00005ae40, 0xc000128fc0, 0x1, 0x1, ...)
	github.com/tendermint/tendermint@v0.32.8/proxy/app_conn.go:65 +0x6b
github.com/tendermint/tendermint/consensus.(*Handshaker).ReplayBlocks(0xc000fa3238, 0xa, 0x170000, 0x173efdc, 0x6, 0xc000fd3580, 0x32, 0x0, 0x0, 0x0, ...)
	github.com/tendermint/tendermint@v0.32.8/consensus/replay.go:318 +0x666
github.com/tendermint/tendermint/consensus.(*Handshaker).Handshake(0xc000fa3238, 0x1af0dc0, 0xc0001f8230, 0x203001, 0x203001)
	github.com/tendermint/tendermint@v0.32.8/consensus/replay.go:269 +0x485
github.com/tendermint/tendermint/node.doHandshake(0x1aef940, 0xc007d51da0, 0xa, 0x0, 0x173efdc, 0x6, 0xc000fd3580, 0x32, 0x0, 0x0, ...)
	github.com/tendermint/tendermint@v0.32.8/node/node.go:281 +0x19a
github.com/tendermint/tendermint/node.NewNode(0xc0000f2500, 0x1acc940, 0xc0000db360, 0xc00db685f0, 0x1aaad60, 0xc00d95ff20, 0xc00d838520, 0x18e8718, 0xc00db60700, 0x1ad6a40, ...)
	github.com/tendermint/tendermint@v0.32.8/node/node.go:596 +0x343
github.com/oasislabs/oasis-core/go/consensus/tendermint.(*tendermintService).lazyInit.func2(0xc001020500, 0x0)
	github.com/oasislabs/oasis-core/go@/consensus/tendermint/tendermint.go:1015 +0x2bf
github.com/oasislabs/oasis-core/go/consensus/tendermint.(*tendermintService).Start(0xc00056de00, 0x2586e60, 0x152f980)
	github.com/oasislabs/oasis-core/go@/consensus/tendermint/tendermint.go:233 +0x89
github.com/oasislabs/oasis-core/go/oasis-node/cmd/node.newNode(0x1741200, 0x0, 0x0, 0x0)
	github.com/oasislabs/oasis-core/go@/oasis-node/cmd/node/node.go:740 +0x238a
github.com/oasislabs/oasis-core/go/oasis-node/cmd/node.NewNode(...)
	github.com/oasislabs/oasis-core/go@/oasis-node/cmd/node/node.go:454
github.com/oasislabs/oasis-core/go/oasis-node/cmd/node.Run(0x258a700, 0xc000f754a0, 0x0, 0x2)
	github.com/oasislabs/oasis-core/go@/oasis-node/cmd/node/node.go:82 +0x32
github.com/spf13/cobra.(*Command).execute(0x258a700, 0xc00003c190, 0x2, 0x2, 0x258a700, 0xc00003c190)
	github.com/spf13/cobra@v0.0.5/command.go:830 +0x2aa
github.com/spf13/cobra.(*Command).ExecuteC(0x258a700, 0x3f, 0x0, 0x0)
	github.com/spf13/cobra@v0.0.5/command.go:914 +0x2fb
github.com/spf13/cobra.(*Command).Execute(...)
	github.com/spf13/cobra@v0.0.5/command.go:864
github.com/oasislabs/oasis-core/go/oasis-node/cmd.Execute()
	github.com/oasislabs/oasis-core/go@/oasis-node/cmd/root.go:46 +0x4b
main.main()
	github.com/oasislabs/oasis-core/go@/oasis-node/main.go:9 +0x20
```

Example panic when an entity doesn't have enough stake to have a node registered:

```
panic: mux: InitChain: fatal error in application: '200_registry': registry: genesis node registration failure: staking: insufficient stake

goroutine 1 [running]:
github.com/oasislabs/oasis-core/go/consensus/tendermint/abci.(*abciMux).InitChain(0xc000fca960, 0x0, 0xed5d4d490, 0x0, 0xc0001e80c0, 0x32, 0xc000530400, 0xc0012262a0, 0x1, 0x1, ...)
        github.com/oasislabs/oasis-core/go@/consensus/tendermint/abci/mux.go:442 +0xedd
github.com/tendermint/tendermint/abci/client.(*localClient).InitChainSync(0xc00f31c480, 0x0, 0xed5d4d490, 0x0, 0xc0001e80c0, 0x32, 0xc000530400, 0xc0012262a0, 0x1, 0x1, ...)
        github.com/tendermint/tendermint@v0.32.8/abci/client/local_client.go:223 +0x101
github.com/tendermint/tendermint/proxy.(*appConnConsensus).InitChainSync(0xc0004ea000, 0x0, 0xed5d4d490, 0x0, 0xc0001e80c0, 0x32, 0xc000530400, 0xc0012262a0, 0x1, 0x1, ...)
        github.com/tendermint/tendermint@v0.32.8/proxy/app_conn.go:65 +0x6b
github.com/tendermint/tendermint/consensus.(*Handshaker).ReplayBlocks(0xc000ff1238, 0xa, 0x170000, 0x173b41c, 0x6, 0xc0001e80c0, 0x32, 0x0, 0x0, 0x0, ...)
        github.com/tendermint/tendermint@v0.32.8/consensus/replay.go:318 +0x666
github.com/tendermint/tendermint/consensus.(*Handshaker).Handshake(0xc000ff1238, 0x1aed900, 0xc00052e1c0, 0x203000, 0x203000)
        github.com/tendermint/tendermint@v0.32.8/consensus/replay.go:269 +0x485
github.com/tendermint/tendermint/node.doHandshake(0x1aec380, 0xc0011c4780, 0xa, 0x0, 0x173b41c, 0x6, 0xc0001e80c0, 0x32, 0x0, 0x0, ...)
        github.com/tendermint/tendermint@v0.32.8/node/node.go:281 +0x19a
github.com/tendermint/tendermint/node.NewNode(0xc0001eadc0, 0x1ac9420, 0xc000535400, 0xc00dabb4d0, 0x1aa7840, 0xc00da76540, 0xc00d7a7400, 0x18e4d18, 0xc00daa95e0, 0x1ad3560, ...)
        github.com/tendermint/tendermint@v0.32.8/node/node.go:596 +0x343
github.com/oasislabs/oasis-core/go/consensus/tendermint.(*tendermintService).lazyInit.func2(0xc00000f3e0, 0x0)
        github.com/oasislabs/oasis-core/go@/consensus/tendermint/tendermint.go:1015 +0x2bf
github.com/oasislabs/oasis-core/go/consensus/tendermint.(*tendermintService).Start(0xc0004e0600, 0x2783e80, 0x152b880)
        github.com/oasislabs/oasis-core/go@/consensus/tendermint/tendermint.go:233 +0x89
github.com/oasislabs/oasis-core/go/oasis-node/cmd/node.newNode(0x173d700, 0x0, 0x0, 0x0)
        github.com/oasislabs/oasis-core/go@/oasis-node/cmd/node/node.go:740 +0x238a
github.com/oasislabs/oasis-core/go/oasis-node/cmd/node.NewNode(...)
        github.com/oasislabs/oasis-core/go@/oasis-node/cmd/node/node.go:454
github.com/oasislabs/oasis-core/go/oasis-node/cmd/node.Run(0x2787720, 0xc001047f60, 0x0, 0x2)
        github.com/oasislabs/oasis-core/go@/oasis-node/cmd/node/node.go:82 +0x32
github.com/spf13/cobra.(*Command).execute(0x2787720, 0xc0000ce160, 0x2, 0x2, 0x2787720, 0xc0000ce160)
        github.com/spf13/cobra@v0.0.5/command.go:830 +0x2aa
github.com/spf13/cobra.(*Command).ExecuteC(0x2787720, 0x3f, 0x0, 0x0)
        github.com/spf13/cobra@v0.0.5/command.go:914 +0x2fb
github.com/spf13/cobra.(*Command).Execute(...)
        github.com/spf13/cobra@v0.0.5/command.go:864
github.com/oasislabs/oasis-core/go/oasis-node/cmd.Execute()
        github.com/oasislabs/oasis-core/go@/oasis-node/cmd/root.go:46 +0x4b
main.main()
        github.com/oasislabs/oasis-core/go@/oasis-node/main.go:9 +0x20
```

### DIFF
--- a/go/registry/api/sanity_check.go
+++ b/go/registry/api/sanity_check.go
@@ -11,12 +11,18 @@ import (
 	"github.com/oasislabs/oasis-core/go/common/entity"
 	"github.com/oasislabs/oasis-core/go/common/logging"
 	"github.com/oasislabs/oasis-core/go/common/node"
+	"github.com/oasislabs/oasis-core/go/common/quantity"
 	epochtime "github.com/oasislabs/oasis-core/go/epochtime/api"
 	"github.com/oasislabs/oasis-core/go/oasis-node/cmd/common/flags"
+	staking "github.com/oasislabs/oasis-core/go/staking/api"
 )
 
 // SanityCheck does basic sanity checking on the genesis state.
-func (g *Genesis) SanityCheck(baseEpoch epochtime.EpochTime) error {
+func (g *Genesis) SanityCheck(
+	baseEpoch epochtime.EpochTime,
+	stakeLedger map[signature.PublicKey]*staking.Account,
+	stakeThresholds map[staking.ThresholdKind]quantity.Quantity,
+) error {
 	logger := logging.GetLogger("genesis/sanity-check")
 
 	if !flags.DebugDontBlameOasis() {
@@ -41,9 +47,26 @@ func (g *Genesis) SanityCheck(baseEpoch epochtime.EpochTime) error {
 	}
 
 	// Check nodes.
-	_, err = SanityCheckNodes(logger, &g.Parameters, g.Nodes, seenEntities, runtimesLookup, true, baseEpoch)
+	nodeLookup, err := SanityCheckNodes(logger, &g.Parameters, g.Nodes, seenEntities, runtimesLookup, true, baseEpoch)
 	if err != nil {
 		return err
+	}
+
+	if !g.Parameters.DebugBypassStake {
+		entities := []*entity.Entity{}
+		for _, ent := range seenEntities {
+			entities = append(entities, ent)
+		}
+		runtimes, err := runtimesLookup.AllRuntimes(context.Background())
+		if err != nil {
+			return fmt.Errorf("registry: sanity check failed: could not obtain all runtimes from runtimesLookup: %w", err)
+		}
+		nodes, err := nodeLookup.Nodes(context.Background())
+		if err != nil {
+			return fmt.Errorf("registry: sanity check failed: could not obtain node list from nodeLookup: %w", err)
+		}
+		// Check stake.
+		return SanityCheckStake(entities, stakeLedger, nodes, runtimes, stakeThresholds, true)
 	}
 
 	return nil
@@ -170,6 +193,127 @@ func SanityCheckNodes(
 	}
 
 	return nodeLookup, nil
+}
+
+// SanityCheckStake ensures entities' stake accumulator claims are consistent
+// with general state and entities have enough stake for themselves and all
+// their registered nodes and runtimes.
+func SanityCheckStake(
+	entities []*entity.Entity,
+	accounts map[signature.PublicKey]*staking.Account,
+	nodes []*node.Node,
+	runtimes []*Runtime,
+	stakeThresholds map[staking.ThresholdKind]quantity.Quantity,
+	isGenesis bool,
+) error {
+	// Entities' escrow accounts for checking claims and stake.
+	generatedEscrows := make(map[signature.PublicKey]*staking.EscrowAccount)
+
+	// Generate escrow account for all entities.
+	for _, entity := range entities {
+		var escrow *staking.EscrowAccount
+		acct, ok := accounts[entity.ID]
+		if ok {
+			// Generate an escrow account with the same active balance and shares number.
+			escrow = &staking.EscrowAccount{
+				Active: staking.SharePool{
+					Balance:     acct.Escrow.Active.Balance,
+					TotalShares: acct.Escrow.Active.TotalShares,
+				},
+			}
+		} else {
+			// No account is associated with this entity, generate an empty escrow account.
+			escrow = &staking.EscrowAccount{}
+		}
+
+		// Add entity stake claim.
+		escrow.StakeAccumulator.AddClaimUnchecked(StakeClaimRegisterEntity, []staking.ThresholdKind{staking.KindEntity})
+
+		generatedEscrows[entity.ID] = escrow
+	}
+
+	for _, node := range nodes {
+		// Add node stake claims.
+		generatedEscrows[node.EntityID].StakeAccumulator.AddClaimUnchecked(StakeClaimForNode(node.ID), StakeThresholdsForNode(node))
+	}
+	for _, rt := range runtimes {
+		// Add runtime stake claims.
+		generatedEscrows[rt.EntityID].StakeAccumulator.AddClaimUnchecked(StakeClaimForRuntime(rt.ID), StakeThresholdsForRuntime(rt))
+	}
+
+	// Compare entities' generated escrow accounts with actual ones.
+	for _, entity := range entities {
+		var generatedEscrow, actualEscrow *staking.EscrowAccount
+		generatedEscrow = generatedEscrows[entity.ID]
+		acct, ok := accounts[entity.ID]
+		if ok {
+			actualEscrow = &acct.Escrow
+		} else {
+			// No account is associated with this entity, generate an empty escrow account.
+			actualEscrow = &staking.EscrowAccount{}
+		}
+
+		if isGenesis {
+			// For a Genesis document, check if the entity has enough stake for all its stake claims.
+			// NOTE: We can't perform this check at an arbitrary point since the entity could
+			// reclaim its stake from the escrow but its nodes and/or runtimes will only be
+			// ineligible/suspended at the next epoch transition.
+			if err := generatedEscrow.CheckStakeClaims(stakeThresholds); err != nil {
+				expected := "unknown"
+				expectedQty, err2 := generatedEscrow.StakeAccumulator.TotalClaims(stakeThresholds, nil)
+				if err2 == nil {
+					expected = expectedQty.String()
+				}
+				return fmt.Errorf("insufficient stake for account %s (expected: %s got: %s): %w",
+					entity.ID,
+					expected,
+					generatedEscrow.Active.Balance,
+					err,
+				)
+			}
+		} else {
+			// Otherwise, compare the expected accumulator state with the actual one.
+			// NOTE: We can't perform this check for the Genesis document since it is not allowed to
+			// have non-empty stake accumulators.
+			expectedClaims := generatedEscrows[entity.ID].StakeAccumulator.Claims
+			actualClaims := actualEscrow.StakeAccumulator.Claims
+			if len(expectedClaims) != len(actualClaims) {
+				return fmt.Errorf("incorrect number of stake claims for account %s (expected: %d got: %d)",
+					entity.ID,
+					len(expectedClaims),
+					len(actualClaims),
+				)
+			}
+			for claim, expectedThresholds := range expectedClaims {
+				thresholds, ok := actualClaims[claim]
+				if !ok {
+					return fmt.Errorf("missing claim %s for account %s", claim, entity.ID)
+				}
+				if len(thresholds) != len(expectedThresholds) {
+					return fmt.Errorf("incorrect number of thresholds for claim %s for account %s (expected: %d got: %d)",
+						claim,
+						entity.ID,
+						len(expectedThresholds),
+						len(thresholds),
+					)
+				}
+				for i, expectedThreshold := range expectedThresholds {
+					threshold := thresholds[i]
+					if threshold != expectedThreshold {
+						return fmt.Errorf("incorrect threshold in position %d for claim %s for account %s (expected: %s got: %s)",
+							i,
+							claim,
+							entity.ID,
+							expectedThreshold,
+							threshold,
+						)
+					}
+				}
+			}
+		}
+	}
+
+	return nil
 }
 
 // Runtimes lookup used in sanity checks.
--- a/go/registry/api/sanity_check.go
+++ b/go/registry/api/sanity_check.go
@@ -41,15 +41,20 @@ func (g *Genesis) SanityCheck(baseEpoch epochtime.EpochTime) error {
 	}
 
 	// Check nodes.
-	return SanityCheckNodes(logger, &g.Parameters, g.Nodes, seenEntities, runtimesLookup, true, baseEpoch)
+	_, err = SanityCheckNodes(logger, &g.Parameters, g.Nodes, seenEntities, runtimesLookup, true, baseEpoch)
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
 
 // SanityCheckEntities examines the entities table.
 // Returns lookup of entity ID to the entity record for use in other checks.
 func SanityCheckEntities(logger *logging.Logger, entities []*entity.SignedEntity) (map[signature.PublicKey]*entity.Entity, error) {
 	seenEntities := make(map[signature.PublicKey]*entity.Entity)
-	for _, sent := range entities {
-		entity, err := VerifyRegisterEntityArgs(logger, sent, true)
+	for _, signedEnt := range entities {
+		entity, err := VerifyRegisterEntityArgs(logger, signedEnt, true)
 		if err != nil {
 			return nil, fmt.Errorf("entity sanity check failed: %w", err)
 		}
@@ -69,8 +74,8 @@ func SanityCheckRuntimes(
 ) (RuntimeLookup, error) {
 	// First go through all runtimes and perform general sanity checks.
 	seenRuntimes := []*Runtime{}
-	for _, srt := range runtimes {
-		rt, err := VerifyRegisterRuntimeArgs(params, logger, srt, isGenesis)
+	for _, signedRt := range runtimes {
+		rt, err := VerifyRegisterRuntimeArgs(params, logger, signedRt, isGenesis)
 		if err != nil {
 			return nil, fmt.Errorf("runtime sanity check failed: %w", err)
 		}
@@ -78,8 +83,8 @@ func SanityCheckRuntimes(
 	}
 
 	seenSuspendedRuntimes := []*Runtime{}
-	for _, srt := range suspendedRuntimes {
-		rt, err := VerifyRegisterRuntimeArgs(params, logger, srt, isGenesis)
+	for _, signedRt := range suspendedRuntimes {
+		rt, err := VerifyRegisterRuntimeArgs(params, logger, signedRt, isGenesis)
 		if err != nil {
 			return nil, fmt.Errorf("runtime sanity check failed: %w", err)
 		}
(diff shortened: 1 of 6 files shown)


Yawning:
I guess?  What's the planned resolution for this when the sanity check fails?

And what's up with the (likely unrelated) CI failure?

go/staking/api/sanity_check.go
```diff
@@ -73,21 +73,21 @@ func SanityCheckDelegations(account *Account, delegations map[signature.PublicKe
 	sharesExpected := account.Escrow.Active.TotalShares
 
 	if shares.Cmp(&sharesExpected) != 0 {
-		return fmt.Errorf("staking: sanity check failed: all shares of all delegations for account don't add up to account's total active shares in escrow")
+		return fmt.Errorf("staking: sanity check failed: all shares of all delegations (%s) for account with ID: %s don't add up to account's total active shares in escrow (%s)", shares, id, sharesExpected)
```
pro-wh: but the caller will put the ID in their message

also below

go/registry/api/sanity_check.go
```diff
@@ -43,6 +54,28 @@ func (g *Genesis) SanityCheck(baseEpoch epochtime.EpochTime) error {
...
+	seenEntities map[signature.PublicKey]*entity.Entity,
+	stakeLedger map[signature.PublicKey]*staking.Account,
+	stakeThresholds map[staking.ThresholdKind]quantity.Quantity,
+) error {
+	for _, entity := range seenEntities {
+		acct := stakeLedger[entity.ID]
+		if acct == nil {
+			return fmt.Errorf("registry: sanity check failed: no account associated with entity with ID: %v", entity.ID)
+		}
+		entityThreshold := stakeThresholds[staking.KindEntity]
```
pro-wh: could move this out of the loop. maybe even out of the function

go/registry/api/sanity_check.go
```diff
@@ -43,6 +54,28 @@ func (g *Genesis) SanityCheck(baseEpoch epochtime.EpochTime) error {
...
+func SanityCheckEntitiesStake(
+	logger *logging.Logger,
+	seenEntities map[signature.PublicKey]*entity.Entity,
+	stakeLedger map[signature.PublicKey]*staking.Account,
+	stakeThresholds map[staking.ThresholdKind]quantity.Quantity,
+) error {
+	for _, entity := range seenEntities {
+		acct := stakeLedger[entity.ID]
+		if acct == nil {
+			return fmt.Errorf("registry: sanity check failed: no account associated with entity with ID: %v", entity.ID)
```
pro-wh: could this be overkill in cases where the threshold is zero?

go/registry/api/sanity_check.go
tjanez: Good point, agreed.

go/registry/api/sanity_check.go
tjanez: Yea, I'll change it to only run `SanityCheckEntitiesStake()` if `stakeThresholds[staking.KindEntity]` > 0.

pro-wh:
ah yeah we are supposed to deregister them if they get slashed below the threshold

go/registry/api/sanity_check.go
tjanez: Done.

go/registry/api/sanity_check.go
tjanez: Done.

go/staking/api/sanity_check.go
tjanez: The thing is that `SanityCheckDelegations()` is also called in:
https://github.com/oasislabs/oasis-core/blob/a2db858c3e7c3edb87c36818837c96a2c96401b3/go/staking/api/sanity_check.go#L191-L200

where it didn't print out the account id or other useful info previously.

(Same for other functions, `SanityCheckDebondingDelegations()` and `SanityCheckAccountShares()`.)

I've updated `go/consensus/tendermint/apps/supplementarysanity/checks.go` to not return duplicate info.

tjanez:
This is ready for another review.

Updating various tests made this PR a bit larger, but I tried to split things down into commits, so I suggest reviewing by commits.

go/staking/api/sanity_check.go
pro-wh: oh, thanks

go/genesis/genesis_test.go
```diff
@@ -412,6 +436,53 @@ func TestGenesisSanityCheck(t *testing.T) {
...
+	// NOTE: We will break the staking part of testDoc and gradually fix it to
+	// test all related registry & staking genesis checks.
+	d = *testDoc
+	d.Staking.Ledger = make(map[signature.PublicKey]*staking.Account)
+	d.Staking.Delegations = make(map[signature.PublicKey]map[signature.PublicKey]*staking.Delegation)
+
+	d.Registry.Entities = []*entity.SignedEntity{signedTestEntity}
+	err = d.SanityCheck()
+	require.NotNil(err, "entity without an account should be rejected")
+	require.Regexp("no account associated with entity with ID", err)
```
pro-wh: I think this will be the first we have of this kind of check

go/genesis/genesis_test.go
```diff
@@ -412,6 +436,53 @@ func TestGenesisSanityCheck(t *testing.T) {
...
+
+	// Add account for test entity to staking ledger.
+	d.Staking.Ledger[validPK] = &staking.Account{
+		General: staking.GeneralAccount{
+			Balance: *quantity.NewQuantity(),
```
pro-wh: sometimes `*quantity.NewQuantity()`, sometimes `stakingTests.QtyFromInt(0)`

go/genesis/genesis_test.go
```diff
@@ -658,35 +729,48 @@ func TestGenesisSanityCheck(t *testing.T) {
 	// Test staking genesis checks.
 	// NOTE: There doesn't seem to be a way to generate invalid Quantities, so
 	// we're just going to test the code that checks if things add up.
+
+	// NOTE: We will replace the staking part of testDoc with the one from
+	// stakingTests and use stakingTests' entity.
```
pro-wh: why this?

go/registry/api/sanity_check.go
```diff
@@ -33,6 +39,14 @@ func (g *Genesis) SanityCheck(baseEpoch epochtime.EpochTime) error {
 		return err
 	}
 
+	// Check that all entities have enough stake to be registered.
+	qZero := quantity.NewQuantity()
+	if entityThreshold := stakeThresholds[staking.KindEntity]; entityThreshold.Cmp(qZero) > 0 {
```
pro-wh: in other places, we use `!qqqq.IsZero()`

tjanez:
> ah now we're not sure after all, if we want to delete an entity if it gets slashed below the threshold. cc @peterjgilbert @ravenac95

Hmm... are you worried about a potentially invalid state dump (i.e. genesis file)?

Do we currently de-register entities that fall below the staking thresholds when they are slashed?
If not, this could be a problem.

> code-wise though: this doesn't do a full accumulation. could this check miss cases where the amount escrowed is enough, but there isn't enough overall for the entity plus, for example, some nodes?

Good catch! Indeed, the current check, e.g. misses the case when an entity has enough stake to be registered itself, but not enough stake to also register its nodes.

We could augment the sanity checks to create a map of `StakeAccumulator`s for each entity and then add `StakeClaim`s as appropriate (i.e. one for the entity, one for each node's role, ... do we also want to add stake claims for compute and key manager runtimes?).
Finally, the sanity checks would check if an entity's escrow active balance is >= `StakeAccumulator.TotalClaims()`.

go/registry/api/sanity_check.go
tjanez: Aha, ok. I will update the code to use that form.

go/genesis/genesis_test.go
tjanez: We already have 2 checks like that :slightly_smiling_face: :
https://github.com/oasislabs/oasis-core/blob/798fbe38f4665a1cb78cec2d605bf84dcfa51f95/go/worker/common/host/sandboxed_test.go#L222-L234

tjanez: From my experience, just checking for a presence of an error and not checking for occurrence of a particular error is quite bad.

Frequently, a developer will break code in a new way and the tests will seem to pass even if they shouldn't because the error that occurred is not the one that was expected.

I plan to make a follow-up PR to update these tests to actually check for particular errors (and not just any error).

go/staking/tests/tester.go
```diff
@@ -326,25 +397,26 @@ func testEscrowEx(
...
-	_ = dstAcc.Escrow.Active.Balance.Add(&escrow.Tokens)
 	newDstAcc, err := backend.AccountInfo(context.Background(), &api.OwnerQuery{Owner: dstID, Height: consensusAPI.HeightLatest})
 	require.NoError(err, "dst: AccountInfo - after")
 	if !srcID.Equal(dstID) {
 		require.Equal(dstAcc.General.Balance, newDstAcc.General.Balance, "dst: general balance - after")
 		require.Equal(dstAcc.General.Nonce, newDstAcc.General.Nonce, "dst: nonce - after")
 	}
 	require.Equal(dstAcc.Escrow.Active.Balance, newDstAcc.Escrow.Active.Balance, "dst: active escrow balance - after")
-	require.Equal(dstAcc.Escrow.Active.Balance, newDstAcc.Escrow.Active.TotalShares, "dst: active escrow total shares - after")
+	require.Equal(dstAcc.Escrow.Active.TotalShares, newDstAcc.Escrow.Active.TotalShares, "dst: active escrow total shares - after")
```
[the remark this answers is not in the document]
tjanez: Indeed, since tokens and shares don't map 1:1, I updated the code to check `TotalShares` instead of `Balance` when checking the amount of shares in the account.

go/registry/api/sanity_check.go
tjanez: Done.

go/genesis/genesis_test.go
tjanez: I've updated the code to consistently use `*quantity.NewQuantity()`.

go/genesis/genesis_test.go
tjanez: The `testDoc` used `DebugGenesisState` from `go/staking/tests/debug/debug_stake.go` before:
https://github.com/oasislabs/oasis-core/blob/798fbe38f4665a1cb78cec2d605bf84dcfa51f95/go/genesis/genesis_test.go#L70-L72

Using that for the newly added tests for registry genesis checks was not practical, but I kept them for the staking genesis checks.

I don't have a strong preference though. We could also adopt the staking genesis checks to use the new `testDoc` without replacing its staking part.

go/staking/tests/tester.go
```diff
@@ -83,45 +155,40 @@ func StakingClientImplementationTests(t *testing.T, backend api.Backend, consens
...
 
 	acc, err := backend.AccountInfo(context.Background(), &api.OwnerQuery{Owner: SrcID, Height: consensusAPI.HeightLatest})
 	require.NoError(err, "src: AccountInfo")
-	require.Equal(testTotalSupply, acc.General.Balance, "src: general balance")
-	require.True(acc.Escrow.Active.Balance.IsZero(), "src: active escrow balance")
+	require.Equal(state.srcAccountGeneralBalance, acc.General.Balance, "src: general balance")
+	require.Equal(state.srcAccountEscrowActiveBalance, acc.Escrow.Active.Balance, "src: active escrow balance")
 	require.True(acc.Escrow.Debonding.Balance.IsZero(), "src: debonding escrow balance")
-	require.EqualValues(0, acc.General.Nonce, "src: nonce")
+	require.EqualValues(state.srcAccountNonce, acc.General.Nonce, "src: nonce")
```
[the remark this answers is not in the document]
tjanez: :slightly_smiling_face:  The tests' state is initialized via fixed values (i.e. without calling `update()`).

The subsequent runs are indeed tautological since the `update()` method uses the same functions to update the state as are used here to check the state.

Should we only run `testInitialEnv()` on the first run of the tests? Should we skip it altogether?

go/registry/api/sanity_check.go
```diff
@@ -147,43 +175,151 @@ func SanityCheckNodes(
...
+
+		// Add entity stake claim.
+		escrow.StakeAccumulator.AddClaimUnchecked(StakeClaimRegisterEntity, []staking.ThresholdKind{staking.KindEntity})
+
+		generatedEscrows[entity.ID] = escrow
+	}
+
+	for _, node := range nodes {
+		// Add node stake claims.
+		generatedEscrows[node.EntityID].StakeAccumulator.AddClaimUnchecked(StakeClaimForNode(node.ID), StakeThresholdsForNode(node))
```
pro-wh: it's already checked that `node.EntityID` is a valid entity ID, right?

go/registry/api/sanity_check.go
```diff
@@ -147,43 +175,151 @@ func SanityCheckNodes(
...
+		}
+
+		// Check if entity has enough stake for all stake claims.
+		if err := generatedEscrow.CheckStakeClaims(stakeThresholds); err != nil {
+			var expected = "unknown"
+			expectedQty, err := generatedEscrow.StakeAccumulator.TotalClaims(stakeThresholds, nil)
+			if err == nil {
+				expected = expectedQty.String()
+			}
+			return fmt.Errorf("insufficient stake for account %s (expected: %s got: %s)", entity.ID, expected, generatedEscrow.Active.Balance)
```
pro-wh: could we `%w` the error from `CheckStateClaims`?

or explain why we'd rather throw away that info

go/registry/api/sanity_check.go
tjanez: Yes, it is already checked by `SanityCheckEntities()` which in turn calls `VerifyRegisterEntityArgs()` for each signed entity which finally checks the ID in:
https://github.com/oasislabs/oasis-core/blob/b9fe8b7e4ea570b9aa69ec549ec1f3da5ffdcd70/go/registry/api/api.go#L363-L370

go/registry/api/sanity_check.go
tjanez: Good catch. I've added `%w` to the `fmt.Errorf()` call.

tjanez:
> Good catch! Indeed, the current check, e.g. misses the case when an entity has enough stake to be registered itself, but not enough stake to also register its nodes.
> 
> We could augment the sanity checks to create a map of `StakeAccumulator`s for each entity and then add `StakeClaim`s as appropriate (i.e. one for the entity, one for each node's role, ... do we also want to add stake claims for compute and key manager runtimes?).
> Finally, the sanity checks would check if an entity's escrow active balance is >= `StakeAccumulator.TotalClaims()`.

I've implemented that, please take a look.

go/registry/api/sanity_check.go
```diff
@@ -151,43 +179,173 @@ func SanityCheckNodes(
...
+		} else {
+			// No account is associated with this entity, generate an empty escrow account.
+			actualEscrow = &staking.EscrowAccount{}
+		}
+
+		if isGenesis {
+			// For a Genesis document, check if the entity has enough stake for all its stake claims.
+			// NOTE: We can't perform this check at an arbitrary point since the entity could
+			// reclaim its stake from the escrow but its nodes and/or runtimes will only be
+			// deregistered/suspend at the next epoch transition.
```
kostko, line 260:
```suggestion
			// ineligible/suspended at the next epoch transition.
```

tjanez:
> Hmm... are you worried about a potentially invalid state dump (i.e. genesis file)?
> 
> Do we currently de-register entities that fall below the staking thresholds when they are slashed?
> If not, this could be a problem.

I've filled a follow-up issue that will address this: #2886.

go/registry/api/sanity_check.go
```diff
@@ -151,43 +179,173 @@ func SanityCheckNodes(
...
 		nodeLookup.nodesCertHashes[h] = node
 	}
 
+	return nodeLookup, nil
+}
+
+// SanityCheckStake ensures entities' stake accumulator claims are consistent
+// with general state and entities have enough stake for themselves and all
+// their registered nodes and runtimes.
+func SanityCheckStake(
```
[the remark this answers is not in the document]
tjanez: Agreed :+1:.

tjanez:
@pro-wh, could you take another look?

PR #2749, opened 2020-03-02
Allow runtimes to have a custom dispatcher

### LINKED ISSUE #2281: Allow runtimes to have a custom dispatcher
Runtimes should be allowed to customize the batch dispatch process.

*Estimated cost: 1 sprint*

jberci:
Closes #2281.

### DIFF
--- a/runtime/src/dispatcher.rs
+++ b/runtime/src/dispatcher.rs
@@ -40,7 +40,9 @@ use crate::{
         StorageContext,
     },
     transaction::{
-        dispatcher::Dispatcher as TxnDispatcher, tree::Tree as TxnTree, types::TxnBatch,
+        dispatcher::{Dispatcher as TxnDispatcher, NoopDispatcher as TxnNoopDispatcher},
+        tree::Tree as TxnTree,
+        types::TxnBatch,
         Context as TxnContext,
     },
     types::{Body, ComputedBatch},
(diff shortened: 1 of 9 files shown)


runtime/src/dispatcher.rs
```diff
@@ -155,16 +160,13 @@ impl Dispatcher {
...
-                &mut rpc_dispatcher,
-                &mut txn_dispatcher,
-            );
-        }
+        let mut txn_dispatcher: Box<dyn TxnDispatcher> = if let Some(txn) =
+            initializer.init(&protocol, &self.rak, &mut rpc_demux, &mut rpc_dispatcher)
+        {
+            txn
+        } else {
+            Box::new(TxnMethDispatcher::new())
```
pro-wh: what if we made the initializer do this internally

runtime/src/transaction/dispatcher.rs
```diff
@@ -216,8 +232,34 @@ impl Dispatcher {
         self.finalizer = Some(Box::new(finalizer));
     }
 
-    /// Dispatches a batch of runtime requests.
-    pub fn dispatch_batch(
+    /// Dispatches a raw runtime invocation request.
+    pub fn dispatch(&self, call: &Vec<u8>, ctx: &mut Context) -> Vec<u8> {
```
pro-wh: are there any external users of this?

runtime/src/dispatcher.rs
jberci: Not sure we can... whoever constructs a concrete dispatcher will need its specific interface, so it would be problematic if the initializer itself returned the general boxed trait already.

runtime/src/transaction/dispatcher.rs
jberci: Not that I could see...

runtime/src/dispatcher.rs
pro-wh: ok what about yet another implementation that's specifically a no-op dispatcher

PR #2753, opened 2020-03-05
Remove locked accounts

pro-wh:
We expect not to need this feature.

### DIFF
--- a/go/consensus/tendermint/apps/staking/transactions.go
+++ b/go/consensus/tendermint/apps/staking/transactions.go
@@ -1,28 +1,22 @@
 package staking
 
 import (
-	"fmt"
-
 	"github.com/oasislabs/oasis-core/go/common/cbor"
 	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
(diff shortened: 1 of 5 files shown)


go/consensus/tendermint/apps/staking/transactions.go
```diff
@@ -37,20 +31,17 @@ func (app *stakingApplication) transfer(ctx *abci.Context, state *stakingState.M
 	if err != nil {
 		return err
 	}
-	if err = ctx.Gas().UseGas(1, staking.GasOpTransfer, params.GasCosts); err != nil {
+	if err := ctx.Gas().UseGas(1, staking.GasOpTransfer, params.GasCosts); err != nil {
```
Yawning: Shouldn't need this change, `err` is already declared.

pro-wh:
locked accounts will become unlocked. unless we hooked up the parser to error out on unknown fields.

incidentally, we never saw any public network use this feature

go/consensus/tendermint/apps/staking/transactions.go
pro-wh: ok

PR #2758, opened 2020-03-11
abci fuzzing workspace

pro-wh:
where the fuzzer specifies ~~the genesis document and~~ tendermint events on the consensus connection


pro-wh:
that's it for fuzzing through abci

PR #2760, opened 2020-03-12
go/staking/api: Add more details to sanity check errors

tjanez:
Extracted from #2748.

### DIFF
--- a/go/consensus/tendermint/apps/supplementarysanity/checks.go
+++ b/go/consensus/tendermint/apps/supplementarysanity/checks.go
@@ -149,10 +149,9 @@ func checkStaking(state *iavl.MutableTree, now epochtime.EpochTime) error {
 	if err != nil {
 		return fmt.Errorf("Delegations: %w", err)
 	}
-	for acct, delegations := range delegationses {
-		err = staking.SanityCheckDelegations(st.Account(acct), delegations)
-		if err != nil {
-			return fmt.Errorf("SanityCheckDelegations %s: %w", acct, err)
+	for id, delegations := range delegationses {
+		if err = staking.SanityCheckDelegations(id, st.Account(id), delegations); err != nil {
+			return err
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


go/consensus/tendermint/apps/supplementarysanity/checks.go
```diff
@@ -149,10 +149,9 @@ func checkStaking(state *iavl.MutableTree, now epochtime.EpochTime) error {
 	if err != nil {
 		return fmt.Errorf("Delegations: %w", err)
 	}
-	for acct, delegations := range delegationses {
-		err = staking.SanityCheckDelegations(st.Account(acct), delegations)
-		if err != nil {
-			return fmt.Errorf("SanityCheckDelegations %s: %w", acct, err)
+	for id, delegations := range delegationses {
+		if err = staking.SanityCheckDelegations(id, st.Account(id), delegations); err != nil {
+			return err
```
pro-wh: :information_source: responsibility for reporting which account in the error is pushed down to SanityCheckDelegations. This PR makes similar changes for other functions too.

go/staking/api/sanity_check.go
```diff
@@ -73,21 +73,21 @@ func SanityCheckDelegations(account *Account, delegations map[signature.PublicKe
 	sharesExpected := account.Escrow.Active.TotalShares
 
 	if shares.Cmp(&sharesExpected) != 0 {
-		return fmt.Errorf("staking: sanity check failed: all shares of all delegations for account don't add up to account's total active shares in escrow")
+		return fmt.Errorf("staking: sanity check failed: all shares of all delegations (%s) for account with ID: %s don't add up to account's total active shares in escrow (%s)", shares, id, sharesExpected)
```
pro-wh: :information_source: in some cases, error diagnostics is the only thing that uses the `id` parameter. but adding the argument also makes these sanity check functions uniform with ones that already had an `id` argument.

go/staking/api/sanity_check.go
```diff
@@ -189,33 +189,30 @@ func (g *Genesis) SanityCheck(now epochtime.EpochTime) error { // nolint: gocycl
...
+		if err := SanityCheckDebondingDelegations(id, acct, delegations); err != nil {
 			return err
 		}
 	}
 
 	// Check the above two invariants for each account as well.
 	for id, acct := range g.Ledger {
-		err := SanityCheckAccountShares(acct, g.Delegations[id], g.DebondingDelegations[id])
-		if err != nil {
+		if err := SanityCheckAccountShares(id, acct, g.Delegations[id], g.DebondingDelegations[id]); err != nil {
```
pro-wh: :information_source: some error checking code is changed to this more compact form

PR #2763, opened 2020-03-13
staking/api/commission: fix possible panic in validation check

ptrus:
The validation check would panic whenever the number of bound steps was greater than `rate_steps + 2`.

### DIFF
--- a/.changelog/2763.bugfix.md
+++ b/.changelog/2763.bugfix.md
@@ -1,0 +1,4 @@
+staking/api/commission: fix possible panic in validation check
+
+The validation check would panic whenever the number of bound steps was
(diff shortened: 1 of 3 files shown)


Yawning:
Add an error case test?

ptrus:
> Add an error case test?

Not sure wym. The panic happened regardless if the bounds were vaild or not. The updated test does test this (e.g. it panics on old code & no error in new code). 

Also the specific error is already tested here: https://github.com/oasislabs/oasis-core/blob/f2dee245b174ac0384908010a498e0fe9a9650a3/go/staking/api/commission_test.go#L521

PR #2767, opened 2020-03-13
go/staking/tests: Add escrow and delegations to debug genesis state

tjanez:
Extracted from #2748.

Introduce `stakingTestsState` that holds the current state of staking tests.

Since staking tests are run as part of `TestNode` tests which spin up a real node, subsequent tests are run in a modified (i.e. non-genesis) state.

This change enables the staking implementation tests (`StakingImplementationTest`, `StakingClientImplementationTests`) to use a new `stakingTestsState` that is initialized to the current staking state.

Remove `testInitialEnv()` since it has become tautological.

Extract remaining useful checks from the removed `testInitialEnv()`:
- `testThresholds()`,
- `testLastBlockFees()`.

### DIFF
--- a/go/staking/tests/tester.go
+++ b/go/staking/tests/tester.go
@@ -1,4 +1,4 @@
-// Pacakge tests is a collection of staking token backend implementation tests.
+// Package tests is a collection of staking token backend implementation tests.
 package tests
 
 import (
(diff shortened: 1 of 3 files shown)


go/staking/tests/tester.go
```diff
@@ -83,51 +138,43 @@ func StakingClientImplementationTests(t *testing.T, backend api.Backend, consens
 	} {
 		t.Run(tc.n, func(t *testing.T) { tc.fn(t, backend, consensus) })
 	}
+
+	// Update tests' state for subsequent node tests (which re-use the state).
+	state.update(backend, consensus)
 }
 
 func testInitialEnv(t *testing.T, backend api.Backend, consensus consensusAPI.Backend) {
```
tjanez: The comment from https://github.com/oasislabs/oasis-core/pull/2748#discussion_r389916325 on `testInitialEnv()` becoming tautological still applies.

I don't have a strong opinion on whether we should leave it as is or perhaps:
1. Add a `firstRun` variable to the state and only run the tests in `testInitialEnv()` the first time?
2. Remove `testInitialEnv()` altogether?

pro-wh: what runs before this subtest to alter the state?

go/staking/tests/tester.go
```diff
@@ -26,11 +26,62 @@ import (
...
+	srcAccountEscrowActiveBalance quantity.Quantity
+	srcAccountEscrowActiveShares  quantity.Quantity
+
+	destAccountGeneralBalance quantity.Quantity
+}
+
+func (s *stakingTestsState) update(backend api.Backend, consensus consensusAPI.Backend) {
+	totalSupply, err := backend.TotalSupply(context.Background(), consensusAPI.HeightLatest)
+	if err != nil {
+		panic(err)
```
Yawning: Why should this panic?  make it take a `* testing.T`, and use assertions.

tjanez: I have no strong preferences. Ok, I'll make it use ordinary testing assertions.

tjanez: Done.

go/staking/tests/tester.go
tjanez: `TestNode` first runs `Staking` tests and then `StakingClient` tests:
https://github.com/oasislabs/oasis-core/blob/e7d25bf0c9c69c22de5268633182e60a948d828f/go/oasis-node/node_test.go#L238-L240

These functions in turn run `StakingImplementationTest` and `StakingClientImplementationTests` which internally run (almost) the same tests on the same state:
https://github.com/oasislabs/oasis-core/blob/e7d25bf0c9c69c22de5268633182e60a948d828f/go/staking/tests/tester.go#L40-L69

https://github.com/oasislabs/oasis-core/blob/e7d25bf0c9c69c22de5268633182e60a948d828f/go/staking/tests/tester.go#L71-L86

Hence, the need to update the state so that individual test functions (i.e. `testTransfer`, `testSelfTransfer`, ...) can work the second (and 3rd, ...) time.

tjanez: @pro-wh, ping

pro-wh: ok probably go without testInitialEnv then. I'm starting to get the feeling it was there as a cheap solution before something like this `stakingTestsState` system that you're adding.

maybe in the future we'll even have a way to run tests in their own independent state

tjanez: > ok probably go without testInitialEnv then. I'm starting to get the feeling it was there as a cheap solution before something like this `stakingTestsState` system that you're adding.

Ok, agreed.

> maybe in the future we'll even have a way to run tests in their own independent state

Yes, that would be very neat.

tjanez: I ended up extracting the staking thresholds checks from `testInitialEnv()` as `testThresholds()` since they seem useful.

Please, take another look.

tjanez:
> idea though: maybe make the `state` local to a test, where we make a zeroed out one at the beginning of a test, call `update`, and then use it within the test.
> 
> currently we have a global one that we call `update` on at the end of tests

Good idea, I've implemented it.

Please, take another look.

go/staking/tests/tester.go
pro-wh: testThresholds looks good

go/staking/tests/tester.go
```diff
@@ -48,86 +89,58 @@ func StakingImplementationTests(
 	entitySigner signature.Signer,
 	runtimeID common.Namespace,
 ) {
+	state := newStakingTestsState(t, backend, consensus)
```
pro-wh: (here and below) shouldn't this be on the inside of the loop? or do tests mutate this?

tjanez: Currently, tests don't mutate state but yes, perhaps it would be even cleaner to create a new state for each test that is run.

pro-wh: but the tests transfer things around and then nothing calls `update`, so wouldn't that go wrong?

tjanez: > but the tests transfer things around and then nothing calls `update`, so wouldn't that go wrong?

Yea, tests move things around but they currently handle account updates manually, e.g.:
https://github.com/oasislabs/oasis-core/blob/5435fade6371883b3dc215cd7fec8737b91997a9/go/staking/tests/tester.go#L355-L358

PR #2759, opened 2020-03-11
txsource: runtime workload

ptrus:
Runtime part of: #2506

TODO: 
- [x] validate runtime transaction results

PR #2769, opened 2020-03-16
txsource/queries: workload doing various queries

ptrus:
Queries workload, part of: #2506

- ~Also removes `GetNodeList()` client query as it's almost identical to the `GetNodes` one (minus the sorting) and it's not used anywhere at the moment - can revert that commit if desired.~ (extracted out)
- adds `LastBlockFees` query method to staking client, so that client is able to obtain all funds and confirm total supply matches correctly

Feel free to comment in case any relevant queries/checks should be added.

TODO:
- [x] bunch of todos in code
- [x] requires #2759  merged in


pro-wh:
> Also removes GetNodeList() client query as it's almost identical to the GetNodes one (minus the sorting) and it's not used anywhere at the moment - can revert that commit if desired.

weird, why did we even have both?

go/oasis-node/cmd/debug/txsource/workload/queries.go
```diff
@@ -0,0 +1,455 @@
...
+	}
+
+	for {
+		block, err := q.consensus.GetBlock(ctx, consensus.HeightLatest)
+		if err != nil {
+			return fmt.Errorf("consensus.GetBlock error: %w", err)
+		}
+
+		// Select height at which queries should be done. Earliest (height=1)
+		// is special cased with increased probability.
```
pro-wh: why extra likely on the earliest?

should we also make it extra likely to ask for the latest?

go/oasis-node/cmd/debug/txsource/workload/queries.go
```diff
@@ -0,0 +1,455 @@
...
+		var height int64
+		p := rng.Float32()
+		switch {
+		case p < queriesEarliestHeightRatio:
+			height = 1
+		default:
+			// [1, block.Height]
+			height = rng.Int63n(block.Height) + 1
```
pro-wh: we don't exercise passing `0` for the latest. probably due to the routines doing multiple queries with this parameter and requiring some consistency

go/consensus/tendermint/registry/registry.go
```diff
@@ -125,10 +125,6 @@ func (tb *tendermintBackend) WatchRuntimes(ctx context.Context) (<-chan *api.Run
 	return typedCh, sub, nil
 }
 
-func (tb *tendermintBackend) GetNodeList(ctx context.Context, height int64) (*api.NodeList, error) {
-	return tb.getNodeList(ctx, height)
```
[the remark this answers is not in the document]
ptrus: the `WatchNodeLists` which now that i looked into it also isn't used anywhere, maybe we should just get rid of it all together yes, in any case i think it would be better extracting this in a separate  PR, so will remove the changes from here.

go/oasis-node/cmd/debug/txsource/workload/queries.go
ptrus: We could (however not `consensus.HeightLatest` without special casing that case in some of the checks). I actually had that implemented as well but then figured that most of the latest queries are already implicitly tested by other workloads and sanity checks, and that this workload should instead focus on querying older data (as it was mentioned in the initial idea for it: https://github.com/oasislabs/oasis-core/issues/2506#issuecomment-594524062).

With that in mind, special casing the height 1 still does make some sense i think.

go/oasis-node/cmd/debug/txsource/workload/queries.go
ptrus: Yeah and also since that should be (implicitly at least) already tested in other workloads / sanity checks. see: https://github.com/oasislabs/oasis-core/pull/2769#discussion_r395551939

go/oasis-node/cmd/debug/txsource/workload/queries.go
```diff
@@ -0,0 +1,455 @@
...
+	q.logger.Debug("Doing staking queries",
+		"height", height,
+	)
+
+	total, err := q.staking.TotalSupply(ctx, height)
+	if err != nil {
+		return fmt.Errorf("staking.TotalSupply: %w", err)
+	}
+	if total.IsZero() {
+		return fmt.Errorf("staking.TotalSupply is zero")
```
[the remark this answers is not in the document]
ptrus: good idea, will do

go/oasis-node/cmd/debug/txsource/txsource.go
```diff
@@ -120,7 +116,7 @@ func doRun(cmd *cobra.Command, args []string) error {
 	}
 
 	logger.Debug("entering workload", "name", name)
-	if err = w.Run(ctx, rng, conn, cnsc, rtc, fundingAccount); err != nil {
+	if err = w.Run(ctx, rng, conn, cnsc, fundingAccount); err != nil {
```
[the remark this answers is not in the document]
ptrus: agreed, can revert this change if we feel like we'll be adding more runtime workloads soon.

go/oasis-node/cmd/debug/txsource/workload/queries.go
```diff
@@ -0,0 +1,455 @@
...
+	// Query commiteess for the runtime.
+	committees, err := q.scheduler.GetCommittees(ctx, &scheduler.GetCommitteesRequest{
+		Height:    height,
+		RuntimeID: q.runtimeID,
+	})
+	if err != nil {
+		return fmt.Errorf("GetCommittees at height %d: %w", height, err)
+	}
+	// First two epochs shouldn't have runtime committees, following epochs
+	// should all have existing committess for the simple-keyvalue runtime.
```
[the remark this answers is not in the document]
ptrus: In the E2E-tests we currently have only validators in the genesis documents, meaning there will be no other committees in the first epoch (epoch=0). In epoch 0 the key-manager will register, but not yet compute/storage workers, since those are waiting for key-manager committee to be available, see: https://github.com/oasislabs/oasis-core/blob/7c800a8cf4c0350de769b59e45b0547499a8dce1/go/worker/common/committee/node.go#L291 (which will happen at start of epoch 1). 
This means that compute/storage nodes will get registered during epoch 1 and the committee should be elected from epoch 2 onward.

Maybe i should explain this in the comment, and move the threshold epoch at which committee is expected into a constant.

---

However indeed below should `if` should be changed to `if epoch > 1`.

.changelog/2769.feature.2.md
```diff
@@ -0,0 +1,3 @@
+go/staking: add LastBlockFees query method
+
+LastBlockFees returns the collected fees for previous block.
```
kostko: Slightly unrelated, but this is also something that should be part of the genesis state dump as otherwise it could cause the total token amount to become inconsistent on dump/restore.

ptrus: oh we don't do that at the moment :o will do a PR that adds that tomorrow :grin:

go/oasis-node/cmd/debug/txsource/txsource.go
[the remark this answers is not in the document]
pro-wh: ok let's put it back if we have more runtime tests

go/oasis-node/cmd/debug/txsource/workload/queries.go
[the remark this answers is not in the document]
pro-wh: thanks for the elaboration

go/oasis-node/cmd/debug/txsource/workload/queries.go
pro-wh: ok

go/oasis-node/cmd/debug/txsource/workload/queries.go
pro-wh: resolving with the above thread

ptrus:
thanks for the reviews @pro-wh

PR #2773, opened 2020-03-19
transfer workload include burns

ptrus:
- adds staking.Burn transactions to the transfer workload
- extracts repeated "estimateGas, fund caller, sign tx, submit tx" action into a helper function

### DIFF
--- a/go/oasis-node/cmd/debug/txsource/workload/transfer.go
+++ b/go/oasis-node/cmd/debug/txsource/workload/transfer.go
@@ -19,19 +19,94 @@ import (
 
 const (
 	// NameTransfer is the name of the transfer workload.
+	//
+	// Transfer workload continiously submits transfer and burn transactions.
 	NameTransfer = "transfer"
 
-	transferNumAccounts  = 10
-	transferAmount       = 1
-	transferFundInterval = 10
-	transferGasCost      = 10
+	transferNumAccounts = 10
+	transferAmount      = 1
+	transferBurnAmount  = 10
 )
 
-var transferLogger = logging.GetLogger("cmd/txsource/workload/transfer")
+type transfer struct {
+	logger *logging.Logger
 
-type transfer struct{}
+	consensus consensus.ClientBackend
 
-func (transfer) Run(
+	accounts []struct {
+		signer          signature.Signer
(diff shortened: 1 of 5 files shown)


go/oasis-node/cmd/debug/txsource/workload/workload.go
```diff
@@ -33,6 +33,60 @@ func FundAccountFromTestEntity(ctx context.Context, logger *logging.Logger, cnsc
...
+	// Fund caller to cover transaction fees.
+	feeAmount := int64(tx.Fee.Gas) * gasPrice
+	if err := tx.Fee.Amount.FromInt64(feeAmount); err != nil {
+		return fmt.Errorf("fee amount from int64: %w", err)
+	}
+	if err := transferFunds(ctx, logger, cnsc, fundingAccount, caller.Public(), feeAmount); err != nil {
+		return fmt.Errorf("account funding failure: %w", err)
+	}
+
+	// Submit tx.
```
pro-wh:
```suggestion
	// Sign tx.
```

go/oasis-node/cmd/debug/txsource/workload/transfer.go
```diff
@@ -21,10 +21,9 @@ const (
 	// NameTransfer is the name of the transfer workload.
 	NameTransfer = "transfer"
 
-	transferNumAccounts  = 10
-	transferAmount       = 1
-	transferFundInterval = 10
```
pro-wh: :information_source: the fate of this is that we now fund for every transaction, via a new helper

go/oasis-node/cmd/debug/txsource/workload/transfer.go
```diff
@@ -21,10 +21,9 @@ const (
...
-	transferGasCost      = 10
```
pro-wh: :information_source: the fate of this is that the new helper does a proper gas estimation for each transaction

ptrus:
> hm, is there some other kind of description somewhere we can use better to indicate that the "transfer" workload is 50-50 transfers and burns?

added a comment to the name constant since i couldn't come up with a workload name better indicating it

PR #2774, opened 2020-03-20
buildkite/longtests: build rust code with debug flags

ptrus:
Longtests pipeline doesn't build rust code in debug mode resulting in tests not working in CI. While at it, also parallelize the build similarly as in code pipeline.

Note: this was not noticeable before #2759 as none of the txsource/workloads tested runtime level transactions.


ptrus:
this now works: https://buildkite.com/oasislabs/oasis-core-long-tests/builds/122#_ (note: manually stopeed the job after it sucesfully started)

PR #2777, opened 2020-03-24
go/staking: include LastBlockFees in genesis

ptrus:
Also adds a dump-restore version of the gas-fees test.

### DIFF
--- a/go/oasis-test-runner/scenario/e2e/gas_fees_staking.go
+++ b/go/oasis-test-runner/scenario/e2e/gas_fees_staking.go
@@ -106,8 +106,13 @@ func (sc *gasFeesImpl) Run(childEnv *env.Env) error {
 		return err
 	}
 
+	// Include fees from genesis.
+	totalFees, err := sc.getInitialCommonPoolBalance(ctx)
+	if err != nil {
+		return err
+	}
+
 	// Run some operations that charge fees.
-	var totalFees quantity.Quantity
 	for _, t := range []struct {
 		name string
 		fn   func(context.Context, signature.Signer) (*quantity.Quantity, error)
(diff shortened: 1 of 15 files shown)


Yawning:
Will this break dump/restore?

ptrus:
yeah in the first begin block the `GetLastCommitInfo` will have no votes, so we probably need to also carry over that as well? (or handle differently the initial fee disbursement)

go/oasis-test-runner/scenario/e2e/gas_fees_staking.go
```diff
@@ -106,8 +148,13 @@ func (sc *gasFeesImpl) Run(childEnv *env.Env) error {
 		return err
 	}
 
+	// Include fees from genesis.
+	totalFees, err := sc.getInitialCommonPoolBalance(ctx)
```
pro-wh: does this require that the entire common pool come from fees or something?

.changelog/2777.feature.md
```diff
@@ -0,0 +1,6 @@
+go/staking: include LastBlockFees in genesis
```
pro-wh: oh upgrade this to breaking

go/oasis-test-runner/scenario/e2e/gas_fees_staking.go
ptrus: no doesn't matter, in any case common pool will get distributed to nodes. Updated the comment to not explicitly mention fees and also updated the staking-genesis doc used to have both non-empty initial `last_block_fees` and `common_pool`

.changelog/2777.feature.md
ptrus: hm so not sure if this is the case, under the "breaking" label in github we have a note: "Breaking change requiring state wipe", i don't think that is the case here, also a genesis without last_block_fees does still work. However the change does break forward compatibility as genesis files with last_block_fees field would not work with old versions :thinking:

pro-wh: oh ok

PR #2783, opened 2020-03-25
Various fixes

kostko:
Contains fixes for issues discovered during long-term testing of #2674 which are also present in master.

### DIFF
--- a/go/oasis-node/cmd/debug/txsource/workload/queries.go
+++ b/go/oasis-node/cmd/debug/txsource/workload/queries.go
@@ -6,6 +6,7 @@ import (
 	"math/rand"
 	"time"
 
+	flag "github.com/spf13/pflag"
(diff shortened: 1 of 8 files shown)


go/oasis-node/cmd/debug/txsource/workload/queries.go
```diff
@@ -27,10 +28,17 @@ const (
 	// NameQueries is the name of the queries workload.
 	NameQueries = "queries"
 
+	// CfgNumKeptVersions is the number of last versions that the nodes are keeping (e.g., due to
```
ptrus: Maybe just note somewhere that runtime queries will still be done for all versions

kostko: :+1:

PR #2784, opened 2020-03-25
go/oasis-node/cmd/debug/consim: Initial import

Yawning:


kostko:
Yeah we should run this as part of the daily test pipeline (see `.buildkite/longtests.pipeline.yml`).

Yawning:
> this doesn't exercise the full functionality of staking, which looks at commit info stuff. the current simulator leaves those fields blank

At some point, this becomes an exercise in re-implementing tendermint.

> Yeah we should run this as part of the daily test pipeline (see .buildkite/longtests.pipeline.yml).

Do you honestly thing this will ever do anything more than burn CI resources?  I highly doubt that this codebase will catch anything.

kostko:
I don't see why this would catch anything more than what the existing long-term workload does, it will more probably catch less. But if we have it, we should at least run it somewhere, otherwise why have it at all?

pro-wh:
> > this doesn't exercise the full functionality of staking, which looks at commit info stuff. the current simulator leaves those fields blank
> 
> At some point, this becomes an exercise in re-implementing tendermint.

I'm talking about sending fake results of a commit, not actually running a decentralized fault tolerance protocol.

> > Yeah we should run this as part of the daily test pipeline (see .buildkite/longtests.pipeline.yml).
> 
> Do you honestly thing this will ever do anything more than burn CI resources? I highly doubt that this codebase will catch anything.

a short smoketest on CI, not a long workload like this is intended to do. it'll catch code changes that break this.

pro-wh:
> I don't see why this would catch anything more than what the existing long-term workload does, it will more probably catch less. But if we have it, we should at least run it somewhere, otherwise why have it at all?

ok this PR is gonna need a description

PR #2786, opened 2020-03-26
go/consensus/tendermint: Use our notion of latest height

kostko:
Do not let Tendermint determine the latest height as that completely ignores
ABCI processing so it can return a block for which local state does not yet
exist.


pro-wh:
https://docs.tendermint.com/master/spec/abci/abci.html#query

> `Height (int64)`: The block height for which you want the query (default=0 returns data for the latest committed block). Note that this is the height of the block containing the application's Merkle root hash, which represents the state as it was after committing the block at Height-1

we'd be changing this part of tendermint's query semantics

PR #2794, opened 2020-03-30
go staking: three-way fee split

pro-wh:
We should give more to the previous block proposer, which is the block that first ran the transactions that paid the fees in the `LastBlockFees`. Currently they only get paid as a voter.

# Proposed design

## Summary

The fee for block $h$ is split into $k$ portions, where $k$ is the number of validators at block $h$, one for each of those validators.
Each portion is split by constant weights into a _proposer share_, a _voter share_, and a _next proposer share_.
At the end of block $h$, pay the proposer share of all portions to the proposer of block $h$.
At the beginning of block $h+1$, for each portion where its validator's vote is included in block $h+1$, pay the voter share to the validator and pay the next proposer share to the proposer of block $h+1$.

## Changes to genesis document

In `.staking.params`:

- Changed: `fee_split_*` is renamed to `fee_split_weight_*`
- Added: `fee_split_weight_next_propose`, a `quantity.Quantity`
- Changed: `fee_split_weight_propose` controls the proposer share (previously controlled the next proposer share)

Setting idea:

- `fee_split_weight_propose`: 2
- `fee_split_weight_vote`: 1
- `fee_split_weight_next_propose`: 1

## Changes to block context

- Added: `proposerEntityKey`, the _block proposer_, with the entity of the proposer (needed to remember the proposer from `BeginBlock` for use in `EndBlock`)

## Changes to ABCI state

- Changed: `lastBlockFeesKeyFmt`, the _last block fees_, now stores only the voter shares and next proposer shares

## Algorithms

In cases where resolving an entity ID from a node address fails, skip movements to its balance.

In `BeginBlock`:

- Get the proposer address $A_p$ from the request
- Get the last commit number $k$ of validators from the request
- Get the address $V$ of nodes that voted from the request
- Resolve entity ID $I_p$ from node address $A_p$
- Get $f_{v,q}$ from the last block fees
- Compute a portion size $s=f_{v,q}/k$ (rounds down) where $k$ is the number of validators
- Compute a share size $s_q=s*w_q/(w_v+w_q)$ (rounds down)
- Compute a share size $s_v=s-s_q$
- For each voter address $v_i \in V$:
  - Resolve the entity ID $I_v_i$ from node address $v_i$
  - Move $s_v$ from the last block fees to $I_v_i$'s general balance
- Move $s_q*|V|$ from the last block fees to $I_p$'s general balance
- Move the remaining amount from the last block fees to the common pool
- Store $I_p$ as the block proposer

In `EndBlock`:

- Compute a portion of the fee $f_{v,q}=f*(w_v+w_q)/(w_p+w_v+w_q)$ (rounds down) where $f$ is the accumulated fee, $w_p$ is the proposer share, $w_v$ is the voter share, and $w_q$ is the next proposer share
- Move $f_{v,q}$ from the fee accumulator to the last block fees
- Get $I_p$ from the block proposer
- Move the remaining amount from the fee accumulator to $I_p$'s general balance
- Move the remaining amount from the fee accumulator to the common pool (in case resolving $I_p$ failed)

### DIFF
--- a/go/oasis-test-runner/scenario/e2e/gas_fees_staking.go
+++ b/go/oasis-test-runner/scenario/e2e/gas_fees_staking.go
@@ -190,6 +190,22 @@ func (sc *gasFeesImpl) runTests(ctx context.Context) error {
 		return err
 	}
 
+	// As of the last time this comment was updated:
+	// - total fees has 150 tokens from genesis common pool
+	// - total fees has 50 tokens from genesis last block fees
+	// - for each of 12 transactions that pay for gas:
+	//   - 10 tokens paid for gas in a block on its own
+	//   - (2+2)/(1+2+2) = 80% => 8 tokens persisted for VQ share
+	//   - 10 - 8 = 2 tokens paid to P
+	//   - VQ share divided into 3 validator portions, for 2 tokens each
+	//   - (2)/(2+2) = 50% => 1 token per validator for Q
+	//   - 2 - 1 = 1 token per validator for V
+	//   - remaining 2 tokens moved to common pool
+	// - 150 + 50 + 12 * 10 = 320 tokens `total_fees`
+	// - 12 * 2 = 24 tokens paid for P role
+	// - 12 * 1 * 3 = 36 tokens paid for V roles
+	// - 12 * 1 * 3 = 36 tokens paid for Q role
+	// - 24 + 36 + 36 = 96 tokens `disbursed_fees`
 	sc.logger.Info("making sure that fees have been disbursed",
 		"total_fees", totalFees,
 		"disbursed_fees", newTotalEntityBalance,
--- a/.changelog/2794.breaking.md
+++ b/.changelog/2794.breaking.md
@@ -1,0 +1,16 @@
+go/staking: Three-way fee split
+
+We should give more to the previous block proposer, which is the block
+that first ran the transactions that paid the fees in the
+`LastBlockFees`.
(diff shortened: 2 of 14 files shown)


go/oasis-test-runner/scenario/e2e/gas_fees_staking.go
```diff
@@ -190,6 +190,22 @@ func (sc *gasFeesImpl) runTests(ctx context.Context) error {
 		return err
 	}
 
+	// As of the last time this comment was updated:
```
ptrus: Maybe should add a sanity check that this is indeed the case by adding a manually computed with this comment. 

Would also help with keeping this updated. 

Also note that this test is also run with a dump-restore variant so in the second run the expected result will differ.

pro-wh: what happens in the dump-restore variant?

ptrus: i think the `last_block_fees` & `common_pool` in the dumped genesis will be empty and the rest stays the same?

pro-wh: https://github.com/oasislabs/oasis-core/pull/2798

ptrus: thanks!

PR #2796, opened 2020-03-31
go/txsource/queries: get runtime earliest round from runtime state

ptrus:
Another option would be adding support for querying "earliest" round to the runtime client (which could also include runtime pruning), not sure which option we prefer.

### DIFF
--- a/go/oasis-node/cmd/debug/txsource/workload/queries.go
+++ b/go/oasis-node/cmd/debug/txsource/workload/queries.go
@@ -439,11 +439,11 @@ func (q *queries) Run(gracefulExit context.Context, rng *rand.Rand, conn *grpc.C
 		)
 		return fmt.Errorf("Runtime unmarshal: %w", err)
 	}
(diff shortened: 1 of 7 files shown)


go/oasis-node/cmd/debug/txsource/workload/queries.go
```diff
@@ -416,6 +439,11 @@ func (q *queries) Run(gracefulExit context.Context, rng *rand.Rand, conn *grpc.C
 		)
 		return fmt.Errorf("Runtime unmarshal: %w", err)
 	}
+	rtState, ok := doc.RootHash.RuntimeStates[q.runtimeID]
```
kostko: You could also use the `GetGenesisBlock` query on the roothash backend. Maybe we should expose this via the runtime client as well.

ptrus: :+1: yeah, i can do that

ptrus: Done, added `GetGenesisBlock` client method, please check again

PR #2799, opened 2020-04-01
go/consensus: Introduce gas cost based on tx size

### LINKED ISSUE #2761: Charge gas based on consensus tx size
We should charge gas based on tx size (could be for each byte or in buckets). This will also require a change to `EstimateGas` to take this into account.

kostko:
Fixes #2761

### DIFF
--- a/go/consensus/tendermint/abci/gas.go
+++ b/go/consensus/tendermint/abci/gas.go
@@ -2,6 +2,7 @@ package abci
 
 import (
 	"errors"
+	"fmt"
 	"math"
 
 	"github.com/oasislabs/oasis-core/go/consensus/api/transaction"
@@ -53,7 +54,7 @@ func (ga *basicGasAccountant) UseGas(multiplier int, op transaction.Op, costs tr
 	}
 
 	if ga.usedGas+amount > ga.maxUsedGas {
-		return ErrOutOfGas
+		return fmt.Errorf("%w (limit: %d wanted: %d)", ErrOutOfGas, ga.maxUsedGas, ga.usedGas+amount)
 	}
 
 	ga.usedGas += amount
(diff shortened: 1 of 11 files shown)


go/consensus/tendermint/abci/gas.go
```diff
@@ -53,7 +54,7 @@ func (ga *basicGasAccountant) UseGas(multiplier int, op transaction.Op, costs tr
 	}
 
 	if ga.usedGas+amount > ga.maxUsedGas {
-		return ErrOutOfGas
+		return fmt.Errorf("%w (limit: %d wanted: %d)", ErrOutOfGas, ga.maxUsedGas, ga.usedGas+amount)
```
ptrus, line 57: :+1: had this change locally a couple of times already :grin:

go/oasis-node/cmd/genesis/genesis.go
```diff
@@ -720,6 +724,7 @@ func init() {
 	initGenesisFlags.String(cfgConsensusMaxBlockSizeBytes, "21mb", "tendermint maximum block size (in bytes)")
 	initGenesisFlags.Uint64(cfgConsensusMaxBlockGas, 0, "tendermint max gas used per block")
 	initGenesisFlags.Uint64(cfgConsensusMaxEvidenceAge, 100000, "tendermint max evidence age (in blocks)")
+	initGenesisFlags.Uint64(CfgConsensusGasCostsTxByte, 1, "consensus gas costs: each transaction byte")
```
ptrus, line 727: Maybe:
```suggestion
	initGenesisFlags.Uint64(CfgConsensusGasCostsTxByte, 1, "consensus gas costs of each transaction byte")
```

PR #2800, opened 2020-04-01
go/storage/mkvs: Use cbor.UnmarshalTrusted for internal metadata

kostko:


go/common/cbor/cbor.go
```diff
@@ -27,14 +27,23 @@ var (
...
 		IndefLength: cbor.IndefLengthForbidden,
 		TagsMd:      cbor.TagsForbidden,
 	}
 
-	encMode cbor.EncMode
-	decMode cbor.DecMode
+	// decOptionsTrusted are decoding options for TRUSTED inputs. They are only used when explicitly
+	// requested by using the UnmarshalTrusted method.
+	decOptionsTrusted = cbor.DecOptions{
+		MaxArrayElements: 2097152,
```
ptrus, line 40: Maybe add a comment on why this number? Any reason to not just use the max allowed?

kostko:
> hm, i guess it's already tested in the upstream lib, maybe a test case for the storage code that would fail with old code.

Yeah we don't currently have CBOR tests as those are well tested upstream. We could add some basic API tests for CBOR. But for the database case the problem is that we would need to generate huge data sets (e.g., trees with more than 132k updated nodes) to trigger this which would probably slow tests down. I'll try and see how that would look like.

go/common/cbor/cbor.go
kostko: No special reason, I just multiplied the previous number by 16, could use the max yeah.

kostko:
Ok, it's not that bad, 132k elements takes 2 seconds to process, commit and finalize (locally, probably more on CI?) so we can leave this test in.

go/common/cbor/cbor.go
```diff
@@ -27,14 +27,23 @@ var (
...
 		TagsMd:      cbor.TagsForbidden,
 	}
 
-	encMode cbor.EncMode
-	decMode cbor.DecMode
+	// decOptionsTrusted are decoding options for TRUSTED inputs. They are only used when explicitly
+	// requested by using the UnmarshalTrusted method.
+	decOptionsTrusted = cbor.DecOptions{
+		MaxArrayElements: 134217728, // Maximum allowed.
+		MaxMapPairs:      134217728, // Maximum allowed.
```
pro-wh, line 41: what are the defaults for these on the untrusted `decOptions`?

kostko: [The default is 131072.](https://github.com/fxamacker/cbor#decoding-options) Maybe we should consider making this explicit and/or even changing it (e.g., lowering it) for untrusted settings?

pro-wh: eh, maybe a comment if anything

ridicoulous: @kostko hello. I have an issue with calling [StateToGenesis beacon api](https://github.com/oasisprotocol/oasis-core/blob/9b98fdcad2c64935df98a98a9aba55de9cbe0f06/go/beacon/api/grpc.go#L219): `rpc error: code = Internal desc = grpc: failed to unmarshal the received message cbor: exceeded max number of key-value pairs 131072 for CBOR map` at block `7188323`. How I can explicitly call UnmarshalTrusted method at this point?

PR #2802, opened 2020-04-01
 go consensus: shorten gas import

pro-wh:
switch to more concise `FromUint64`


pro-wh:
coverage is wrong. this line is hit

PR #2814, opened 2020-04-02
go/txsource/queries: Wait for indexed blocks before GetBlockByHash

kostko:


go/oasis-node/cmd/debug/txsource/workload/queries.go
```diff
@@ -363,6 +363,18 @@ func (q *queries) doRuntimeQueries(ctx context.Context, rng *rand.Rand) error {
 		)
 		return fmt.Errorf("runtimeClient.GetBlock, round: %d: %w", round, err)
 	}
+	// GetBlockByHash requires that the block was actually indexed, so wait for it.
+	err = q.runtime.WaitBlockIndexed(ctx, &runtimeClient.WaitBlockIndexedRequest{
```
ptrus, line 367: Is there any default timeout for this or should we add one? 
Actually should we just add a (reasonably long) timeout for each loop iteration in the "main" workload loop, to ensure the workload doesn't get stuck and seem like it works?

kostko: Yeah we should add a general per-loop timeout. I'll add it in this PR.

kostko: Added.

PR #2824, opened 2020-04-06
go/common/crypto/signature/signer/composite: Initial import

Yawning:
This adds a composite signer factory that can aggregate multiple signer
factories.  This could be used (for example), to use multiple signer
backends simultaneously, depending on the key role.

Eg: The P2P link signer could use a local file, while the consensus
signer can be backed by a remote HSM.

Fixes 2684.

### DIFF
--- a/go/oasis-node/cmd/common/signer/signer.go
+++ b/go/oasis-node/cmd/common/signer/signer.go
@@ -4,13 +4,16 @@ package signer
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	flag "github.com/spf13/pflag"
 	"github.com/spf13/viper"
 
 	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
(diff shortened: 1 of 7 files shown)


go/oasis-node/cmd/common/signer/signer.go
```diff
@@ -33,6 +35,8 @@ const (
 	cfgSignerRemoteClientCert = "signer.remote.client.certificate"
 	cfgSignerRemoteClientKey  = "signer.remote.client.key"
 	cfgSignerRemoteServerCert = "signer.remote.server.certificate"
+
+	cfgSignerCompositeBackends = "signer.composite.backends"
```
kostko, line 39: Can we have a test for this argument in E2E tests?

Yawning: That's mostly useless at the moment.  I added unit tests for the ctor instead, please take a look.

kostko: :heavy_check_mark:

PR #2828, opened 2020-04-07
Include oasis-core-runtime-loader and oasis-net-runner in releases

### LINKED ISSUE #2780: Additional artifacts in oasis-core releases
To facilitate runtime development, we should consider adding artifacts needed for testing runtimes to `oasis-core` releases:
- oasis-net-runner
- oasis-core-runtime-loader

kostko:
Fixes #2780 

The current solution for including Rust artifacts does not seem the best to me, but Goreleaser is slightly awkward. I tested this by running `make release`, probably I also need to change the release workflow to include the Rust compiler.

TODO
* [x] Update release workflow to include the Rust compiler.
* [x] Test updated release workflow.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,10 +22,15 @@ jobs:
         uses: actions/setup-go@v1
         with:
           go-version: "1.13.x"
+      - name: Set up Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly-2020-02-16
+          override: true
       - name: Install oasis-node prerequisites
         run: |
           sudo apt-get update
-          sudo apt-get install make libseccomp-dev
+          sudo apt-get install make libseccomp-dev protobuf-compiler
       - name: Install GoReleaser
         run: |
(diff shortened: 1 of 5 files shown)


kostko:
Tested the release flow [on gh-actions-temp repository to generate a fake 20.5 release](https://github.com/oasislabs/gh-actions-temp/releases/tag/v20.5). Seems to work.

.github/workflows/release.yml
```diff
@@ -22,10 +22,15 @@ jobs:
         uses: actions/setup-go@v1
         with:
           go-version: "1.13.x"
+      - name: Set up Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly-2020-02-16
```
peterjgilbert, line 28: Assume this needs to be kept in sync with other config?

kostko: Yeah but the same is true for the used Go version.

peterjgilbert: got it

PR #2837, opened 2020-04-13
keyamanger-runtime: replace with test/simple-keymanager

ptrus:
Common keymanager initalization code is extracted into the `keymanager-lib`
crate. This enables for the actual key manager implementation to only
provide a set of key manager policy signers.

Aditionally the `keymanager-runtime` crate is removed and replaced with
a test `simple-keymanager` runtime that is used in E2E tests. 

(not sure if "simple-keymanager" is the correct name)

TODO: 
- [x] PR in oais-runtime that adds a keymanager-runtime there (https://github.com/oasislabs/oasis-runtime/pull/985)


.changelog/2837.internal.md
```diff
@@ -0,0 +1,7 @@
+keyamanger-runtime: replace with test/simple-keymanager
```
kostko, line 1:
```suggestion
keymanger-runtime: replace with test/simple-keymanager
```

keymanager-api-common/src/lib.rs
```diff
@@ -69,3 +72,28 @@ impl SignedPolicySGX {
         Ok(self.policy.clone())
     }
 }
+
+/// Returns the defult set of the trusted policy signers for key manager.
+pub fn default_trusted_policy_signers() -> TrustedPolicySigners {
```
kostko, line 77: Why do we have this method, it seems dangerous? Shouldn't this be in the specific key manager only (e.g., in `simple-keymanager`)?

keymanager-client/src/client.rs
```diff
@@ -94,7 +94,7 @@ impl RemoteClient {
         keys_cache_sizes: usize,
     ) -> Self {
         #[cfg(target_env = "sgx")]
-        init_trusted_policy_signers();
+        set_trusted_policy_signers(default_trusted_policy_signers()); // TODO: configurable?
```
kostko, line 97: Yes this must be configurable otherwise you can't instantiate proper clients at all.

Why not pass the policy as an argument?

tests/runtimes/simple-keymanager/src/main.rs
```diff
@@ -0,0 +1,9 @@
+use oasis_core_keymanager_api_common::*;
+use oasis_core_keymanager_lib::keymanager::*;
+use oasis_core_runtime::{common::version::Version, version_from_cargo};
+
+fn main() {
+    let init = new_keymanager(default_trusted_policy_signers());
+    // Start the runtime.
```
kostko, line 7: Probably no need for this comment.

keymanager-api-common/src/lib.rs
ptrus: yeah i just moved/renamed from ` keymanager-runtime/api/src/lib.rs` will put it into the `simple-keymanager-api` (and oasis-runtime keymanager)

kostko: Sounds good. Also for the oasis-runtime one we should have an actual policy someday ;-)

keymanager-client/src/client.rs
ptrus: >Why not pass the policy as an argument?

That was my question as well, hence the comment. Will change it

keymanager-api-common/src/lib.rs
ptrus: done

keymanager-client/src/client.rs
ptrus: done

PR #2839, opened 2020-04-13
go: Bump tendermint to v0.33.3-oasis

### LINKED ISSUE #2834: Bump Tendermint Core to 0.33.x
After we get rid of the IAVL dependency (see #2674) we should bump Tendermint Core to 0.33.x. This is a breaking upgrade as the block format has changed.

Yawning:
* [x] Bump the tendermint dependency to the fork based on v0.33.3
 * [x] Fix oasis-core 
 * [x] Fix the random tool in extra that also broke (untested)
 * [x] Max evidence age is now both block and time based.  Support configuring the latter
 * [x] Use the actual tendermint fork tag once @kostko makes one.
 * [x] Add one of those changelog fragments.

Fixes #2834


kostko:
> Use the actual tendermint fork tag once @kostko makes one.

The tag `v0.33.3-oasis1` is now there.

go/extra/stats/cmd/stats.go
```diff
@@ -293,13 +293,7 @@ func getStats(ctx context.Context, consensus consensusAPI.ClientBackend, registr
 		}
 
 		// Go over all signatures for a block.
-		// XXX: In tendermint master (not yet released) the signatures are
-		// obtained in LastCommit.Signatures.
-		for _, sig := range tmBlockMeta.LastCommit.Precommits {
-			if sig == nil {
```
kostko, line 299: Should this skip if `sig.Absent()` is true?

Yawning: Dunno, not my tool.

Yawning: Done.

go/extra/stats/cmd/stats.go
```diff
@@ -293,11 +293,9 @@ func getStats(ctx context.Context, consensus consensusAPI.ClientBackend, registr
...
-				logger.Debug("skipping nil-votes")
+		for _, sig := range tmBlockMeta.LastCommit.Signatures {
+			if sig.Absent() {
```
kostko, line 297: @ptrus can you confirm that this is ok?

ptrus: yeah seems like `nil` votes match the new `Absent()` votes, so seems fine :+1:

PR #2841, opened 2020-04-14
Add `GetEvents` to backends

### LINKED ISSUE #2778: Support querying for all events at specific height
Currently it is only possible to subscribe to events via the various `Watch*` methods exposed by the backends. We should add a new method to each backend's API to retrieve all events at a specific height.

For the Tendermint backend this can be implemented by querying ABCI responses at specific height (using `GetBlockResults`) and converting that into events (note that some events require additional ABCI state queries).

An alternative (or a later addition as a cache) would be to separately index events in a database (e.g., a Badger database which stores per-height events as `<height> <backend-id>`) and then just return them when requested. This would also need to take pruning into account to prune old data based on the existing configured consensus state pruning policy.

abukosek:
Fixes #2778.

### DIFF
--- a/go/consensus/tendermint/registry/registry.go
+++ b/go/consensus/tendermint/registry/registry.go
@@ -169,94 +169,7 @@ func (tb *tendermintBackend) GetEvents(ctx context.Context, height int64) (*[]ap
 	for _, txResults := range results.TxsResults {
 		tmEvents = append(tmEvents, txResults.Events...)
 	}
-	events := []api.Event{}
-	for _, tmEv := range tmEvents {
-		// Ignore events that don't relate to the registry app.
-		if tmEv.GetType() != app.EventType {
-			continue
-		}
-
-		for _, pair := range tmEv.GetAttributes() {
-			key := pair.GetKey()
-			val := pair.GetValue()
-			if bytes.Equal(key, app.KeyRuntimeRegistered) {
-				// Runtime registered event.
-				var rt api.Runtime
-				if err := cbor.Unmarshal(val, &rt); err != nil {
-					return nil, errors.Wrap(err, "registry: corrupt RuntimeRegistered event")
-				}
-				evt := api.Event{
-					RuntimeEvent: &api.RuntimeEvent{Runtime: &rt},
-				}
-				events = append(events, evt)
-			} else if bytes.Equal(key, app.KeyEntityRegistered) {
-				// Entity registered event.
-				var ent entity.Entity
-				if err := cbor.Unmarshal(val, &ent); err != nil {
-					return nil, errors.Wrap(err, "registry: corrupt EntityRegistered event")
-				}
-				evt := api.Event{
-					EntityEvent: &api.EntityEvent{
-						Entity:         &ent,
-						IsRegistration: true,
-					},
-				}
-				events = append(events, evt)
-			} else if bytes.Equal(key, app.KeyEntityDeregistered) {
-				// Entity deregistered event.
-				var dereg app.EntityDeregistration
-				if err := cbor.Unmarshal(val, &dereg); err != nil {
-					return nil, errors.Wrap(err, "registry: corrupt EntityDeregistered event")
-				}
-				evt := api.Event{
-					EntityEvent: &api.EntityEvent{
-						Entity:         &dereg.Entity,
-						IsRegistration: false,
-					},
-				}
-				events = append(events, evt)
-			} else if bytes.Equal(key, app.KeyNodeRegistered) {
-				// Node registered event.
-				var n node.Node
-				if err := cbor.Unmarshal(val, &n); err != nil {
-					return nil, errors.Wrap(err, "registry: corrupt NodeRegistered event")
-				}
-				evt := api.Event{
-					NodeEvent: &api.NodeEvent{
-						Node:           &n,
-						IsRegistration: true,
-					},
-				}
-				events = append(events, evt)
-			} else if bytes.Equal(key, app.KeyNodesExpired) {
-				// Nodes expired event.
-				var nodes []*node.Node
-				if err := cbor.Unmarshal(val, &nodes); err != nil {
-					return nil, errors.Wrap(err, "registry: corrupt NodesExpired event")
-				}
-				evt := api.Event{
-					NodesExpiredEvent: &api.NodesExpiredEvent{
-						Nodes: nodes,
-					},
-				}
-				events = append(events, evt)
-			} else if bytes.Equal(key, app.KeyNodeUnfrozen) {
-				// Node unfrozen event.
-				var nid signature.PublicKey
-				if err := cbor.Unmarshal(val, &nid); err != nil {
-					return nil, errors.Wrap(err, "registry: corrupt NodeUnfrozen event")
-				}
-				evt := api.Event{
-					NodeUnfrozenEvent: &api.NodeUnfrozenEvent{
-						NodeID: nid,
-					},
-				}
-				events = append(events, evt)
-			}
-		}
-	}
-
-	return &events, nil
+	return tb.onABCIEvents(ctx, tmEvents, height, false)
 }
 
 func (tb *tendermintBackend) worker(ctx context.Context) {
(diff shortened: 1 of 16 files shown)


go/consensus/tendermint/registry/registry.go
```diff
@@ -150,6 +152,113 @@ func (tb *tendermintBackend) StateToGenesis(ctx context.Context, height int64) (
...
+	for _, tmEv := range tmEvents {
+		// Ignore events that don't relate to the registry app.
+		if tmEv.GetType() != app.EventType {
+			continue
+		}
+
+		for _, pair := range tmEv.GetAttributes() {
+			key := pair.GetKey()
+			val := pair.GetValue()
+			if bytes.Equal(key, app.KeyRuntimeRegistered) {
```
kostko, line 182: Could we unify this decoding with what is done in `onABCIEvents`? Also elsewhere.

go/consensus/tendermint/registry/registry.go
```diff
@@ -189,72 +211,123 @@ func (tb *tendermintBackend) onEventDataNewBlock(ctx context.Context, ev tmtypes
...
 
 func (tb *tendermintBackend) onEventDataTx(ctx context.Context, tx tmtypes.EventDataTx) {
-	tb.onABCIEvents(ctx, tx.Result.Events, tx.Height)
+	_, _ = tb.onABCIEvents(ctx, tx.Result.Events, tx.Height, true)
 }
 
-func (tb *tendermintBackend) onABCIEvents(ctx context.Context, events []abcitypes.Event, height int64) {
-	for _, tmEv := range events {
+func (tb *tendermintBackend) onABCIEvents(ctx context.Context, tmEvents []abcitypes.Event, height int64, doBroadcast bool) (*[]api.Event, error) { // nolint: gocyclo
+	events := []api.Event{}
```
kostko, line 222:
```suggestion
	var events []api.Event
```

go/consensus/tendermint/registry/registry.go
```diff
@@ -189,72 +211,123 @@ func (tb *tendermintBackend) onEventDataNewBlock(ctx context.Context, ev tmtypes
...
 }
 
 func (tb *tendermintBackend) onEventDataTx(ctx context.Context, tx tmtypes.EventDataTx) {
-	tb.onABCIEvents(ctx, tx.Result.Events, tx.Height)
+	_, _ = tb.onABCIEvents(ctx, tx.Result.Events, tx.Height, true)
 }
 
-func (tb *tendermintBackend) onABCIEvents(ctx context.Context, events []abcitypes.Event, height int64) {
-	for _, tmEv := range events {
+func (tb *tendermintBackend) onABCIEvents(ctx context.Context, tmEvents []abcitypes.Event, height int64, doBroadcast bool) (*[]api.Event, error) { // nolint: gocyclo
```
kostko, line 221: A slice can already be nil so no need to return a pointer to a slice. Also in other places, including the backend interfaces.
```suggestion
func (tb *tendermintBackend) onABCIEvents(ctx context.Context, tmEvents []abcitypes.Event, height int64, doBroadcast bool) ([]api.Event, error) { // nolint: gocyclo
```

go/consensus/tendermint/registry/registry.go
```diff
@@ -264,22 +337,46 @@ func (tb *tendermintBackend) onABCIEvents(ctx context.Context, events []abcitype
...
+				evt := api.Event{
+					NodeUnfrozenEvent: &api.NodeUnfrozenEvent{
+						NodeID: nid,
+					},
+				}
+				events = append(events, evt)
 			}
 		}
 	}
+	return &events, nil
```
kostko, line 379:
```suggestion
	return events, nil
```

go/consensus/tendermint/registry/registry.go
```diff
@@ -189,72 +211,123 @@ func (tb *tendermintBackend) onEventDataNewBlock(ctx context.Context, ev tmtypes
...
+			if bytes.Equal(key, app.KeyNodesExpired) {
+				// Nodes expired event.
 				var nodes []*node.Node
-				if err := cbor.Unmarshal(pair.GetValue(), &nodes); err != nil {
+				if err := cbor.Unmarshal(val, &nodes); err != nil {
 					tb.logger.Error("worker: failed to get nodes from tag",
 						"err", err,
 					)
+					if !doBroadcast {
+						return nil, errors.Wrap(err, "registry: corrupt NodesExpired event")
```
kostko, line 240: Use `fmt.Errorf` for all new code (also elsewhere).

go/oasis-node/cmd/debug/txsource/workload/queries.go
```diff
@@ -232,6 +232,12 @@ func (q *queries) doRegistryQueries(ctx context.Context, rng *rand.Rand, height
 		}
```
kostko, line 232: Also add a `GetEvents` call for roothash.

go/registry/api/api.go
```diff
@@ -281,6 +284,30 @@ type NodeEvent struct {
 	IsRegistration bool
 }
 
+// RuntimeEvent signifies new runtime registration.
+type RuntimeEvent struct {
+	Runtime *Runtime
```
kostko, line 289: Add explicit serialization field annotations to all event structures (also pre-existing ones are missing these). For example here it would be something like:
```suggestion
	Runtime *Runtime `json:"runtime"`
```

go/staking/api/api.go
```diff
@@ -151,6 +154,13 @@ type EscrowEvent struct {
 	Reclaim *ReclaimEscrowEvent `json:"reclaim,omitempty"`
 }
 
+// Event signifies a staking event, returned via GetEvents.
+type Event struct {
+	TransferEvent *TransferEvent
```
kostko, line 159: Add explicit serialization field annotations.

abukosek:
Thanks for the review, I've addressed all the comments, so please re-review :)

go/consensus/tendermint/registry/registry.go
```diff
@@ -189,72 +211,135 @@ func (tb *tendermintBackend) onEventDataNewBlock(ctx context.Context, ev tmtypes
...
+					}
+				} else {
+					evt := api.Event{
+						NodesExpiredEvent: &api.NodesExpiredEvent{
+							Nodes: nodes,
+						},
+					}
+					events = append(events, evt)
+
+					// For compatibility, we also emit NodeEvents with
```
kostko, line 261: Why have both?

go/consensus/tendermint/tendermint.go
```diff
@@ -835,19 +835,36 @@ func (t *tendermintService) GetHeight(ctx context.Context) (int64, error) {
...
+
+	result, err := t.client.BlockResults(&tmHeight)
 	if err != nil {
 		return nil, fmt.Errorf("tendermint: block results query failed: %w", err)
 	}
 
 	return result, nil
 }
 
+func (t *tendermintService) GetEvents(ctx context.Context, height int64) ([]roothashAPI.Event, error) {
```
kostko, line 864: Why is this here? It gets you roothash events?

Oh I see, you added it here because the roothash service is not exposed directly. This doesn't seem right. If anywhere this should be added to the `RuntimeClient` API instead. But seeing what this contains I would be ok with just not adding it (and no need to test it in txsource for now as long as there is a unit test).

go/registry/api/api.go
```diff
@@ -270,20 +273,44 @@ func NewRegisterRuntimeTx(nonce uint64, fee *transaction.Fee, sigRt *SignedRunti
...
+}
+
+// NodeUnfrozenEvent signifies when node becomes unfrozen.
+type NodeUnfrozenEvent struct {
+	NodeID signature.PublicKey `json:"node_id"`
+}
+
+// Event is a registry event returned via GetEvents.
+type Event struct {
+	RuntimeEvent      *RuntimeEvent      `json:"runtime_event,omitempty"`
```
kostko, line 304: The serialized fields could omit the `_event` suffix (as the whole struct contains events).

go/staking/api/api.go
```diff
@@ -151,6 +154,13 @@ type EscrowEvent struct {
 	Reclaim *ReclaimEscrowEvent `json:"reclaim,omitempty"`
 }
 
+// Event signifies a staking event, returned via GetEvents.
+type Event struct {
+	TransferEvent *TransferEvent `json:"transfer_event,omitempty"`
```
kostko, line 159: Same as above, omit `_event` suffix.

go/consensus/tendermint/registry/registry.go
abukosek: Because the rest of the code seems to expect these events and not the expiration events, especially in tests.  I can remove this if it doesn't make sense.

go/consensus/tendermint/tendermint.go
abukosek: Yes, there is no RoothashClient.  I've removed this code now.

go/consensus/tendermint/registry/registry.go
kostko: Yeah but why have the separate expiration events at all?

abukosek: Oh, you meant that.  Don't know, I just exported all the events that the underlying layer had, maybe it will be useful to the developers that have requested this feature?

kostko: But the two are basically equivalent (there is no other way to deregister a node, so node deregistration event is the expiration event) if you ignore aggregation (list vs. multiple events). I would rather have just one.

abukosek: OK, I'll remove the expiration event then :)

abukosek: Done!

kostko:
Aw you forgot to rebase :/

PR #2843, opened 2020-04-15
go/oasis-test-runner/oasis: Add a keymanager replication test

Yawning:
* [x] Add a basic replication test.
    * [x] Spin up a replica.
    * [x] Explicitly query the replica or something.
 * [x] Fix bugs.
    * [x] Oh my god, the key manager worker host doesn't support enclave rpc.
    * [x] Key manager client should support km->km.
    * [x] Fix the key manager grpc access control.

### DIFF
--- a/go/oasis-net-runner/fixtures/fixtures_test.go
+++ b/go/oasis-net-runner/fixtures/fixtures_test.go
@@ -20,6 +20,15 @@ func TestDefaultFixture(t *testing.T) {
 	require.Nil(t, err)
 	require.NotNil(t, data)
 
+	// As cool as having tests cases is, having to regenerate test data
+	// every single time the default fixture changes is incredibly
+	// annoying.
+	//
+	// May this pearl of wisdom serve as a guiding light for the next
+	// unfortunate victim.
+	//
+	// $ ./oasis-net-runner dump-fixture > /tmp/fuckfuckfuckfuckfuck
+
 	storedData, err := ioutil.ReadFile(defaultFixturePath)
 	require.Nil(t, err)
 	require.NotNil(t, storedData)
--- a/go/oasis-test-runner/oasis/keymanager.go
+++ b/go/oasis-test-runner/oasis/keymanager.go
@@ -131,14 +131,15 @@ func (net *Network) NewKeymanagerPolicy(cfg *KeymanagerPolicyCfg) (*KeymanagerPo
 		return nil, errors.Wrap(err, "oasis/keymanager: failed to create keymanager policy subdir")
 	}
 
-	net.keymanagerPolicy = &KeymanagerPolicy{
+	newPol := &KeymanagerPolicy{
 		net:     net,
 		dir:     policyDir,
 		runtime: cfg.Runtime,
 		serial:  cfg.Serial,
 	}
+	net.keymanagerPolicies = append(net.keymanagerPolicies, newPol)
 
-	return net.keymanagerPolicy, nil
+	return newPol, nil
 }
 
 // Keymanager is an Oasis key manager.
@@ -149,6 +150,7 @@ type Keymanager struct { // nolint: maligned
 
 	runtime *Runtime
 	entity  *Entity
+	policy  *KeymanagerPolicy
 
 	tmAddress        string
 	consensusPort    uint16
@@ -165,6 +167,7 @@ type KeymanagerCfg struct {
 
 	Runtime *Runtime
 	Entity  *Entity
+	Policy  *KeymanagerPolicy
 }
 
 // IdentityKeyPath returns the paths to the node's identity key.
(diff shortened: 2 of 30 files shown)


go/worker/keymanager/handler.go
```diff
@@ -3,22 +3,71 @@ package keymanager
...
+		h.w.logger.Error("failed to create remote client",
+			"err", err,
+		)
+		return
+	}
+
+	select {
+	case <-h.w.ctx.Done():
+		h.w.logger.Error("failed to wait for key manager",
+			"err", err,
```
kostko, line 51: Probably should be `h.w.ctx.Err()` instead of `err`.

go/worker/keymanager/handler.go
```diff
@@ -3,22 +3,71 @@ package keymanager
...
+	sync.Mutex
+
 	w            *Worker
+	remoteClient *client.Client
 	localStorage localstorage.LocalStorage
 }
 
+func (h *hostHandler) initRemoteClient(commonWorker *workerCommon.Worker) {
+	// Wait for the runtime (XXX: Do we need to do this?)
+	if _, err := h.w.runtime.RegistryDescriptor(h.w.ctx); err != nil {
```
kostko, line 33: I don't think this is needed, the key manager client should watch for registry descriptor changes internally.

Yawning:
Well, replication works.  The rust side simple encryption test client will panic when the query goes to the replica, but that's an orthogonal problem.  Either this doesn't retry, or the retry logic is busted.

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: RpcFailure(RpcStatus { status: Unknown, details: Some("worker/keymanager: error from runtime: decrypt error") })', tests/clients/simple-keyvalue-enc/src/main.rs:99:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

kostko:
> The rust side simple encryption test client will panic when the query goes to the replica, but that's an orthogonal problem. Either this doesn't retry, or the retry logic is busted

Rust part of the retry mechanism should work and is also covered in unit tests. Currently the Rust client will only retry three times though so depending on the situation this may not be enough. It could also be a problem with the round robin node selection policy.

go/oasis-test-runner/oasis/keymanager.go
```diff
@@ -247,9 +299,18 @@ func (net *Network) NewKeymanager(cfg *KeymanagerCfg) (*Keymanager, error) {
 		return nil, errors.Wrap(err, "oasis/keymanager: failed to create keymanager subdir")
 	}
 
+	// HACK HACK HACK: Not sure how to fit this into the fixture stuff.
```
kostko, line 302: What's the problem? The only major difference to a `KeyManagerPolicyCfg` is that the `Runtime` would be referenced by index (so it can be serialized) instead of being a pointer to a runtime object.

I would change this slightly so that the network would have a list of key manager policies, similar to how it has a list of runtimes. Then at network setup time, you would iterate over them and set them up. The fixture would then just initialize that list, similar to how it does with runtimes.

go/oasis-test-runner/scenario/e2e/keymanager_replicate.go
```diff
@@ -0,0 +1,149 @@
...
+
+// KeymanagerReplicate is the keymanager replication scenario.
+var KeymanagerReplicate scenario.Scenario = newKmReplicateImpl()
+
+type kmReplicateImpl struct {
+	runtimeImpl
+}
+
+func newKmReplicateImpl() scenario.Scenario {
+	// BUG: This should use `simple-keyvalue-enc-client`, but the client
```
kostko, line 24: Can you file an issue for this?

Yawning: #2885

go/oasis-net-runner/fixtures/fixtures_test.go
```diff
@@ -20,6 +20,15 @@ func TestDefaultFixture(t *testing.T) {
 	require.Nil(t, err)
 	require.NotNil(t, data)
 
+	// As cool as having tests cases is, having to regenerate test data
+	// every single time the default fixture changes is incredibly
+	// annoying.
+	//
+	// May this pearl of wisdom serve as a guiding light for the next
```
kostko, line 27: :rofl:

go/oasis-test-runner/oasis/keymanager.go
```diff
@@ -18,6 +20,125 @@ const (
...
+func (pol *KeymanagerPolicy) provision() error {
+	if pol.runtime.teeHardware == node.TEEHardwareInvalid {
+		// No policy document.
+		pol.statusArgs = append(pol.statusArgs, "--keymanager.policy.file", "")
+	} else {
+		// Policy signed with test keys.
+		policyPath := filepath.Join(pol.dir.String(), kmPolicyFile)
+		policyArgs := []string{
+			"keymanager", "init_policy",
+			"--debug.dont_blame_oasis",
```
matevz, line 53: Please use "--"+`CfgDebugDontBlameOasis` instead of duplicating the literal.

go/oasis-test-runner/oasis/keymanager.go
```diff
@@ -18,6 +20,125 @@ const (
...
+			"--keymanager.policy.file", policyPath,
```
matevz, line 54: "--"+`CfgPolicyFile` ?

go/oasis-test-runner/oasis/keymanager.go
Yawning: Was like that from when you wrote it, but sure.

go/oasis-test-runner/oasis/keymanager.go
Yawning: Likewise, was like that originally, but sure.

PR #2849, opened 2020-04-16
go/oasis-test-runner/txsource: Generate a new random seed on each run

kostko:
I remembered we don't actually change the seed :grin:

### DIFF
--- a/go/oasis-test-runner/scenario/e2e/txsource.go
+++ b/go/oasis-test-runner/scenario/e2e/txsource.go
@@ -3,6 +3,8 @@ package e2e
 import (
 	"context"
 	"crypto"
+	cryptoRand "crypto/rand"
+	"encoding/hex"
 	"fmt"
 	"math"
 	"math/rand"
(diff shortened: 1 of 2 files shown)


go/oasis-test-runner/scenario/e2e/txsource.go
```diff
@@ -270,6 +284,7 @@ func (sc *txSourceImpl) startWorkload(childEnv *env.Env, errCh chan error, name
 		"--" + workload.CfgRuntimeID, runtimeID.String(),
 		"--" + txsource.CfgWorkload, name,
 		"--" + txsource.CfgTimeLimit, sc.timeLimit.String(),
+		"--" + txsource.CfgSeed, sc.seed,
```
pro-wh, line 287: how did the txsource process used to get the seed?

ptrus: used the default https://github.com/oasislabs/oasis-core/blob/master/go/oasis-node/cmd/debug/txsource/txsource.go#L136

PR #2852, opened 2020-04-17
go/oasis-test-runner: Test dynamic KM runtime registration

### LINKED ISSUE #2840: Test dynamic key manager runtime registration
After #2516 we should update the `runtime-dynamic` E2E test (there are TODOs in the code) to actually test dynamic key manager runtime registration.

abukosek:
Fixes #2840.

TODO:
- [X] Add helper for keymanager CLI stuff to oasis-test-runner.
- [X] Fix keymanager tendermint app to actually register the UpdatePolicy method.
- [X] Handle non-existing previous KM status as new in UpdatePolicy transaction.
- [X] Convert `signature.PublicKey` to `common.Namespace` in EnclavePolicySGX and related.
- [X] Fix `keymanager gen_update` command to accept the consensus txn flags and the policy & policy signatures flags.
- [X] Fix `keymanager gen_update` command to properly deserialize policy file (should use CBOR, but previous code had JSON).
- [X] Add `GetEnclaveIdentity` and `Kind` helpers to oasis-test-runner's Runtime struct.
- [X] Dynamically register KM runtime in `runtime-dynamic` e2e test.
- [X] Figure out why it doesn't work (forgot to bump the nonces in the rest of the `runtime-dynamic` test, oops).
- [X] Figure out why it hangs in SGX tests waiting on nodes (tendermint keymanager app didn't handle tags that weren't at the beginning of blocks).
- [X] Make the keymanager tendermint app's `updatePolicy` actually update the policy :smile:
- [x] Figure out why it still hangs in SGX (not enough signatures).

### DIFF
--- a/go/oasis-test-runner/scenario/e2e/runtime_dynamic.go
+++ b/go/oasis-test-runner/scenario/e2e/runtime_dynamic.go
@@ -6,11 +6,14 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/oasislabs/oasis-core/go/common"
 	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
 	"github.com/oasislabs/oasis-core/go/common/quantity"
+	"github.com/oasislabs/oasis-core/go/common/sgx"
 	consensus "github.com/oasislabs/oasis-core/go/consensus/api"
 	"github.com/oasislabs/oasis-core/go/consensus/api/transaction"
 	epochtime "github.com/oasislabs/oasis-core/go/epochtime/api"
+	keymanager "github.com/oasislabs/oasis-core/go/keymanager/api"
 	"github.com/oasislabs/oasis-core/go/oasis-test-runner/env"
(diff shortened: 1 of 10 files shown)


go/oasis-test-runner/scenario/e2e/runtime_dynamic.go
```diff
@@ -103,14 +100,101 @@ func (sc *runtimeDynamicImpl) Run(childEnv *env.Env) error { // nolint: gocyclo
...
+		// do an epoch transition instead (to complete the KM runtime
+		// registration).
+		if err := sc.epochTransition(ctx); err != nil {
+			return err
+		}
+	}
+
+	// Wait for enough nodes to register, then make another epoch transition.
+	if havePolicy {
+		numNodes += len(sc.net.Keymanagers())
```
kostko, line 179: Hm, I think both cases (SGX and non-SGX) should wait for the key manager to register before performing an epoch transition.

abukosek: Thanks, fixed! :)

PR #2868, opened 2020-04-23
go consensus: stake weighted voting

pro-wh:
Now here's something we want to experiment with: validator voting power proportional to entity stake.

### DIFF
--- a/go/consensus/tendermint/apps/scheduler/genesis.go
+++ b/go/consensus/tendermint/apps/scheduler/genesis.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
 	"github.com/oasislabs/oasis-core/go/common/node"
-	consensus "github.com/oasislabs/oasis-core/go/consensus/api"
 	abciAPI "github.com/oasislabs/oasis-core/go/consensus/tendermint/api"
 	registryState "github.com/oasislabs/oasis-core/go/consensus/tendermint/apps/registry/state"
 	schedulerState "github.com/oasislabs/oasis-core/go/consensus/tendermint/apps/scheduler/state"
@@ -32,7 +31,7 @@ func (app *schedulerApplication) InitChain(ctx *abciAPI.Context, req types.Reque
 	if doc.Scheduler.Parameters.DebugStaticValidators {
 		ctx.Logger().Warn("static validators are configured")
 
-		var staticValidators []signature.PublicKey
+		staticValidators := make(map[signature.PublicKey]int64)
 		for _, v := range req.Validators {
 			tmPk := v.GetPubKey()
 
@@ -53,7 +52,8 @@ func (app *schedulerApplication) InitChain(ctx *abciAPI.Context, req types.Reque
 				return fmt.Errorf("scheduler: invalid static validator public key: %w", err)
 			}
 
-			staticValidators = append(staticValidators, id)
+			// Use a flat vote weight in this simplified configuration.
+			staticValidators[id] = 1
 		}
 
 		// Add the current validator set to ABCI, so that we can query it later.
@@ -95,7 +95,7 @@ func (app *schedulerApplication) InitChain(ctx *abciAPI.Context, req types.Reque
 
 	// Assemble the list of the tendermint genesis validators, and do some
 	// sanity checking.
-	var currentValidators []signature.PublicKey
+	currentValidators := make(map[signature.PublicKey]int64)
 	for _, v := range req.Validators {
 		tmPk := v.GetPubKey()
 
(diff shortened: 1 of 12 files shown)


pro-wh:
oh we're missing part of bypass-stake support for e2e tests.

when it tries to initialize the validator set from genesis

pro-wh:
oh, correction: they don't use "debug bypass stake." the e2e test genesis doc is just not high enough fidelity to run now

go/consensus/tendermint/apps/scheduler/genesis.go
```diff
@@ -143,10 +135,17 @@ func (app *schedulerApplication) InitChain(ctx *abciAPI.Context, req types.Reque
 	//
 	// Sort of stupid it needs to be done this way, but tendermint doesn't
 	// appear to pass ABCI the validator set anywhere other than InitChain.
-	if err := state.PutCurrentValidators(ctx, currentValidators); err != nil {
+	if err = state.PutCurrentValidators(ctx, currentValidators); err != nil {
 		return fmt.Errorf("failed to set validator set: %w", err)
 	}
 
+	if !doc.Scheduler.Parameters.DebugBypassStake {
```
kostko, line 142: This check should (also) be performed in the genesis document sanity check so it fails early.

go/consensus/tendermint/apps/scheduler/genesis.go
```diff
@@ -134,7 +126,7 @@ func (app *schedulerApplication) InitChain(ctx *abciAPI.Context, req types.Reque
 		ctx.Logger().Debug("adding validator to current validator set",
 			"id", id,
 		)
-		currentValidators = append(currentValidators, n.Consensus.ID)
+		currentValidators[n.Consensus.ID] = v.Power
```
kostko, line 129: Shouldn't this instead derive voting power from stake and make sure it is correct?

go/consensus/tendermint/apps/scheduler/query.go
```diff
@@ -58,8 +57,8 @@ func (sq *schedulerQuerier) Validators(ctx context.Context) ([]*scheduler.Valida
 	// Since we use flat voting power for now, doing it this way saves
```
kostko, line 57: This comment should be changed/removed as it is wrong now that the state stores a map of public keys to voting power.

go/scheduler/api/api.go
```diff
@@ -148,6 +148,27 @@ func (c *Committee) EncodedMembersHash() hash.Hash {
 	return hh
 }
 
+// TokensPerVotingPower is the ratio of base units staked to validator power.
+var TokensPerVotingPower quantity.Quantity
+
+func VotingPowerFromTokens(t *quantity.Quantity) (int64, error) {
```
kostko, line 154: Add documentation comment.

pro-wh:
info on the test failure: the `queries` workload of the short txsource e2e test encounters this:

```
Error: workload queries: scheduler queries error: Not enough validators at height 38, expected at least: 1, got: 0
```

go/consensus/tendermint/apps/scheduler/genesis.go
pro-wh: oh, in case something goes wrong in between?

go/consensus/tendermint/apps/scheduler/genesis.go
pro-wh: oh yeah, we should

go/consensus/tendermint/apps/scheduler/genesis.go
kostko: Yeah I know we derive this from our genesis doc anyway so it seems redundant, but on the other hand we also double-check that validators are in the registry.

go/consensus/tendermint/apps/scheduler/query.go
pro-wh: ok removing

go/scheduler/api/api.go
pro-wh: ok

go/consensus/tendermint/apps/scheduler/genesis.go
pro-wh: :vomiting_face:

pro-wh:
info on the test failure:
```
mux: BeginBlock: fatal error in application: '100_staking': disburse fees voters and next proposer: staking: failed to query last block fees: unavailable/corrupted state: context canceled
```

kostko:
This is not the reason for the test failure. The context is cancelled during node shutdown, so this is expected when the test is done.

pro-wh:
ah, the algorithm for diffing the current and previous committees was wrong

pro-wh:
this diffing algorithm doesn't even have good coverage

PR #2873, opened 2020-04-28
go/consensus/tendermint: sync-worker additionally check block timestamps

ptrus:
Fixes: https://github.com/oasislabs/oasis-core/issues/2861

Sync-worker relies on Tendermint fast-sync to determine if the node is still
catching up. This PR adds aditional condition that the latest block is not
older than 1 minute. This prevents cases where node stops fast-syincing
before it has actually caught up.

TODO:
- [x] although not directly testing the change itself done in this PR it would probably be good to have an E2E test that starts a client node at a later time (e.g. after 2 minutes or so), and ensures the node catches up and starts

### DIFF
--- a/go/consensus/tendermint/tendermint.go
+++ b/go/consensus/tendermint/tendermint.go
@@ -127,6 +127,12 @@ const (
 	// StateDir is the name of the directory located inside the node's data
 	// directory which contains the tendermint state.
 	StateDir = "tendermint"
+
+	// Time difference threshold used when considering if node is done with
+	// initial syncing. If difference is greater than the specified threshold
+	// the node is considered not yet synced.
+	// NOTE: this is only used during the initial sync.
+	syncWorkerLastBlockTimeDiffThreshold = 1 * time.Minute
 )
 
 var (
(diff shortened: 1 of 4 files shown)


go/consensus/tendermint/tendermint.go
```diff
@@ -1173,17 +1179,37 @@ func (t *tendermintService) syncWorker() {
...
+					t.Logger.Error("Failed to get tendermint block",
+						"err", err,
+					)
+					return
+				}
+
+				now := time.Now()
+				// No committed blocks or latest block within threshold.
+				if tmBlock == nil || now.Sub(tmBlock.Header.Time) < syncWorkerLastBlockTimeDiffThreshold {
+					close(t.syncedCh)
```
kostko, line 1204: Consider logging something when this happens.

ptrus: done, also added a test

go/oasis-test-runner/scenario/e2e/late_start.go
```diff
@@ -0,0 +1,84 @@
...
+	cmd, err := startClient(childEnv, s.net, resolveClientBinary(s.clientBinary), s.clientArgs)
+	if err != nil {
+		return err
+	}
+	clientErrCh := make(chan error)
+	go func() {
+		clientErrCh <- cmd.Wait()
+	}()
+
+	// Wait for the client to exit.
```
kostko, line 73: Consider just calling `return sc.wait(childEnv, cmd, clientErrCh)` here to avoid code duplication.

Issue #2862, opened 2020-04-21
Registration worker can fall behind during initial sync


ptrus:
Registration worker can fall behind with epoch registrations, mostly noticeable in case where the node exists sync mode before it is actually caught up (related: https://github.com/oasislabs/oasis-core/issues/2861) - where one epoch registration attempt could take more time than it takes to sync one epoch worth of blocks.  

Consider using an overwriting channel ([RingChannel](https://github.com/eapache/channels/blob/master/ring_channel.go)) in `WatchEpochs` cases where it is safe to skip epochs - the registration worker & possibly others.

PR #2876, opened 2020-04-29
go/epochtime: add WatchLatestEpoch method

ptrus:
Fixes: #2862

Pubsub framework is extended to support subscriptions based on bounded
ring channels. Bounded subscriptions are used in the new epochtime
WatchLatestEpoch method. The method is similar to the existing WatchEpochs
method, with the change that unread epochs get overridden with latest
epoch.

Registration worker is changed to use the WatchLatestEpoch method to
prevent trying to register for old epochs in case the worker falls
behind.

XXX: any other places where using a ring channel (e.g. skipping past/outdated channel items) would be preferable?

### DIFF
--- a/go/common/pubsub/pubsub.go
+++ b/go/common/pubsub/pubsub.go
@@ -13,7 +13,7 @@ type broadcastedValue struct {
 }
 
 type cmdCtx struct {
-	ch              *channels.InfiniteChannel
+	ch              channels.Channel
 	errCh           chan error
 	onSubscribeHook OnSubscribeHook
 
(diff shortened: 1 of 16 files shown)


go/common/pubsub/pubsub.go
```diff
@@ -77,37 +77,52 @@ func (s *Subscription) Close() {
...
+	return b.SubscribeEx(int64(channels.Infinity), nil)
+}
+
+// SubscribeBuff subscribes to the Broker's broadcasts, and returns a
+// subscription handle that can be used to receive broadcasts.
+//
+// Buffer controlls the capacity of a ring buffer - when buffer is full the
+// oldest value will be discarded. In case buffer is negative (or zero) an
+// unbounded channel is used.
+func (b *Broker) SubscribeBuff(buffer int64) *Subscription {
```
kostko, line 106:
```suggestion
func (b *Broker) SubscribeBuffered(buffer int64) *Subscription {
```

Issue #2880, opened 2020-05-03
State sync for the Tendermint consensus backend


kostko:
We should support state sync for the Tendermint consensus backend. Also requires at least basic support for consensus light clients (see #2440).

Issue #2440, opened 2019-11-28
Light consensus client/node


kostko:
We should support a light consensus client so that compute nodes which are not validators don't need to sync and execute everything.

Requires #1898 to make it easy to do remote consensus queries after verifying the headers and getting the state roots.

Related to #2349 to avoid compute nodes from doing irrelevant work.

## Linked Issues
* [x] [Basic consensus light client support](https://github.com/oasislabs/oasis-core/pull/2881)
* [x] Tendermint light client provider (part of #2882).
* [x] [Consensus light client interface should expose read-only state](https://github.com/oasisprotocol/oasis-core/issues/3077)
* [x] [go/consensus/tendermint: Refactor internal event handling](https://github.com/oasisprotocol/oasis-core/pull/3091)
* [x] [Make sure TendermintService is only used internally](https://github.com/oasisprotocol/oasis-core/issues/3115)
* [x] [Introduce the notion of consensus backend mode](https://github.com/oasisprotocol/oasis-core/issues/3116)
* [x] [Add non-blocking consensus SubmitTx variant](https://github.com/oasisprotocol/oasis-core/issues/3152)
* [x] [Move `SubmitEvidence` to `LightClientBackend`](https://github.com/oasisprotocol/oasis-core/issues/3169)
* [ ] Light version of the Tendermint consensus backend which uses the light client internally.

PR #2881, opened 2020-05-03
Basic consensus light client support

kostko:
Initial implementation of #2440, required for #2880.

TODO
* [x] Cleanup (extracted from branch which also bumps Tendermint to an unreleased version, see #2882).
* [x] More tests where missing.

### DIFF
--- a/go/storage/mkvs/checkpoint/checkpoint.go
+++ b/go/storage/mkvs/checkpoint/checkpoint.go
@@ -50,6 +50,10 @@ type ChunkProvider interface {
 type GetCheckpointsRequest struct {
 	Version   uint16           `json:"version"`
 	Namespace common.Namespace `json:"namespace"`
+
+	// RootVersion specifies an optional root version to limit the request to. If specified, only
+	// checkpoints for roots with the specific version will be considered.
+	RootVersion *uint64 `json:"root_version,omitempty"`
 }
 
 // Creator is a checkpoint creator.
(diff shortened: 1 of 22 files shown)


go/worker/consensus/worker.go
```diff
@@ -0,0 +1,95 @@
...
+		commonWorker: commonWorker,
+		quitCh:       make(chan struct{}),
+		logger:       logging.GetLogger("worker/consensus"),
+	}
+
+	if w.enabled {
+		// Register the consensus light client service.
+		consensus.RegisterLightService(commonWorker.Grpc.Server(), commonWorker.Consensus)
+
+		// TODO: We could introduce a role bit for nodes providing these public services so that you
```
kostko, line 79: Any thoughts on whether we should do this now? It would just be allocating a role bit that wouldn't otherwise be used for anything (e.g., any node could signal it and there would be no other requirements for now).

Yawning: Probably should, if only to make finding nodes for this less annoying.

kostko: Yeah that was my thought as well, will add it here.

go/storage/mkvs/checkpoint/checkpoint.go
```diff
@@ -125,16 +133,25 @@ type Metadata struct {
 	Chunks  []hash.Hash `json:"chunks"`
 }
 
+// EncodedHash returns the encoded cryptographic hash of the checkpoint metadata.
+func (m *Metadata) EncodedHash() hash.Hash {
+	var hh hash.Hash
+
+	hh.From(m)
```
Yawning, line 140: Doesn't need to be now, and we do this all over the place, but at some point we should add `func NewFrom(v interface{}) Hash`.

kostko: Ah, good idea! I'll do this in a separate PR.

kostko: Done in #2890.

Issue #1733, opened 2019-05-21
Run SGX tests with IAS and check AVRs


tjanez:
Now that we [also run End-to-End tests on intel-sgx TEE](https://github.com/oasislabs/ekiden/pull/1543), we should also run them with IAS (Intel Attestation Services) and check that AVR (Attestation Verification Report) verification works.

## Details

To enable this, the relevant Buildkite agents will need to have Oasis' IAS credentials provisioned.
Lastly, the tests will need to unset the `OASIS_UNSAFE_SKIP_AVR_VERIFY` environment variable during building of SGX artifacts and during running the SGX tests.

## Acceptance Criteria
* End-to-End test running on intel-sgx TEE run successfully with AVR verification enabled.

PR #2883, opened 2020-05-04
go/oasis-test-runner: add support for running tests with non-mock IAS

ptrus:
Fixes: #1733

TODO:
- [x] test on actual sgx instance 
- [x] update to work with latest (v4) IAS API (ref: https://api.trustedservices.intel.com/documents/sgx-attestation-api-spec.pdf)
  - no more client certs, but client authentication is done with an API key
  - new statuses: `SW_HARDENING_NEEDED` & `CONFIGURATION_AND_SW_HARDENING_NEEDED`
  - advisories moved from response headers into actual report
- [x] setup a job that runs a test with actual IAS (probably daily on master)

### DIFF
--- a/go/common/sgx/ias/avr.go
+++ b/go/common/sgx/ias/avr.go
@@ -11,8 +11,6 @@ import (
 	"net/url"
 	"time"
 
-	"github.com/pkg/errors"
-
 	"github.com/oasislabs/oasis-core/go/common/sgx"
 )
 
@@ -32,14 +30,16 @@ const TimestampFormat = "2006-01-02T15:04:05.999999999"
 
 var (
 	isvQuoteFwdMap = map[string]ISVEnclaveQuoteStatus{
-		"OK":                     QuoteOK,
-		"SIGNATURE_INVALID":      QuoteSignatureInvalid,
-		"GROUP_REVOKED":          QuoteGroupRevoked,
-		"SIGNATURE_REVOKED":      QuoteSignatureRevoked,
-		"KEY_REVOKED":            QuoteKeyRevoked,
-		"SIGRL_VERSION_MISMATCH": QuoteSigRLVersionMismatch,
-		"GROUP_OUT_OF_DATE":      QuoteGroupOutOfDate,
-		"CONFIGURATION_NEEDED":   QuoteConfigurationNeeded,
+		"OK":                                    QuoteOK,
+		"SIGNATURE_INVALID":                     QuoteSignatureInvalid,
+		"GROUP_REVOKED":                         QuoteGroupRevoked,
+		"SIGNATURE_REVOKED":                     QuoteSignatureRevoked,
+		"KEY_REVOKED":                           QuoteKeyRevoked,
+		"SIGRL_VERSION_MISMATCH":                QuoteSigRLVersionMismatch,
+		"GROUP_OUT_OF_DATE":                     QuoteGroupOutOfDate,
+		"CONFIGURATION_NEEDED":                  QuoteConfigurationNeeded,
+		"SW_HARDENING_NEEDED":                   QuoteSwHardeningNeeded,
+		"CONFIGURATION_AND_SW_HARDENING_NEEDED": QuoteConfigurationAndSwHardeningNeeded,
 	}
 	isvQuoteRevMap = make(map[ISVEnclaveQuoteStatus]string)
 
@@ -88,6 +88,8 @@ const (
 	QuoteSigRLVersionMismatch
 	QuoteGroupOutOfDate
 	QuoteConfigurationNeeded
+	QuoteSwHardeningNeeded
+	QuoteConfigurationAndSwHardeningNeeded
 )
 
 // UnmarshalText implements the encoding.TextUnmarshaler interface.
@@ -208,6 +210,8 @@ type AttestationVerificationReport struct {
 	PlatformInfoBlob      string                `json:"platformInfoBlob"`
 	Nonce                 string                `json:"nonce"`
 	EPIDPseudonym         []byte                `json:"epidPseudonym"`
+	AdvisoryURL           string                `json:"advisoryURL"`
+	AdvisoryIDs           []string              `json:"advisoryIDs"`
 }
 
 // Quote decodes and returns the enclave quote component of an Attestation
@@ -227,10 +231,10 @@ func (a *AttestationVerificationReport) validate() error { // nolint: gocyclo
 	)
 
 	if _, err := time.Parse(TimestampFormat, a.Timestamp); err != nil {
-		return errors.Wrap(err, "ias/avr: invalid timestamp")
+		return fmt.Errorf("ias/avr: invalid timestamp: %w", err)
 	}
 
-	// TODO: Enforce version once version 3 test vectors are available.
+	// TODO: Enforce version once version 4 test vectors are available.
 
 	if a.ISVEnclaveQuoteStatus == quoteFieldMissing {
 		return fmt.Errorf("ias/avr: missing isvEnclaveQuoteStatus")
@@ -241,10 +245,10 @@ func (a *AttestationVerificationReport) validate() error { // nolint: gocyclo
 	case quoteLen:
 		untrustedQuote, err := a.Quote()
 		if err != nil {
-			return errors.Wrap(err, "ias/avr: malformed quote")
+			return fmt.Errorf("ias/avr: malformed quote: %w", err)
 		}
 		if err = untrustedQuote.Verify(); err != nil {
-			return errors.Wrap(err, "ias/avr: invalid quote")
+			return fmt.Errorf("ias/avr: invalid quote: %w", err)
 		}
 	default:
 		return fmt.Errorf("ias/avr: invalid isvEnclaveQuoteBody length")
@@ -260,7 +264,7 @@ func (a *AttestationVerificationReport) validate() error { // nolint: gocyclo
 
 	if a.PSEManifestStatus != nil {
 		switch a.ISVEnclaveQuoteStatus {
-		case QuoteOK, QuoteGroupOutOfDate, QuoteConfigurationNeeded:
+		case QuoteOK, QuoteGroupOutOfDate, QuoteConfigurationNeeded, QuoteSwHardeningNeeded, QuoteConfigurationAndSwHardeningNeeded:
 		default:
 			return fmt.Errorf("ias/avr: unexpected pseManifestStatus")
 		}
@@ -268,7 +272,7 @@ func (a *AttestationVerificationReport) validate() error { // nolint: gocyclo
 
 	pseHash, err := hex.DecodeString(a.PSEManifestHash)
 	if err != nil {
-		return errors.Wrap(err, "ias/avr: failed to decode pseManifestHash")
+		return fmt.Errorf("ias/avr: failed to decode pseManifestHash: %w", err)
 	}
 	switch len(pseHash) {
 	case 0, pseManifestHashLen:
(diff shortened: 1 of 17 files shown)


go/common/sgx/ias/avr.go
```diff
@@ -208,6 +212,9 @@ type AttestationVerificationReport struct {
 	PlatformInfoBlob      string                `json:"platformInfoBlob"`
 	Nonce                 string                `json:"nonce"`
 	EPIDPseudonym         []byte                `json:"epidPseudonym"`
+	// TODO: these were moved from response headers into the report, anything to do with these?
```
kostko, line 215: I think not?

go/ias/http/http.go
```diff
@@ -230,59 +238,21 @@ func New(cfg *Config) (api.Endpoint, error) {
...
-			return nil, errors.Wrap(err, "ias: failed to verify client certificate CA")
-		}
-	}
-
 	e := &httpEndpoint{
 		httpClient: &http.Client{
-			Transport: &http.Transport{
-				TLSClientConfig: tlsConfig,
-			},
+			Timeout: 10 * time.Second,
```
kostko, line 243: Arbitrary timeout? Move it to a constant.

go/oasis-node/cmd/ias/proxy.go
```diff
@@ -209,35 +204,18 @@ func iasEndpointFromFlags() (ias.Endpoint, error) {
...
-
-		cfg.IsProduction = viper.GetBool(cfgIsProduction)
-	} else {
+	if viper.GetBool(cfgDebugMock) {
 		if !flags.DebugDontBlameOasis() {
 			return nil, fmt.Errorf("ias: refusing to mock IAS responses")
 		}
 		cfg.DebugIsMock = true
+	} else {
+		apikey := viper.GetString(cfgAuthAPIKey)
```
kostko, line 213:
```suggestion
		apiKey := viper.GetString(cfgAuthAPIKey)
```

go/oasis-test-runner/oasis/oasis.go
```diff
@@ -249,6 +249,15 @@ type NetworkCfg struct { // nolint: maligned
 	// genesis document for authenticating runtime IDs.
 	IASUseRegistry bool `json:"ias_use_registry,omitempty"`
 
+	// IASMock specifies if Mock IAS Proxy should be used.
```
kostko, line 252: Lots of IAS-related options now, could we group them under a single `IAS` field and have a dedicated structure?

go/ias/http/http.go
```diff
@@ -185,11 +194,8 @@ func (e *mockEndpoint) Cleanup() {
 
 // Config is the IAS HTTP endpoint configuration.
 type Config struct {
-	// AuthCert is the IAS authentication certificate (and private key).
-	AuthCert *tls.Certificate
-
-	// AuthCertCA is the CA cert for the IAS authentication certificate.
-	AuthCertCA *x509.Certificate
+	// SubscriptionKey is the IAS API key used for client authentication.
+	SubscriptionKey string
```
kostko, line 198: Nice API keys instead of TLS certificates.

go/ias/http/http.go
```diff
@@ -63,42 +82,33 @@ func (e *httpEndpoint) VerifyEvidence(ctx context.Context, evidence *api.Evidenc
...
 	}
 
 	// Dispatch the request via HTTP.
-	u := *e.baseURL
-	u.Path = path.Join(u.Path, "/attestation/sgx/v3/report")
-	resp, err := ctxhttp.Post(ctx, e.httpClient, u.String(), "application/json", bytes.NewReader(reqPayload))
+	resp, err := e.doIASRequest(ctx, "POST", "/attestation/v4/report", "application/json", bytes.NewReader(reqPayload))
 	if err != nil {
-		return nil, errors.Wrap(err, "ias: http POST failed")
+		logger.Error("ias request error", "err", err, "request", reqPayload)
```
kostko, line 91: Could log the error in `doIASRequest` and also log the method/path.

go/ias/http/http.go
```diff
@@ -63,42 +82,33 @@ func (e *httpEndpoint) VerifyEvidence(ctx context.Context, evidence *api.Evidenc
...
+		return nil, fmt.Errorf("ias: http POST failed: %w", err)
 	}
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
-		return nil, errors.Wrapf(err, "ias: http POST returned error: %s", http.StatusText(resp.StatusCode))
+		logger.Error("ias request error", "status", resp.StatusCode, "request", reqPayload, "resp", resp)
```
kostko, line 97: Could check status and log the error in `doIASRequest`.

go/ias/http/http.go
```diff
@@ -230,59 +238,21 @@ func New(cfg *Config) (api.Endpoint, error) {
...
+		subscriptionKey: cfg.SubscriptionKey,
+		trustRoots:      ias.IntelTrustRoots,
 		spidInfo: api.SPIDInfo{
 			SPID:               spidBin,
 			QuoteSignatureType: cfg.QuoteSignatureType,
 		},
 	}
 	if cfg.IsProduction {
-		e.baseURL, _ = url.Parse("https://as.sgx.trustedservices.intel.com/")
+		e.baseURL, _ = url.Parse("https://api.trustedservices.intel.com/sgx")
```
kostko, line 253: Maybe move these URLs to constants?

go/ias/http/http.go
```diff
@@ -63,42 +82,33 @@ func (e *httpEndpoint) VerifyEvidence(ctx context.Context, evidence *api.Evidenc
...
 	if err != nil {
-		return nil, errors.Wrap(err, "ias: failed to marshal")
+		return nil, fmt.Errorf("ias: failed to marshal: %w", err)
 	}
 
 	// Dispatch the request via HTTP.
-	u := *e.baseURL
-	u.Path = path.Join(u.Path, "/attestation/sgx/v3/report")
-	resp, err := ctxhttp.Post(ctx, e.httpClient, u.String(), "application/json", bytes.NewReader(reqPayload))
+	resp, err := e.doIASRequest(ctx, "POST", "/attestation/v4/report", "application/json", bytes.NewReader(reqPayload))
```
kostko, line 89: Maybe move these paths to constants?

go/common/sgx/ias/avr.go
```diff
@@ -231,6 +237,7 @@ func (a *AttestationVerificationReport) validate() error { // nolint: gocyclo
 	}
 
 	// TODO: Enforce version once version 3 test vectors are available.
+	// Enforce to version 4 now?
```
ptrus, line 240: Should we enforce version here now?

.buildkite/scripts/daily_sgx_ias.sh
```diff
@@ -0,0 +1,27 @@
...
+OASIS_IAS_APIKEY=$(cat ~/.oasis-ias/api_key)
+export OASIS_IAS_APIKEY
+OASIS_IAS_SPID=$(cat ~/.oasis-ias/spid)
+export OASIS_IAS_SPID
+set -x
+
+if [[ $BUILDKITE_RETRY_COUNT == 0 ]]; then
+    make all
+    # XXX: add more tests?
+    .buildkite/scripts/test_e2e.sh -t e2e/runtime/runtime-encryption
```
ptrus, line 17: actually building everything and running this one test case takes ~5-7mins, we could probably run this on PR branches (assuming we can increase the number of sgx ci agents by 1 probably so it would run in parallel, and that IAS doesn't randomly fail at times)

go/common/sgx/ias/avr.go
kostko: I think we should enforce version 4, yeah.

cc @Yawning about the test vectors.

.buildkite/scripts/daily_sgx_ias.sh
kostko: Maybe we can start by always running this in the master branch? And then if it seems stable and relatively fast we can change it to run on all PRs?

go/oasis-test-runner/scenario/e2e/runtime.go
```diff
@@ -102,6 +107,10 @@ func (sc *runtimeImpl) Parameters() *flag.FlagSet {
 	f.StringVar(&sc.runtimeBinaryDir, cfgRuntimeBinaryDir, sc.runtimeBinaryDir, "path to the runtime binaries directory")
 	f.StringVar(&sc.runtimeLoader, cfgRuntimeLoader, sc.runtimeLoader, "path to the runtime loader")
 	f.StringVar(&sc.TEEHardware, cfgTEEHardware, sc.TEEHardware, "TEE hardware to use")
+	// XXX: ideally iasMock would default to true, but the Clone/Parameters
```
kostko, line 110: Hm, can you explain a bit more what would need to change? cc @matevz

go/oasis-test-runner/oasis/oasis.go
```diff
@@ -213,6 +213,16 @@ type Network struct {
 	errCh chan error
 }
 
+// IASCfg is the Oasis test network IAS configuration.
+type IASCfg struct {
+	// UseRegistry specifies whether the IAS proxy should use the registry
+	// instead of the genesis document for authenticating runtime IDs.
+	UseRegistry bool `json:"ias_use_registry,omitempty"`
```
kostko, line 220:
```suggestion
	UseRegistry bool `json:"use_registry,omitempty"`
```

go/oasis-test-runner/oasis/oasis.go
```diff
@@ -213,6 +213,16 @@ type Network struct {
...
+
+	// Mock specifies if Mock IAS Proxy should be used.
+	Mock bool `json:"ias_mock,omitempty"`
```
kostko, line 223:
```suggestion
	Mock bool `json:"mock,omitempty"`
```

.buildkite/scripts/daily_sgx_ias.sh
ptrus: Good idea, i'll change it to that, thanks!

go/oasis-test-runner/scenario/e2e/runtime.go
ptrus: `Parameters()` always returns a new flagset (which feels wrong, i think new flagset should only be created in `Clone()` - but didn't want to change that in this PR).

The above in conjunction with:
https://github.com/oasislabs/oasis-core/blob/a408af740b9ba84b14743a821572a702dc59f557/go/oasis-test-runner/cmd/root.go#L133-L135

The already overridden parameter gets replaced with the default value every time this is called, if the existing var isn't passed in as `default`. This can be fixed by calling `sCloned.Parameters()` once and storing the returned flagset. 

But didn't check if any other code expects the same - and a more proper fix would be `Parameters()` not creating a new flagset i think.

kostko: Ok, feel free to convert your comment to an issue and reference it in the comment ;-)

go/oasis-test-runner/oasis/oasis.go
ptrus: thanks for spotting these two :-)

.buildkite/scripts/daily_sgx_ias.sh
ptrus: Updated, it runs as part of the main pipeline now: https://buildkite.com/oasislabs/oasis-core-ci/builds/3746#cfbade48-fb2e-4bea-afe6-e116cfc5d768 and since i was able to spawn one more sgx agent and the test seems pretty stable I enabled it to run on all PR branches

kostko:
Wait, what's up with coverage?

ptrus:
> Wait, what's up with coverage?

fixed 

> increase coverage by 0.27%

Also the IAS-proxt http is now covered: https://codecov.io/gh/oasislabs/oasis-core/src/ptrus%2Ffeature%2Fci-use-ias/go/ias/http/http.go

go/common/sgx/ias/avr.go
ptrus: i didn't find anything, so i just updated the comment for now :grin:

PR #2884, opened 2020-05-04
Matevz/feature/new metrics

matevz:
Resolves https://github.com/oasislabs/oasis-core/issues/2842


Yawning:
So as I noted elsewhere, why not instantiate prometheus and query the metric metadata (https://prometheus.io/docs/prometheus/latest/querying/api/#querying-metric-metadata).

At a minimum if you are going to parse source files for annotations, use `go/parser`...

matevz:
> So as I noted elsewhere, why not instantiate prometheus and query the metric metadata (https://prometheus.io/docs/prometheus/latest/querying/api/#querying-metric-metadata).
> 
> At a minimum if you are going to parse source files for annotations, use `go/parser`...

This is now implemented in https://github.com/oasislabs/oasis-core/pull/2877

go/consensus/tendermint/tendermint.go
```diff
@@ -1280,6 +1286,50 @@ func (t *tendermintService) worker() {
...
+		case <-sub.Cancelled():
+			return
+		case v := <-sub.Out():
+			ev := v.Data().(tmtypes.EventDataNewBlock)
+			blk = ev.Block
+		}
+
+		// Tendermint uses specific public key encoding.
+		pubKey := t.consensusSigner.Public()
+		myAddr := []byte(crypto.PublicKeyToTendermint(&pubKey).Address())
```
Yawning, line 1314: Just cache this, it doesn't change currently.

.changelog/2842.feature.md
```diff
@@ -0,0 +1,12 @@
+Add new consensus-related Prometheus metrics
+
+Four new metrics have been added:
+
+- `oasis_worker_epoch_number` is the current epoch number as seen by the
+  worker.
+- `oasis_registry_node_registered` is a binary metric which denotes, if the
+  node is registered validator.
```
ptrus, line 8: i think the code just checks if the node is registered (not a registered validator)

go/registry/metrics.go
```diff
@@ -100,12 +111,48 @@ func (m *MetricsUpdater) updatePeriodicMetrics(ctx context.Context) {
...
+	// Check, if node is a registered validator.
+	// XXX: Is there an existing nodeIdentity instance?
+	dataDir, err := cmdCommon.DataDirOrPwd()
+	if err != nil {
+		m.logger.Error("failed to query data directory",
+			"err", err,
+		)
+		return
+	}
+	nodeSignerFactory, err := fileSigner.NewFactory(dataDir, signature.SignerNode, signature.SignerP2P, signature.SignerConsensus)
```
Yawning, line 124: Oh hell no.

go/registry/metrics.go
```diff
@@ -33,10 +37,17 @@ var (
 			Help: "Number of registry runtimes.",
 		},
 	)
-	registeryCollectors = []prometheus.Collector{
+	registryNodeRegistered = prometheus.NewGauge(
```
kostko, line 40: This should probably be moved to the registration worker in `go/worker/registration` as that will trigger registration in the first place. No need to check every time whether the node is registered.

go/registry/metrics.go
matevz: Moved to go/worker/registration.

go/consensus/metrics/metrics.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/spf13/viper"
+
+	"github.com/oasislabs/oasis-core/go/oasis-node/cmd/common/metrics"
+)
+
+var (
+	SignedBlocks = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "oasis_consensus_signed_blocks",
+			Help: "Number of signed blocks by the node.",
```
kostko, line 16:
```suggestion
			Help: "Number of blocks signed by the node.",
```

go/consensus/metrics/metrics.go
```diff
@@ -0,0 +1,42 @@
...
+		},
+		[]string{"backend"},
+	)
+	ProposedBlocks = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "oasis_consensus_proposed_blocks",
+			Help: "Number of proposed blocks by the node.",
```
kostko, line 23:
```suggestion
			Help: "Number of blocks proposed by the node.",
```

go/consensus/tendermint/tendermint.go
```diff
@@ -257,6 +263,9 @@ func (t *tendermintService) Start() error {
 		}
 		go t.syncWorker()
 		go t.worker()
+		if viper.GetString(cmmetrics.CfgMetricsAddr) != "" {
```
kostko, line 266: Make this based on the configured metrics mode instead and enable it if the mode is not none.

go/consensus/tendermint/tendermint.go
```diff
@@ -1280,6 +1289,49 @@ func (t *tendermintService) worker() {
...
+		}
+
+		// Was block proposed by our node.
+		if bytes.Equal(myAddr, blk.ProposerAddress) {
+			metrics.ProposedBlocks.With(labelTendermint).Inc()
+		}
+
+		// Was block voted for by our node. Ignore if there was no previous block.
+		if blk.LastCommit != nil {
+			for _, a := range blk.LastCommit.Signatures {
```
kostko, line 1325: Rename `a` to `sig`.

go/consensus/tendermint/tendermint.go
```diff
@@ -1280,6 +1289,49 @@ func (t *tendermintService) worker() {
...
+				if bytes.Equal(myAddr, a.ValidatorAddress) {
```
kostko, line 1326: This is not correct as it also includes missing votes. If `sig.IsAbsent()` or `sig.BlockIDFlag == tmtypes.BlockIDFlagNil` the vote should not be counted.

go/worker/registration/worker.go
```diff
@@ -656,6 +672,7 @@ func (w *Worker) registerNode(epoch epochtime.EpochTime, hook RegisterNodeHook)
 	}
 
 	w.logger.Info("node registered with the registry")
+	workerNodeRegistered.Set(1.0)
```
kostko, line 675: What about resetting to zero in case the registration transaction failed (e.g., a few lines above, but basically any of the error cases)? I would move setting this metric to where `registerNode` is called, so that you can check the error code (and if there is an error you need to reset it to zero).

Makefile
```diff
@@ -65,7 +69,7 @@ fmt-go:
 fmt: $(fmt-targets)
 
 # Lint code, commits and documentation.
-lint-targets := lint-go lint-git lint-md lint-changelog
+lint-targets := lint-go lint-git lint-md lint-changelog check-docs
```
kostko, line 72: Can you change the ci-lint workflow to actually run this shiny new `check-docs` target (maybe rename to `lint-docs` to be consistent)? See `.github/workflows/ci-lint.yml`.

go/worker/registration/worker.go
```diff
@@ -240,7 +254,13 @@ func (w *Worker) registrationLoop() { // nolint: gocyclo
 			default:
 			}
 
-			return w.registerNode(epoch, hook)
+			var err error
+			if err = w.registerNode(epoch, hook); err == nil {
```
kostko, line 258: Would be more idiomatic if this was a switch statement.

kostko:
Don't forget to squash!

PR #2888, opened 2020-05-05
go/runtime/client: Actually store the created key manager client

### LINKED ISSUE #2885: go/keymanager/client(?): Intermittent errors when committe size > 1
I have been told that there is retry logic, and both the rust enclave RPC and go key manager clients appear to have something that looks like retry logic.

That said, when there is more than one node in a key manager committee, a simple non-enclave key manager request (via gRPC) from rust code intermittently fails with the following panic:

```
Testing key manager connection via gRPC transport...
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: RpcFailure(RpcStatus { status: Unknown, details: Some("worker/keymanager: error from runtime: decrypt error") })', tests/clients/simple-keyvalue-enc/src/main.rs:99:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Node logs: https://buildkite.com/oasislabs/oasis-core-ci/builds/3614 (The keymanager-replication tests in the 6/7 non-sgx group)

I assume something, somewhere should be re-establishing the session as part of the retry logic, or something somewhere shouldn't be as happy to swap to a different node.

kostko:
Should fix #2885


go/runtime/client/client.go
```diff
@@ -459,6 +459,8 @@ func (c *runtimeClient) CallEnclave(ctx context.Context, request *enclaverpc.Cal
 		var km *keymanager.Client
 		c.Lock()
 		if km = c.kmClients[rt.ID()]; km == nil {
+			c.logger.Debug("creating new key manager client instance")
```
pro-wh, line 462: maybe log the km.ID() too

PR #2892, opened 2020-05-06
go scheduler: increase tokens per voting power

pro-wh:
we'll need this to fit under tendermint's maximum total voting power limit

### DIFF
--- a/go/scheduler/api/api.go
+++ b/go/scheduler/api/api.go
@@ -262,8 +262,8 @@ func (g *Genesis) SanityCheck(stakingTotalSupply *quantity.Quantity) error {
 }
 
 func init() {
-	// 2 allows for up to 1.8e19 base units to be staked.
-	if err := TokensPerVotingPower.FromUint64(2); err != nil {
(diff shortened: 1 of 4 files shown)


pro-wh:
I'm adding the sanity check and tests and changelog

pro-wh:
@kostko the new sanity check references the limit in code, which acts to explain why `16` allows up to 1.8e19.

go/scheduler/api/api.go
```diff
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"strings"
 
+	tmtypes "github.com/tendermint/tendermint/types"
```
kostko, line 9: Ow, this brings in tendermint to the consensus agnostic part (the scheduler API). I am fine with defining the constant such that it satisfies the Tendermint limits and mention that the constant is such that it works with the Tendermint backend, but we should not import tendermint here. In case this ever changes in Tendermint, the tendermint-specific sanity check (that imports tendermint) will fail.

pro-wh: ok

PR #2893, opened 2020-05-07
runtime: Do SGX enclave signing ourself

Yawning:
* [x] Add the ability to generate signed SIGSTRUCTs, without using the fortanix SGX SDK.
 * [x] Generate and pass the signature to the runtime loader (with a mock key for now).


go/runtime/host/sgx/sgx.go
```diff
@@ -74,30 +87,80 @@ type sgxProvisioner struct {
...
+	}
+
+	// If the path to an existing SIGSTRUCT is provided, load it.
+	if rtExtra, ok := rtCfg.Extra.(*RuntimeExtra); ok {
+		sig, err = ioutil.ReadFile(rtExtra.SignaturePath)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to load SIGSTRUCT: %w", err)
+		}
+	} else {
+		// HACK: Just generate one with the test key for now.
```
kostko, line 111: Maybe log something?

go/worker/common/config.go
```diff
@@ -42,6 +42,9 @@ var (
 	// runtime IDs to corresponding resource paths (type of the resource depends on the
 	// provisioner).
 	CfgRuntimePaths = "worker.runtime.paths"
+	// CfgRuntimeSignatures configures signatures for supported runtimes.  The value should
+	// be a map of runtime IDs to correspondign resource paths.
```
kostko, line 46:
```suggestion
	// be a map of runtime IDs to corresponding resource paths.
```

go/worker/common/config.go
```diff
@@ -42,6 +42,9 @@ var (
...
+	CfgRuntimeSignatures = "worker.runtime.signatures"
```
kostko, line 47: Since this is SGX-specific, maybe move this to `worker.runtime.sgx.signatures` as the above loader option?

go/worker/common/config.go
```diff
@@ -188,17 +191,34 @@ func NewConfig(ias ias.Endpoint) (*Config, error) {
...
+			// right now that needs this anyway, the non-SGX provisioner
+			// currently ignores this.
+			if sigPath := runtimeSignatures[runtimeID]; sigPath != "" {
+				runtimeHostCfg.Extra = &hostSgx.RuntimeExtra{
+					SignaturePath: sigPath,
+				}
+			} else {
+				// HACK HACK HACK: Allow dummy SIGSTRUCT generation.
+				runtimeHostCfg.Extra = &hostSgx.RuntimeExtra{
+					UnsafeDebugGenerateSigstruct: true,
```
kostko, line 217: Require don't blame oasis? I know that if don't blame oasis is not set, getting the dummy key will fail anyway, but maybe do it here for good measure?

PR #2896, opened 2020-05-07
go stats: new availability v3 formula

pro-wh:
here's the implementation of the new availability score v3 for May

### DIFF
--- a/.changelog/2896.feature.md
+++ b/.changelog/2896.feature.md
@@ -1,0 +1,3 @@
+go/stats: New availability v3 formula
+
+Here's the implementation of the new availability score v3 for May.
--- a/go/extra/stats/cmd/stats.go
+++ b/go/extra/stats/cmd/stats.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"os"
@@ -10,6 +11,9 @@ import (
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
 	"github.com/spf13/viper"
+	tmamino "github.com/tendermint/go-amino"
+	tmrpctypes "github.com/tendermint/tendermint/rpc/core/types"
+	tmtypes "github.com/tendermint/tendermint/types"
 
 	"github.com/oasislabs/oasis-core/go/common/cbor"
 	"github.com/oasislabs/oasis-core/go/common/crypto/signature"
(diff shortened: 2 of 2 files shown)


.changelog/2896.feature.md
```diff
@@ -0,0 +1,3 @@
+go/stats: New availability v3 formula
```
kostko, line 1:
```suggestion
go/extra/stats: New availability v3 formula
```

go/extra/stats/cmd/stats.go
```diff
@@ -248,74 +282,84 @@ func getStats(ctx context.Context, consensus consensusAPI.ClientBackend, registr
...
-				logger.Error("failed to query registry",
+		tmBlockMeta := getTmBlockMetaOrExit(ctx, consensus, height)
+
+		// Process the commit that is put on chain in this block.
+		if height > 1 {
+			// Commit is for previous height.
+			lastCommitHeight := height - 1
+
+			// Get validators.
+			// Hypothesis: this gets the validator set with priorities after they're incremented to get the round 0
```
kostko, line 293: I assume you checked that this is actually the case?

go/extra/stats/cmd/stats.go
```diff
@@ -248,74 +282,84 @@ func getStats(ctx context.Context, consensus consensusAPI.ClientBackend, registr
...
 
-				// Add proposal.
-				if err := stats.addNodeProposal(previousProposerAddr); err != nil {
-					logger.Error("failure adding proposal",
-						"err", err,
-					)
-					os.Exit(1)
+			// Go over all signatures for a block.
+			for _, sig := range tmBlockMeta.LastCommit.Signatures {
+				if sig.Absent() {
```
kostko, line 320: While we're here, what about signatures for nil? Are those recorded too (I believe with the Tendermint upgrade they are) and should they be treated as "not signing"?

pro-wh: don't know

go/extra/stats/cmd/stats.go
pro-wh: I get the gist of it now

go/extra/stats/cmd/stats.go
peterjgilbert: If you voted for `nil` in a round that succeeded, you failed to sign the block that was committed. Is this any more useful that not signing?

pro-wh: it's an online conscientious objection

pro-wh: I'll make it not count. only popular answer wins ):

PR #2898, opened 2020-05-08
metrics: Ignore labels with empty values

matevz:
Workaround for https://github.com/prometheus/pushgateway/issues/344

Prometheus push gateway rejects labels with empty values (e.g. when SGX is not used `tee_hardware` label is empty). This PR removes labels with empty values before sending them to push gateway.

### DIFF
--- a/go/oasis-node/cmd/common/metrics/metrics.go
+++ b/go/oasis-node/cmd/common/metrics/metrics.go
@@ -272,6 +272,18 @@ func GetDefaultPushLabels(ti *env.TestInstanceInfo) map[string]string {
 		for k, v := range flags.GetStringMapString(CfgMetricsLabels) {
 			labels[k] = v
 		}
+
+		// Remove empty label values - workaround for
+		// https://github.com/prometheus/pushgateway/issues/344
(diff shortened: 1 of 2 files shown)


go/oasis-node/cmd/common/metrics/metrics.go
```diff
@@ -272,6 +272,18 @@ func GetDefaultPushLabels(ti *env.TestInstanceInfo) map[string]string {
 		for k, v := range flags.GetStringMapString(CfgMetricsLabels) {
 			labels[k] = v
 		}
+
+		// Remove empty label values - workaround for
+		// https://github.com/prometheus/pushgateway/issues/344
```
kostko, line 277: File issue to remove this workaround once fixed upstream.

matevz:
oasis-core-daily-benchmarks now work back with this fix.

go/oasis-node/cmd/common/metrics/metrics.go
matevz: Opened https://github.com/oasislabs/oasis-core/issues/2899

PR #2906, opened 2020-05-12
common/flags: Fix parsing of metrics.labels, if provided in config .yml

### LINKED ISSUE #2905: metrics.labels not parsed correctly for .yml configs
Running `oasis-node --metrics.labels foo1=bar1,foo2=bar2` works. However, making a yaml file:
```
metrics:
  labels:
    foo1: bar1
    foo2: bar2
```
complains that no labels have been passed.

This regression was caused by #2687.

matevz:
Fixes #2905


kostko:
Actually, looks like this was supposed to be fixed in viper 1.7.0 (see https://github.com/spf13/viper/pull/874), can you test?

matevz:
Fixed upstream, removed drop-in.

PR #2907, opened 2020-05-12
Refactor the runtime host APIs

### LINKED ISSUE #2801: Make runtime host API more flexible
In order to make the runtime host API more flexible to support different execution models in the future, we should refactor the existing worker host into the following parts:
  * `go/runtime/host/protocol` (from existing `go/worker/common/host/protocol`)
  * `go/runtime/host/manager.go` (common part of `go/worker/common/host/sandboxed.go` that manages Oasis Core Runtime Enclaves, regardless of what sandboxing/confidentiality platform they use)
  * `go/runtime/host/sgx` (provisioner for Intel SGX using Bubblewrap and runtime-loader).

While refactoring the API we should consider use cases where the runtime abstraction (VM/enclave) is pre-provisioned (e.g., externally provisioned Oasis Core VM and one or more Oasis Core Runtime VMs together with some interconnect).

In addition to the runtime host protocol channel there should also be a separate channel (pipe) for logs from the runtime (related to #1819). The actual transport should be abstracted.

kostko:
Fixes #2801 

Several changes were made to the runtime host APIs used to provision and
communicate with runtimes:

* The runtime host implementation has been moved to `go/runtime/host`.

* Some of the runtime host protocol types have been changed, all references
  to `Worker` in messages were renamed to `Runtime` to make it more clear
  what they refer to. Additionally, the `Error` message type has been changed
  to include additional fields (module and code) to make it easier to remap
  errors automatically.

  This makes it a BREAKING change for any existing runtimes.

* Provisioning of a runtime is now performed by a `Provisioner` which is an
  interface. Implementations exist for (sandboxed) ELF binaries and Intel SGX
  enclaves. The implementations have been refactored to be more composable,
  so for example the SGX implementation only implements the SGX-related bits
  but uses the regular provisioner otherwise.

* Configuration options for hosted runtimes have changed so existing configs
  will need to be _updated_ as follows:

    * The `--worker.runtime.backend` option has been renamed to
      `--worker.runtime.provisioner`.

    * The `--worker.runtime.loader` option has been renamed to
      `--worker.runtime.sgx.loader` and is now only required for supporting
      SGX runtimes. Non-SGX runtimes no longer need a loader.

    * The `--worker.runtime.binary` option has been renamed to
      `--worker.runtime.paths` and the value format has changed to be either
      a YAML map or a set of comma-separated key-value pairs separated by
      `=` (e.g., `<runtime-ID>=/path/to/binary`).

* The key manager worker has been slightly changed to use the common runtime
  provisioning code. The following configuration options have been _removed_:

    * `--worker.keymanager.tee_hardware` as the TEE hardware is inferred
      from the runtime descriptor.

    * `--worker.keymanager.runtime.loader` and
      `--worker.keymanager.runtime.binary` as the common options mentioned
      above should be used instead.


go/runtime/host/protocol/connection.go
```diff
@@ -0,0 +1,461 @@
...
+	c.state = state
+}
+
+// Implements Connection.
+func (c *connection) Close() {
+	if c.getState() != stateReady {
+		return
+	}
+
+	if err := c.conn.Close(); err != nil {
```
Yawning, line 121: Isn't this race condition prone?  `Close` and `Call` both dereference `c.conn`, and the critical section is too small.

go/runtime/host/protocol/types.go
```diff
@@ -53,27 +55,27 @@ type Body struct {
...
-	WorkerRPCCallResponse                *WorkerRPCCallResponse                `json:",omitempty"`
-	WorkerLocalRPCCallRequest            *WorkerLocalRPCCallRequest            `json:",omitempty"`
-	WorkerLocalRPCCallResponse           *WorkerLocalRPCCallResponse           `json:",omitempty"`
-	WorkerCheckTxBatchRequest            *WorkerCheckTxBatchRequest            `json:",omitempty"`
-	WorkerCheckTxBatchResponse           *WorkerCheckTxBatchResponse           `json:",omitempty"`
-	WorkerExecuteTxBatchRequest          *WorkerExecuteTxBatchRequest          `json:",omitempty"`
-	WorkerExecuteTxBatchResponse         *WorkerExecuteTxBatchResponse         `json:",omitempty"`
-	WorkerAbortRequest                   *Empty                                `json:",omitempty"`
-	WorkerAbortResponse                  *Empty                                `json:",omitempty"`
+	// Runtime interface.
```
Yawning, line 58: Not sure about how this will work out on the rust side, but would this be cleaner if things were grouped in a `struct Runtime`?

go/runtime/host/sandbox/bwrap.go
```diff
@@ -0,0 +1,154 @@
...
+
+	bwrapPath = "/usr/bin/bwrap"
+)
+
+type bwrap struct {
+	*naked
+}
+
+// NewBubbleWrap creates a Bubblewrap-based sandbox.
+func NewBubbleWrap(cfg Config) (Process, error) {
```
Yawning, line 26: Something I found myself wishing for was a way to easily mount byte buffers as files in the container (`--ro-bind-data FD DEST`).

go/runtime/host/host.go
```diff
@@ -0,0 +1,98 @@
...
+	"context"
+
+	"github.com/oasislabs/oasis-core/go/common"
+	"github.com/oasislabs/oasis-core/go/common/node"
+	"github.com/oasislabs/oasis-core/go/common/pubsub"
+	"github.com/oasislabs/oasis-core/go/common/version"
+	"github.com/oasislabs/oasis-core/go/runtime/host/protocol"
+)
+
+// Config contains common configuration for the provisioned runtime.
```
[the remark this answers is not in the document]
kostko: Good call, so one option would be to add a provisioner-specific `Extra interface{}` field.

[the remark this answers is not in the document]
Yawning: That works for me.

go/runtime/host/sandbox/bwrap.go
kostko: Added the `BindData` option.

go/runtime/host/host.go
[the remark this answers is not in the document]
kostko: Added.

go/runtime/host/sandbox/bwrap_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	t.Run("BindData", testBwrapBindData)
+}
+
+func testBwrapBindData(t *testing.T) {
+	require := require.New(t)
+
+	// Bind some data from a buffer and run /bin/cat inside the sandbox to make sure it's there.
+	var stdout bytes.Buffer
+	p, err := NewBubbleWrap(Config{
```
ptrus, line 26: this is mostly duplicated with the `naked_test` should this test instead just run the naked test with bwrap sandbox?

go/runtime/host/sandbox/bwrap_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"io"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestBubbleWrapSandbox(t *testing.T) {
+	// Skip test if there is no bwrap binary available.
+	if _, err := os.Stat(bwrapPath); os.IsNotExist(err) {
```
ptrus, line 14: We might want to instead use `+build` tags and handle this check elsewhere (e.g. a check in the makefile) - and on CI always build with the tag - that way we could enforce that this test will always be run on CI and it wouldn't get skipped without us noticing (e.g. someone not installing bwrap on a CI agent in future and missing this). 

Similar for other places (e.g. a `test-sgx` build tag instead of relying on env vars set in the sgx_tests)

go/runtime/host/sandbox/provisioner_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/oasislabs/oasis-core/go/runtime/host"
+	"github.com/oasislabs/oasis-core/go/runtime/host/tests"
+)
+
+var envRuntimePath = os.Getenv("OASIS_TEST_RUNTIME_HOST_RUNTIME_PATH")
+
+func skipIfMissingDeps(t *testing.T) {
+	// Skip test if there is no bwrap binary available.
+	if _, err := os.Stat(bwrapPath); os.IsNotExist(err) {
```
ptrus, line 15: shouldn't the `"Naked"` case be run even if there's no `bwrap`?

go/runtime/host/sandbox/sandbox.go
```diff
@@ -0,0 +1,60 @@
+// Package sandbox implements a process sandboxing mechanism.
```
ptrus, line 1: It might be nicer moving `Process` (and it's implementations naked & bwrap) into `go/runtime/host/sandbox/process` package

kostko: Good idea.

go/runtime/host/sandbox/bwrap_test.go
kostko: I just changed it so that bwrap must always be available to run these tests.

go/runtime/host/sandbox/bwrap_test.go
kostko: Done.

PR #2911, opened 2020-05-14
staking: Emit events when disbursing fees and rewards

### LINKED ISSUE #2909: Emit staking events when disbursing fees/rewards
Currently the staking service does not emit events when performing operations like disbursing fees and rewards. This should be fixed to emit proper events.

abukosek:
Fixes #2909.

Staking events are now generated when disbursing fees and rewards.
There are two new special account IDs -- `CommonPoolAccountID` and
`FeeAccumulatorAccountID` (both defined in `go/staking/api/api.go`),
which are used only in events to signify the common pool and the fee
accumulator respectively.
These account IDs are invalid by design to prevent misusing them
anywhere else.

### DIFF
--- a/go/common/crypto/signature/signature.go
+++ b/go/common/crypto/signature/signature.go
@@ -666,3 +666,22 @@ func BuildPublicKeyBlacklist(allowTestKeys bool) {
 		blacklistedPublicKeys.Store(pk, true)
 	}
 }
+
+// NewBlacklistedKey returns the PublicKey from the given hex or panics.
+// The given key is also added to the blacklist.
+func NewBlacklistedKey(hex string) PublicKey {
+	var pk PublicKey
+	if err := pk.UnmarshalHex(hex); err != nil {
+		panic(err)
+	}
+
+	// Make sure that the key doesn't already exist in the blacklist.
+	if pk.isBlacklisted() {
+		panic("key already exists in blacklist, use another")
+	}
+
+	// Blacklist key.
+	blacklistedPublicKeys.Store(pk, true)
+
+	return pk
+}
--- a/go/consensus/tendermint/apps/staking/state/state.go
+++ b/go/consensus/tendermint/apps/staking/state/state.go
@@ -22,6 +22,9 @@ import (
 var (
 	// AppName is the ABCI application name.
 	AppName = "100_staking"
+	// KeyAddEscrow is an ABCI event attribute key for AddEscrow calls
+	// (value is an api.AddEscrowEvent).
+	KeyAddEscrow = []byte("add_escrow")
 	// KeyTakeEscrow is an ABCI event attribute key for TakeEscrow calls
 	// (value is an app.TakeEscrowEvent).
 	KeyTakeEscrow = []byte("take_escrow")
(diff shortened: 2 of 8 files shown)


go/consensus/tendermint/apps/staking/state/state.go
```diff
@@ -782,6 +782,7 @@ func (s *MutableState) AddRewards(
 			if err = quantity.Move(&ent.Escrow.Active.Balance, commonPool, q); err != nil {
 				return errors.Wrap(err, "transferring to active escrow balance from common pool")
 			}
+			// XXX: Should we emit an event here?
```
kostko, line 785: Yes, this is reward disbursement.

go/consensus/tendermint/apps/staking/state/state.go
```diff
@@ -897,6 +898,7 @@ func (s *MutableState) AddRewardSingleAttenuated(
 		if err = quantity.Move(&ent.Escrow.Active.Balance, commonPool, q); err != nil {
 			return errors.Wrap(err, "transferring to active escrow balance from common pool")
 		}
+		// XXX: Should we emit an event here?
```
kostko, line 901: Yes, this is reward disbursement.

go/common/crypto/signature/signature.go
```diff
@@ -666,3 +666,12 @@ func BuildPublicKeyBlacklist(allowTestKeys bool) {
 		blacklistedPublicKeys.Store(pk, true)
 	}
 }
+
+// NewBlacklistedKey returns the PublicKey from the given hex or panics.
+func NewBlacklistedKey(hex string) PublicKey {
```
kostko, line 671: Based on the name this should actually blacklist the key.

abukosek: Thanks, fixed!

go/staking/tests/tester.go
```diff
@@ -179,8 +179,14 @@ func testTransfer(t *testing.T, state *stakingTestsState, backend api.Backend, c
 	err = consensusAPI.SignAndSubmitTx(context.Background(), consensus, srcSigner, tx)
 	require.NoError(err, "Transfer")
 
+TransferWaitLoop:
 	select {
 	case ev := <-ch:
+		if ev.From.Equal(api.CommonPoolAccountID) || ev.To.Equal(api.CommonPoolAccountID) || ev.From.Equal(api.FeeAccumulatorAccountID) || ev.To.Equal(api.FeeAccumulatorAccountID) {
+			// This is not the transfer we're looking for.
+			goto TransferWaitLoop
```
kostko, line 187: Use a `for` and continue? Also could you instead test that these fee-related events are there instead of just skipping them? Same below.

go/common/crypto/signature/signature.go
```diff
@@ -666,3 +666,14 @@ func BuildPublicKeyBlacklist(allowTestKeys bool) {
...
 }
+
+// NewBlacklistedKey returns the PublicKey from the given hex or panics.
+// The given key is also added to the blacklist.
+func NewBlacklistedKey(hex string) PublicKey {
+	var pk PublicKey
+	if err := pk.UnmarshalHex(hex); err != nil {
+		panic(err)
+	}
+	blacklistedPublicKeys.Store(pk, true)
```
kostko, line 677: @Yawning suggested to panic in case the key already exists -- to catch ID reuse. This is a good idea, also in this case you probably need to change the IDs to use something unique as those small order points are already blacklisted.

abukosek: Good idea!  I've implemented it and changed the IDs.  The new IDs have a prefix of `1abe11ed` (for "labelled") -- **c**ommon p**ool** continues with `c00l` and **fee** **acc**umulator continues with `feeacc` :)

go/staking/tests/tester.go
abukosek: Done!

Issue #2581, opened 2020-01-22
Version signed (entity/node/runtime) descriptors


kostko:
All registry descriptors that are signed should be explicitly versioned so that we can have code that supports multiple versions of the descriptors as we can't do migrations on signed data without invalidating the signatures.

The most recent change that breaks node descriptors is #2514 which changes the node role bitmask so previous node descriptors are now invalid. This makes it impossible to upgrade from 20.1 to 20.2 without regenerating all descriptors.

PR #2914, opened 2020-05-15
go/registry: Avoid storing full TLS certificates

### LINKED ISSUE #2556: registry: avoid storing full tls certificates
Currently we store whole certificates in registry, during #2475 an idea came up that we could potentially only store certificate hashes:

> ...But I guess the ony way to avoid it would be to completely override how TLS authentication is performed so that only certificate hashes would be compared ...

Taken from: https://github.com/oasislabs/oasis-core/pull/2475#discussion_r360342043

Should probably at least evaluate the approach.

kostko:
Fixes #2556 

Previously the node registry descriptor contained full TLS certificates for
talking with nodes via gRPC. This changes it so that only TLS public keys are
used when verifying peer certificates for TLS authentication.

This makes the registry descriptors smaller and also makes it easier to pass
around TLS identities (as public keys are much shorter).

Obviously, this change BREAKS the consensus protocol and all previously
signed node descriptors.

The following configuration changes are needed due to this change:

- In `oasis-node registry node` CLI, the `--node.committee_address` option
  has been renamed to `--node.tls_address` and the format has changed from
  `<certificate>@ip:port` to `<pubkey>@ip:port`.

- For configuring sentry nodes on the workers, the
  `--worker.sentry.cert_file` has been _removed_. Instead, the
  `--worker.sentry.address` now takes the same address format as specified
  above (`<pubkey>@ip:port`).

Previously signed node descriptors (v0) are considered valid at genesis time
iff the node is exclusively a validator node as indicated by the role bits.
Other nodes will need to be removed from genesis.

TODO
* [x] Blocked by #2581.
* [X] Limited support for old node descriptors at genesis.
* [x] Tests with old node descriptors at genesis.
* [x] Document what configuration changes are required.


docs/oasis-node/metrics.md
```diff
@@ -59,9 +59,9 @@ oasis_codec_size | Summary | CBOR codec message size (bytes). | [common/cbor](..
 oasis_consensus_proposed_blocks | Counter | Number of blocks proposed by the node. | [consensus/metrics](../../go/consensus/metrics/metrics.go#L17)
 oasis_consensus_signed_blocks | Counter | Number of blocks signed by the node. | [consensus/metrics](../../go/consensus/metrics/metrics.go#L10)
 oasis_finalized_rounds | Counter | Number of finalized rounds. | [roothash](../../go/roothash/metrics.go#L15)
-oasis_grpc_calls | Counter | Number of gRPC calls. | [common/grpc](../../go/common/grpc/grpc.go#L46)
-oasis_grpc_latency | Summary | gRPC call latency (seconds). | [common/grpc](../../go/common/grpc/grpc.go#L53)
-oasis_grpc_stream_writes | Counter | Number of gRPC stream writes. | [common/grpc](../../go/common/grpc/grpc.go#L60)
+oasis_grpc_calls | Counter | Number of gRPC calls. | [common/grpc](../../go/common/grpc/grpc.go#L48)
```
Yawning, line 62: Oh god.  This is going to be a perpetual source of churn isn't it.

PR #2915, opened 2020-05-15
{consensus,control}: Add GetStatus method to API

### LINKED ISSUE #2902: Add Status method to consensus backend API
We should add a `Status` method to the consensus client backend API which returns things like:

* Latest block height, hash and timestamp.
* Genesis block height and hash.
* List of peers (backend-dependent).
* Other things that I am forgetting.

abukosek:
Closes #2902.

* consensus: Add GetStatus method to API

    A new `GetStatus` method has been added to the consensus API.
    It returns useful information about the latest block, the genesis
    block, and the node itself.

* control: Add GetStatus method to API

    A new `GetStatus` method has been added to the control API.
    It returns the software version and the status of the consensus layer.

### DIFF
--- a/go/consensus/api/api.go
+++ b/go/consensus/api/api.go
@@ -83,6 +83,9 @@ type ClientBackend interface {
 
 	// GetGenesisDocument returns the original genesis document.
 	GetGenesisDocument(ctx context.Context) (*genesis.Document, error)
+
+	// GetStatus returns the current status overview.
+	GetStatus(ctx context.Context) (*Status, error)
 }
 
 // Block is a consensus block.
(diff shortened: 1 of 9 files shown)


go/consensus/tendermint/tendermint.go
```diff
@@ -699,6 +700,43 @@ func (t *tendermintService) GetTransactions(ctx context.Context, height int64) (
...
+	}
+
+	peers := make([]string, len(addrs))
+	for _, addr := range addrs {
+		peers = append(peers, addr.String())
+	}
+
+	return &consensusAPI.Status{
+		NodeVersion:   version.SoftwareVersion,
+		NodePeers:     peers,
```
Yawning, line 731: `NodePeers` is an odd name for "list of addresses the consensus backend is listening on.

go/consensus/tendermint/tendermint.go
```diff
@@ -699,6 +700,43 @@ func (t *tendermintService) GetTransactions(ctx context.Context, height int64) (
...
+	if err != nil {
+		return nil, err
+	}
+
+	addrs, err := t.GetAddresses()
+	if err != nil {
+		return nil, err
+	}
+
+	peers := make([]string, len(addrs))
```
Yawning, line 724: This creates a slice of strings, pre-populated with `len(addrs)` empty strings, that you are then appending to.

go/consensus/tendermint/tendermint.go
abukosek: Yeah, that is just a placeholder for testing until I implement the actual thing :)

go/consensus/tendermint/tendermint.go
abukosek: Oops, I'll fix that, thanks!

go/consensus/api/api.go
```diff
@@ -100,6 +103,19 @@ type Block struct {
 	Meta cbor.RawMessage `json:"meta"`
 }
 
+// Status is the current status overview.
+type Status struct {
+	NodeVersion string   `json:"node_version"`
```
kostko, line 108: What has node version to do with a consensus backend? Maybe there should be a separate `GetStatus` API on the node's control service (see `go/control/api`) which returns something like this (including the `GetStatus` response for the consensus backend among other things):
```golang
type Status struct {
    SoftwareVersion string `json:"software_version"`
    // ... maybe some other versions

    Consensus consensus.Status `json:"consensus"`
    // ...
}
```
Consensus could then expose fields related to it like the consensus protocol version.

abukosek: Good idea, I'll add that.

go/consensus/api/api.go
```diff
@@ -100,6 +103,27 @@ type Block struct {
 	Meta cbor.RawMessage `json:"meta"`
 }
 
+// Status is the current status overview.
+type Status struct {
+	// ConsensusVersion is the version of the consensus protocol that the node is using.
+	ConsensusVersion string `json:"consensus_version"`
```
kostko, line 109: Maybe also include a `Backend` field that specifies the consensus backend identifier (as a string).

abukosek: Done! :)

Issue #2517, opened 2020-01-06
Tests for key manager upgrades


kostko:
Depends on #2516.

We should add E2E tests for key manager runtime upgrades (without production keys).

The process of upgrading the KM is the following:
* First make sure an initial key manager is registered and operational.
* Build an upgraded key manager runtime enclave (anything which changes MRENCLAVE should be fine).
* Update the key manager policy to include the MRENCLAVE for the new runtime and wait for the policy to propagate.
* Leave old key manager nodes running.
* Spin up new key manager nodes with upgraded enclave.
* Wait for the key manager nodes to replicate the master secret.
* Verify that the key manager nodes work.
* Take down old key manager nodes.
* Verify that the key manager nodes work.

Issue #2130, opened 2019-09-19
Add IsReady/WaitReady to client protocol


kostko:
We should add `IsReady` and `WaitReady` RPC methods to the client protocol that could be used to check whether the node has finished initializing and is ready to process requests.

Currently there already exist `IsSynced` and `WaitSync` which check for consensus sync, but these new methods would be useful for the worker nodes to additionally check that the hosted runtimes have also started successfully.

A node would be ready iff everything of the following is true:
* Consensus has finished syncing.
* All configured runtimes are registered.
* All configured workers have finished initializing.
* All hosted runtime processes are ready to process requests.

There should also be a corresponding `debug client wait-ready` command.

ptrus:
Just posting two more code references that can be done with this issue:

https://github.com/oasislabs/oasis-core/blob/3d5e7dc02d5150fe0d5dcb54aa773984bf0a3872/go/oasis-test-runner/scenario/e2e/runtime.go#L556-L559

https://github.com/oasislabs/oasis-core/blob/3d5e7dc02d5150fe0d5dcb54aa773984bf0a3872/go/worker/keymanager/worker.go#L186-L189

PR #2920, opened 2020-05-18
KM upgrade E2E test

ptrus:
Closes: #2517 

TODO:
- [x] instead of using the `OASIS_UNSAFE_KM_POLICY_THRESHOLD` build-time-flag to ensure different binaries, just duplicate the test-keymanager crate with a change, it will make the code simpler
- [x] will be done in a separate PR. Last part of the test (running the client again with the new km), is not working yet, requires: https://github.com/oasislabs/oasis-core/issues/2919 (potentially do it as part of this PR)
  - will be handled in a separate PR
- [x] address the remaining todo: https://github.com/oasislabs/oasis-core/pull/2920/files#diff-59b3c0087bae96951be3b7509d5cd9b7R560-R563

### DIFF
--- a/.changelog/2517.bugfix.md
+++ b/.changelog/2517.bugfix.md
@@ -1,0 +1,6 @@
+go/worker/keymanager: retry initialization in case of failure
+
+The keymanager worker registers only after the initialization either fails or
+succeeds. In case the worker needs to replicate the first initialization will
+always fail, since other nodes' access control prevents it from replicating.
+In that case the initialization should be retried.
--- a/go/oasis-test-runner/oasis/oasis.go
+++ b/go/oasis-test-runner/oasis/oasis.go
@@ -68,6 +68,7 @@ type Node struct { // nolint: maligned
 	termErrorOk bool
 	doStartNode func() error
 	isStopping  bool
+	noAutoStart bool
 
 	disableDefaultLogWatcherHandlerFactories bool
 	logWatcherHandlerFactories               []log.WatcherHandlerFactory
@@ -166,6 +167,8 @@ func (n *Node) handleExit(cmdErr error) error {
 type NodeCfg struct { // nolint: maligned
 	AllowEarlyTermination bool
 	AllowErrorTermination bool
+
+	NoAutoStart bool
 
 	DisableDefaultLogWatcherHandlerFactories bool
 	LogWatcherHandlerFactories               []log.WatcherHandlerFactory
@@ -289,7 +292,7 @@ func (net *Network) Runtimes() []*Runtime {
 	return net.runtimes
 }
 
-// Keymanager returns the keymanagers associated with the network.
+// Keymanagers returns the keymanagers associated with the network.
 func (net *Network) Keymanagers() []*Keymanager {
 	return net.keymanagers
 }
@@ -379,7 +382,7 @@ func (net *Network) AddLogWatcher(node *Node) error {
 	return nil
 }
 
-// CloseLogWatchers closes all log watchers and checks if any errors were reported
+// CheckLogWatchers closes all log watchers and checks if any errors were reported
 // while the log watchers were running.
 func (net *Network) CheckLogWatchers() (err error) {
 	for _, w := range net.logWatchers {
@@ -396,7 +399,7 @@ func (net *Network) CheckLogWatchers() (err error) {
 }
 
 // Start starts the network.
-func (net *Network) Start() error {
+func (net *Network) Start() error { // nolint: gocyclo
 	net.logger.Info("starting network")
 
 	// Figure out if the IAS proxy is needed by peeking at all the
@@ -467,16 +470,22 @@ func (net *Network) Start() error {
 		}
 	}
 
-	net.logger.Debug("starting seed node")
-	if err = net.seedNode.startNode(); err != nil {
-		net.logger.Error("failed to start seed node",
-			"err", err,
-		)
-		return err
+	if !net.seedNode.noAutoStart {
+		net.logger.Debug("starting seed node")
+		if err = net.seedNode.startNode(); err != nil {
+			net.logger.Error("failed to start seed node",
+				"err", err,
+			)
+			return err
+		}
 	}
 
 	net.logger.Debug("starting validator node(s)")
 	for _, v := range net.validators {
+		if v.noAutoStart {
+			continue
+		}
+
 		if err = v.startNode(); err != nil {
 			net.logger.Error("failed to start validator",
 				"err", err,
@@ -494,8 +503,12 @@ func (net *Network) Start() error {
 		time.Sleep(validatorStartDelay)
 	}
 
+	net.logger.Debug("starting keymanager(s)")
 	for _, km := range net.keymanagers {
-		net.logger.Debug("starting keymanager")
+		if km.noAutoStart {
+			continue
+		}
+
 		if err = km.startNode(); err != nil {
 			net.logger.Error("failed to start keymanager node",
 				"err", err,
@@ -506,6 +519,10 @@ func (net *Network) Start() error {
 
 	net.logger.Debug("starting storage node(s)")
 	for _, v := range net.storageWorkers {
+		if v.noAutoStart {
+			continue
+		}
+
 		if err = v.startNode(); err != nil {
 			net.logger.Error("failed to start storage worker",
 				"err", err,
@@ -516,6 +533,10 @@ func (net *Network) Start() error {
 
 	net.logger.Debug("starting compute node(s)")
 	for _, v := range net.computeWorkers {
+		if v.noAutoStart {
+			continue
+		}
+
 		if err = v.startNode(); err != nil {
 			net.logger.Error("failed to start compute worker",
 				"err", err,
@@ -526,16 +547,25 @@ func (net *Network) Start() error {
 
 	net.logger.Debug("starting sentry node(s)")
 	for _, v := range net.sentries {
+		if v.noAutoStart {
+			continue
+		}
+
 		if err = v.startNode(); err != nil {
 			net.logger.Error("failed to start sentry node",
 				"err", err,
 			)
 			return err
 		}
+
 	}
 
 	net.logger.Debug("starting client node(s)")
 	for _, v := range net.clients {
+		if v.noAutoStart {
+			continue
+		}
+
 		if err = v.startNode(); err != nil {
 			net.logger.Error("failed to start client node",
 				"err", err,
@@ -546,6 +576,10 @@ func (net *Network) Start() error {
 
 	net.logger.Debug("starting byzantine node(s)")
 	for _, v := range net.byzantine {
+		if v.noAutoStart {
+			continue
+		}
+
 		if err = v.startNode(); err != nil {
 			net.logger.Error("failed to start byzantine node",
 				"err", err,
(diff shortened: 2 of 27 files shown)


go/oasis-test-runner/scenario/e2e/runtime.go
```diff
@@ -557,6 +557,12 @@ func (sc *runtimeImpl) initialEpochTransitions() error {
 			return fmt.Errorf("failed to wait for nodes: %w", err)
 		}
 
+		// Temporary....
+		// TODO: with the change that keymanager initially registers without
+		// status, before it is initialized, we should wait here that keymanager
+		// is initialized. Add a method to wait for KM initialization?
+		time.Sleep(10 * time.Second)
```
ptrus, line 564: If the "register as a keymanager on the first status update" fix is ok (see changelog fragment), i'll fix this properly by waiting for KM initialization, but didn't want to implement that before the fix is acked, as it might not be needed if we decide for a different fix (e.g. maybe just retrying the initialization)

go/worker/keymanager/worker.go
```diff
@@ -175,21 +175,18 @@ func (w *Worker) callLocal(ctx context.Context, data []byte) ([]byte, error) {
...
-				rt := n.AddOrUpdateRuntime(w.runtime.ID())
-				rt.Version = startedEvent.Version
-				rt.ExtraInfo = nil
-				rt.Capabilities.TEE = startedEvent.CapabilityTEE
-				return nil
-			})
-		}
-	}()
+	if !w.initialSyncDone {
+		// Send an initial node registration without any extra info, so that
```
kostko, line 179: Can you explain a bit why this helps? So previously this would be called after init fails and now it is called immediately. But `SetAvailable` does not block until everyone has updated their access control lists so I'm not sure why calling it earlier would make a difference? If there is an observed difference it could be a timing quirk -- e.g., that this works because there is enough delay in trying to initialize the key manager, but it doesn't seem robust?

If this is actually problematic, wouldn't it be better that if `updateStatus` fails, but both "status" and "startedEvent" are available, the key manager worker loop would create a ticker that would retry key manager initialization? Of course you could have both, the change you implemented here (to release registration earlier) and also retry in case there are no policy updates, but initializing the key manager at a slightly later time would work due to propagation delay?

Otherwise could this cause the key manager to never initialize due to unfortunate timing and no new status updates?

go/oasis-test-runner/scenario/e2e/runtime.go
kostko: Wouldn't something be needed in either case with the "key manager always registers"? E.g., even if you retry, the retry could happen later than the key manager actually registering with correct `ExtraInfo`? So maybe we should do #2130 and then you could call `WaitReady` on the key manager node(s) to check if they are really ready?

go/worker/keymanager/worker.go
ptrus: > But SetAvailable does not block until everyone has updated their access control lists so I'm not sure why calling it earlier would make a difference? If there is an observed difference it could be a timing quirk -- e.g., that this works because there is enough delay in trying to initialize the key manager, but it doesn't seem robust?

iirc the runtime does have some retry logic so this did always work, although I agree that in practice it could still fail.

> key manager worker loop would create a ticker that would retry key manager initialization

yeah i think this is a more proper solution, the keymanager should retry initialization in case `status`  and `startedEvent` are available and the initialization has not yet succeed.

> Of course you could have both

I think i will revert the current change just so I don't have to deal with fixing the initialization epochs in the e2e tets where mock epochtime is used

go/oasis-test-runner/scenario/e2e/runtime.go
ptrus: Yeah would be, but if i revert the "km always registers" and only replace it it with retrying the initialization no changes are needed here - so i think i'll do just that.

ptrus: Once we have #2130 we should change the code here to wait for ready, and we can also bring back the "km always registers" fix, since i think is generally a bit nicer than the current "hack" to only register after initial failure :-)

kostko: Maybe mention this in #2130 (and add a comment in the code) so that we don't forget about it?

ptrus: done, i'll mention on the issue once this is merged so i can permalink to code

go/worker/keymanager/worker.go
ptrus: done

tests/runtimes/simple-keymanager-upgrade/src/api.rs
```diff
@@ -0,0 +1,26 @@
...
+                .iter()
+                {
+                    let private_key = OasisPrivateKey::from_test_seed(seed.to_string());
+                    set.insert(private_key.public_key());
+                }
+            }
+
+            set
+        },
+        threshold: 9002,
```
Yawning, line 24: At some point, we probably should get rid of the hack that sets the threshold to 2 when using test keys.  Also either leave a comment here, or pick a different way to force the MRENCLAVE to change.

go/worker/keymanager/worker.go
```diff
@@ -287,6 +302,11 @@ func (w *Worker) updateStatus(status *api.Status, startedEvent *host.StartedEven
 	w.logger.Info("Key manager initialized",
 		"checksum", hex.EncodeToString(signedInitResp.InitResponse.Checksum),
 	)
+	if w.initTicker != nil {
+		w.initTickerCh = make(chan time.Time)
```
kostko, line 306: Why not set the channel to `nil`, it should always block then (e.g., select case should never trigger).

go/worker/keymanager/worker.go
```diff
@@ -459,12 +479,17 @@ func (w *Worker) worker() { // nolint: gocyclo
...
 			// Forward status update to key manager runtime.
-			if err := w.updateStatus(currentStatus, currentStartedEvent); err != nil {
+			if err = w.updateStatus(currentStatus, currentStartedEvent); err != nil {
 				w.logger.Error("failed to handle status update",
 					"err", err,
 				)
 				continue
 			}
+		case <-w.initTickerCh:
+			if err = w.updateStatus(currentStatus, currentStartedEvent); err != nil {
```
kostko, line 489: Maybe check that `currentStatus` and `currentStartedEvent` are still available?

go/worker/keymanager/worker.go
ptrus: :+1: also fixed the init code to initialize it to a nil channel.

go/worker/keymanager/worker.go
ptrus: good idea, done :+1:

.changelog/2517.bugfix.md
```diff
@@ -0,0 +1,6 @@
+go/worker/keymanager: retry initialization in case of failure
+
+The keymanager worker registers only after the initialization either fails or
+succeeds. In case the worker needs to replicate the first initialization will
+always fail, since other nodes' access control prevent it from replicating. In
```
kostko, line 5:
```suggestion
always fail, since other nodes' access control prevents it from replicating. In
```

go/oasis-test-runner/oasis/oasis.go
```diff
@@ -467,109 +470,132 @@ func (net *Network) Start() error {
...
-		//
-		// If you don't attempt to start the Tendermint Prometheus HTTP server
-		// (even if it is doomed to fail due to node already listening on the
-		// port), and you launch all the validators near simultaniously, there
-		// is a high chance that at least one of the validators will get upset
-		// and start refusing connections.
-		time.Sleep(validatorStartDelay)
+	net.logger.Debug("starting validator node(s)")
+	for _, v := range net.validators {
+		if !v.noAutoStart {
```
kostko, line 485: Could avoid the additional indentation by doing it like (also elsewhere):
```golang
if v.noAutoStart {
    continue
}
```

go/oasis-test-runner/scenario/e2e/keymanager_upgrade.go
```diff
@@ -0,0 +1,313 @@
...
+	"github.com/oasislabs/oasis-core/go/oasis-test-runner/scenario"
+	registry "github.com/oasislabs/oasis-core/go/registry/api"
+)
+
+// KeymanagerUpgrade is the keymanager upgrade scenario.
+var KeymanagerUpgrade scenario.Scenario = newKmUpgradeImpl()
+
+type kmUpgradeImpl struct {
+	runtimeImpl
+	nonce uint64
```
kostko, line 26: Blank line above.

go/oasis-test-runner/scenario/e2e/keymanager_upgrade.go
```diff
@@ -0,0 +1,313 @@
...
+	// Sanity check fixture.
+	if err := func() error {
+		if oldKMRuntime.Kind() != registry.KindKeyManager {
+			return fmt.Errorf("old keymanager runtime not KeyManagerKind")
+		}
+		if newKMRuntime.Kind() != registry.KindKeyManager {
+			return fmt.Errorf("new keymanager runtime not KeyManagerKind")
+		}
+		if oldKMRuntime.ID() != newKMRuntime.ID() {
+			return fmt.Errorf("keymanager runtimes ID missmatch")
```
kostko, line 94:
```suggestion
			return fmt.Errorf("keymanager runtimes ID mismatch")
```

go/oasis-test-runner/scenario/e2e/keymanager_upgrade.go
```diff
@@ -0,0 +1,313 @@
...
+	kmPolicySig2Path := filepath.Join(childEnv.Dir(), "km_policy_sig2.pem")
+	kmPolicySig3Path := filepath.Join(childEnv.Dir(), "km_policy_sig3.pem")
+	kmUpdateTxPath := filepath.Join(childEnv.Dir(), "km_gen_update.json")
+
+	oldKMRuntime := sc.net.Runtimes()[0]
+	newKMRuntime := sc.net.Runtimes()[2]
+	// Sanity check fixture.
+	if err := func() error {
+		if oldKMRuntime.Kind() != registry.KindKeyManager {
+			return fmt.Errorf("old keymanager runtime not KeyManagerKind")
```
kostko, line 88:
```suggestion
			return fmt.Errorf("old keymanager runtime not of kind KindKeyManager")
```

go/oasis-test-runner/scenario/e2e/runtime.go
```diff
@@ -553,6 +553,11 @@ func (sc *runtimeImpl) initialEpochTransitions() error {
 			"num_nodes", numNodes,
 		)
 
+		// TODO: once #2130 is done, aditionally wait for nodes to be Ready here.
+		// As generally, for example a keymanager can register (with missing
+		// extra info) before actually being initialized and triggering an
+		// epoch transition at that point would be to soon.
```
kostko, line 559:
```suggestion
		// epoch transition at that point would be too soon.
```

go/worker/keymanager/worker.go
```diff
@@ -459,12 +479,19 @@ func (w *Worker) worker() { // nolint: gocyclo
...
 			// Forward status update to key manager runtime.
-			if err := w.updateStatus(currentStatus, currentStartedEvent); err != nil {
+			if err = w.updateStatus(currentStatus, currentStartedEvent); err != nil {
 				w.logger.Error("failed to handle status update",
 					"err", err,
 				)
 				continue
 			}
+		case <-w.initTickerCh:
+			if currentStatus != nil && currentStartedEvent != nil {
```
kostko, line 489: Instead of indenting just continue and outdent the updateStatus below.

go/oasis-test-runner/oasis/oasis.go
```diff
@@ -546,30 +576,39 @@ func (net *Network) Start() error {
 
 	net.logger.Debug("starting byzantine node(s)")
 	for _, v := range net.byzantine {
-		if err = v.startNode(); err != nil {
-			net.logger.Error("failed to start byzantine node",
-				"err", err,
-			)
-			return err
+		if !v.noAutoStart {
```
kostko, line 579: Continue as above?

go/oasis-test-runner/oasis/oasis.go
```diff
@@ -546,30 +576,39 @@ func (net *Network) Start() error {
...
-	// Use the first validator as a controller.
-	if len(net.validators) >= 1 {
-		if net.controller, err = NewController(net.validators[0].SocketPath()); err != nil {
-			net.logger.Error("failed to create controller",
-				"err", err,
-			)
-			return fmt.Errorf("oasis: failed to create controller: %w", err)
+	// Use the first started validator as a controller.
+	for _, v := range net.validators {
+		if !v.noAutoStart {
```
kostko, line 591: Continue as above?

go/oasis-test-runner/oasis/oasis.go
```diff
@@ -546,30 +576,39 @@ func (net *Network) Start() error {
...
-	if len(net.clients) >= 1 {
-		if net.clientController, err = NewController(net.clients[0].SocketPath()); err != nil {
-			net.logger.Error("failed to create client controller",
-				"err", err,
-			)
-			return fmt.Errorf("oasis: failed to create client controller: %w", err)
+
+	// Create a client controller for the first started client node.
+	for _, v := range net.clients {
+		if !v.noAutoStart {
```
kostko, line 604: Continue as above?

ptrus:
thx for the reviews

PR #2921, opened 2020-05-19
go/oasis-node: Add the `debug dumpdb` command

### LINKED ISSUE #2359: oasis-node: Write a tool to extract a state dump from the state db
The system is designed to fail-safe and crash the node when certain conditions are encountered (eg: was unable to elect a validator set).

It should be possible to use the on-disk database to create a genesis dump, that can be used for the basis of a redeployment (with some manual intervention to fix the condition that caused the hard-failure in the first place).

Yawning:
Fixes #2359

 * [x] Add support for making storage backends read-only.
 * [x] Refactor timekeeping so that state queries can happen without an ApplicationState.
 * [x] Write the tool.
 * [x] Write test cases.

### DIFF
--- a/go/oasis-node/cmd/debug/dumpdb/dumpdb.go
+++ b/go/oasis-node/cmd/debug/dumpdb/dumpdb.go
@@ -1,0 +1,387 @@
+// Package dumpdb implements the dumpdb sub-command.
+package dumpdb
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
+	"github.com/spf13/viper"
+
+	beacon "github.com/oasislabs/oasis-core/go/beacon/api"
+	"github.com/oasislabs/oasis-core/go/common/logging"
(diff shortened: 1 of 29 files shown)


go/oasis-node/cmd/debug/dumpdb/dumpdb.go
```diff
@@ -0,0 +1,383 @@
...
+	// document without manual intervention, and only the state that
+	// would be exported by the normal dump process will be present
+	// in the dump.
+	qs := &dumpQueryState{
+		ldb:    ldb,
+		height: dumpVersion,
+	}
+	doc := &genesis.Document{
+		Height:    qs.BlockHeight(),
+		Time:      time.Now(), // XXX: Make this deterministic?
```
kostko, line 148: Making this deterministic would require looking into the Tendermint block store and fetching the corresponding block header. I had some code to do this stashed somewhere, I'll try to dig it up.

Yawning: To be honest since any documents generated by the tool will require manual intervention anyway, I'm ok with this just being the dump time for now.

kostko: That's fine.

PR #2922, opened 2020-05-19
metrics: Add caller label to oasis_codec_size

### LINKED ISSUE #1771: Metrics for inter-node communication
We should add Prometheus metrics tracking inter-node communication, for example:
* Rate and size of messages received via P2P.
* Rate of EnclaveRPC calls.

matevz:
Resolves #1771


go/common/cbor/codec.go
```diff
@@ -40,15 +37,16 @@ type MessageReader struct {
 }
 
 // Read deserializes a single CBOR-encoded Message from the underlying reader.
-func (c *MessageReader) Read(msg interface{}) error {
+func (c *MessageReader) Read(msg interface{}, caller string) error {
```
kostko, line 40: Uh I don't like this, wouldn't the caller always be the same for the lifetime of the MessageReader? If anything it should be a field in the MessageReader.

go/common/cbor/codec.go
```diff
@@ -21,16 +21,13 @@ var (
 			Name: "oasis_codec_size",
 			Help: "CBOR codec message size (bytes).",
 		},
-		[]string{"call"},
+		[]string{"call", "caller"},
```
kostko, line 24: Instead of calling it "caller" maybe say "module"?

go/common/cbor/codec.go
```diff
@@ -10,7 +10,18 @@ import (
 )
 
 // Maximum message size.
-const maxMessageSize = 104857600 // 100 MiB
+const (
+	maxMessageSize = 104857600 // 100 MiB
+
+	// ModuleP2P corresponds to a peer-to-peer message context.
+	ModuleP2P = "p2p"
```
kostko, line 17: These modules should not be defined in the codec package, they should be where the components use the codec.

go/common/cbor/codec.go
```diff
@@ -21,21 +32,25 @@ var (
...
 	}
 
-	labelRead  = prometheus.Labels{"call": "read"}
-	labelWrite = prometheus.Labels{"call": "write"}
-
 	metricsOnce sync.Once
 )
 
+// MessageWrapper is a wrapper for MessageReader or MessageWriter.
+type MessageWrapper struct {
```
kostko, line 46: All of this just for a single common string field? Consider just copying the field, it will simplify the code.

go/runtime/host/protocol/connection.go
```diff
@@ -409,7 +446,7 @@ func (c *connection) initConn(conn net.Conn) {
 	}
 
 	c.conn = conn
-	c.codec = cbor.NewMessageCodec(conn)
+	c.codec = cbor.NewMessageCodec(conn, cbor.ModuleRuntimeHost)
```
kostko, line 449: There is already this thing called `moduleName` here, use it?

go/runtime/host/protocol/connection.go
```diff
@@ -171,6 +192,22 @@ func (c *connection) Call(ctx context.Context, body *Body) (*Body, error) {
 }
 
 func (c *connection) call(ctx context.Context, body *Body) (*Body, error) {
+	start := time.Now()
+	defer func() {
+		if viper.GetString(metrics.CfgMetricsMode) != metrics.MetricsModeNone {
+			var call string
+			// Determine request/response type by taking the first non-empty field.
```
kostko, line 199: Could this part (extracting the name of the body request) be moved to the Body struct, e.g. a `String() string` method or something.

go/runtime/host/protocol/connection.go
```diff
@@ -5,23 +5,44 @@ import (
...
 )
 
 const moduleName = "rhp/internal"
 
 // ErrNotReady is the error reported when the Runtime Host Protocol is not initialized.
-var ErrNotReady = errors.New(moduleName, 1, "rhp: not ready")
+var (
+	ErrNotReady = errors.New(moduleName, 1, "rhp: not ready")
+
+	rhpLatency = prometheus.NewSummaryVec(
```
kostko, line 32: Could you also add counters for successes and failures as you can get that by just checking if there was an error.

go/runtime/host/protocol/connection.go
```diff
@@ -167,10 +203,25 @@ func (c *connection) Call(ctx context.Context, body *Body) (*Body, error) {
 		return nil, ErrNotReady
 	}
 
-	return c.call(ctx, body)
+	b, err := c.call(ctx, body)
+	if viper.GetString(metrics.CfgMetricsMode) != metrics.MetricsModeNone {
+		if err != nil {
```
kostko, line 208: This will miss calls that use the internal `call` directly. Instead of doing it here to catch the error you can use named return values like so:
```golang
func (c *connection) call(ctx context.Context, body *Body) (result *Body, err error) {
```

And in the defer handler you then have the `err` available so you can check its value (e.g., is it non-nil) and update the corresponding counter. This way you also avoid having the metrics-enabled check twice.

matevz: Good idea!

matevz: :+1:

PR #2925, opened 2020-05-20
genesis: Set initial height to 1

abukosek:
The genesis block should have an initial height of 1, since the value 0 is reserved for the latest block in queries (`go/consensus/api/api.go`, `HeightLatest`).


go/oasis-node/cmd/genesis/genesis.go
```diff
@@ -157,6 +157,7 @@ func doInitGenesis(cmd *cobra.Command, args []string) {
 		ChainID:   chainID,
 		Time:      time.Now(),
 		HaltEpoch: epochtime.EpochTime(viper.GetUint64(cfgHaltEpoch)),
+		Height:    1,
```
kostko, line 160: Note that this height is currently unused as the underlying consensus backend (Tendermint) does not yet support starting at heights different from 1.

abukosek: It's exported in the genesis document and the `GetStatus` method, which is used by external services (e.g. our Rosetta gateway) to determine the height of the genesis block, so having a mismatching value here breaks other things :)

abukosek:
Good idea, I'll add that.

abukosek:
Done, this is now ready for review :)

abukosek:
After a discussion with Yawning, I'm closing this in favor of returning a hardcoded value of `1` as the genesis height in `GetStatus`.

PR #2926, opened 2020-05-20
consensus: Fix GenesisHeight in GetStatus

abukosek:
This PR fixes the `GenesisHeight` in `GetStatus`, so that it always returns 1 as the genesis height.

### DIFF
--- a/go/consensus/tendermint/tendermint.go
+++ b/go/consensus/tendermint/tendermint.go
@@ -710,12 +710,12 @@ func (t *tendermintService) GetTransactions(ctx context.Context, height int64) (
 }
 
 func (t *tendermintService) GetStatus(ctx context.Context) (*consensusAPI.Status, error) {
-	genDoc, err := t.GetGenesisDocument(ctx)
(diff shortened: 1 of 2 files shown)


go/consensus/tendermint/tendermint.go
```diff
@@ -710,12 +710,10 @@ func (t *tendermintService) GetTransactions(ctx context.Context, height int64) (
 }
 
 func (t *tendermintService) GetStatus(ctx context.Context) (*consensusAPI.Status, error) {
-	genDoc, err := t.GetGenesisDocument(ctx)
-	if err != nil {
-		return nil, err
-	}
-
-	genBlk, err := t.GetBlock(ctx, genDoc.Height)
+	// Genesis block is hardcoded as block 1, since tendermint doesn't have
```
kostko, line 713: Note that this is only true until Tendermint starts supporting restoring from a non-zero height. Maybe mention this and reference https://github.com/tendermint/tendermint/issues/2543.

PR #2927, opened 2020-05-20
go: update dependencies depending on 'websocket@v1.4.0'

ptrus:
TODO: 
- [x] ~tendermint `v0.33.4` still relies on `websocket@v1.4.0`. Although we use a a fork (via replace directive), the tool doesn't seem to be aware of it so only updating the dependency in our fork might not be enough: https://github.com/oasislabs/tendermint/pull/7~ 
  - actually not needed since we need to use a `replace` directive here anyway


go/go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/oasislabs/oasis-core/go
 
 replace (
-	github.com/tendermint/tendermint => github.com/oasislabs/tendermint v0.33.4-oasis1
+	github.com/tendermint/tendermint => github.com/oasislabs/tendermint v0.33.4-ptrus // TODO: update tag.
```
kostko, line 4: Do we also need a replace directive for websocket here (since there is one in the fork) or are they transitive?

ptrus: Yeah seems that indeed it is needed, will add it. I tested this by creating an invalid replace directive in our fork: https://github.com/oasislabs/tendermint/commit/ecece7f01b43f28cf6a004f5e5b2a0fce908b437  (which causes the build to break). However the oasis-core/go build did still succeed when depending on that version.

Which probably means that any project that would depend on `oasis-core/go` should also propagate the replaces we use in `go.mod` (otherwise the code would actually use the non replaced dependencies??)

ptrus: also this means we can avoid updating the dependencies in the fork, which is probably preferable as it would only cause pain on the next update of the tendermint fork

ptrus:
Also opened a cleanup task for this: https://github.com/oasislabs/oasis-core/issues/2931

.nancy-ignore
```diff
@@ -0,0 +1,5 @@
+# Since nancy doesn't respect replace directive, this check will fail until
+# upstream tendermint depends on websocket@1.4.0. We force a fixed version of
+# the websocket in go.mod. Remove once upstream dependencies do not depend on
+# websocket@1.4.0 and replace directive is not needed anymore.
+5f259e63-3efb-4c47-b593-d175dca716b0
```
kostko, line 5: What's this magic UUID?

ptrus: https://ossindex.sonatype.org/vuln/5f259e63-3efb-4c47-b593-d175dca716b0

I'll add a comment with the link above it. Also tests passed and this is only change that i'll do, so i'll just merge it without waiting for the tests again

kostko: Yeah go go go :+1:

Issue #2928, opened 2020-05-20
Change Account ids to truncated hash of the public key


tjanez:
##### SUMMARY

Transition Account ids which are currently just Entities' public keys to a truncated hash of the corresponding public key.

Use the first 20 bytes of SHA512/256(public key).

##### ISSUE TYPE
- Feature Idea

##### ADDITIONAL INFORMATION

Practically all significant blockchain projects use some form of hash over the public key rather then the public key itself for an account's id/address:
- Bitcoin: RIPEMD160(SHA256(public key)
- Ethereum: last 20 bytes of Keccak256(public key)

See https://www.susanka.eu/coins-crypto/ for more examples.

PR #2929, opened 2020-05-21
go/common/keyformat: Add support for hashed key elements

kostko:
Add support for hashed key elements and hash some of the key elements.

Staking state is omitted as it will not be needed once we do #2928 (as all account identifiers will already be hashed then).

### DIFF
--- a/go/common/keyformat/hashed.go
+++ b/go/common/keyformat/hashed.go
@@ -1,0 +1,80 @@
+package keyformat
+
+import (
+	"encoding"
+	"fmt"
+
+	"github.com/oasislabs/oasis-core/go/common/crypto/hash"
+)
+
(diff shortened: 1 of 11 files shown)


go/common/keyformat/hashed.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func (h *hashedFormat) getData(v interface{}) ([]byte, error) {
+	switch t := v.(type) {
+	case encoding.BinaryMarshaler:
+		return t.MarshalBinary()
+	case []byte:
+		return t, nil
+	default:
+		panic(fmt.Sprintf("unsupported type: %T", t))
```
Yawning, line 45: If this can return an error, and the only caller returns an error, should this really be panicing?

PR #2930, opened 2020-05-21
Make (some) CBOR encodings more compact

kostko:
We could do this for more of our CBOR types to make encodings more compact, although this will break external tooling (khm Ledger khm) depending on specific encoding. But better now than later.


Yawning:
I'm ok with the idea, and if we're going to do it, the sooner the better.

PR #2937, opened 2020-05-22
oasis-test-runner: Use explicit flagsets for scenario parameters

### LINKED ISSUE #2897: Cannot set default values for E2E Scenarios Parameters
`Parameters()` methods always create new flagset instances, for example:
https://github.com/oasislabs/oasis-core/blob/35b7f62efec57e7fe638ad0aeee8e32962dbfc47/go/oasis-test-runner/scenario/e2e/e2e.go#L56-L61

That is problematic in conjunction with:
https://github.com/oasislabs/oasis-core/blob/35b7f62efec57e7fe638ad0aeee8e32962dbfc47/go/oasis-test-runner/cmd/root.go#L130-L138

IF a flag doesn't set existing parameter state as the default value, e.g.:
`fs.StringVar(&sc.nodeBinary, cfgNodeBinary, "<some-default>", "path to the node binary")`

The parameter will get overridden with the default value every time `Parameters()` is called. 

---

Suggested fix:
`Parameters()` methods creating new flagset instances is wrong, instead the flagset should be stored in the scenario struct and `Parameters()` should just return the stored flagset. However, `Clone()` should create a new instance of the flagset.

matevz:
Fixes #2897

### DIFF
--- a/go/oasis-test-runner/env/env.go
+++ b/go/oasis-test-runner/env/env.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os/exec"
 	"path/filepath"
+	"reflect"
 	"sync"
 	"syscall"
 	"time"
@@ -25,6 +26,9 @@ type CleanupFn func()
 // ParameterFlagSet is a wrapper for flag.FlagSet to produce nicer JSON output.
 type ParameterFlagSet struct {
 	flag.FlagSet
+
+	name          string
+	errorHandling flag.ErrorHandling
 }
 
 // TestInstanceInfo contains information of the current test run.
@@ -50,6 +54,34 @@ func (pfs *ParameterFlagSet) MarshalJSON() ([]byte, error) {
 	})
 
 	return json.Marshal(ps)
+}
+
+// Clone clones the parameter flagset.
+//
+// This function is usually called when cloning scenarios where we also want to clone existing
+// parameter values.
+func (pfs *ParameterFlagSet) Clone() *ParameterFlagSet {
+	newPfs := NewParameterFlagSet(pfs.name, pfs.errorHandling)
+	pfs.VisitAll(func(f *flag.Flag) {
+		// Clone concrete members of Flag.
+		fl := *f
+		// Clone Value interface.
+		fl.Value = reflect.New(reflect.TypeOf(fl.Value).Elem()).Interface().(flag.Value)
+		// And actual flag value.
+		_ = fl.Value.Set(f.Value.String())
+		newPfs.AddFlag(&fl)
+	})
+
+	return newPfs
+}
+
+// NewParameterFlagSet returns new instance of ParameterFlagSet.
+func NewParameterFlagSet(name string, eh flag.ErrorHandling) *ParameterFlagSet {
+	return &ParameterFlagSet{
+		FlagSet:       *flag.NewFlagSet(name, eh),
+		name:          name,
+		errorHandling: eh,
+	}
 }
 
 // Env is a (nested) test environment.
--- a/go/oasis-test-runner/scenario/e2e/e2e.go
+++ b/go/oasis-test-runner/scenario/e2e/e2e.go
@@ -12,14 +12,13 @@ import (
 )
 
 const (
+	// cfgNodeBinary is the path to oasis-node executable.
 	cfgNodeBinary = "node.binary"
 )
 
 var (
-	// E2eParamsDummy is a dummy instance of e2eImpl used to register e2e-wise parameters.
-	E2eParamsDummy *e2eImpl = &e2eImpl{name: "e2e"}
-
-	logger = logging.GetLogger("e2e/common")
+	// E2eParamsDummy is a dummy instance of e2eImpl used to register global e2e flags.
+	E2eParamsDummy *e2eImpl = newE2eImpl("")
 )
 
 // e2eImpl is a base class for tests involving oasis-node.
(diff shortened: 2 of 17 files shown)


go/oasis-test-runner/scenario/e2e/runtime.go
```diff
@@ -69,60 +67,49 @@ type runtimeImpl struct {
...
+		e2eImpl:      *newE2eImpl(fullName),
+		clientBinary: clientBinary,
+		clientArgs:   clientArgs,
+	}
+
+	sc.flags.String(cfgClientBinaryDir, "", "path to the client binaries directory")
+	sc.flags.String(cfgRuntimeBinaryDir, "", "path to the runtime binaries directory")
+	sc.flags.String(cfgRuntimeLoader, "oasis-core-runtime-loader", "path to the runtime loader")
+	sc.flags.String(cfgTEEHardware, "", "TEE hardware to use")
+	sc.flags.Bool(cfgIasMock, true, "if mock IAS service should be used")
```
ptrus, line 89: :+1:

go/oasis-test-runner/scenario/e2e/gas_fees_staking.go
```diff
@@ -65,7 +65,11 @@ func (sc *gasFeesImpl) Fixture() (*oasis.NetworkFixture, error) {
 	}
 
 	var tee node.TEEHardware
-	err = tee.FromString(sc.TEEHardware)
+	teeStr, err := sc.flags.GetString(cfgTEEHardware)
+	if err != nil {
```
kostko, line 69: Huh, now we need to handle errors in all places where fetching flags? Does that mean that if you pass an invalid type (e.g., a string where an integer value is expected), the error will only be emitted once it hits the `Get*` call? Can you check?

It would be nice if this could be simplified.

matevz:
Cloning scenario and its flagset is not as simple as I thought. Doing
```
newSc.flags =  flag.NewFlagSet(sc.name, flag.ContinueOnError)
newSc.flags.AddFlagSet(sc.flags)
```
keeps old references to values so the flagset is not correctly cloned. I will need to make separate helpers per scenario for copying the flags...

matevz:
> Cloning scenario and its flagset is not as simple as I thought. Doing
> 
> ```
> newSc.flags =  flag.NewFlagSet(sc.name, flag.ContinueOnError)
> newSc.flags.AddFlagSet(sc.flags)
> ```
> 
> keeps old references to values so the flagset is not correctly cloned. I will need to make separate helpers per scenario for copying the flags...

This was now solved in d676397195e2bd68faf62a5f88aca7d7dda40a3c

go/oasis-test-runner/scenario/e2e/gas_fees_staking.go
matevz: Actually all setter functions for flags are called when setting up scenarios (when generating individual parameter sets), so if you pass invalid value (e.g. string instead of int), it will correctly complain.

matevz: > It would be nice if this could be simplified.

The idea was that flags are now explicit to separate them from other non-settable scenario fields. Before they were just wrappers around scenario variables and initialization/usage was unclear.

matevz: I added `getTEEHardware()` helper which does this now.

kostko: > Actually all setter functions for flags are called when setting up scenarios (when generating individual parameter sets), so if you pass invalid value (e.g. string instead of int), it will correctly complain.

So when can this actually return an error?

matevz: > So when can this (e.g., the `Get*` methods) actually return an error?

If a programmer registers a flag as fs.String(cfgMyFlag) and then he calls GetInt(cfgMyFlag). The new `getTEEHardware()` helper returns error, if TEE flag is invalid (it's not either `""` or `"intel-sgx"`).

kostko: This should be a panic (programmer error) not a runtime error.

go/oasis-test-runner/scenario/e2e/e2e.go
```diff
@@ -27,37 +28,50 @@ type e2eImpl struct {
...
-		nodeBinary: "oasis-node",
+	// Empty scenario name is used for registering global parameters only.
+	fullName := "e2e"
+	if name != "" {
+		fullName += "/" + name
+	}
+
+	sc := &e2eImpl{
+		name:   fullName,
+		logger: logging.GetLogger("scenario/e2e/" + name),
```
ptrus, line 43: if `name == ""` then this will be `"scenario/e2e/"`? maybe use `"scenario/fullName"`

go/oasis-test-runner/scenario/e2e/gas_fees_staking.go
matevz: We'll just ignore those for now as `viper` does.

go/oasis-test-runner/scenario/e2e/e2e.go
```diff
@@ -12,52 +12,64 @@ import (
...
-		name:       sc.name,
-		logger:     sc.logger,
-		nodeBinary: sc.nodeBinary,
+	newSc := &e2eImpl{
+		net:    sc.net,
+		name:   sc.name,
+		logger: sc.logger,
+		flags:  newE2eFlags(sc.name),
 	}
+	scenario.CopyFlagValues(newSc.flags, sc.flags)
```
kostko, line 62: This seems slightly awkward, can you instead define your own flagset like this:
```golang
type FlagSet struct {
    flag.FlagSet
}
```
And then define a `Clone()` method on it that returns a cloned flagset? Put it under the `env` package.

Actually I see that you already do this for `env.ParameterFlagSet`, why not use that type everywhere (e.g., make `Parameters()` return a `*ParameterFlagSet`) and add a `Clone()` method?

matevz: I pushed a draft version of the Clone() method. This seems to work, but it's ugly - I need to use `reflect` package to copy the interface.

matevz: cleaned up, ready for review

go/oasis-test-runner/env/env.go
```diff
@@ -52,6 +53,27 @@ func (pfs *ParameterFlagSet) MarshalJSON() ([]byte, error) {
 	return json.Marshal(ps)
 }
 
+// Clone clones the parameter flagset.
+//
+// This function is usually called when cloning scenarios where we also want to clone existing
+// parameter values.
+func (pfs *ParameterFlagSet) Clone() *ParameterFlagSet {
+	// XXX: pfs.name and pfs.errorHandling are private. Use reflection.
```
kostko, line 61: Is this just to get the two parameters from the constructor? Why not add a `NewParameterFlagSet` function and record these two arguments in `ParameterFlagSet` instead?

go/oasis-test-runner/env/env.go
```diff
@@ -25,6 +26,8 @@ type CleanupFn func()
 // ParameterFlagSet is a wrapper for flag.FlagSet to produce nicer JSON output.
 type ParameterFlagSet struct {
 	flag.FlagSet
+	name          string
```
kostko, line 29: Blank line above.

go/oasis-test-runner/scenario/e2e/runtime.go
```diff
@@ -250,13 +232,23 @@ func (sc *runtimeImpl) start(childEnv *env.Env) (<-chan error, *exec.Cmd, error)
 	return clientErrCh, cmd, nil
 }
 
+// getTEEHardware returns TEEHardware based on viper parameter.
```
kostko, line 235:
```suggestion
// getTEEHardware returns the configured TEE hardware.
```

PR #2941, opened 2020-05-26
metrics: Fix prometheus push issues

### LINKED ISSUE #2936: oasis-node crashes, if metrics in push mode and PGW crashes
Currently, if oasis-node is run in metrics push mode and push gateway unexpectedly crashes, so does oasis-node.

Improve oasis-node Prometheus pusher code as follows:
1. If PGW is not responding, try to reconnect.
2. When starting oasis-node in push mode and if PGW is not accessible, try to reconnect instead of failing to start.

Yawning:
* [x] go/oasis-node/cmd/common/metrics: Re-create the pusher on failure
 * ~investigate mystery crashes~
 * [x] Deprecate official push support (Feature-gate behind a debug flag).

Fixes #2936

### DIFF
--- a/.changelog/2936.bugfix.1.md
+++ b/.changelog/2936.bugfix.1.md
@@ -1,0 +1,5 @@
+go/oasis-node/cmd/common/metrics: Re-create the pusher on failure
+
+When using prometheus' push client, any single failure causes the client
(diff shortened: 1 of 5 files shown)


Yawning:
The best way to fix mystery crashes with little to no information is to just remove features.

Issue #2919, opened 2020-05-18
Notify runtimes of its key manager policy updates


kostko:
Currently runtimes are unaware of any key manager policy updates. For example, the runtime (simple-keyvalue) used in E2E tests initializes its key manager policy on startup and passes it to the key manager client.

If the key manager policy changes while the runtime is running (e.g., due to the key manager being upgraded, see #2517), the runtime is not notified and keeps using the old policy. This will cause EnclaveRPC sessions to fail when enclave identity changes.

One possible way of resolving this is extending the Runtime Host Protocol (RHP) with a method for notifying the runtime of a new key manager policy:
* Currently, the runtime can query the policy at any point by sending a `HostKeyManagerPolicyRequest` message over RHP.
* We could turn this around (removing the above message) and introduce a `RuntimeKeyManagerPolicyUpdateRequest` that the host uses to notify the runtime of a new policy.

Watching policy updates and propagating them to the runtime should most likely be part of the common runtime host handler (see `go/worker/common/committee/runtime_host.go`).

PR #2947, opened 2020-05-27
notify runtimes of its key manager policy updates

ptrus:
Fixes: #2919

This will also require (minor) changes to the oasis-runtime (updating the initalizer with the keymanager-update handle)

### DIFF
--- a/.changelog/2919.bugfix.md
+++ b/.changelog/2919.bugfix.md
@@ -1,0 +1,6 @@
+runtime: Notify runtimes of its key manager policy updates
+
+Before runtimes were unaware of any key-manager policy updates. The runtime
+only queried for the active key-manager policy at startup. This is now changed
+so that the host notifies runtimes of any key-manager policy changes and
+runtime updates the policies.
--- a/client/src/rpc/client.rs
+++ b/client/src/rpc/client.rs
@@ -1,7 +1,10 @@
 //! Enclave RPC client.
-use std::sync::{
-    atomic::{AtomicBool, Ordering},
-    Arc, Mutex,
+use std::{
+    collections::HashSet,
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc, Mutex,
+    },
 };
 
 use failure::{Fail, Fallible};
@@ -19,7 +22,7 @@ use tokio_executor::spawn;
 #[cfg(not(target_env = "sgx"))]
 use oasis_core_runtime::common::runtime::RuntimeId;
 use oasis_core_runtime::{
-    common::cbor,
+    common::{cbor, sgx::avr::EnclaveIdentity},
     protocol::Protocol,
     rpc::{
         session::{Builder, Session},
@@ -371,6 +374,15 @@ impl RpcClient {
                     Err(err)
                 }),
         )
+    }
+
+    /// Update session enclaves if changed.
+    pub fn update_enclaves(&self, enclaves: Option<HashSet<EnclaveIdentity>>) {
+        let mut session = self.inner.session.lock().unwrap();
+        if session.builder.get_remote_enclaves() != &enclaves {
+            session.builder = session.builder.clone().remote_enclaves(enclaves);
+            session.reset();
+        }
     }
 }
 
--- a/go/worker/common/committee/runtime_host.go
+++ b/go/worker/common/committee/runtime_host.go
@@ -3,19 +3,26 @@ package committee
 import (
 	"context"
 	"errors"
-	"fmt"
+	"sync"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/opentracing/opentracing-go"
 
 	"github.com/oasislabs/oasis-core/go/common/cbor"
-	consensus "github.com/oasislabs/oasis-core/go/consensus/api"
+	"github.com/oasislabs/oasis-core/go/common/logging"
 	keymanagerApi "github.com/oasislabs/oasis-core/go/keymanager/api"
 	keymanagerClient "github.com/oasislabs/oasis-core/go/keymanager/client"
-	registry "github.com/oasislabs/oasis-core/go/registry/api"
+	"github.com/oasislabs/oasis-core/go/runtime/host"
 	"github.com/oasislabs/oasis-core/go/runtime/host/protocol"
 	"github.com/oasislabs/oasis-core/go/runtime/localstorage"
 	runtimeRegistry "github.com/oasislabs/oasis-core/go/runtime/registry"
 	storage "github.com/oasislabs/oasis-core/go/storage/api"
+)
+
+const (
+	retryInterval = 1 * time.Second
+	maxRetries    = 15
 )
 
 var (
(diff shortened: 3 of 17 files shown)


.changelog/2919.bugfix.md
```diff
@@ -0,0 +1,6 @@
+notify runtimes of its key manager policy updates
```
kostko, line 1:
```suggestion
runtime: Notify runtimes of its key manager policy updates
```

client/src/rpc/client.rs
```diff
@@ -372,6 +375,13 @@ impl RpcClient {
                 }),
         )
     }
+
+    /// Updates session enclaves.
+    pub fn update_enclaves(&self, enclaves: Option<HashSet<EnclaveIdentity>>) -> () {
```
kostko, line 380:
```suggestion
    pub fn update_enclaves(&self, enclaves: Option<HashSet<EnclaveIdentity>>) {
```

client/src/rpc/client.rs
```diff
@@ -372,6 +375,13 @@ impl RpcClient {
...
+        let mut session = self.inner.session.lock().unwrap();
+        session.builder = session.builder.clone().remote_enclaves(enclaves);
```
kostko, line 382: Maybe it should check if the enclave identities differ from the existing ones and do nothing in case they don't.

go/worker/common/runtime_host.go
```diff
@@ -63,14 +66,18 @@ func (n *RuntimeHostNode) GetHostedRuntime() host.Runtime {
...
+// notifiers when provisioning hosted runtimes.
 type RuntimeHostHandlerFactory interface {
 	// GetRuntime returns the registered runtime for which a runtime host handler is to be created.
 	GetRuntime() runtimeRegistry.Runtime
 
 	// NewRuntimeHostHandler creates a new runtime host handler.
 	NewRuntimeHostHandler() protocol.Handler
+
+	// NewNotifier creates a new runtime host notifier.
+	// TODO: return a notifier and start it later.
```
kostko, line 79: Yeah it would be nicer if it returned a notifier interface.

go/worker/common/committee/runtime_host.go
```diff
@@ -125,6 +106,99 @@ func (n *Node) GetRuntime() runtimeRegistry.Runtime {
...
+			retry := backoff.WithMaxRetries(backoff.NewConstantBackOff(retryInterval), maxRetries)
+			err := backoff.Retry(call, backoff.WithContext(retry, h.ctx))
+			if err != nil {
+				h.logger.Error("backoff error",
+					"err", err,
+				)
+				continue
+			}
+
+			if response.Error != nil {
```
kostko, line 178: Doesn't `h.host.Call` above already handle this case and wraps it in an `error` return value? So this should never be triggered.

go/worker/common/committee/runtime_host.go
```diff
@@ -125,6 +106,99 @@ func (n *Node) GetRuntime() runtimeRegistry.Runtime {
...
+					return err
+				}
+				response = resp
+				return nil
+			}
+
+			retry := backoff.WithMaxRetries(backoff.NewConstantBackOff(retryInterval), maxRetries)
+			err := backoff.Retry(call, backoff.WithContext(retry, h.ctx))
+			if err != nil {
+				h.logger.Error("backoff error",
```
kostko, line 173: Probably should say something like "failed dispatching key manager policy update to runtime".

keymanager-client/src/client.rs
```diff
@@ -85,8 +83,9 @@ impl RemoteClient {
 
     /// Create a new key manager client with runtime-internal transport.
     ///
-    /// Using this method automatically obtains valid key manager enclave identities via the
-    /// worker-host protocol.
+    /// Using this method valid enclave identities won't be preset and should
+    /// be obtained via the worker-host protocol and updated with the set_policy
+    /// method.
```
kostko, line 88: Maybe also say that until `set_policy` is called session establishment will fail (I assume this is the case)?

keymanager-client/src/client.rs
```diff
@@ -149,6 +132,21 @@ impl RemoteClient {
             keys_cache_sizes,
         )
     }
+
+    /// Sets client allowed enclaves from key manager policy.
```
kostko, line 136:
```suggestion
    /// Set client allowed enclaves from key manager policy.
```

keymanager-client/src/client.rs
```diff
@@ -149,6 +132,21 @@ impl RemoteClient {
...
+    pub fn set_policy(&self, signed_policy_raw: Vec<u8>) -> () {
```
kostko, line 137:
```suggestion
    pub fn set_policy(&self, signed_policy_raw: Vec<u8>) {
```

keymanager-client/src/client.rs
```diff
@@ -149,6 +132,21 @@ impl RemoteClient {
...
+        let untrusted_policy: SignedPolicySGX = match cbor::from_slice(&signed_policy_raw) {
+            Ok(sp) => sp,
+            Err(err) => panic!("error obtaining list of KM enclaves: {}", err),
```
kostko, line 140: Should this method maybe just return an error instead of panicking? And then the caller can decide to panic. E.g. just make this method return `Fallible<()>` and then use `cbor::from_slice(&signed_policy_raw)?`.

runtime/src/dispatcher.rs
```diff
@@ -142,6 +142,16 @@ impl Dispatcher {
         Ok(())
     }
 
+    pub fn update_runtime_keymanager_policy(
```
kostko, line 145: How is this different from `queue_request` above?

runtime/src/protocol.rs
```diff
@@ -324,6 +324,14 @@ impl Protocol {
                 self.dispatcher.queue_request(ctx, id, req)?;
                 Ok(None)
             }
+            // #[cfg(target_env = "sgx")] XXX: only if sgx?
```
kostko, line 327: I guess it is fine if this is supported all the time.

runtime/src/rpc/dispatcher.rs
```diff
@@ -211,4 +217,20 @@ impl Dispatcher {
             },
         }
     }
+
+    // Handle key manager policy update.
+    pub fn handle_km_policy_update(&self, signed_policy_raw: Vec<u8>) -> () {
```
kostko, line 222:
```suggestion
    pub fn handle_km_policy_update(&self, signed_policy_raw: Vec<u8>) {
```

runtime/src/rpc/dispatcher.rs
```diff
@@ -211,4 +217,20 @@ impl Dispatcher {
...
+        match &self.km_policy_handler {
+            Some(handler) => handler(signed_policy_raw),
```
kostko, line 224: Could simplify this by just saying `self.km_policy_handler.and_then(|handler| handler(signed_policy_raw));`.

runtime/src/rpc/dispatcher.rs
```diff
@@ -211,4 +217,20 @@ impl Dispatcher {
...
+            None => (),
+        }
+    }
+
+    /// Update key manager policy update handler.
+    pub fn set_keymanager_policy_update_handler(
+        &mut self,
+        f: Option<Box<KeyManagerPolicyHandler>>,
+    ) -> () {
```
kostko, line 233:
```suggestion
    ) {
```

go/worker/common/committee/runtime_host.go
ptrus: yeah looks like it does, will fix it (also in he `worker/keymanager/worker.go` where i took this from :grin:

kostko: Ooops, probably missed it when changing it to use `Call` (previously it used a method that didn't do this automatically). Thanks!

keymanager-client/src/client.rs
```diff
@@ -4,16 +4,15 @@ use std::{
...
 #[cfg(not(target_env = "sgx"))]
 use grpcio::Channel;
 use io_context::Context;
 use lru::LruCache;
-#[cfg(target_env = "sgx")]
 use std::iter::FromIterator;
 
-#[cfg(target_env = "sgx")]
-use oasis_core_runtime::{common::cbor, protocol::ProtocolError, types::Body};
+use oasis_core_runtime::common::cbor;
```
kostko, line 15: Move import to group below?

keymanager-client/src/client.rs
```diff
@@ -4,16 +4,15 @@ use std::{
...
 };
 
+use failure::Fallible;
 use futures::{future, prelude::*};
 #[cfg(not(target_env = "sgx"))]
 use grpcio::Channel;
 use io_context::Context;
 use lru::LruCache;
-#[cfg(target_env = "sgx")]
 use std::iter::FromIterator;
```
kostko, line 13: Move `std` import to group above.

runtime/src/rpc/dispatcher.rs
```diff
@@ -211,4 +217,19 @@ impl Dispatcher {
             },
         }
     }
+
+    // Handle key manager policy update.
```
kostko, line 221:
```suggestion
    /// Handle key manager policy update.
```

tests/runtimes/simple-keyvalue/src/main.rs
```diff
@@ -255,11 +255,21 @@ fn main() {
             1024,
             trusted_policy_signers(),
         ));
+        let initializer_client = km_client.clone();
```
kostko, line 258:
```suggestion
        let initializer_km_client = km_client.clone();
```

PR #2951, opened 2020-05-28
go/oasis-node: Signer configuration fixes

### LINKED ISSUE #2949: Composite signer configuration is not possible in YAML
The problem is that we currently define the signer backend argument as simply `signer` and all the other arguments are nested under it, like `signer.composite.backends`. And while you can do `--signer composite --signer.composite.backends 1:file` you cannot this in YAML as `signer` would need to be both a string and a map.

We should probably rename `signer` to `signer.backend`.

### LINKED ISSUE #2950: oasis-node doesn't parse signer related flags
##### SUMMARY
Passing --signer flag to oasis-node will throw: 
`unknown flag: --signer`

##### ISSUE TYPE
- Bug Report

##### OASIS NODE VERSION

```Software version: 20.6
Runtime protocol version: 0.13.0
Consensus protocol version: 0.25.0
Committee protocol version: 0.9.0
Tendermint core version: 0.33.4
ABCI library version: 0.16.2
Go toolchain version: 1.14.3
```

##### OS / ENVIRONMENT
Ubuntu 18.04

##### STEPS TO REPRODUCE
oasis-node --config ./etc/config.yml --signer file 

##### ACTUAL RESULTS

```
unknown flag: --signer
```

##### EXPECTED RESULTS
All signer related configs should be correctly parsed.

kostko:
Fixes #2949
Fixes #2950 

This may require updates to the operator docs after 20.7 is released as `--signer` has been renamed to `--signer.backend`.

### DIFF
--- a/.changelog/2949.bugfix.md
+++ b/.changelog/2949.bugfix.md
@@ -1,0 +1,6 @@
+go/oasis-node: Fix signer configuration via YAML
+
+The `--signer` argument has been renamed to `--signer.backend` to allow
+signer configuration to be passed via a YAML config. Previously, this would
+be impossible as `signer` would need to be both a string and a map at the
+same time when set via a YAML config.
--- a/.changelog/2950.bugfix.md
+++ b/.changelog/2950.bugfix.md
@@ -1,0 +1,1 @@
+go/oasis-node: Fix parsing of signer configuration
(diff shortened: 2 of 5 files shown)


.changelog/2949.bugfix.md
```diff
@@ -0,0 +1,6 @@
+go/oasis-node: Fix signer configuration via YAML
+
+The `--signer` argument has been renamed to `--signer.backend` to allow
```
ptrus, line 3: Wondering if it would be useful to have a "configuration change" fragment type? (iirc currently we just have the label)

kostko: Good idea, I'll open an issue.

PR #2957, opened 2020-05-29
oasis-node: Add support for IsReady and WaitReady RPC methods

### LINKED ISSUE #2130: Add IsReady/WaitReady to client protocol

matevz:
Fixes #2130


.changelog/2130.internal.md
```diff
@@ -0,0 +1,11 @@
+client protocol: Add IsReady() and WaitReady() RPC methods
```
kostko, line 1:
```suggestion
go/control: Add IsReady() and WaitReady() RPC methods
```

.changelog/2130.internal.md
```diff
@@ -0,0 +1,11 @@
...
+
+Beside `IsSynced()` and `WaitSynced()` which are triggered when a node is
+registered, new `IsReady()` and `WaitReady()` methods have been added to
```
kostko, line 4: The `*Synced` methods are not triggered when a node is registered, only when the consensus backend is synced. Only the new `*Ready` calls include node registration.

go/control/api/api.go
```diff
@@ -48,9 +52,12 @@ type Status struct {
 }
 
 // Shutdownable is an interface the node presents for shutting itself down.
+// TODO: Rename Shutdownable to Node? And Node to NodeImpl?
 type Shutdownable interface {
 	// RequestShutdown is the method called by the control server to trigger node shutdown.
 	RequestShutdown() (<-chan struct{}, error)
+	// Ready returns a channel that is closed once node is ready.
```
kostko, line 59: Blank line above this one.

go/control/api/api.go
```diff
@@ -48,9 +52,12 @@ type Status struct {
 }
 
 // Shutdownable is an interface the node presents for shutting itself down.
+// TODO: Rename Shutdownable to Node? And Node to NodeImpl?
```
kostko, line 55: Could also rename this interface to `ControlledNode` and leave the `Node` as-is.

go/control/api/grpc_debug.go
```diff
@@ -99,6 +99,10 @@ func (c *debugControllerClient) WaitNodesRegistered(ctx context.Context, count i
 	return c.conn.Invoke(ctx, methodWaitNodesRegistered.FullName(), count, nil)
 }
 
+func (c *debugControllerClient) WaitReady(ctx context.Context, count int) error {
```
kostko, line 102: Why add this here, this seems wrong, the method is for the non-debug part?

go/oasis-node/cmd/debug/control/control.go
```diff
@@ -36,6 +37,14 @@ var (
 		Run:   doWaitNodes,
 	}
 
+	controlWaitReadyCmd = &cobra.Command{
+		Use:   "wait-ready",
+		Short: "wait for node to become ready",
+		Long: "Wait for node consensus to be synced and runtimes being registered, " +
```
kostko, line 43:
```suggestion
		Long: "Wait for the consensus backend to be synced and runtimes being registered, " +
```

go/oasis-node/cmd/node/node.go
```diff
@@ -166,6 +167,50 @@ func (n *Node) RequestShutdown() (<-chan struct{}, error) {
 	return n.RegistrationWorker.Quit(), nil
 }
 
+func (n *Node) Ready() <-chan struct{} {
```
kostko, line 170: Add documentation comment for public method.

go/oasis-node/cmd/node/node.go
```diff
@@ -166,6 +167,50 @@ func (n *Node) RequestShutdown() (<-chan struct{}, error) {
...
+	return n.readyCh
+}
+
+func (n *Node) WaitReady() error {
```
kostko, line 174: Why is this method public, it seems to be used internally and calling it from the outside can cause a panic due to closing an already closed channel? Also my suggestion is to change this to not return an error, log it instead and below just call `go n.waitReady()` instead of having that wrapper closure.

go/oasis-node/cmd/node/node.go
```diff
@@ -413,6 +458,13 @@ func (n *Node) startWorkers(logger *logging.Logger) error {
 		}
 	}
 
+	// Close readyCh once all workers and runtimes are initialized.
+	go func() {
```
kostko, line 462: Change this as mentioned in the other comment.

go/oasis-test-runner/scenario/e2e/runtime.go
```diff
@@ -607,34 +607,53 @@ func (sc *runtimeImpl) initialEpochTransitions() error {
...
+	for _, n := range sc.net.ComputeWorkers() {
+		if err := n.WaitReady(ctx); err != nil {
+			return fmt.Errorf("failed to wait for a compute worker: %w", err)
+		}
+	}
+	sc.logger.Info("waiting for byzantine nodes to initialize",
+		"num_byzantine", len(sc.net.Byzantine()),
+	)
+	for _, n := range sc.net.Byzantine() {
+		if err := n.WaitReady(ctx); err != nil {
```
kostko, line 654: I don't think this will work for Byzantine nodes? Byzantine nodes do not use the regular node APIs? Feel free to keep using `WaitNodesRegistered` here instead of adding support for `WaitReady` to Byzantine nodes (which doesn't make sense).

go/worker/keymanager/worker.go
```diff
@@ -178,29 +184,20 @@ func (w *Worker) callLocal(ctx context.Context, data []byte) ([]byte, error) {
...
-				rt.Capabilities.TEE = startedEvent.CapabilityTEE
-				return nil
-			})
-
-			// If initialization failed setup a retry ticker.
-			if w.initTicker == nil {
-				w.initTicker = backoff.NewTicker(backoff.NewExponentialBackOff())
-				w.initTickerCh = w.initTicker.C
-			}
+		<-w.Initialized()
```
kostko, line 188: Huh, I don't get this? This seems to defeat the point of this handler (pre-registration on failures) and may lead to deadlocks (`updateStatus` must be called to close `initCh` so if it failed then this deferred handler will block forever as it will block the event loop from propagating more status updates). Also wasn't the point mentioned in the removed comment that this should be changed to always pre-register?

cc @ptrus @Yawning

Yawning: This change breaks the replication code in various ways yes.  Key managers need to either register with a populated `ExtraInfo` field (ready to service requests), or with an empty one (replication needed).

Trampling over the field like in this PR will ensure that newly added replicas will never be used, till the nodes are restarted (I think).

go/control/api/grpc_debug.go
matevz: Oops, forgot to remove this.

go/worker/keymanager/worker.go
matevz: `SetAvailable()` is now first called for pre-registration with empty extra-info (to enable replication) and then again after successful initialization with populated extra-info.

go/worker/keymanager/worker.go
```diff
@@ -178,32 +184,23 @@ func (w *Worker) callLocal(ctx context.Context, data []byte) ([]byte, error) {
...
-				return nil
-			})
-
-			// If initialization failed setup a retry ticker.
-			if w.initTicker == nil {
-				w.initTicker = backoff.NewTicker(backoff.NewExponentialBackOff())
-				w.initTickerCh = w.initTicker.C
-			}
+		// If initialization failed setup a retry ticker.
+		if w.initTicker == nil {
```
ptrus, line 189: before, this happened only if initialization failed, now it sets up a retry ticker even if it succeeds?

go/worker/keymanager/worker.go
```diff
@@ -178,32 +184,23 @@ func (w *Worker) callLocal(ctx context.Context, data []byte) ([]byte, error) {
...
+			w.initTicker = backoff.NewTicker(backoff.NewExponentialBackOff())
+			w.initTickerCh = w.initTicker.C
 		}
 	}()
 
+	// Pre-register to allow replication.
+	w.roleProvider.SetAvailable(func(n *node.Node) error {
```
ptrus, line 196: I think it might make more sense to move this out of `updateStatus` function, probably to where we watch for the started event and do it once we receive the started event. (that way this will also be done only once, and not on every status update - which is i think what we want)

go/worker/keymanager/worker.go
```diff
@@ -178,29 +184,11 @@ func (w *Worker) callLocal(ctx context.Context, data []byte) ([]byte, error) {
...
-				return nil
-			})
-
-			// If initialization failed setup a retry ticker.
-			if w.initTicker == nil {
-				w.initTicker = backoff.NewTicker(backoff.NewExponentialBackOff())
-				w.initTickerCh = w.initTicker.C
-			}
+		// If initialization failed setup a retry ticker.
+		if !w.initialSyncDone && w.initTicker == nil {
```
kostko, line 189: This seems wrong? The condition should be `if !initOk && w.initTicker == nil` to match what was there before.

go/worker/keymanager/worker.go
```diff
@@ -408,6 +395,15 @@ func (w *Worker) worker() { // nolint: gocyclo
 					continue
 				}
 
+				// Pre-register to allow replication.
+				w.roleProvider.SetAvailable(func(n *node.Node) error {
```
kostko, line 399: Why did you move this here, shouldn't it be part of `updateStatus`?

ptrus: i suggested that, i think we want to do the "pre-registration" (needed for replication) when the runtime is started and only once - so its a bit more suited here? (while the updateStatus would do that on every status update?)

kostko: Yeah after thinking about the state transitions, you're right about this only mattering here. But it still would need to check if `w.enclaveStatus` is set. So you can keep it here but add the check to not overwrite existing registrations.

go/worker/keymanager/worker.go
matevz: Reintroduced `initOk` back.

go/worker/keymanager/worker.go
matevz: Done, but the deadlock persists :thinking:

go/worker/storage/committee/node.go
```diff
@@ -540,16 +540,20 @@ func (n *Node) worker() { // nolint: gocyclo
 	defer close(n.diffCh)
 
 	// Wait for the common node to be initialized.
+	n.logger.Info("matevz0")
```
kostko, line 543: Feel free to remove these debug statements :-)

matevz: Oops :wink:

go/worker/keymanager/worker.go
```diff
@@ -302,12 +291,22 @@ func (w *Worker) updateStatus(status *api.Status, startedEvent *host.StartedEven
...
+	w.roleProvider.SetAvailableWithCallback(func(n *node.Node) error {
 		rt := n.AddOrUpdateRuntime(w.runtime.ID())
 		rt.Version = startedEvent.Version
 		rt.ExtraInfo = cbor.Marshal(signedInitResp)
 		rt.Capabilities.TEE = startedEvent.CapabilityTEE
 		return nil
+	}, func(context.Context) error {
+		w.logger.Info("Key manager registered")
+
+		// Signal that we are initialized.
```
matevz, line 303: Maybe mention here or where `initCh` is declared that KeyManager triggers `Initialized()` when it is ready to serve requests + is registered (in contrast to other workers where it's not mandatory to be registered).

go/worker/storage/committee/node.go
```diff
@@ -596,6 +596,7 @@ func (n *Node) worker() { // nolint: gocyclo
 
 	heap.Init(outOfOrderDiffs)
 
+	n.logger.Info("about to close init channel")
```
matevz, line 599: This is a debugging remainder. Can be removed.

go/worker/storage/worker.go
```diff
@@ -200,6 +200,8 @@ func (s *Worker) Start() error {
 			<-r.Initialized()
 		}
 
+		s.logger.Info("storage worker runtimes started, waiting for InitialRegistration")
```
matevz, line 203: This was also for debugging. Can be removed.

go/worker/keymanager/worker.go
kostko: Good idea, changed the documentation comment for the `Initialized` method.

go/keymanager/client/client.go
```diff
@@ -80,14 +81,26 @@ func (c *Client) CallRemote(ctx context.Context, data []byte) ([]byte, error) {
...
-		if status.Code(err) == codes.PermissionDenied {
+		switch {
+		case err == nil:
+		case status.Code(err) == codes.PermissionDenied:
 			// Calls can fail around epoch transitions, as the access policy
 			// is being updated, so we must retry.
 			return err
+		case status.Code(err) == codes.Unavailable:
+			// XXX: HACK: Find a better way to determine the root cause.
+			if strings.Contains(err.Error(), "tls: bad public key") {
```
kostko, line 92: Unfortunately, due to gRPC's use of wrapping RPC errors (see [here](https://github.com/grpc/grpc-go/blob/754ee590a4f386d0910d887f3b8776354042260b/rpc_util.go#L782)) it is not possible to get the original error.

PR #2940, opened 2020-05-26
Change Account ids/addresses to truncated hash of the public key

tjanez:
Closes #2928.

TODO:
- [x] Use Bech32 encoding for marshaling/unmarshaling addresses to text form.
- [x] Implement address context separation in the form of `H(<context> || <pubkey>)`, e.g. `H("oasis-core/address: staking" || <pubkey>)`.
- [x] Add versioning to Account addresses.
- [x] Implement a notion of *reserved* addresses.
- [x] Add staking transaction checks for *reserved* addresses.
- [x] Choose Oasis staking account address Bech32 human readable part.
- [x] Fix E2E tests.

PR #2959, opened 2020-05-29
go/consensus/genesis: Add a public key blacklist

Yawning:
* [x] Basic changes to crypto/signature to support such a thing
 * [x] Alter the genesis document format
 * [x] Make the registry genesis checks blacklist aware
 * [x] Figure out wtf to do for the staking ledger (Resolved: Ignore it till the address format changes)

### DIFF
--- a/go/common/crypto/signature/signer_test.go
+++ b/go/common/crypto/signature/signer_test.go
@@ -85,3 +85,18 @@ func TestContext(t *testing.T) {
 	_, err = PrepareSignerMessage(unregCtx, []byte("message"))
 	require.NoError(err, "PrepareSignerMessage should work with unregisered context (bypassed)")
 }
+
+func TestBlacklist(t *testing.T) {
+	require := require.New(t)
+
+	var pk PublicKey
+	err := pk.UnmarshalHex("d75a980182b10ab7d54bfed3c964073a0ee172f3daa62325af021a68f707511a")
+	require.NoError(err, "UnmarshalHex")
+	require.True(pk.IsValid(), "test key should initially be valid")
+
+	require.False(pk.IsBlacklisted(), "test key should not initially be blacklisted")
+	err = pk.Blacklist()
+	require.NoError(err, "Blacklist")
+	require.True(pk.IsBlacklisted(), "test key should now be blacklisted")
+	require.False(pk.IsValid(), "test key should now be invalid")
+}
--- a/go/consensus/genesis/genesis.go
+++ b/go/consensus/genesis/genesis.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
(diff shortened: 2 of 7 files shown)


Yawning:
This is mostly done, but I need to figure out how we want to handle genesis documents that contain staking ledger entries corresponding to blacklisted public keys.

As far as I know we want to allow this (a public key being invalid shouldn't obliterate the corresponding account state, just the ability to do transactions), but this might be a moot point based on how #2940 ends up being implemented.

go/consensus/genesis/genesis.go
```diff
@@ -41,5 +43,14 @@ func (g *Genesis) SanityCheck() error {
 		return fmt.Errorf("consensus: sanity check failed: timeout commit must be >= 1ms")
 	}
 
+	// Check for duplicate entries in the pk blacklist.
+	m := make(map[signature.PublicKey]bool)
+	for _, v := range g.PublicKeyBlacklist {
+		if m[v] {
```
kostko, line 49: Maybe also check if the key is already statically blacklisted and fail the sanity check? As otherwise `Blacklist()` will fail later.

go/consensus/genesis/genesis.go
```diff
@@ -5,13 +5,15 @@ import (
...
 	"github.com/oasislabs/oasis-core/go/consensus/api/transaction"
 )
 
 // Genesis contains various consensus config flags that should be part of the genesis state.
 type Genesis struct {
-	Backend    string     `json:"backend"`
-	Parameters Parameters `json:"params"`
+	Backend            string                `json:"backend"`
+	Parameters         Parameters            `json:"params"`
+	PublicKeyBlacklist []signature.PublicKey `json:"public_key_blacklist,omitempty"`
```
kostko, line 16: Will we want to change the blacklist anytime in the future (e.g., via an unspecified governance mechanism)? If yes, then we may want to have this under `Parameters` instead.

go/common/crypto/signature/signer_test.go
```diff
@@ -85,3 +85,18 @@ func TestContext(t *testing.T) {
 	_, err = PrepareSignerMessage(unregCtx, []byte("message"))
 	require.NoError(err, "PrepareSignerMessage should work with unregisered context (bypassed)")
 }
+
+func TestBlacklist(t *testing.T) {
```
matevz, line 89: Testing whether you cannot submit transactions (and that transaction is not accepted by other nodes) with blacklisted key would be useful.

kostko: You could do that by using `show_tx` (as long as the blacklist is part of the genesis document), you only would need to load the blacklist there as well (similar to how we set the chain domain separation context from the genesis doc).

PR #2967, opened 2020-06-02
go/runtime/host/sandbox: Retry Call in case the runtime is not yet ready

kostko:
Issue discovered in long-term tests.

### DIFF
--- a/go/runtime/host/sandbox/sandbox.go
+++ b/go/runtime/host/sandbox/sandbox.go
@@ -98,15 +98,26 @@ func (r *sandboxedRuntime) ID() common.Namespace {
 }
 
 // Implements host.Runtime.
-func (r *sandboxedRuntime) Call(ctx context.Context, body *protocol.Body) (*protocol.Body, error) {
-	r.RLock()
-	conn := r.conn
-	r.RUnlock()
+func (r *sandboxedRuntime) Call(ctx context.Context, body *protocol.Body) (rsp *protocol.Body, err error) {
+	callFn := func() error {
+		r.RLock()
+		conn := r.conn
+		r.RUnlock()
 
-	if conn == nil {
-		return nil, fmt.Errorf("runtime is not ready")
+		if conn == nil {
+			return fmt.Errorf("runtime is not ready")
+		}
+		rsp, err = r.conn.Call(ctx, body)
+		if err != nil {
+			// All protocol-level errors are permanent.
+			return backoff.Permanent(err)
+		}
+		return nil
 	}
-	return r.conn.Call(ctx, body)
+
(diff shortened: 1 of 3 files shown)


go/runtime/host/sandbox/sandbox.go
```diff
@@ -98,15 +98,26 @@ func (r *sandboxedRuntime) ID() common.Namespace {
...
 
 // Implements host.Runtime.
-func (r *sandboxedRuntime) Call(ctx context.Context, body *protocol.Body) (*protocol.Body, error) {
-	r.RLock()
-	conn := r.conn
-	r.RUnlock()
-
-	if conn == nil {
-		return nil, fmt.Errorf("runtime is not ready")
+func (r *sandboxedRuntime) Call(ctx context.Context, body *protocol.Body) (rsp *protocol.Body, err error) {
```
ptrus, line 101: I think that with this the retry in [runtime_host](https://github.com/oasisprotocol/oasis-core/blob/83e9a9bd118f0e8261bd2116b5ad15fe7b961527/go/worker/common/committee/runtime_host.go#L181) might not be needed anymore? (and i think we also do similar retry in km-worker that might not be needed anymore either)

kostko: Oh I didn't know those retries were due to the runtime not being ready, otherwise I would have suggested this fix sooner. Yeah if that's the only reason for that retry I'll just go ahead and remove it.

ptrus: Yeah i'm pretty sure i only added the retry due to failures if runtime was not ready yet - so that one can be removed (i don't think there's any expected reason for failure there). 

I misremembered for keymanager, we only retry in the km-client for the remote calls, so that should stay as it is.

kostko: Done, please take a look.

go/worker/common/committee/runtime_host.go
```diff
@@ -164,21 +163,7 @@ func (n *computeRuntimeHostNotifier) watchPolicyUpdates() {
...
-					)
-					return err
-				}
-				response = resp
-				return nil
-			}
-
-			retry := backoff.WithMaxRetries(backoff.NewConstantBackOff(retryInterval), maxRetries)
-			err := backoff.Retry(call, backoff.WithContext(retry, n.ctx))
+			response, err := n.host.Call(n.ctx, req)
```
ptrus, line 166: can also remove the `retryInterval` & `maxRetries` vars

kostko: :+1:

PR #2970, opened 2020-06-02
ci: Automatically rebuild development/CI Docker images

### LINKED ISSUE #295: Regularly build docker image
we ended up with the docker image no longer building last week without anyone having broken.

We should have a regularly scheduled task in the background that attempts re-building our images so that we can be alerted when anything breaks out from under us.

kostko:
Fixes #295

### DIFF
--- a/.buildkite/longtests.pipeline.yml
+++ b/.buildkite/longtests.pipeline.yml
@@ -1,6 +1,6 @@
 # Copied from pipeline.yml.
 docker_plugin_default_config: &docker_plugin_default_config
-  image: "oasislabs/testing:0.3.0"
+  image: "oasisprotocol/oasis-core-ci:master"
   always_pull: true
   workdir: /workdir
   volumes:
--- a/.buildkite/pipeline.sh
+++ b/.buildkite/pipeline.sh
@@ -38,8 +38,22 @@ pr_and_no_code_related_changes() {
       ':(exclude)towncrier.toml'
 }
 
-if pr_and_no_code_related_changes; then
-    cat .buildkite/code-skip.pipeline.yml
+# Determine the oasis-core-ci Docker image tag to use for tests.
+if [[ -n $BUILDKITE_PULL_REQUEST_BASE_BRANCH ]]; then
+  branch=$BUILDKITE_PULL_REQUEST_BASE_BRANCH
 else
-    cat .buildkite/code.pipeline.yml
+  branch=$BUILDKITE_BRANCH
 fi
+branch=${branch//\//-}
+
+export DOCKER_OASIS_CORE_CI_BASE_TAG=$branch
+
+# Decide which pipeline to use.
+pipeline=.buildkite/code.pipeline.yml
+if pr_and_no_code_related_changes; then
+    pipeline=.buildkite/code-skip.pipeline.yml
+fi
+
+# Upload the selected pipeline.
+cat $pipeline | buildkite-agent pipeline upload
+
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,53 @@
+# NOTE: This name appears in GitHub's Checks API and in workflow's status badge.
+name: docker
+
+# Trigger the workflow when:
+on:
+  # A push occurs to one of the matched branches and at least one modified file matches the configured paths.
+  push:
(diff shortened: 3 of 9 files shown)


.buildkite/pipeline.sh
```diff
@@ -38,8 +38,22 @@ pr_and_no_code_related_changes() {
       ':(exclude)towncrier.toml'
 }
 
+# Helper that allows some variables to be injected into the pipeline YAML file.
```
tjanez, line 41: Hmm... this seems a bit ugly.

Could we just export an environment variable and it would be substituted by Buildkite's "standard" environment variable substitution https://buildkite.com/docs/agent/v3/cli-pipeline#environment-variable-substitution?

kostko: Yeah that would be much nicer but I'm not sure if that works in arbitrary places like the docker image field?

kostko: Cool, seems to work.

kostko:
The Buildkite CI will fail as it needs a pipeline configuration change that will break all other jobs, so I haven't made it yet. Once approved I'll change Buildkite config and merge this.

.buildkite/longtests.pipeline.yml
```diff
@@ -1,6 +1,6 @@
 # Copied from pipeline.yml.
```
tjanez, line 1: Hmm... could you file a new issue to stop copying these parts of YAML files but rather break them out to a separate YAML file and combine them when uploading the pipelines via `cat $pipeline-common $pipeline | buildkite-agent pipeline upload`?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,51 @@
+# NOTE: This name appears in GitHub's Checks API and in workflow's status badge.
+name: docker
+
+# Trigger the workflow when:
+on:
+  # A push occurs to one of the matched branches.
```
tjanez, line 6:
```suggestion
  # A push occurs to one of the matched branches and at least one modified file matches the configured paths.
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,51 @@
...
+  push:
+    branches:
+      - master
+      - stable/*
+    paths:
+      - docker/**
```
tjanez, line 12: I think it would be useful to also re-build the Docker images daily (even when there are no new commits) so we get the latest security fixes from the updated base Docker images, e.g. `ubuntu:18.04`, and early detection when some part of the Docker image process starts to fail.

Something like:
https://github.com/oasisprotocol/oasis-core/blob/d0a8e24f1a022347067012f06677800e7124fe59/.github/workflows/ci-reproducibility.yml#L11-L13

Note that [`on.schedule` events](https://help.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events-schedule) have the `GITHUB_REF` set to repository's default branch so this will only work for rebuilding the `master` images.

tjanez: This is actually necessary to do since we use path filtering (on `docker/**`) when building the Docker images for the `master` or `stable/*` branches and those images will thus be only very rarely rebuilt.

kostko: Added.

.buildkite/longtests.pipeline.yml
kostko: #2972

tjanez: Thanks!

.github/workflows/docker.yml
```diff
@@ -0,0 +1,53 @@
+# NOTE: This name appears in GitHub's Checks API and in workflow's status badge.
+name: docker
+
+# Trigger the workflow when:
```
tjanez, line 4: I think this triggering configuration means that Docker images for `stable/*` branches will only be built the first time a new `stable/*` branch is pushed to GitHub and never again, unless there is a manual change to something in the `docker/` directory.

In my opinion, this is fine for now, but perhaps file an issue so that we document this current limitation and perhaps address it in the future, if it will become a frequent issue or [when GitHub Actions gain support for triggering workflows manually](https://github.community/t/github-actions-manual-trigger-approvals/16233) and we can just document how to do that when needed.

PR #2971, opened 2020-06-03
staking: Add WatchEvents

### LINKED ISSUE #2944: Provide contextual block information for Watch* call returns
<!--- First, verify that your feature was not already discussed on GitHub -->
<!--- This template is only for *small* changes. For more complex features, please follow instructions in https://github.com/oasislabs/oasis-core/blob/master/CONTRIBUTING.md#feature-requests -->
<!--- Complete *all* sections as described, this form is processed automatically -->

##### SUMMARY

Currently when subscribing to `Watch*` style API endpoints It's impossible to correlate the values with the actual transactions or blocks that they occurred around. For example, if I subscribe to `WatchTransfers` I will receive a channel that forwards me Transfer events shaped like so:

```
type TransferEvent struct {                                                                                                                                                                                                       
    From   signature.PublicKey `json:"from"`                                                                                                                            
    To     signature.PublicKey `json:"to"`                                                                                                                      
    Tokens quantity.Quantity   `json:"tokens"`                                                                                                                      
}
```

The problem is I have no way to tell which block this ocurred in, at what height, what time, or the associated transactions.

An alternative is to not use the Watch methods and instead call `GetEvents` on every block which will also include a TxHash, but this is much slower when trying to process the whole chain and there is no way to lookup a TxHash using the API yet.

What would be ideal is for common chain information to be included. Things like block height are common to all chains and so I think would be fine to expose regardless of the backing consensus used.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
<!--- Write the name of the component the issue applies to (e.g. go/worker/storage) below, use your best guess if unsure -->

Any API module providing `Watch*`.

##### ADDITIONAL INFORMATION

I'm currently avoiding any `Watch*` API endpoints because they don't provide enough context to be useful, this means instead I'm calling `Get*` style API endpoints. The problem there though is the more calls I make during a block the slower processing the chain becomes.

abukosek:
Closes #2944.

### DIFF
--- a/.changelog/2944.feature.md
+++ b/.changelog/2944.feature.md
@@ -1,0 +1,5 @@
+staking: Add WatchEvents
+
+A new method was added to the staking API, `WatchEvents`.  It returns
+a channel that produces a stream of staking `Event`s.
+The `Event` structure was also extended to include the block height.
--- a/go/Makefile
+++ b/go/Makefile
@@ -55,7 +55,7 @@ fmt:
 # Lint.
 lint:
 	@$(ECHO) "$(CYAN)*** Running Go linters...$(OFF)"
-	@env -u GOPATH golangci-lint run --timeout 2m
+	@env -u GOPATH golangci-lint run --timeout 4m
 
 # Test.
 test-targets := test-unit test-node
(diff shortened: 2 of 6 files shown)


abukosek:
Thanks!

PR #2993, opened 2020-06-10
release-process: create a stable branch when preparing a new release

ptrus:
Currently we create a stable release branch only once we need to backport some
changes. This PR changes the release process to create the stable branch when
creating a new release. This will ensure CI jobs hooked to stable/ branches,
such as building a release tagged CI docker image will be run for every
release.

NOTE: this could probably be automated within the `tag-next-release` make step (with an additional check to only run this when creating a new release and not when backporting), but I'm afraid i'm not enough of a makefile enthusiast to tackle that :/ - can open an issue though.

### DIFF
--- a/.changelog/2993.process.md
+++ b/.changelog/2993.process.md
@@ -1,0 +1,7 @@
+release: Create a stable branch when preparing a new release
+
+Currently we create a stable release branch only once we need to backport some
+changes. This PR changes the release process to create the stable branch when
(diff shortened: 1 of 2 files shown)


.changelog/2993.process.md
```diff
@@ -0,0 +1,7 @@
+release: create a stable branch when preparing a new release
```
tjanez, line 1:
```suggestion
release: Create a stable branch when preparing a new release
```

PR #2997, opened 2020-06-10
ci: Remove push path filters for dockerfile build action

ptrus:


.github/workflows/docker.yml
```diff
@@ -20,9 +24,6 @@ on:
       - stable/*
     paths:
       - docker/**
-  # Or every day at 04:00 UTC (for the default/master branch).
```
kostko, line 23: Why remove this? We should still rebuild in case there were no pushes to master.

ptrus: ah yeah i thought there might be enough rebuilds without this :-) will add it back

PR #3004, opened 2020-06-11
Grab bag of more stuff from the backlog

### LINKED ISSUE #1599: bwrap location should not be hardcoded
## Details

We use [bwrap](https://github.com/projectatomic/bubblewrap)  for running the ekiden worker in the restricted environment. Currently the location is hardcoded to `/usr/bin/bwrap`. If you install bwrap manually, then it's in `/usr/local/bin/bwrap` or even some third location. Make this path configurable.

## Acceptance Criteria
* workerBubblewrapBinary should be read from the config file. If not passed, it could be determined by `which bwrap` command.

### LINKED ISSUE #2469: Rename Rust rpc module(s) to enclave_rpc
The EnclaveRPC client and server are currently located under a `rpc` module in `oasis-core-client`/`oasis-core-runtime` crates. As the term "rpc" is too generic, we should rename the modules to something like `enclave_rpc` to make it more clear.

Yawning:
* [x] go/worker/common: Allow specifying the path to the bwrap binary
 * [x] runtime: Rename the rpc crate to enclave_rpc
 * [x] client: Rename the rpc crate to enclave_rpc

Notes:
 * While the original issue suggests using `which` to figure out where bwrap is I am opting against that as unwise.  If people are so inclined as to install bwrap by hand, they can set the config option themselves.

Fixes #1599
Fixes #2469


go/runtime/host/sandbox/sandbox_test.go
```diff
@@ -11,6 +11,8 @@ import (
 var envRuntimePath = os.Getenv("OASIS_TEST_RUNTIME_HOST_RUNTIME_PATH")
 
 func TestProvisionerSandbox(t *testing.T) {
+	const bwrapPath = "/usr/bin/bwrap" // Sensible systems only.
```
kostko, line 14: I assume the same fix is required in `go/runtime/host/sgx/sgx_test.go` (currently not tested in CI).

Issue #2991, opened 2020-06-09
Attempt to handle runtime interruptions gracefully


kostko:
Currently if an epoch transition happens while the runtime is processing a transaction batch, [the runtime will be restarted](https://github.com/oasisprotocol/oasis-core/blob/master/go/worker/compute/executor/committee/node.go#L518) (handled [here](https://github.com/oasisprotocol/oasis-core/blob/15ae821004d1a239e2e5a16c6df5b98546169e0d/go/runtime/host/sandbox/sandbox.go#L337)), forcing RAK regeneration for SGX runtimes.

We should make the compute worker wait "a bit" for the processing to complete -- one way this could be done is to implement `RuntimeAbortRequest` (in the Rust runtime) to set some kind of an abort flag (that the transaction dispatcher thread can check and abort) and block until the current batch has been processed.

If the `RuntimeAbortRequest` does not return soon enough (within `runtimeInterruptTimeout`), the runtime will be restarted (this is already implemented, the only missing part is that the `RuntimeAbortRequest` is currently a no-op that returns an error).

PR #3008, opened 2020-06-12
runtime: Handle runtime interruptions gracefully

ptrus:
Closes: #2991

TODO: 
- [x] test sgx txsource test to ensure it doesn't encounter timeouts in case interrupt happens


runtime/src/dispatcher.rs
```diff
@@ -253,6 +268,10 @@ impl Dispatcher {
                         signed_policy_raw,
                     );
                 }
+                Ok((_ctx, _id, Body::RuntimeAbortRequest {})) => {
+                    info!(self.logger, "Received abort request");
+                    break 'dispatch;
```
kostko, line 273: Hm, won't terminating the dispatcher bring the runtime down? The abort should be handled like the following:
* Abort is considered finished once the current request that is being handled by the dispatcher is completed (e.g., after this big `match` is done processing whatever it is currently processing). It shouldn't interrupt the dispatcher as then you need to set it up again and this probably leads to restarting the runtime? Instead, the loop should just continue here with the next `rx.recv()` (after signalling that an abort request has been processed).
* It would be cool (read: optional) if the default transaction dispatcher (`MethodDispatcher` under `runtime/src/transaction/dispatcher.rs`) could abort while in the middle of processing a batch, (e.g. during `dispatch_batch`). This could be done by just having a synchronized flag (e.g., an `AtomicBool`) that you could check to see if you need to stop processing the batch and return immediately with an error.

This second part is optional and the whole thing would be useful even if it is not implemented.

ptrus: Ah, i thought we actually want to restart the runtime, but yeah there is no reason for it actually so it makes more sense avoiding that and just waiting for the currently processed batch to finish.

Yeah can also do the 2nd part as part of this PR :+1:

runtime/src/dispatcher.rs
```diff
@@ -253,6 +269,11 @@ impl Dispatcher {
                         signed_policy_raw,
                     );
                 }
+                Ok((_ctx, _id, Body::RuntimeAbortRequest {})) => {
+                    info!(self.logger, "Received abort request");
+                    txn_dispatcher.abort_batch();
```
kostko, line 274: This won't work as the txn dispatcher executes in the same thread as where you process the abort request. The `abort_batch` would need to be called from a different thread (e.g., the protocol thread). (It is fine if you handle the request here to notify the requestor that abort is done, it is just that this will currently always wait for the whole batch to complete processing.)

One of the problems with that is that the transaction dispatcher is not reachable from there.

runtime/src/transaction/dispatcher.rs
```diff
@@ -292,7 +304,10 @@ impl Dispatcher for MethodDispatcher {
         &self,
         batch: &TxnBatch,
         mut ctx: Context,
-    ) -> (TxnBatch, Vec<Tags>, Vec<RoothashMessage>) {
+    ) -> Result<(TxnBatch, Vec<Tags>, Vec<RoothashMessage>)> {
+        // Clear any old abort batch state.
+        self.abort_batch.swap(false, Ordering::SeqCst);
```
kostko, line 309: Why not just `store` as you're not interested in the current value?

runtime/src/transaction/dispatcher.rs
```diff
@@ -172,9 +175,11 @@ pub trait Dispatcher {
         &self,
         batch: &TxnBatch,
         ctx: Context,
-    ) -> (TxnBatch, Vec<Tags>, Vec<RoothashMessage>);
+    ) -> Result<(TxnBatch, Vec<Tags>, Vec<RoothashMessage>)>;
```
kostko, line 178: Note: this probably breaks oasis-runtime.

go/runtime/host/host.go
```diff
@@ -53,7 +53,7 @@ type Runtime interface {
 	Start() error
 
 	// Restart attempts to restart a runtime so that it will be ready to service new requests.
-	Restart(ctx context.Context) error
+	Restart(ctx context.Context, force bool) error
```
ptrus, line 56: Would it make more sense to rename this to `Abort` as now it does not actually do a `Restart` unless it's necessary? (or if the `force` flag is on)

kostko: Yeah good idea, and mention in the documentation comment in which case the runtime is restarted (e.g., if the abort fails or the force flag is set).

runtime/src/dispatcher.rs
```diff
@@ -142,7 +153,20 @@ impl Dispatcher {
         Ok(())
     }
 
-    fn run(&self, initializer: Box<dyn Initializer>, rx: channel::Receiver<QueueItem>) {
+    /// Waits for dispatcher to abort.
+    pub fn wait_abort(&self) -> Result<()> {
```
kostko, line 157: Maybe call it `abort_and_wait` (with doc comment like "Signals to dispatcher that it should abort and waits for the abort to complete.").

runtime/src/dispatcher.rs
```diff
@@ -253,6 +278,10 @@ impl Dispatcher {
                         signed_policy_raw,
                     );
                 }
+                Ok((_ctx, _id, Body::RuntimeAbortRequest {})) => {
+                    info!(self.logger, "Received abort request");
```
kostko, line 282: Add a comment explaining that if this "abort request" is processed here this means that the abort has already been processed.

kostko: On another thought, since you have this `abort_and_wait` why do you need to process the request here at all? Maybe just at the beginning of the loop (e.g., before `rx.recv()`) check if the abort flag is set and if so, clear it and emit something to `abort_tx`? This seems cleaner than having this strange request here.

runtime/src/protocol.rs
```diff
@@ -268,9 +268,13 @@ impl Protocol {
                 info!(self.logger, "Received worker shutdown request");
                 Err(ProtocolError::MethodNotSupported.into())
             }
-            Body::RuntimeAbortRequest {} => {
+            req @ Body::RuntimeAbortRequest {} => {
                 info!(self.logger, "Received worker abort request");
-                Err(ProtocolError::MethodNotSupported.into())
+                self.can_handle_runtime_requests()?;
+                self.dispatcher.queue_request(ctx, id, req)?;
```
kostko, line 274: In case you implement the above, this `queue_request` would not be needed.

runtime/src/transaction/dispatcher.rs
```diff
@@ -292,7 +307,12 @@ impl Dispatcher for MethodDispatcher {
         &self,
         batch: &TxnBatch,
         mut ctx: Context,
-    ) -> (TxnBatch, Vec<Tags>, Vec<RoothashMessage>) {
+    ) -> Result<(TxnBatch, Vec<Tags>, Vec<RoothashMessage>)> {
+        // Clear any old abort batch state.
```
kostko, line 311: Maybe this should move to the general dispatcher (see my previous comments).

runtime/src/dispatcher.rs
```diff
@@ -303,7 +334,8 @@ impl Dispatcher {
         let (mut outputs, mut tags, messages) =
             StorageContext::enter(&mut cache.mkvs, untrusted_local.clone(), || {
                 txn_dispatcher.dispatch_batch(&inputs, txn_ctx)
-            });
+            })
+            .unwrap();
```
kostko, line 338: Do I understand this correctly that the runtime will panic when a batch is aborted? This doesn't seem right? It should just propagate the error back, saying that the batch was aborted.

ptrus: fixed

runtime/src/dispatcher.rs
```diff
@@ -142,7 +153,19 @@ impl Dispatcher {
         Ok(())
     }
 
-    fn run(&self, initializer: Box<dyn Initializer>, rx: channel::Receiver<QueueItem>) {
+    /// Signals to dispatcher that it should abort and waits for the abort to
+    // complete.
```
kostko, line 157:
```suggestion
    /// complete.
```

PR #3009, opened 2020-06-12
Add a CLI command for converting entities' ids to their staking account addresses

### LINKED ISSUE #3003: Add a CLI command for converting entities' ids to their staking account addresses
##### SUMMARY
After https://github.com/oasisprotocol/oasis-core/pull/2940, account address are no longer represented by entities' ids, i.e. public keys, directly. Instead, they are represented by a truncated hash of a corresponding entity's public key, prefixed by a 1 byte address version.
Also, they are no longer encoded with Base64 encoding, but rather use the Bech32 encoding.

It would be useful to provide a CLI command that would convert a Base64-encoded entity id (public key) to its Bech32-encoded staking account address.

##### ISSUE TYPE
- Feature Idea

##### ADDITIONAL INFORMATION
Also consider augmenting the output of other CLI commands with staking account addresses where appropriate.

tjanez:
Closes #3003.

### DIFF
--- a/.changelog/3003.feature.md
+++ b/.changelog/3003.feature.md
@@ -1,0 +1,4 @@
+go/oasis-node/cmd/stake: Add `pubkey2address` commmand
+
+Add `oasis-node stake pubkey2address` CLI command for converting a public key
+(e.g. an entity's ID) to a staking account address.
--- a/go/oasis-node/cmd/stake/stake.go
+++ b/go/oasis-node/cmd/stake/stake.go
@@ -9,8 +9,10 @@ import (
 
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
+	"github.com/spf13/viper"
 	"google.golang.org/grpc"
 
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
 	"github.com/oasisprotocol/oasis-core/go/common/errors"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
 	"github.com/oasisprotocol/oasis-core/go/common/quantity"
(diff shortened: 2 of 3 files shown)


.changelog/3003.feature.md
```diff
@@ -0,0 +1,4 @@
+go/oasis-node/cmd/stake: Add `account get_address` commmand
+
+Add `oasis-node stake account get_address` CLI command for obtaining a
```
tjanez, line 3: Feedback on the (sub)comand's name (or position in the hierarchy) are welcome.

kostko: Maybe call it `pubkey2address` (or `pk2addr`) or something to be more obvious what it does? Having it under stake/account seems fine.

go/oasis-node/cmd/stake/account.go
```diff
@@ -22,6 +23,9 @@ const (
 	// CfgAccountAddr configures the account address.
 	CfgAccountAddr = "stake.account.address"
 
+	// CfgAccountAddr configures the entity ID.
+	CfgEntityID = "entity_id"
```
kostko, line 27: I would rather name this "public key" or something as it can be either an entity or a node public key (nodes also have balances, e.g., to pay fees).

.changelog/3003.feature.md
tjanez: > Maybe call it `pubkey2address` (or `pk2addr`) or something to be more obvious what it does?

Agreed. I've changed it to `pubkey2address`.

> Having it under stake/account seems fine.

On a second though, I think putting it directly under `oasis-node stake` makes more sense.

go/oasis-node/cmd/stake/account.go
tjanez: Agreed.

tjanez:
> Add E2E test to the `stake_cli` scenario, otherwise looks good!

Done. Please, take another look.

PR #3014, opened 2020-06-15
worker/compute: Retry local dispatch between workers

### LINKED ISSUE #3000: Local dispatch between workers should be retried
Currently there are special cases when the local node has multiple roles simultaneously:
* [txn scheduler and executor](https://github.com/oasisprotocol/oasis-core/blob/3ffcef81af553ce198de5d08188c58be79ed98f2/go/worker/compute/executor/committee/node.go#L296)
* [executor and merge](https://github.com/oasisprotocol/oasis-core/blob/3ffcef81af553ce198de5d08188c58be79ed98f2/go/worker/compute/merge/committee/node.go#L275)

In these cases the local submission does not go via P2P but via a direct function call. However such dispatch is never retried in case of failures (the failure cases are the same as with P2P dispatch). The context that should be used in both cases is the [`roundCtx`](https://github.com/oasisprotocol/oasis-core/blob/3ffcef81af553ce198de5d08188c58be79ed98f2/go/worker/common/committee/group.go#L46) specified in the local `Group` (currently not exposed).

abukosek:
Fixes #3000.

### DIFF
--- a/go/worker/compute/merge/committee/node.go
+++ b/go/worker/compute/merge/committee/node.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
 
@@ -273,12 +274,20 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
 // HandleResultsFromExecutorWorkerLocked processes results from an executor worker.
 // Guarded by n.commonNode.CrossNode.
 func (n *Node) HandleResultsFromExecutorWorkerLocked(spanCtx opentracing.SpanContext, commit *commitment.ExecutorCommitment) {
-	// TODO: Context.
-	if err := n.handleResultsLocked(context.TODO(), commit); err != nil {
-		n.logger.Warn("failed to handle results from local executor worker",
-			"err", err,
-		)
-	}
+	// Spawn retry in a goroutine to prevent blocking processing.
+	go func() {
+		call := func() error {
+			n.commonNode.CrossNode.Lock()
+			defer n.commonNode.CrossNode.Unlock()
+			return n.handleResultsLocked(n.roundCtx, commit)
+		}
+		bkoff := backoff.WithContext(backoff.NewExponentialBackOff(), n.roundCtx)
+		if err := backoff.Retry(call, bkoff); err != nil {
+			n.logger.Warn("failed to handle results from local executor worker",
+				"err", err,
+			)
+		}
+	}()
 }
 
 // Guarded by n.commonNode.CrossNode.
--- a/.changelog/3000.bugfix.md
+++ b/.changelog/3000.bugfix.md
@@ -1,0 +1,5 @@
+worker/compute: Retry local dispatch between workers
+
+If a node has multiple roles simultaneously, the local submissions
+don't go through P2P, but via a direct function call.
+This function call is now retried in case of errors.


go/worker/common/committee/group.go
```diff
@@ -255,6 +255,11 @@ type Group struct {
 	logger *logging.Logger
 }
 
+// RoundContext returns the active epoch's round context.
+func (g *Group) RoundContext() context.Context {
+	return g.activeEpoch.roundCtx
```
kostko, line 260: Access to `g.activeEpoch` requires the read lock to be held. Also what happens in case `g.activeEpoch` is `nil`?

go/worker/compute/executor/committee/node.go
```diff
@@ -511,7 +518,22 @@ func (n *Node) startProcessingBatchLocked(
 			batchRuntimeProcessingTime.With(n.getMetricLabels()).Observe(time.Since(rtStartTime).Seconds())
 		}()
 
-		rsp, err := rt.Call(ctx, rq)
+		roundCtx := n.commonNode.Group.RoundContext()
+		var rsp *protocol.Body
+		call := func() error {
+			var grr error
+			rsp, grr = rt.Call(ctx, rq)
```
kostko, line 525: Why did you add a retry here? This should not be needed.

The only methods that need retry are (as mentioned in #3000 with links):
* `HandleBatchFromTransactionSchedulerLocked` in executor.
* `HandleResultsFromExecutorWorkerLocked` in merge.

go/worker/common/committee/group.go
abukosek: Oops, will add the lock.  Where this is called from, the `activeEpoch` shouldn't be `nil`, though.

go/worker/compute/executor/committee/node.go
abukosek: `HandleBatchFromTransactionSchedulerLocked` calls `maybeStartProcessingBatchLocked`, which calls `startProcessingBatchLocked`, which can fail only in this `rt.Call`.  There doesn't seem to be any other place to put the retry but here.

kostko: Retrying here does not make sense as the `Call` internally includes a retry in case the runtime is not available. If it can't fail elsewhere, maybe retry is not needed for the transaction scheduler case then? I'll take another look.

kostko: Yeah it seems that `HandleBatchFromTransactionSchedulerLocked` does not need retry.

abukosek: OK, I'll remove it then :)

go/worker/common/committee/group.go
```diff
@@ -255,6 +255,16 @@ type Group struct {
 	logger *logging.Logger
 }
 
+// RoundContext returns the active epoch's round context.
+func (g *Group) RoundContext() (*context.Context, error) {
+	g.RLock()
+	defer g.RUnlock()
+	if g.activeEpoch == nil {
```
kostko, line 262: Empty line above.

go/worker/compute/merge/committee/node.go
```diff
@@ -76,6 +77,12 @@ var (
 	infiniteTimeout = time.Duration(math.MaxInt64)
 )
 
+// Parameters for retrying local submission calls.
+const (
+	retryInterval = 1 * time.Second
+	maxRetries    = 15
```
kostko, line 83: This should be [the same as in P2P](https://github.com/oasisprotocol/oasis-core/blob/cfc8b1a0d12e0e1fe425d987b6149643b844a015/go/worker/common/p2p/p2p.go#L194) which just uses `NewExponentialBackOff`. No need to limit the retries as the round context will be cancelled in case the round advances.

go/worker/common/committee/group.go
```diff
@@ -255,6 +255,16 @@ type Group struct {
 	logger *logging.Logger
 }
 
+// RoundContext returns the active epoch's round context.
+func (g *Group) RoundContext() (*context.Context, error) {
```
kostko, line 259: `context.Context` is an interface and so it can be nil, no need to return a pointer to it.

go/worker/compute/merge/committee/node.go
```diff
@@ -273,8 +280,28 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
+		// which shouldn't happen in our case.
+		n.logger.Warn("HandleResultsFromExecutorWorkerLocked called with no active epoch")
+		return
+	}
+
+	call := func() error {
+		grr := n.handleResultsLocked(*ctx, commit)
+		switch grr {
+		case errIncorrectState:
+			return backoff.Permanent(grr)
```
kostko, line 295: Why no retry in this case? The P2P code path retries even if the state is incorrect.

abukosek: Fixed :)

go/worker/compute/merge/committee/node.go
```diff
@@ -273,8 +274,21 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
 // HandleResultsFromExecutorWorkerLocked processes results from an executor worker.
 // Guarded by n.commonNode.CrossNode.
 func (n *Node) HandleResultsFromExecutorWorkerLocked(spanCtx opentracing.SpanContext, commit *commitment.ExecutorCommitment) {
-	// TODO: Context.
-	if err := n.handleResultsLocked(context.TODO(), commit); err != nil {
+	ctx, err := n.commonNode.Group.RoundContext()
```
kostko, line 277: Actually I just noticed that there already eixsts a `n.roundCtx` that you can use, so this is not needed.

abukosek: Haha, this PR is turning into a 2-line fix :D

go/worker/compute/merge/committee/node.go
```diff
@@ -273,8 +274,9 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
 // HandleResultsFromExecutorWorkerLocked processes results from an executor worker.
 // Guarded by n.commonNode.CrossNode.
 func (n *Node) HandleResultsFromExecutorWorkerLocked(spanCtx opentracing.SpanContext, commit *commitment.ExecutorCommitment) {
-	// TODO: Context.
-	if err := n.handleResultsLocked(context.TODO(), commit); err != nil {
+	call := func() error { return n.handleResultsLocked(n.roundCtx, commit) }
+	bkoff := backoff.WithContext(backoff.NewExponentialBackOff(), n.roundCtx)
+	if err := backoff.Retry(call, bkoff); err != nil {
```
kostko, line 279: This probably needs to be changed to perform retry in an async fashion, otherwise this can block all processing and the error condition will never resolve (as the lock is held all the time. So probably just spawn retry in a separate go routine and make sure to take the lock in the call function each time.

go/worker/compute/merge/committee/node.go
kostko: :-)

PR #3039, opened 2020-06-22
worker/sentry: Replace policy watcher with UpdatePolicies API method

### LINKED ISSUE #2820: sentry: Replace policy watcher with API command
Currently, the sentry gRPC worker (worker/sentry/grpc/worker.go) watches for policy changes, but that is unnecessary, since the upstream node can simply push the changes to the sentry using a new sentry API command (e.g. `UpdatePolicies`) in a similar way as was recently done with `SetUpstreamTLSCertificates` in PR #2675.

This would also enable simplification of the sentry gRPC worker code (e.g. not storing the connection dialed in `initConnection()` in the worker struct, removing other cruft, etc.).

abukosek:
Closes #2820.

Previously, the sentry gRPC worker watched for policy changes, but
now all policy changes are pushed to the sentry node via a new
sentry API method, `UpdatePolicies`.

This also gets rid of a lot of cruft in the sentry gRPC worker.


go/common/grpc/policy/policy.go
```diff
@@ -115,6 +115,22 @@ func (c *DynamicRuntimePolicyChecker) SetAccessPolicy(policy accessctl.Policy, r
 	}
 }
 
+// GetServicePolicies returns the service policies.
+func (c *DynamicRuntimePolicyChecker) GetServicePolicies() api.ServicePolicies {
+	c.RLock()
+	defer c.RUnlock()
+
+	policies := make(map[common.Namespace]accessctl.Policy, len(c.accessPolicies))
```
kostko, line 123: Add a comment (similar to above) why this needs to be a copy.

go/sentry/api/grpc.go
```diff
@@ -22,6 +24,12 @@ var (
 	// methodGetUpstreamTLSPubKeys is the GetUpstreamTLSPubKeys method.
 	methodGetUpstreamTLSPubKeys = serviceName.NewMethod("GetUpstreamTLSPubKeys", nil)
 
+	// methodUpdatePolicies is the UpdatePolicies method.
+	methodUpdatePolicies = serviceName.NewMethod("UpdatePolicies", grpcPolApi.ServicePolicies{})
+
+	// methodGetPolicyCheckers is the GetPolicyCheckers method.
+	methodGetPolicyCheckers = serviceName.NewMethod("GetPolicyCheckers", nil)
```
kostko, line 31: How can this be exposed via gRPC? It returns a `*policy.DynamicRuntimePolicyChecker` which is not serializable? This method should probably be removed from the gRPC API.

I assume this method is needed for local use on the sentry node itself? To make this more clear, the sentry `Backend` interface should also be split into two interfaces:
* `Backend` containing all the remotely accessible methods.
* `LocalBackend` that includes `Backend` but also additional methods like `GetPolicyCheckers`.

The gRPC `RegisterService` and `NewSentryClient` would then take/return a `Backend` while the sentry node would keep a `LocalBackend`.

go/sentry/sentry.go
```diff
@@ -82,6 +87,25 @@ func (b *backend) GetUpstreamTLSPubKeys(ctx context.Context) ([]signature.Public
...
+	}
+
+	return nil
+}
+
+func (b *backend) GetPolicyCheckers(ctx context.Context) (map[cmnGrpc.ServiceName]*policy.DynamicRuntimePolicyChecker, error) {
+	b.RLock()
+	defer b.RUnlock()
+
+	return b.grpcPolicyCheckers, nil
```
kostko, line 106: This returns a reference to the map so the above lock doesn't protect it. Making a copy of the whole map seems unnecessary, maybe just change the method to take a service name argument and only return a single policy checker (the `*policy.DynamicRuntimePolicyChecker` uses a mutex internally so sharing that should be fine)?

go/worker/keymanager/worker.go
```diff
@@ -592,6 +593,32 @@ func (crw *clientRuntimeWatcher) updateExternalServicePolicyLocked(snapshot *com
 
 	crw.w.grpcPolicy.SetAccessPolicy(policy, crw.node.Runtime.ID())
 	crw.w.logger.Debug("worker/keymanager: new normal runtime access policy in effect", "policy", policy)
+
+	// Notify the sentry nodes of the new policy.
+	for _, addr := range crw.w.commonWorker.GetConfig().SentryAddresses {
```
kostko, line 598: This seems like some common functionality, could it be handled in a single place instead of each piece of code creating a new connection to the sentry nodes?

It should work something like this:
* All the different places call `SetAccessPolicy` (as it is now, e.g. see above).
* There is a global `PolicyWatcher` that is used with all such policies. As far as I can see the policy watcher stuff is already done it should just be changed to no longer be exposed over gRPC and only used locally?
* The policy watcher is such that it will notify the sentry nodes over the single connection that it has. This connection should be shared with the registration worker as well as there is no sense in having different connections to the sentry workers from so many places.

go/worker/sentry/grpc/worker.go
```diff
@@ -153,67 +135,15 @@ func (g *Worker) authFunction() auth.AuthenticationFunction {
...
-			g.logger.Error("unable to dial upstream node",
-				"err", err,
-			)
-			return
-		}
-	}
-
-	// Initialize policy watcher.
-	g.policyWatcher = policyAPI.NewPolicyWatcherClient(g.conn)
-	ch, sub, err := g.policyWatcher.WatchPolicies(g.ctx)
```
kostko, line 195: Since there is no longer need for a policy watcher client (the `PolicyWatcherClient` interface), it and all uses should be removed. However the `PolicyWatcher` interface still seems useful to centralize policy updates (see my other comment).

go/worker/storage/committee/node.go
```diff
@@ -360,6 +363,32 @@ func (n *Node) updateExternalServicePolicyLocked(snapshot *committee.EpochSnapsh
 	// Update storage gRPC access policy for the current runtime.
 	n.grpcPolicy.SetAccessPolicy(policy, n.commonNode.Runtime.ID())
 	n.logger.Debug("set new storage gRPC access policy", "policy", policy)
+
+	// Notify the sentry nodes of the new policy.
+	for _, addr := range n.workerCommonCfg.SentryAddresses {
```
kostko, line 368: See above, with this centralized it shouldn't be needed at all as calling `SetAcceessPolicy` above will propagate the policy.

go/worker/keymanager/worker.go
abukosek: This is now done and ready for re-review :)

The policy watcher now handles pushing policy updates to the sentry nodes, however, I've left the registry worker as-is, since it only connects to the sentry nodes when rotating certificates and it needs to establish a new connection anyways in this case, since the certs have changed.

go/common/grpc/policy/watcher.go
```diff
@@ -2,35 +2,83 @@ package policy
```
kostko, line 2: This now seems like a bad place to put this -- previously it was a generic policy watcher that worked over gRPC, but now it is specific to sentry nodes. It should move somewhere under `go/sentry`, maybe even `go/sentry/client`?

go/common/grpc/policy/watcher.go
```diff
@@ -2,35 +2,83 @@ package policy
...
+			var err error
 
-func (c *policyWatcher) WatchPolicies(ctx context.Context) (<-chan api.ServicePolicies, pubsub.ClosableSubscription, error) {
-	typedCh := make(chan api.ServicePolicies)
-	sub := c.policyNotifier.Subscribe()
-	sub.Unwrap(typedCh)
+			if c.sentryClients[idx] != nil {
+				client = c.sentryClients[idx]
+			} else {
+				client, err = sentryClient.New(addr, c.id)
```
kostko, line 42: Is there any reason why we wouldn't initialize all the clients in the constructor and only reference them here instead of creating new connections each time?

go/common/grpc/policy/watcher.go
```diff
@@ -2,35 +2,83 @@ package policy
...
+	id            *identity.Identity
+	sentryClients []*sentryClient.Client
+
+	logger *logging.Logger
 }
 
 func (c *policyWatcher) PolicyUpdated(service grpc.ServiceName, accessPolicies map[common.Namespace]accessctl.Policy) {
-	c.policyNotifier.Broadcast(api.ServicePolicies{Service: service, AccessPolicies: accessPolicies})
-}
+	// Notify the sentry nodes of the new policy.
```
kostko, line 33: Can we make sure that this doesn't block the caller?

go/common/grpc/policy/watcher.go
```diff
@@ -2,35 +2,83 @@ package policy
...
+		}
+	}
 }
 
 // NewPolicyWatcher retruns a new policy watcher.
-func NewPolicyWatcher() api.PolicyWatcher {
+func NewPolicyWatcher(sentryAddrs []node.TLSAddress, id *identity.Identity) api.PolicyWatcher {
 	return &policyWatcher{
-		policyNotifier: pubsub.NewBroker(true),
+		ctx:           context.TODO(),
```
kostko, line 78: Accept context as argument?

go/common/grpc/policy/watcher.go
```diff
@@ -2,35 +2,83 @@ package policy
...
+		sentryAddrs:   sentryAddrs,
+		id:            id,
+		sentryClients: make([]*sentryClient.Client, len(sentryAddrs)),
+		logger:        logging.GetLogger("worker/policywatcher"),
```
kostko, line 82: Should probably be something like `sentry/policywatcher`.

go/sentry/localbackend/localbackend.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"context"
+
+	cmnGrpc "github.com/oasisprotocol/oasis-core/go/common/grpc"
+	"github.com/oasisprotocol/oasis-core/go/common/grpc/policy"
+	"github.com/oasisprotocol/oasis-core/go/sentry/api"
+)
+
+// LocalBackend is a local sentry backend implementation.
```
kostko, line 11: The interface should be part of the general API (e.g., `go/sentry/api`), why the move here?

go/worker/common/worker.go
```diff
@@ -276,8 +276,7 @@ func New(
 		return nil, err
 	}
 
-	grpcPolicyWatcher := policy.NewPolicyWatcher()
-	policyAPI.RegisterService(grpc.Server(), grpcPolicyWatcher)
+	grpcPolicyWatcher := policy.NewPolicyWatcher(cfg.SentryAddresses, identity)
```
kostko, line 279: Consider creating a context here that gets cancelled during `Stop()`.

go/common/grpc/policy/watcher.go
abukosek: Good idea, I'll move it.

go/common/grpc/policy/watcher.go
abukosek: Yes -- also, it's not creating new connections each time, they're cached.
Basically, this code does lazy connecting and also handles reconnecting if the connection is interrupted (e.g. when certs are rotated).  If we connected only in the constructor, then it would stop working after a while or it might not work at all in the beginning, since the sentry node might not be up yet.

go/common/grpc/policy/watcher.go
abukosek: OK, I'll move it to a goroutine.

go/sentry/localbackend/localbackend.go
abukosek: Because it caused import cycle hell and this was the cleanest way to solve it.  This might improve after moving the policy watcher to the sentry worker, will revisit this later.

go/common/grpc/policy/watcher.go
kostko: Make sure to add locks around shared state (e.g., `c.sentryClients`).

go/common/grpc/policy/watcher.go
kostko: I overlooked that, looks ok then.

go/common/grpc/policy/watcher.go
```diff
@@ -2,35 +2,83 @@ package policy
...
 )
 
 var _ api.PolicyWatcher = (*policyWatcher)(nil)
 
 type policyWatcher struct {
-	policyNotifier *pubsub.Broker
+	ctx context.Context
+
+	sentryAddrs   []node.TLSAddress
+	id            *identity.Identity
```
kostko, line 26: Call this `identity` to make it less confusing.

go/sentry/localbackend/localbackend.go
kostko: Ok, if nothing else it should at least be under `go/sentry/api/local`.

abukosek: OK, after moving the policy watcher, I was able to move this back to `go/sentry/api` :)

abukosek:
Comments addressed, ready for re-review :)

go/sentry/policywatcher/watcher.go
```diff
@@ -0,0 +1,93 @@
...
+					"err", err,
+					"sentry_address", addr,
+				)
+			}
+		}
+	}()
+}
+
+// NewPolicyWatcher retruns a new policy watcher.
+func NewPolicyWatcher(ctx context.Context, sentryAddrs []node.TLSAddress, id *identity.Identity) api.PolicyWatcher {
```
kostko, line 85: Since the package name is `policywatcher`, maybe just name this method `New` as it will be used like `policywatcher.New` anyway.

abukosek: Makes sense, fixed! :)

PR #3050, opened 2020-06-25
go/common/crypto/signature/signers/ledger: Descriptive error on user reject

tjanez:
Make Ledger signer return a more descriptive error message when a user rejects a transaction on the Ledger device.

For example, previously, the returned error was:
```
ts=2020-06-25T10:26:59.930888297Z level=error module=cmd/common/consensus caller=consensus.go:107 msg="failed to sign transaction" err="[APDU_CODE_COMMAND_NOT_ALLOWED] Command not allowed (no current EF)"
```

Now, it is:
```
ts=2020-06-25T10:27:24.693936203Z level=error module=cmd/common/consensus caller=consensus.go:107 msg="failed to sign transaction" err="transaction was rejected on Ledger device"
```

### DIFF
--- a/go/common/crypto/signature/signers/ledger/ledger_signer.go
+++ b/go/common/crypto/signature/signers/ledger/ledger_signer.go
@@ -2,8 +2,11 @@
 package ledger
 
 import (
+	"errors"
 	"fmt"
 	"io"
+
+	ledger "github.com/zondax/ledger-go"
 
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
 	ledgerCommon "github.com/oasisprotocol/oasis-core/go/common/ledger"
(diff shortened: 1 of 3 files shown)


go/oasis-node/cmd/common/consensus/consensus.go
```diff
@@ -104,6 +109,10 @@ func SignAndSaveTx(tx *transaction.Transaction) {
 
 	sigTx, err := transaction.Sign(signer, tx)
 	if err != nil {
+		if viper.GetString(cmdSigner.CfgSigner) == "ledger" && err.Error() == ledgerRejectTxErrorMsg {
```
kostko, line 112: This seems like the wrong place to put it? It should go into the implementation of the ledger signer.

tjanez: Agreed. I've moved it to the ledger signer implementation.

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -125,7 +132,15 @@ func (s *Signer) ContextSign(context signature.Context, message []byte) ([]byte,
 	if err != nil {
 		return nil, err
 	}
-	return s.device.SignEd25519(s.path, preparedContext, message)
+	signature, err := s.device.SignEd25519(s.path, preparedContext, message)
+	if err != nil {
+		if err.Error() == ledgerRejectTxErrorMsg {
```
kostko, line 137: Maybe use `errors.Is` instead. Also consider using a switch statement instead of these nested ifs.

tjanez: I've removed nested ifs and used a switch statement.

I couldn't use `errors.Is()` because at the moment, ledger-go doesn't use proper Go error semantics and doesn't expose errors as variables.

PR #3052, opened 2020-06-25
go/common/cbor: Add support for generic versioned blobs

Yawning:


.changelog/3052.internal.md
```diff
@@ -0,0 +1,4 @@
+go/common/cbor: Add support for generic versioned blobs
+
+Having to push, get a PR number, and then write these things is stupid
```
kostko, line 3: You can predict what number you'll get or you can open an issue so you have a known identifier.

Issue #3053, opened 2020-06-25
Improve term used for identifying connected Ledger device/app and document how it is derived


tjanez:
##### SUMMARY
Currently, listing connected Ledger devices with `oasis-node  signer ledger list_devices` outputs something like:
```
============ Device found
Oasis App Version : 312e312e32
Oasis App Address : oasis1jlnjum5rasu7hxxhayvf2yat5e32pz3ppwvhfv8hr969sjpuw9sgn54g9
```

The "Oasis App Address" is actually the Bech32-encoded public key (as of [Oasis Ledger App version 1.1.2](https://github.com/Zondax/ledger-oasis/releases/tag/v1.1.2)) of the account for Ledger's HD (hierarchical deterministic) wallet's [BIP44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) path for:

```
<mnemonic> / <purpose>' / <coin-type>' / <account>' / <change> / <address-index>
```

where `<address-index>` is set to `0`.

Once [Oasis Ledger App](https://github.com/Zondax/ledger-oasis/) is updated for the new [Staking Account Address](https://github.com/oasisprotocol/oasis-core/blob/master/docs/consensus/staking.md#accounts), this will be Bech32-encoded staking account address for the Ledger's HD wallet's account with address index of `0`.

We need to come up with a less confusing term for identifying the connected Ledger device/app and document how it is derived.

##### ISSUE TYPE
- Feature Idea

tjanez:
Implemented in https://github.com/oasisprotocol/oasis-core-ledger/pull/46.

PR #3058, opened 2020-06-26
Ledger validator app support

raynaudoe:
**On this PR**:
- Support for using Ledger Validator app for consensus role.

### DIFF
--- a/go/common/crypto/signature/signers/ledger/ledger_signer.go
+++ b/go/common/crypto/signature/signers/ledger/ledger_signer.go
@@ -29,12 +29,26 @@ var (
 	_ signature.SignerFactory     = (*Factory)(nil)
 	_ signature.Signer            = (*Signer)(nil)
 
-	// SignerDerivationRootPath is the derivation path prefix used for
-	// generating the signature key on the Ledger device.
-	SignerDerivationRootPath = []uint32{ledgerCommon.PathPurpose, SignerPathCoinType, SignerPathAccount, SignerPathChange}
+	// SignerEntityDerivationRootPath is the BIP-0032 path prefix used for generating
+	// an Entity signer.
+	SignerEntityDerivationRootPath = []uint32{
+		ledgerCommon.PathPurposeBIP44,
+		SignerPathCoinType,
+		SignerPathAccount,
+		SignerPathChange,
+	}
+	// SignerConsensusDerivationRootPath is the derivation path prefix used for
+	// generating a consensus signer.
+	SignerConsensusDerivationRootPath = []uint32{
+		ledgerCommon.PathPurposeOther,
+		SignerPathCoinType,
+		ledgerCommon.PathSubPurposeConsensus,
+		SignerPathAccount,
+	}
 
 	roleDerivationRootPaths = map[signature.SignerRole][]uint32{
-		signature.SignerEntity: SignerDerivationRootPath,
+		signature.SignerEntity:    SignerEntityDerivationRootPath,
+		signature.SignerConsensus: SignerConsensusDerivationRootPath,
 	}
 
 	// NOTE: The 0x6986 ISO 7816 error code is returned by the Oasis Ledger App
@@ -94,7 +108,7 @@ func (fac *Factory) Load(role signature.SignerRole) (signature.Signer, error) {
 	if !ok {
 		return nil, fmt.Errorf("role %d is not supported when using the Ledger backed signer", role)
 	}
-	device, err := ledgerCommon.ConnectToDevice(fac.address)
+	device, err := ledgerCommon.ConnectToDevice(fac.address, append(pathPrefix, fac.index))
 	if err != nil {
 		return nil, err
 	}
--- a/go/common/ledger/ledger.go
+++ b/go/common/ledger/ledger.go
@@ -6,8 +6,17 @@ import (
 )
 
 const (
-	// PathPurpose is set to 44 to indicate use of the BIP-0044 specification.
-	PathPurpose uint32 = 44
+	// PathPurposeBIP44 is set to 44 to indicate the use of the BIP-0044's hierarchy:
+	// https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki.
+	PathPurposeBIP44 uint32 = 44
+	// PathPurposeOther is set to 43 to indicate the use of (proposed)
+	// BIP-0043's extension that specifies the purpose for non-Bitcoin uses:
+	// https://github.com/bitcoin/bips/pull/523/files
+	PathPurposeOther uint32 = 43
+
+	// PathSubPurposeConsensus is set to 0 to indicate it is to be used for
+	// consensus-related things.
+	PathSubPurposeConsensus uint32 = 0
 
 	// ListingPathCoinType is set to 474, the index registered to Oasis in the SLIP-0044 registry.
 	ListingPathCoinType uint32 = 474
@@ -21,7 +30,9 @@ const (
 
 var (
 	// ListingDerivationPath is the path used to list and connect to devices by address.
-	ListingDerivationPath = []uint32{PathPurpose, ListingPathCoinType, ListingPathAccount, ListingPathChange, ListingPathIndex}
+	ListingDerivationPath = []uint32{
+		PathPurposeBIP44, ListingPathCoinType, ListingPathAccount, ListingPathChange, ListingPathIndex,
+	}
 )
 
 // Device is a Ledger device.
@@ -33,6 +44,6 @@ func ListDevices() {
 }
 
 // ConnectToDevice attempts to connect to a Ledger device by address, which is derived by ListingDerivationPath.
-func ConnectToDevice(address string) (*Device, error) {
-	return ledger.ConnectLedgerOasisApp(address, ListingDerivationPath)
+func ConnectToDevice(address string, derivationPath []uint32) (*Device, error) {
+	return ledger.ConnectLedgerOasisApp(address, derivationPath)
 }
--- a/go/go.mod
+++ b/go/go.mod
@@ -16,24 +16,17 @@ require (
 	github.com/blevesearch/bleve v1.0.9
 	github.com/btcsuite/btcutil v1.0.2
 	github.com/cenkalti/backoff/v4 v4.0.0
-	github.com/cznic/b v0.0.0-20181122101859-a26611c4d92d // indirect
-	github.com/cznic/mathutil v0.0.0-20181122101859-297441e03548 // indirect
-	github.com/cznic/strutil v0.0.0-20181122101858-275e90344537 // indirect
-	github.com/davidlazar/go-crypto v0.0.0-20200604182044-b73af7476f6c // indirect
 	github.com/dgraph-io/badger/v2 v2.0.3
 	github.com/eapache/channels v1.1.0
 	github.com/fxamacker/cbor/v2 v2.2.1-0.20200526031912-58b82b5bfc05
 	github.com/go-kit/kit v0.10.0
-	github.com/golang/protobuf v1.4.0
+	github.com/golang/protobuf v1.4.2
 	github.com/golang/snappy v0.0.1
 	github.com/grpc-ecosystem/go-grpc-middleware v1.0.1-0.20190118093823-f849b5445de4
 	github.com/hpcloud/tail v1.0.0
-	github.com/ipfs/go-log/v2 v2.1.1 // indirect
 	github.com/libp2p/go-libp2p v0.9.6
 	github.com/libp2p/go-libp2p-core v0.6.0
 	github.com/libp2p/go-libp2p-pubsub v0.3.1
-	github.com/libp2p/go-libp2p-swarm v0.2.7 // indirect
-	github.com/mr-tron/base58 v1.2.0 // indirect
 	github.com/multiformats/go-multiaddr v0.2.2
 	github.com/multiformats/go-multiaddr-net v0.1.5
 	github.com/oasisprotocol/deoxysii v0.0.0-20200527154044-851aec403956
@@ -42,9 +35,7 @@ require (
 	github.com/prometheus/client_golang v1.5.1
 	github.com/prometheus/common v0.9.1
 	github.com/prometheus/procfs v0.0.8
-	github.com/remyoudompheng/bigfft v0.0.0-20190512091148-babf20351dd7 // indirect
 	github.com/seccomp/libseccomp-golang v0.9.1
-	github.com/smartystreets/assertions v1.0.0 // indirect
 	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.7.0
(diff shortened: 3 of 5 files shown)


go/go.mod
```diff
@@ -24,16 +24,13 @@ require (
...
 	github.com/fxamacker/cbor/v2 v2.2.1-0.20200526031912-58b82b5bfc05
 	github.com/go-kit/kit v0.10.0
-	github.com/golang/protobuf v1.4.0
+	github.com/golang/protobuf v1.4.2
 	github.com/golang/snappy v0.0.1
 	github.com/grpc-ecosystem/go-grpc-middleware v1.0.1-0.20190118093823-f849b5445de4
 	github.com/hpcloud/tail v1.0.0
-	github.com/ipfs/go-log/v2 v2.1.1 // indirect
-	github.com/libp2p/go-libp2p v0.9.6
+	github.com/libp2p/go-libp2p v0.10.0
```
kostko, line 31: It seems like this contains unrelated dependency updates, could you only include changes relevant to this PR?

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -10,6 +10,7 @@ import (
 
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
 	ledgerCommon "github.com/oasisprotocol/oasis-core/go/common/ledger"
+	ledger "github.com/zondax/ledger-oasis-go"
```
kostko, line 13: This should be grouped in the above group of third-party imports. Also there already seems to be a `ledger` import above that points to `github.com/zondax/ledger-go` so you probably need to rename this one.

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -32,9 +33,13 @@ var (
 	// SignerDerivationRootPath is the derivation path prefix used for
 	// generating the signature key on the Ledger device.
 	SignerDerivationRootPath = []uint32{ledgerCommon.PathPurpose, SignerPathCoinType, SignerPathAccount, SignerPathChange}
+	// ConsensusDerivationRootPath is the derivation path prefix used for
+	// ledger Validator app.
+	ConsensusDerivationRootPath = []uint32{ledger.PathPurposeConsensus, SignerPathCoinType, SignerPathAccount, SignerPathChange}
```
kostko, line 38: Should this point to `ledgerCommon.PathPurposeConsensus`? Also the existing `ledgerCommon.PathPurpose` should be renamed to something like `ledgerCommon.PathPurposeEntity` (to signify that it is for use with the entity key while the consensus one is used with the consensus key).

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -94,7 +99,7 @@ func (fac *Factory) Load(role signature.SignerRole) (signature.Signer, error) {
 	if !ok {
 		return nil, fmt.Errorf("role %d is not supported when using the Ledger backed signer", role)
 	}
-	device, err := ledgerCommon.ConnectToDevice(fac.address)
+	device, err := ledgerCommon.ConnectToDevice(fac.address, append(pathPrefix, fac.index))
```
kostko, line 102: Why does this need to be changed, the path prefix in `ConnectToDevice` is only used to identify a specific device so it shouldn't matter which derivation path is used for that part? There was some mention of the need to refactor this (see #3053).

kostko:
Thanks for your contribution, added some comments.

go/common/crypto/signature/signers/ledger/ledger_signer.go
raynaudoe: Yeah, mi bad, I'll delete the `ledger-oasis-go` since is not necessary after doing the requested fix below.

go/common/crypto/signature/signers/ledger/ledger_signer.go
jleni: @kostko 
The standard (consumer) app and the validator app are completely different and independent apps.
`consumer: 44'/474'/.....`
`validator: 474'/474'/....`
based on the path we communicate to the correct app.
We could hide this behind some flag but I think it is better to keep ledger-oasis-go generic and let oasis-core to have the flexibility to use any path in the future?

kostko: Yeah I agree, but this specific path in `ConnectToDevice` is just for discovering the device, so it doesn't really matter which path is used here? Because the only reason why we use a path at all is that there is apparently no other reason to configure a specific ledger device to be used (e.g., if we used something like a device index, in case multiple devices are connected, this wouldn't be deterministic)?

raynaudoe: `Path` is actually needed to infer the `appMode` (Consumer/Validator). Without that info you can't connect to the ledger (at least directly without any 'guessing') because some info (like CLA) changes in relation of the `appMode`.
Another option would be having a way to list all available devices, print their app addresses and also print if that device has a validator app installed. Then pass that address with the flag `--signer.ledger.address `.

raynaudoe: `pathPrefix` is needed by `ledger-oasis-go` in order to know to which app (validator or consumer) should do the connection to.

go/common/crypto/signature/signers/ledger/ledger_signer.go
raynaudoe: Ok, fixing that!

go/go.mod
raynaudoe: Ok, should be cleaner now.

raynaudoe:
Ok, now both `go.mod` and `go.sum` are the same as in master branch.

go/common/ledger/ledger.go
```diff
@@ -6,9 +6,10 @@ import (
 )
 
 const (
-	// PathPurpose is set to 44 to indicate use of the BIP-0044 specification.
-	PathPurpose uint32 = 44
-
+	// PathPurposeEntity is set to 44 to indicate use of the BIP-0044 specification.
+	PathPurposeEntity uint32 = 44
+	// PathPurposeConsensus is set to 474 to be used with ledger Validator app
```
tjanez, line 11: Why is this set to `474`? Is there any unofficial agreement like [SLIP-0044](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) for registered coin types for [BIP44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki)?

kostko:
Cool, I see that you merged master into this branch. Please instead rebase on master and squash commits into one as these all pertain to the single feature.

go/common/ledger/ledger.go
```diff
@@ -21,7 +22,7 @@ const (
 
 var (
 	// ListingDerivationPath is the path used to list and connect to devices by address.
-	ListingDerivationPath = []uint32{PathPurpose, ListingPathCoinType, ListingPathAccount, ListingPathChange, ListingPathIndex}
+	ListingDerivationPath = []uint32{PathPurposeEntity, ListingPathCoinType, ListingPathAccount, ListingPathChange, ListingPathIndex}
```
tjanez, line 25: Since this hasn't changed, it means `oasis-node` will still list connected Ledger devices using the `<mnemonic> / 44' / 474' / 0' / 0 / 0` [BIP44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) path (reference: #3053).

raynaudoe: Yes, in order to also list Validator devices `ListDevices` should take into account a `ListingDerivationPath` with `PathPurposeConsensus`

go/common/ledger/ledger.go
jleni: no, we want to completely separate the validator app from anything coin related

tjanez: > No, we want to completely separate the validator app from anything coin related.

Agreed.

> We can choose any other number for purpose if really necessary, but it is important that the path does not collide with normal BIP44 paths.

I agree it is important that this path doesn't collide with any other BIP44 paths. That's why I asked if there is any unofficial community agreement on how to "register" a purpose.

I actually found one proposal to augment BIP43 by a core Bitcoin developer which makes sense: https://github.com/bitcoin/bips/pull/523/.

It proposes non-Bitcoin cryptocoins should use:
```
<mnemonic> / 43' / <coin-type>' / <subpurpose>' /
```

where `<coin-type>` is the non-Bitcoin's coin type as registered in [SLIP-0044](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) (`474` in our case) and `<subpurpose>` signifies the purpose, e.g. consensus, and would have a number that is unique in the Oasis' namespace. We free to choose whatever we like here.

jleni: well... basically, if you don't start with 44' then you are in the realm of BIP32 where the structure is actually free.

tjanez: If this was arbitrarily chosen, I would prefer following the proposal I mentioned above and use:
```
<mnemonic> / 43' / 474' / <subpurpose>' /
```
where `<subpurpose>` is e.g. `0` for consensus.

jleni: This break several parts of the code and integration libraries because we are already using 5 elements.

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -29,12 +29,16 @@ var (
 	_ signature.SignerFactory     = (*Factory)(nil)
 	_ signature.Signer            = (*Signer)(nil)
 
-	// SignerDerivationRootPath is the derivation path prefix used for
+	// SignerEntityDerivationRootPath is the derivation path prefix used for
```
tjanez, line 32:
```suggestion
	// SignerDerivationRootPath is the BIP-0032 path prefix used for generating
	// an Entity signer.
```

go/common/crypto/signature/signers/ledger/ledger_signer.go
```diff
@@ -29,12 +29,16 @@ var (
...
 	// generating the signature key on the Ledger device.
-	SignerDerivationRootPath = []uint32{ledgerCommon.PathPurpose, SignerPathCoinType, SignerPathAccount, SignerPathChange}
+	SignerEntityDerivationRootPath = []uint32{ledgerCommon.PathPurposeEntity, SignerPathCoinType, SignerPathAccount, SignerPathChange}
+	// SignerConsensusDerivationRootPath is the derivation path prefix used for
+	// ledger Validator app.
+	SignerConsensusDerivationRootPath = []uint32{ledgerCommon.PathPurposeConsensus, SignerPathCoinType, SignerPathAccount, SignerPathChange}
```
tjanez, line 37:
```suggestion
	SignerEntityDerivationRootPath = []uint32{
		ledgerCommon.PathPurposeBIP44,
		SignerPathCoinType,
		SignerPathAccount,
		SignerPathChange,
	}
	// SignerConsensusDerivationRootPath is the derivation path prefix used for
	// generating a consensus signer.
	SignerConsensusDerivationRootPath = []uint32{
		ledgerCommon.PathPurposeNonBitcoin,
		SignerPathCoinType,
		ledgerCommon.PathSubPurposeConsensus,
		SignerPathAccount,
	}
```

go/common/ledger/ledger.go
```diff
@@ -6,9 +6,10 @@ import (
...
 
 const (
-	// PathPurpose is set to 44 to indicate use of the BIP-0044 specification.
-	PathPurpose uint32 = 44
-
+	// PathPurposeEntity is set to 44 to indicate use of the BIP-0044 specification.
+	PathPurposeEntity uint32 = 44
+	// PathPurposeConsensus is set to 43 to be used with ledger Validator app
+	PathPurposeConsensus uint32 = 43
 	// ListingPathCoinType is set to 474, the index registered to Oasis in the SLIP-0044 registry.
```
tjanez, line 13: Then I would add the consensus subpurpose like I mentioned in https://github.com/oasisprotocol/oasis-core/pull/3058#discussion_r448258724:

```go
	// PathSubPurposeConsensus is set to 0 to indicate it is to be used for
	// consensus-related things.
	PathSubPurposeConsensus uint32 = 0
```

go/go.mod
```diff
@@ -57,8 +57,8 @@ require (
 	github.com/uber/jaeger-client-go v2.16.0+incompatible
 	github.com/uber/jaeger-lib v2.0.0+incompatible // indirect
 	github.com/whyrusleeping/go-logging v0.0.1
-	github.com/zondax/ledger-go v0.11.0
-	github.com/zondax/ledger-oasis-go v0.3.0
+	github.com/zondax/ledger-go v0.12.0
```
tjanez, line 60: Can you single out these package updates (go.mod and go.sum changes) to a separate commit?

go/common/ledger/ledger.go
tjanez: I don't have a strong preference on how many path elements are used, just that they are prefixed with `43' / 474' / <subpurpose>' / `.

I tried to keep the number of elements at 5 in my suggested changes: https://github.com/oasisprotocol/oasis-core/pull/3058#discussion_r449126499.

go/common/ledger/ledger.go
```diff
@@ -6,9 +6,10 @@ import (
 )
 
 const (
-	// PathPurpose is set to 44 to indicate use of the BIP-0044 specification.
-	PathPurpose uint32 = 44
-
+	// PathPurposeEntity is set to 44 to indicate use of the BIP-0044 specification.
+	PathPurposeEntity uint32 = 44
+	// PathPurposeConsensus is set to 43 to be used with ledger Validator app
+	PathPurposeConsensus uint32 = 43
```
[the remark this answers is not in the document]
jleni: IMO, making this change will add a lot of confusion to the codebase
I don't think bitcoin it even important here.. in this code base there is an specific meaning beyond the BIP.

go/go.mod
jleni: They were in a different commit and I was just asked to squash everything...

go/common/ledger/ledger.go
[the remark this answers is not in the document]
tjanez: Thanks for your comment. Agreed, the name is not the best. I've changed it to `PathPurposeOther`.

go/common/ledger/ledger.go
tjanez: Added.

go/go.sum
```diff
@@ -780,6 +783,7 @@ github.com/oasisprotocol/deoxysii v0.0.0-20200527154044-851aec403956 h1:etZXZf8f
 github.com/oasisprotocol/deoxysii v0.0.0-20200527154044-851aec403956/go.mod h1:cE5EgXTIhq5oAVdZ7LZd1FjTRLALPEzv93CWzBtDkyI=
 github.com/oasisprotocol/ed25519 v0.0.0-20200528083105-55566edd6df0 h1:qmiMZ6ZhkeQZkV/Huajj+QBAu1jX0HTGsOwi+eXTGY8=
 github.com/oasisprotocol/ed25519 v0.0.0-20200528083105-55566edd6df0/go.mod h1:IZbb50w3AB72BVobEF6qG93NNSrTw/V2QlboxqSu3Xw=
+github.com/oasisprotocol/oasis-core/go v0.0.0-20200623153002-9e61aea5195b/go.mod h1:/8dnlXMWVE88E6q4Dq/JoQyQhBfJzHtnebCmmPu2SLI=
```
kostko, line 786: I don't like this circular dependency as I'm not sure how this will interact with modules, but let's accept this for now and we'll fix it later.

PR #3062, opened 2020-06-29
go/tests/sentry: Add access control sanity checks

ptrus:


go/oasis-test-runner/oasis/oasis.go
```diff
@@ -734,10 +733,6 @@ func (net *Network) startOasisNode(
 	args = append(args, baseArgs...)
 	args = append(args, extraArgs.vec...)
 
-	if !strings.HasPrefix(node.Name, "sentry-") && !strings.HasPrefix(node.Name, "ias-proxy") && len(net.byzantine) == 0 {
```
ptrus, line 737: So i'm not sure if the `&& len(net.byzantine) == 0` flag was needed before, since in the new code i also enable cert rotation in byzantine tests and tests don't fail

go/oasis-test-runner/oasis/client.go
```diff
@@ -23,6 +23,7 @@ func (client *Client) startNode() error {
 	args := newArgBuilder().
 		debugDontBlameOasis().
 		debugAllowTestKeys().
+		workerCertificateRotation(true).
```
kostko, line 26: The client doesn't actually register, so this is not needed?

go/oasis-test-runner/scenario/e2e/runtime/sentry.go
```diff
@@ -130,3 +157,246 @@ func (s *sentryImpl) Fixture() (*oasis.NetworkFixture, error) {
...
+	}
+	opts := grpc.WithTransportCredentials(creds)
+	conn, err := cmnGrpc.Dial(address, opts) // nolint: staticcheck
+	if err != nil {
+		return nil, err
+	}
+	return conn, nil
+}
+
+func (s *sentryImpl) loadNodeIdentity(dataDir string) (*identity.Identity, error) {
```
kostko, line 188: Is this something that would be useful to have as a method on nodes?

go/oasis-test-runner/scenario/e2e/runtime/sentry.go
```diff
@@ -130,3 +157,246 @@ func (s *sentryImpl) Fixture() (*oasis.NetworkFixture, error) {
...
+	}
+
+	// Check Storage-1 endpoint with Storage-0 client certificates.
+	opts = &cmnGrpc.ClientOptions{
+		CommonName:    identity.CommonName,
+		ServerPubKeys: map[signature.PublicKey]bool{},
+		Certificates:  []tls.Certificate{*storage0Identity.GetTLSCertificate()},
+	}
+	// XXX: we need to skip client authentication of the server keys as storage-1 is using ephemeral
+	// tls keys which aren't known.
```
kostko, line 330: Actually, with the new `GetStatus` endpoint you can now fetch them by calling `GetStatus` on the target node and inspecting the retrieved `Identity` field (or even the node registration descriptor) in [the response](https://pkg.go.dev/github.com/oasisprotocol/oasis-core/go/control/api?tab=doc#Status).

go/oasis-test-runner/oasis/client.go
ptrus: This is currently consistent with previous code: https://github.com/oasisprotocol/oasis-core/pull/3062#discussion_r447062679 but yeah can remove it

go/oasis-test-runner/scenario/e2e/runtime/sentry.go
ptrus: Registration descriptor wouldn't work as i believe its sentry node key is used there, but using the `Identity` field should work yeah. Nice wasn't aware of the endpoint, in that case i also don't need the node with cert rotation disabled - but will probably keep the "refactor" in (moving the flag from oasis.go into node configs).

go/oasis-test-runner/scenario/e2e/runtime/sentry.go
ptrus: Indeed, moved it there

go/oasis-test-runner/scenario/e2e/runtime/sentry.go
ptrus: Fixed, thanks for the suggestion, no need for the hack with skipping client auth anymore.

.changelog/3062.feature.md
```diff
@@ -0,0 +1,6 @@
+go/control: List all valid publickeys in identity status
```
kostko, line 1:
```suggestion
go/control: List all valid TLS public keys in identity status
```

go/oasis-test-runner/scenario/e2e/runtime/sentry.go
```diff
@@ -130,3 +154,233 @@ func (s *sentryImpl) Fixture() (*oasis.NetworkFixture, error) {
...
+	storage1Client = storage.NewStorageClient(conn)
+	_, err = storage1Client.GetCheckpoints(ctx, emptyGetCheckpointsReq)
+	s.Logger.Debug("storage1.GetCheckpoints with storage0 cert", "err", err)
+	if err != nil {
+		return errors.New("storage1 checkpoints endpoint should allow connection with storage0 certificate")
+	}
+
+	// Sanity check Sentry-4 storage endpoint. All nodes are allowed to access
+	// storage sentry read only endpoints. Only active storage committee nodes
+	// are nodes are allowed to access checkpoint endpoints.
```
kostko, line 329:
```suggestion
	// are allowed to access checkpoint endpoints.
```

go/oasis-test-runner/scenario/e2e/runtime/sentry.go
```diff
@@ -130,3 +154,233 @@ func (s *sentryImpl) Fixture() (*oasis.NetworkFixture, error) {
...
+	sentry4Client := storage.NewStorageClient(conn)
+	_, err = sentry4Client.SyncGet(ctx, &storage.GetRequest{})
+	s.Logger.Debug("sentry4.SyncGet with validator0 sentry cert", "err", err)
+	// XXX: since we make an invalid request the request does fail, but ensure
+	// it makes past access control.
+	if status.Code(err) == codes.PermissionDenied {
+		return errors.New("sentry4 storage read-only endpoint should allow connections with validator0 certificate")
+	}
+	_, err = sentry4Client.GetCheckpoints(ctx, emptyGetCheckpointsReq)
+	s.Logger.Debug("sentry4.SyncGet with validator0 sentry cert", "err", err)
```
kostko, line 353:
```suggestion
	s.Logger.Debug("sentry4.GetCheckpoints with validator0 sentry cert", "err", err)
```

PR #3063, opened 2020-06-29
Runtime upgrade E2E test

### LINKED ISSUE #2520: Tests for runtime upgrades
We should specifically have E2E tests for performing (confidential) runtime upgrades. Also see #2517.

It should do something like the follows:
* First make sure an initial runtime is registered and operational.
* Build an upgraded runtime enclave (anything which changes MRENCLAVE should be fine).
* Update the key manager policy to inlcude the MRENCLAVE for the new runtime and wait for the policy to propagate.
* Shut down the compute nodes running the old runtime, replace the runtime and bring up compute nodes with the new runtime.
  * Could also be done so we have both old and new runtimes running at the same time and make sure that both old and new instances can access confidential data.
* Make sure everything works and the new runtime can access confidential data.
* Update the key manager policy to only include the new runtime.
* Make sure everything works and the new runtime can still access confidential data.

jberci:
Closes #2520


tests/runtimes/simple-keyvalue/Cargo.toml
```diff
@@ -4,6 +4,16 @@ version = "0.3.0-alpha"
 authors = ["Oasis Labs Inc. <info@oasislabs.com>"]
 edition = "2018"
 
+[[bin]]
+name = "simple-keyvalue"
```
ptrus, line 8: Nice! Can you similarly update the `simple-keymanager-upgrade` so we don't need two crates? (similar `println` change should be fine there as well)

jberci: Will do. The `println`s aren't actually needed, from what I've seen (the ones in here are just for debugging the PR).

go/oasis-test-runner/scenario/e2e/runtime/runtime_upgrade.go
```diff
@@ -0,0 +1,287 @@
...
+	sc.Logger.Info("waiting for new compute nodes to register",
+		"num_nodes", sc.Net.NumRegisterNodes(),
+	)
+
+	if err = sc.Net.Controller().WaitNodesRegistered(ctx, sc.Net.NumRegisterNodes()); err != nil {
+		return fmt.Errorf("failed to wait for nodes: %w", err)
+	}
+
+	sc.Logger.Info("wait for few epochs")
+	var waitEpoch epochtime.EpochTime
```
ptrus, line 252: I think instead of waiting for some epochs, this can now be replaced by waiting for the Compute workers to be ready (see the `WaitReady` method added in: https://github.com/oasisprotocol/oasis-core/pull/2957). Similarly in the `keymanager-runtime-upgrade` upgrade (which should wait for the newly spawned keymanager workers to be ready) - i think it will still need to wait for 1 more epoch transition after the nodes are ready.

tests/fixture-data/net-runner/default.json
```diff
@@ -194,7 +194,8 @@
                 "disable_check_tx": false,
                 "prune_num_kept": 0,
                 "tendermint_recover_corrupted_wal": false
-            }
+            },
+            "RuntimeOverride": null
```
kostko, line 198: See the serializer annotations comment.

tools/bin/cargo-elf2sgxs.rs
```diff
@@ -71,72 +71,80 @@ fn real_main() -> Result<()> {
...
-        let target_modified = target_meta.modified()?;
-
-        if target_modified > src_modified {
-            println!(
-                "{} {}",
-                Green.bold().paint(format!("{:>12}", "elf2sgxs")),
-                White.dimmed().paint("(skipped due to newer target file)"),
-            );
-            return Ok(());
+    target_path.push("bin");
```
kostko, line 82: Maybe call this `"<binary-name-placeholder>"` and add a comment because at first glance it looks like the binaries will go into a new `bin` directory (but this is in fact immediately pop-ed).

go/oasis-test-runner/scenario/e2e/runtime/runtime_upgrade.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+func (sc *runtimeUpgradeImpl) Fixture() (*oasis.NetworkFixture, error) {
+	f, err := sc.runtimeImpl.Fixture()
+	if err != nil {
+		return nil, err
+	}
+
+	// Get number of compute runtimes.
+	var computeIndex int
```
kostko, line 45: Use a negative sentry value and return an error in case there is no compute runtimes? Then you can also remove the duplicate kind check below.

go/oasis-test-runner/scenario/e2e/runtime/runtime_upgrade.go
```diff
@@ -0,0 +1,276 @@
...
+
+	// Load the upgraded runtime binary.
+	newRuntimeBinary, err := sc.resolveRuntimeBinary("simple-keyvalue-upgrade")
+	if err != nil {
+		return nil, fmt.Errorf("error resolving binary: %w", err)
+	}
+
+	// Setup the upgraded runtime (first is keymanager, others should be generic compute).
+	runtimeFix := f.Runtimes[computeIndex]
+	if runtimeFix.Kind != registry.KindCompute {
```
kostko, line 61: Duplicate kind check? See comment above.

go/oasis-test-runner/scenario/e2e/runtime/runtime_upgrade.go
```diff
@@ -0,0 +1,276 @@
...
+		if f.Runtimes[i].Kind == registry.KindCompute {
+			computeIndex = i
+			break
+		}
+	}
+
+	// Load the upgraded runtime binary.
+	newRuntimeBinary, err := sc.resolveRuntimeBinary("simple-keyvalue-upgrade")
+	if err != nil {
+		return nil, fmt.Errorf("error resolving binary: %w", err)
```
kostko, line 56:
```suggestion
		return nil, fmt.Errorf("error resolving upgraded binary: %w", err)
```

go/oasis-test-runner/scenario/e2e/runtime/runtime_upgrade.go
```diff
@@ -0,0 +1,276 @@
...
+	runtimeFix.Binaries = append([]string{newRuntimeBinary}, runtimeFix.Binaries...)
+
+	// The upgraded runtime will be registered later.
+	runtimeFix.ExcludeFromGenesis = true
+	newComputeIndex := len(f.Runtimes)
+	f.Runtimes = append(f.Runtimes, runtimeFix)
+
+	// Add the upgraded compute rutimes, will be started later.
+	workers := len(f.ComputeWorkers)
+	for i := range f.ComputeWorkers {
```
kostko, line 73: What about storage workers, those have runtime-specific provisions as well? I guess this doesn't matter as the runtime identifiers are duplicated, but would probably be cleaner to support a `Runtimes` option for storage workers as well?

go/oasis-test-runner/scenario/e2e/runtime/runtime_upgrade.go
```diff
@@ -0,0 +1,276 @@
...
+		f.ComputeWorkers[i].RuntimeOverride = []int{computeIndex}
+	}
+	for i := 0; i < workers; i++ {
+		f.ComputeWorkers = append(f.ComputeWorkers, oasis.ComputeWorkerFixture{Entity: 1, NoAutoStart: true, RuntimeOverride: []int{newComputeIndex}})
+	}
+
+	f.Network.IAS.UseRegistry = true
```
kostko, line 80: Add comment why this is needed.

go/oasis-test-runner/scenario/e2e/runtime/runtime_upgrade.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	// Generate and update a policy that will allow the new runtime to run.
+	if err = sc.applyUpgradePolicy(childEnv); err != nil {
+		return fmt.Errorf("updating policies: %w", err)
+	}
+
+	// Kill old compute workers.
+	sc.Logger.Info("stopping old runtimes")
+	for i := 0; i < computeWorkers/2; i++ {
```
kostko, line 210: Instead of using `computeWorkers / 2` would it be better to have an index where new workers start?

go/oasis-test-runner/scenario/e2e/runtime/runtime_upgrade.go
```diff
@@ -0,0 +1,276 @@
...
+	if err = cli.Consensus.SubmitTx(newTxPath); err != nil {
+		return fmt.Errorf("failed to update compute runtime: %w", err)
+	}
+
+	// Wait for the new nodes to register.
+	sc.Logger.Info("waiting for new compute nodes to register",
+		"num_nodes", sc.Net.NumRegisterNodes(),
+	)
+
+	if err = sc.Net.Controller().WaitNodesRegistered(ctx, sc.Net.NumRegisterNodes()); err != nil {
```
kostko, line 244: Below `WaitReady` includes registration so this is not needed.

go/oasis-test-runner/oasis/storage.go
```diff
@@ -115,7 +119,17 @@ func (worker *Storage) startNode() error {
 		workerStorageCheckpointCheckInterval(worker.checkpointCheckInterval).
 		appendNetwork(worker.net).
 		appendEntity(worker.entity)
-	for _, v := range worker.net.runtimes {
+	var runtimeArray []*Runtime
+	if len(worker.runtimes) > 0 {
+		worker.net.logger.Error("SKV runtime overrides present")
```
kostko, line 124: Was this for debugging? Should be removed (also below).

go/oasis-test-runner/scenario/e2e/runtime/runtime_upgrade.go
kostko: This was resolved, but it seems like the storage workers are still not updated here in the fixture?

PR #3065, opened 2020-06-29
go staking: increase reward amount denominator

pro-wh:
The denominator turned out not to be fine enough. This PR increases it. To keep the same reward rate, the numerators from the genesis block would need to be increased too. cc @bennetyee


pro-wh:
This new value `100_000_000` is chosen so that the genesis block can specify reward precisely enough to control about a hundredth of a percent APR for rewards that are delivered per-epoch.

Issue #2961, opened 2020-06-01
Registry query for nodes by consensus address


wjdfx:
In use of gRPC. In order to better process block data, methodGetBlock should include some info related to the proposer entity address or proposer nodes address.

Now I use this  method
methodGetBlock = serviceName.NewMethod("GetBlock", int64(0))
got the result :
```
{
 "version": {
    "block": 10,
    "app": 1638400
 },
 "chain_id": "41e79e2847b9188e6bf0246dc86a01a01cf11bbed6ca666b45",
 "height": 274755,
 "time": 1590845914,
 "last_block_id": {
    "hash": "Oylj7JUe36vHGbwvY9HFI0U1iZQj1VyNg/S7dnMkyQA=",
    "parts": {
       "hash": "yFZus9sNLqBq+14438QRZx1WuiKJc7Ug8pYphX0SGP4=",
       "total": 1
    }
 },
 "last_commit_hash": "0+MePs++luveOL9QG8j9ZLCR710YZ7GmFZ6QGvsJBf4=",
 "data_hash": null,
 "validators_hash": "U1+J1ANtw0wM61+xdMltO4ghx3w14H9JHGxZ9OSiFPg=",
 "next_validators_hash": "U1+J1ANtw0wM61+xdMltO4ghx3w14H9JHGxZ9OSiFPg=",
 "consensus_hash": "BICRvH3cKD93v7+R1zxE2ljD34qcvIZ0Bdi389qtoi8=",
 "app_hash": "VHEOn10lDB6HZ7rXOv8WAU3QzS0t7Hr6dZljhc5hHJU=",
 "last_results_hash": null,
 "evidence_hash": null,
 "proposer_address": "hOwR6glpRGOpwtaO4x6CAfgDyAg="
}
```

From the rusult, we can see  block proposer only has node's consensus public key: is "hOwR6glpRGOpwtaO4x6CAfgDyAg=", No entity address information, no node address information. There is no way to related them.

kostko:
The proposer address corresponds to (in case of the Tendermint consensus backend) a Tendermint address computed from the node's consensus key. This information is available in `.Consensus.ID` in the node descriptor.

So all the information is already there you only need an index from consensus addresses to node IDs. Such an index exists internally, but is currently not exposed and it may be a good idea to expose a way to query this information.

SmartStake:
[...]

PR #3067, opened 2020-06-30
go/registry/api: Add `GetNodeByConsensusAddress` method

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/2961

### DIFF
--- a/.changelog/2961.feature.md
+++ b/.changelog/2961.feature.md
@@ -1,0 +1,4 @@
+go/registry/api: Add `GetNodeByConsensusAddress` method
+
+`GetNodeByConsensusAddress` can be used to query nodes by their Consensus
+address.
--- a/go/consensus/tendermint/apps/registry/query.go
+++ b/go/consensus/tendermint/apps/registry/query.go
@@ -18,6 +18,7 @@ type Query interface {
 	Entity(context.Context, signature.PublicKey) (*entity.Entity, error)
 	Entities(context.Context) ([]*entity.Entity, error)
 	Node(context.Context, signature.PublicKey) (*node.Node, error)
+	NodeByConsensusAddress(context.Context, []byte) (*node.Node, error)
 	NodeStatus(context.Context, signature.PublicKey) (*registry.NodeStatus, error)
 	Nodes(context.Context) ([]*node.Node, error)
 	Runtime(context.Context, common.Namespace) (*registry.Runtime, error)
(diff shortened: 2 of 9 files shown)


.changelog/2961.feature.md
```diff
@@ -0,0 +1,7 @@
+go/registry/api: Add `GetNodeByConesnusID` method
```
kostko, line 1:
```suggestion
go/registry/api: Add `GetNodeByConsensusID` method
```

.changelog/2961.feature.md
```diff
@@ -0,0 +1,7 @@
...
+
+`GetNodeByConensusID` can be used to query nodes by their Consensus ID.
```
kostko, line 3:
```suggestion
`GetNodeByConsensusID` can be used to query nodes by their Consensus ID.
```

.changelog/3067.feature.md
```diff
@@ -0,0 +1,4 @@
+go/registry/api: Remove `GetNodeList` method
```
kostko, line 1: This should probably not be a feature but `breaking` or `internal`.

go/consensus/tendermint/apps/registry/query.go
```diff
@@ -71,6 +73,10 @@ func (rq *registryQuerier) Node(ctx context.Context, id signature.PublicKey) (*n
 	return node, nil
 }
 
+func (rq *registryQuerier) NodeByConsensusID(ctx context.Context, id signature.PublicKey) (*node.Node, error) {
+	return rq.state.NodeByConsensusAddress(ctx, []byte(tmcrypto.PublicKeyToTendermint(&id).Address()))
```
kostko, line 77: Slight problem with this is if you only have a "consensus address" (which is an internal Tendermint detail) and want to lookup the node -- it seems this is not possible but was one of the primary drivers for this feature (see #2961).

We could instead expose `NodeByConsensusAddress` that would take a `[]byte` argument and mention in the documentation comment that what a "consensus address" actually is depends on the consensus backend.

.changelog/2961.feature.md
```diff
@@ -0,0 +1,7 @@
...
+
+To query the node by tendermint consensus address use the
+`oasis-core/go/consensus/tendermint/crypto.PublicKeyFromTendermint`
+helper to first convert the tendermint address into a public key.
```
kostko, line 7: This is not actually correct, in case of Tendermint the address is a truncated hash of the public key, so going back is not possible.

ptrus: Ah yeah i missed this, thanks, fixed by instead exposing the `GetNodeByConsensusAddress` method!

go/registry/api/api.go
```diff
@@ -196,6 +196,10 @@ type Backend interface {
 	// GetNodes gets a list of all registered nodes.
 	GetNodes(context.Context, int64) ([]*node.Node, error)
 
+	// GetNodeByConsensusAddress returns the Node by Consensus address at the
+	// specified block height.
```
kostko, line 200:
```suggestion
	// GetNodeByConsensusAddress looks up a node by its consensus address at the
	// specified block height. The nature and format of the consensus address depends
	// on the specific consensus backend implementation used.
```

PR #3072, opened 2020-06-30
go/oasis-node: add consensus estimate_gas

### LINKED ISSUE #2723: Add CLI command for estimating gas cost of a transaction
##### SUMMARY
When a user creates a transaction using CLI, he needs to specify ``transaction.fee.gas``, i.e. the maximum amount of gas a transaction can spend, but he has to estimate the gas cost of a transaction manually.

We already support automatically estimating a transaction's gas cost:
https://github.com/oasislabs/oasis-core/blob/ffa12899c2fc44aed9848ab87d6709a8805b2bc3/go/consensus/api/api.go#L89-L90

We just need to expose it via CLI.

##### ISSUE TYPE
- Feature Idea

pro-wh:
fixes #2723 

> We just need to expose it via CLI.

in fact this is made even easier now that the command is further exposed over consensus ClientBackend

~~doesn't exactly live up to the original request though, because you'd still have to set a gas and sign it first before you would run the signed transaction through this file~~

### DIFF
--- a/go/oasis-node/cmd/consensus/consensus.go
+++ b/go/oasis-node/cmd/consensus/consensus.go
@@ -4,6 +4,7 @@ package consensus
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"os"
 
@@ -36,6 +37,12 @@ var (
 		Use:   "show_tx",
 		Short: "Show the content a pre-signed transaction",
 		Run:   doShowTx,
+	}
+
+	estimateGasCmd = &cobra.Command{
+		Use:   "estimate_gas",
+		Short: "Estimate how much gas a transactionw will use",
+		Run:   doEstimateGas,
 	}
 
 	logger = logging.GetLogger("cmd/consensus")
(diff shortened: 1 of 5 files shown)


pro-wh:
this looks like it would work, but the test fails with a signature verification error when trying to open the signed transaction. any thoughts on why that signature might not verify?

```
ts=2020-06-30T22:09:57.899708251Z level=info module=scenario/e2e/stake-cli caller=cli.go:79 msg="launching subcommand" binary=/home/wh0/work/ekiden/repo/go/oasis-node/oasis-node args="consensus estimate_gas --transaction.file /tmp/oasis-test-runner057607525/e2e/stake-cli/stake_transfer.json --address unix:/tmp/oasis-test-runner057607525/internal-505783007.sock --debug.allow_test_keys"
level=error module=test-runner caller=root.go:391 ts=2020-06-30T22:09:57.935110649Z msg="failed to run test case" err="root: failed to run test case: error while running Transfer test: failed to estimate gas: exit status 1 stderr [ts=2020-06-30T22:09:57.934306379Z level=error module=cmd/consensus caller=consensus.go:124 msg=\"failed to open signed transaction\" err=\"signed: signature verification failed\"\n]" test=e2e/stake-cli run_id=0
Error: root: failed to run test case: root: failed to run test case: error while running Transfer test: failed to estimate gas: exit status 1 stderr [ts=2020-06-30T22:09:57.934306379Z level=error module=cmd/consensus caller=consensus.go:124 msg="failed to open signed transaction" err="signed: signature verification failed"
]
```

go/oasis-node/cmd/consensus/consensus.go
```diff
@@ -103,11 +110,43 @@ func doShowTx(cmd *cobra.Command, args []string) {
 	sigTx.PrettyPrint("", os.Stdout)
 }
 
+func doEstimateGas(cmd *cobra.Command, args []string) {
+	if err := cmdCommon.Init(); err != nil {
+		cmdCommon.EarlyLogAndExit(err)
+	}
+
```
kostko, line 117: This is missing `cmdConsensus.InitGenesis()` which initializes the chain domain separation context so any signature verifications will fail. Adding it also means that you need to specify the path to the genesis file.

kostko:
> doesn't exactly live up to the original request though, because you'd still have to set a gas and sign it first before you would run the signed transaction through this file

Why not accept an unsigned transaction and allow specifying the signer public key to the estimate gas command?

pro-wh:
> Why not accept an unsigned transaction and allow specifying the signer public key as an argument to the estimate gas subcommand?

I didn't find any documentation on how to generate an unsigned transaction.

Skipping an unnecessary signing step would be strictly better. A problem would remain though. An unsigned transaction would still have to be created with a gas amount.

kostko:
> I didn't find any documentation on how to generate an unsigned transaction.

So there is definitely no CLI to do that, but we should just add an option like `--transaction.unsigned` to all transaction-generating subcommands. Transaction signing and serialization is currently performed here for all transactions:
https://github.com/oasisprotocol/oasis-core/blob/99e3a697874e5b3ea3790c026ac5795403286162/go/oasis-node/cmd/common/consensus/consensus.go#L88

So you could easily add this option in the common consensus CLI code and skip the signing step in case the option is passed.

> Skipping an unnecessary signing step would be strictly better. A problem would remain though. An unsigned transaction would still have to be created with a gas amount.

This is true, but it doesn't need to be anything in particular (e.g., it can be left as unspecified) because the simulation logic always sets the gas limit to the maximum possible value in order to perform estimation.

pro-wh:
thanks. let me see if I can add that `transaction.unsigned` idea as part of this PR

pro-wh:
ok added those. estimate_gas command now takes public key and unsigned transaction.

Issue #3047, opened 2020-06-24
Transaction status via RPC 


Messer4:
##### SUMMARY

When extracting data from the chain GetTransactions RPC doesnt return Tx status and some data aggregations based on failed txs can be false.
It would be awesome to have Tx Status included in Tx struct.
##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

GetTransactions RPC method

kostko:
Thanks for the feature suggestion, it seems like a generally useful thing to introduce something like a `GetTransactionResults` that takes a height and returns execution results for all transactions in a given height. Alternatively we could also have something like `GetTransactionsWithResults` to reduce the number of round trips required.

Messer4:
Second one could be more useful because reduce number of requests by RPC and increase speed of data processing

PR #3073, opened 2020-07-01
go/consensus/api: Add `GetTransactionsWithResults` method

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3047

### DIFF
--- a/go/consensus/api/api.go
+++ b/go/consensus/api/api.go
@@ -13,6 +13,7 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common/node"
 	"github.com/oasisprotocol/oasis-core/go/common/pubsub"
 	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"
+	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction/results"
 	epochtime "github.com/oasisprotocol/oasis-core/go/epochtime/api"
 	genesis "github.com/oasisprotocol/oasis-core/go/genesis/api"
 	keymanager "github.com/oasisprotocol/oasis-core/go/keymanager/api"
@@ -76,6 +77,11 @@ type ClientBackend interface {
 	//
 	// NOTE: Any of these transactions could be invalid.
 	GetTransactions(ctx context.Context, height int64) ([][]byte, error)
+
+	// GetTransactionsWithResults returns a list of transactions and their
+	// execution results, contained within a consensus block at a specific
+	// height.
+	GetTransactionsWithResults(ctx context.Context, height int64) (*TransactionsWithResults, error)
 
 	// WatchBlocks returns a channel that produces a stream of consensus
 	// blocks as they are being finalized.
@@ -243,4 +249,12 @@ type EstimateGasRequest struct {
 type GetSignerNonceRequest struct {
 	AccountAddress staking.Address `json:"account_address"`
 	Height         int64           `json:"height"`
+}
+
+// TransactionsWithResults is GetTransactionsWithResults response.
+//
+// Results[i] are the results of executing Transactions[i].
+type TransactionsWithResults struct {
+	Transactions [][]byte          `json:"transactions"`
+	Results      []*results.Result `json:"results"`
 }
--- a/go/consensus/api/transaction/results/results.go
+++ b/go/consensus/api/transaction/results/results.go
@@ -1,0 +1,34 @@
+package results
+
+import (
+	"github.com/oasisprotocol/oasis-core/go/common/errors"
+	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
+	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
+	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
+)
+
+// Event is a consensus service event that may be emitted during processing of
+// a transaction.
+type Event struct {
+	Staking  *staking.Event  `json:"staking,omitempty"`
+	Registry *registry.Event `json:"registry,omitempty"`
+	RootHash *roothash.Event `json:"roothash,omitempty"`
+}
+
+// Error is a transaction execution error.
+type Error struct {
+	Module  string `json:"module,omitempty"`
+	Code    uint32 `json:"code,omitempty"`
+	Message string `json:"message,omitempty"`
+}
+
+// Result is a transaction execution result.
+type Result struct {
+	Error  Error    `json:"error"`
+	Events []*Event `json:"events"`
+}
+
+// IsSuccess returns true if transaction execution was successful.
+func (r *Result) IsSuccess() bool {
+	return r.Error.Code == errors.CodeNoError
+}
--- a/go/consensus/tendermint/roothash/roothash.go
+++ b/go/consensus/tendermint/roothash/roothash.go
@@ -9,6 +9,8 @@ import (
 	"sync"
 
 	"github.com/eapache/channels"
+	"github.com/hashicorp/go-multierror"
+	tmabcitypes "github.com/tendermint/tendermint/abci/types"
 	tmrpctypes "github.com/tendermint/tendermint/rpc/core/types"
 	tmtypes "github.com/tendermint/tendermint/types"
 
@@ -19,7 +21,6 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
 	"github.com/oasisprotocol/oasis-core/go/common/pubsub"
 	consensus "github.com/oasisprotocol/oasis-core/go/consensus/api"
-	tmapi "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/api"
 	app "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/apps/roothash"
 	"github.com/oasisprotocol/oasis-core/go/consensus/tendermint/service"
 	"github.com/oasisprotocol/oasis-core/go/roothash/api"
@@ -144,24 +145,6 @@ func (tb *tendermintBackend) WatchBlocks(id common.Namespace) (<-chan *api.Annot
 	return monotonicCh, sub, nil
 }
 
-func (tb *tendermintBackend) getBlockFromFinalizedTag(ctx context.Context, rawValue []byte, height int64) (*block.Block, *app.ValueFinalized, error) {
-	var value app.ValueFinalized
-	if err := cbor.Unmarshal(rawValue, &value); err != nil {
-		return nil, nil, fmt.Errorf("roothash: corrupt finalized tag: %w", err)
-	}
-
-	block, err := tb.getLatestBlockAt(ctx, value.ID, height)
-	if err != nil {
-		return nil, nil, fmt.Errorf("roothash: failed to fetch block: %w", err)
-	}
-
-	if block.Header.Round != value.Round {
-		return nil, nil, fmt.Errorf("roothash: tag/query round mismatch (tag: %d, query: %d)", value.Round, block.Header.Round)
-	}
-
-	return block, &value, nil
-}
-
 func (tb *tendermintBackend) WatchAllBlocks() (<-chan *block.Block, *pubsub.Subscription) {
 	sub := tb.allBlockNotifier.Subscribe()
 	ch := make(chan *block.Block)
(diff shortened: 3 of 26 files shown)


go/consensus/api/api.go
```diff
@@ -244,3 +249,11 @@ type GetSignerNonceRequest struct {
 	AccountAddress staking.Address `json:"account_address"`
 	Height         int64           `json:"height"`
 }
+
+// TransactionsWithResults is GetTransactionsWithResults response.
+//
+// Results[i] are the results of executing Transactions[i].
+type TransactionsWithResults struct {
+	Transactions [][]byte `json:"transactions"`
+	Results      [][]byte `json:"results"`
```
kostko, line 258: I don't think we should expose raw, backend-dependent results as this seems like exposing internal details (e.g., things like Tendermint tags and their values used for events are internal details).

I think it would be better to define a consensus backend-agnostic results format and convert results to that format. The types mentioned below should live in the `transaction` package (`go/consensus/api/transaction`). For now it should have fields like:
```golang
type Result struct {
	Error Error `json:"error"`
}

func (r Result) IsSuccess() bool {
	return r.Error.Code == 0 // consider making this zero a constant in go/common/errors.go, also see line 43 there
}
```

Where `Error` is a type that supports our error encoding convention used in all the protocols (see `go/common/erorrs`), e.g.:
```golang
type Error struct {
	Module  string `json:"module,omitempty"`
	Code    uint32 `json:"code,omitempty"`
	Message string `json:"message,omitempty"`
}
```

We could also expose things like emitted events in addition to errors (e.g., by having an `Events []Event` field in `Result`), but this would require a union type like the following:
```golang
type Event struct {
	Staking  *staking.Event `json:"staking,omitempty"`
	Registry *registry.Event `json:"registry,omitempty"`
	RootHash *roothash.Event `json:"roothash,omitempty"`
}
```
And it would get populated by calling a lightweight version of `GetEvents` of individual services (it wouldn't be `GetEvents` directly as that fetches all transactions, it would require a slight refactor to allow processing of a single tx, e.g. similar to `onEventDataTx`).

ptrus: Thanks for the suggestions, please check the changes in`go/consensus/tendermint/staking/staking.go` before i continue impl. for other backends & clean things up.

go/consensus/api/transaction/results/results.go
```diff
@@ -0,0 +1,33 @@
+package results
+
+import (
+	"github.com/oasisprotocol/oasis-core/go/common/errors"
+	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
+	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
+	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
+)
+
+// Event is a transaction event.
```
kostko, line 10:
```suggestion
// Event is a consensus service event that may be emitted during processing of a transaction.
```

go/consensus/tendermint/staking/staking.go
```diff
@@ -242,35 +236,59 @@ func (tb *tendermintBackend) worker(ctx context.Context) {
...
+		if ev.Burn != nil {
+			tb.burnNotifier.Broadcast(&ev.Burn)
+		}
+		tb.eventNotifier.Broadcast(ev)
+	}
 }
 
-func (tb *tendermintBackend) onABCIEvents(ctx context.Context, tmEvents []tmapi.EventWithHash, height int64, doBroadcast bool) ([]api.Event, error) {
+// EventsFromTendermintEvents extracts staking events from tendermint events.
+func EventsFromTendermintEvents(
```
kostko, line 271:
```suggestion
func EventsFromTendermint(
```

go/consensus/tendermint/staking/staking.go
```diff
@@ -242,35 +236,59 @@ func (tb *tendermintBackend) worker(ctx context.Context) {
...
+func (tb *tendermintBackend) onEventDataTx(ctx context.Context, ev tmtypes.EventDataTx) {
+	events, errs := EventsFromTendermintEvents(ev.Tx, ev.Height, ev.Result.Events)
+	for _, err := range errs {
+		tb.logger.Error("error processing staking event", "err", err)
 	}
+	tb.notify(events)
+}
 
-	_, _ = tb.onABCIEvents(ctx, events, tx.Height, true)
+func (tb *tendermintBackend) notify(events []api.Event) {
```
kostko, line 255:
```suggestion
func (tb *tendermintBackend) notifyEvents(events []api.Event) {
```

go/consensus/tendermint/staking/staking.go
```diff
@@ -242,35 +236,59 @@ func (tb *tendermintBackend) worker(ctx context.Context) {
...
+	for _, ev := range events {
+		if ev.Transfer != nil {
```
kostko, line 257: I think we should consider removing these event-specific notifiers and Watch endpoints as they don't seem all that useful now that we have the general one. But that's probably for a different issue.

ptrus: Ok, i'll open an issue for it.

go/roothash/api/api.go
```diff
@@ -162,15 +162,23 @@ type ExecutionDiscrepancyDetectedEvent struct {
...
 }
 
+// FinalizedEvent is a finalized event.
+type FinalizedEvent struct {
+	Value []byte `json:"value"`
+}
+
 // Event is a roothash event.
 type Event struct {
+	RuntimeID common.Namespace `json:"runtime_id"`
```
ptrus, line 172: These are most prob. breaking changes (will add a fragment), are we fine with including the runtime ID in events? 

Also the added `FinalizedEvent` i could probably work around changing these two things if we don't like ti.

cc @kostko

go/oasis-node/cmd/debug/txsource/workload/queries.go
```diff
@@ -69,6 +72,87 @@ type queries struct {
 	runtimeGenesisRound uint64
 }
 
+func (q *queries) sanityCheckTransactionEvents(ctx context.Context, height int64, txEvents []*results.Event) error {
+	// Ensure transaction events match querying backend GetEvents methods.
+	// TODO: Registry events don't include any transaction info, so we cannot
```
ptrus, line 77: Should we add this (e.g. adding some tx info - maybe the transaction hash as we do for staking, to registry events as well)?

ptrus:
@kostko this is ready for more reviews, i posted 2 questions as comments (will also split into multiple commits & fragments after this round of reviews)

go/registry/api/api.go
```diff
@@ -303,12 +303,18 @@ type NodeUnfrozenEvent struct {
 	NodeID signature.PublicKey `json:"node_id"`
 }
 
+// NodeListEpochEvent is the per-epoch node list event.
```
kostko, line 306: I don't think we should expose this?

go/staking/api/api.go
```diff
@@ -150,6 +150,11 @@ type OwnerQuery struct {
 	Owner  Address `json:"owner"`
 }
 
+// TransactionEventsQuery is an transaction event query.
+type TransactionEventsQuery struct {
```
kostko, line 154: Where is this used?

go/oasis-node/cmd/debug/txsource/workload/queries.go
kostko: Yeah we should.

go/roothash/api/api.go
kostko: Yeah I think it makes sense to include RuntimeID as the `GetEvents` endpoint returns events for all runtimes. While we are here, can we update all the internal Tendermint events emitted by the roothash service to include a "runtime ID" attribute (e.g., `KeyRuntimeID`)? This would allow us to only subscribe to events for specific runtimes.

go/roothash/api/api.go
```diff
@@ -162,15 +162,23 @@ type ExecutionDiscrepancyDetectedEvent struct {
 type MergeDiscrepancyDetectedEvent struct {
 }
 
+// FinalizedEvent is a finalized event.
+type FinalizedEvent struct {
+	Value []byte `json:"value"`
```
kostko, line 167: Why is this an opaque value? Should this just have the round number?

go/consensus/tendermint/roothash/roothash.go
```diff
@@ -413,182 +422,163 @@ func (tb *tendermintBackend) onEventDataNewBlock(
...
 		for _, pair := range tmEv.GetAttributes() {
 			key := pair.GetKey()
 			val := pair.GetValue()
 
 			switch {
 			case bytes.Equal(key, app.KeyFinalized):
-				// A runtime block has been finalized.
-				if !doBroadcast {
-					// Ignore finalization event when querying events in GetEvents.
+				if !includeFinalized {
```
kostko, line 531: I don't think this flag is needed now that processing is separate? I think the only reason for it was to avoid triggering the special processing of finalized events. Otherwise I think it is fine to include finalization events in `GetEvents`.

ptrus: yeah indeed it was now only here to not include finalization events in `GetEvents`, will remove it

go/staking/api/api.go
ptrus: leftover, removed

go/registry/api/api.go
ptrus: agreed, I added an `InternalEvent` type for registry that includes the internal events that registry backend needs to process but aren't exposed (currently only such event type is the NodeListEpochEvent).

go/roothash/api/api.go
ptrus: Agreed, fixed

go/consensus/tendermint/roothash/roothash.go
```diff
@@ -413,182 +420,181 @@ func (tb *tendermintBackend) onEventDataNewBlock(
...
-					notifiers := tb.getRuntimeNotifiers(value.ID)
-					notifiers.eventNotifier.Broadcast(&ev)
-				} else {
-					events = append(events, ev)
-				}
+				ev := api.Event{RuntimeID: value.ID, Height: height, TxHash: txHash, MergeCommitted: &value.Event}
+				events = append(events, ev)
+			case bytes.Equal(key, app.KeyRuntimeID):
+				// Runtime ID attribute.
+				// TODO: Every event should contain this attribute, but not used atm
```
ptrus, line 588: See this comments please @kostko  (should we remove the `RuntimeID` field in each of the events, now that we have this attribute? or should we keep it unused for now, maybe just add a check for it to exist for every event?)

kostko: So the `RuntimeID` field would mostly be used for filtering events in queries (in this case the runtime ID that it contains as value should be text-serialized) not for populating it from anywhere else. So it is fine if it is unused during processing (could add a sanity check to make sure it is there though).

go/registry/api/api.go
```diff
@@ -305,12 +306,24 @@ type NodeUnfrozenEvent struct {
 
 // Event is a registry event returned via GetEvents.
 type Event struct {
+	TxHash hash.Hash `json:"tx_hash,omitempty"`
```
kostko, line 309: What about height?

go/registry/api/api.go
```diff
@@ -305,12 +306,24 @@ type NodeUnfrozenEvent struct {
...
 	NodeUnfrozenEvent *NodeUnfrozenEvent `json:"node_unfrozen,omitempty"`
 }
 
+// NodeListEpochInternalEvent is the per-epoch node list event.
+type NodeListEpochInternalEvent struct {
+	Height int64 `json:"height"`
+}
+
+// InternalEvent is a registry app internal event that is not exposed.
+type InternalEvent struct {
```
kostko, line 323: Why even expose this in the general API? If it is internal then it is backend-specific so it should just be defined inside the backend implementation and not exposed.

go/roothash/api/api.go
```diff
@@ -162,15 +162,22 @@ type ExecutionDiscrepancyDetectedEvent struct {
...
+// FinalizedEvent is a finalized event.
+type FinalizedEvent struct {
+	Round uint64 `json:"round"`
+}
+
 // Event is a roothash event.
 type Event struct {
-	Height int64     `json:"height,omitempty"`
-	TxHash hash.Hash `json:"tx_hash,omitempty"`
+	RuntimeID common.Namespace `json:"runtime_id"`
```
kostko, line 172: Maybe move the RuntimeID after the Height and TxHash fields to be consistent.

go/consensus/tendermint/roothash/roothash.go
```diff
@@ -413,182 +420,181 @@ func (tb *tendermintBackend) onEventDataNewBlock(
...
 	}
+	tb.processEvents(ctx, events, blockHistory)
+}
+
+func (tb *tendermintBackend) processEvents(ctx context.Context, events []api.Event, blockHistory map[common.Namespace]api.BlockHistory) {
+	for _, evp := range events {
+		// Redeclare variable on every iteration since the pointer to it gets broadcasted.
+		ev := evp
+
+		// Notify non-finalzed events.
```
kostko, line 449:
```suggestion
		// Notify non-finalized events.
```

go/registry/api/api.go
ptrus: added

go/registry/api/api.go
ptrus: yeah agreed & fixed

PR #3089, opened 2020-07-08
Add round sync metrics to storage worker

### LINKED ISSUE #3088: Add storage worker sync metrics
We should add metrics for the following storage worker metrics (prefix `oasis_worker_storage_`):
* last fully synced and finalized round
* last synced but not finalized round
* last to-be-synced round

jberci:
Closes #3088

### DIFF
--- a/go/worker/storage/committee/node.go
+++ b/go/worker/storage/committee/node.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 
 	"github.com/eapache/channels"
+	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/accessctl"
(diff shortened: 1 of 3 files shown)


go/worker/storage/committee/node.go
```diff
@@ -11,6 +11,8 @@ import (
 
 	"github.com/eapache/channels"
 
+	"github.com/prometheus/client_golang/prometheus"
```
kostko, line 14: Move to group above.

go/worker/storage/committee/node.go
```diff
@@ -40,6 +42,42 @@ var (
...
+		[]string{"round"},
+	)
+
+	storageWorkerCollectors = []prometheus.Collector{
+		storageWorkerLastFullRound,
+		storageWorkerLastSyncedRound,
+		storageWorkerLastPendingRound,
+	}
+
+	labelLastFullRound    = prometheus.Labels{"round": "last_full"}
```
kostko, line 76: Can you explain the role of these labels? Shouldn't you instead have labels for things like the runtime ID? See:
* https://github.com/oasisprotocol/oasis-core/blob/0fd92a9d9f0a980c68bc8392e59e1ea8aa80074a/go/worker/compute/executor/committee/node.go#L69
* https://github.com/oasisprotocol/oasis-core/blob/0fd92a9d9f0a980c68bc8392e59e1ea8aa80074a/go/worker/compute/executor/committee/node.go#L187-L191

PR #3094, opened 2020-07-09
go/consensus/api/transaction: Support multi-sig accounts

Yawning:
* [x] Add the v0 envelope format
 * [x] Use the v0 envelope for transactions
 * [x] Kludge the existing tooling so that at least accounts with only 1 signer work
 * [x] Unbreak entity node signing
 * [x] Make the node/entity/runtime descriptors use `cbor.Versioned`
 * [x] Make the tests pass
 * [x] Restore the stupid PrettyPrint crap
 * [x] Update transaction test vector generators (staking/registry `gen_vectors`)
 * ~Update the tooling to support multi-sig accounts~

Not going to fully update the tooling, since my approach to writing tools is "a jumbled mess of command line flags, that tries to be as user-hostile as possible", and apparently people actually are expected to use the things.


go/consensus/api/submission.go
```diff
@@ -51,6 +53,8 @@ type SubmissionManager interface {
 	//
 	// It also automatically handles retries in case the nonce was incorrectly estimated.
 	SignAndSubmitTx(ctx context.Context, signer signature.Signer, tx *transaction.Transaction) error
+
+	// TODO: Add SubmitSignedTx
```
kostko, line 57: Why would this be needed, if you have a signed tx and don't have the corresponding signers, you can just use the regular `SubmitTx` on the consensus backend?

go/common/node/node.go
```diff
@@ -53,9 +54,9 @@ type Node struct { // nolint: maligned
 	// ID is the public key identifying the node.
 	ID signature.PublicKey `json:"id"`
 
-	// EntityID is the public key identifying the Entity controlling
-	// the node.
-	EntityID signature.PublicKey `json:"entity_id"`
+	// EntityAddress is the account address of the entity controlling
+	// the runtime.
```
kostko, line 58:
```suggestion
	// the node.
```

go/consensus/api/transaction/v0/envelope.go
```diff
@@ -105,7 +112,7 @@ type Envelope struct {
 
 // Open opens an envelope, and returns the payload after verifying the
 // Account and signatures.
-func (e *Envelope) Open(dst interface{}) error {
+func (e *Envelope) Open(context signature.Context, dst interface{}) error {
```
kostko, line 115: If the multisig envelope is now something general (since it seems it supports different contexts as it is used for entities), should it be moved somewhere outside the consensus API (e.g., somewhere under common/crypto)?

go/consensus/tendermint/apps/registry/state/state.go
```diff
@@ -23,7 +25,7 @@ var (
 	// signedEntityKeyFmt is the key format used for signed entities.
 	//
 	// Value is CBOR-serialized signed entity.
-	signedEntityKeyFmt = keyformat.New(0x10, keyformat.H(&signature.PublicKey{}))
+	signedEntityKeyFmt = keyformat.New(0x10, keyformat.H(&staking.Address{}))
```
kostko, line 28: Since addresses are hashed, additional hashing via `H(...)` shoud not be needed. Also below.

This should also simplify some of the stuff below (e.g., the need for `hashAddr`).

go/consensus/tendermint/abci/mux.go
```diff
@@ -679,11 +680,14 @@ func (mux *abciMux) EstimateGas(caller signature.PublicKey, tx *transaction.Tran
...
 
-	ctx.SetTxSigner(caller)
+	ctx.SetTxAccount(caller)
 	mockSignedTx := transaction.SignedTransaction{
-		Signed: signature.Signed{
-			Blob: cbor.Marshal(tx),
-			// Signature is fixed-size, so we can leave it as default.
+		Envelope: v0.Envelope{
+			Account: *caller,
+			Signatures: []*signature.RawSignature{
```
kostko, line 687: So this always assumes a single signer, so EstimateGas will underestimate for multiple signatures. I guess we can fix this later.

go/consensus/api/transaction/v0/envelope.go
```diff
@@ -0,0 +1,201 @@
...
+
+// Verify validates an account descriptor for well-formedness.
+func (acc *Account) Verify() error {
+	if acc.V != 0 {
+		return fmt.Errorf("consensus/transaction/v0: invalid version: %v", acc.V)
+	}
+	if len(acc.Signers) == 0 {
+		return fmt.Errorf("consensus/transaction/v0: no account signers")
+	}
+
```
kostko, line 44: Should we add a maximum number of allowed signers? Should additional signers cost extra gas (beyond just the txes being larger) or is verifying extra signatures not really that more expensive since we're doing batch verification?

go/consensus/api/submission.go
Yawning: Sure, ok.

go/consensus/api/transaction/v0/envelope.go
Yawning: Yeah that's probably a good idea.  The construct itself is generic to a point.  There is a caveat that things that care about replay prevention (eg: actual transactions) need to include protection in the payload since it's trivial to re-submit an envelope with different numbers of signatures.

go/consensus/api/transaction/v0/envelope.go
Yawning: This "depends".  If the number of signatures is larger than the internal/unexposed minimum batch size, and all the signatures are correct, then it's fast.  Otherwise it validates the signatures serially.

This does mean the cases where this actually can be slow, there's no one to charge (any one signature invalidates the envelope), so this is probably ok for now.

go/consensus/tendermint/abci/mux.go
Yawning: It now assumes everyone signs the tx.

go/consensus/api/transaction/v0/envelope.go
Yawning: I'll move this once I'm not in the middle of working on another part of this.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -217,7 +216,17 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
...
 	// Charge gas for node registration if signed by entity. For node-signed
 	// registrations, the gas charges are pre-paid by the entity.
-	isEntitySigned := sigNode.MultiSigned.IsSignedBy(newNode.EntityID)
+	//
+	// Yes, the detection logic is awful.  See VerifyRegisterNodeArgs,
+	// where the same thing is done for a justification.
+	var isEntitySigned bool
+	for _, v := range sigNode.MultiSigned.Signatures {
+		addr := staking.NewAddress(multisig.NewAccountFromPublicKey(v.PublicKey))
+		if addr == untrustedNode.EntityAddress {
```
pro-wh, line 225: would `newNode.EntityAddress` have worked here?

Yawning: Sure.  There's essentially no difference, but I'll use newNode for consistency.

go/common/crypto/multisig/multisig_test.go
```diff
@@ -0,0 +1,197 @@
...
+
+	// Define a random test payload
+	type testPayload struct {
+		TestData string
+	}
+	data := testPayload{
+		TestData: "Hasta el suelo, Mi amigo linoleo, Linoleo",
+	}
+	payload := cbor.Marshal(data)
+	_ = payload
```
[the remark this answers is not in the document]
Yawning: Yeah.

tests/fixture-data/consim/genesis.json
```diff
@@ -1,328 +1 @@
...
...:"TEST","token_value_exponent":0,"total_supply":"200000000000","common_pool":"0","last_block_fees":"0","ledger":{"oasis1qz5lltqdla3n9qfl3qm6q82cgd8tweauxucvwlzp":{"general":{"balance":"100000000000"},"escrow":{"active":{"balance":"100000000000","total_shares":"1"},"debonding":{"balance":"0","total_shares":"0"},"commission_schedule":{},"stake_accumulator":{}}}},"delegations":{"oasis1qz5lltqdla3n9qfl3qm6q82cgd8tweauxucvwlzp":{"oasis1qz5lltqdla3n9qfl3qm6q82cgd8tweauxucvwlzp":{"shares":"1"}}}},"keymanager":{},"scheduler":{"params":{"min_validators":1,"max_validators":100,"max_validators_per_entity":1,"debug_bypass_stake":false,"debug_static_validators":false,"reward_factor_epoch_election_any":"0"}},"beacon":{"params":{"debug_deterministic":false}},"consensus":{"backend":"tendermint","params":{"timeout_commit":1000000000,"skip_timeout_commit":false,"empty_block_interval":0,"max_tx_size":32768,"max_block_size":22020096,"max_block_gas":0,"max_evidence_age_blocks":100000,"max_evidence_age_time":172800000000000,"state_checkpoint_interval":10000,"state_checkpoint_num_kept":2,"state_checkpoint_chunk_size":8388608,"gas_costs":{"tx_byte":1}}},"halt_epoch":18446744073709551615,"extra_data":null}
```
pro-wh, line 1: why did we have so many accounts before?

Yawning: Don't remember, it doesn't need them.

go/oasis-node/cmd/genesis/genesis.go
```diff
@@ -345,15 +345,8 @@ func AppendRegistryState(doc *genesis.Document, entities, runtimes, nodes []stri
 	if flags.DebugTestEntity() {
 		l.Warn("registering debug test entity")
 
-		ent, signer, err := entity.TestEntity()
-		if err != nil {
-			l.Error("failed to retrive test entity",
```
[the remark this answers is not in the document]
Yawning: There's a decent number of places where, `TestEntity` having the same signature as `Load` is used.  (I thought about doing this, but decided not to make a large branch even larger).

go/genesis/genesis_test.go
```diff
@@ -398,23 +409,23 @@ func TestGenesisSanityCheck(t *testing.T) {
...
 	d = *testDoc
 	te = *testEntity
 	te.Nodes = []signature.PublicKey{invalidPK}
-	signedBrokenEntity = signEntityOrDie(signer, &te)
+	signedBrokenEntity = signEntityOrDie(signer, validAccount, &te)
 	d.Registry.Entities = []*entity.SignedEntity{signedBrokenEntity}
 	require.Error(d.SanityCheck(), "test entity's invalid node public key should be rejected")
 
 	d = *testDoc
-	te = *testEntity
```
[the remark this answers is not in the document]
Yawning: What?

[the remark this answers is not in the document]
pro-wh: we used to reset te here, and now we don't. why don't we?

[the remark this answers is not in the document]
Yawning: Oh I see.  It doesn't make a difference, since signature validation happens super early.  I'll add it back in though.

go/consensus/api/transaction/transaction.go
```diff
@@ -148,64 +149,75 @@ type PrettyTransaction struct {
...
+}
+
 // PrettyPrint writes a pretty-printed representation of the type
 // to the given writer.
 func (s SignedTransaction) PrettyPrint(prefix string, w io.Writer) {
 	fmt.Fprintf(w, "%sHash: %s\n", prefix, s.Hash())
 
-	fmt.Fprintf(w, "%sSigner: %s\n", prefix, s.Signature.PublicKey)
-	fmt.Fprintf(w, "%s        (signature: %s)\n", prefix, s.Signature.Signature)
+	// TODO: Someone that cares probably can figure out a better way
```
kostko, line 199: Why wouldn't this just show the stringified staking address derived from the account?

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -217,7 +216,17 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
 
 	// Charge gas for node registration if signed by entity. For node-signed
 	// registrations, the gas charges are pre-paid by the entity.
-	isEntitySigned := sigNode.MultiSigned.IsSignedBy(newNode.EntityID)
+	//
+	// Yes, the detection logic is awful.  See VerifyRegisterNodeArgs,
```
kostko, line 220: Since this check seems suboptimal, could we avoid doing it in case we are not in debug mode (e.g., `DebugAllowEntitySignedNodeRegistration`)? My understanding is that `VerifyRegisterNodeArgs` will fail in case entity signing is used and that debug-only flag is false?

go/consensus/api/transaction/transaction.go
Yawning: From `go/staking/api.go`: `"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"`

go/consensus/tendermint/apps/registry/transactions.go
Yawning: I don't think it's particularly performance intensive or anything, it's just awful from a design standpoint...

Yawning: Fixed anyway.

Yawning:
> Any measurable performance impact from this change?

No idea, shouldn't be though.

> Some changes like using account addresses instead of public keys for entity identifiers seem useful for future proofness.

Moving to `cbor.Versioned` is also good here, but people will cry about entity regeneration if I wanted to extract that from this branch.

go/oasis-node/cmd/genesis/genesis.go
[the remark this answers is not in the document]
pro-wh: ah good point

Yawning:
Welp, we're not doing this (for real this time).  I'll close the PR but leave the branch lying around for a bit, so that I can salvage stuff out of it when people inevitably change their mind again in a few months.

pro-wh:
https://github.com/oasisprotocol/oasis-sdk/issues/91 hi from 2021

PR #3096, opened 2020-07-10
go: Use gofumpt instead of gofmt to format Go

### LINKED ISSUE #3095: Use gofumpt for stricter formatting
We should use [gofumpt](https://github.com/mvdan/gofumpt) and gofumports to enforce some stricter formatting rules. Some of them we already flag manually during reviews, but having an automated formatter seems better. This also requires bumping our golangci-lint version to at least 1.28.x.

kostko:
Fixes #3095

### DIFF
--- a/docker/development/Dockerfile
+++ b/docker/development/Dockerfile
@@ -4,8 +4,9 @@ FROM ubuntu:18.04
 ARG GO_VERSION=1.14.3
 ARG GO_PROTOC_VERSION=3.6.1
 ARG GO_PROTOC_GEN_GO_VERSION=1.21.0
-ARG GOLANGCILINT_VERSION=1.23.6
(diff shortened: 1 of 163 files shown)


docker/development/Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM ubuntu:18.04
 ARG GO_VERSION=1.14.3
 ARG GO_PROTOC_VERSION=3.6.1
 ARG GO_PROTOC_GEN_GO_VERSION=1.21.0
-ARG GOLANGCILINT_VERSION=1.23.6
+ARG GOLANGCILINT_VERSION=1.28.2
```
ptrus, line 7: Would be nice to also install `gofumpt` into the dev docker image?

kostko: Good idea, added.

Issue #3093, opened 2020-07-09
Node's runtime watcher should include suspended runtimes


ptrus:
Currently a node for a suspended runtime will never register itself as it will never see the suspended runtime.

Fix this by adding support for querying suspended runtimes to the registry API (add an `includeSuspended` flag to the `WatchRuntimes`/`GetRuntimes` methods) and update the runtime node watcher to also include the suspended runtimes.

PR #3099, opened 2020-07-10
go/registry: Add support for querying suspended runtimes

ptrus:
Fixes: #3093

TODO:
- [x] ~fix the runtime_dynamic test~ 
  - actually cannot reporduce this in the runtime dynamic test (or locally anymore), so it might have worked fine even before (even if the node fast-syncs it still get's all the events during fast syncing so it does get notified of the suspended runtimes). Nevertheless this change is still useful.

### DIFF
--- a/go/oasis-node/cmd/registry/runtime/runtime.go
+++ b/go/oasis-node/cmd/registry/runtime/runtime.go
@@ -86,13 +86,17 @@ const (
 	// Staking parameters flags.
 	CfgStakingThreshold = "runtime.staking.threshold"
 
+	// List runtimes flags.
+	CfgIncludeSuspended = "include_suspended"
+
 	runtimeGenesisFilename = "runtime_genesis.json"
 )
 
 var (
-	outputFlags   = flag.NewFlagSet("", flag.ContinueOnError)
-	runtimeFlags  = flag.NewFlagSet("", flag.ContinueOnError)
-	registerFlags = flag.NewFlagSet("", flag.ContinueOnError)
+	outputFlags      = flag.NewFlagSet("", flag.ContinueOnError)
+	runtimeFlags     = flag.NewFlagSet("", flag.ContinueOnError)
+	runtimeListFlags = flag.NewFlagSet("", flag.ContinueOnError)
+	registerFlags    = flag.NewFlagSet("", flag.ContinueOnError)
 
 	runtimeCmd = &cobra.Command{
 		Use:   "runtime",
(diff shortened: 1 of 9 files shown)


go/oasis-node/cmd/registry/runtime/runtime.go
```diff
@@ -30,6 +30,7 @@ import (
 	cmdFlags "github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common/flags"
 	cmdGrpc "github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common/grpc"
 	cmdSigner "github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common/signer"
+	"github.com/oasisprotocol/oasis-core/go/registry/api"
```
kostko, line 33: This is already imported on the next line as `registry`.

go/oasis-node/cmd/registry/runtime/runtime.go
```diff
@@ -86,13 +87,19 @@ const (
 	// Staking parameters flags.
 	CfgStakingThreshold = "runtime.staking.threshold"
 
+	// List runtimes flags.
+	// TODO: maybe automatically include suspended if -v (verbose) flag is set?
+	// and remove this flag.
+	CfgIncludeSuspended = "include-suspended"
```
kostko, line 93:
```suggestion
	CfgIncludeSuspended = "include_suspended"
```

go/registry/tests/tester.go
```diff
@@ -526,8 +526,13 @@ func testRegistryEntityNodes( // nolint: gocyclo
...
 	require := require.New(t)
 
-	existingRuntimes, err := backend.GetRuntimes(context.Background(), consensusAPI.HeightLatest)
-	require.NoError(err, "GetRuntimes")
+	query := &api.GetRuntimesQuery{Height: consensusAPI.HeightLatest, IncludeSuspended: false}
+	existingRuntimes, err := backend.GetRuntimes(context.Background(), query)
+	require.NoError(err, "GetRuntimes(includeSuspended=false)")
+	query.IncludeSuspended = true
+	existingAllRuntimes, err := backend.GetRuntimes(context.Background(), query)
+	require.NoError(err, "GetRuntimes(includeSuspended=false)")
```
kostko, line 534:
```suggestion
	require.NoError(err, "GetRuntimes(includeSuspended=true)")
```

PR #3102, opened 2020-07-11
go/consensus/tendermint: Use P2P key for Tendermint P2P

### LINKED ISSUE #3101: Use P2P key for Tendermint P2P connections
Currently the Tendermint consensus backend uses the node's identity key for P2P connections. It should use the node's P2P key instead.

kostko:
Fixes #3101 

Previously the Tendermint consensus backend used the node's identity key for
Tendermint P2P connections. This has now been changed to use the node's P2P
key instead as that can be made ephemeral in the future.


Yawning:
Isn't this breaking?

kostko:
Yeah the changelog fragment marks it as such.

PR #3106, opened 2020-07-13
go/tests/e2e: add missing Clone() override to late-start

ptrus:
Note: not overriding `Clone` in a scenario that embeds a different scenario actually causes embedded scenario test to be run (since `Clone` returns the embedded "parent" scenario).

Did check all other test scenarios and all do override the `Clone()` method.

### DIFF
--- a/.changelog/3106.internal.md
+++ b/.changelog/3106.internal.md
@@ -1,0 +1,1 @@
+go/tests/e2e: Add missing Clone() override to late-start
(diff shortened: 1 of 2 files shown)


Yawning:
Is there a way to prevent this from happening in the future (that isn't "remove the Clone() thing as a bad idea")?

ptrus:
> Is there a way to prevent this from happening in the future (that isn't "remove the Clone() thing as a bad idea")?

If we want to keep supporting the scenario flag combinations, we need it:

https://github.com/oasisprotocol/oasis-core/blob/c56a606134b58c11d8af8ea396ef98a333f53848/go/oasis-test-runner/cmd/root.go#L107-L110

I didn't implement this, also it's not for me to decide, anyway i think we can open a separate issue to discuss this.

Issue #3090, opened 2020-07-08
Monitor and report resource usage in daily longterm tests


ptrus:

kostko:
We do have support for that now with the added metrics, would just need to wire up the long term tests with some prometheus instance? We also have tooling for doing automated comparisons (e.g., #2536).

ptrus:
yeah exactly

PR #3109, opened 2020-07-14
longtests: setup prometheus monitoring

ptrus:
Fixes: #3090

### DIFF
--- a/.changelog/3090.internal.md
+++ b/.changelog/3090.internal.md
@@ -1,0 +1,1 @@
+ci/longtests: Setup prometheus monitoring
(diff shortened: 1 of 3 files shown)


.changelog/3090.internal.md
```diff
@@ -0,0 +1 @@
+longtests: Setup prometheus monitoring
```
kostko, line 1:
```suggestion
ci/longtests: Setup prometheus monitoring
```

ptrus:
> I assume the part that runs cmp will be separate?
> Also see the .buildkite/benchmarks.pipeline.yml.

yeah i'll look into fixing that pipeline (and setting up `cmp`) separately. This does get's us metrics at least that we can check/monitor manually.

PR #3111, opened 2020-07-16
Enable pretty-printers to print stake amounts in tokens

### LINKED ISSUE #3037: Display stake amount in token units in CLI command output
##### SUMMARY

To have a better UX when using the `oasis-node` CLI with a hardware wallet (e.g. Ledger) or on its own, it should also support displaying stake amounts in token unit (rather than base units used internally).

For example, `oasis stake info` currently displays something like:
```
Total supply: 10000000000000000000
Common pool: 7999217230119682890
Last block fees: 0
Staking threshold (entity): 100000000000
Staking threshold (node-validator): 100000000000
Staking threshold (node-compute): 100000000000
Staking threshold (node-storage): 100000000000
Staking threshold (node-keymanager): 100000000000
Staking threshold (runtime-compute): 100000000000
Staking threshold (runtime-keymanager): 100000000000
```

It could display something like:
```
Total supply: AMBER 10000000000
Common pool: AMBER 7999217230.11968289
Last block fees: 0
Staking threshold (entity): AMBER 100
Staking threshold (node-validator): AMBER 100
Staking threshold (node-compute): AMBER 100
Staking threshold (node-storage): AMBER 100
Staking threshold (node-keymanager): AMBER 100
Staking threshold (runtime-compute): AMBER 100
Staking threshold (runtime-keymanager): AMBER 100
```

where `AMBER` is the token's ticker symbol for the currently running [Oasis Amber Network](https://docs.oasis.dev/operators/amber-network.html).

We could support configuring both:
- amount of *base token* units one token unit is worth (e.g. 1000000000),
- token's ticker symbol (e.g. `AMBER`).

##### ISSUE TYPE
- Feature Idea

tjanez:
Closes #3037.

Example outputs of update `oasis-node stake` CLI commands:

```
[tadej@toronto staking-docs-network]$ oasis-node-dev stake info -a $ADDR
Token's ticker symbol: AMBER
Token's value base-10 exponent: 9
Total supply: AMBER 10000000000.0
Common pool: AMBER 7999217230.11968289
Last block fees: AMBER 0.0
Staking threshold (entity): AMBER 100.0
Staking threshold (node-validator): AMBER 100.0
Staking threshold (node-compute): AMBER 100.0
Staking threshold (node-storage): AMBER 100.0
Staking threshold (node-keymanager): AMBER 100.0
Staking threshold (runtime-compute): AMBER 100.0
Staking threshold (runtime-keymanager): AMBER 100.0
```

```
[tadej@toronto staking-docs-network]$ oasis-node-dev stake account info   -a $ADDR   --stake.account.address oasis1qp26ua0g6t30eus5mkxuv08fkds37jcqs5fs9tu7
General Account:
  Balance: AMBER 10000.000000001
  Nonce: 0
Escrow Account:
  Active:
    Balance: AMBER 0.0
    Total Shares: 0
  Debonding:
    Balance: AMBER 0.0
    Total Shares: 0
  Commission Schedule:
    Rates: (none)
    Rate Bounds: (none)
  Stake Accumulator:
    Claims: (none)
```

### DIFF
--- a/go/oasis-node/cmd/stake/stake.go
+++ b/go/oasis-node/cmd/stake/stake.go
@@ -70,6 +70,40 @@ func doConnect(cmd *cobra.Command) (*grpc.ClientConn, api.Backend) {
 	return conn, client
 }
 
+func getTokenSymbol(ctx context.Context, cmd *cobra.Command, client api.Backend) string {
+	symbol, err := client.TokenSymbol(ctx)
+	if err != nil {
+		logger.Error("failed to query token's symbol",
+			"err", err,
+		)
+		os.Exit(1)
+	}
+	return symbol
+}
+
+func getTokenValueExponent(ctx context.Context, cmd *cobra.Command, client api.Backend) uint8 {
+	exp, err := client.TokenValueExponent(ctx)
+	if err != nil {
+		logger.Error("failed to query token's value exponent",
+			"err", err,
+		)
+		os.Exit(1)
+	}
+	return exp
+}
+
+func getAccount(ctx context.Context, cmd *cobra.Command, addr api.Address, client api.Backend) *api.Account {
+	acct, err := client.Account(ctx, &api.OwnerQuery{Owner: addr, Height: consensus.HeightLatest})
+	if err != nil {
+		logger.Error("failed to query account",
+			"address", addr,
+			"err", err,
+		)
+		os.Exit(1)
+	}
+	return acct
+}
+
 func doInfo(cmd *cobra.Command, args []string) {
 	if err := cmdCommon.Init(); err != nil {
 		cmdCommon.EarlyLogAndExit(err)
@@ -80,6 +114,15 @@ func doInfo(cmd *cobra.Command, args []string) {
 
 	ctx := context.Background()
 
+	symbol := getTokenSymbol(ctx, cmd, client)
+	fmt.Printf("Token's ticker symbol: %s\n", symbol)
+
+	exp := getTokenValueExponent(ctx, cmd, client)
+	fmt.Printf("Token's value base-10 exponent: %d\n", exp)
+
+	ctx = context.WithValue(ctx, api.PrettyPrinterContextKeyTokenSymbol, symbol)
+	ctx = context.WithValue(ctx, api.PrettyPrinterContextKeyTokenValueExponent, exp)
+
 	totalSupply, err := client.TotalSupply(ctx, consensus.HeightLatest)
 	if err != nil {
 		logger.Error("failed to query total supply",
@@ -87,7 +130,9 @@ func doInfo(cmd *cobra.Command, args []string) {
 		)
 		os.Exit(1)
 	}
-	fmt.Printf("Total supply: %v\n", totalSupply)
+	fmt.Print("Total supply: ")
+	api.PrettyPrintAmount(ctx, *totalSupply, os.Stdout)
+	fmt.Println()
 
 	commonPool, err := client.CommonPool(ctx, consensus.HeightLatest)
 	if err != nil {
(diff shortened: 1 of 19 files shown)


go/oasis-node/cmd/stake/stake.go
```diff
@@ -90,6 +90,28 @@ func doWithRetries(cmd *cobra.Command, descr string, fn func() error) {
 	os.Exit(1)
 }
 
+func getTokenSymbol(ctx context.Context, cmd *cobra.Command, client api.Backend) string {
+	var symbol string
+	doWithRetries(cmd, "query token's symbol", func() error {
```
kostko, line 95: What's up with these retries all over the staking cmds, these should never be needed. Can we open an issue to get rid of this and not introduce more in new code?

go/oasis-test-runner/scenario/e2e/stake_cli.go
```diff
@@ -81,6 +82,8 @@ var (
 	// we can expect the amount of reclaimed base units to equal the number of reclaimed escrow
 	// shares.
 	reclaimEscrowAmount int64 = reclaimEscrowShares
+
+	ctx = context.Background()
```
kostko, line 86: Having a global like this sounds bad as `ctx` is commonly used and this makes it global in the whole `e2e` package.

go/oasis-test-runner/scenario/e2e/stake_cli.go
```diff
@@ -826,3 +828,9 @@ func (sc *stakeCLIImpl) genAmendCommissionScheduleTx(childEnv *env.Env, nonce in
 	}
 	return nil
 }
+
+func init() {
```
kostko, line 832: Why do this in global init, I would much rather have this per test case.

go/staking/api/api.go
```diff
@@ -88,6 +94,13 @@ var (
 		MethodAmendCommissionSchedule,
 	}
 
+	// ContextKeyTokenSymbol is the key to retrieve the token's ticker symbol
+	// value from a context.
+	ContextKeyTokenSymbol = contextKey("staking/token-symbol")
```
kostko, line 99: The `Context` prefix sounds too general as we have other kinds of contexts. Maybe explicitly name this `PrettyPrinterContext...`.

Also consider moving all this pretty printing stuff into its own `prettyprint.go` file.

go/oasis-node/cmd/stake/stake.go
tjanez: Old code fixed in #3113. I've also removed references in this PR.

go/oasis-test-runner/scenario/e2e/stake_cli.go
tjanez: Agreed. I've made it per test case now.

go/oasis-test-runner/scenario/e2e/stake_cli.go
tjanez: Agreed. I've made it per test case now.

go/staking/api/api.go
tjanez: > The `Context` prefix sounds too general as we have other kinds of contexts. Maybe explicitly name this `PrettyPrinterContext...`.

Yes, good point. I've renamed it.

> Also consider moving all this pretty printing stuff into its own `prettyprint.go` file.

Did you mean just the new `ConvertToTokenAmount()` and `PrettyPrintAmount()` functions?

For `PrettyPrint()` and `PrettyType()` methods, I would rather have them defined in the same place as the corresponding types.

If we would like to make the `go/staking/api/api.go` file smaller, I would rather break out logical pieces (e.g. events-related stuff, transaction-related stuff, threshold-related stuff, or similar).

kostko: I mean just the new functions.

tjanez: > I mean just the new functions (and context keys).

Aha, ok. I can move those to `prettyprint.go`.

tjanez: > > I mean just the new functions (and context keys).
> Aha, ok. I can move those to `prettyprint.go`.

Done.

PR #3120, opened 2020-07-21
go/common/crypto/signature: Add and use the plugin backed signer

Yawning:
Bloating the repository with a ton of different HSM (etc) signing backends doesn't make sense, and is a maintenance burden.  Use the runtime's built-in DLSO support to be able to separate out the non-esssential implementations.

 * [x] Add the plugin signer implementation
 * [x] Change oasis-node to instantiate the plugin signer.
 * [x] Add tests for the plugin signer.

After merge:
 * Move the ledger implementation off into a separate repository (https://github.com/oasisprotocol/oasis-core-ledger/issues/13)
 * Remove the integrated ledger signer implementation.
 * Remove the ledger oasis-node sub command.

Notes:
 * The plugin loader enforces (rather strictly) that the package versions of plugins are the same as what the loading application was built with.  This will require everything be built with `-trimpath`, but we should be doing that already anyway.  See https://github.com/golang/go/issues/27751 for more information.
 * This is undocumented because anyone that can't figure it out, has zero business writing anything that is even vaguely security critical.  That said there's tests and an example now.
 * golang-ci-lint got mad when linting the plugin because it doesn't have `main` defined, so it is skipped for now.

### DIFF
--- a/go/.golangci.yml
+++ b/go/.golangci.yml
@@ -24,3 +24,10 @@ linters:
     - varcheck
     - govet
     - maligned
+
+run:
+  skip-dirs:
+    # golang-ci-lint requires that files compile for certain linters
+    # to run, and Go plugins do not compile unless `-buildmode=plugin`
+    # is set, which linters do not do.
+    - oasis-test-runner/scenario/pluginsigner/example_signer_plugin
--- a/go/common/crypto/signature/signers/plugin/plugin.go
+++ b/go/common/crypto/signature/signers/plugin/plugin.go
@@ -1,0 +1,61 @@
+// Package plugin implements the Go plugin signature signer.
+package plugin
+
+import (
+	"fmt"
+	"plugin"
+
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
+)
+
+const (
+	// EntryPoint is the entry point (FactoryCtor) expected to be defined
+	// in each signer plugin.
+	EntryPoint = "GetPluginCtor"
+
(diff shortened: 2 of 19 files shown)


go/.golangci.yml
```diff
@@ -24,3 +24,7 @@ linters:
     - varcheck
     - govet
     - maligned
+
+run:
+  skip-dirs:
+    - oasis-test-runner/scenario/pluginsigner/example_signer_plugin
```
kostko, line 30: Maybe add a comment why this is skipped (as mentioned in the PR comment).

go/common/crypto/signature/signers/plugin/plugin.go
```diff
@@ -0,0 +1,56 @@
...
+const (
+	// EntryPoint is the entry point (FactoryCtor) expected to be defined
+	// in each signer plugin.
+	EntryPoint = "GetPluginCtor"
+
+	// SignerName is the name used to identify the plugin signer.
+	SignerName = "plugin"
+)
+
+type FactoryCtor func() signature.SignerFactoryCtor
```
kostko, line 20: Where is this used, is it there just for documentation purposes to show the type of the entry point? Maybe add a documentation comment to explain that.

Yawning: It was supposed to be actually used, but using it causes the plugin loader to error out because the runtime is too stupid to figure out that yes, the symbol does indeed match the function signature.  Added some comments.

PR #3124, opened 2020-07-22
go/e2e/txsource: periodically restart runtime nodes

ptrus:
TODO:
- [x] setup 2 keymanagers for the long-running test and also enable restarts there
- [x] setup 4 compute nodes
- [x] keep one of each type of nodes always running (so we can observe the memory/cpu/disk usage of that node, compared to the others)
- [x] revert the `scripts/daily_txsource.sh` change before merging

Fixes:
* .buildkite/longtests.pipeline.yml missing new artifact: extracted into: #3125
* go/consensus/tendermint/roothash: ignore non tracked runtimes when reindexing

### DIFF
--- a/.changelog/3124.bugfix.md
+++ b/.changelog/3124.bugfix.md
@@ -1,0 +1,1 @@
+go/consensus/tendermint/roothash: Ignore non-tracked runtimes when reindexing
--- a/go/consensus/tendermint/roothash/roothash.go
+++ b/go/consensus/tendermint/roothash/roothash.go
@@ -343,6 +343,10 @@ func (sc *serviceClient) reindexBlocks(currentHeight int64, bh api.BlockHistory)
 						return fmt.Errorf("failed to unmarshal finalized event: %w", err)
 					}
 
+					// Only process finalized events for tracked runtimes.
+					if sc.trackedRuntime[value.ID] == nil {
+						continue
+					}
 					if err = sc.processFinalizedEvent(sc.ctx, height, value.ID, &value.Round, false); err != nil {
 						return fmt.Errorf("failed to process finalized event: %w", err)
 					}
@@ -419,6 +423,10 @@ func (sc *serviceClient) DeliverEvent(ctx context.Context, height int64, tx tmty
 			continue
 		}
 
+		// Only process finalized events for tracked runtimes.
+		if sc.trackedRuntime[ev.RuntimeID] == nil {
+			continue
+		}
 		if err = sc.processFinalizedEvent(ctx, height, ev.RuntimeID, &ev.FinalizedEvent.Round, true); err != nil {
 			return fmt.Errorf("roothash: failed to process finalized event: %w", err)
 		}
@@ -435,6 +443,13 @@ func (sc *serviceClient) processFinalizedEvent(
 	reindex bool,
 ) (err error) {
 	tr := sc.trackedRuntime[runtimeID]
+	if tr == nil {
+		sc.logger.Error("runtime not tracked",
+			"runtime_id", runtimeID,
+			"tracked_runtimes", sc.trackedRuntime,
+		)
+		return fmt.Errorf("roothash: runtime not tracked: %s", runtimeID)
+	}
 	defer func() {
 		// If there was an error, flag the tracked runtime for reindex.
 		if err == nil {
--- a/go/oasis-node/cmd/debug/txsource/workload/runtime.go
+++ b/go/oasis-node/cmd/debug/txsource/workload/runtime.go
@@ -37,6 +37,8 @@ const (
 	runtimeGetExistingRatio = 0.9
 	// Ratio of remove requests that should delete an existing key.
 	runtimeRemoveExistingRatio = 0.5
+
+	runtimeRequestTimeout = 120 * time.Second
 )
 
 // RuntimeFlags are the runtime workload flags.
@@ -112,15 +114,14 @@ func (r *runtime) submitRuntimeRquest(ctx context.Context, rtc runtimeClient.Run
 		"request", req,
 	)
 
-	// Wait for a maximum of 30 seconds as invalid submissions may block
+	// Wait for a maximum of 'runtimeRequestTimeout' as invalid submissions may block
 	// forever.
-	submitCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	submitCtx, cancel := context.WithTimeout(ctx, runtimeRequestTimeout)
 	out, err := rtc.SubmitTx(submitCtx, rtx)
+	cancel()
 	if err != nil {
-		cancel()
 		return nil, fmt.Errorf("failed to submit runtime transaction: %w", err)
 	}
-	cancel()
 
 	if err = cbor.Unmarshal(out, &rsp); err != nil {
 		return nil, fmt.Errorf("malformed tx output from runtime: %w", err)
--- a/go/oasis-test-runner/scenario/e2e/runtime/txsource.go
+++ b/go/oasis-test-runner/scenario/e2e/runtime/txsource.go
@@ -148,11 +148,10 @@ func (sc *txSourceImpl) Fixture() (*oasis.NetworkFixture, error) {
 	f.Network.StakingGenesis = "tests/fixture-data/txsource/staking-genesis.json"
 
 	if sc.nodeRestartInterval > 0 {
-		// If node restarts enabled, do not enable round timeouts and merge
+		// If node restarts enabled, do not enable round timeouts and
 		// discrepancy log watchers.
 		f.Network.DefaultLogWatcherHandlerFactories = []log.WatcherHandlerFactory{
 			oasis.LogAssertNoRoundFailures(),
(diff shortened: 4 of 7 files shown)


go/consensus/tendermint/roothash/roothash.go
```diff
@@ -419,8 +426,15 @@ func (sc *serviceClient) DeliverEvent(ctx context.Context, height int64, tx tmty
 			continue
 		}
 
-		if err = sc.processFinalizedEvent(ctx, height, ev.RuntimeID, &ev.FinalizedEvent.Round, true); err != nil {
-			return fmt.Errorf("roothash: failed to process finalized event: %w", err)
+		if tr := sc.trackedRuntime[ev.RuntimeID]; tr != nil {
```
kostko, line 429: Same as above, consider using a continue statement and not emit a debug entry each time.

go/consensus/tendermint/roothash/roothash.go
```diff
@@ -435,6 +449,13 @@ func (sc *serviceClient) processFinalizedEvent(
 	reindex bool,
 ) (err error) {
 	tr := sc.trackedRuntime[runtimeID]
+	if tr == nil {
+		sc.logger.Error("runtime not tracked",
+			"runtime_id", runtimeID,
+			"trackedRuntimes", sc.trackedRuntime,
```
kostko, line 455:
```suggestion
			"tracked_runtimes", sc.trackedRuntime,
```
Also does including `sc.trackedRuntime` actually output a list or do you actually need to use something like `fmt.Sprintf("%+v", sc.trackedRuntime)`?

.changelog/3124.bugfix.md
```diff
@@ -0,0 +1 @@
+go/consensus/tendermint/rotohash: Ignore non tracked rutnimes when reindexing
```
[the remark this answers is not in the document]
kostko: (Also fix the commit message.)

kostko:
Nice catch!

.changelog/3124.bugfix.md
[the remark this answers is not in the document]
ptrus: ok, about time i setup spellcheck locally :-)

go/consensus/tendermint/roothash/roothash.go
```diff
@@ -342,9 +342,16 @@ func (sc *serviceClient) reindexBlocks(currentHeight int64, bh api.BlockHistory)
...
-
-					if err = sc.processFinalizedEvent(sc.ctx, height, value.ID, &value.Round, false); err != nil {
-						return fmt.Errorf("failed to process finalized event: %w", err)
+					// Only process finalized events for tracked runtimes.
+					if tr := sc.trackedRuntime[value.ID]; tr != nil {
+						if err = sc.processFinalizedEvent(sc.ctx, height, value.ID, &value.Round, false); err != nil {
+							return fmt.Errorf("failed to process finalized event: %w", err)
+						}
+					} else {
+						logger.Debug("runtime not tracked",
```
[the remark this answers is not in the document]
ptrus: yeah i will skip it :+1:

go/consensus/tendermint/roothash/roothash.go
ptrus: seems like it does `"tracked_runtimes":{"gAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=":{}`

go/consensus/tendermint/roothash/roothash.go
```diff
@@ -342,9 +342,16 @@ func (sc *serviceClient) reindexBlocks(currentHeight int64, bh api.BlockHistory)
 						)
 						return fmt.Errorf("failed to unmarshal finalized event: %w", err)
 					}
-
-					if err = sc.processFinalizedEvent(sc.ctx, height, value.ID, &value.Round, false); err != nil {
-						return fmt.Errorf("failed to process finalized event: %w", err)
+					// Only process finalized events for tracked runtimes.
+					if tr := sc.trackedRuntime[value.ID]; tr != nil {
```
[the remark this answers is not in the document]
ptrus: agreed :+1:

go/oasis-test-runner/scenario/e2e/runtime/txsource.go
```diff
@@ -148,11 +148,10 @@ func (sc *txSourceImpl) Fixture() (*oasis.NetworkFixture, error) {
 	f.Network.StakingGenesis = "tests/fixture-data/txsource/staking-genesis.json"
 
 	if sc.nodeRestartInterval > 0 {
-		// If node restarts enabled, do not enable round timeouts and merge
+		// If node restarts enabled, do not enable round timeouts and
```
ptrus, line 151: I'm not sure if `RoundFailures` are also expected to happen with runtime nodes getting restarted? (i guess will see after the test run i'm currently running will finish).

go/consensus/tendermint/roothash/roothash.go
kostko: Cool!

.changelog/3124.intenral.md
```diff
@@ -0,0 +1 @@
+go/e2e/txsource: Periodically restart runtime nodes
```
kostko, line 1: Fix filename (typo in "internal").

go/consensus/tendermint/roothash/roothash.go
```diff
@@ -342,7 +342,11 @@ func (sc *serviceClient) reindexBlocks(currentHeight int64, bh api.BlockHistory)
 						)
 						return fmt.Errorf("failed to unmarshal finalized event: %w", err)
 					}
-
+					tr := sc.trackedRuntime[value.ID]
```
kostko, line 345: `tr` doesn't seem to be used, why not have it inline in the `if` statement? Also below.

go/oasis-test-runner/scenario/e2e/runtime/txsource.go
kostko: This might have been there due to there not being enough validators (during a validator restart) which caused things to time out and fail? Maybe we fixed this with adding more validators, but I guess it can still happen in case we are unlucky and the proposer is the one being restarted (as this will cause things to fall over only after a timeout).

.changelog/3124.intenral.md
ptrus: :+1: would think the `lint fragments` check catch this :-)

go/oasis-test-runner/scenario/e2e/runtime/txsource.go
ptrus: so current change in this PR removes the `LogAssertNoExecutionDiscrepancyDetected()` - since this can now happen due to runtimes nodes being restarted.
 
And I am/was wondering if `RoundFailures` should also be removed, not sure if it can happen now?

ptrus: Based on the one full 6 hour run on this PR it looks like it doesn't happen, so the check can be kept?

kostko: Yeah let's leave it in for now and we'll see if it causes problems.

go/oasis-node/cmd/debug/txsource/workload/runtime.go
```diff
@@ -112,15 +113,22 @@ func (r *runtime) submitRuntimeRquest(ctx context.Context, rtc runtimeClient.Run
 		"request", req,
 	)
 
-	// Wait for a maximum of 30 seconds as invalid submissions may block
-	// forever.
-	submitCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
-	out, err := rtc.SubmitTx(submitCtx, rtx)
+	// Retry in case a runtime node is just getting restarted.
```
kostko, line 116: This is interesting, why does the workload need to handle retries? Shouldn't the client node handle this correctly as client nodes are not restarted? What exactly happens, does it just time out? If so, we may want to just increase the timeout value (to 120s?) instead of doing retries here. This is because retrying here could mask bugs in the client node not handling errors/retries correctly.

ptrus: hm, pretty sure it failed on the first compute node restart when i tried it the first time. iirc it timed out - didn't look to much into it - but i will re-check.

kostko: Maybe try with an increased timeout (e.g., 120s).

ptrus: yeah seems to work. will revert the retries.

ptrus:
hm got a: `failed to run scenario: log watcher compute-3/log: log assertion failed: timeout detected` in the short-txsource test on CI: https://buildkite.com/oasisprotocol/oasis-core-ci/builds/918#2cdbef61-4395-4526-894a-bc040620f727 (where we don't do any restarts, so only change are few more nodes).

kostko:
Based on a quick look at the logs, the timeout has to do with there being more storage nodes and one of them returning "permission denied" for some updates. Need to investigate this more.

PR #3127, opened 2020-07-23
 go/tendermint/roothash: Skip pruned heights when reindexing

ptrus:
* makes roothash reindexing aware of pruning
* adds a e2e/test scenario that triggers reindexing (with prunning)

TODO:
- [x] bunch of minor todo's/cleanups

### DIFF
--- a/go/consensus/tendermint/abci/prune.go
+++ b/go/consensus/tendermint/abci/prune.go
@@ -16,10 +16,13 @@ const (
 
 	pruneNone  = "none"
 	pruneKeepN = "keep_n"
+
+	// LogEventABCIPruneDelete is a log event value that signals an ABCI pruning
+	// delete event.
+	LogEventABCIPruneDelete = "tendermint/abci/prune"
 )
 
-// PruneStrategy is the strategy to use when pruning the ABCI mux iAVL
-// state.
+// PruneStrategy is the strategy to use when pruning the ABCI mux state.
 type PruneStrategy int
 
 const (
@@ -151,6 +154,7 @@ func (p *genericPruner) doPrune(ctx context.Context, latestVersion uint64) error
 		p.logger.Debug("Prune: Delete",
 			"latest_version", latestVersion,
 			"pruned_version", i,
+			logging.LogEvent, LogEventABCIPruneDelete,
 		)
 
 		err := p.ndb.Prune(ctx, i)
--- a/go/oasis-test-runner/oasis/compute.go
+++ b/go/oasis-test-runner/oasis/compute.go
@@ -2,8 +2,8 @@ package oasis
 
 import (
 	"fmt"
+	"sync"
 
-	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
 	storageClient "github.com/oasisprotocol/oasis-core/go/storage/client"
 	commonWorker "github.com/oasisprotocol/oasis-core/go/worker/common"
 )
@@ -19,6 +19,8 @@ const (
 
 // Compute is an Oasis compute node.
 type Compute struct { // nolint: maligned
+	sync.RWMutex
+
 	Node
 
 	entity *Entity
@@ -41,6 +43,13 @@ type ComputeCfg struct {
 	RuntimeProvisioner string
 
 	Runtimes []int
+}
+
+// UpdateRuntimes updates the worker node runtimes.
+func (worker *Compute) UpdateRuntimes(runtimes []int) {
+	worker.Lock()
+	defer worker.Unlock()
+	worker.runtimes = runtimes
 }
 
 // IdentityKeyPath returns the path to the node's identity key.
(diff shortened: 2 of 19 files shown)


go/consensus/tendermint/abci/prune.go
```diff
@@ -16,6 +16,9 @@ const (
 
 	pruneNone  = "none"
 	pruneKeepN = "keep_n"
+
+	// LogEventABCIPruneDelete is a log event value that singals an ABCI pruning delete event.
```
kostko, line 20:
```suggestion
	// LogEventABCIPruneDelete is a log event value that signals an ABCI pruning delete event.
```

go/consensus/tendermint/abci/prune.go
```diff
@@ -16,6 +16,9 @@ const (
...
+	LogEventABCIPruneDelete = "tendermint/abci/prune"
 )
 
 // PruneStrategy is the strategy to use when pruning the ABCI mux iAVL
```
kostko, line 24: While we're here, can we remove the "iAVL" bit as we no longer use it :-)

go/consensus/tendermint/abci/mux.go
```diff
@@ -268,6 +268,11 @@ func (a *ApplicationServer) State() api.ApplicationQueryState {
 	return a.mux.state
 }
 
+// GetLastRetainedVersion returns the earliest retained version.
+func (a *ApplicationServer) GetLastRetainedVersion(ctx context.Context) uint64 {
+	return a.mux.state.GetLastRetainedVersion()
```
kostko, line 273: Could we avoid this separate method by moving `GetLastRetainedVersion` to `api.ApplicationQueryState` (or `api.ApplicationState` and make `State()` return the latter)?

go/oasis-test-runner/oasis/compute.go
```diff
@@ -43,6 +43,11 @@ type ComputeCfg struct {
 	Runtimes []int
 }
 
+// UpdateRuntimes updates the worker node runtimes.
+func (worker *Compute) UpdateRuntimes(runtimes []int) {
+	worker.runtimes = runtimes
```
kostko, line 48: Do we need any locking here?

go/oasis-test-runner/oasis/compute.go
```diff
@@ -99,7 +104,9 @@ func (worker *Compute) startNode() error {
 	var runtimeArray []*Runtime
 	if len(worker.runtimes) > 0 {
 		for _, idx := range worker.runtimes {
-			runtimeArray = append(runtimeArray, worker.net.runtimes[idx])
+			if idx >= 0 {
```
kostko, line 107: What's the reason for having negative indices instead of just omitting those indices?

go/roothash/api/api.go
```diff
@@ -33,6 +33,9 @@ const (
 	LogEventRoundFailed = "roothash/round_failed"
 	// LogEventMessageUnsat is a log event value that signals a roothash message was not satisfactory.
 	LogEventMessageUnsat = "roothash/message_unsat"
+	// LogEventRoothashRuntimeReindexing is a log event value that singals a roothash runtime
+	// reindexing was run.
+	LogEventRoothashRuntimeReindexing = "roothash/runtime_reindexing"
```
kostko, line 38: Since this is already in the roothash api package, maybe just call it "LogEventHistoryReindexing"?
```suggestion
	LogEventHistoryReindexing = "roothash/history_reindexing"
```

go/oasis-test-runner/scenario/e2e/runtime/reindex.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis/cli"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario"
+	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
+)
+
+const (
+	reindexComputePruneNumKept = 50
+)
+
+// Reindex is the scenario that triggers reindexing.
```
kostko, line 20: Maybe call it "history reindex" (also update the filename in this case)?

go/oasis-test-runner/oasis/compute.go
ptrus: Since current logic is that if no runtimes are set, it will use all runtimes configured in the network cfg: https://github.com/oasisprotocol/oasis-core/blob/02a2481df9a8a93d3faa1df8a7f65978dd763bb9/go/oasis-test-runner/oasis/compute.go#L100-L106

Can look into how many test cases rely on that, and fix it up across scenarios, to not need the negative index hack.

go/oasis-test-runner/oasis/compute.go
ptrus: Technically yes, but practically probably not - doesn't hurt though so can add it.

go/oasis-test-runner/oasis/compute.go
ptrus: fixed

PR #3128, opened 2020-07-23
go: Remove the integrated ledger support

Yawning:
This has been moved to a separate repository.

Note: Yes this updates a bunch of dependencies.  The build got super mad when I messed with go.mod.


go/oasis-node/cmd/signer/signer.go
```diff
@@ -1,61 +0,0 @@
...
-	exportCmd = &cobra.Command{
-		Use:   "export",
-		Short: "export the public key from signer as an empty entity",
-		Run:   doExport,
-	}
-
-	logger = logging.GetLogger("cmd/signer")
-)
-
-func doExport(cmd *cobra.Command, args []string) {
```
kostko, line 30: Is this no longer needed or is this something that should be moved to the signer-specific cmd? Seems generic though.

Yawning: Oh whoops.  Forgot we had that.

PR #3130, opened 2020-07-27
Add GOVERNANCE.md

kostko:


kostko:
> Can we call this "Proposed Contribution Guidelines" instead of Governance? I think Governance has a different meaning that focuses a bit more on voting by the community and notes about how stake influences voting power, etc.

Sure, we can do that. I've updated the title to be "Project Contribution Guidelines". Should it still be located in a `GOVERNANCE.md` file (as mentioned [here](https://opensource.guide/leadership-and-governance/#how-do-i-formalize-these-leadership-roles))?

kostko:
Hm, the updated title does seem too similar with what is in [`CONTRIBUTING.md`](https://github.com/oasisprotocol/oasis-core/blob/master/CONTRIBUTING.md) which contains actual ways to contribute, not the governance process.

GOVERNANCE.md
```diff
@@ -0,0 +1,98 @@
...
+the majority of decisions to be made without resorting to a formal vote.
+
+[Architectural Decision Records (ADRs)]: docs/adr/index.md
+
+### ADRs and Lazy Consensus
+
+Decision making typically involves the following steps:
+
+* proposal,
+* discussionm
```
ptrus, line 79:
```suggestion
discussion,
```

GOVERNANCE.md
```diff
@@ -0,0 +1,98 @@
...
+* identifying requirements
+* providing graphics and web design
+* programming
+* assisting with project infrastructure
+* writing documentation
+* fixing bugs
+* adding features
+
+Contributors engage with the project through the issue tracker, writing or
+editing documentation, and participating in the community Slack channels. They
```
tjanez, line 34: To be consistent with section "Contribution Process"
```suggestion
editing documentation, and participating in the Oasis Community Slack channels. They
```
Also, convert `Oasis Community Slack channels` to a link.

GOVERNANCE.md
```diff
@@ -0,0 +1,98 @@
+# Project Contribution Guidelines
```
tjanez, line 1: Hmm... shouldn't this be "Governance" or something similar?

Contributing Guidelines is in `CONTRIBUTING.md`...

GOVERNANCE.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Overview
+
+This is a meritocratic, consensus-based community project. Anyone with an
+interest in the project can join the community, contribute to the project design
+and participate in the decision making process. This document describes how that
+participation takes place and how to set about earning merit within the project
+community.
+
```
tjanez, line 10: Was this Governance document inspired by some existing project's governance document?

If yes, it would be nice to tell about it here so that seasoned contributors have easier understanding and comparison to a governance document of an established project.

GOVERNANCE.md
kostko: it was like that before, see comment by @Afauvre and my reply above.

GOVERNANCE.md
kostko: Good idea, added to the top-level section, please take another look.

GOVERNANCE.md
kostko: Changed it back to "Governance Model".

GOVERNANCE.md
tjanez: Looks good, thanks!

GOVERNANCE.md
tjanez: Aha, I didn't read that before.

I think "Governance Model" and `GOVERNANCE.md` is fine since this governs how the Oasis Core project is run.

Of course, there is the _governance_ of the Oasis Network but that should be described in a different place, probably a document in https://github.com/oasisprotocol/docs.

GOVERNANCE.md
```diff
@@ -0,0 +1,109 @@
...
+have shown a willingness and ability to participate in the project as a team
+player. Typically, a potential committer will need to show that they have an
+understanding of the project, its objectives and its strategy. They will also
+have provided valuable contributions to the project over a period of time.
+
+## Contribution Process
+
+Anyone can contribute to the project, regardless of their skills, as there are
+many ways to contribute. The [Oasis Community Slack channels] are the best place
+to ask questions and the Oasis Protocol GitHub is the best place to see open
```
tjanez, line 67: Looking at the doc again, it would make sense to also make "Oasis Protocol GitHub" a link for consistency.

GOVERNANCE.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Overview
+
+This is a meritocratic, consensus-based community project. Anyone with an
+interest in the project can join the community, contribute to the project design
+and participate in the decision making process. This document describes how that
+participation takes place and how to set about earning merit within the project
+community.
+
+## Roles And Responsibilities
```
pro-wh, line 19:
```suggestion
## Roles and Responsibilities
```

GOVERNANCE.md
```diff
@@ -0,0 +1,111 @@
...
+have shown a willingness and ability to participate in the project as a team
+player. Typically, a potential committer will need to show that they have an
+understanding of the project, its objectives and its strategy. They will also
+have provided valuable contributions to the project over a period of time.
+
+## Contribution Process
+
+Anyone can contribute to the project, regardless of their skills, as there are
+many ways to contribute. The [Oasis Community Slack channels] are the best place
+to ask questions and the [Oasis Protocol GitHub] is the best place to see open
```
pro-wh, line 67:
```suggestion
to ask questions, and the [Oasis Protocol GitHub] is the best place to see open
```

PR #3135, opened 2020-07-28
Kill `tests/fixture-data` with fire

### LINKED ISSUE #3110: Kill `tests/fixture-data` with fire
JSON documents with account addresses that come from god only knows where is a monumental pain in the ass to maintain.

As cute as the idea is, we should probably just hard code the documents into the relevant test cases so we can do things like "Automatically derive the addresses from public keys", so that the next victim that has to update these things en-mass isn't stuck trying to figure out where they came from.

Yawning:
Fixes #3110

Note: There still is random mystery accounts of doom, but now they all live in a single go file in the e2e test sub-directory.  In an ideal world, these things will be derived properly, but to be honest, beyond the vague notion of "generated in a deterministic notion by something", I don't really have a good idea where they came from.


.buildkite/scripts/test_consim.sh
```diff
@@ -21,13 +21,25 @@ WORKDIR=$PWD
...
+CONSIM_GENESIS=/tmp/consim-genesis.json
+
+# Generate the consensus sim genesis document.
+${WORKDIR}/go/oasis-node/oasis-node \
+    genesis init \
+    --debug.test_entity \
+    --debug.dont_blame_oasis \
+    --chain.id test \
+    --debug.allow_test_keys \
+    --staking.token_symbol FUK \
```
kostko, line 33: Why not resurrect Buffy the BUF :grin:

go/oasis-node/cmd/common/genesis/staking.go
```diff
@@ -0,0 +1,145 @@
...
+// AppendableStakingState is a staking genesis state that can be appended
+// to an existing genesis document.
+type AppendableStakingState struct {
+	State           staking.Genesis
+	DebugTestEntity bool
+}
+
+// Append appends the staking genesis state to an existing genesis document.
+// Any pre-existing state will be overwritten.
+func (st *AppendableStakingState) Append(doc *genesis.Document) error {
```
kostko, line 27: Maybe it's just me, but it would be more clear if this was named `AppendTo`?

go/oasis-test-runner/oasis/runtime.go
```diff
@@ -205,3 +225,19 @@ func deriveMrEnclave(f string) (*sgx.MrEnclave, error) {
 
 	return &m, nil
 }
+
+func marshalWriteLog(wl storage.WriteLog) ([]byte, error) {
+	// Arrrrrgh.  mkvs/writelog.LogEntry has a custom UnmarshalJSON
+	// but no MarshalJSON equivalent.
```
kostko, line 231: Should we instead add it there?

go/oasis-test-runner/scenario/e2e/debond.go
```diff
@@ -37,7 +38,41 @@ func (s *debondImpl) Fixture() (*oasis.NetworkFixture, error) {
...
+	f.Network.StakingGenesis = &staking.Genesis{
+		Parameters: staking.ConsensusParameters{
+			CommissionScheduleRules: staking.CommissionScheduleRules{
+				RateChangeInterval: 10,
+				RateBoundLead:      30,
+				MaxRateSteps:       4,
+				MaxBoundSteps:      12,
+			},
+		},
+		TotalSupply: stakingTests.QtyFromInt(1000),
```
kostko, line 50: There's also that `quantity.NewFromUint64` that we could use instead of importing tests (and `QtyFromInt` could just be removed).

go/oasis-test-runner/scenario/e2e/runtime/runtime_dynamic.go
```diff
@@ -63,7 +78,12 @@ func (sc *runtimeDynamicImpl) Fixture() (*oasis.NetworkFixture, error) {
 	}
 	// Test storage genesis state for compute runtimes. Also test with a non-zero round.
 	f.Runtimes[1].GenesisRound = 42
-	f.Runtimes[1].GenesisState = "tests/fixture-data/runtime-dynamic/runtime-genesis-state.json"
+	f.Runtimes[1].GenesisState = storage.WriteLog{
+		{
+			Key:   mustUnbase64("Z2VuZXNpcyBzdGF0ZQ=="),
+			Value: mustUnbase64("aGVsbG8gd29ybGQ="),
```
kostko, line 84:
```suggestion
			Key:   []byte("genesis state"),
			Value: []byte("hello world"),
```
:grin:

kostko: Also, should make these strings constants as they are tested to be equal below on lines 263+.

go/oasis-test-runner/scenario/e2e/gas_fees_staking.go
```diff
@@ -63,9 +64,32 @@ func (sc *gasFeesImpl) Fixture() (*oasis.NetworkFixture, error) {
...
-			NodeBinary:              f.Network.NodeBinary,
-			EpochtimeMock:           true,
-			StakingGenesis:          "tests/fixture-data/gas-fees/staking-genesis.json",
+			NodeBinary:    f.Network.NodeBinary,
+			EpochtimeMock: true,
+			StakingGenesis: &staking.Genesis{
+				Parameters: staking.ConsensusParameters{
+					DebondingInterval: 1,
+					GasCosts: transaction.Costs{
+						"transfer":       10,
```
kostko, line 73: This should probably use the gas op constants defined in the staking api, e.g. this one would be `staking.GasOpTransfer`. Also elsewhere.

go/oasis-test-runner/oasis/fixture.go
```diff
@@ -191,9 +192,10 @@ type RuntimeFixture struct { // nolint: maligned
 	Entity     int                  `json:"entity"`
 	Keymanager int                  `json:"keymanager"`
 
-	Binaries     []string `json:"binaries"`
-	GenesisState string   `json:"genesis_state"`
-	GenesisRound uint64   `json:"genesis_round"`
+	Binaries         []string         `json:"binaries"`
+	GenesisState     storage.WriteLog `json:"genesis_state,omitempty"`
+	GenesisStatePath string           `json:"genesis_state_path"`
```
kostko, line 197: Maybe just omitempty this one and the round below?

go/oasis-test-runner/scenario/e2e/test_accounts.go
```diff
@@ -0,0 +1,33 @@
+package e2e
+
+import stakingTests "github.com/oasisprotocol/oasis-core/go/staking/tests/debug"
+
+// The people that made all the test staking genesis documents neglected
+// to really add comments where these accounts came from.  Presumably they
+// are deterministically generated identities of various things.  Someone
```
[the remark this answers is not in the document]
Yawning: #3141

PR #3140, opened 2020-07-29
Add consensus backend mode

### LINKED ISSUE #3116: Introduce the notion of consensus backend mode
The consensus backend should have a notion of a "mode" (configured via the `--consensus.tendermint.mode` flag) which should support the following options:
* `full` runs a regular full consensus node (default),
* (once implemented) `light` runs a light client consensus node (see #2440),
* `seed` runs a seed-only consensus node (replaces the `--tendermint.p2p.seed_mode`).

kostko:
Fixes #3116


go/consensus/tendermint/full/full.go
```diff
@@ -0,0 +1,1466 @@
+// Package full implements a full Tendermint consensus node.
```
kostko, line 1: :information_source: This is just moved from `../tendermint.go`.

PR #3142, opened 2020-07-29
go/badger: bump badger version

ptrus:
TODO:
- [x] run the longtests and comapre cpu/memory/iops usage
  - looks much better during compaction & also fixes cpu spikes at node restarts (there was a bug with badger vlog replay that replayed all vlogs) 
- [x] wait for the 72hours test to finish

### DIFF
--- a/.changelog/3142.internal.md
+++ b/.changelog/3142.internal.md
@@ -1,0 +1,1 @@
+go/badger: bump badger version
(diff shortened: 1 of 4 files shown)


kostko:
Looking at the failure, this may require disabling conflict detection by setting the `WithDetectConflicts(false)` (previously this was implied by managed mode, but now it is not).

ptrus:
> Looking at the failure, this may require disabling conflict detection by setting the WithDetectConflicts(false) (new option, previously this was implied by managed mode, but now it is not).

That fixed the failure, thanks!

kostko:
Just noticed that Badger master has some additional fixes in comparison to rc3.

ptrus:
Ok, bumped badger in this PR to use that commit as well.

PR #3143, opened 2020-07-30
go/oasis-node: Fix the interactive prompt to be correct

Yawning:
* Implicitly set `--assume_yes` if not running in a tty
 * Don't crash when not running in a tty


.changelog/9000.internal.md
```diff
@@ -0,0 +1,18 @@
+fixme: add a creative title
```
kostko, line 1: Could just add an empty trivial file as it fixes something introduced in this release.

Yawning: Yeah I just have that there so I don't get annoyed by ci-lint when I'm still working on a branch.

PR #3148, opened 2020-07-30
go/common/crypto/signature/signers/plugin: Rewrite

Yawning:
The Go runtime library's `plugin` is a steaming pile of shit.  Switch
to something that lets people write plugins that live in a separate
module, that import sub-packages from the module that is loading the
plugin.

### DIFF
--- a/go/common/crypto/signature/signers/plugin/plugin.go
+++ b/go/common/crypto/signature/signers/plugin/plugin.go
@@ -3,31 +3,66 @@ package plugin
 
 import (
 	"fmt"
-	"plugin"
+	"io"
+	"net/rpc"
+	"os/exec"
+	"sync"
+	"syscall"
+
+	hclog "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-plugin"
 
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
 )
 
 const (
-	// EntryPoint is the entry point (FactoryCtor) expected to be defined
-	// in each signer plugin.
-	EntryPoint = "GetPluginCtor"
-
-	// SignerName is the name used to identify the plugin signer.
+	// SignerName is the name used to identify the plugin backed signer.
 	SignerName = "plugin"
+
+	pluginName = "oasis_core_signer"
 )
 
-// FactoryCtor is the function signature of the EntryPoint symbol
-// that is expected to be defined in each plugin.
-type FactoryCtor func() signature.SignerFactoryCtor
+var (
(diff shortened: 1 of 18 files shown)


go/oasis-node/main.go
```diff
@@ -2,9 +2,14 @@
 package main
 
 import (
+	"github.com/hashicorp/go-plugin"
+
 	"github.com/oasisprotocol/oasis-core/go/oasis-node/cmd"
 )
 
 func main() {
+	// If we use go-plugin, we are supposed to clean clients up.
```
kostko, line 11: Probably the same in remote signer?

Yawning: And in the tests yeah.  I'll deal with that after I unbreak the tests.

go/common/crypto/signature/signers/plugin/rpc.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
+)
+
+var _ Signer = (*rpcClient)(nil)
+
+type rpcServer struct {
+	impl Signer
+}
+
+// XXX: Do the arg types need to be public?
```
kostko: Do they need to be? All the encoding/decoding is done in this package.

go/common/crypto/signature/signers/plugin/plugin.go
```diff
@@ -3,33 +3,68 @@ package plugin
...
+	return plugin.HandshakeConfig{
+		ProtocolVersion:  1,
+		MagicCookieKey:   "OASIS_CORE_SIGNER_PLUGIN",
+		MagicCookieValue: name,
+	}
+}
 
 // FactoryConfig is the plugin factory configuration.
 type FactoryConfig struct {
+	// Name is the expected human readable name of the plugin (eg: "ledger", "memory")
```
kostko, line 44: Why does the name need to be specified here? Shouldn't it be provided by the plugin?

go/common/crypto/signature/signers/plugin/rpc.go
Yawning: Yes.

go/common/crypto/signature/signers/plugin/plugin.go
Yawning: I could remove it entirely instead and just have every plugin named `signer` as far as the transport goes.  Figured I'd make users explicitly know what plugin the random binary they're passing to the node actually is.

PR #3153, opened 2020-07-31
go/consensus: Add SubmitTxNoWait method

### LINKED ISSUE #3152: Add non-blocking consensus SubmitTx variant
The current `SubmitTx` method also blocks until the transaction is included in a block (or the context is cancelled). We should add a non-blocking `SubmitTxNoWait` method which only validates against the local mempool and does not wait for the transaction to be included in a block.

kostko:
Fixes #3152 

The new method allows submitting a transaction without waiting for it to be
included in a block.

### DIFF
--- a/.changelog/3152.feature.md
+++ b/.changelog/3152.feature.md
@@ -1,0 +1,4 @@
+go/consensus: Add SubmitTxNoWait method
+
+The new method allows submitting a transaction without waiting for it to be
+included in a block.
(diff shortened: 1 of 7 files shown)


Yawning:
Do we need another call for this vs just adding an argument to the existing method?

kostko:
> Do we need another call for this vs just adding an argument to the existing method?

Yeah thought about that, but the blocking `SubmitTx` should only be available over the non-light interface (as it requires state and some more resources to track the transaction) and having a separate method makes it easier to split the two between interfaces.

Issue #3146, opened 2020-07-30
Dont allow registering a non-TEE keymanager unless it's a test runtime


ptrus:
Unless the runtime ID has the test bit set, don't allow a keymanager runtime to register if it is not a TEE runtime.

PR #3154, opened 2020-07-31
go/registry: Require SGX for non-test keymanager runtimes

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3146

### DIFF
--- a/go/consensus/tendermint/apps/registry/transactions.go
+++ b/go/consensus/tendermint/apps/registry/transactions.go
@@ -540,20 +540,6 @@ func (app *registryApplication) registerRuntime( // nolint: gocyclo
 		return registry.ErrIncorrectTxSigner
 	}
 
-	// If TEE is required, check if runtime provided at least one enclave ID.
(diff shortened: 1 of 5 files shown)


go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -540,20 +540,6 @@ func (app *registryApplication) registerRuntime( // nolint: gocyclo
 		return registry.ErrIncorrectTxSigner
 	}
 
-	// If TEE is required, check if runtime provided at least one enclave ID.
```
ptrus, line 543: This was moved into `VerifyRegisterRuntimeArgs` since currently it was skipped by sanity checks and only checked on the register transactions.

.changelog/3146.feature.md
```diff
@@ -0,0 +1,4 @@
+go/registry: Require SGX for non-test keymanager runtimes
```
kostko, line 1: This is breaking.

Issue #3145, opened 2020-07-30
Keymanager runtime compute runtime policies


ptrus:
Keymanager runtime usage is currently only restricted via the keymanager policies, while in registry any compute runtime can specify any keymanager runtime. 

Each keymanager runtime should be able to control which compute runtimes are allowed to use it as a keymanager in the registry.

ptrus:
Not needed, instead the keymanager worker will ignore runtimes from registry that are not in policy document.

PR #3158, opened 2020-07-31
go/registry/runtimes: Add AllowedComputeRuntimes to keymanager runtimes

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3145

### DIFF
--- a/go/registry/api/runtime.go
+++ b/go/registry/api/runtime.go
@@ -208,6 +208,13 @@ func (s *RuntimeStakingParameters) ValidateBasic(runtimeKind RuntimeKind) error 
 	return nil
 }
 
+// KeyManagerParameters are the parameters specific to key manager runtimes.
+type KeyManagerParameters struct {
+	// AllowedComputeRuntimes sets which compute runtimes are allowed to
+	// register with this key manager runtime.
+	AllowedComputeRuntimes map[common.Namespace]bool `json:"allowed_compute_runtimes"`
+}
+
 const (
 	// LatestRuntimeDescriptorVersion is the latest entity descriptor version that should be used
 	// for all new descriptors. Using earlier versions may be rejected.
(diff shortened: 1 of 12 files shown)


go/registry/api/runtime.go
```diff
@@ -256,6 +263,9 @@ type Runtime struct { // nolint: maligned
 	// Storage stores parameters of the storage committee.
 	Storage StorageParameters `json:"storage,omitempty"`
 
+	// KeyManager parameters stores paratemers specific to Keymanager runtime.
+	KeyManagerParameters KeyManagerParameters `json:"key_manager_parameters,omitempty"`
```
ptrus, line 267: Would rather just use `KeyManager` (as we do for other types of parameters), but we already use it for specifying keymanager runtime id. 

Since this is already "breaking" could also rename the `KeyManager` field to `KeyManagerID` here, and use `KeyManager` for parameters.

Yawning: I would just rename the field tbh.

Yawning:
Wait, why does this need another field.  Just parse the policy document.

go/registry/api/runtime.go
kostko: Yeah, just rename the field.

kostko:
> Just parse the policy document.

Yeah that could work as well, it would require the registration code to look into the keymanager backend state though. On the other hand, there would then only be a single source of truth which seems like a good thing.

ptrus:
Yeah now that we don't allow non-sgx keymanagers (https://github.com/oasisprotocol/oasis-core/pull/3154), i guess we can indeed use the policy document.

ptrus:
I'm not sure if the complexity this would bring (mostly by needing the km service state in registry) is worth doing though, i think that in that case we rather do not do any such control on the registry level, and leave it up to the policies to handle this (keep it as it is).

kostko:
Maybe it would be better to make sure that the key manager worker (not only the enclave) ignores any runtimes which are not in the policy document but are trying to use the key manager? This would make the key manager reject requests one layer earlier.

ptrus:
> Maybe it would be better to make sure that the key manager worker (not only the enclave) ignores any runtimes which are not in the policy document but are trying to use the key manager? This would make the key manager reject requests one layer earlier.

Ok, will open a separate PR for that and close this one

PR #3160, opened 2020-08-03
go/storage/mkvs/checkpoint: Remove empty version directories

kostko:
When all root checkpoints are removed for a specific version, the version dir
itself should also be removed.

### DIFF
--- a/go/storage/mkvs/checkpoint/file.go
+++ b/go/storage/mkvs/checkpoint/file.go
@@ -138,16 +138,16 @@ func (fc *fileCreator) GetCheckpoints(ctx context.Context, request *GetCheckpoin
 	return cps, nil
 }
 
-func (fc *fileCreator) GetCheckpoint(ctx context.Context, request *GetCheckpointRequest) (*Metadata, error) {
+func (fc *fileCreator) GetCheckpoint(ctx context.Context, version uint16, root node.Root) (*Metadata, error) {
 	// Currently we only support a single version.
-	if request.Version != checkpointVersion {
+	if version != checkpointVersion {
 		return nil, ErrCheckpointNotFound
 	}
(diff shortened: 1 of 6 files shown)


go/storage/mkvs/checkpoint/file.go
```diff
@@ -162,22 +162,43 @@ func (fc *fileCreator) GetCheckpoint(ctx context.Context, request *GetCheckpoint
...
-	checkpointDir := filepath.Join(
-		fc.dataDir,
-		strconv.FormatUint(request.Root.Version, 10),
-		request.Root.Hash.String(),
-	)
-	if _, err := os.Stat(checkpointDir); err != nil {
+	versionDir := filepath.Join(fc.dataDir, strconv.FormatUint(root.Version, 10))
+	checkpointDir := filepath.Join(versionDir, root.Hash.String())
+	checkpointFilename := filepath.Join(checkpointDir, checkpointMetadataFile)
+	if err := os.Remove(checkpointFilename); err != nil {
```
Yawning, line 174: If you're going to `RemoveAll` the directory that contains the checkpoint file, isn't explicitly removing the checkpoint dir redundant?

kostko: The idea is that `RemoveAll` could take some time if there are a lot of chunks. If we remove the metadata file first, this will prevent the checkpoint from being listed in `GetCheckpoints` etc. while it is being removed. The alternative would be to rename the checkpoint directory before removing it (and then skipping such directories in `GetCheckpoints`).

In practice it may not make a big difference if removals are fast.

Yawning: Ok, that seems fine.

PR #3161, opened 2020-08-03
go/registry/metrics: Runtimes metrics: query for runtimes

ptrus:
Runtime events are also triggered when a runtime is resumed, however we never subtract a runtime being suspended.

### DIFF
--- a/go/consensus/tendermint/full/full.go
+++ b/go/consensus/tendermint/full/full.go
@@ -161,15 +161,14 @@ type fullService struct { // nolint: maligned
 
 	stateDb tmdb.DB
 
-	beacon          beaconAPI.Backend
-	epochtime       epochtimeAPI.Backend
-	keymanager      keymanagerAPI.Backend
-	registry        registryAPI.Backend
-	registryMetrics *registry.MetricsUpdater
-	roothash        roothashAPI.Backend
-	staking         stakingAPI.Backend
-	scheduler       schedulerAPI.Backend
-	submissionMgr   consensusAPI.SubmissionManager
+	beacon        beaconAPI.Backend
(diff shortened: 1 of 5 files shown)


ptrus:
> One downside here is that this will cause the query to be issued every 10 seconds.

But we already do the same for `Nodes` & `Entities`, feels weird we'd only want to optimize the runtimes one (which will most likely also always be the smallest one).

ptrus:
Anyway fixed.

kostko:
Oh I didn't see that, sorry. So in this case I would be for keeping this consistent and the thing you proposer would actually be better, just maybe adjust the update interval (the current one seems excessive, something like 1 min should be enough).

I would however disable the metrics updater in case metrics are disabled to avoid needless queries as currently it seems this is always running even if the metrics are not actually reported anywhere.

go/consensus/tendermint/full/full.go
```diff
@@ -922,10 +922,12 @@ func (t *fullService) initialize() error {
 		return err
 	}
 	t.registry = scRegistry
-	t.registryMetrics = registry.NewMetricsUpdater(t.ctx, t.registry)
+	if cmmetrics.Enabled() {
+		t.registryMetrics = registry.NewMetricsUpdater(t.ctx, t.registry)
```
kostko, line 926: While we're here, we can probably also just get rid of the `t.registryMetrics` member variable and use a local one instead (as it doesn't seem to be used anywhere else besides getting registered with the service manager on the next line).

ptrus: fixed

kostko: Thanks!

PR #3162, opened 2020-08-03
go/worker/keymanager: Ignore runtimes not in policy document

ptrus:
Fixes:
>Maybe it would be better to make sure that the key manager worker (not only the enclave) ignores any runtimes which are not in the policy document but are trying to use the key manager? This would make the key manager reject requests one layer earlier.

TODO:
- [ ] ~potentially also implement stopping clientRuntimeWatchers for runtimes that were removed from policy docs~
  - can do in future if needed

### DIFF
--- a/go/worker/keymanager/worker.go
+++ b/go/worker/keymanager/worker.go
@@ -18,6 +18,7 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common/node"
 	"github.com/oasisprotocol/oasis-core/go/common/pubsub"
 	"github.com/oasisprotocol/oasis-core/go/common/service"
+	consensus "github.com/oasisprotocol/oasis-core/go/consensus/api"
 	"github.com/oasisprotocol/oasis-core/go/keymanager/api"
 	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
 	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
@@ -64,6 +65,9 @@ type Worker struct { // nolint: maligned
 
 	runtime            runtimeRegistry.Runtime
 	runtimeHostHandler protocol.Handler
+
+	clientRuntimes       map[common.Namespace]*clientRuntimeWatcher
+	clientRuntimesQuitCh chan *clientRuntimeWatcher
 
 	commonWorker  *workerCommon.Worker
 	roleProvider  registration.RoleProvider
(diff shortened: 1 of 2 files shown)


go/worker/keymanager/worker.go
```diff
@@ -345,6 +349,112 @@ func extractMessageResponsePayload(raw []byte) ([]byte, error) {
...
+	case status.Policy != nil:
+		for _, enc := range status.Policy.Policy.Enclaves {
+			if _, ok := enc.MayQuery[rt.ID]; ok {
+				found = true
+				break
+			}
+		}
+	}
+	if !found {
+		w.logger.Debug("runtime not found in keymanager policy, skipping",
```
kostko, line 383: Maybe log as a warning or move above "seen new runtime ..." info after the policy check.

go/worker/keymanager/worker.go
```diff
@@ -486,68 +603,36 @@ func (w *Worker) worker() { // nolint: gocyclo
 				)
 				continue
 			}
+			// New runtimes can be allowed with the policy update.
+			if err = w.recheckAllRuntimes(currentStatus); err != nil {
+				w.logger.Error("failed rechecking runtimes",
+					"err", err,
+				)
+				return
```
kostko, line 611: Should this `continue` instead of terminating the whole key manager worker?

go/worker/keymanager/worker.go
```diff
@@ -486,68 +603,36 @@ func (w *Worker) worker() { // nolint: gocyclo
...
+				w.logger.Error("failed rechecking runtimes",
 					"err", err,
 				)
-				continue
-			}
-
-			crw := &clientRuntimeWatcher{
-				w:    w,
-				node: node,
+				return
```
kostko, line 626: Same as above, `continue`.

go/worker/keymanager/worker.go
ptrus: Yeah maybe that's more in line with the rest of the code. Although we do `return` already if the `WatchEvents` fails (line 596), and similarly i made the `recheckAllRuntimes` return an error only in case the registry query fails. But probably still better to `continue` here so will change it.

go/worker/keymanager/worker.go
ptrus: Yeah makes sense, changed this one into a warning.

PR #3165, opened 2020-08-04
fix HMAC_DRBG edge case

keks:
The [NIST SP 800-90A] spec of HMAC_DRBG uses "`!= Null`" checks, but "`Null`" is never defined in the document. In the current implementation, it was interpreted as `nil.` However, the [NIST examples document] uses "`<empty>`" values, so `[]byte{}` should be passed in the test cases. If that happens, the `nil` checks do not work as expected. Therefore, instead of checking equality with `nil`, it should be checked whether the length of `additional_info` is zero.

As additional evidence I cite the [openssl implementation], which checks for zero length instead of NULL pointer equality.

I do not expect this to have a concrete security impact, but still it would be nice if the implementation was correct.

[NIST SP 800-90A]: https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-90Ar1.pdf
[NIST examples document]: https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/HMAC_DRBG.pdf
[openssl implementation]: https://github.com/openssl/openssl/pull/6779/files#diff-c44c9681c4fc49b1d7e9e74578085212R77-R79


go/common/crypto/drbg/hmac_drbg_test.go
```diff
@@ -36,7 +36,7 @@ func TestDrbgNISTExample1(t *testing.T) {
 	)
 
 	// HMAC_DRBG_Generate
-	v, err := drbg.generate(readSz, nil)
+	v, err := drbg.generate(readSz, []byte{})
```
kostko, line 39: Should we test both `nil` and `[]byte{}` in tests? They should behave identically now.

Yawning: Yeah I would test both just to prevent regressions.

keks:
I duplicated the tests and in one of them use `nil` and in the other `[]byte{}`. Additionally, I added a changelog fragement and capitalized the commit subject line. Unfortunately, the URLs are wider than 80 characters, and I wasn't sure how to wrap them.

Finally, I'm not 100% sure if any part of Oasis uses an empty slice somewhere in it's code, so I left a TODO in the fragment. I can update it if you have a definite answer for me.

Yawning:
> Finally, I'm not 100% sure if any part of Oasis uses an empty slice somewhere in it's code, so I left a TODO in the fragment. I can update it if you have a definite answer for me.

No, we don't, and in fact, there is no way for the edge case to be triggered.  Since the affected routines are package private, it's fairly easy to confirm this via manual review.

 * `Drbg.Read` always passes `nil` as `additionalInput`
 * `New` always passes a non-0 length slice (`seedMaterial`) as `additionalInput`

keks:
I briefly skimmed the output of `git grep drbg` but from my experience the package maintainers usually have a much better overview :slightly_smiling_face: 

I feel like the URLs are helpful in the commit messages so I'm hesitant to remove them in order to make the linter happy. How do you want to proceed?

kostko:
> I feel like the URLs are helpful in the commit messages so I'm hesitant to remove them in order to make the linter happy. How do you want to proceed?

Good point, I just fixed the linter in #3168, once that's merged, you can rebase and the linter should pass. I'll also trigger the non-lint CI workflow again after your rebase.

keks:
@kostko 

Sorry this took so long. I think the linter regex doesn't like the hash fragment. [On line 10, it says:](https://github.com/oasisprotocol/oasis-core/pull/3168/files#diff-21ca6cf7d49f93265a8ba2e38f32a2c2R10)

```
 URL_RE_RAW = r'http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+' 
```

Note that the third bracket after `://` we have `[$-_@.&+]`, where `$-_` means all characters between `$` (0x24) and `_` (0x5f). However, the hash sign `#` is 0x23, so it doesn't match this bracket expression.

Do you want to fix the regex or should we skip the more precise reference and use a link without hash fragment?

kostko:
Aw sorry, I only tested with the first URL. Feel free to fix the regexp in this PR, just in a separate commit.

kostko:
Thanks again! I also triggered the other CI workflow to run now.

Issue #1931, opened 2019-07-17
Slash stake for malicious txn schedulers


abukosek:
In `go/roothash/api/commitment/pool.go` when we discover that a txn sched has signed two different batches for the same committee, we should slash its stake.  Currently, only a warning is printed.

One possible design is to introduce a `Slasher` interface which is passed to the methods that need to decide whether to slash and then call into that interface when slashable offences occur.

## Acceptance Criteria
* Stake is slashed for malicious txn schedulers.

kostko:
Superseded by #3685.

PR #3179, opened 2020-08-07
go/roothash: Drop support for multiple committees

kostko:
Since there is currently no transaction scheduler implementation which would
support multiple committees, there is no sense in having the merge node as it
could be a source of bugs.

The merge node is also the only client for the Merge* storage operations, so
they can just be removed in order to reduce the exposed API surface.


go/worker/common/committee/group.go
```diff
@@ -609,13 +503,8 @@ func (g *Group) PublishScheduledBatch(
 		return nil, fmt.Errorf("group: not leader of txn scheduler committee")
 	}
 
-	xc := g.activeEpoch.executorCommittees[committeeID]
-	if xc == nil {
-		return nil, fmt.Errorf("group: invalid executor committee")
-	}
-
+	// XXX: this should be moved to the txn scheduler and passed as argument
```
pro-wh, line 513: not just constructing this message, this whole method seems very specialized, e.g. with checking membership as txn scheduler leader

go/worker/common/committee/group.go
```diff
@@ -567,11 +466,7 @@ func (g *Group) HandlePeerMessage(unusedPeerID signature.PublicKey, msg *p2p.Mes
 	return g.handler.HandlePeerMessage(ctx, msg)
 }
 
-func (g *Group) publishLocked(
-	spanCtx opentracing.SpanContext,
-	ci *CommitteeInfo,
```
pro-wh, line 572: why was this already unused? where have we been sending these messages?

go/worker/compute/executor/committee/fault.go
```diff
@@ -1,186 +0,0 @@
...
-
-	// TODO: Once we have P2P gossipsub also look at gossiped commitments in addition to consensus.
-
-	for {
-		select {
-		case <-ctx.Done():
-			return
-		case <-timer.C:
-			// Local round timeout expired.
-			d.logger.Warn("local round timeout expired without seeing our commitment")
```
pro-wh, line 104: what was this file, and why are we removing it as part of this change?

go/worker/compute/executor/committee/node.go
```diff
@@ -700,71 +687,39 @@ func (n *Node) proposeBatchLocked(batch *protocol.ComputedBatch) {
 		return
 	}
 
-	// Publish commitment to merge committee.
-	spanPublish := opentracing.StartSpan("PublishExecuteFinished(commitment)",
```
pro-wh, line 704: ugh goodbye tracing

go/worker/storage/committee/node.go
```diff
@@ -372,17 +372,12 @@ func (n *Node) updateExternalServicePolicyLocked(snapshot *committee.EpochSnapsh
...
-			executorCommitteePolicy.AddRulesForCommittee(&policy, xc, snapshot.Nodes())
-		}
+	if xc := snapshot.GetExecutorCommittee(); xc != nil {
+		executorCommitteePolicy.AddRulesForCommittee(&policy, xc, snapshot.Nodes())
 	}
 	if tsc := snapshot.GetTransactionSchedulerCommittee(); tsc != nil {
 		txnSchedulerCommitteePolicy.AddRulesForCommittee(&policy, tsc, snapshot.Nodes())
 	}
-	if mc := snapshot.GetMergeCommittee(); mc != nil {
-		mergeCommitteePolicy.AddRulesForCommittee(&policy, mc, snapshot.Nodes())
```
pro-wh, line 384: oh it is removed in a later commit

go/worker/common/committee/group.go
```diff
@@ -440,18 +444,21 @@ func (g *Group) HandlePeerMessage(unusedPeerID signature.PublicKey, msg *p2p.Mes
...
 
-		return g.activeEpoch.roundCtx, nil
+		return nil
 	}()
 	if err != nil {
 		return err
 	}
 
+	ctx, cancel := context.WithTimeout(context.Background(), peerMessageProcessTimeout)
+	defer cancel()
```
pro-wh, line 460: what's the context of this commit? why do we now want to allow handling of messages while we're not in a round?

also, why `defer cancel()` here? is it know that `HandlePeerMessage` doesn't do anything after returning?

go/worker/storage/committee/policy.go
```diff
@@ -22,10 +22,7 @@ var (
 		},
 	}
 	mergeCommitteePolicy = &committee.AccessPolicy{
-		Actions: []accessctl.Action{
-			accessctl.Action(api.MethodMerge.FullName()),
-			accessctl.Action(api.MethodMergeBatch.FullName()),
-		},
+		Actions: []accessctl.Action{},
```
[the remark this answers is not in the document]
kostko: You seem to be reviewing some of the changes that were superseded in a later commit.

go/roothash/api/commitment/pool.go
```diff
@@ -256,35 +228,33 @@ func (p *Pool) addOpenExecutorCommitment(
...
-				// don't match -- txn sched is malicious!
-				// TODO: Slash stake! (issue #1931)
-				logger.Warn("txn sched signed two different batches for the same committee ID",
-					"committee_id", cb.CommitteeID,
-				)
-			}
+		existingTxnSchedSig := cb.TxnSchedSig
+		if currentTxnSchedSig.PublicKey.Equal(existingTxnSchedSig.PublicKey) &&
+			!currentTxnSchedSig.Signature.Equal(existingTxnSchedSig.Signature) &&
+			!cb.InputRoot.Equal(&com.Body.InputRoot) {
```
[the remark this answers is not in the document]
kostko: Whoops, the other one should be `body` not `com.Body`.

go/roothash/api/commitment/pool.go
```diff
@@ -256,35 +228,33 @@ func (p *Pool) addOpenExecutorCommitment(
...
-				// Same committe, same txn sched, but txn sched signatures
-				// don't match -- txn sched is malicious!
-				// TODO: Slash stake! (issue #1931)
-				logger.Warn("txn sched signed two different batches for the same committee ID",
-					"committee_id", cb.CommitteeID,
-				)
-			}
+		existingTxnSchedSig := cb.TxnSchedSig
+		if currentTxnSchedSig.PublicKey.Equal(existingTxnSchedSig.PublicKey) &&
+			!currentTxnSchedSig.Signature.Equal(existingTxnSchedSig.Signature) &&
```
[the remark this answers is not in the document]
kostko: It doesn't matter as both work for a fixed-size array, but `Equal` seems more consistent with the rest.

go/roothash/api/commitment/pool_test.go
```diff
@@ -157,9 +155,31 @@ func TestPoolSingleCommitment(t *testing.T) {
...
+	for _, tc := range []struct {
+		name        string
+		fn          func(*ComputeBody)
+		expectedErr error
+	}{
+		{"BlockBadRound", func(b *ComputeBody) { b.Header.Round-- }, ErrNotBasedOnCorrectBlock},
+		{"BlockBadPreviousHash", func(b *ComputeBody) { b.Header.PreviousHash.FromBytes([]byte("invalid")) }, ErrNotBasedOnCorrectBlock},
+		{"StorageSigs1", func(b *ComputeBody) { b.StorageSignatures = nil }, ErrBadStorageReceipts},
+	} {
+		invalidBody := body
```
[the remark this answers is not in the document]
kostko: How come? Doesn't this create a copy?

go/roothash/api/commitment/pool.go
```diff
@@ -764,10 +764,10 @@ func (m *MultiPool) CheckEnoughCommitments() error {
 }
 
 // GetExecutorCommitments returns a list of executor commitments in the pool.
-func (m *MultiPool) GetExecutorCommitments() (result []ExecutorCommitment) {
+func (m *MultiPool) GetOpenExecutorCommitments() (result []OpenExecutorCommitment) {
```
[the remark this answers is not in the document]
kostko: This no longer exists, you seem to be reviewing stuff that was superseded in a later commit.

go/worker/compute/executor/committee/fault.go
kostko: Because the point of the fault detector was to independently submit executor commits in case the merge committee is faulty. If there is no merge committee, this is no longer needed.

go/roothash/api/commitment/pool_test.go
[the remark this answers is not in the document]
pro-wh: it's not something one would be able to verify by looking at this line. one would need to examine the struct fields, recursively. and also when making changes to those structs, one would have to know about places like these

go/worker/compute/executor/committee/fault.go
pro-wh: oh
:information_source: now executors submit directly to consensus

go/worker/common/committee/group.go
kostko: Yeah it was unused since we migrated to gossipsub. Messages are published to all P2P peers subscribing to the runtime-specific topic via gossipsub.

go/worker/common/committee/group.go
kostko: Yeah, we should probably just expose a more generic `Publish`. I was planning to do that but forgot, thanks for finding this comment ;)

go/worker/compute/executor/committee/node.go
kostko: Just for the merge part which is obviously gone now. The other tracing should still be there.

pro-wh: we still submit the result, but we no longer trace it

go/worker/common/committee/group.go
kostko: Fixed.

go/roothash/api/commitment/pool.go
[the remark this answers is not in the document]
pro-wh: thanks

so did we want to exclude cases where the inputs are different but the signatures were the same?

go/roothash/api/commitment/pool.go
```diff
@@ -661,143 +502,3 @@ func (p *Pool) GetExecutorCommitments() (result []ExecutorCommitment) {
...
-	return
-}
-
-// ResetCommitments resets the commitments in the pool and clears their discrepancy
-// flags.
-func (m *MultiPool) ResetCommitments() {
-	for _, p := range m.Committees {
-		p.ResetCommitments()
-	}
-}
```
pro-wh, line 803: what were the changes to this file since the last push?

go/worker/compute/executor/committee/node.go
kostko: We can't trace submissions to the consensus layer, but I guess we could leave a span open until the consensus layer confirms. Will change it.

kostko: Hm, we didn't trace merge submissions to consensus before though.

pro-wh: oh ok let's continue not to. we can add it later if we want to see

go/roothash/api/commitment/pool.go
kostko: The only changes were fixes addressing your comments above. GitHub should really add a feature to show diffs to what you acked before.

go/roothash/api/commitment/pool.go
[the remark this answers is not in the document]
pro-wh: oh, I missed that we previously didn't compare the values

go/roothash/api/commitment/pool.go
```diff
@@ -256,35 +228,33 @@ func (p *Pool) addOpenExecutorCommitment(
...
-			}
+		existingTxnSchedSig := cb.TxnSchedSig
+		if currentTxnSchedSig.PublicKey.Equal(existingTxnSchedSig.PublicKey) &&
+			!currentTxnSchedSig.Signature.Equal(existingTxnSchedSig.Signature) &&
+			!cb.InputRoot.Equal(&body.InputRoot) {
+			// TODO: Signal to slash stake! (issue #1931).
+			logger.Warn("transaction scheduler signed two different batches for the same round",
+				"txn_scheduler_id", currentTxnSchedSig.PublicKey,
+				"input_root_a", cb.InputRoot,
+				"input_root_b", com.Body.InputRoot,
```
pro-wh, line 256: body here too

go/roothash/api/commitment/pool.go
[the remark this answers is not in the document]
pro-wh: why do we now additionally compare the values?

go/roothash/api/commitment/pool.go
pro-wh: thanks

go/roothash/api/commitment/pool.go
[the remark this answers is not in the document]
kostko: Just to make it explicit what is being checked, if the values change the signatures will also be different. In any case, this check doesn't make sense until we actually have slashing (see #1931), so I've removed it.

go/roothash/api/commitment/executor.go
```diff
@@ -32,6 +32,7 @@ var (
 //
 // Keep the roothash RAK validation in sync with changes to this structure.
 type ComputeResultsHeader struct {
+	Round        uint64           `json:"round"`
```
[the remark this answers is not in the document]
kostko: Yeah, it is just to make it easier to determine if the commit is old. Also it will be useful when we decide to slash -- signing two different commits for the same round should be slashable.

go/roothash/api/commitment/pool_test.go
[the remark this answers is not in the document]
kostko: Changed it to an explicit generate + signature copy.

go/worker/common/committee/group.go
kostko: Yeah the previous context that was used was actually incorrect. `HandlePeerMessage` should dispatch longer-lasting operations to background goroutines with their own contexts (all the current implementations do). I'll add an explicit comment to the interface.

go/worker/compute/txnscheduler/algorithm/batching/batching.go
```diff
@@ -44,29 +44,6 @@ type config struct {
...
-	// with multiple committees will currently cause the rounds to fail as all other
-	// committees will be idle.
-	var committeeID *hash.Hash
-	func() {
-		// Guarding against EpochTransition() modifying current epoch.
-		s.RLock()
-		defer s.RUnlock()
-
-		// We cannot schedule anything until there is an epoch transition.
-		if s.epoch == nil {
```
ptrus, line 57: by removing this, i think the `s.epoch` field can be removed as well (and with that also the `EpochTransition` alg. interface method, and also the lock on the batching state)

kostko: Good idea, removed.

go/worker/common/committee/group.go
kostko: Done.

pro-wh: thanks for adding that info

go/roothash/api/commitment/executor.go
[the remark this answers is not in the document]
pro-wh: ok

PR #3180, opened 2020-08-10
Make the random beacon suck less

Yawning:
* [x] Implement PVSS
 * [x] Rewrite all the consensus bits
   * [x] Switch to unified epochtime/beacon
   * [x] epochtime
   * [x] beacon
 * [x] Write the new PVSS node worker
 * [x] Alter the tooling to support the new parameters
   * [x] oasis-node genesis
   * [x] oasis-test-runner
 * [x] Test
   * [x] Make old tests pass ~(Blocked on #3404)~
   * [x] Add new tests for the trivial failure cases
 * [x] Backfill protocol error handling
   * [x] Slashing
   * [x] Worker should retry submitting failed tx(es)
   * [x] The beacon tx(es) should cost gas
   * [x] Disable new node eligibility for next election on end of commit phase
   * [x] Round failure should disable runtimes
 * [ ] Review/pending changes
   * [x] Unify Threshold/PVSSThreshold
   * [ ] Refactoring required to support the transition

To be done later (file issues after merge):
 * [ ] Make getting the historical epoch suck less
 * [ ] Add more test cases
 * [ ] Improve performance by adding in the optimizations from the SCRAPE paper.
 * [ ] Add a beacon gRPC service, just so people can query the consensus parameters.

I'm not the most thrilled by the performance from the PVSS scheme when used without pairing based crypto, but it has several nice properties:

 * It doesn't chew up ridiculous amounts of consensus state, even with a relatively large number of nodes.
 * All protocol messages can be independently verified for correctness, on-chain.
 * The final entropy generation can be pushed on chain.
 * The number of broadcast messages (transactions) is not massive.
 * Someone else implemented dleq based PVSS so it's easy to implement.


go/common/crypto/scrape/scrape_test.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+func TestScrape(t *testing.T) {
+	t.Run("Basic", func(t *testing.T) { doTestBasic(t, 0, 0, 0) })
+	t.Run("MissingCommits", func(t *testing.T) { doTestBasic(t, 3, 0, 0) })
+	t.Run("MissingReveals", func(t *testing.T) { doTestBasic(t, 0, 3, 0) })
+	t.Run("MissingVerifierReveals", func(t *testing.T) { doTestBasic(t, 0, 0, 3) })
+}
+
+func doTestBasic(t *testing.T, truncateCommits, truncateReveals, truncateVerReveals int) {
```
kostko, line 26: One useful thing to test would be serialization round-trips of these various types to make sure there are no surprises.

Yawning: Yeah I just came to the realization that types that contain `kyber.Scalar`/`kyber.Point` don't handle s11n well because they're interface types.

Yawning: Fixed and now tested.

go/common/node/node.go
```diff
@@ -68,6 +69,13 @@ type Node struct { // nolint: maligned
 	// consensus member.
 	Consensus ConsensusInfo `json:"consensus"`
 
+	// Beacon contains information for this node's participation
+	// in the random beacon protocol.
+	//
+	// TODO: This is optional for now, make mandatory once enough
+	// nodes provide this field.
+	Beacon *BeaconInfo `json:"beacon,omitempty"`
```
kostko, line 77: We should include this in 20.10 in some form so that nodes will already be able to include this field (and the inner `point` as `[]byte`) before there is actually support for the new beacon protocol. Is this structure relatively stable? Otherwise we could also use `Beacon cbor.RawMessage` for now.

kostko: #3256

go/consensus/tendermint/apps/beapoch/backend_scrape.go
```diff
@@ -0,0 +1,431 @@
...
+	MethodSCRAPECommit = transaction.NewMethodName(AppName, "SCRAPECommit", SCRAPECommit{})
+
+	// MethodSCRAPEReveal is the method name for a SCRAPE reveal.
+	MethodSCRAPEReveal = transaction.NewMethodName(AppName, "SCRAPEReveal", SCRAPEReveal{})
+)
+
+// SCRAPECommit is a SCRAPE commitment transaction payload.
+type SCRAPECommit struct {
+	Epoch epochtime.EpochTime `json:"epoch"`
+	Round int                 `json:"round"`
```
kostko, line 38: Should use an architecture-independent type (also below).

go/consensus/tendermint/apps/beapoch/backend_scrape.go
```diff
@@ -0,0 +1,431 @@
...
+func (impl *backendSCRAPE) ExecuteTx(
+	ctx *api.Context,
+	state *beapochState.MutableState,
+	params *beapochState.ConsensusParameters,
+	tx *transaction.Transaction,
+) error {
+	if tx.Method != MethodSCRAPEReveal && tx.Method != MethodSCRAPECommit {
+		return fmt.Errorf("beapoch: invalid method: %s", tx.Method)
+	}
+
```
kostko, line 302: Gas costs? (I know that in practice only validators can call these methods, so they could opt to pay zero gas by only including these transactions when they are the proposer.)

go/consensus/tendermint/apps/beapoch/state/state_scrape.go
```diff
@@ -0,0 +1,62 @@
...
+	epochtime "github.com/oasisprotocol/oasis-core/go/epochtime/api"
+)
+
+var scrapeStateKeyFmt = keyformat.New(0x42)
+
+// RoundState is a SCRAPE round state.
+type RoundState int
+
+const (
+	StateInvalid RoundState = iota
```
kostko, line 20: Since this is serializable I would rather we use explicit values instead of `iota`.

go/consensus/tendermint/apps/beapoch/state/state_scrape.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
+	"github.com/oasisprotocol/oasis-core/go/common/keyformat"
+	abciAPI "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/api"
+	epochtime "github.com/oasisprotocol/oasis-core/go/epochtime/api"
+)
+
+var scrapeStateKeyFmt = keyformat.New(0x42)
+
+// RoundState is a SCRAPE round state.
+type RoundState int
```
kostko, line 17:
```suggestion
type RoundState uint8
```

kostko (commented):
Just some minor nits.

go/common/crypto/scrape/scrape.go
```diff
@@ -0,0 +1,512 @@
...
+}
+
+// MayReveal returns true iff it is possible to proceed to the reveal
+// step, and the total number of distinct valid commitments received.
+func (inst *Instance) MayReveal() (bool, int) {
+	totalCommits := len(inst.Commits)
+	return totalCommits >= inst.Threshold, totalCommits
+}
+
+// Reveal executes the commit phase of the protocol, generating a reveal
```
kostko, line 245:
```suggestion
// Reveal executes the reveal phase of the protocol, generating a reveal
```

go/consensus/tendermint/apps/beacon/api.go
```diff
@@ -1,25 +1,51 @@
...
+
+	// MethodSetEpoch is the method name for setting epochs.
+	MethodSetEpoch = transaction.NewMethodName(AppName, "SetEpoch", beacon.EpochTime(0))
+
+	// Methods is a list of all methods supported by the beacon application.
+	Methods = []transaction.MethodName{
+		MethodSetEpoch,
+		MethodSCRAPECommit,
+		MethodSCRAPEReveal,
+	}
```
kostko, line 44: The reason why `MethodSetEpoch` was previously defined here (in an internal Tendermint-specific package) is because it was only used by mock epochtime. For non-mock methods it may be useful to actually define them in the beacon API package (e.g., `go/beacon/api/scrape.go`) and add methods for generating transactions.

That is unless the SCRAPE-based implementation is supposed to be an internal detail of the Tendermint consensus backend.

go/consensus/tendermint/apps/beacon/backend_insecure.go
```diff
@@ -0,0 +1,228 @@
...
+			// fantastic, but it's only for one epoch.
+			ctx.Logger().Debug("onBeaconEpochChange: using block hash as entropy")
+			entropy = req.Hash
+		} else {
+			// Use the previous commit hash as the entropy input, under the theory
+			// that the merkle root of all the commits that went into the last
+			// block is harder for any single validator to game than the block
+			// hash.
+			//
+			// TODO: This still isn't ideal, and an entirely different beacon
```
kostko, line 135: This comment should probably be removed as this PR adds a SCRAPE-based beacon :-)

kostko (commented):
19/106 :confused:

go/consensus/tendermint/apps/beacon/api.go
Yawning: I have no strong opinions of moving this or not.  It is deeply tied to the tendermint implementation for now, so I am ok with leaving things here.  When we get another consensus implementation (if ever), moving it is trivial as well.

kostko: So I would like to avoid imports of consensus-backend internal stuff outside `go/consensus/tendermint`. Currently the beacon worker (in `go/worker/beacon`) needs these so maybe it would make sense to move them out to `go/beacon/api/scrape.go`. As you said it should be trivial and I think it makes the most sense. The "mock set epoch" method can stay here as it is not needed anywhere else.

kostko: (The other option would be moving the beacon worker into the Tendermint consensus backend and not exposing any of the SCRAPE stuff in `go/beacon/api/scrape.go`.)

Yawning: I sort of think that we're fooling ourselves if we think that we are actually going to support a different consensus backend, and that the code will be a lot simpler if we assume we are not  going to do that ever, but ok.

go/consensus/tendermint/apps/beacon/backend_scrape.go
```diff
@@ -0,0 +1,502 @@
...
+func (impl *backendSCRAPE) ExecuteTx( //nolint: gocyclo
+	ctx *api.Context,
+	state *beaconState.MutableState,
+	params *beaconState.ConsensusParameters,
+	tx *transaction.Transaction,
+) error {
+	if tx.Method != beacon.MethodSCRAPEReveal && tx.Method != beacon.MethodSCRAPECommit {
+		return fmt.Errorf("beacon: invalid method: %s", tx.Method)
+	}
+
```
kostko, line 325: I think you could make sure that each validator only proposes its own commits/reveals by rejecting transactions in case `ctx.IsCheckOnly()` is true and the `staking.NewAddress(ctx.TxSigner())` is not equal to `ctx.AppState().OwnTxSignerAddress()`. This will prevent other validator's commits/reveals from being included in the local mempool and thus in the blocks we propose.

Then the only other required thing would be to prevent the mempool from gossiping commit/reveal txes to other peers (as other proposers could otherwise still include them). (We don't need to do this second part in this PR to keep it simpler.)

go/consensus/tendermint/apps/beacon/beapoch.go
```diff
@@ -0,0 +1,126 @@
+// Package beacon implements the combined beacon and epochtime
```
kostko, line 1: This file is still named `beapoch.go`, should this just be moved under `beacon.go`?

go/consensus/tendermint/apps/beacon/state/state.go
```diff
@@ -23,6 +32,18 @@ var (
 	parametersKeyFmt = keyformat.New(0x41)
 )
 
+// ConsensusParmeters is the beacon consensus parameters.
+type ConsensusParameters struct {
+	Backend    string                     `json:"backend"`
```
kostko, line 37: Why isn't the `Backend` field in `beacon.ConsensusParameters` and then this separate structure would not be needed?

go/consensus/tendermint/apps/beacon/state/state_scrape.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"context"
+
+	beacon "github.com/oasisprotocol/oasis-core/go/beacon/api"
+	"github.com/oasisprotocol/oasis-core/go/common/cbor"
+	"github.com/oasisprotocol/oasis-core/go/common/keyformat"
+	abciAPI "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/api"
+)
+
+var scrapeStateKeyFmt = keyformat.New(0x42)
```
kostko, line 12: Add documentation comment on what this contains.

kostko (commented):
104/107

go/consensus/tendermint/apps/beacon/backend_scrape.go
Yawning: Done.

go/worker/beacon/worker.go
```diff
@@ -0,0 +1,588 @@
...
+			w.logger.Error("commit: participant node missing point",
+				"id", id,
+			)
+			return
+		}
+		points = append(points, node.Beacon.Point)
+	}
+	h, cmpH := hash.NewFrom(points), hash.NewFrom(scrapeState.Instance.Participants)
+	if !h.Equal(&cmpH) {
+		w.logger.Error("commit: participant point list mismatch")
```
pro-wh, line 202: does anything use the event that doesn't already know the correct participant point list?

go/worker/beacon/worker.go
```diff
@@ -0,0 +1,588 @@
...
+	w.persistState(&roundState{
+		Epoch:    ev.Epoch,
+		Round:    ev.Round,
+		State:    ev.State,
+		Instance: instance,
+		CommitTx: tx,
+	})
+
+	// Submit the commit tx.
+	// XXX: Add a retry?
```
pro-wh, line 249: will especially need this if we go with the design where you have to wait to be proposer before you can submit your commit/reveal

go/consensus/tendermint/apps/beacon/backend_scrape.go
Yawning: It's a problem if there is a large number of commits and or reveals in a single block, because commit/reveal processing is rather CPU intensive.

go/worker/beacon/worker.go
Yawning: What do you mean?  In theory the registry contains the points for each node, and in theory the consensus layer isn't producing busted events, so a simple search would suffice I suppose.  The check is cheap (the number of participants is unlikely to exceed 20 or so)...

go/worker/beacon/worker.go
Yawning: I mean, this is in the issue todo list for a reason.

go/worker/beacon/worker.go
```diff
@@ -0,0 +1,653 @@
...
+			)
+			return
+		}
+		if node.Beacon == nil {
+			w.logger.Error("commit: participant node missing point",
+				"id", id,
+			)
+			return
+		}
+		points = append(points, node.Beacon.Point)
```
kostko, line 208: If this is us (based on the node ID), the point should correspond to `w.identity.BeaconScalar.Point()`, right? I guess `scrape.New` will already check for participation in `SetScalar` so an additional check may be redundant.

go/worker/beacon/worker.go
```diff
@@ -0,0 +1,653 @@
...
+	}
+
+	if scrapeState.BadParticipants[w.identity.NodeSigner.Public()] {
+		w.logger.Error("reveal: node identity listed as bad participant")
+		return
+	}
+
+	// Handle all of the commits.
+	//
+	// Note: This could be better optimized by having the consensus layer
```
kostko, line 281: Maybe (once ready for merge) file a follow-up issue for a possible optimization.

go/worker/beacon/worker.go
Yawning: Huh?  `scrape.New` needs the Points (public keys) of every participant.  This assembles it based off the registry data (instead of blindly trusting `scrapeState`).

kostko: I'm just saying that in theory the point in the registry descriptor (`node.Beacon.Point`) for our own node may not correspond to the local point (`w.identity.BeaconScalar.Point()`). Currently the validation for that is when you do `scrape.New` as that checks that the specified private key is among the participants, and maybe that is enough. Otherwise we may want to also check here/above that the point is the one that we have locally?

go/consensus/tendermint/apps/beacon/backend_scrape.go
```diff
@@ -0,0 +1,819 @@
...
+	if err != nil {
+		return fmt.Errorf("beacon: failed to query node list: %w", err)
+	}
+
+	var ids []signature.PublicKey
+	for _, node := range nodes {
+		ids = append(ids, node.ID)
+	}
+
+	if err = state.SetSCRAPENodesSnapshot(ctx, ids); err != nil {
```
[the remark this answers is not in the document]
Yawning: It's super annoying to do if we want to continue to support the mock epoch, because it is not guaranteed to know `that specific epoch` when this happens.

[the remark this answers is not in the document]
Yawning: So while I took a stab at addressing this, I'm still against changing it as it complicates the code and breaks the e2e tests (timeouts).  I'm not sure if it's a bug in the changes or just "it takes even longer now", and I'm beyond reluctant to keep increasing the timeout (this stuff takes forever already).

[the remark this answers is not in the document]
kostko: Ok, feel free to revert, we can look at this later.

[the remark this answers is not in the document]
Yawning: Seems to work now, after I debugged it a bit.

go/beacon/api/api.go
```diff
@@ -15,41 +16,152 @@ const (
...
+	case BackendSCRAPE:
+		params := g.Parameters.SCRAPEParameters
+		if params == nil {
+			return fmt.Errorf("beacon: sanity check failed: SCRAPE backend not configured")
+		}
+
+		if params.Participants <= 1 {
+			return fmt.Errorf("beacon: sanity check failed: SCRAPE participants must be > 1")
+		}
+		if n := params.Threshold; n <= 1 || n > params.Participants {
```
bennetyee, line 137: (Using $\LaTeX$ style to refer to stuff from the paper and `this` to refer to code.)

In the SCRAPE paper, the thresholds used appears to be $t=n/2$ or `params.Participants/2`.  To make auditing easier, I suggest renaming the `n` here (and/or below at L140) to be `t`, and please add comments why we would want to allow the full range of possible values.  Is this threshold $|\mathcal{C}|$?  It appears to be from the comment in `go/common/crypto/scrape/scrape.go`, where it says this threshold probably should be at least `params.PVSSThreshold`, so this should probably be sanityy checked here, unless we don't want to enforce it for some reason (please provide reason somewhere).  The `New` there does not enforce this either.

Is $n$ or `params.Participants` expected to be about the size of the validator committee?

kostko: > Is $n$ or `params.Participants` expected to be about the size of the validator committee?

Depends on how big the validator committee is, usually this would be a subset of the committee due to performance reasons. E.g. at most ~20.

go/beacon/api/api.go
```diff
@@ -15,41 +16,152 @@ const (
...
+	// messages on epoch transitions. If an epoch transition happens
+	// before the previous epoch is read from the channel, the old
+	// epochs are overwritten.
+	//
+	// Upon subscription the current epoch is sent immediately.
+	WatchLatestEpoch() (<-chan EpochTime, *pubsub.Subscription)
+
+	// GetBeacon gets the beacon for the provided block height.
+	// Calling this method with height `0` should return the
+	// beacon for the latest finalized block.
```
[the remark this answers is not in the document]
Yawning: That's the existing API, and all of our accessors (in other subsystems) work in a similar manner.  Nothing should actually calls `GetBeacon` at the moment anyway, so we could just remove the call (though it is useful for testing).

go/beacon/api/api.go
Yawning: Truth be told, unless there's a strong reason not to do so, I'm tempted to unify Threshold and PVSSThreshold since I have a hard time envisioning a world where they are set to different values.

Yawning: Regarding the allowable values, as far as my understanding of the protocol goes, like all of these sort of commit/reveal type schemes, the output is unbiased as long as 1 participant is honest.  Naturally we will likely use something that is more than 1 in practice, but if people wish to do something less sensible that's up to them.

bennetyee: I agree that unifying the two thresholds could be reasonable.  Generally the goal is to be able to withstand a large number of dishonest nodes while allowing some honest node failures (or eclipsing).  If `PVSSThreshold` were too low then it means if there are that many dishonest nodes they can reconstruct votes of other participants in private and choosing their own $s_j$ adaptively to influence $\rho$.  So the number of dishonest nodes has to be less than `PVSSThreshold`, and the number of honest but unavailable nodes has to be less than `n - Threshold`.

I think this is accurate:

```
|---------------------|------------|-----------|-------------------|
0                                                                  n
|<- dishonest ->      |     <- honest ->                           |
| PVSSThreshold                    |                               |
|   available / Threshold                      |     down          |
```
while |dishonest| < `Threshold` < `PVSSThreshold` is feasible, it doesn't make sense since if `PVSSThreshold` nodes are available to run Reconstruction (to publish $\tilde{s}^a_j$) then they will have posted their shares with $\textsf{Open}(s^j,r_j)$ and PVSS will never need to be used.

go/common/crypto/scrape/s11n.go
```diff
@@ -0,0 +1,426 @@
...
+		return fmt.Errorf("scrape/s11n: invalid PubVerShare VG: %w", err)
+	}
+	if err := pvs.VH.isWellFormed(); err != nil {
+		return fmt.Errorf("scrape/s11n: invalid PubVerShare VH: %w", err)
+	}
+
+	return nil
+}
+
+func (pvs *PubVerShare) toKyber(index int) *pvss.PubVerShare {
```
bennetyee, line 354: precondition `pvs.isWellFormed() == nil`

go/common/crypto/scrape/scrape.go
```diff
@@ -0,0 +1,504 @@
...
+
+// Config is the configuration for an execution of the SCRAPE protocol.
+type Config struct {
+	// PrivateKey is the scalar to use as the private key.
+	PrivateKey *Scalar
+
+	// Participants is the vector of public keys of all participants
+	// in the protocol.
+	//
+	// Note: This must be in constent order across all participants,
```
bennetyee, line 45: s/constent/consistent/

go/beacon/api/scrape.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// SCRAPEBackend is a Backend that is backed by SCRAPE.
+type SCRAPEBackend interface {
+	Backend
+
+	// GetSCRAPEState gets the SCRAPE beacon round state for the
+	// provided block height.  Calling this method with height `0`
+	// should return the beacon for the latest finalized block.
+	GetSCRAPEState(context.Context, int64) (*SCRAPEState, error)
```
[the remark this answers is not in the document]
Yawning: The `height` input parameter specifies "at which block height the query should be executed" (common idiom throughout the code).  `SCRAPEState.Height` is the height at which the state was changed last.

So there's no relation between the heights beyond "The SCRAPEState that was returned, was what the consensus layer thought it was at the height that was passed in".

go/beacon/api/scrape.go
```diff
@@ -0,0 +1,144 @@
...
+
+	// WatchLatestSCRAPEEvent returns a channel that produces a
+	// stream of mesages on SCRAPE round events.  If a round
+	// transition happens before the previous round event is read
+	// from the channel, old events are overwritten.
```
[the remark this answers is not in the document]
Yawning:
```
	if ev.Epoch != w.state.Epoch || ev.Round != w.state.Round {
		w.logger.Error("epoch/round mismatch",
			"epoch", w.state.Epoch,
			"round", w.state.Round,
		)
		return false
	}
```

Like that?

go/beacon/api/scrape.go
```diff
@@ -0,0 +1,144 @@
...
+
+	Participants []signature.PublicKey `json:"participants,omitempty"`
+}
+
+func (ev *SCRAPEEvent) FromState(state *SCRAPEState) {
+	ev.Height = state.Height
+	ev.Epoch = state.Epoch
+	ev.Round = state.Round
+	ev.State = state.State
+	ev.Participants = state.Participants
```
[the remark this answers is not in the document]
Yawning: Eh, "don't do that then".  The routine's only in the API because it's used by both the consensus backend and the node worker, which both are written correctly.

The beacon in general isn't intended to be a general use externally exposed component anyway...

go/common/crypto/scrape/s11n.go
Yawning: Sure, and this is checked already...

go/consensus/tendermint/apps/beacon/api.go
```diff
@@ -1,25 +1,51 @@
...
+	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"
+	"github.com/oasisprotocol/oasis-core/go/consensus/tendermint/api"
+	beaconState "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/apps/beacon/state"
+	genesis "github.com/oasisprotocol/oasis-core/go/genesis/api"
+)
 
 const (
 	// AppID is the unique application identifier.
-	AppID uint8 = 0x04
+	AppID uint8 = 0x08
```
kostko, line 15: This should probably not change as the state prefixes are still `0x30`/`0x40`?. Also see discussion in https://github.com/oasisprotocol/oasis-core/pull/3423#discussion_r520704948. Either this or all the state key prefixes should change to `0x80` (in `beacon/state`).

go/beacon/api/scrape.go
```diff
@@ -0,0 +1,144 @@
...
+	// GasOpSCRAPECommit is the gas operation identifier for a commit.
+	GasOpSCRAPECommit transaction.Op = "scrape_commit"
+	// GasOpSCRAPEReveal is the gas operation identifier for a reveal.
+	GasOpSCRAPEReveal transaction.Op = "scrape_reveal"
+)
+
+// DefaultGasCosts are the "default" gas costs for operations.
+var DefaultGasCosts = transaction.Costs{
+	GasOpSCRAPECommit: 1000,
+	GasOpSCRAPEReveal: 1000,
```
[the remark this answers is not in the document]
Yawning: Per discussion I'm just going to remove the gas later, once the reactor is fixed (the Tx-es are functionally free anyway because the validator submits their own commit/reveal).

go/oasis-node/cmd/genesis/genesis.go
```diff
@@ -709,13 +731,19 @@ func init() {
...
+	initGenesisFlags.Bool(CfgBeaconDebugDeterministic, false, "enable deterministic beacon output (UNSAFE)")
+	initGenesisFlags.Bool(CfgBeaconDebugMockBackend, false, "use debug mock Epoch time backend")
+	initGenesisFlags.Int64(CfgBeaconInsecureTendermintInterval, 86400, "Epoch interval (in blocks)")
+	initGenesisFlags.Uint64(CfgBeaconSCRAPEParticipants, 10, "number of participants in a SCRAPE round")
+	initGenesisFlags.Uint64(CfgBeaconSCRAPEThreshold, 7, "threshold entropy contributors")
+	initGenesisFlags.Uint64(CfgBeaconSCRAPEPVSSThreshold, 7, "PVSS secret recovery threshold")
+	initGenesisFlags.Int64(CfgBeaconSCRAPECommitInterval, 43200, "SCRAPE round commit interval (in blocks)")
+	initGenesisFlags.Int64(CfgBeaconSCRAPERevealInterval, 34560, "SCRAPE round reveal interval (in blocks)")
+	initGenesisFlags.Int64(CfgBeaconSCRAPETransitionDelay, 8639, "SCRAPE round transition delay (in blocks)")
+	initGenesisFlags.StringSlice(CfgBeaconSCRAPEDebugForcedParticipant, nil, "SCRAPE forced participant public keys")
```
kostko, line 744: Hide the debug option?

go/oasis-test-runner/scenario/e2e/byzantine_beacon.go
```diff
@@ -0,0 +1,152 @@
...
+	f.Network.Beacon.SCRAPEParameters = &beacon.SCRAPEParameters{
+		DebugForcedParticipants: []signature.PublicKey{
+			pks[0],
+		},
+	}
+
+	// Make sure the byzantine node does at least 1 round (in)correctly.
+	f.ByzantineNodes[0].LogWatcherHandlerFactories = []log.WatcherHandlerFactory{
+		oasis.LogAssertEvent(byzantine.LogEventBeaconRoundCompleted, "byzantine node executed no rounds"),
+		// oasis.LogAssertEvent("fuck", "fucks not given"),
```
kostko, line 100: :grin:

kostko (commented):
Reviewed the Byzantine beacon tests, looking good!

go/beacon/api/api.go
```diff
@@ -15,41 +16,155 @@ const (
...
+
 	// GetBeacon gets the beacon for the provided block height.
-	// Calling this method with height `0`, should return the
-	// beacon for latest finalized block.
+	// Calling this method with height `consensus.HeightLatest` should
+	// return the beacon for the latest finalized block.
 	GetBeacon(context.Context, int64) ([]byte, error)
 
 	// StateToGenesis returns the genesis state at specified block height.
 	StateToGenesis(context.Context, int64) (*Genesis, error)
```
ptrus, line 72: Can we also include `ConsensusParameters(context.Context, int64) (*ConsensusParameters, error)` here, to make it consistent with other backends. #3565 added this method to some backends that were missing it (e.g. Scheduler), but for Epochtime/Beacon it was deferred for this PR (mostly because Epochtime was missing a grpc service, so it didn't made sense temporary adding it, as this PR would obsolete it anyway).

#3565 also added `EpochtimeConsensusParameters` to the consensus backend, which can be removed/replaced with the ConsesusParameters method on the new beacon backend.

Yawning: Was it really necessary to add the EpochtimeConsensusParameters stuff while I still have this WIP branch that massively refactors everything.

ptrus: > Was it really necessary

yeah before there was no way to query consensus parameters for some backends without dumping the whole state. And one of the tests needed the epochtime consensus parameters.

go/beacon/api/scrape.go
```diff
@@ -0,0 +1,146 @@
...
+	Participants uint32 `json:"participants"`
+	Threshold    uint32 `json:"threshold"`
+
+	CommitInterval  int64 `json:"commit_interval"`
+	RevealInterval  int64 `json:"reveal_interval"`
+	TransitionDelay int64 `json:"transition_delay"`
+
+	GasCosts transaction.Costs `json:"gas_costs,omitempty"`
+
+	DebugForcedParticipants []signature.PublicKey `json:"forced_participants,omitempty"`
```
kostko, line 32:
```suggestion
	DebugForcedParticipants []signature.PublicKey `json:"debug_forced_participants,omitempty"`
```

go/consensus/api/api.go
```diff
@@ -110,13 +109,13 @@ type ClientBackend interface {
...
+	WaitEpoch(ctx context.Context, epoch beacon.EpochTime) error
 
 	// GetEpoch returns the current epoch.
-	GetEpoch(ctx context.Context, height int64) (epochtime.EpochTime, error)
+	GetEpoch(ctx context.Context, height int64) (beacon.EpochTime, error)
 
-	// EpochtimeConsensusParameters returns the epochtime consensus parameters.
-	EpochtimeConsensusParameters(ctx context.Context, height int64) (*epochtime.ConsensusParameters, error)
+	// BeaconConsensusParameters returns the beacon consensus parameters.
+	BeaconConsensusParameters(ctx context.Context, height int64) (*beacon.ConsensusParameters, error)
```
kostko, line 118: This should probably be removed in favor of a beacon gRPC service. If not in this PR, please open an issue for it.

go/consensus/tendermint/apps/beacon/slashing.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	// Slash runtime node entity.
+	entityAddr := staking.NewAddress(node.EntityID)
+	totalSlashed, err := stakeState.SlashEscrow(ctx, entityAddr, &penalty.Amount)
+	if err != nil {
+		return fmt.Errorf("beacon: error slashing account %s: %w", entityAddr, err)
+	}
+	if totalSlashed.IsZero() && totalSlashed.Cmp(&penalty.Amount) != 0 {
+		return fmt.Errorf("beacon: nothing to slash from account %s", entityAddr)
```
kostko, line 101: The second condition probably shouldn't trigger an error as the penalty could be higher than the minimum amount required for a validator. Otherwise this can likely be used as a DoS vector under a specific stake distribution.

I assume the first condition (non-zero slashing) is fine since only active validators can be slashed and those need at least the minimum amount of stake (either in escrow or debonding). But in case a network would be configured with a zero threshold for validators, this could be used to trigger a chain halt.

Yawning: Well, no because the amount check boils down to "AND 0 == penalty.Amount", but I removed it anyway to support more exotic configurations.

go/consensus/api/api.go
Yawning: I added filing an issue to the post-merge TODO.

go/common/node/node.go
```diff
@@ -71,8 +72,9 @@ type Node struct { // nolint: maligned
 	// Beacon contains information for this node's participation
 	// in the random beacon protocol.
 	//
-	// NOTE: This is reserved for future use.
-	Beacon cbor.RawMessage `json:"beacon,omitempty"`
+	// TODO: This is optional for now, make mandatory once enough
```
kostko, line 75: Should we make this required in master? And optional in the backported version.

go/consensus/tendermint/apps/beacon/backend_pvss.go
```diff
@@ -0,0 +1,893 @@
...
+				goodParticipants[idx] = true
+			}
+			for idx, id := range pvssState.Participants {
+				switch goodParticipants[idx] {
+				case true:
+					// TODO: Incentivise participation?  Should there be a
+					// reward for participating in the beacon, beyond "time
+					// needs to advance for various rewards to be distributed"?
+				case false:
+					// Slash for failing to participate fully.  A node that
```
kostko, line 326: Does this take into account that the node only gets slashed in case it was able to propose a block during this period?

go/common/node/node.go
Yawning: Depends on how the transition will happen.

kostko (approved):

PR #3181, opened 2020-08-10
go/worker/storage: Add initial sync from checkpoints

### LINKED ISSUE #2479: Storage node sync should use checkpoints
After #1920 which adds a runtime-specific pruning policy which supports querying which checkpoints should be available, the storage node sync should use checkpoints when it needs to sync too far back.

~Storage checkpointing also needs to be fixed to use explicit chunks where each chunk is independently verifiable to be part of the given root. Changing this a bit will also enable us to easily use our MKVS with Tendermint's state sync (assuming we also do #1898).~ (This was done in #2659.)

jberci:
Closes #2479

### DIFF
--- a/go/oasis-test-runner/oasis/log.go
+++ b/go/oasis-test-runner/oasis/log.go
@@ -7,6 +7,7 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/log"
 	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
 	upgrade "github.com/oasisprotocol/oasis-core/go/upgrade/api"
+	workerStorage "github.com/oasisprotocol/oasis-core/go/worker/storage/committee"
 )
 
 // LogAssertEvent returns a handler which checks whether a specific log event was
@@ -81,8 +82,14 @@ func LogEventABCIPruneDelete() log.WatcherHandlerFactory {
 	return LogAssertEvent(abci.LogEventABCIPruneDelete, "expected ABCI pruning to be done")
 }
 
-// LogAssertRoothashRoothashReindexing returns a handler witch checks wether roothash reindexing was
+// LogAssertRoothashRoothashReindexing returns a handler which checks whether roothash reindexing was
 // run based on JSON log output.
 func LogAssertRoothashRoothashReindexing() log.WatcherHandlerFactory {
 	return LogAssertEvent(roothash.LogEventHistoryReindexing, "roothash runtime reindexing not detected")
+}
+
+// LogAssertCheckpointSync returns a handler which checks whether initial storage sync from
+// a checkpoint was successful or not.
+func LogAssertCheckpointSync() log.WatcherHandlerFactory {
+	return LogAssertEvent(workerStorage.LogEventCheckpointSyncSuccess, "checkpoint sync did not succeed")
 }
--- a/go/storage/api/grpc.go
+++ b/go/storage/api/grpc.go
@@ -478,7 +478,7 @@ func (c *storageClient) GetCheckpointChunk(ctx context.Context, chunk *checkpoin
 
 	for {
 		var part []byte
-		switch stream.RecvMsg(&part) {
+		switch err = stream.RecvMsg(&part); err {
 		case nil:
 		case io.EOF:
 			return nil
--- a/go/storage/client/init.go
+++ b/go/storage/client/init.go
@@ -47,14 +47,16 @@ func New(
 	schedulerBackend scheduler.Backend,
 	registryBackend registry.Backend,
 	runtime registry.RuntimeDescriptorProvider,
+	extraWatcherOpts ...committee.WatcherOption,
 ) (api.Backend, error) {
+	watcherOpts := append(extraWatcherOpts, committee.WithAutomaticEpochTransitions())
 	committeeWatcher, err := committee.NewWatcher(
 		ctx,
 		schedulerBackend,
 		registryBackend,
 		namespace,
 		scheduler.KindStorage,
-		committee.WithAutomaticEpochTransitions(),
+		watcherOpts...,
 	)
 	if err != nil {
 		return nil, fmt.Errorf("storage/client: failed to create committee watcher: %w", err)
--- a/go/storage/mkvs/checkpoint/restorer.go
+++ b/go/storage/mkvs/checkpoint/restorer.go
@@ -40,6 +40,22 @@ func (rs *restorer) StartRestore(ctx context.Context, checkpoint *Metadata) erro
 	return nil
 }
 
+func (rs *restorer) AbortRestore(ctx context.Context) error {
+	rs.Lock()
+	defer rs.Unlock()
+
+	if rs.currentCheckpoint == nil {
+		return ErrNoRestoreInProgress
+	}
+
+	rs.pendingChunks = nil
+	rs.currentCheckpoint = nil
+
+	// TODO: also properly remove leftover nodes from the database (#3241).
+
+	return nil
+}
+
 func (rs *restorer) GetCurrentCheckpoint() *Metadata {
 	rs.Lock()
 	defer rs.Unlock()
--- a/go/worker/storage/committee/checkpoint_sync.go
+++ b/go/worker/storage/committee/checkpoint_sync.go
@@ -1,0 +1,460 @@
+package committee
+
+import (
+	"bytes"
+	"container/heap"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"sort"
+	"sync"
+	"time"
+
+	"github.com/cenkalti/backoff/v4"
+
+	"github.com/oasisprotocol/oasis-core/go/runtime/committee"
+	schedulerApi "github.com/oasisprotocol/oasis-core/go/scheduler/api"
+	storageApi "github.com/oasisprotocol/oasis-core/go/storage/api"
+	storageClient "github.com/oasisprotocol/oasis-core/go/storage/client"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/checkpoint"
+)
+
+const (
+	retryInterval = 1 * time.Second
+	maxRetries    = 30
+
+	checkpointStatusDone = 0
+	checkpointStatusNext = 1
+	checkpointStatusBail = 2
+
+	// LogEventCheckpointSyncSuccess is a log event value that signals that checkpoint sync was successful.
+	LogEventCheckpointSyncSuccess = "worker/storage/checkpoint-sync-success"
+)
+
+// ErrNoUsableCheckpoints is the error returned when none of the checkpoints could be synced.
+var ErrNoUsableCheckpoints = errors.New("storage: no checkpoint could be synced")
+
+type restoreResult struct {
+	done bool
+	err  error
+}
+
+type chunkHeap struct {
+	array  []*checkpoint.ChunkMetadata
+	length int
+}
+
+func (h chunkHeap) Len() int           { return h.length }
+func (h chunkHeap) Less(i, j int) bool { return h.array[i].Index < h.array[j].Index }
+func (h chunkHeap) Swap(i, j int)      { h.array[i], h.array[j] = h.array[j], h.array[i] }
+
+func (h *chunkHeap) Push(x interface{}) {
+	h.array[h.length] = x.(*checkpoint.ChunkMetadata)
+	h.length++
+}
+
+func (h *chunkHeap) Pop() interface{} {
+	h.length--
+	ret := h.array[h.length]
+	h.array[h.length] = nil
+	return ret
+}
+
+// goWithCommittee runs the given operation with all the connections to the storage committee.
+func (n *Node) goWithCommittee(
+	committeeClient committee.Client,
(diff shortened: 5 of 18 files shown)


go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+	}
+}
+
+func (n *Node) syncCheckpoint() error {
+	// Fetch metadata first from the current committee.
+	req := &checkpoint.GetCheckpointsRequest{
+		Namespace: n.commonNode.Runtime.ID(),
+	}
+	metadata, err := n.storageClient.GetCheckpoints(n.ctx, req)
```
kostko, line 219: We should fetch checkpoints from multiple nodes and combine all of them. Otherwise this could end up with a bad storage node and you wouldn't be able to sync.

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	committeeClient, err := committee.NewClient(n.ctx, committeeWatcher.Nodes(), committee.WithClientAuthentication(n.commonNode.Identity))
+	if err != nil {
+		return fmt.Errorf("can't create committee client: %w", err)
+	}
+
+	// Try all the checkpoints now, from most recent backwards.
+	for _, check := range metadata {
+		if !n.localStorage.NodeDB().HasRoot(check.Root) {
+			n.logger.Debug("checkpoint doesn't have a known root", "root", check.Root)
```
kostko, line 249: Hm, this sounds wrong, you will only sync checkpoints for which you already have all the data?

You probably wanted to use `n.commonNode.Runtime.History().GetBlock` to check if you have the block and at the same time, you should filter out all the checkpoints which are lower or equal than `n.syncedState.LastBlock.Round` (because for those you already have all the data synced).

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -0,0 +1,268 @@
...
+
+		// GetCheckpointChunk errors.
+		switch {
+		case err == nil:
+		case errors.Is(err, checkpoint.ErrChunkNotFound):
+			errorCh <- checkpointStatusNext
+			return
+		default:
+			n.logger.Error("can't fetch chunk from storage node", "node", clientConn.Node.ID, "chunk", chunk.Index)
+			// XXX retry
```
kostko, line 99: Yeah there should be some maximum number of retries per node.

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -0,0 +1,268 @@
...
+			return
+		}
+
+		// RestoreChunk errors
```
kostko, line 103:
```suggestion
		// RestoreChunk errors.
```

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -0,0 +1,268 @@
...
+			}
+		}()
+		err := conn.GetCheckpointChunk(ctx, chunk, wr)
+		result := <-restoreCh
+
+		// GetCheckpointChunk errors.
+		switch {
+		case err == nil:
+		case errors.Is(err, checkpoint.ErrChunkNotFound):
+			errorCh <- checkpointStatusNext
```
kostko, line 95: This doesn't seem right, if the chunk is not found, the storage node could be faulty, so the chunk should be returned and given to a different node?

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -0,0 +1,268 @@
...
+			chunkReturnCh <- nil
+			return
+		case errors.Is(result.err, checkpoint.ErrChunkCorrupted):
+			chunkReturnCh <- chunk
+			return
+		case errors.Is(result.err, checkpoint.ErrChunkProofVerificationFailed):
+			errorCh <- checkpointStatusNext
+			return
+		case result.err != nil:
+			n.logger.Error("chunk restoration failed", "node", clientConn.Node.ID, "chunk", chunk.Index, "err", result.err)
```
kostko, line 116: What kinds of failures can this be? Should you abort the checkpoint sync process in this case?

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -0,0 +1,268 @@
...
+		var ok bool
+		select {
+		case <-ctx.Done():
+			return
+		case chunk, ok = <-chunkDispatchCh:
+			if !ok {
+				return
+			}
+		}
+
```
kostko, line 78: I think currently `chunk` can be nil here (see my other comment), but this is not handled?

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -0,0 +1,268 @@
...
+			if returned == nil {
+				// Restoration completed, no more chunks.
+				return checkpointStatusDone, nil
+			}
+			heap.Push(chunks, returned)
+
+		case status := <-errorCh:
+			return status, nil
+
+		case outChan <- next:
```
kostko, line 200: Should this only try sending `next` if it's not nil? Otherwise it seems like it will keep sending nils? And if it sends a nil it will currently crash.

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -0,0 +1,268 @@
...
+			next = nil
+
+		case <-workersDone:
+			// No usable committee connections left, bail.
+			return checkpointStatusBail, storageClient.ErrStorageNotAvailable
```
kostko, line 205: This could in theory be related to a checkpoint being faulty, so I think we should rather return `checkpointStatusNext` here?

go/worker/storage/committee/node.go
```diff
@@ -660,11 +660,32 @@ func (n *Node) worker() { // nolint: gocyclo
...
+	registeredCh := make(chan interface{})
+	n.roleProvider.SetAvailableWithCallback(func(nd *node.Node) error {
 		nd.AddOrUpdateRuntime(n.commonNode.Runtime.ID())
 		return nil
+	}, func(ctx context.Context) error {
+		close(registeredCh)
+		return nil
 	})
 
+	// Wait for the registration to finish, because we'll need to ask
```
kostko, line 672: Also log (in debug level) that we are waiting for our registration to finish.

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -0,0 +1,390 @@
...
+			retList[cursor] = list[i]
+			cursor++
+		}
+	}
+
+	return retList[:cursor], nil
+}
+
+func (n *Node) checkCheckpointUsable(cp *checkpoint.Metadata) bool {
+	if n.syncedState.LastBlock.IORoot.Namespace.Equal(&cp.Root.Namespace) && n.syncedState.LastBlock.IORoot.Version >= cp.Root.Version {
```
kostko, line 323: Doesn't accessing `n.syncedState` require holding a lock (if caller should lock, rename the method to `...Locked`)? Also it would be more clear if namespace comparisons were against ` n.commonNode.Runtime.ID()`?

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -0,0 +1,390 @@
...
+		// We already have this root.
+		return false
+	}
+	blk, err := n.commonNode.Runtime.History().GetBlock(n.ctx, cp.Root.Version)
+	if err != nil {
+		n.logger.Error("can't get block information for checkpoint, skipping", "err", err, "root", cp.Root)
+		return false
+	}
+	if blk.Header.Namespace.Equal(&cp.Root.Namespace) && blk.Header.IORoot.Equal(&cp.Root.Hash) {
```
kostko, line 332: Why is this only checking IO root checkpoints? Shouldn't there also be state root checkpoints? Those are actually the more important ones.

go/oasis-test-runner/oasis/log.go
```diff
@@ -75,8 +76,14 @@ func LogEventABCIPruneDelete() log.WatcherHandlerFactory {
...
 }
 
-// LogAssertRoothashRoothashReindexing returns a handler witch checks wether roothash reindexing was
+// LogAssertRoothashRoothashReindexing returns a handler which checks whether roothash reindexing was
 // run based on JSON log output.
 func LogAssertRoothashRoothashReindexing() log.WatcherHandlerFactory {
 	return LogAssertEvent(roothash.LogEventHistoryReindexing, "roothash runtime reindexing not detected")
 }
+
+// LogAssertCheckpointSyncreturns a handler which checks whether initial storage sync from
```
kostko, line 85:
```suggestion
// LogAssertCheckpointSync returns a handler which checks whether initial storage sync from
```

kostko (commented):
Tests seem to be failing?

go/storage/api/grpc.go
```diff
@@ -478,7 +478,7 @@ func (c *storageClient) GetCheckpointChunk(ctx context.Context, chunk *checkpoin
 
 	for {
 		var part []byte
-		switch stream.RecvMsg(&part) {
+		switch err = stream.RecvMsg(&part); err {
```
kostko, line 481: :facepalm:

go/worker/storage/committee/node.go
```diff
@@ -649,14 +661,45 @@ func (n *Node) worker() { // nolint: gocyclo
...
+		// Try syncing again. The main reason for this is the sync failing due to a
+		// checkpoint pruning race condition (where nodes list a checkpoint which is
+		// then deleted just before we request its chunks). One retry is enough.
+		n.logger.Info("first checkpoint sync failed, trying once more", "err", err)
+		err = n.syncCheckpoints()
+	}
+	if err != nil {
+		n.logger.Info("checkpoint sync failed", "err", err)
+	} else {
+		n.logger.Info("checkpoint sync succeeded",
```
kostko, line 697: Maybe also add some log keys related to the current synced round?

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -0,0 +1,432 @@
...
+	}
+
+	cancel, doneCh, err := n.goWithCommittee(committeeClient, worker)
+	if err != nil {
+		return checkpointStatusBail, fmt.Errorf("can't fetch chunks from committee nodes: %w", err)
+	}
+	defer cancel()
+
+	if err := n.localStorage.Checkpointer().StartRestore(n.ctx, check); err != nil {
+		return checkpointStatusNext, fmt.Errorf("can't start checkpoint restore: %w", err)
```
kostko, line 210: This should probably result in aborting the checkpoint process.

go/storage/mkvs/checkpoint/restorer.go
```diff
@@ -40,6 +40,22 @@ func (rs *restorer) StartRestore(ctx context.Context, checkpoint *Metadata) erro
...
+	defer rs.Unlock()
+
+	if rs.currentCheckpoint == nil {
+		return ErrNoRestoreInProgress
+	}
+
+	rs.pendingChunks = nil
+	rs.currentCheckpoint = nil
+
+	// TODO: also properly remove leftover nodes from the database.
```
kostko, line 54: Open an issue and reference it here.

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -0,0 +1,460 @@
...
+
+	err = n.localStorage.Checkpointer().StartRestore(n.ctx, check)
+	switch {
+	case err == nil:
+		break
+	case errors.Is(err, checkpoint.ErrRestoreAlreadyInProgress):
+		n.logger.Info("a restore is already in progress, trying to abort")
+		err = n.localStorage.Checkpointer().AbortRestore(n.ctx)
+		if err != nil {
+			return checkpointStatusNext, fmt.Errorf("can't abort previous checkpoint restore: %w", err)
```
kostko, line 217: As below, this should probably abort the checkpoint process.

go/storage/mkvs/checkpoint/restorer.go
jberci: Issue #3241

kostko: I meant reference it in the above TODO comment.

jberci: Also nice to have it here tho, for more context.

go/storage/client/init.go
```diff
@@ -54,6 +54,7 @@ func New(
 		namespace,
 		scheduler.KindStorage,
 		committee.WithAutomaticEpochTransitions(),
+		committee.WithFilter(committee.IgnoreNodeFilter(ident.NodeSigner.Public())),
```
kostko, line 57: Actually, this should probably not be done in general (e.g., a compute node that is also a storage node may need to commit to self), only specifically for the storage node worker connecting to other nodes. You could make `New` take `committeeOpts ...committee.WatcherOption` and only enable it in the storage node worker.

kostko:
Could you add a configuration flag, e.g. `worker.storage.checkpoint_sync` (which defaults to `true`) so that we can disable checkpoint sync (e.g., in tests).

kostko (approved):

PR #3183, opened 2020-08-11
oasis-node/cmd: Add consensus benchmark subcommand

abukosek:


pro-wh:
a little reminiscent of https://github.com/oasisprotocol/oasis-core/blob/master/go/oasis-node/cmd/debug/txsource/workload/parallel.go the parallel workload. this is more robust to failures and adds measurements so we can get a transaction result out. a long time ago we were using a scheme where we'd take the middle 80% of the transactions to get the peak rate.

abukosek:
Thanks for taking a look at this.  I can add the transaction cutting, but IMO it's not needed -- in my test runs, only the first block at the start of the benchmark had fewer transactions, the other blocks were always full.  As long as there are enough samples (so that the duration of the benchmark is long enough), it won't significantly affect the results.

pro-wh:
some usability issues from trying to run this today:

- still tries to load signer even when running with skip_funding. shouldn't need it, I think. I worked around this by enabling that test entity flag
- too scary to run this with no feedback that it's working
- I'd suggest not doing gas estimation in the benchmark loop
- something crashes on integer divide by zero when no transaction succeeds. could we be more graceful?
- I think we shouldn't automatically and immediately delete the output log. and if we keep them, we should name them with the parameters, which is currently the number of accounts

abukosek:
Thanks for the feedback, I'll fix these issues tomorrow!
Gas estimation is only done once per account (or whenever there's an error), but I suppose it would make more sense to pre-set the cached value to the one obtained at the beginning, since it's unlikely to change.

Issue #1677, opened 2019-05-03
Internal transaction retry


pro-wh:
The compute committee used to have an in internal retry mechanism to return re-queue transactions from an aborted batch. This is not implemented in the updated system with separate transaction scheduler and compute worker.

## Details

some alternatives of how it might work:

* transaction scheduler watches over the compute node to see which batches get aborted. could be nice to leave the transaction conceptually in the transaction scheduler's custody even after it is submitted to the compute committee.
* the compute nodes actively send them back to the transaction scheduler
  * dedupe in transaction scheduler to handle multiple compute nodes doing this, as well as client retries?
  * use libp2p? we already track the transaction scheduler leader's p2p address. or maybe connect through the client QueueCall method

## Acceptance Criteria
* If we reimplement this, a transaction should survive a transient batch abort without end-to-end retry.

Issue #1654, opened 2019-04-26
Shorten transaction scheduler leader 


pro-wh:
From #1640, the transaction scheduler committee has a single leader for an entire general purpose epoch. We need it to rotate on a shorter interval.

## Details

Two current proposals:
1. Change leader among the committee on a new "sub-epoch" system
2. Let different kinds of committees have different epoch times, with the transaction scheduler committee having a shorter epoch time (related to storage committees too #1585 )

Clients will have to be aware of this too.

blocks #1577

## Acceptance Criteria
* we can shorten how long a node will be transaction scheduler leader without shortening the compute committee lifetime

PR #3184, opened 2020-08-11
go/worker/txnscheduler: Support multiple txnscheduler workers

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/1654 
Fixes: https://github.com/oasisprotocol/oasis-core/issues/1677

Adds support for multiple txnschedulers in runtime committees:
- txnscheduler grpc endpoint removed, the per-runtime libp2p gossipsub is used for submitting transactions
- current implementation: ~added `LeadersInterval` runtime parameter~ - txnschedulers switch between leaders in round-robin fashion (switching on each runtime round)
- leader role removed from scheduler api
- txnscheduler worker removed (merged into exectuor)
- txnscheduler committee removed 

TODO: 
- [x] actually following on #3179 drop support for txnschedulers
- [x] cleanup - see comments in code.
- [x] byzantine-node tests
- [x] ~add support for roothash timeout in case the current leader is not submitting batches~ - open issue, will handle in separate PR
  - https://github.com/oasisprotocol/oasis-core/issues/3199


go/worker/common/p2p/dispatch.go
```diff
@@ -153,13 +156,15 @@ func (h *topicHandler) dispatchMessage(peerID core.PeerID, m *queuedMsg, isIniti
 	// Perhaps this should reject the message, but it is possible that
 	// the local node is just behind.  This does result in stale messages
 	// getting retried though.
-	if err = h.handler.AuthenticatePeer(m.from, m.msg); err != nil {
-		return err
-	}
+	for _, handler := range h.handlers {
```
kostko, line 159: No lock on the handlers? The `topicHandler` should probably have a RW lock and the RLock should be taken here while a Lock should be taken in `RegisterHandler`.

go/worker/common/p2p/p2p.go
```diff
@@ -111,13 +111,21 @@ func (p *P2P) RegisterHandler(runtimeID common.Namespace, handler Handler) {
 	p.Lock()
 	defer p.Unlock()
```
kostko, line 113: Consider updating the method documentation to specify what happens if multiple handlers are registered for a runtime.

go/worker/common/p2p/dispatch.go
```diff
@@ -153,13 +156,15 @@ func (h *topicHandler) dispatchMessage(peerID core.PeerID, m *queuedMsg, isIniti
...
+		if err = handler.AuthenticatePeer(m.from, m.msg); err != nil {
```
kostko, line 160: This should probably be split into two traversals, first we authenticate the peer using all handlers and stop if any fails. Otherwise partial dispatches could be possible.

Then we dispatch the message to all handlers. The question is should we always dispatch to all first before returning an error in case one of the handlers failed or not?

go/worker/common/p2p/noop.go
```diff
@@ -0,0 +1,19 @@
+package p2p
+
+import (
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
+)
+
+// NoOp P2P Handler that can be used in publishing/relaying-only clients.
+type NoOp struct {
```
kostko, line 8: Maybe rename to `NoopHandler` or `BaseHandler`.

go/worker/common/p2p/peermgmt.go
```diff
@@ -92,7 +92,11 @@ func (mgr *PeerManager) isPeerAuthorized(id core.PeerID) bool {
 	mgr.RLock()
 	defer mgr.RUnlock()
 
-	return mgr.peers[id] != nil
+	// TODO: if we're txnscheduler anyone can publish, unless we would use a
+	// separate gossipsub network for that.
+	// TODO: add a settable 'allowAll' flag, that we can toggle when we
```
kostko, line 97: Not sure if this is a good idea as it could result in a flaky gossipsub topology.

go/worker/common/p2p/peermgmt.go
```diff
@@ -92,7 +92,11 @@ func (mgr *PeerManager) isPeerAuthorized(id core.PeerID) bool {
 	mgr.RLock()
 	defer mgr.RUnlock()
 
-	return mgr.peers[id] != nil
+	// TODO: if we're txnscheduler anyone can publish, unless we would use a
```
kostko, line 95: Hm, right, this makes it more likely that random people can mess with libp2p. But I guess that's what the (not yet implemented) P2P proxy/sentry is for. @Yawning any thoughts?

go/worker/compute/txnscheduler/committee/node.go
```diff
@@ -139,9 +141,76 @@ func (n *Node) getMetricLabels() prometheus.Labels {
...
+		n.commonNode.CrossNode.Lock()
+		height := n.commonNode.CurrentHeight
+		n.commonNode.CrossNode.Unlock()
+		if !epochSnapshot.IsTransactionSchedulerLeader(height) {
+			n.logger.Error("unable to handle message, not leader anymore",
+				"err", api.ErrNotLeader,
+				"current_epoch", epochSnapshot.GetEpochNumber(),
+				"current_height", height,
+			)
+			return true, p2pError.Permanent(api.ErrNotLeader)
```
kostko, line 162: This should probably return a `nil` error as you do want to relay the message instead of dropping it.

go/worker/compute/txnscheduler/committee/node.go
```diff
@@ -139,9 +141,76 @@ func (n *Node) getMetricLabels() prometheus.Labels {
...
+// QueueCall queues a runtime call.
+func (n *Node) QueueCall(call []byte) error {
+	n.algorithmMutex.RLock()
+	defer n.algorithmMutex.RUnlock()
+
+	if n.algorithm == nil || !n.algorithm.IsInitialized() {
+		return api.ErrNotReady
+	}
+	if err := n.algorithm.ScheduleTx(call); err != nil {
+		return err
```
kostko, line 206: Failing to queue (e.g., due to already being in the queue) should prevent relaying (e.g., a permanent P2P error). Currently I see that the batching algorithm does not propagate the error, but since there are no more client acks, it would be better to just return an error and make it permanent here.

go/worker/common/p2p/peermgmt.go
Yawning: I wouldn't use `allowAll`.

go/worker/common/p2p/peermgmt.go
Yawning: Oh gross.  What's "all nodes" in this context, the entire internet?

kostko: Yeah, so previously it was "the entire internet can connect to txnscheduler gRPC" now it is "the entire internet can connect to libp2p". It is more scary I guess, but not sure how it could be avoided as there needs to be a way for anyone to submit transactions?

ptrus: Yeah, well technically the nodes could still limit access on a different layer to trusted "client nodes" (e.g. firewalling ... would in any case make sense at least in the case of permissioned runtimes - but could do the same for others as well) - i guess the only way to avoid that would be having clients also needing to register.

kostko: And everything available publicly should run behind proxies (we don't have libp2p sentries yet though).

go/consensus/tendermint/apps/roothash/transactions.go
```diff
@@ -123,7 +154,7 @@ func (app *rootHashApplication) executorCommit(
 	}
 
 	for _, commit := range cc.Commits {
-		if err = rtState.Round.AddExecutorCommitment(ctx, &commit, sv, nl); err != nil { // nolint: gosec
+		if err = rtState.Round.AddExecutorCommitment(ctx, app.state.BlockHeight(), &commit, sv, nl); err != nil { // nolint: gosec
```
kostko, line 157: Use `ctx.BlockHeight()` (since you already pass `ctx` the called method could already access that).

go/consensus/tendermint/apps/scheduler/scheduler.go
```diff
@@ -487,10 +483,11 @@ func (app *schedulerApplication) electCommittee(
...
 	err = schedulerState.NewMutableState(ctx.State()).PutCommittee(ctx, &scheduler.Committee{
-		Kind:      kind,
-		RuntimeID: rt.ID,
-		Members:   members,
-		ValidFor:  epoch,
+		Kind:           kind,
+		RuntimeID:      rt.ID,
+		Members:        members,
+		ValidFor:       epoch,
+		LeaderInterval: leaderInterval,
```
kostko, line 490: Yeah this can probably be removed as it just duplicates the configuration value from the runtime descriptor. Unless we would want the scheduler to be able to adjust this interval, but I don't see why we would need that.

go/runtime/client/watcher.go
```diff
@@ -115,30 +118,47 @@ func (w *blockWatcher) watch() {
...
 		case blk := <-blocks:
 			current = blk.Block
 			height = blk.Height
 
+		case blk := <-consBlocks:
+			if blk == nil {
+				break
+			}
+			// TODO: maybe make this more proper - remember for each request
+			// at which height it was made, and retry after X intervals?
```
kostko, line 150: Yeah that would be better.

go/runtime/client/watcher.go
```diff
@@ -216,17 +238,13 @@ func (w *blockWatcher) Stop() {
 	close(w.stopCh)
 }
 
-func newWatcher(common *clientCommon, id common.Namespace) (*blockWatcher, error) {
+func newWatcher(common *clientCommon, id common.Namespace, p2pSvc *p2p.P2P) (*blockWatcher, error) {
 	committeeWatcher, err := committee.NewWatcher(
```
kostko, line 242: Why is the committee watcher (and committee client) even used now (since tx dispatch happens over p2p anyway)?

go/worker/common/committee/node.go
```diff
@@ -108,8 +108,9 @@ type Node struct {
 
 	// Mutable and shared between nodes' workers.
 	// Guarded by .CrossNode.
-	CrossNode    sync.Mutex
-	CurrentBlock *block.Block
+	CrossNode          sync.Mutex
+	CurrentBlock       *block.Block
+	CurrentBlockHeight int64
```
kostko, line 113: This is slightly confusing as the `CurrentBlock` above refers to the runtime's block, while this block height refers to the current consensus layer block.

ptrus:
> So this removes the transaction scheduler worker but not the transaction scheduler committee? There is no need for a separate committee AFAIK, it makes more sense that one of the executor nodes is the transaction scheduler (proposer).

yeah but removing the committee will be straightforward so i left it for after i have the rest done :-)

go/consensus/tendermint/apps/scheduler/scheduler.go
ptrus: yeah will remove, there's a todo somewhere :)

go/runtime/client/watcher.go
ptrus: probably not needed, will remove

go/roothash/api/commitment/pool.go
```diff
@@ -205,8 +210,8 @@ func (p *Pool) addOpenExecutorCommitment(
 	}
 
 	// Verify that the txn scheduler signature for current commitment is valid.
-	if err := sv.VerifyCommitteeSignatures(scheduler.KindComputeTxnScheduler, []signature.Signature{body.TxnSchedSig}); err != nil {
-		logger.Debug("executor commitment has bad transaction scheduler signers",
+	if err := sv.VerifyTxnSchedulerSignature(body.TxnSchedSig, height); err != nil {
```
kostko, line 213: The pool should probably remember the height at which the first valid commitment was received for the current round and then use that for subsequent commitments (until reset at end of round)?

go/worker/compute/executor/committee/node.go
```diff
@@ -188,22 +214,65 @@ func (n *Node) getMetricLabels() prometheus.Labels {
...
 		// Before opening the signed dispatch message, verify that it was
 		// actually signed by the current transaction scheduler.
-		epoch := n.commonNode.Group.GetEpochSnapshot()
-		txsc := epoch.GetTransactionSchedulerCommittee()
-		if !txsc.PublicKeys[sbd.Signature.PublicKey] {
+		n.commonNode.CrossNode.Lock()
+		height := n.commonNode.CurrentHeight
+		n.commonNode.CrossNode.Unlock()
+		txsc := epoch.GetTransactionSchedulerLeader(height)
+		if !txsc.PublicKey.Equal(sbd.Signature.PublicKey) {
```
kostko, line 269: This should probably use `epoch.VerifyTxnSchedulerSignature`.

kostko: Also for the height this needs to take into account the current "round height" (if any), similar to what the pool does. The easiest way to do that is to first update `ExecutorCommittedEvent` to include a `CurrentRoundHeight` and then make sure that the executor node handles the `ExecutorCommittedEvent` event in `HandleNewEventLocked` and updates its view of the round height. If there is no current round height, it should use the current height.

go/worker/compute/executor/committee/node.go
```diff
@@ -290,15 +360,20 @@ func (n *Node) queueBatchBlocking(
...
-func (n *Node) HandleBatchFromTransactionSchedulerLocked(
+func (n *Node) handleInternalBatchLocked(
 	batchSpanCtx opentracing.SpanContext,
 	ioRoot hash.Hash,
 	batch transaction.RawBatch,
 	txnSchedSig signature.Signature,
 	inputStorageSigs []signature.Signature,
-) {
+) error {
+	// If we are not waiting for a batch, don't do anything.
```
kostko, line 372: Why is this needed now?

go/worker/compute/executor/committee/txnscheduling/algorithm/batching/batching.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
 	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
-	"github.com/oasisprotocol/oasis-core/go/worker/compute/txnscheduler/algorithm/api"
+	"github.com/oasisprotocol/oasis-core/go/worker/compute/executor/committee/txnscheduling/algorithm/api"
```
kostko, line 11: Maybe consider moving the `algorithm` package to `go/runtime/scheduling` (and rename to `scheduling`) as it seems quite independent from the actual worker.

go/worker/compute/executor/committee/node.go
ptrus: indeed not needed anymore since we already do the check in `Dispatch`

go/registry/api/runtime.go
```diff
@@ -98,6 +98,11 @@ type TxnSchedulerParameters struct {
 	// GroupSize is the size of the committee.
 	GroupSize uint64 `json:"group_size"`
```
kostko, line 99: Remove group size?

go/roothash/api/commitment/pool.go
```diff
@@ -243,6 +270,14 @@ func (p *Pool) addOpenExecutorCommitment(
 	if p.ExecuteCommitments == nil {
 		p.ExecuteCommitments = make(map[signature.PublicKey]OpenExecutorCommitment)
 	}
+	// If this is the first commitment for the height, save it.
```
kostko, line 266:
```suggestion
	// If this is the first commitment for the round, save it.
```

go/runtime/client/client.go
```diff
@@ -497,15 +414,18 @@ func (c *runtimeClient) Cleanup() {
 func New(
 	ctx context.Context,
 	dataDir string,
+	identity *identity.Identity,
```
kostko, line 417: Where is identity used?

go/worker/compute/executor/committee/node.go
```diff
@@ -424,6 +485,229 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
+		return errNotReady
+	}
+	resp, err := rt.Call(ctx, checkRq)
+	switch {
+	case err != nil:
+		n.logger.Error("CheckTx: runtime call error",
+			"err", err,
+		)
+		return err
+	case resp == nil:
```
kostko, line 516: This can never be nil (if it was `err` would have an error), but the underlying message can be:
```suggestion
	case resp.RuntimeCheckTxBatchResponse == nil:
```

go/worker/compute/executor/committee/node.go
```diff
@@ -424,6 +485,229 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
+		return err
+	}
+
+	incomingQueueSize.With(n.getMetricLabels()).Set(float64(n.txnSchedulingAlgorithm.UnscheduledSize()))
+
+	return nil
+}
+
+// RemoveTxBatch removes a runtime batch.
+func (n *Node) RemoveTxBatch(batch [][]byte) error {
```
kostko, line 566: Why is this public?

go/worker/compute/executor/committee/node.go
```diff
@@ -424,6 +485,229 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
+		n.logger.Error("CheckTx: runtime failed with error",
+			"err", result.Error,
+		)
+		return fmt.Errorf("%w: %s", errCheckTxFailed, *result.Error)
+	}
+
+	return nil
+}
+
+// QueueTx queues a runtime tx.
```
kostko, line 548:
```suggestion
// QueueTx queues a runtime transaction for scheduling.
```

go/worker/compute/executor/committee/node.go
```diff
@@ -424,6 +485,229 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
+	if _, ok := n.state.(StateWaitingForBatch); !ok {
+		return errIncorrectState
+	}
+	if n.commonNode.CurrentBlock == nil {
+		return errNoBlocks
+	}
+
+	epoch := n.commonNode.Group.GetEpochSnapshot()
+	round := n.commonNode.CurrentBlock.Header.Round
+	// If we are not a leader or we don't have any blocks, don't do anything.
```
kostko, line 597:
```suggestion
	// If we are not a leader don't do anything.
```

go/worker/compute/executor/committee/txnscheduling/algorithm/batching/incoming_queue.go
```diff
@@ -86,6 +86,28 @@ func (q *incomingQueue) addCallLocked(call []byte, callHash hash.Hash) {
...
+	defer q.Unlock()
+
+	var newQueue [][]byte
+	for _, call := range batch {
+		callHash := hash.NewFromBytes(call)
+		if _, ok := q.callHashes[callHash]; !ok {
+			continue
+		}
+		delete(q.callHashes, callHash)
+		for _, existing := range q.queue {
```
kostko, line 100: This seems like an expensive way to update the list as it seems like O(n^2). You could optimize it to O(2n) by just moving this loop out (since you check against `q.callHashes` anyway). Hmm, cctually it looks like this would duplicate transactions in the queue?

go/worker/compute/executor/committee/txnscheduling/algorithm/batching/incoming_queue.go
```diff
@@ -86,6 +86,28 @@ func (q *incomingQueue) addCallLocked(call []byte, callHash hash.Hash) {
...
+			existingHash := hash.NewFromBytes(call)
+			if _, ok := q.callHashes[existingHash]; ok {
+				newQueue = append(newQueue, existing)
+			}
+		}
+	}
+	q.queue = newQueue
```
kostko, line 107: This does not recompute `q.queueSizeBytes`.

go/worker/compute/executor/init.go
```diff
@@ -1,16 +1,42 @@
...
 	workerCommon "github.com/oasisprotocol/oasis-core/go/worker/common"
 	"github.com/oasisprotocol/oasis-core/go/worker/compute"
+	txnSchedulerAlgorithm "github.com/oasisprotocol/oasis-core/go/worker/compute/executor/committee/txnscheduling/algorithm"
 	"github.com/oasisprotocol/oasis-core/go/worker/registration"
 )
 
+const (
+	// CfgScheduleCheckTxEnabled enables checking each transaction before
+	// scheduling it.
+	CfgScheduleCheckTxEnabled = "worker.exectutor.schedule_check_tx.enabled"
```
kostko, line 16:
```suggestion
	CfgScheduleCheckTxEnabled = "worker.executor.schedule_check_tx.enabled"
```

go/worker/compute/executor/init.go
```diff
@@ -1,16 +1,42 @@
...
+)
+
+// Flags has the configuration flags.
+var Flags = flag.NewFlagSet("", flag.ContinueOnError)
+
+// ScheduleCheckTxEnabled reads CfgScheduleCheckTxEnabled flag from viper.
+func ScheduleCheckTxEnabled() bool {
```
kostko, line 23: Is there a reason why this needs to be public?

ptrus: hm it was public before, but yeah doesn't seem like it needs to be.

.changelog/3184.cfg.md
```diff
@@ -0,0 +1,5 @@
+Added `worker.p2p.enabled` flag to explicitly enable P2P worker.
+
+Note: compute workers will automatically enable the P2P worker. The P2P worker
+needs to be manually enabled on nodes that need submitting runtime
```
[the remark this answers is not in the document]
ptrus: `--runtime.supported` is also configured for storage nodes, which don't need `p2p`, that's why i chose not to use that.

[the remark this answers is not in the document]
kostko: I see, ok let's leave it as-is for now.

go/worker/compute/executor/committee/txnscheduling/algorithm/batching/incoming_queue.go
ptrus: yes of course (it was late when i pushed this :grin:)

go/worker/compute/executor/committee/node.go
```diff
@@ -187,23 +213,61 @@ func (n *Node) getMetricLabels() prometheus.Labels {
 }
 
 // HandlePeerMessage implements NodeHooks.
-func (n *Node) HandlePeerMessage(ctx context.Context, message *p2p.Message) (bool, error) {
-	if message.TxnSchedulerBatch != nil {
+func (n *Node) HandlePeerMessage(ctx context.Context, message *p2p.Message, isOwn bool) (bool, error) {
+	n.logger.Debug("received peer message", "message", message, "isOwn", isOwn)
```
kostko, line 217:
```suggestion
	n.logger.Debug("received peer message", "message", message, "is_own", isOwn)
```

go/worker/compute/executor/committee/node.go
```diff
@@ -187,23 +213,61 @@ func (n *Node) getMetricLabels() prometheus.Labels {
...
+		if err != nil {
+			n.logger.Error("unable to Queue transaction",
+				"err", err,
+			)
+			return true, p2pError.Permanent(err)
+		}
+		return true, nil
+
+	case message.TxnSchedulerBatch != nil:
+		// Ignore own messages.
```
kostko, line 249:
```suggestion
		// Ignore own messages as those are handled via handleInternalBatchLocked.
```

go/common/version/version.go
```diff
@@ -68,15 +68,15 @@ var (
 
 	// RuntimeCommitteeProtocol versions the P2P protocol used by the runtime
 	// committee members.
-	RuntimeCommitteeProtocol = Version{Major: 0, Minor: 11, Patch: 0}
+	RuntimeCommitteeProtocol = Version{Major: 0, Minor: 12, Patch: 0}
```
kostko, line 71: This was already bumped since 20.9, so no need to bump it again.

go/common/version/version.go
```diff
@@ -68,15 +68,15 @@ var (
...
 
 	// ConsensusProtocol versions all data structures and processing used by
 	// the epochtime, beacon, registry, roothash, etc. modules that are
 	// backend by consensus.
 	//
 	// NOTE: Any change in the major or minor versions are considered
 	//       breaking changes for the protocol.
-	ConsensusProtocol = Version{Major: 0, Minor: 27, Patch: 0}
+	ConsensusProtocol = Version{Major: 0, Minor: 28, Patch: 0}
```
kostko, line 79: This was already bumped since 20.9, so no need to bump it again.

.changelog/3184.cfg.md
```diff
@@ -0,0 +1,5 @@
+Added `worker.p2p.enabled` flag to explicitly enable P2P worker
```
kostko, line 1: File should probably be named `.cfg.1.md` (as you now also have the .2).

go/roothash/api/commitment/pool_test.go
```diff
@@ -795,14 +805,14 @@ func generateStorageReceiptSignature(t *testing.T, blk *block.Block, body *Compu
...
-	dispatch := &TxnSchedulerBatch{
+	dispatch := &ProposedBatch{
 		IORoot:            body.InputRoot,
 		StorageSignatures: body.InputStorageSigs,
 		Header:            childBlk.Header,
 	}
 	sk, err := memorySigner.NewSigner(rand.Reader)
 	require.NoError(t, err, "NewSigner")
-	signedDispatch, err := signature.SignSigned(sk, TxnSchedulerBatchSigCtx, dispatch)
+	signedDispatch, err := signature.SignSigned(sk, ProposedBatchSignatureContext, dispatch)
```
kostko, line 815: Could use `SignProposedBatch` as well.

go/worker/compute/executor/committee/node.go
```diff
@@ -44,6 +46,12 @@ var (
 	errIncorrectRole      = errors.New("executor: incorrect role")
 	errIncorrectState     = errors.New("executor: incorrect state")
 	errMsgFromNonTxnSched = errors.New("executor: received txn scheduler dispatch msg from non-txn scheduler")
+
+	// Transaction scheduling errors.
+	errNoBlocks            = fmt.Errorf("executor: no blocks")
+	errNotReady            = fmt.Errorf("executor: runtime not ready")
+	errCheckTxFailed       = fmt.Errorf("executor: CheckTx failed")
+	errNotSchedulingLeader = fmt.Errorf("executor: not scheduling leader")
```
kostko, line 54:
```suggestion
	errNotTxnScheduler = fmt.Errorf("executor: not transaction scheduler in this round")
```

go/worker/compute/executor/committee/node.go
```diff
@@ -417,13 +496,244 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
+		spanPublish.Finish()
+		n.logger.Error("failed to publish batch to committee",
+			"err", err,
+		)
+		return err
+	}
+	crash.Here(crashPointLeaderBatchPublishAfter)
+	spanPublish.Finish()
+
+	if epoch.IsExecutorMember() {
```
kostko, line 724: I think this will currently always be true as a transaction scheduler is one of the executors? So it should be safe to remove this check and add a comment saying something like "Also process the batch locally.".

go/worker/compute/executor/init.go
```diff
@@ -1,16 +1,52 @@
...
+	// scheduling it.
+	CfgScheduleCheckTxEnabled = "worker.executor.schedule_check_tx.enabled"
+
+	cfgMaxQueueSize = "worker.executor.schedule_max_queue_size"
+)
+
+// Flags has the configuration flags.
+var Flags = flag.NewFlagSet("", flag.ContinueOnError)
+
+func scheduleCheckTxEnabled() bool {
```
kostko, line 23: Since this is only used below, you could just use viper there directly instead of having these one-line functions?

go/oasis-test-runner/scenario/e2e/runtime/byzantine.go
```diff
@@ -29,6 +30,8 @@ var (
 		oasis.LogAssertExecutionDiscrepancyDetected(),
 	}, oasis.ByzantineSlot3IdentitySeed)
 	// ByzantineExecutorStraggler is the byzantine executor straggler scenario.
+	// XXX: currently ByzantineSlot3IdentitySeed is used, to ensure we are not the tx scheduler
+	// in the round. Once that case is handled we should test both scenarios.
```
kostko, line 34: Agreed that we should test both, is the plan to do it in this PR? If not, open an issue so that we don't forget about it.

go/oasis-test-runner/scenario/e2e/runtime/txsource.go
```diff
@@ -478,6 +478,12 @@ func (sc *txSourceImpl) Run(childEnv *env.Env) error {
 		return fmt.Errorf("WaitNodesRegistered: %w", err)
 	}
 
+	// Wait for second epoch, so that runtimes are up and running.
```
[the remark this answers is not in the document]
ptrus: It was always needed but this is only problematic if the default epoch time is increased so that getting to the 2nd epoch lasts more that the timeouts in the workload-runtime test (120 secs). I noticed this when I was running some tests with increased epoch time.

go/oasis-test-runner/scenario/e2e/runtime/byzantine.go
ptrus: So this PR currently doesn't yet handle the case of scheduler straggling (without waiting for epoch transition).
Once that is handled we can test it here. So the question is if we want to implement that in this or separate PR.
I have not strong feelings (in any case i plan to work on that next), other than it might be more manageable (for reviewing) if it's split in a separate PR. 

If we decide to do it in a following PR, then this PR is ready for final review & merging (I'll squash the commits in 1-2 commits, update the fragments). All tests are passing now, including the 6 hour longtest.

kostko: Yeah let's do it in a separate PR.

ptrus: https://github.com/oasisprotocol/oasis-core/issues/3199

go/consensus/tendermint/apps/roothash/transactions.go
```diff
@@ -90,7 +118,7 @@ func (app *rootHashApplication) getRuntimeState(
 			"round", rtState.CurrentBlock.Header.Round,
 		)
 
-		rtState.Round.Transition(rtState.CurrentBlock)
+		rtState.Round.Transition(rtState.CurrentBlock, ctx.BlockHeight())
```
kostko, line 121: With the `Round` now recording the block height at which the given round was finalized, we should probably not trigger the transition here? It should instead be triggered when the roothash block is finalized (e.g., in `tryFinalizeExecutor` and `emitEmptyBlock`). Otherwise the `CurrentBlockHeight` will be wrong and will not be updated until the first commit is received for the next round.

go/runtime/scheduling/api/api.go
```diff
@@ -6,14 +6,14 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/runtime/transaction"
 )
 
-// Algorithm defines an algorithm for scheduling incoming transaction.
-type Algorithm interface {
-	// Initialize initializes the internal transaction scheduler state.
-	// Algorithm should use the provided transaction dispatcher to dispatch
-	// scheduled transactions.
+// Scheduler defines an algorithm for scheduling incoming transaction.
```
kostko, line 9:
```suggestion
// Scheduler defines an algorithm for scheduling incoming transactions.
```

go/runtime/scheduling/api/api.go
```diff
@@ -32,6 +39,9 @@ type Algorithm interface {
 	// IsQueued returns if a transaction is queued.
 	IsQueued(hash.Hash) bool
 
+	// RemoveTxBatch removes a transaction batch.
+	RemoveTxBatch(tx [][]byte) error
```
kostko, line 43: Maybe name consistently, there is `ScheduleBatch`, `ScheduleTx` and `RemoveTxBatch`. Either add `Tx` everywhere or remove it from everywhere (no strong preference).

go/runtime/scheduling/batching/incoming_queue.go
```diff
@@ -119,24 +149,24 @@ func (q *incomingQueue) AddBatch(batch transaction.RawBatch) error {
...
 
-	// Then add all calls if checks passed.
-	for i, call := range batch {
+		// Check if there is room in the queue.
+		if uint64(len(q.queue)) >= q.maxQueueSize {
+			errs = multierror.Append(errs, fmt.Errorf("failed inserting call: %d, error: %w", i, errQueueFull))
+			return errs
+		}
+
+		// Then add all calls if checks passed.
```
kostko, line 165:
```suggestion
		// Then add a call if checks passed.
```

go/consensus/tendermint/apps/roothash/transactions.go
ptrus: fixed! Note did also have to move the `Round.Reset()` out of `round.Transition()` since in `tryFinalizeExecutor` we also set `.Finalized = true`  and i don't think we want to reset that until we're in `postProcessFinalizedBlock`

go/consensus/tendermint/apps/roothash/roothash.go
```diff
@@ -497,6 +502,7 @@ func (app *rootHashApplication) tryFinalizeExecutor(
 		// Messages omitted on purpose.
 
 		rtState.Round.ExecutorPool.ResetCommitments()
+		rtState.Round.Transition(blk, ctx.BlockHeight())
 		rtState.Round.Finalized = true
```
kostko, line 506: What is the point of this finalized flag now? It is set to true here and then set to false in `postProcessFinalizedBlock` (in `Reset`) which is executed immediately after? If we change the current block immediately, any further commitments will fail anyway as they will be for the wrong block/round? I think we can just get rid of the finalized flag.

go/worker/compute/executor/committee/node.go
```diff
@@ -417,13 +496,239 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
+	}
+
+	incomingQueueSize.With(n.getMetricLabels()).Set(float64(n.scheduler.UnscheduledSize()))
+
+	return nil
+}
+
+// removeTxBatch removes a batch from scheduling queue.
+func (n *Node) removeTxBatch(batch [][]byte) error {
+	// XXX: remove batch can only happen after a batch was already scheduled, meaning
```
kostko, line 593: This is fine as long as there is no way for the scheduler to change (e.g., due to a runtime descriptor update) -- currently there is no way for it to change once set. In that case we would want to hold the mutex here.

go/worker/compute/executor/committee/node.go
```diff
@@ -846,14 +1158,48 @@ func (n *Node) worker() {
...
 
+	// Initialize transaction scheduling algorithm.
+	runtime, err := n.commonNode.Runtime.RegistryDescriptor(n.ctx)
+	if err != nil {
+		n.logger.Error("failed to fetch runtime registry descriptor",
+			"err", err,
+		)
+		return
+	}
+	scheduler, err := scheduling.New(
```
kostko, line 1169: Hm, not related to this PR, but what happens if the runtime descriptor is changed? The scheduler will never update its configuration. Maybe we should open an issue for this.

go/worker/compute/executor/committee/node.go
ptrus: Yeah that's what we also have in the comment in `schedulerMutex` decleration, and since we explicitly state that there i decided to follow that. However i'm also fine with leaving that here for future proofness.

go/worker/compute/executor/committee/node.go
ptrus: oh yeah that's true, currently it wouldn't take affect until the worker is restarted.  i'll open an issue

go/consensus/tendermint/apps/roothash/roothash.go
kostko: Actually why do we even copy the `CurrentBlock` into `Round` since it is already available in runtime state? We would just need to change `Round`'s `AddExecutorCommitment` to take the block as an argument?

ptrus: Didn't even notice that was the only use case for that flag, thanks! fixed

kostko: Cool, thanks! See also my comment about `CurrentBlock` in `Round`, I think we can remove the duplicate block as well?

ptrus: > Actually why do we even copy the CurrentBlock into Round since it is already available in runtime state? 

No idea :-) 

> We would just need to change Round's AddExecutorCommitment to take the block as an argument?

Yeah can do that

go/worker/compute/executor/committee/node.go
ptrus: https://github.com/oasisprotocol/oasis-core/issues/3203

go/consensus/tendermint/apps/roothash/roothash.go
ptrus: fixed as well

go/consensus/tendermint/apps/roothash/state/round.go
```diff
@@ -13,13 +12,11 @@ import (
 type Round struct {
```
kostko, line 12: Sorry to point this out again, but now that current block and finalized have been removed, why do we even have the round type? It looks like it just forwards everything to the `ExecutorPool`? Could we just move `ExecutorPool` and `CurrentBlockHeight` directly to `RuntimeState` and remove `Round`?

go/registry/api/runtime.go
```diff
@@ -226,7 +223,8 @@ type Runtime struct { // nolint: maligned
 	// Executor stores parameters of the executor committee.
 	Executor ExecutorParameters `json:"executor,omitempty"`
 
-	// TxnScheduler stores parameters of the transactions scheduler committee.
+	// TxnScheduler stores transaction scheduling parameters of executor
```
kostko, line 226:
```suggestion
	// TxnScheduler stores transaction scheduling parameters of the executor
```

go/consensus/tendermint/apps/roothash/state/round.go
ptrus: fixed :-)

ptrus:
Did some more (minor) changes see: https://github.com/oasisprotocol/oasis-core/pull/3184/commits/5738d832f12fef9fec1a852316d5c6e806adc7ea (in a separate commit for now, for easier review)

go/runtime/scheduling/api/api.go
kostko: Was this reverted, I remember this was already fixed :-)

kostko (approved):

go/runtime/scheduling/api/api.go
ptrus: Yeah me to :thinking:

ptrus:
Really thanks for all the great reviews @kostko !

PR #3209, opened 2020-08-24
tests/txsource: Run queries workload against all nodes

kostko:
* tests/txsource: Run queries workload against all nodes
* tests/txsource/queries: Verify state integrity
* go/registry: Correctly propagate lookup errors

### DIFF
--- a/go/oasis-test-runner/oasis/oasis.go
+++ b/go/oasis-test-runner/oasis/oasis.go
@@ -200,6 +200,11 @@ func (n *Node) handleExit(cmdErr error) error {
 	}
 }
 
+// Consensus returns the node's consensus configuration.
+func (n *Node) Consensus() ConsensusFixture {
+	return n.consensus
+}
+
 // SetConsensusStateSync configures wheteher a node should perform
 func (n *Node) SetConsensusStateSync(cfg *ConsensusStateSyncCfg) {
 	n.Lock()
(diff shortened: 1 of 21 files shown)


go/oasis-test-runner/scenario/e2e/runtime/txsource.go
```diff
@@ -485,10 +505,35 @@ func (sc *txSourceImpl) Run(childEnv *env.Env) error {
...
+	errCh := make(chan error, len(sc.clientWorkloads)+len(sc.allNodeWorkloads)+2)
+	for _, name := range sc.clientWorkloads {
+		if err := sc.startWorkload(childEnv, errCh, name, &sc.Net.Clients()[0].Node); err != nil {
+			return fmt.Errorf("failed to start client workload %s: %w", name, err)
+		}
+	}
+	for _, name := range sc.allNodeWorkloads {
+		// Determine nodes that we should run the workload on.
+		var nodes []*oasis.Node
+		for _, v := range sc.Net.Validators() {
```
ptrus, line 517: Maybe this can be extracted into a method of the `Network` struct, in case it'll be useful elsewhere.
Also consider moving this out of the above loop.

ptrus (approved):

go/oasis-test-runner/scenario/e2e/runtime/txsource.go
kostko: Good idea, will refactor this a bit.

Yawning (approved):

go/oasis-test-runner/oasis/oasis.go
```diff
@@ -360,6 +365,27 @@ func (net *Network) Byzantine() []*Byzantine {
 	return net.byzantine
 }
 
+// Nodes returns all the nodes associated with the network.
+func (net *Network) Nodes() []*Node {
```
ptrus, line 369: Note that this doesn't return seed and sentry nodes (maybe it's worth mentioning this in comment)

kostko: Updated.

ptrus (approved):

PR #3211, opened 2020-08-24
go/roothash: Change runtime round timeout to be in blocks

### LINKED ISSUE #3210: Runtime round timeout should be in blocks instead of time
In order to simplify the timeout processing logic and avoid spurious timeouts when consensus stalls, runtime round timeouts should be in blocks instead of time.

kostko:
Fixes #3210


go/oasis-node/cmd/registry/runtime/runtime.go
```diff
@@ -537,7 +537,7 @@ func init() {
 	runtimeFlags.Uint64(CfgExecutorGroupSize, 1, "Number of workers in the runtime executor group/committee")
 	runtimeFlags.Uint64(CfgExecutorGroupBackupSize, 0, "Number of backup workers in the runtime executor group/committee")
 	runtimeFlags.Uint64(CfgExecutorAllowedStragglers, 0, "Number of stragglers allowed per round in the runtime executor group")
-	runtimeFlags.Duration(CfgExecutorRoundTimeout, 10*time.Second, "Executor committee round timeout for this runtime")
+	runtimeFlags.Int64(CfgExecutorRoundTimeout, 5, "Executor committee round timeout for this runtime (in consensus blocks)")
```
ptrus, line 540: Maybe add an extra note to the `.changelog/3210.breaking.md`:

> Note: Existing deployments will need to alter the state dump and update the executor round timeout in all registered compute runtimes.

go/roothash/api/commitment/pool.go
```diff
@@ -420,9 +421,9 @@ func (p *Pool) TryFinalize(
 	defer func() {
 		if rearmTimer {
 			// All timeouts are rounded to nearest second to ensure stable serialization.
```
ptrus, line 423: Outdated comment

go/consensus/tendermint/apps/roothash/roothash.go
```diff
@@ -580,13 +526,49 @@ func (app *rootHashApplication) tryFinalizeBlock(
...
+		}
+	}(rtState.ExecutorPool.NextTimeout)
+
 	finalizedBlock := app.tryFinalizeExecutor(ctx, rtState, forced)
 	if finalizedBlock == nil {
 		return nil
 	}
 
-	if err := app.postProcessFinalizedBlock(ctx, rtState, finalizedBlock); err != nil {
+	if err = app.postProcessFinalizedBlock(ctx, rtState, finalizedBlock); err != nil {
```
ptrus, line 571: hm i don't think the `app.postProcessFinalizedBlock` can even ever return an error, maybe update the function definition.

Also the initial error check can then be skipped in the deffered function

go/oasis-node/cmd/registry/runtime/runtime.go
kostko: Good idea, added.

ptrus:
I think it should be also pretty simple adding a "TestRoundTimeout" test to the `roothash/test/tester` for this (we currently only test successful round there):

Something like:
- get latest roothash block
- wait for runtime.Exectuor.RoundTimeout consensus blocks
- ensure that timeout happened and round failed

(just added a similar test in my "proposer timeouts", that's why i know :-) )

go/oasis-node/cmd/registry/runtime/runtime.go
```diff
@@ -372,7 +372,7 @@ func runtimeFromFlags() (*registry.Runtime, signature.Signer, error) { // nolint
 			GroupSize:         viper.GetUint64(CfgExecutorGroupSize),
 			GroupBackupSize:   viper.GetUint64(CfgExecutorGroupBackupSize),
 			AllowedStragglers: viper.GetUint64(CfgExecutorAllowedStragglers),
-			RoundTimeout:      viper.GetDuration(CfgExecutorRoundTimeout),
+			RoundTimeout:      viper.GetInt64(CfgExecutorRoundTimeout),
```
ptrus, line 375: not related to this PR but i also just noticed that we don't verify any (other than `GroupSize`) of the `ExecutorParameters` when registering a runtime.

ptrus: Specifically for `RoundTimeout` should probably allow only `0<=X<=max`, `max` mostly so we don't need to worry about overflowing. Maybe also not allow non sensible small values such us `<5`?

kostko: Validation for RoundTimeout (actually exactly for "<5" :grin:) has already been added in this PR?

ptrus: Oh, sorry missed that :-) maybe also impose a sensible max limit? (and add validation for other executor params that miss it / or we can open an issue for the rest)

kostko: Refactored the runtime descriptor validity checks a bit, please take another look.

ptrus: looks good :+1:

ptrus (approved):
Looks good (there's one more minor pending comment)

go/roothash/api/commitment/pool.go
```diff
@@ -445,6 +445,11 @@ func (p *Pool) TryFinalize(
 			// process the round, assuming that it is possible to do so.
 			if isTimeoutAuthoritative {
 				p.Discrepancy = true
+				// Arm the timer, but increase the roundTimeout as the backup workers should be
+				// given some more time to do the computation.
+				rearmTimer = true
+				// roundTimeout = 1.5*roundTimeout
+				roundTimeout = (15 * roundTimeout) / 10
```
ptrus, line 452: maybe move the `15` into a constant

kostko: Good idea, added numerator/denominator constants.

go/consensus/tendermint/apps/roothash/roothash.go
kostko: Changed.

Issue #3199, opened 2020-08-19
Support roothash timeout triggered by executors


ptrus:
Following on #3184 add support for executors to trigger a roothash timeout in case there are new transactions waiting to be proposed, but the current transaction scheduler does not propose a batch.

The timeout should only be allowed if sufficient consensus blocks (configurable via a runtime parameter) have passed since the round started and there were no proposed batches by the current scheduler.

PR #3212, opened 2020-08-25
go/roothash: Add support for executor triggered timeouts

ptrus:
Fixes: #3199

TODO:
- [x] add byzantine tests for proposers

### DIFF
--- a/go/consensus/tendermint/apps/roothash/transactions.go
+++ b/go/consensus/tendermint/apps/roothash/transactions.go
@@ -6,12 +6,14 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
+	"github.com/oasisprotocol/oasis-core/go/common/logging"
 	abciAPI "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/api"
 	tmapi "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/api"
 	registryState "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/apps/registry/state"
 	roothashState "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/apps/roothash/state"
 	schedulerState "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/apps/scheduler/state"
 	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
+	"github.com/oasisprotocol/oasis-core/go/roothash/api/block"
 	"github.com/oasisprotocol/oasis-core/go/roothash/api/commitment"
 	scheduler "github.com/oasisprotocol/oasis-core/go/scheduler/api"
 )
@@ -68,7 +70,7 @@ func (sv *roothashSignatureVerifier) VerifyTxnSchedulerSignature(sig signature.S
 	if committee == nil {
 		return roothash.ErrInvalidRuntime
 	}
-	scheduler, err := roothash.GetTransactionScheduler(committee, round)
+	scheduler, err := commitment.GetTransactionScheduler(committee, round)
 	if err != nil {
 		return fmt.Errorf("roothash: error getting transaction scheduler: %w", err)
 	}
@@ -109,6 +111,72 @@ func (app *rootHashApplication) getRuntimeState(
 	nl := registryState.NewMutableState(ctx.State())
 
 	return rtState, sv, nl, nil
+}
+
+func (app *rootHashApplication) executorProposerTimeout(
+	ctx *abciAPI.Context,
+	state *roothashState.MutableState,
+	rpt *roothash.ExecutorProposerTimeoutRequest,
+) error {
+	if ctx.IsCheckOnly() {
+		return nil
+	}
+
+	// Charge gas for this transaction.
+	params, err := state.ConsensusParameters(ctx)
+	if err != nil {
+		ctx.Logger().Error("failed to fetch consensus parameters",
+			"err", err,
+		)
+		return err
+	}
+	if err = ctx.Gas().UseGas(1, roothash.GasOpProposerTimeout, params.GasCosts); err != nil {
+		return err
+	}
+
+	rtState, sv, nl, err := app.getRuntimeState(ctx, state, rpt.ID)
+	if err != nil {
+		return err
+	}
+
+	// Ensure enough blocks have passed since round start.
+	proposerTimeout := rtState.Runtime.TxnScheduler.ProposerTimeout
+	currentBlockHeight := rtState.CurrentBlockHeight
+	if height := ctx.BlockHeight(); height < currentBlockHeight+proposerTimeout {
+		ctx.Logger().Error("failed requesting proposer round timeout, timeout not allowed yet",
+			"height", height,
+			"current_block_height", currentBlockHeight,
+			"proposer_timeout", proposerTimeout,
+		)
+		return roothash.ErrProposerTimeoutNotAllowed
+	}
+
+	// Ensure request is valid.
+	if err = rtState.ExecutorPool.CheckProposerTimeout(ctx, rtState.CurrentBlock, sv, nl, ctx.TxSigner(), rpt.Round); err != nil {
+		ctx.Logger().Error("failed requesting proposer round timeout",
+			"err", err,
+			"round", rtState.CurrentBlock.Header.Round,
+			"request", rpt,
+		)
+		return err
+	}
+
+	// Timeout triggered by executor node, emit empty error block.
+	ctx.Logger().Error("proposer round timeout",
+		"round", rpt.Round,
+		"err", err,
+		logging.LogEvent, roothash.LogEventRoundFailed,
+	)
+	if err = app.emitEmptyBlock(ctx, rtState, block.RoundFailed); err != nil {
+		return fmt.Errorf("failed to emit empty block: %w", err)
+	}
+
+	// Update runtime state.
+	if err = state.SetRuntimeState(ctx, rtState); err != nil {
+		return fmt.Errorf("failed to set runtime state: %w", err)
+	}
+
+	return nil
 }
 
 func (app *rootHashApplication) executorCommit(
--- a/go/oasis-node/cmd/debug/byzantine/executor.go
+++ b/go/oasis-node/cmd/debug/byzantine/executor.go
@@ -3,6 +3,7 @@ package byzantine
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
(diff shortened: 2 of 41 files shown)


go/worker/compute/executor/committee/node.go
```diff
@@ -601,6 +611,45 @@ func (n *Node) removeTxBatch(batch [][]byte) error {
...
+		case nil:
+			n.logger.Info("executor timeout request finalized")
+		default:
+			n.logger.Error("failed to submit executor timeout request",
+				"err", err,
+			)
+		}
+	}()
+
+	n.transitionLocked(StateWaitingForTimeout{})
```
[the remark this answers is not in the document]
ptrus: I added it mostly to avoid re-proposing timeout after we do it once for the round, but i think it could be fine as well if we just stay in the WaitingForBatch state.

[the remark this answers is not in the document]
kostko: Oh I see, yeah that makes sense, but maybe do it with a flag (that is cleared on the next block or when the timeout request succeeds/fails) instead of a new state.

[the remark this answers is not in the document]
ptrus: yeah that makes more sense, thanks!

go/worker/compute/executor/committee/node.go
```diff
@@ -508,6 +509,15 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
+	case StateWaitingForTimeout:
+		// A new block means the round has timeouted.
+		n.logger.Info("considering the round timeouted",
+			"round", blk.Header.Round,
+			"header_hash", blk.Header.EncodedHash(),
+			"header_type", blk.Header.HeaderType,
+		)
+
+		n.transitionLocked(StateWaitingForBatch{})
 	}
```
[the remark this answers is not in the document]
ptrus: good idea, added it (but it will only schedule immediately if it already has a full batch ready, otherwise it will still wait for the next "force" flush, which i think makes sense)

go/roothash/api/commitment/pool.go
```diff
@@ -250,6 +285,55 @@ func (p *Pool) addOpenExecutorCommitment(
...
+) (*OpenProposerTimeout, error) {
+	// Check the timeout signature and de-serialize it.
+	openTimeout, err := timeout.Open()
+	if err != nil {
+		return nil, p2pError.Permanent(err)
+	}
+
+	// Ensure timeout is for correct round.
+	if openTimeout.Round < block.Header.Round {
+		return nil, p2pError.Permanent(ErrTimeoutNotCorrectRound)
```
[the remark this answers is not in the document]
ptrus: yeah of course, removed the p2p error wrap

go/roothash/api/commitment/executor.go
```diff
@@ -132,6 +136,47 @@ func (m *ComputeBody) VerifyStorageReceipt(ns common.Namespace, receipt *storage
 	return nil
 }
 
+// ProposerTimeout is a roothash proposer timeout requested by an executor
+// worker.
+type ProposerTimeout struct {
```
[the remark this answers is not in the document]
ptrus: Yeah agreed, simplified all of this to use the transaction signer

go/consensus/tendermint/apps/roothash/transactions.go
```diff
@@ -111,6 +113,73 @@ func (app *rootHashApplication) getRuntimeState(
...
+
+	// Charge gas for this transaction.
+	params, err := state.ConsensusParameters(ctx)
+	if err != nil {
+		ctx.Logger().Error("failed to fetch consensus parameters",
+			"err", err,
+		)
+		return err
+	}
+	if err = ctx.Gas().UseGas(1, roothash.GasOpComputeCommit, params.GasCosts); err != nil {
```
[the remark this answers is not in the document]
ptrus: Done

go/roothash/api/commitment/pool.go
```diff
@@ -250,6 +292,42 @@ func (p *Pool) addOpenExecutorCommitment(
...
+	id signature.PublicKey,
+	round uint64,
+) error {
+	if p.Committee == nil {
+		return ErrNoCommittee
+	}
+	if p.Committee.Kind != scheduler.KindComputeExecutor {
+		return ErrInvalidCommitteeKind
+	}
+
```
kostko, line 310: There is no more round check, it should probably still check if the round matches?

ptrus: Oh yeah of course, i didn't mean to remove that.

go/oasis-test-runner/scenario/e2e/runtime/txsource.go
```diff
@@ -478,12 +476,6 @@ func (sc *txSourceImpl) Run(childEnv *env.Env) error {
 		return fmt.Errorf("WaitNodesRegistered: %w", err)
 	}
 
-	// Wait for second epoch, so that runtimes are up and running.
```
ptrus, line 481: I moved this step directly into the workloads that do runtime queries (`runtime` & `queries` workloads). A benefit for this is that other workloads now start sooner. 

Though, the main reason for moving this is so that `runtimes` workload starts doing queries sooner - what happened before (rarely) was that the runtime workload would submit the first runtime transaction more than `ProposerTimeout` (20) blocks after the 2nd epoch started, leading to nodes triggering proposer timeout on the first transaction received (since triggering a proposer timeout is faster operation than the proposer  proposing a batch and submitting a commitment). With this change hopefully that doesn't happen and we can keep the `AssertNoRoundFailures` check in for the txsource-short workload.

Am also considering adding a short delay before exectuor nodes trigger "proposer timeout", to avoid triggering a timeout always when there are no transactions for `ProposerTimeout` blocks.

go/oasis-node/cmd/debug/byzantine/executor.go
```diff
@@ -40,6 +43,92 @@ func newComputeBatchContext() *computeBatchContext {
 	return &computeBatchContext{}
 }
 
+func (cbc *computeBatchContext) receiveTransactions(ph *p2pHandle, timeout time.Duration) []*api.Tx {
+	var req p2pReqRes
+	txs := []*api.Tx{}
+
+RECEIVE_TRANSACTIONS:
```
kostko, line 50:
```suggestion
ReceiveTransactions:
```

go/roothash/api/api.go
```diff
@@ -52,12 +51,19 @@ var (
...
 	ErrRuntimeSuspended = errors.New(ModuleName, 5, "roothash: runtime is suspended")
 
+	// ErrProposerTimeoutNotAllowed is the error returned when proposer timeout is not allowed.
+	ErrProposerTimeoutNotAllowed = errors.New(ModuleName, 6, "roothash: proposer timeout not allowed")
+
 	// MethodExecutorCommit is the method name for executor commit submission.
 	MethodExecutorCommit = transaction.NewMethodName(ModuleName, "ExecutorCommit", ExecutorCommit{})
 
+	// MethodExecutorProposerTimeout is the method name for executor.
+	MethodExecutorProposerTimeout = transaction.NewMethodName(ModuleName, "MethodExecutorProposerTimeout", ExecutorProposerTimeoutRequest{})
```
kostko, line 61:
```suggestion
	MethodExecutorProposerTimeout = transaction.NewMethodName(ModuleName, "ExecutorProposerTimeout", ExecutorProposerTimeoutRequest{})
```

go/roothash/api/commitment/executor.go
```diff
@@ -18,6 +18,10 @@ var (
 	// worker commitments.
 	ExecutorSignatureContext = signature.NewContext("oasis-core/roothash: executor commitment", signature.WithChainSeparation())
 
+	// RequestProposerTimeoutSignatureContext is the signature context used to
+	// sign executor proposer timeout requests.
+	RequestProposerTimeoutSignatureContext = signature.NewContext("oasis-core/roothash: executor proposer timeout request", signature.WithChainSeparation())
```
kostko, line 23: This is probably not needed anymore?

go/worker/compute/executor/committee/node.go
```diff
@@ -508,6 +532,14 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
 
 		// Record time taken for successfully processing a batch.
 		batchProcessingTime.With(n.getMetricLabels()).Observe(time.Since(state.batchStartTime).Seconds())
+
+		// Check if we are a proposer and if so try to immediately schedule a new batch.
```
kostko, line 536: This should probably be moved outside the switch, in case the current state after the switch is `StateWaitingForBatch`?

ptrus: hm, yeah i think that makes more sense indeed

go/runtime/scheduling/simple/simple.go
```diff
@@ -95,9 +95,9 @@ func (s *scheduler) RemoveTxBatch(tx [][]byte) error {
 	return s.incomingQueue.RemoveBatch(tx)
 }
 
-func (s *scheduler) Flush() error {
-	// Force schedule a batch.
-	if err := s.scheduleBatch(true); err != nil {
+func (s *scheduler) Flush(force bool) error {
+	// Schedule a batch.
+	if err := s.scheduleBatch(force); err != nil {
 		// XXX: Log a warning here as the exected common failures are
```
kostko, line 101:
```suggestion
		// XXX: Log a warning here as the expected common failures are
```

go/worker/common/p2p/dispatch.go
```diff
@@ -178,6 +178,7 @@ func (h *topicHandler) dispatchMessage(peerID core.PeerID, m *queuedMsg, isIniti
 		// Dispatch the message to the handler.
 		// XXX: Could also dispatch the message to all handlers, and only fail after.
 		if err = handler.HandlePeerMessage(m.from, m.msg, m.peerID == h.p2p.host.ID()); err != nil {
+			h.logger.Error("handling message error", "message", m.msg, "from", m.from, "err", err)
```
kostko, line 181:
```suggestion
			h.logger.Error("failed to handle message", "message", m.msg, "from", m.from, "err", err)
```

go/worker/compute/executor/committee/node.go
```diff
@@ -140,6 +143,9 @@ type Node struct { // nolint: maligned
 	schedulerMutex sync.RWMutex
 	scheduler      schedulingAPI.Scheduler
 
+	proposingTimeout     bool
+	proposingTimeoutLock sync.RWMutex
```
kostko, line 147: Is a separate lock required or do all callers (except line 699) take the CrossNode lock anyway? In this case, just mention that this is guarded by the CrossNode lock (as below) and remove the methods?

go/worker/compute/executor/committee/node.go
```diff
@@ -505,9 +532,16 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
-		// Record time taken for successfully processing a batch.
-		batchProcessingTime.With(n.getMetricLabels()).Observe(time.Since(state.batchStartTime).Seconds())
+	// Clear the potentially set "is proposing timeout" flag from the previous round.
+	n.clearIsProposingTimeout()
+
+	// Check if we are a proposer and if so try to immediately schedule a new batch.
+	if n.commonNode.Group.GetEpochSnapshot().IsTransactionScheduler(blk.Header.Round) {
+		go func() {
+			n.scheduler.Flush(false)
+		}()
```
kostko, line 544:
```suggestion
		go n.scheduler.Flush(false)
```

kostko (approved):

PR #3225, opened 2020-09-01
go/common/crypto/signature: Use descriptive names for Signer roles 

tjanez:
The `--signer.composite.backends` CLI flag previously accepted integer-indexed Signer roles, e.g:

```
--signer.composite.backends 1:file,2:file,3:file,4:plugin
```

Now, it only accepts descriptive string names for Signer roles, e.g.:

```
--signer.composite.backends entity:file,node:file,p2p:file,consensus:plugin
```

### DIFF
--- a/go/common/crypto/signature/signer.go
+++ b/go/common/crypto/signature/signer.go
@@ -2,11 +2,11 @@ package signature
 
 import (
 	"crypto/sha512"
+	"encoding"
 	"errors"
 	"fmt"
 	"io"
 	"os"
-	"strconv"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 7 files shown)


go/common/crypto/signature/signer.go
```diff
@@ -156,27 +162,53 @@ func SetChainContext(rawContext string) {
...
+		return SignerP2PName
+	case SignerConsensus:
+		return SignerConsensusName
+	default:
+		return "[unknown signer role]"
 	}
+}
+
+// MarshalText encodes a SignerRole into text form.
+func (role SignerRole) MarshalText() ([]byte, error) {
```
kostko, line 194: Is this used anywhere? Consider adding a simple round-trip test where you marshal/unmarshal all the roles.

tjanez: > Is this used anywhere?

Currently, it's not used but I added it for completeness.

> Consider adding a simple round-trip test where you marshal/unmarshal all the roles.

Done.

kostko (approved):

PR #3238, opened 2020-09-04
go: Update Go dependencies

ptrus:


kostko (approved):

kostko:
Should we enable [dependabot](https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates) so we get these bumps automatically?

ptrus:
> Should we enable dependabot so we get these bumps automatically?

Good idea, i'll open an issue and look into adding that

PR #3246, opened 2020-09-04
ci/ci-reproducibility: fix output path

ptrus:
This matches the behavior from actions/download-artifact@v1 (ref: https://github.com/actions/download-artifact#compatibility-between-v1-and-v2)


kostko (approved):
If the changelog fragment is empty, it should be a trivial one.

PR #3250, opened 2020-09-07
go/control: Add per-runtime status report

### LINKED ISSUE #3234: Add runtime status to node status response
The node status returned by the `GetStatus` control endpoint should be updated to also include a map of runtimes supported by the running node and then for each runtime at least the following information:
* Genesis round, timestamp and state root.
* Latest round, timestamp and state root.
* Flag whether the node is currently in the executor committee.
* Flag whether the node is currently in the storage committee.
* If the node is a storage node, it should also include the information about the latest synced round.

kostko:
Fixes #3234

### DIFF
--- a/.changelog/3234.feature.md
+++ b/.changelog/3234.feature.md
@@ -1,0 +1,6 @@
+go/control: Add per-runtime status report
+
+An additional `runtimes` field has been added to the output of the
+`control.GetStatus` method (e.g., as reported by `control status`
+subcommand). The field contains a map of supported runtime IDs to their
+runtime-specific status reports.
--- a/docs/oasis-node/cli.md
+++ b/docs/oasis-node/cli.md
@@ -10,38 +10,150 @@ Run
 oasis-node control status
 ```
 
-to get information like the following:
+to get information like the following (example taken from a runtime compute
+node):
 
 ```json
 {
-  "software_version": "20.8",
+  "software_version": "20.10",
   "identity": {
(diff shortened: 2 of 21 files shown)


.changelog/3234.feature.md
```diff
@@ -0,0 +1,6 @@
+go/control: Add per-runtime status report
+
+An additional `runtimes` field has been added to the output of the
```
ptrus, line 3: Might be nice to also update the example command docs, with a runtime node example https://github.com/oasisprotocol/oasis-core/blob/master/docs/oasis-node/cli.md

ptrus (approved):

abukosek (approved):

ptrus:
Should we also add a CLI E2E test for this command (similarly as we have stake/registry/identity cli e2e tests). (could be that we already do this, but i didn't see it in a quick search, only some indirect testing of it in `e2e/node_shutdown`)

.changelog/3234.feature.md
kostko: Done.

kostko:
So we do test the `GetStatus` API call (in the [queries txsource workload](https://github.com/oasisprotocol/oasis-core/blob/99f3bcede29339f2a0cda9acd94dbddd71e3e590/go/oasis-node/cmd/debug/txsource/workload/queries.go#L658)), but not the CLI subcommand itself.

Yawning (approved):

PR #3252, opened 2020-09-08
go/worker/storage: Always use fresh nodes for policy updates

kostko:


ptrus (approved):

ptrus:
Btw is there a way to disable specifically the "Unchanged files with check annotations" from codecov. I think the annotations for files that weren't changed in a PR don't make that much sense.

kostko:
> Btw is there a way to disable specifically the "Unchanged files with check annotations" from codecov. I think the annotations for files that weren't changed in a PR don't make that much sense.

Agreed, but I don't see that option, it seems you can only [disable all annotations](https://docs.codecov.io/docs/github-checks-beta#disabling-github-checks-patch-annotations).

PR #3262, opened 2020-09-10
ADR 0003: Consensus/Runtime Token Transfer

kostko:


docs/adr/0003-consensus-runtime-token-transfer.md
```diff
@@ -0,0 +1,379 @@
...
+- `amount` specifies the amount of base units to transfer.
+
+If neither `from` nor `to` are specified the message is invalid.
+
+If there is not enough balance in the given source account (either the runtime
+sub account or the runtime account), the message is invalid.
+
+#### Withdraw
+
+The withdraw message enables transfers from a runtime sub account to a general
```
bennetyee, line 220: could the name be "refund", "release", or something like that instead of "withdraw".  the actor for an withdrawal action is usually the depositor, and i think it might be confusing since (i think of) the depositor is the owner of the general account, and this message is only sent by the runtime.

kostko: Yeah good point, "refund" sounds good.

kostko: This has been refactored to the allowance model (where an account holder can approve beneficiaries so they can withdraw) where I think withdraw makes more sense. Please take another look.

docs/adr/0003-consensus-runtime-token-transfer.md
```diff
@@ -0,0 +1,460 @@
...
+- `amount` specifies the amount of base units to withdraw.
+
+The transaction signer implicitly specifies the destination general account.
+Upon executing the withdrawal the following actions are performed:
+
+- If either the `disable_transfers` or `disable_allowances` staking consensus
+  parameters are set to `true`, the method fails with `ErrForbidden`.
+
+- Address specified by `from` is compared with the transaction signer address.
+  If the addresses are the same, the method fails with `ErrInvalidArgument`.
```
pro-wh, line 173: can we have a similar check in `Approve`?

docs/adr/0003-consensus-runtime-token-transfer.md
```diff
@@ -0,0 +1,460 @@
...
+
+**Fields:**
+
+- `from` specifies the account address to withdraw from.
+- `amount` specifies the amount of base units to withdraw.
+
+The transaction signer implicitly specifies the destination general account.
+Upon executing the withdrawal the following actions are performed:
+
+- If either the `disable_transfers` or `disable_allowances` staking consensus
```
pro-wh, line 169: this, whereas Approve doesn't check `disable_transfers`. but what would be the purpose of having approves without withdraws? maybe we should use the same criteria for both.

docs/adr/0003-consensus-runtime-token-transfer.md
```diff
@@ -0,0 +1,460 @@
...
+
+- The account indicated by the signer is loaded.
+
+- If the approval would create a new allowance and the maximum number of
+  allowances for an account has been reached, the method fails with
+  `ErrInvalidArgument`.
+
+- The set of approvals is updated so that the allowance is updated as specified
+  by `amount_change`/`negative`. In case the change would cause the allowance to
+  be equal to zero or negative, the allowance is removed.
```
pro-wh, line 125: this will be the only place where we succeed if an amount asked for was too much. are there any specific use cases this leniency is meant to support?

docs/adr/0003-consensus-runtime-token-transfer.md
```diff
@@ -0,0 +1,460 @@
...
+
+- `amount` is deducted from the source general account balance. If this would
+  cause the balance to go negative, the method fails with
+  `ErrInsufficientBalance`.
+
+- `amount` is added to the destination general account balance.
+
+- Both source and destination accounts are saved.
+
+- The corresponding `TransferEvent` is emitted.
```
pro-wh, line 191: the allowance would change, without any event about it

docs/adr/0003-consensus-runtime-token-transfer.md
kostko: Good idea!

docs/adr/0003-consensus-runtime-token-transfer.md
kostko: Agreed.

docs/adr/0003-consensus-runtime-token-transfer.md
kostko: So the thing is that you specify how much the allowance changes. So it could happen that you look at the allowance and see that it is X and you want to set it to zero, so you submit a tx T1 `{amount_change: X, negative: true}`. But before that tx is executed, tx T2 is executed which withdraws Y < X. In this case T1 would fail instead of setting the allowance to zero which may not be desired.

docs/adr/0003-consensus-runtime-token-transfer.md
kostko: Good point, updated `ApprovalEvent` to `AllowanceChangeEvent` and made sure it is also emitted here.

docs/adr/0003-consensus-runtime-token-transfer.md
kostko: Updated.

docs/adr/0003-consensus-runtime-token-transfer.md
kostko: Updated.

ptrus (approved):

tjanez (approved):

bennetyee (approved):

abukosek (approved):

PR #3273, opened 2020-09-14
[BACKPORT/20.10.x] Pretty print commission schedule rates and rate bounds as enumerated lists

tjanez:
Backport of #3265.

### DIFF
--- a/.changelog/3265.feature.1.md
+++ b/.changelog/3265.feature.1.md
@@ -1,0 +1,1 @@
+go/staking/api: Add `PrettyPrintCommissionScheduleIndexInfixes()` helper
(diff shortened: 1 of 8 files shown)


ptrus:
Can you also include https://github.com/oasisprotocol/oasis-core/pull/3264 please? (probably can even backport everything since last release, since i think there's no breaking changes).

tjanez:
> Can you also include #3264 please? (since you're already backporting things). Probably could even backport everything since last release, since i think there's no breaking changes.

Yea, sure, I could.

> Ah i see you are backporting only commits from a single PR (at first look i thought you're backporting multiple things), i can also backport that PR myself.

Yea, only from a single PR. Probably will make things more transparent if we backport each PR separately.

If we'll need this a lot, it might be handy to prepare a bot / GitHub action / Make target or something :slightly_smiling_face:.

kostko (approved):

Issue #3274, opened 2020-09-14
Add a LRU cache of last seen runtime transactions


kostko:
To enable a basic form of transaction replay prevention, the basic transaction scheduler should add a LRU cache (with the cache size being configurable via a CLI argument) of last seen runtime transactions keyed by transaction hash.

PR #3279, opened 2020-09-15
go/worker/executor: cache of last seen runtime transactions

ptrus:
Closes: #3274

### DIFF
--- a/go/worker/compute/executor/committee/node.go
+++ b/go/worker/compute/executor/committee/node.go
@@ -11,6 +11,7 @@ import (
 	"github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
 
+	"github.com/oasisprotocol/oasis-core/go/common/cache/lru"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/crash"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
@@ -52,6 +53,7 @@ var (
 	errNotReady        = fmt.Errorf("executor: runtime not ready")
 	errCheckTxFailed   = fmt.Errorf("executor: CheckTx failed")
 	errNotTxnScheduler = fmt.Errorf("executor: not transaction scheduler in this round")
+	errDuplicateTx     = fmt.Errorf("executor: duplicate transaction")
 
 	// Duration to wait before submitting the propose timeout request.
 	proposeTimeoutDelay = 2 * time.Second
@@ -131,6 +133,8 @@ var (
 // Node is a committee node.
 type Node struct { // nolint: maligned
 	*commonWorker.RuntimeHostNode
+
+	lastScheduledCache *lru.Cache
 
 	scheduleCheckTxEnabled bool
 	scheduleMaxQueueSize   uint64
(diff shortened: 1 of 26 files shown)


go/worker/compute/executor/committee/node.go
```diff
@@ -593,10 +607,28 @@ func (n *Node) QueueTx(call []byte) error {
 	if n.scheduler == nil || !n.scheduler.IsInitialized() {
 		return errNotReady
 	}
+
+	callHash := hash.NewFromBytes(call)
+	if _, b := n.lastScheduledCache.Peek(callHash); b {
```
kostko, line 612: Why `Peek` instead of `Get`?

go/worker/compute/executor/committee/node.go
```diff
@@ -432,11 +436,21 @@ func (n *Node) HandleEpochTransitionLocked(epoch *committee.EpochSnapshot) {
 	if !epoch.IsExecutorWorker() {
 		// Clear incoming queue if we are not a worker anymore.
 		n.scheduler.Clear()
+		n.lastScheduledCache.Clear()
 		incomingQueueSize.With(n.getMetricLabels()).Set(0)
 	}
 
 	switch {
 	case epoch.IsExecutorMember():
+		if state, ok := n.state.(StateWaitingForFinalize); ok {
```
kostko, line 445: Hm on a related note, in `HandleNewBlockLocked` when in `StateWaitingForFinalize` state we currently always call `removeTxBatch` but we likely need to check that the round succeeded (header type is "normal") and that this was actually the round that we processed (e.g. by comparing the IO roots) as otherwise we can incorrectly remove a batch that was not actually finalized?

go/worker/compute/executor/committee/node.go
```diff
@@ -1423,6 +1456,11 @@ func NewNode(
 		return nil, err
 	}
 
+	cache, err := lru.New(lru.Capacity(lastScheduledCacheSize, false))
```
kostko, line 1459: If cache size is specified as zero this should probably disable the cache instead of making it unlimited (which sounds dangerous) which is what happens currently?

go/worker/compute/executor/committee/node.go
ptrus: Yeah good point, i did notice that, but then only checked what happens on epoch transition, which is handled fine (we don't delete the batch since we transition the state before that already), didnt thought about other failures though. Will fix it

ptrus: done

go/worker/compute/executor/committee/node.go
```diff
@@ -492,28 +510,46 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
-		// A new block means the round has been finalized.
-		n.logger.Info("considering the round finalized",
-			"round", blk.Header.Round,
-			"header_hash", blk.Header.EncodedHash(),
-		)
-		// Record time taken for successfully processing a batch.
-		batchProcessingTime.With(n.getMetricLabels()).Observe(time.Since(state.batchStartTime).Seconds())
+		func() {
+			defer n.transitionLocked(StateWaitingForBatch{})
+			// A new block means the round has been finalized.
```
kostko, line 520: Empty line above.

go/worker/compute/executor/committee/node.go
```diff
@@ -1412,7 +1467,10 @@ func NewNode(
 	roleProvider registration.RoleProvider,
 	scheduleCheckTxEnabled bool,
 	scheduleMaxQueueSize uint64,
+	lastScheduledCacheSize uint64,
 ) (*Node, error) {
+	var err error
```
kostko, line 1472: Why is this needed, isn't `err` defined when `NewRuntimeHostNode` is called below?

go/worker/compute/executor/committee/node.go
kostko: Looks good, thanks!

kostko (approved):

go/worker/compute/executor/committee/node.go
ptrus: You are right, thx :+1:

Issue #3275, opened 2020-09-14
Use hash-based message ids for P2P messages


kostko:
Currently we use the default message ids in libp2p pubsub, which defaults to source + sequence number. We should instead change it to be the hash of the payload. This can be configured by using `pubsub.WithMessageIdFn`.

PR #3280, opened 2020-09-15
go/worker/p2p: Use hash of the message payload as message ID

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3275

### DIFF
--- a/go/worker/common/p2p/p2p.go
+++ b/go/worker/common/p2p/p2p.go
@@ -13,12 +13,14 @@ import (
 	"github.com/libp2p/go-libp2p-core/network"
 	"github.com/libp2p/go-libp2p-core/transport"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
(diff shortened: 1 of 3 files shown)


go/worker/common/p2p/p2p.go
```diff
@@ -201,6 +203,9 @@ func New(ctx context.Context, identity *identity.Identity, consensus consensus.B
 		pubsub.WithFloodPublish(true),
 		pubsub.WithPeerOutboundQueueSize(viper.GetInt(CfgP2PPeerOutboundQueueSize)),
 		pubsub.WithValidateQueueSize(viper.GetInt(CfgP2PValidateQueueSize)),
+		pubsub.WithMessageIdFn(func(pmsg *pb.Message) string {
+			return hash.NewFromBytes(pmsg.Data).String()
```
kostko, line 207: No need to encode in hex via `String()`, just convert the hash bytes slice to a (binary) string. It should be fine here.

kostko (approved):

PR #3282, opened 2020-09-16
Ignore some nancy false positives from 2018

kostko:


abukosek (approved):

.nancy-ignore
```diff
@@ -3,3 +3,12 @@
 # the websocket in go.mod: https://github.com/oasisprotocol/oasis-core/issues/2931
 # https://ossindex.sonatype.org/vuln/5f259e63-3efb-4c47-b593-d175dca716b0
 5f259e63-3efb-4c47-b593-d175dca716b0
+
+# For some reason these are false positives in the golang/golang.org/x/net, we
+# use the latest version of this package which is not vulnerable.
```
ptrus, line 8: Maybe it would be also useful to link to the relevant issue here:  https://github.com/sonatype-nexus-community/nancy/issues/189

ptrus (approved):

.nancy-ignore
kostko: Indeed!

Issue #2515, opened 2020-01-06
Runtimes updating their own runtime descriptor


kostko:
To support updating runtime descriptors via a more decentralized runtime governance, we should consider allowing runtimes (e.g., any current committee) to update parts of their own runtime descriptor. This should be configurable similar to the addmission policy to have both entity-governed (where an entity private key can perform updates) and self-governed (where only the runtime can perform updates) runtimes.

Runtimes would probably signal updates via a roothash message.

Note that currently the runtime descriptors are signed by the entity which prevents any modifications.

kostko:
This was done.

PR #3284, opened 2020-09-17
ADR 0004: Runtime Governance

kostko:
See #2515

### DIFF
--- a/docs/adr/0004-runtime-governance.md
+++ b/docs/adr/0004-runtime-governance.md
@@ -1,0 +1,298 @@
+# ADR 0004: Runtime Governance
+
+## Changelog
+
+- 2020-10-07: Add per-role max node limits, minimum required election pool size
+- 2020-09-30: Add entity whitelist admission policy max nodes limit
+- 2020-09-17: Initial draft
+
+## Status
+
+Accepted
+
+## Context
+
+Currently all runtimes can only be governed by a single entity -- the runtime
+owner. In this regard governance means being able to update certain fields in
+the runtime descriptor stored by the consensus layer registry service. On one
+hand the runtime descriptor contains security-critical parameters and on the
+other there needs to be a mechanism through which the runtimes can be upgraded
+(especially so for TEE-based runtimes where a specific runtime binary is
+enforced via remote attestation mechanisms).
+
+This proposal extends runtime governance options and enables a path towards
+runtimes that can define their own governance mechanisms. This proposal assumes
+that [ADR 0003] has been adopted and runtimes can have their own accounts in the
+staking module.
+
+## Decision
+
+This proposal takes a simplistic but powerful approach which allows each runtime
+to choose its governance model upon its first registration. It does so through
+a newly introduced field in the runtime descriptor which indicates how the
+runtime descriptor can be updated in the future.
+
+### Runtime Descriptor
+
+The runtime descriptor version is bumped to `2`. Version `1` descriptors are
+accepted at genesis and are converted to the new format by assuming the entity
+governance model as that is the only option in v1. All new runtime registrations
+must use the v2 descriptor.
+
+#### Governance Model
+
+This proposal updates the runtime descriptor by adding fields as follows:
+
+```golang
+type Runtime struct {
+    // GovernanceModel specifies the runtime governance model.
+    GovernanceModel RuntimeGovernanceModel `json:"governance_model"`
+
+    // ... existing fields omitted ...
+}
+
+// RuntimeGovernanceModel specifies the runtime governance model.
+type RuntimeGovernanceModel uint8
+
+const (
+    GovernanceEntity    RuntimeGovernanceModel = 1
+    GovernanceRuntime   RuntimeGovernanceModel = 2
+    GovernanceConsensus RuntimeGovernanceModel = 3
+)
+
+// ... some text serialization methods omitted ...
+```
+
+The `governance_model` field can specifiy one of the following governance
+models:
+
+- **Entity governance (`GovernanceEntity`).** This causes the runtime to behave
+  exactly as before, the runtime owner (indicated by `entity_id` in the runtime
+  descriptor) is the only one who can update the runtime descriptor via
+  `registry.RegisterRuntime` method calls.
+
+  The runtime owner is also the one that needs to provide the required stake
+  in escrow in order to avoid the runtime from being suspended. As before note
+  that anyone can delegate the required stake to the runtime owner in order to
+  enable runtime operation (but the owner can always prevent the runtime from
+  operating by performing actions which would cause the stake claims to no
+  longer be satisfied).
+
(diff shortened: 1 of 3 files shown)


docs/adr/0004-runtime-governance.md
```diff
@@ -0,0 +1,225 @@
...
+### Consensus Parameters
+
+#### Registry
+
+This proposal introduces the following new consensus parameters in the registry
+module:
+
+- `disable_runtime_governance_models` (set of `RuntimeGovernanceModel`)
+  specifies the set of runtime governance models that are not allowed to be used
+  when updating registrations (either via method calls or via runtime messages).
```
pro-wh, line 196: or when creating registrations?

docs/adr/0004-runtime-governance.md
```diff
@@ -0,0 +1,225 @@
...
+- **Consensus layer governance (`GovernanceConsensus`).** In this case only the
+  consensus layer itself can update the runtime descriptor either through a
+  network upgrade or via a consensus layer governance mechanism not specified by
+  this proposal.
+
+  Runtimes using this governance model are never suspended and do not need to
+  provide stake in escrow.
+
+  Runtimes using this governance model cannot be registered via regular registry
+  method calls (doing so must return `ErrForbidden`). Instead such a runtime can
```
pro-wh, line 103: I think this includes when an owning entity or runtime tries to update its own descriptor, is that right?

docs/adr/0004-runtime-governance.md
```diff
@@ -0,0 +1,225 @@
...
+```
+registry.RegisterRuntime
+```
+
+The body of a register runtime transaction must be a `Runtime` descriptor.
+The signer of the transaction must be the owning entity key.
+
+Registering a runtime may require sufficient stake in either the owning entity's
+(when entity governance is used) or the runtime's (when runtime governance is
+used) escrow account.
```
pro-wh, line 158: what is the workflow for creating a runtime with runtime governance? do you create it that way, and it's immediately suspended and then you do something? or maybe you create it under entity governance, and you transfer some tokens and then you update it to runtime governance?

docs/adr/0004-runtime-governance.md
```diff
@@ -0,0 +1,225 @@
...
+### State
+
+This proposal introduces/updates the following consensus state in the registry
+module:
+
+#### Stored Runtime Descriptors
+
+Since the runtime descriptors can now be updated by actors other than the
+initial registering entity, it does not make sense to store signed runtime
+descriptors. The value of storage key prefixed with `0x13` which previously
```
pro-wh, line 116: is this similarly true of the genesis document format?

docs/adr/0004-runtime-governance.md
```diff
@@ -0,0 +1,225 @@
...
+## Decision
+
+This proposal takes a simplistic but powerful approach which allows each runtime
+to choose its governance model upon its first registration. It does so through
+a newly introduced field in the runtime descriptor which indicates how the
+runtime descriptor can be updated in the future.
+
+Additionally it also introduces an additional safeguard for (all) runtime
+descriptor updates -- a configurable delay period after which any modifications
+to a runtime descriptor take effect instead of them taking effect immediately.
```
pro-wh, line 35: what does this safeguard against?

docs/adr/0004-runtime-governance.md
kostko: The create/update is the same method call, but yeah we can say "when calling `registry.RegisterRuntime`" to be more clear instead.

docs/adr/0004-runtime-governance.md
kostko: Correct.

docs/adr/0004-runtime-governance.md
kostko: You can escrow tokens to the runtime account before registering the runtime itself. And then you can register it with runtime governance directly.

docs/adr/0004-runtime-governance.md
kostko: Yes, everywhere where there were signed runtime descriptors they now need to be plain descriptors.

docs/adr/0004-runtime-governance.md
kostko: In theory it would give users time to e.g., move tokens out of the runtime in case they don't agree with the upgrade (security parameter change/MRENCLAVE change/etc.). Of course there are ways to prevent this, so it may not be worth adding the complexity.

docs/adr/0004-runtime-governance.md
kostko: Added a section on genesis document changes.

docs/adr/0004-runtime-governance.md
```diff
@@ -0,0 +1,246 @@
...
+- 2020-09-17: Initial draft
+
+## Status
+
+Proposed
+
+## Context
+
+Currently all runtimes can only be governed by a single entity -- the runtime
+owner. In this regard governance means being able to update certain fields in
```
bennetyee, line 14: We probably need to enumerate the fields that can be updated and if update follows the same rules (range of allowed values, interaction with other values, etc) as creation.  E.g., could `GovernanceModel` be updated?

hydai:
Hi, I would like to request a feature of the runtime governance.

Is it possible to limit the number of compute nodes per entity?

As a ParaTime developer, we will encounter the following issues if the number of compute nodes per entity is not limited:

1. Someone can create as many as possible compute nodes with malicious modification. When the number of compute nodes per entity is not limited, all of the active compute nodes can be taken by this maliciously modified version in one epoch time.
2. If we want to provide rewards for the active compute nodes, an entity which creates more compute nodes may increase its possibility to join the active group.

docs/adr/0004-runtime-governance.md
```diff
@@ -0,0 +1,292 @@
...
+the given runtime for each role.
+
+```golang
+type EntityWhitelistConfig struct {
+    // MaxNodes is the maximum number of nodes that an entity can register under
+    // the given runtime for a specific role. If the map is empty or absent, the
+    // number of nodes is unlimited. If the map is present and non-empty, the
+    // the number of nodes is restricted to the specified maximum (where zero
+    // means no nodes allowed), any missing roles imply zero nodes.
+    MaxNodes map[node.RolesMask]uint16 `json:"max_nodes,omitempty"`
```
ptrus, line 120: This should probably only allow `RolesMask` values that present a single role (e.g. where single bit is set)? Otherwise there could be conflicting entries.

ptrus: Ah addressed bellow :grin:

kostko: Yeah this is already specified below:
> Each key (roles mask) in the `max_nodes` map must specify a single role, otherwise the runtime descriptor is rejected with `ErrInvalidArgument`.

peterjgilbert (approved):

PR #3285, opened 2020-09-17
go/storage/mkvs: Add node tracking to chunk restore

### LINKED ISSUE #3241: Clean stale nodes in checkpointer after aborting a restore
At the moment, `AbortRestore` just clears internal bureaucracy and leaves data from successfully restored chunks in the database. Those should all be cleared, probably by keeping a nearby list of restored hashes that were actually inserted in the db.

jberci:
Closes #3241

### DIFF
--- a/go/storage/mkvs/checkpoint/restorer.go
+++ b/go/storage/mkvs/checkpoint/restorer.go
@@ -31,6 +31,10 @@ func (rs *restorer) StartRestore(ctx context.Context, checkpoint *Metadata) erro
 		return ErrRestoreAlreadyInProgress
 	}
 
+	if err := rs.ndb.StartMultipartInsert(checkpoint.Root.Version); err != nil {
+		return err
+	}
+
 	rs.currentCheckpoint = checkpoint
 	rs.pendingChunks = make(map[uint64]bool)
 	for idx := range checkpoint.Chunks {
@@ -44,16 +48,10 @@ func (rs *restorer) AbortRestore(ctx context.Context) error {
 	rs.Lock()
 	defer rs.Unlock()
 
-	if rs.currentCheckpoint == nil {
-		return ErrNoRestoreInProgress
-	}
-
 	rs.pendingChunks = nil
 	rs.currentCheckpoint = nil
 
-	// TODO: also properly remove leftover nodes from the database (#3241).
-
-	return nil
+	return rs.ndb.AbortMultipartInsert()
 }
 
 func (rs *restorer) GetCurrentCheckpoint() *Metadata {
@@ -95,12 +93,7 @@ func (rs *restorer) RestoreChunk(ctx context.Context, idx uint64, r io.Reader) (
 	case errors.Is(err, ErrChunkProofVerificationFailed):
 		// Chunk was as specified in the manifest but did not match the reported root. In this case
 		// we need to abort processing the given checkpoint.
-		rs.Lock()
-		defer rs.Unlock()
-
-		rs.pendingChunks = nil
-		rs.currentCheckpoint = nil
-
+		_ = rs.AbortRestore(ctx)
 		return false, err
 	default:
 		return false, err
@@ -116,7 +109,6 @@ func (rs *restorer) RestoreChunk(ctx context.Context, idx uint64, r io.Reader) (
 	if len(rs.pendingChunks) == 0 {
 		rs.pendingChunks = nil
 		rs.currentCheckpoint = nil
-
 		return true, nil
 	}
 
--- a/go/storage/mkvs/db/badger/badger.go
+++ b/go/storage/mkvs/db/badger/badger.go
@@ -3,6 +3,7 @@ package badger
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 
@@ -20,7 +21,13 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/writelog"
 )
 
-const dbVersion = 3
+const (
+	dbVersion = 3
+
+	// multipartVersionNone is the value used for the multipart version in metadata
+	// when no multipart restore is in progress.
+	multipartVersionNone uint64 = 0
+)
 
 var (
 	// nodeKeyFmt is the key format for nodes (node hash).
@@ -46,6 +53,13 @@ var (
 	//
 	// Value is CBOR-serialized metadata.
 	metadataKeyFmt = keyformat.New(0x04)
+	// multipartRestoreNodeLogKeyFmt is the key format for the nodes inserted during a chunk restore.
+	// Once a set of chunks is fully restored, these entries should be removed. If chunk restoration
+	// is interrupted for any reason, the nodes associated with these keys should be removed, along
+	// with these entries.
+	//
+	// Value is empty.
+	multipartRestoreNodeLogKeyFmt = keyformat.New(0x05, &hash.Hash{})
 )
 
 // New creates a new BadgerDB-backed node database.
@@ -87,6 +101,12 @@ func New(cfg *api.Config) (api.NodeDB, error) {
 		return nil, fmt.Errorf("mkvs/badger: failed to load metadata: %w", err)
 	}
 
+	// Cleanup any multipart restore remnants, since they can't be used anymore.
+	if err = db.cleanMultipartLocked(true); err != nil {
+		_ = db.db.Close()
+		return nil, fmt.Errorf("mkvs/badger: failed to clean leftovers from multipart restore: %w", err)
+	}
+
 	db.gc = cmnBadger.NewGCWorker(db.logger, db.db)
 
 	return db, nil
@@ -99,6 +119,8 @@ type badgerNodeDB struct { // nolint: maligned
 
 	readOnly         bool
 	discardWriteLogs bool
+
+	multipartVersion uint64
 
 	db *badger.DB
 	gc *cmnBadger.GCWorker
@@ -141,6 +163,7 @@ func (d *badgerNodeDB) load() error {
 				d.meta.value.Namespace,
 			)
 		}
+		return nil
 	case badger.ErrKeyNotFound:
 	default:
 		return err
@@ -160,6 +183,76 @@ func (d *badgerNodeDB) sanityCheckNamespace(ns common.Namespace) error {
 	if !ns.Equal(&d.namespace) {
 		return api.ErrBadNamespace
 	}
+	return nil
+}
+
+// Assumes metaUpdateLock is held when called.
+func (d *badgerNodeDB) cleanMultipartLocked(removeNodes bool) error {
+	var version uint64
+
+	if d.multipartVersion != multipartVersionNone {
+		version = d.multipartVersion
+	} else {
+		version = d.meta.getMultipartVersion()
+	}
+	if version == multipartVersionNone {
+		// No multipart in progress, but it's not an error to call in a situation like this.
+		return nil
+	}
+
+	txn := d.db.NewTransactionAt(tsMetadata, false)
+	defer txn.Discard()
+
+	opts := badger.DefaultIteratorOptions
+	opts.Prefix = multipartRestoreNodeLogKeyFmt.Encode()
+	it := txn.NewIterator(opts)
+	defer it.Close()
+
+	metaBatch := d.db.NewWriteBatchAt(tsMetadata)
+	defer metaBatch.Cancel()
+	nodeBatch := d.db.NewWriteBatchAt(versionToTs(version))
+	defer nodeBatch.Cancel()
+
+	var logged bool
+	for it.Rewind(); it.Valid(); it.Next() {
+		key := it.Item().Key()
+		if removeNodes {
+			if !logged {
+				d.logger.Info("removing some nodes from a multipart restore")
+				logged = true
+			}
+			var hash hash.Hash
+			if !multipartRestoreNodeLogKeyFmt.Decode(key, &hash) {
+				panic("mkvs/badger: bad iterator")
+			}
+			if err := nodeBatch.Delete(nodeKeyFmt.Encode(&hash)); err != nil {
+				return err
+			}
+		}
+		if err := metaBatch.Delete(key); err != nil {
+			return err
+		}
+	}
+
+	// Flush both batches first. If anything fails, having corrupt
+	// multipart info in d.meta shouldn't hurt us next run.
+	if err := nodeBatch.Flush(); err != nil {
+		return err
+	}
+	if err := metaBatch.Flush(); err != nil {
+		return err
+	}
+
+	metaTx := d.db.NewTransactionAt(tsMetadata, true)
+	defer metaTx.Discard()
+	if err := d.meta.setMultipartVersion(metaTx, 0); err != nil {
+		return err
+	}
+	if err := metaTx.CommitAt(tsMetadata, nil); err != nil {
+		return err
+	}
+
+	d.multipartVersion = multipartVersionNone
 	return nil
 }
 
@@ -414,6 +507,10 @@ func (d *badgerNodeDB) Finalize(ctx context.Context, version uint64, roots []has
 	d.metaUpdateLock.Lock()
 	defer d.metaUpdateLock.Unlock()
 
+	if d.multipartVersion != multipartVersionNone && d.multipartVersion != version {
+		return api.ErrInvalidMultipartVersion
+	}
+
 	// Version batch collects removals at the version timestamp.
 	versionBatch := d.db.NewWriteBatchAt(versionToTs(version))
 	defer versionBatch.Cancel()
@@ -563,6 +660,13 @@ func (d *badgerNodeDB) Finalize(ctx context.Context, version uint64, roots []has
 	if err := tx.CommitAt(tsMetadata, nil); err != nil {
 		return fmt.Errorf("mkvs/badger: failed to commit metadata: %w", err)
 	}
+
+	// Clean multipart metadata if there is any.
+	if d.multipartVersion != multipartVersionNone {
+		if err := d.cleanMultipartLocked(false); err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
@@ -573,6 +677,10 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
 
 	d.metaUpdateLock.Lock()
 	defer d.metaUpdateLock.Unlock()
+
+	if d.multipartVersion != multipartVersionNone {
+		return api.ErrMultipartInProgress
+	}
 
 	// Make sure that the version that we try to prune has been finalized.
 	lastFinalizedVersion, exists := d.meta.getLastFinalizedVersion()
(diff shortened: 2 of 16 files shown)


go/storage/mkvs/db/badger/badger.go
```diff
@@ -672,6 +686,56 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
 	return nil
 }
 
+func (d *badgerNodeDB) StartMultipartInsert() error {
```
kostko, line 689: This doesn't seem to actually toggle `d.multipart` anywhere?

go/storage/mkvs/db/badger/badger.go
```diff
@@ -672,6 +686,56 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
...
+	defer d.metaUpdateLock.Unlock()
+	txn := d.db.NewTransactionAt(tsMetadata, true)
+	defer txn.Discard()
+
+	opts := badger.DefaultIteratorOptions
+	opts.Prefix = multipartRestoreNodeLogKeyFmt.Encode()
+	it := txn.NewIterator(opts)
+	defer it.Close()
+	for it.Rewind(); it.Valid(); it.Next() {
+		key := it.Item().Key()
```
kostko, line 722: If something is found to remove, do log that we are cleaning up after a previously failed multipart insert (just once not for each item).

go/storage/mkvs/db/badger/badger.go
```diff
@@ -672,6 +686,56 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
...
+		if removeNodes {
+			var hash hash.Hash
+			if !multipartRestoreNodeLogKeyFmt.Decode(key, &hash) {
+				panic("mkvs/badger: bad iterator")
+			}
+			if err := txn.Delete(nodeKeyFmt.Encode(&hash)); err != nil {
```
kostko, line 728: Consider using a write batch (see `WriteBatch`) for deletion otherwise this needs to handle going over the transaction size limit (e.g., `ErrTxnTooBig`).

go/storage/mkvs/db/badger/badger.go
```diff
@@ -712,8 +784,9 @@ func (d *badgerNodeDB) Close() {
 type badgerBatch struct {
 	api.BaseBatch
 
-	db  *badgerNodeDB
-	bat *badger.WriteBatch
+	db    *badgerNodeDB
+	bat   *badger.WriteBatch
+	nodes *badger.WriteBatch
```
kostko, line 789: Maybe call it `multipartNodes` so that it is more clear what this is for?

go/storage/mkvs/db/badger/badger.go
```diff
@@ -891,6 +972,11 @@ func (s *badgerSubtree) PutNode(depth node.Depth, ptr *node.Pointer) error {
 	if err = s.batch.bat.Set(nodeKeyFmt.Encode(&h), data); err != nil {
 		return err
 	}
+	if s.batch.nodes != nil {
+		if err = s.batch.nodes.Set(multipartRestoreNodeLogKeyFmt.Encode(&h), []byte{}); err != nil {
```
kostko, line 976: There is a slight problem with doing this, namely you can only delete nodes in case they did not exist before otherwise you can end up removing existing nodes? You could just check that `ptr.Node.GetCreatedVersion()` is greater than the last version stored in the database (before the restore started) and only add the node in this case?

Hm but there may actually be some other issues when we support restoring checkpoints with pre-existing state in the database (e.g., it won't be actually possible to prune past the gap between the pre-restore version and the restored version). We may need to update `Prune` and add some tests to make sure this works.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -672,6 +686,56 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
 	return nil
 }
 
+func (d *badgerNodeDB) StartMultipartInsert() error {
+	if d.multipart {
```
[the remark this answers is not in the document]
jberci: Lapsus when I was adding the locks, thanks!

go/storage/mkvs/db/badger/badger.go
```diff
@@ -672,6 +686,56 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
...
+
+	txn := d.db.NewTransactionAt(tsMetadata, false)
+	defer txn.Discard()
+
+	opt := badger.DefaultIteratorOptions
+	opt.Prefix = multipartRestoreNodeLogKeyFmt.Encode()
+	it := txn.NewIterator(opt)
+	defer it.Close()
+	for it.Rewind(); it.Valid(); it.Next() {
+		return api.ErrMultipartInProgress
```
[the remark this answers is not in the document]
jberci: Doesn't, thanks. The choice is between doing the cleanup when opening the database or when trying to start a new restore. I had the former in mind and originally wrote the latter...

go/storage/mkvs/checkpoint/restorer.go
```diff
@@ -101,6 +103,7 @@ func (rs *restorer) RestoreChunk(ctx context.Context, idx uint64, r io.Reader) (
 		rs.pendingChunks = nil
 		rs.currentCheckpoint = nil
 
+		_ = rs.ndb.FinishMultipartInsert(false)
```
kostko, line 106: Shouldn't the argument be `true` as this is an abort? Could also just call `rs.AbortRestore(ctx)` here (and remove the separate lock/nilification)?

go/storage/mkvs/db/badger/badger.go
```diff
@@ -672,6 +686,57 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
...
+
+	txn := d.db.NewTransactionAt(tsMetadata, false)
+	defer txn.Discard()
+
+	opts := badger.DefaultIteratorOptions
+	opts.Prefix = multipartRestoreNodeLogKeyFmt.Encode()
+	it := txn.NewIterator(opts)
+	defer it.Close()
+
+	batch := d.db.NewWriteBatch()
```
kostko, line 714: This won't work, you need `NewWriteBatchAt(tsMetadata)`

go/storage/mkvs/db/badger/badger.go
```diff
@@ -680,11 +745,19 @@ func (d *badgerNodeDB) NewBatch(oldRoot node.Root, version uint64, chunk bool) a
 	// thing to do would be to either crank up MaxTableSize or maybe split
 	// the transaction out.
 
+	var logBatch *badger.WriteBatch
+	if d.multipart {
```
kostko, line 749: Read from `d.multipart` should require a mutex to be held?

go/storage/mkvs/db/badger/badger.go
```diff
@@ -680,11 +745,19 @@ func (d *badgerNodeDB) NewBatch(oldRoot node.Root, version uint64, chunk bool) a
...
 	return &badgerBatch{
-		db:      d,
-		bat:     d.db.NewWriteBatchAt(versionToTs(version)),
-		oldRoot: oldRoot,
-		chunk:   chunk,
+		db:             d,
+		bat:            d.db.NewWriteBatchAt(versionToTs(version)),
+		multipartNodes: logBatch,
+		oldRoot:        oldRoot,
+		chunk:          chunk,
```
kostko, line 760: Could we make sure that `chunk` must equal `d.multipart` otherwise batch operations error? It would be even better if `NewBatch` could return an error and then we could fail early.

go/storage/mkvs/checkpoint/restorer.go
```diff
@@ -117,7 +120,8 @@ func (rs *restorer) RestoreChunk(ctx context.Context, idx uint64, r io.Reader) (
 		rs.pendingChunks = nil
 		rs.currentCheckpoint = nil
 
-		return true, nil
+		err := rs.ndb.FinishMultipartInsert(false)
```
kostko, line 123: Actually this needs to be deferred until `Finalize` is called as otherwise you could still leave nodes in the database if the process crashes between this call and `Finalize`.

I suggest you change the API a bit as follows:
* There is no `FinishMultipartInsert(bool)`, instead there is `AbortMultipartInsert()` which behaves the same as `FinishMultipartInsert(false)`.
* `StartMultipartInsert` is changed to take a version which is stored together with the `multipart` flag. All versions passed to `NewBatch`, `Commit` and `Finalize` calls must match the `multipartVersion` otherwise this is an error.
* If you want to finish the multipart insert successfully you need to call `Finalize` with the `multipartVersion`.
* Calling `Prune` while a multipart insert is in progress is an error.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -163,6 +179,76 @@ func (d *badgerNodeDB) sanityCheckNamespace(ns common.Namespace) error {
...
+// Assumes metaUpdateLock is held when called.
+func (d *badgerNodeDB) cleanMultipartLocked(removeNodes bool) error {
+	var version uint64
+
+	if d.multipart {
+		version = d.multipartVersion
+	} else {
+		version = d.meta.getMultipartVersion()
+	}
+	if version == 0 {
```
kostko, line 191: Hm so a non-zero version is used as a multipart indicator? Why is a separate `d.multipart` flag needed then? Can it happen that `d.multipart == true && d.multipartVersion == 0`?

go/storage/mkvs/db/badger/badger.go
```diff
@@ -163,6 +179,76 @@ func (d *badgerNodeDB) sanityCheckNamespace(ns common.Namespace) error {
...
+		// No multipart in progress, but it's not an error to call in a situation like this.
+		return nil
+	}
+
+	txn := d.db.NewTransactionAt(tsMetadata, false)
+	defer txn.Discard()
+	opts := badger.DefaultIteratorOptions
```
kostko, line 198: Empty line above.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -672,20 +773,69 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
...
-func (d *badgerNodeDB) NewBatch(oldRoot node.Root, version uint64, chunk bool) api.Batch {
+func (d *badgerNodeDB) StartMultipartInsert(version uint64) error {
+	d.metaUpdateLock.Lock()
+	defer d.metaUpdateLock.Unlock()
+
+	if d.multipart {
+		if d.multipartVersion != version {
+			return api.ErrMultipartInProgress
+		}
+		// Multipart already initialized, so this was probably
```
kostko, line 784:
```suggestion
		// Multipart already initialized at the same version, so this was probably
```

go/storage/mkvs/db/badger/badger.go
```diff
@@ -672,20 +773,69 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
...
 	// the transaction out.
 
-	return &badgerBatch{
-		db:      d,
-		bat:     d.db.NewWriteBatchAt(versionToTs(version)),
-		oldRoot: oldRoot,
-		chunk:   chunk,
+	d.metaUpdateLock.Lock()
+	defer d.metaUpdateLock.Unlock()
+	if d.multipart && d.multipartVersion != version {
```
kostko, line 821: Empty line above.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -672,20 +773,69 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
...
+		return nil, api.ErrInvalidMultipartVersion
+	}
+
```
kostko, line 824: Since this method can now return errors, you can move the read only check from the top of the `Commit` method here.

go/storage/mkvs/db/badger/badger.go
jberci: I had just the flag initially, then left it in for slightly easier reasoning. But yeah, it's redundant.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -672,20 +776,73 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
 	return nil
 }
 
-func (d *badgerNodeDB) NewBatch(oldRoot node.Root, version uint64, chunk bool) api.Batch {
+func (d *badgerNodeDB) StartMultipartInsert(version uint64) error {
+	d.metaUpdateLock.Lock()
+	defer d.metaUpdateLock.Unlock()
+
```
kostko, line 782: Fail if `version` is `noMultipart`.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -672,20 +776,73 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
...
+		return nil, api.ErrReadOnly
+	}
+
+	d.metaUpdateLock.Lock()
+	defer d.metaUpdateLock.Unlock()
+
+	if d.multipartVersion != noMultipart && d.multipartVersion != version {
+		return nil, api.ErrInvalidMultipartVersion
+	}
+
```
kostko, line 831: Make sure that `chunk` is equal to `d.multipartVersion != noMultipart`, return error instead.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -891,6 +1056,11 @@ func (s *badgerSubtree) PutNode(depth node.Depth, ptr *node.Pointer) error {
 	if err = s.batch.bat.Set(nodeKeyFmt.Encode(&h), data); err != nil {
 		return err
 	}
+	if s.batch.multipartNodes != nil {
+		if err = s.batch.multipartNodes.Set(multipartRestoreNodeLogKeyFmt.Encode(&h), []byte{}); err != nil {
```
kostko, line 1060: As mentioned before, this can remove pre-existing nodes (e.g., if `h` existed before, this is not checked anywhere). You should only add a node to `multipartRestoreNodeLogKeyFmt` in case it didn't exist before. Also add a test for this to make sure (e.g., create a root at version 1, then restore a root at version 2 which shares some nodes with the root at version 1, but abort the multipart restore instead of finalizing it, then iterate over root at version 1 to make sure everything is still there).

go/storage/mkvs/db/badger/badger_test.go
```diff
@@ -0,0 +1,235 @@
...
+	"os"
+	"strconv"
+	"testing"
+
+	"github.com/dgraph-io/badger/v2"
+	"github.com/stretchr/testify/require"
+
+	"github.com/oasisprotocol/oasis-core/go/common"
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
+	"github.com/oasisprotocol/oasis-core/go/storage/api"
```
kostko, line 17: You should instead directly import the relevant package under `storage/mkvs`, the `api` package only re-exports.

go/storage/mkvs/db/badger/badger_test.go
```diff
@@ -0,0 +1,235 @@
...
+)
+
+var (
+	nodePrefix = nodeKeyFmt.Encode()
+
+	logPrefix = multipartRestoreNodeLogKeyFmt.Encode()
+
+	testNs = common.NewTestNamespaceFromSeed([]byte("badger nodedb test ns"), 0)
+
+	dbCfg = (&api.Config{
```
kostko, line 30: You should instead use `mkvs/db/api.Config`.

go/storage/mkvs/db/badger/badger_test.go
```diff
@@ -0,0 +1,235 @@
...
+type test struct {
+	require  *require.Assertions
+	ctx      context.Context
+	badgerdb *badgerNodeDB
+	restorer checkpoint.Restorer
+	ckMeta   *checkpoint.Metadata
+	ckNodes  keySet
+}
+
+func fillDB(ctx context.Context, require *require.Assertions, values [][]byte, nodedb api.NodeDB) api.Root {
```
kostko, line 55: Just to be consistent (also elsewhere).
```suggestion
func fillDB(ctx context.Context, require *require.Assertions, values [][]byte, ndb api.NodeDB) api.Root {
```

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -385,20 +375,19 @@ func (n *Node) checkCheckpointUsable(cp *checkpoint.Metadata, remainingMask outs
...
 	defer n.syncedLock.Unlock()
 
-	n.syncedState.LastBlock.Round = check.Root.Version
-	n.syncedState.LastBlock.Namespace = check.Root.Namespace
-	if mask == maskIO {
-		n.syncedState.LastBlock.IORoot = check.Root
-	} else if mask == maskState {
-		n.syncedState.LastBlock.StateRoot = check.Root
-	}
+	n.syncedState = syncState
```
kostko, line 382: This does not seem to update the persistent state (see `case finalized := <-n.finalizeCh` in `node.go`)?

go/storage/mkvs/db/badger/badger.go
```diff
@@ -672,20 +779,83 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
 	return nil
 }
 
-func (d *badgerNodeDB) NewBatch(oldRoot node.Root, version uint64, chunk bool) api.Batch {
+func (d *badgerNodeDB) StartMultipartInsert(version uint64) error {
+	d.metaUpdateLock.Lock()
+	defer d.metaUpdateLock.Unlock()
+
+	if version == multipartVersionNone {
+		return api.ErrInvalidMultipartVersion
```
kostko, line 787: Add a test that makes sure this fails.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -672,20 +779,83 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
...
+	}
+
+	if d.multipartVersion != multipartVersionNone {
+		if d.multipartVersion != version {
+			return api.ErrMultipartInProgress
```
kostko, line 792: Add a test that makes sure this fails.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -672,20 +779,83 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
...
 	// thing to do would be to either crank up MaxTableSize or maybe split
 	// the transaction out.
 
-	return &badgerBatch{
-		db:      d,
-		bat:     d.db.NewWriteBatchAt(versionToTs(version)),
-		oldRoot: oldRoot,
-		chunk:   chunk,
+	if d.readOnly {
+		return nil, api.ErrReadOnly
```
kostko, line 829: Add a test that makes sure this fails.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -672,20 +779,83 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
...
+	}
+
+	d.metaUpdateLock.Lock()
+	defer d.metaUpdateLock.Unlock()
+
+	if d.multipartVersion != multipartVersionNone && d.multipartVersion != version {
+		return nil, api.ErrInvalidMultipartVersion
```
kostko, line 836: Add a test that makes sure this fails.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -672,20 +779,83 @@ func (d *badgerNodeDB) Prune(ctx context.Context, version uint64) error {
...
 	}
+	if chunk != (d.multipartVersion != multipartVersionNone) {
+		return nil, api.ErrMultipartInProgress
```
kostko, line 839: Add a test that makes sure this fails.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -758,14 +933,13 @@ func (ba *badgerBatch) RemoveNodes(nodes []node.Node) error {
...
-	// XXX: Ideally this would fail at batch creation.
-	if ba.db.readOnly {
-		return api.ErrReadOnly
-	}
-
 	ba.db.metaUpdateLock.Lock()
 	defer ba.db.metaUpdateLock.Unlock()
 
+	if ba.db.multipartVersion != multipartVersionNone && ba.db.multipartVersion != root.Version {
+		return api.ErrInvalidMultipartVersion
```
kostko, line 940: Add a test that makes sure this fails.

kostko (approved):

PR #3290, opened 2020-09-18
go/oasis-node/cmd/genesis: Add convert CLI command

tjanez:
It can be used to convert a given genesis file to the canonical form.

### DIFF
--- a/.changelog/3290.doc.md
+++ b/.changelog/3290.doc.md
@@ -1,0 +1,3 @@
+Document [`oasis-node genesis` CLI commands]
+
+[`oasis-node genesis` CLI commands]: docs/oasis-node/cli.md#genesis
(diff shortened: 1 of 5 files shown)


Yawning:
Why do we need this if it's a one time thing (tooling produces canonical output by default)?

tjanez:
> Why do we need this if it's a one time thing (tooling produces canonical output by default)?

I guess it could be useful more than once (I'm guessing you meant for converting the proposed Mainnet genesis file) since people could use external tools to assemble a genesis file or do some modifications to a genesis file and then use this CLI command to convert it back to the canonical form.

Yawning:
I am against this.

 * The utility seems dubious at best.
 * There is no need for this to be a sub-command of oasis-node.
 * There are already too many sub-commands that quite honestly should be moved to elsewhere.

The first step to fixing the sub-command situation is, not actively making the problem worse.

tjanez:
> I am against this.

Yea, agreed. The usefulness vs. maintenance burden ratio is too low.

If we need this for a one-off thing, I can do the conversion manually.

Issue #3289, opened 2020-09-18
RHP connection should set write deadline


kostko:
The RHP connection should set a write deadline before writing to make sure that the connection is closed in case the runtime is broken.

PR #3292, opened 2020-09-18
go/runtime/host/protocol: Add write timeouts

ptrus:
Fixes: #3289

### DIFF
--- a/go/runtime/host/protocol/connection.go
+++ b/go/runtime/host/protocol/connection.go
@@ -21,7 +21,11 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common/metrics"
 )
 
-const moduleName = "rhp/internal"
+const (
+	moduleName = "rhp/internal"
+
+	connWriteTimeout = 5 * time.Second
+)
 
 // ErrNotReady is the error reported when the Runtime Host Protocol is not initialized.
 var (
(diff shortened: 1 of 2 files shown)


Yawning:
While I expect this sort of thing to be exceedingly rare, should it be configurable? (No is a fine answer).

go/runtime/host/protocol/connection.go
```diff
@@ -321,12 +325,22 @@ func (c *connection) workerOutgoing() {
...
+					"err", err,
+				)
+			}
 			// Outgoing message, send it.
 			if err := c.codec.Write(msg); err != nil {
 				c.logger.Error("error while sending message",
 					"err", err,
 				)
 			}
+			if err := c.conn.SetWriteDeadline(time.Time{}); err != nil {
```
kostko, line 339: Since `Write` is only ever called here, do we need to reset the deadline?

ptrus:
> should it be configurable

Probably no need specially since on the runtime side the requests get queued so the write should not really get blocked by anything in case everything is fine.

Yawning (approved):

go/runtime/host/protocol/connection.go
ptrus: need no, but seemed like a reasonable thing to do (specially since connection is passed into this struct and not created here).

kostko: Ok.

kostko (approved):

Issue #3245, opened 2020-09-04
Update rust dependencies and remove from dependabot ignore list


ptrus:
Some rust dependencies are out-of-date and are set to be ignored by `dependabot` until updated.

See the list of dependencies that need to be updated: https://github.com/oasisprotocol/oasis-core/blob/master/.github/dependabot.yml#L49-L69

PR #3294, opened 2020-09-21
rust: bump serde crates

ptrus:
Part of: #3245

TODO:
-  [x] `serde_bytes`


runtime/src/enclave_rpc/types.rs
```diff
@@ -25,7 +25,6 @@ impl SessionID {
 /// Frame.
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Frame {
-    #[serde(with = "serde_bytes")]
```
ptrus, line 28: Note: `serde_bytes` now internally uses `serde_bytes::{Serialize,Deserialize}` traits, and is missing implementations for fixed sized byte arrays, so we cannot use it here anymore. 

However `serde_bytes` is not even needed here since we implement "byte-aware" serde serialization and deserialization methods in `impl_bytes!` anyway.

Yawning (approved):

PR #3300, opened 2020-09-22
go/worker/executor: clear tx queue before returning to committee

ptrus:


go/worker/compute/executor/committee/node.go
```diff
@@ -150,8 +150,9 @@ type Node struct { // nolint: maligned
 	// Guarded by .commonNode.CrossNode.
 	proposingTimeout bool
 
-	commonNode *committee.Node
+	prevEpochInCommittee bool
```
kostko, line 153: Should probably also add the notice that the flag is guarded by `.commonNode.CrossNode` (or just move it to the group above).

ptrus: done :+1:

kostko (approved):

PR #3302, opened 2020-09-22
queries workload eof

ptrus:
- don't fail on `EOF` on `GetTransactionsWithResults`
- enable `TendermintRecoverCorruptedWAL` also for non-validators since the upstream recovery doesn't seem to work all that well

### DIFF
--- a/go/oasis-node/cmd/debug/txsource/workload/queries.go
+++ b/go/oasis-node/cmd/debug/txsource/workload/queries.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"math/rand"
 	"time"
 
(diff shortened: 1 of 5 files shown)


go/oasis-node/cmd/debug/txsource/workload/queries.go
```diff
@@ -232,7 +233,17 @@ func (q *queries) doConsensusQueries(ctx context.Context, rng *rand.Rand, height
 			"txs", txs,
 			"txs_with_results", txsWithRes,
 			"height", height,
+			"err", err,
 		)
+		if status := cmnGrpc.GetErrorStatus(err); status != nil {
+			if status.Err() == io.ErrUnexpectedEOF {
+				// XXX: Connection seems to get occasionally reset with
```
kostko, line 240: Maybe also open an issue and reference it in the comment.

kostko (approved):

PR #3304, opened 2020-09-22
go/worker/executor: clear tx queue before becoming a worker

ptrus:
`BackupWorker` also doesn't follow transactions and processed batches so should also reset the queue in that case. 

Not sure how i missed this in: https://github.com/oasisprotocol/oasis-core/pull/3300

### DIFF
--- a/go/worker/compute/executor/committee/node.go
+++ b/go/worker/compute/executor/committee/node.go
@@ -148,8 +148,8 @@ type Node struct { // nolint: maligned
 	scheduler      schedulingAPI.Scheduler
 
 	// Guarded by .commonNode.CrossNode.
-	proposingTimeout     bool
-	prevEpochInCommittee bool
+	proposingTimeout bool
+	prevEpochWorker  bool
 
 	commonNode   *committee.Node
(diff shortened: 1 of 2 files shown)


kostko (approved):

go/worker/compute/executor/committee/node.go
```diff
@@ -457,20 +457,18 @@ func (n *Node) HandleEpochTransitionLocked(epoch *committee.EpochSnapshot) {
 			}
 			incomingQueueSize.With(n.getMetricLabels()).Set(float64(n.scheduler.UnscheduledSize()))
 		}
-		fallthrough
-	case epoch.IsExecutorBackupWorker():
-		if !n.prevEpochInCommittee {
+		if !n.prevEpochWorker {
 			// Clear incoming queue and cache of any stale transactions in case
-			// we were not part of the committee in previous epoch.
+			// we were not part of the worker committee in previous epoch.
```
kostko, line 462:
```suggestion
			// we were not part of the compute committee in previous epoch.
```

PR #3310, opened 2020-09-23
changelog: Assemble changes for release 20.11

ptrus:
TODO: should we bump any protocol versions? (there were not really any breaking changes, could potentially bump some minor versions?)


kostko:
Were there any protocol changes? If not, we should not bump the versions.

ptrus:
> Were there any protocol changes? If not, we should not bump the versions.

Hm yeah actually now that i think about it we did change how the libp2p message ID is computed, that's probably a breaking change and should be a Major version upgrade to `RuntimeCommitteeProtocol`?

kostko:
I think the message ID function is a local-only function [only used for duplicate message detection](https://github.com/libp2p/go-libp2p-pubsub/blob/a3445b756fdb68f372dadd45de65a723d0237ac8/pubsub.go#L1025-L1030), so it shouldn't be problematic.

ptrus:
Yeah this should probably still deserve a `Patch` version update, but since that would currently result in a breaking change to the p2p topic id's https://github.com/oasisprotocol/oasis-core/blob/61de25e20457fffd47df39bf7188096d036db5fb/go/worker/common/p2p/p2p.go#L233-L235 we should probably just not bump versions at this time (should still open an issue that changing the patch version does not result in the change of a topic id, as these should generally be non-breaking changes)

kostko:
> should still open an issue that changing the patch version does not result in the change of a topic id, as these should generally be non-breaking changes

Agreed.

kostko (approved):

PR #3316, opened 2020-09-24
Omit node operator link from release notes

### LINKED ISSUE #3313: Broken link to operator docs in release notes
In release notes we link to a non-existing page:
> If you would like to become a node operator for the Oasis Network, see the [Operator Docs](https://docs.oasis.dev/operators/overview.html).

Should either fix the link or could even consider omitting that part from the release notes.

kostko:
Fixes #3313

### DIFF
--- a/.changelog/3313.internal.md
+++ b/.changelog/3313.internal.md
@@ -1,0 +1,1 @@
+Omit node operator link from release notes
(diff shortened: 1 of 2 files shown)


ptrus (approved):

tjanez (approved):
Yea, this makes sense.

PR #3317, opened 2020-09-24
go/consensus/tendermint: Bump Tendermint Core to v0.34-rc4-oasis3

kostko:
TODO
* [x] Switch to released tag.


Yawning (approved):

pro-wh (approved):
why did we tag -rc4 before upstream did?

Issue #3318, opened 2020-09-24
P2P Dispatch should always authenticate own messages


ptrus:
Authenticating the peer in dispatched messages should be skipped for own messages:
https://github.com/oasisprotocol/oasis-core/blob/1b7ede7f09a62b13abfe1fa9656b491245ab3509/go/worker/common/p2p/dispatch.go#L167-L174

The practical consequence of this is that currently if we are using an executor node for submitting runtime transactions, the node will self-reject messages whenever it's not part of the committee.

PR #3319, opened 2020-09-24
go/worker/p2p: Skip peer authentication for our own messages

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3318


go/worker/common/p2p/dispatch.go
```diff
@@ -165,6 +165,10 @@ func (h *topicHandler) dispatchMessage(peerID core.PeerID, m *queuedMsg, isIniti
 	defer h.handlersLock.RUnlock()
 
 	for _, handler := range h.handlers {
+		// Skip authentication if we are the peer.
```
kostko, line 168: Would probably make more sense to just skip the for loop instead of breaking on the first iteration (as you don't need the `handler` for this check)?

kostko (approved):

PR #3320, opened 2020-09-24
go/worker/executor: don't hold the lock during storage requests

ptrus:
Don't unnecessarily hold the lock during external requests to storage nodes, as those can take long time. 

Issue discovered in the daily tests with long restarts, when the storage node is taken offline, these requests can take up to 15 sec to timeout.

### DIFF
--- a/go/worker/compute/executor/committee/node.go
+++ b/go/worker/compute/executor/committee/node.go
@@ -752,37 +752,43 @@ func (n *Node) proposeTimeoutLocked() error {
 
 // Dispatch dispatches a batch to the executor committee.
 func (n *Node) Dispatch(batch transaction.RawBatch) error {
-	n.commonNode.CrossNode.Lock()
-	defer n.commonNode.CrossNode.Unlock()
+	lastHeader, err := func() (*block.Header, error) {
+		n.commonNode.CrossNode.Lock()
+		defer n.commonNode.CrossNode.Unlock()
 
-	// If we are not waiting for a batch, don't do anything.
-	if _, ok := n.state.(StateWaitingForBatch); !ok {
-		return errIncorrectState
-	}
-	if n.commonNode.CurrentBlock == nil {
-		return errNoBlocks
-	}
-
-	epoch := n.commonNode.Group.GetEpochSnapshot()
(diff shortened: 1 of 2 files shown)


go/worker/compute/executor/committee/node.go
```diff
@@ -792,7 +797,9 @@ func (n *Node) Dispatch(batch transaction.RawBatch) error {
 
 	// Generate the initial I/O root containing only the inputs (outputs and
 	// tags will be added later by the executor nodes).
+	n.commonNode.CrossNode.Lock()
 	lastHeader := n.commonNode.CurrentBlock.Header
```
kostko, line 801: Would be best if this was returned by the function above that takes the lock.

go/worker/compute/executor/committee/node.go
```diff
@@ -841,6 +848,8 @@ func (n *Node) Dispatch(batch transaction.RawBatch) error {
 	}
 	spanInsert.Finish()
 
+	n.commonNode.CrossNode.Lock()
+	defer n.commonNode.CrossNode.Unlock()
```
kostko, line 852: In the section below (sorry can't comment on it), we should use `lastHeader` instead of accessing `n.commonNode.CurrentBlock` to make sure things are consistent (`n.commonNode.CurrentBlock` could change while you are not holding the lock).

After this is changed, the lock can possibly be moved just before the `handleInternalBatchLocked` call (group publish should not require holding the lock).

ptrus: yeah good idea :+1:

go/worker/compute/executor/committee/node.go
```diff
@@ -886,6 +891,8 @@ func (n *Node) Dispatch(batch transaction.RawBatch) error {
 	crash.Here(crashPointBatchPublishAfter)
 	spanPublish.Finish()
 
+	n.commonNode.CrossNode.Lock()
+	defer n.commonNode.CrossNode.Unlock()
```
kostko, line 895: After re-taking the lock this should make sure that the round is still the same and that we are still in the correct state. Both may have changed while we were not holding the lock.

kostko (approved):

PR #3324, opened 2020-09-25
go/worker/executor: reinsert tx batch in case round was not successful

ptrus:
Before we only reinserted the tx batch in case it failed during processing or on epoch transitions. Also reinsert the txs in case the round failed after commitment was already submitted.

### DIFF
--- a/go/worker/compute/executor/committee/node.go
+++ b/go/worker/compute/executor/committee/node.go
@@ -232,6 +232,16 @@ func (n *Node) clearQueuedTxs() {
 	incomingQueueSize.With(n.getMetricLabels()).Set(0)
 }
 
+// Assumes scheduler is initialized.
+func (n *Node) reinsertTxs(txs transaction.RawBatch) {
+	if err := n.scheduler.AppendTxBatch(txs); err != nil {
+		n.logger.Warn("error while reinserting transactions",
+			"err", err,
+		)
+	}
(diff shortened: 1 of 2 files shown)


go/worker/compute/executor/committee/node.go
```diff
@@ -534,15 +539,17 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
+			if header.HeaderType != block.Normal ||
+				!header.IORoot.Equal(&state.proposedIORoot) {
 				n.logger.Error("proposed batch was not finalized",
 					"header_io_root", header.IORoot,
 					"proposed_io_root", state.proposedIORoot,
+					"header_type", header.HeaderType,
 					"batch", state.raw,
 				)
+				// Reinsert the failed batch.
+				n.reinsertTxs(state.raw)
```
kostko, line 552: Hm, aren't the processed transactions only removed below (line 564)? Where have they been removed in this case so they need to be reinserted?

ptrus: In case we are the "proposer" (scheduler), the txs are removed at `Dispatch`, but actually a good point, if we switch the logic to only remove the transactions after a successful round finalization (and not at dispatch time) we can probably make the code a lot simpler.

Will look into doing that

ptrus:
Closed in favor of https://github.com/oasisprotocol/oasis-core/pull/3332

PR #3332, opened 2020-09-28
go/worker/executor: only remove incoming transactions on finalized round

ptrus:


go/runtime/scheduling/simple/incoming_queue.go
```diff
@@ -169,68 +168,66 @@ func (q *incomingQueue) AddBatch(batch transaction.RawBatch) error {
 	return errs
 }
 
-// Take attempts to take a batch from the incoming queue.
-func (q *incomingQueue) Take(force bool) (transaction.RawBatch, error) {
+// GetBatch attempts to get a batch from the incoming queue.
```
kostko, line 171: Add a comment noting that it may return nil in case a batch is not available.

kostko (approved):

go/runtime/scheduling/simple/incoming_queue.go
ptrus: So `transaction.RawBatch` is of type `[][]byte`, so it will return a `nil` slice in that case, which i think it doesn't need a special note?

Unless you mean the fact that there could be txs in the queue but unless there's enough txs for a barch (or `force=true`) it won't return a batch?

Issue #2078, opened 2019-09-11
Slashing


kostko:
Slashing should be implemented based on the set of slashing conditions. Ideally there should be E2E tests that verify that slashing functions correctly (e.g., using the byzantine node to invoke incorrect behavior).

### Slashing conditions

* [x] Runtimes / Compute committee
  * [x] Compute/sign/send a wrong result
  * [x] Sign two commitments for the same round but different I/O root/state root/message root.
* [ ] Runtimes / Storage committee
  * [ ] Return a wrong result to read request
  * [ ] Longer-term loss of availability
* [x] Consensus committee
  * [x] [Signing different votes for the same height](https://github.com/oasislabs/oasis-core/issues/1901)

kostko:
All runtime-related slashing conditions (and thresholds to run nodes) should probably be configurable per-runtime.

kostko:
In addition to slashing, misbehaving nodes should also be frozen (similar to how validators are frozen for double signing), with freezing maybe being configurable per runtime (maybe in addition to there being global parameters?).

PR #3339, opened 2020-09-29
ADR 0005: Runtime Compute Node Slashing

kostko:
See #2078

### DIFF
--- a/docs/adr/0005-runtime-compute-slashing.md
+++ b/docs/adr/0005-runtime-compute-slashing.md
@@ -1,0 +1,372 @@
+# ADR 0005: Runtime Compute Node Slashing
+
+## Changelog
+
+- 2020-10-14: Evidence expiry, duplicate evidence detection
+- 2020-09-28: Initial draft
+
+## Status
+
+Accepted
+
+## Context
+
+The runtime compute nodes make updates to the runtime state by submitting
+commitment messages to the roothash service in the consensus layer where
+discrepancy detection and resolution are performed.
+
+Currently, the compute nodes are never slashed even if they commit incorrect
+results. While integrity is guarded by discrepancy detection and resolution,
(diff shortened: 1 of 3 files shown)


docs/adr/0005-runtime-compute-slashing.md
```diff
@@ -0,0 +1,284 @@
...
+
+Any slashing instructions related to freezing nodes are currently ignored.
+
+### Transaction Methods
+
+This proposal updates the following transaction methods in the roothash module:
+
+#### Evidence
+
+The evidence method allows anyone to submit evidence of runtime node
```
bennetyee, line 175: Anyone can submit evidence of double signing, but there's no evidence format to be committed to support discrepancy resolution based slashing.  Do we want/need this?

docs/adr/0005-runtime-compute-slashing.md
```diff
@@ -0,0 +1,367 @@
...
+  the number of rounds.
+
+## Consequences
+
+### Positive
+
+- Compute nodes can be disincentivized to submit incorrect results by runtimes
+  configuring slashing parameters.
+
+### Negative
```
peterjgilbert, line 351: Something to call out as either a negative or neutral consequence: if a runtime exhibits nondeterminism, this can result in a compute node being slashed. While we specify that runtimes should be deterministic, for non-SGX runtimes we have no way determining whether a discrepancy is due to runtime nondeterminism or a faulty compute node.

peterjgilbert: Ok, it's probably "negative" 😄

kostko: Added.

bennetyee (approved):

peterjgilbert (approved):

PR #3342, opened 2020-09-30
go/executor/node: check if we are in correct state before publishing a message

ptrus:
Currently (since https://github.com/oasisprotocol/oasis-core/pull/3320) this check is made after a node publishes a message, instead this check should be made before node publishes a message to prevent proposing another batch after node starts processing a batch.
Note that `p2p.Publish` is async so this shouldn't increase the time the CrossNode.Lock is held by much.

### DIFF
--- a/.changelog/3342.trivial.md
+++ b/.changelog/3342.trivial.md
@@ -1,0 +1,1 @@
+go/executor/node: ensure correct state before publishing a message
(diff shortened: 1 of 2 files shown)


.changelog/3342.bugfix.md
```diff
@@ -0,0 +1 @@
+go/executor/node: ensure correct state before publishing a message
```
kostko, line 1: Could be changed to trivial as it fixes something that was not released yet.

kostko (approved):

.changelog/3342.bugfix.md
ptrus: oh right, yeah agreed, otherwise it's probably confusing

Issue #3311, opened 2020-09-23
P2P topic IDs should not include Minor/Patch versions of the RuntimeCommitteeProtocol


ptrus:
P2P topic IDs include `RuntimeCommitteeProtocol` version: 
https://github.com/oasisprotocol/oasis-core/blob/61de25e20457fffd47df39bf7188096d036db5fb/go/worker/common/p2p/p2p.go#L233-L235

However since we are following SemVer, `Minor` and `Patch` version changes to the `RuntimeCommitteeProtocol` should be backwards compatible therefore also not resulting in a changed topic ID.

ptrus:
The topics should probably also include chain IDs (or something similar to separate different network meshes).

kostko:
Yeah so while the messages themselves use the chain domain separation context so they wouldn't be valid, we should probably separate the meshes to avoid needless messages.

PR #3346, opened 2020-09-30
p2p topic ids: include chain ID and remove patch version

ptrus:
Fixes: #3311

Note: used `MajorMinor()` since we also used that for runtime-host protocol.

This is a breaking `RuntimeCommitteeProtocol` change.

### DIFF
--- a/go/worker/common/p2p/p2p.go
+++ b/go/worker/common/p2p/p2p.go
@@ -44,6 +44,8 @@ type P2P struct {
 
 	ctx context.Context
 
+	chainContext string
+
 	host   core.Host
 	pubsub *pubsub.PubSub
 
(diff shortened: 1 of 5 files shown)


go/worker/common/p2p/p2p.go
```diff
@@ -212,9 +222,15 @@ func New(ctx context.Context, identity *identity.Identity, consensus consensus.B
...
 
+	doc, err := consensus.GetGenesisDocument(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("worker/common/p2p: failed to get consensus genesis document: %w", err)
+	}
+
 	p := &P2P{
 		PeerManager:       newPeerManager(ctx, host, consensus),
 		ctx:               ctx,
+		chainID:           doc.ChainID,
```
kostko, line 233: This should use `doc.ChainContext()` instead (and maybe call the field `chainContext`).

go/worker/common/p2p/p2p.go
```diff
@@ -158,6 +160,14 @@ func (p *P2P) handleConnection(conn core.Conn) {
 	)
 }
 
+func (p *P2P) topicIDForRuntime(runtimeID common.Namespace) string {
+	return fmt.Sprintf("%s/%s/%s",
```
kostko, line 164: BTW, is there a limit on topic length?

ptrus: Doesn't seem like it. The type is `string` and I don't see any checks for length through the code, also the [specs](https://github.com/libp2p/specs/tree/master/pubsub#the-topic-descriptor) don't specify any such limit.

kostko: Ok, just making sure we don't run into any limits :)

kostko (commented):
> This is a breaking RuntimeCommitteeProtocol change.

Update the version to 2.0.0?

kostko (approved):

PR #3348, opened 2020-09-30
go consensus: add last retained block information

pro-wh:
want this for https://github.com/oasisprotocol/oasis-core-rosetta-gateway/issues/57

### DIFF
--- a/go/consensus/tendermint/full/full.go
+++ b/go/consensus/tendermint/full/full.go
@@ -821,6 +821,17 @@ func (t *fullService) GetStatus(ctx context.Context) (*consensusAPI.Status, erro
 			// We may not be able to fetch the genesis block in case it has been pruned.
 		}
 
+		lastRetainedHeight, err := t.GetLastRetainedVersion(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get last retained height: %w", err)
+		}
+		status.LastRetainedHeight = lastRetainedHeight
+		lastRetainedBlock, err := t.GetBlock(ctx, lastRetainedHeight)
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch last reatianed block: %w", err)
+		}
+		status.LastRetainedHash = lastRetainedBlock.Hash
+
 		// Latest block.
 		latestBlk, err := t.GetBlock(ctx, consensusAPI.HeightLatest)
 		switch err {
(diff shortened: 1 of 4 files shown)


go/consensus/tendermint/full/full.go
```diff
@@ -821,6 +821,17 @@ func (t *fullService) GetStatus(ctx context.Context) (*consensusAPI.Status, erro
 			// We may not be able to fetch the genesis block in case it has been pruned.
 		}
 
+		lastRetainedHeight, err := t.GetLastRetainedVersion(ctx)
```
pro-wh, line 824: is "version" the same number?

abukosek (approved):
Should this change also bump the `ConsensusProtocol` version in `go/common/version/version.go`?

kostko:
> Should this change also bump the ConsensusProtocol version in go/common/version/version.go?

Nope, it's not a breaking consensus change, just adding some fields to the status response.

go/consensus/tendermint/full/full.go
kostko: Yeah "version" is a generic term used by the MKVS database and it can either represent height (consensus state) or round (runtime state).

kostko (approved):

go/consensus/tendermint/full/full.go
```diff
@@ -821,6 +821,21 @@ func (t *fullService) GetStatus(ctx context.Context) (*consensusAPI.Status, erro
...
+			return nil, fmt.Errorf("failed to get last retained height: %w", err)
+		}
+		// Some pruning configurations return 0 instead of a valid block height. Clamp those to the genesis height.
+		if lastRetainedHeight < t.genesis.Height {
+			lastRetainedHeight = t.genesis.Height
+		}
+		status.LastRetainedHeight = lastRetainedHeight
+		lastRetainedBlock, err := t.GetBlock(ctx, lastRetainedHeight)
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch last reatianed block: %w", err)
```
kostko, line 835: This shouldn't fail the method as otherwise no status will be returned. This currently fails the `e2e/early-query` test since there are no blocks available yet (and the status endpoint should work even when there are no blocks). Make sure to just not set the hash as above (and possibly set the last retained height to zero in this case as there are no blocks).

```
failed to get consensus status: failed to fetch last reatianed block: tendermint: block query failed: height 1 must be less than or equal to the current blockchain height 0
```

pro-wh: this is guarded by t.started. is that something different?

pro-wh: oh probably this is from the time between when it starts and when the first block is confirmed

kostko: Yeah that's right.

pro-wh: alright, changing it to handling similar to the above

go/consensus/tendermint/full/full.go
pro-wh: ok good. couldn't figure out if it was off by one or something. or possibly by more than one if the initial height is not 1

Issue #3276, opened 2020-09-14
Use human readable identifiers for slashing reasons


kostko:
When serialized into JSON the slashing reason identifiers should be human readable (e.g., `double-signing` instead of `0`).

PR #3355, opened 2020-10-01
go: human readable slashing reasons

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3276

### DIFF
--- a/go/staking/api/slashing.go
+++ b/go/staking/api/slashing.go
@@ -1,6 +1,8 @@
 package api
 
 import (
+	"fmt"
+
 	"github.com/oasisprotocol/oasis-core/go/common/quantity"
 	epochtime "github.com/oasisprotocol/oasis-core/go/epochtime/api"
 )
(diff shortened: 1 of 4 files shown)


go/staking/api/slashing.go
```diff
@@ -11,20 +13,37 @@ type SlashReason int
 const (
 	// SlashDoubleSigning is slashing due to double signing.
 	SlashDoubleSigning SlashReason = 0
+	SlashMax                       = SlashDoubleSigning
```
kostko, line 16: I don't think we use this anywhere so it could just be removed.

go/staking/api/slashing.go
```diff
@@ -11,20 +13,37 @@ type SlashReason int
...
-		return "double-signing"
+		return SlashDoubleSigningName
 	default:
 		return "[unknown slash reason]"
 	}
 }
 
+// MarshalText encodes a SlashReason into text form.
+func (s SlashReason) MarshalText() ([]byte, error) {
+	return []byte(s.String()), nil
```
kostko, line 33: That's not the best way as it could encode as `[unknown slash reason]`. We should probably use a separate switch and return an error in that case?

go/staking/api/slashing.go
ptrus: yeah indeed it's not :+1:

kostko (approved):
Looks good, maybe just add a serialization round-trip test for all supported reasons and a test with invalid values?

PR #3357, opened 2020-10-01
Prioritize nodes that signed storage receipts

### LINKED ISSUE #3354: Storage client should prioritize node(s) specified in storage receipts
Currently the executor tries to fetch storage from any committee node. Instead of doing this blindly, it should prioritize by fetching from storage nodes specified in the storage receipt as those are the ones that should have the given root. The prioritized node identifiers could be hinted by setting a context key.

Related to #1815.

kostko:
Fixes #3354

### DIFF
--- a/go/storage/client/client.go
+++ b/go/storage/client/client.go
@@ -291,24 +291,40 @@ func (b *storageClientBackend) readWithClient(
 ) (interface{}, error) {
 	var resp interface{}
 	op := func() error {
-		conns := b.committeeClient.GetConnectionsWithMeta()
-		n := len(conns)
-		if n == 0 {
+		conns := b.committeeClient.GetConnectionsMap()
+		if len(conns) == 0 {
 			b.logger.Error("readWithClient: no connected nodes for runtime",
 				"runtime_id", ns,
 			)
 			return ErrStorageNotAvailable
 		}
 
+		var nodes []*committee.ClientConnWithMeta
+		// If a storage node priority hint is set, prioritize overlapping nodes.
+		for _, nodeID := range api.NodePriorityHintFromContext(ctx) {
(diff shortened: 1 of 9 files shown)


go/storage/client/client.go
```diff
@@ -304,10 +304,26 @@ func (b *storageClientBackend) readWithClient(
...
+			for _, nodeID := range nodes {
+				for i, c := range conns {
+					if c.Node.ID.Equal(nodeID) {
+						indices = append(indices, i)
+						break
+					}
+				}
+			}
+
+			nodeOrder = append(indices, nodeOrder...)
```
ptrus, line 321: This will result in duplicates, so if a connection to a "priority" node is not working (or the request errors for whatever reason), it can get queried twice. Which is technically ok, but might be nicer cleaning this of duplicates or at least adding a note about it.

kostko: Yeah I thought about it, but it didn't seem to be worth the added complexity. I'll think about it some more :)

go/storage/client/client.go
```diff
@@ -305,9 +305,33 @@ func (b *storageClientBackend) readWithClient(
...
+					}
+				}
+				if prioritySlot > -1 {
+					prioritySlots--
+					nodeOrder[prioritySlot] = i
+					continue
+				}
+			}
+			// Otherwise add to some random slot after the priority slots.
+			nodeOrder[rng.Intn(n-len(priorityNodes))+len(priorityNodes)] = i
```
ptrus, line 329: I don't think this last part is correct as it can insert into the same random slot multiple times.

kostko: Yeah it's not.

ptrus (approved):

PR #3360, opened 2020-10-01
Compare protocol versions according to SemVer 2.0.0

tjanez:
Since we bumped the protocol versions to version 1.0.0 with the release of Oasis Core 20.10, we also need to modify how we compare and detect backward-incompatible changes to follow SemVer 2.0.0 rules.

From now onwards, only a change in a protocol's major version signifies a backward-incompatible change.

NOTE: The change in this PR would require a major bump in runtime committee protocol's version, but since it was already bumped in #3346, we don't need to bump it again.

### DIFF
--- a/go/consensus/tendermint/full/statesync.go
+++ b/go/consensus/tendermint/full/statesync.go
@@ -59,7 +59,7 @@ func (sp *stateProvider) State(ctx context.Context, height uint64) (tmstate.Stat
 		InitialHeight: sp.genesisDocument.InitialHeight,
 	}
 	// XXX: This will fail in case an upgrade happened in-between.
-	state.Version.Consensus.App = version.ConsensusProtocol.ToU64()
+	state.Version.Consensus.App = version.TendermintAppVersion
 
 	// The snapshot height maps onto the state heights as follows:
 	//
--- a/go/runtime/host/protocol/connection.go
+++ b/go/runtime/host/protocol/connection.go
@@ -530,7 +530,7 @@ func (c *connection) InitHost(ctx context.Context, conn net.Conn) (*version.Vers
 	}
 
 	info := rsp.RuntimeInfoResponse
-	if ver := version.FromU64(info.ProtocolVersion); ver.MajorMinor() != version.RuntimeHostProtocol.MajorMinor() {
+	if ver := version.FromU64(info.ProtocolVersion); ver.Major != version.RuntimeHostProtocol.Major {
 		c.logger.Error("runtime has incompatible protocol version",
 			"version", ver,
 			"expected_version", version.RuntimeHostProtocol,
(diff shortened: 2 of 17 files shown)


go/runtime/host/protocol/connection.go
```diff
@@ -530,7 +530,7 @@ func (c *connection) InitHost(ctx context.Context, conn net.Conn) (*version.Vers
 	}
 
 	info := rsp.RuntimeInfoResponse
-	if ver := version.FromU64(info.ProtocolVersion); ver.MajorMinor() != version.RuntimeHostProtocol.MajorMinor() {
+	if ver := version.FromU64(info.ProtocolVersion); ver.Major != version.RuntimeHostProtocol.Major {
```
ptrus, line 533: can also remove `MajorMinor()` as here, and bellow in p2p are the only places where it's used.

kostko: Actually we could probably just change it to use `.Major` but this would be incompatible with `1.0.0`.

tjanez: > Actually we could probably just change it to use `.Major` but this would be incompatible with `1.0.0`.

I understand. Are there any objections to bumping runtime host protocol version to 2.0.0?

kostko: That statement was referring to the consensus protocol where this would be a breaking change. I don't think we need to bump the RHP version as this doesn't really break anything for the current protocol.

tjanez: > I don't think we need to bump the RHP version as this doesn't really break anything for the current protocol.

Yes, agreed.

go/consensus/tendermint/full/statesync.go
```diff
@@ -59,7 +59,7 @@ func (sp *stateProvider) State(ctx context.Context, height uint64) (tmstate.Stat
 		InitialHeight: sp.genesisDocument.InitialHeight,
 	}
 	// XXX: This will fail in case an upgrade happened in-between.
-	state.Version.Consensus.App = version.ConsensusProtocol.ToU64()
+	state.Version.Consensus.App = version.TendermintAppVersion
```
kostko, line 62: While reviewing this line I discovered that the state sync test was broken (see #3369). Thanks @tjanez and Codecov inline warnings! :-)

kostko:
Rebase and don't forget updates due to https://github.com/oasisprotocol/oasis-core/pull/3361#discussion_r498919678.

tjanez:
> Rebase and don't forget updates due to [#3361 (comment)](https://github.com/oasisprotocol/oasis-core/pull/3361#discussion_r498919678).

Done.

kostko (approved):

Issue #3356, opened 2020-10-01
Seed node panics on oasis-node control queries


ptrus:
Seed node currently panics on most `oasis-node control` methods, since it doesn't override most of the methods from the base backend (https://github.com/oasisprotocol/oasis-core/blob/master/go/consensus/api/base.go). 

Some queries such as `GetStatus()` that do make sense for seed nodes should be implemented. For any others that don't make sense, make sure the node doesn't panic.

PR #3361, opened 2020-10-01
go/seed: implement missing consensus methods

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3356

Seed now supports `oasis-node control` methods and shouldn't panic on any cli requests.

TODO:
- [x] consider adding an e2e test for seed control cli


go/consensus/tendermint/seed/seed.go
```diff
@@ -83,6 +95,106 @@ func (srv *seedService) Cleanup() {
...
+	for _, tmpeer := range tmpeers {
+		p := string(tmpeer.ID()) + "@" + tmpeer.RemoteAddr().String()
+		peers = append(peers, p)
+	}
+	status.NodePeers = peers
+
+	return status, nil
+}
+
+// Unimplemented consensus Backend methods.
```
ptrus, line 125: The only missing methods are now the backends (epochtime, registry, beacon...) and `ConsensusKey` (which is only used in tests anyway). 

If we want, we could implement those remaining methods here (can still just panic on those), and get rid of the `consensus/api.BaseBackend` since it's not used anywhere else.

cc @kostko

go/consensus/tendermint/seed/seed.go
```diff
@@ -10,24 +11,35 @@ import (
...
 	genesis "github.com/oasisprotocol/oasis-core/go/genesis/api"
 	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/syncer"
 )
 
-var logger = logging.GetLogger("consensus/tendermint/seed")
+var (
+	logger = logging.GetLogger("consensus/tendermint/seed")
+
+	errNotSupported = fmt.Errorf("not supported by seed node")
```
kostko, line 35: Consider using `consensus.ErrUnsupported` instead.

go/consensus/tendermint/seed/seed.go
kostko: Agreed, implement the methods and remove the base backend.

kostko (commented):
Cool, some minor nits and yeah a test would be useful (and should be simple to do).

go/consensus/tendermint/common/common.go
```diff
@@ -32,11 +40,60 @@ const (
...
+	// CfgP2PSendRate is the rate at which packets can be sent, in bytes/second.
+	CfgP2PSendRate = "consensus.tendermint.p2p.send_rate"
+	// CfgP2PRecvRate is the rate at which packets can be received, in bytes/second.
+	CfgP2PRecvRate = "consensus.tendermint.p2p.recv_rate"
 )
 
 // Flags has the configuration flags.
 var Flags = flag.NewFlagSet("", flag.ContinueOnError)
 
+// GetAddress returns the configured tendermint address.
```
kostko, line 59: Maybe name something like `GetExternalAddress` to be more clear.

go/consensus/tendermint/seed/seed.go
```diff
@@ -83,6 +99,192 @@ func (srv *seedService) Cleanup() {
 	// No cleanup in particular.
 }
 
+// Synced returns a channel that is closed once synchronization is complete.
```
kostko, line 102:
```suggestion
// Implements Backend.
```

go/consensus/tendermint/seed/seed.go
```diff
@@ -83,6 +99,192 @@ func (srv *seedService) Cleanup() {
...
+func (srv *seedService) Synced() <-chan struct{} {
+	// Seed is always considered synced.
+	ch := make(chan struct{})
+	close(ch)
+	return ch
+}
+
+// SupportedFeatures returns the features supported by this consensus backend.
```
kostko, line 110:
```suggestion
// Implements Backend.
```

go/consensus/tendermint/seed/seed.go
```diff
@@ -83,6 +99,192 @@ func (srv *seedService) Cleanup() {
...
+		Backend:          api.BackendName,
+		Features:         srv.SupportedFeatures(),
+	}
+
+	// List of consensus peers.
+	tmpeers := srv.p2pSwitch.Peers().List()
+	peers := make([]string, 0, len(tmpeers))
+	for _, tmpeer := range tmpeers {
+		p := string(tmpeer.ID()) + "@" + tmpeer.RemoteAddr().String()
+		peers = append(peers, p)
```
kostko, line 128: No peers in E2E tests when querying seed node status (according to Codecov at least)?

go/consensus/tendermint/seed/seed.go
```diff
@@ -92,49 +294,56 @@ func New(dataDir string, identity *identity.Identity, genesisProvider genesis.Pr
...
+	srv.doc = doc
 
 	nodeInfo := p2p.DefaultNodeInfo{
 		ProtocolVersion: p2p.NewProtocolVersion(
-			version.P2PProtocol,
-			version.BlockProtocol,
-			0,
+			tmversion.P2PProtocol,
+			tmversion.BlockProtocol,
+			version.ConsensusProtocol.ToU64(),
```
kostko, line 329: Nice catch! @tjanez don't forget about this one in #3360 (depends on which PR gets merged first).

go/oasis-node/cmd/node/node.go
```diff
@@ -526,7 +525,8 @@ func NewTestNode() (*Node, error) {
 	return newNode(true)
 }
 
-func newNode(testNode bool) (n *Node, err error) { // nolint: gocyclo
+func newNode(testNode bool) (*Node, error) { // nolint: gocyclo
+	var err error
```
kostko, line 529: The reason for having that named `err` return value is for the deferred func below to propagate the error. This prevents that from happening. Maybe add a comment so this doesn't get changed again :-)

ptrus: Oh right, however in that case any objections to doing `(node *Node, err error)` ? (and not using a separate `n` variable for node here). Because otherwise having `n` in scope through the function which is always `nil` is a bit dangerous :grin: that's also what lead me to removing the named returns here. The only "downside" will be that a non-nil `node`  will be returned in case this function errors, but i think that's fine as?

go/consensus/tendermint/seed/seed.go
ptrus: Yeah i looked into this earlier today, so there's two things at work here:
- for inbound peers - seed instantly returns the peers and disconnects (as expected)
- for outbound - since we don't set the `"consensus.tendermint.core.external_address"` flag in the tests, nodes connecting to the seed provide `0.0.0.0:<port>` as their external address, the seed then doesn't insert the nodes to its address book and therefore also never makes outbound connections to those

Actually with current test setup seed will only ever return the peers from the genesis and never includes new nodes (but this is enough for all the tests). I also tested this by disabling the `populateAddrBookFromGenesis` in the seed and in that case none of the tests pass. However then configuring the test nodes to provide the correct external addresses (`127.0.0.1:<port>`), things work again correctly :-) - since this is working as expected i didn't commit the test here. However thinking about it more, it might make sense including such a test (disabling automatically including genesis peers in the seeds addressbook) in the suite so that we properly test seed including new peers in its address book.

ptrus: Fixed by configuring `external_addresses` in tests so seed does outbound requests now, also a check is done in the seed test.

go/oasis-node/cmd/node/node.go
kostko: So we did have it like that but changed it in d0551abb74b878bf62b44115fd917a9786c9357b due to a "possible nil dereference", but can't currently think of a case where this would be problematic.

ptrus: hm, so either whatever triggered that was fixed already (or it was a [false positive](https://github.com/dominikh/go-tools/issues/641)) since it doesn't seem to happen anymore

go/consensus/tendermint/seed/seed.go
ptrus: Also added support for configuring multiple seeds in the e2e tests, and a debug flag for disabling "populating address from genesis", so seed returning new peers is tested as well. Added an E2E test that tests both.

kostko (approved):

go/oasis-node/cmd/node/node.go
ptrus: ah i did found it, so `node` was `nil` in the deferred function. Fixed it by passing `node` into the deferred function via an argument

PR #3369, opened 2020-10-02
go/oasis-test-runner: Fix e2e/consensus-state-sync scenario (take two)

kostko:
Previous "fix" (see #3194) actually made the test silently skip actually performing state sync because it used the wrong validator index.

### DIFF
--- a/go/oasis-test-runner/scenario/e2e/consensus_state_sync.go
+++ b/go/oasis-test-runner/scenario/e2e/consensus_state_sync.go
@@ -9,6 +9,7 @@ import (
 	consensus "github.com/oasisprotocol/oasis-core/go/consensus/api"
 	control "github.com/oasisprotocol/oasis-core/go/control/api"
 	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/env"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/log"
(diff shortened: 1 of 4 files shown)


kostko:
Maybe we could make the test more robust by emitting log events.

go/oasis-test-runner/scenario/e2e/consensus_state_sync.go
```diff
@@ -86,7 +86,7 @@ func (sc *consensusStateSyncImpl) Run(childEnv *env.Env) error {
 
 	// Get the TLS public key from the validators.
 	var consensusNodes []string
-	for _, v := range sc.Net.Validators()[:2] {
+	for _, v := range sc.Net.Validators()[:3] {
```
ptrus, line 89: Instead of hardocing the `3` here, could change to "all but last" (and similarly bellow for the syncing validator could use "last"). That would also be in sync with the fixture where we append a validator, and would continue to work correctly in case the base fixture changes.

ptrus (approved):

go/oasis-test-runner/scenario/e2e/consensus_state_sync.go
kostko: Yeah, will do that and also add a log event assertion to make sure that state sync actually happened.

PR #3372, opened 2020-10-05
go/e2e/queries: correctly ignore UnexpectedEOF errors

ptrus:
Working fix from: #3302 , now with tests

### DIFF
--- a/go/oasis-node/cmd/debug/txsource/workload/queries.go
+++ b/go/oasis-node/cmd/debug/txsource/workload/queries.go
@@ -12,6 +12,7 @@ import (
 	"github.com/spf13/viper"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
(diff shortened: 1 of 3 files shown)


go/oasis-node/cmd/debug/txsource/workload/queries.go
```diff
@@ -239,9 +240,11 @@ func (q *queries) doConsensusQueries(ctx context.Context, rng *rand.Rand, height
 			"txs_with_results", txsWithRes,
 			"height", height,
 			"err", err,
+			"status", cmnGrpc.GetErrorStatus(err),
 		)
-		if status := cmnGrpc.GetErrorStatus(err); status != nil {
-			if status.Err() == io.ErrUnexpectedEOF {
+		if st := cmnGrpc.GetErrorStatus(err); st != nil {
+			s, _ := status.FromError(io.ErrUnexpectedEOF)
+			if st.Err().Error() == s.Err().Error() {
```
kostko, line 247: Ugh this seems awkward.

kostko (approved):

go/oasis-node/cmd/debug/txsource/workload/queries.go
ptrus: Yeah, unfortunately i didn't find a nicer way of doing this (the alternative would be directly matching on string - well which is kinda close to what this is doing, but at least here there's no hardcoded strings here :P)

PR #3373, opened 2020-10-05
Move storage backend initialization to storage worker

### LINKED ISSUE #3323: Client storage backend should be removed as a configurable backend
The `storage.backend` configuration option should only have an effect for storage nodes and should not be needed for client nodes (where it must currently be set to `client`). The storage backend init should also be moved under `go/worker/storage` and the configuration options under `worker.storage.*`.

Note that compute nodes already automatically use the correct storage backend and the client node should do the same automatically.

This would make it easier to run client nodes as there would be one less configuration parameter to misconfigure.

jberci:
Closes #3323


.changelog/3323.breaking.md
```diff
@@ -0,0 +1,4 @@
+go: Move storage backend initialization to storage worker
+
+Notably, this also means that node command line options for storage
+configuration have been renamed (`--storage.*` to `--worker.storage.*`).
```
ptrus, line 4: Add a `cfg` changelog fragment and list the changes please.

ptrus: Also there's an additional configuration change that should also be listed: client's do not need to configure storage backed anymore

go/runtime/registry/registry.go
```diff
@@ -284,21 +319,31 @@ func (r *runtimeRegistry) addSupportedRuntime(ctx context.Context, id common.Nam
 	var ns common.Namespace
 	copy(ns[:], id[:])
 
-	storageBackend, err := storage.New(ctx, path, ns, r.identity,
-		storage.WithConsensus(r.consensus),
-		storage.WithRuntime(rt),
-	)
+	//	impl, err = client.New(ctx, namespace, identity, o.consensus.Scheduler(), o.consensus.Registry(), o.runtime)
```
ptrus, line 322: Don't forget to remove the commented code.

go/runtime/registry/registry.go
```diff
@@ -284,21 +319,31 @@ func (r *runtimeRegistry) addSupportedRuntime(ctx context.Context, id common.Nam
...
+		case <-cancelCh:
+			return nil
+		case <-r.storageSetCh:
+			return rt.storage
+		}
+	}
 
 	// Create runtime tag indexer.
-	tagIndexer, err := tagindexer.New(path, cfg.TagIndexer, history, r.consensus.RootHash(), storageBackend)
+	tagIndexer, err := tagindexer.New(path, cfg.TagIndexer, history, r.consensus.RootHash(), storageWaiter)
```
kostko, line 346: Instead of using the storage waiter, could you just defer starting the tag indexer until storage is fully initialized and then pass the storage backend to the start method? I think this is the only use case for `WaitForBackend` and so it could be removed.

go/runtime/registry/registry.go
```diff
@@ -51,6 +51,10 @@ type Registry interface {
 
 	// Cleanup performs post-termination cleanup.
 	Cleanup()
+
+	// CommitStorage finalizes storage setup for all runtimes and triggers any
+	// delayed storageAPI.WaitForBackend getters.
+	CommitStorage(ctx context.Context) error
```
kostko, line 57: Maybe call this `FinishInitialization` instead?

.changelog/3323.cfg.2.md
```diff
@@ -0,0 +1,6 @@
+go: Change storage backend configuration options
```
kostko, line 1: Why `cfg.2`, there's only one cfg fragment?

go/runtime/registry/registry.go
```diff
@@ -145,11 +161,13 @@ func (r *runtime) stop() {
 	r.cancelCtx()
 	// Close local storage backend.
 	r.localStorage.Stop()
-	// Close storage backend.
-	r.storage.Cleanup()
-	// Close tag indexer service.
-	r.tagIndexer.Stop()
-	<-r.tagIndexer.Quit()
+	if !r.finished {
```
kostko, line 164: Does this require holding a mutex?

go/runtime/registry/registry.go
```diff
@@ -236,6 +256,34 @@ func (r *runtimeRegistry) Cleanup() {
...
+			}
+			rt.storage = storageBackend
+			rt.finished = true
+			defer func() {
+				if rerr != nil {
+					rt.tagIndexer.Stop()
+					storageBackend.Cleanup()
+				}
+			}()
+		}
```
kostko, line 281: Could this part be moved into the `runtime`, so you would call something like `rt.finishInitialization()` here?

go/runtime/registry/registry.go
```diff
@@ -236,6 +256,34 @@ func (r *runtimeRegistry) Cleanup() {
 	}
 }
 
+func (r *runtimeRegistry) FinishInitialization(ctx context.Context) (rerr error) {
+	r.RLock()
+	defer r.RUnlock()
+
+	for _, rt := range r.runtimes {
+		if rt.storage == nil {
```
kostko, line 264:
```suggestion
		if rt.storage != nil {
			continue
		}
```
And outdent the block below.

go/runtime/registry/registry.go
```diff
@@ -195,6 +213,8 @@ type runtimeRegistry struct {
 	identity  *identity.Identity
 
 	runtimes map[common.Namespace]*runtime
+
+	storageSetCh chan struct{}
```
kostko, line 217: This is more like "initialization completed channel"?
```suggestion
	initCh chan struct{}
```

go/runtime/registry/registry.go
kostko, line 264: Should this be taking `rt.RLock()`?

go/runtime/registry/registry.go
```diff
@@ -133,6 +146,9 @@ func (r *runtime) TagIndexer() tagindexer.QueryableBackend {
 }
 
 func (r *runtime) Storage() storageAPI.Backend {
+	if r.storage == nil {
```
kostko, line 149: Should this now be taking `r.RLock()`?

go/runtime/registry/registry.go
```diff
@@ -124,6 +133,10 @@ func (r *runtime) WatchRegistryDescriptor() (<-chan *registry.Runtime, pubsub.Cl
 	return ch, sub, nil
 }
 
+func (r *runtime) RegisterStorage(storage storageAPI.Backend) {
+	r.storage = storage
```
kostko, line 137: Should this now be taking `r.Lock()`?

go/runtime/tagindexer/tagindexer.go
```diff
@@ -121,14 +121,15 @@ func (s *Service) worker() {
...
-func (s *Service) Start() error {
+func (s *Service) Start(storage storage.Backend) error {
 	if _, ok := s.backend.(*nopBackend); ok {
 		// In case this is a nopBackend (which doesn't index anything) avoid the overhead of having
 		// a tag indexer service follow blocks and index them.
 		s.Logger.Info("not starting tag indexer as it is disabled")
 		return nil
 	}
 
+	s.storage = storage
```
kostko, line 132: Could also pass it as an argument to the goroutine.

go/runtime/registry/registry.go
kostko: Also should this panic if storage was already registered? To prevent accidental overwrites.

go/runtime/registry/registry.go
```diff
@@ -95,6 +102,8 @@ type runtime struct {
 	descriptorCh       chan struct{}
 	descriptorNotifier *pubsub.Broker
 
+	finished bool
```
kostko, line 105: Should this be called `initialized`?

kostko (approved):

PR #3377, opened 2020-10-06
go/e2e/workload-queries: don't query block heights that could get pruned

ptrus:


go/oasis-node/cmd/debug/txsource/workload/queries.go
```diff
@@ -705,7 +705,10 @@ func (q *queries) doQueries(ctx context.Context, rng *rand.Rand) error {
 
 	// Determine the earliest height that we can query.
 	earliestHeight := int64(1)
-	if numKept := viper.GetInt64(CfgConsensusNumKeptVersions); numKept < block.Height {
+	// When pruning is enabled add some buffer blocks to the earliest height,
```
kostko, line 708: We already reduce the actual pruning versions by half for this reason:
https://github.com/oasisprotocol/oasis-core/blob/6ae607363ed0b70fda9c7acfed8bf6d92ef73566/go/oasis-test-runner/scenario/e2e/runtime/txsource.go#L550-L551

I noticed that sometimes the control queries (`doControlQueries`) takes an unusually long time to complete so by the time the rest of the queries are executed, the height has already been pruned. We can do two things:
* Move control queries to the end as they are not height-dependent. (Also we should look into why the query can sometimes take so long -- usually they are fast, but sometimes they take a minute or so.)
* Increase `consensusPruneMinKept` in the txsource E2E test (currently it is set to 100 which seems quite low).

ptrus: yeah missed this, also doesn't seem that pruning was problematic for the test failure

PR #3391, opened 2020-10-08
failure indicating commitments

ptrus:
The E2E tests with increased epoch times (#3322) discovered a failure edge case that can happen if the storage committee becomes unavailable after proposer already proposes a batch, but before any of the nodes submit a commitment. In that case the nodes currently abort the processing and wait for a round failure. Since there is no submitted commitments the round failure only happens at the next epoch transition.

This PR adds support for failure indicating commitments (which were first proposed as part of the [ADR 0005](https://github.com/oasisprotocol/oasis-core/blob/cbf413b788e3fd0a0bc20b8193cc3570f99880b9/docs/adr/0005-runtime-compute-slashing.md)), which solve the above mentioned issue since compute nodes can now submit a failure indicating commitment and with it trigger a round failure. 

NOTE: only the subset (support for failure indicating commitments) of the mentioned ADR is implemented here.

TODO:
- [x] probably some more tests
- [x] check for other scenarios where a failure indicating commitment should be submitted by an executor node (currently only the mentioned failure scenario was fixed)

### DIFF
--- a/go/roothash/api/commitment/executor.go
+++ b/go/roothash/api/commitment/executor.go
@@ -4,6 +4,7 @@ package commitment
 import (
 	"bytes"
 	"errors"
+	"fmt"
 
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
@@ -32,11 +33,14 @@ var (
 //
 // Keep the roothash RAK validation in sync with changes to this structure.
 type ComputeResultsHeader struct {
-	Round        uint64           `json:"round"`
-	PreviousHash hash.Hash        `json:"previous_hash"`
-	IORoot       hash.Hash        `json:"io_root"`
-	StateRoot    hash.Hash        `json:"state_root"`
-	Messages     []*block.Message `json:"messages"`
+	Round        uint64    `json:"round"`
+	PreviousHash hash.Hash `json:"previous_hash"`
+
+	// Optional fields (may be absent for failure indication).
+
+	IORoot    *hash.Hash       `json:"io_root,omitempty"`
+	StateRoot *hash.Hash       `json:"state_root,omitempty"`
+	Messages  []*block.Message `json:"messages,omitempty"`
 }
 
 // IsParentOf returns true iff the header is the parent of a child header.
@@ -54,15 +58,42 @@ func (h *ComputeResultsHeader) EncodedHash() hash.Hash {
 	return hash.NewFrom(h)
 }
 
+// ExecutorCommitmentFailure is the executor commitment failure reason.
+type ExecutorCommitmentFailure uint8
+
+const (
+	// FailureNone indicates that no failure has occurred.
+	FailureNone ExecutorCommitmentFailure = 0
+	// FailureUnknown indicates a generic failure.
+	FailureUnknown ExecutorCommitmentFailure = 1
+	// FailureStorageUnavailable indicates that batch processing failed due to
+	// storage being unavailable.
+	FailureStorageUnavailable ExecutorCommitmentFailure = 2
+)
+
 // ComputeBody holds the data signed in a compute worker commitment.
 type ComputeBody struct {
-	Header            ComputeResultsHeader   `json:"header"`
-	StorageSignatures []signature.Signature  `json:"storage_signatures"`
-	RakSig            signature.RawSignature `json:"rak_sig"`
+	Header  ComputeResultsHeader      `json:"header"`
+	Failure ExecutorCommitmentFailure `json:"failure,omitempty"`
 
 	TxnSchedSig      signature.Signature   `json:"txn_sched_sig"`
 	InputRoot        hash.Hash             `json:"input_root"`
 	InputStorageSigs []signature.Signature `json:"input_storage_sigs"`
+
+	// Optional fields (may be absent for failure indication).
+
+	StorageSignatures []signature.Signature   `json:"storage_signatures,omitempty"`
+	RakSig            *signature.RawSignature `json:"rak_sig,omitempty"`
+}
+
+// SetFailure sets failure reason and clears any fields that should be clear
+// in a failure indicating commitment.
+func (m *ComputeBody) SetFailure(failure ExecutorCommitmentFailure) {
+	m.Header.IORoot = nil
+	m.Header.StateRoot = nil
+	m.StorageSignatures = nil
+	m.RakSig = nil
+	m.Failure = failure
 }
 
 // VerifyTxnSchedSignature rebuilds the batch dispatch message from the data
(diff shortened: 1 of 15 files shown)


go/roothash/api/commitment/executor.go
```diff
@@ -54,15 +54,32 @@ func (h *ComputeResultsHeader) EncodedHash() hash.Hash {
...
+	Failure ExecutorCommitmentFailure `json:"failure,omitempty"`
 
 	TxnSchedSig      signature.Signature   `json:"txn_sched_sig"`
 	InputRoot        hash.Hash             `json:"input_root"`
 	InputStorageSigs []signature.Signature `json:"input_storage_sigs"`
+
+	// Optional fields (may be absent for failure indication).
+
+	StorageSignatures []signature.Signature  `json:"storage_signatures,omitempty"`
+	RakSig            signature.RawSignature `json:"rak_sig,omitempty"`
```
kostko, line 82: So the ADR specifies that some of these fields (and in `ComputeResultsHeader`) should be made optional by making them pointers so they can be `nil`. This will require some more checks (and corresponding tests) to make sure they either are or are not nil (depending on the failure field), but it should result in smaller failure messages.

go/roothash/api/commitment/pool.go
```diff
@@ -483,6 +514,9 @@ func (p *Pool) ResolveDiscrepancy() (OpenCommitment, error) {
 	}
 
 	minVotes := (backupNodes / 2) + 1
+	if failuresTally >= minVotes {
+		return nil, ErrInsufficientVotes // TODO: new error?
```
kostko, line 518: Yeah it would be good to have a new error code, not sure how it should be called.

go/roothash/api/commitment/executor.go
ptrus: Yeah that make sense, will update to match the ADR :+1:

go/roothash/api/commitment/pool.go
```diff
@@ -267,30 +239,111 @@ func (p *Pool) addOpenExecutorCommitment(
...
+	switch body.Failure {
+	case FailureNone:
+		// Ensure header fields are present.
+		if header.IORoot == nil {
+			logger.Debug("executor commitment is missing IORoot")
+			return ErrBadRootHashes
+		}
 
+		// Ensure header fields are present.
+		if header.StateRoot == nil {
```
kostko, line 251: Maybe these checks should be moved into a `ValidateBasic() error` method of the `ComputeBody`? It would check that the message is internally consistent (e.g. if it indicates failure the fields are nil and otherwise they are non-nil). This would make it easier to check validity in various places if needed.

go/worker/compute/executor/committee/node.go
```diff
@@ -1167,9 +1167,14 @@ func (n *Node) proposeBatchLocked(processedBatch *processedBatch) {
 	}()
 	storageCommitLatency.With(n.getMetricLabels()).Observe(time.Since(start).Seconds())
 
-	if err != nil {
-		n.abortBatchLocked(err)
-		return
+	if storageErr != nil {
+		n.logger.Error("storage failure, sibmitting failure indicating commitment",
```
kostko, line 1171:
```suggestion
		n.logger.Error("storage failure, submitting failure indicating commitment",
```

go/worker/compute/executor/committee/node.go
```diff
@@ -1167,9 +1167,14 @@ func (n *Node) proposeBatchLocked(processedBatch *processedBatch) {
...
+			"err", storageErr,
+		)
+		proposedResults.RakSig = nil
+		proposedResults.Header.IORoot = nil
+		proposedResults.Header.StateRoot = nil
+		proposedResults.Failure = commitment.FailureStorageUnavailable
```
kostko, line 1177: Could also make a method like `proposedResults.SetFailure(commitment.FailureStorageUnavailable)` which would take care of nilifying the required fields?

kostko (commented):
We should also probably submit a failure in case the runtime fails to process a batch, e.g. here:
https://github.com/oasisprotocol/oasis-core/blob/bb1aa483f9b85bdfaebf90cce3b7b7c1a2256211/go/worker/compute/executor/committee/node.go#L1447

go/worker/compute/executor/committee/node.go
ptrus: good idea, done!

go/roothash/api/commitment/pool.go
ptrus: done

go/roothash/api/commitment/pool.go
ptrus: Used `ErrMajorityFailure` for now

ptrus:
> We should also probably submit a failure in case the runtime fails to process a batch, e.g. here:

Added, not sure how to best test these cases maybe with byzantine-storage tests, need to think about it a bit, but probably can be done in a separate PR.

kostko:
Yeah we should add Byzantine tests for all these cases, can you open an issue for it?

go/roothash/api/commitment/executor.go
```diff
@@ -81,10 +112,54 @@ func (m *ComputeBody) VerifyTxnSchedSignature(header block.Header) bool {
...
+	}
 	return []hash.Hash{
-		m.Header.IORoot,
-		m.Header.StateRoot,
+		ioRoot,
+		stateRoot,
+	}
+}
+
+// ValidateBasic performs basic compute body commitment parameters checks.
```
kostko, line 129:
```suggestion
// ValidateBasic performs basic executor commitment validity checks.
```

go/roothash/api/commitment/pool.go
```diff
@@ -267,28 +239,74 @@ func (p *Pool) addOpenExecutorCommitment(
...
-			"node_id", id,
-			"min_write_replication", p.Runtime.Storage.MinWriteReplication,
-			"num_receipts", len(body.StorageSignatures),
-		)
-		return ErrBadStorageReceipts
-	}
-	if err := sv.VerifyCommitteeSignatures(scheduler.KindStorage, body.StorageSignatures); err != nil {
-		logger.Debug("executor commitment has bad storage receipt signers",
-			"node_id", id,
+	// Validate basic.
```
kostko, line 242: Basic checks should be moved before signature verification (e.g., where RAK checks were before).

go/worker/compute/executor/committee/node.go
```diff
@@ -1167,22 +1172,50 @@ func (n *Node) proposeBatchLocked(processedBatch *processedBatch) {
...
+			proposedIORoot: *proposedResults.Header.IORoot,
+		})
+	default:
+		n.abortBatchLocked(storageErr)
+	}
+
+	crash.Here(crashPointBatchProposeAfter)
+}
+
+func (n *Node) singAndSubmitCommitment(body *commitment.ComputeBody) error {
```
kostko, line 1208:
```suggestion
func (n *Node) signAndSubmitCommitment(body *commitment.ComputeBody) error {
```

go/worker/compute/executor/committee/node.go
```diff
@@ -1291,6 +1318,90 @@ func (n *Node) handleExternalBatchLocked(batch *unresolvedBatch, hdr block.Heade
 	return nil
 }
 
+func (n *Node) handleHostRtEvent(ev *host.Event) {
```
kostko, line 1321:
```suggestion
func (n *Node) handleRuntimeHostEvent(ev *host.Event) {
```

go/worker/compute/executor/committee/node.go
```diff
@@ -1291,6 +1318,90 @@ func (n *Node) handleExternalBatchLocked(batch *unresolvedBatch, hdr block.Heade
...
+			n.logger.Debug("submitting failure indicating commitment")
+
+			// Submit a failure indicating commitment.
+			header := n.commonNode.CurrentBlock.Header
+			commit := &commitment.ComputeBody{
+				Header: commitment.ComputeResultsHeader{
+					Round:        header.Round + 1,
+					PreviousHash: header.EncodedHash(),
+				},
+				Failure:          commitment.FailureUnknown,
```
kostko, line 1374: Should this use `commit.SetFailure` instead?

go/worker/compute/executor/committee/node.go
```diff
@@ -1291,6 +1318,90 @@ func (n *Node) handleExternalBatchLocked(batch *unresolvedBatch, hdr block.Heade
...
+		n.roleProvider.SetUnavailable()
+	default:
+		// Unknown event.
+		n.logger.Warn("unknown worker event",
+			"ev", ev,
+		)
+	}
+}
+
+func (n *Node) handleProcessedBatch(batch *processedBatch) {
```
kostko, line 1352: Could you refactor this function a bit? Two things:
* The state staleness checks are the same between the two cases (failed batch/successful batch) which is currently handled in two inline functions, so no need to duplicate them.
* These inline functions are not needed here (they were needed before as this was done inside a case, but now you already have a function), you can move locking/defer one level up.

go/worker/compute/executor/committee/node.go
```diff
@@ -1384,18 +1495,16 @@ func (n *Node) worker() {
 	// We are initialized.
 	close(n.initCh)
 
-	var runtimeVersion version.Version
 	for {
 		// Check if we are currently processing a batch. In this case, we also
 		// need to select over the result channel.
-		var processingDoneCh chan *processedBatch
-
+		n.processingDoneCh = nil
```
kostko, line 1501: Maybe it would be better to keep the variable here and just pass it as an argument to `n.handleProcessedBatch` to avoid anyone using it elsewhere.

ptrus:
> Yeah we should add Byzantine tests for all these cases, can you open an issue for it?

https://github.com/oasisprotocol/oasis-core/issues/3414

go/worker/compute/executor/committee/node.go
```diff
@@ -1167,22 +1170,48 @@ func (n *Node) proposeBatchLocked(processedBatch *processedBatch) {
...
+			"commit", proposedResults,
+			"err", err,
+		)
 		n.abortBatchLocked(err)
 		return
 	}
 
-	// Sign the commitment and submit.
-	commit, err := commitment.SignExecutorCommitment(n.commonNode.Identity.NodeSigner, proposedResults)
+	// TODO: Add crash point.
```
kostko, line 1189: I don't think we need one as there is one below and nothing interesting happens inbetween.

go/worker/compute/executor/committee/node.go
```diff
@@ -1291,6 +1314,78 @@ func (n *Node) handleExternalBatchLocked(batch *unresolvedBatch, hdr block.Heade
...
+	n.commonNode.CrossNode.Lock()
+	defer n.commonNode.CrossNode.Unlock()
+
+	// To avoid stale events, check if the stored state is still valid.
+	state, ok := n.state.(StateProcessingBatch)
+	if !ok || state.done != processingCh {
+		return
+	}
+
+	if batch == nil || batch.computed == nil {
```
kostko, line 1358: Since this body is longer, maybe invert the check so less stuff is indented?

kostko (approved):

PR #3415, opened 2020-10-15
go/storage: Add type information to storage roots

### LINKED ISSUE #2053: Storage roots should be typed
Currently storage roots are arbitrary Merkle key-value trees without additional semantics. We use storage to store two distinct kinds of roots:
* State roots which contain arbitrary runtime-specific state.
* I/O roots which contain transaction inputs, outputs and emitted tags and have clearly specified semantics (e.g., the format of keys and values).

It would be useful to add a type to storage roots in order to:
* Prevent (accidental or malicious) mixing of roots by making the type be part of all node hash computations. The type should also be part of storage receipts so that everyone can easily verify which kind of root was committed.
* Enable type-specific optimizations and validation (e.g., storage nodes can check that the I/O root actually conforms to I/O root semantics and can provide additional methods to query such roots).

jberci:
Closes #2053

### DIFF
--- a/client/src/transaction/snapshot.rs
+++ b/client/src/transaction/snapshot.rs
@@ -8,7 +8,7 @@ use oasis_core_runtime::{
     common::{cbor, crypto::hash::Hash, namespace::Namespace},
     consensus::roothash::Block,
     storage::{
-        mkvs::{sync::*, Iterator, Prefix, Root, Tree, WriteLog},
+        mkvs::{sync::*, Iterator, Prefix, Root, RootType, Tree, WriteLog},
         MKVS,
     },
     transaction::types::{TxnCall, TxnOutput},
@@ -66,6 +66,7 @@ impl Clone for BlockSnapshot {
             .with_root(Root {
                 namespace: self.block.header.namespace,
                 version: self.block.header.round,
+                root_type: RootType::State,
                 hash: self.block.header.state_root,
             })
             .new(Box::new(read_syncer.clone()));
@@ -86,6 +87,7 @@ impl BlockSnapshot {
             .with_root(Root {
                 namespace: block.header.namespace,
                 version: block.header.round,
+                root_type: RootType::State,
                 hash: block.header.state_root,
             })
             .new(Box::new(read_syncer.clone()));
--- a/go/consensus/tendermint/abci/mux.go
+++ b/go/consensus/tendermint/abci/mux.go
@@ -30,6 +30,7 @@ import (
 	consensusGenesis "github.com/oasisprotocol/oasis-core/go/consensus/genesis"
 	abciState "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/abci/state"
 	"github.com/oasisprotocol/oasis-core/go/consensus/tendermint/api"
+	storageApi "github.com/oasisprotocol/oasis-core/go/storage/api"
 	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/checkpoint"
 	upgrade "github.com/oasisprotocol/oasis-core/go/upgrade/api"
 )
@@ -965,7 +966,7 @@ func (mux *abciMux) ApplySnapshotChunk(req types.RequestApplySnapshotChunk) type
 
 	// Check if we are done with the restoration. In this case, finalize the root.
 	if done {
-		err = mux.state.storage.NodeDB().Finalize(mux.state.ctx, cp.Root.Version, []hash.Hash{cp.Root.Hash})
+		err = mux.state.storage.NodeDB().Finalize(mux.state.ctx, []storageApi.Root{cp.Root})
 		if err != nil {
 			mux.logger.Error("failed to finalize restored root",
 				"err", err,
--- a/go/consensus/tendermint/abci/prune_test.go
+++ b/go/consensus/tendermint/abci/prune_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/storage/mkvs"
 	mkvsDB "github.com/oasisprotocol/oasis-core/go/storage/mkvs/db/api"
 	mkvsBadgerDB "github.com/oasisprotocol/oasis-core/go/storage/mkvs/db/badger"
+	mkvsNode "github.com/oasisprotocol/oasis-core/go/storage/mkvs/node"
 )
 
 func TestPruneKeepN(t *testing.T) {
@@ -31,7 +32,7 @@ func TestPruneKeepN(t *testing.T) {
 		MaxCacheSize: 16 * 1024 * 1024,
 	})
 	require.NoError(err, "New")
-	tree := mkvs.New(nil, ndb)
+	tree := mkvs.New(nil, ndb, mkvsNode.RootTypeState)
 
 	ctx := context.Background()
 	for i := uint64(1); i <= 11; i++ {
@@ -41,7 +42,7 @@ func TestPruneKeepN(t *testing.T) {
 		var rootHash hash.Hash
 		_, rootHash, err = tree.Commit(ctx, common.Namespace{}, i)
 		require.NoError(err, "Commit")
-		err = ndb.Finalize(ctx, i, []hash.Hash{rootHash})
+		err = ndb.Finalize(ctx, []mkvsNode.Root{{Namespace: common.Namespace{}, Version: i, Type: mkvsNode.RootTypeState, Hash: rootHash}})
 		require.NoError(err, "Finalize")
 	}
 
--- a/go/oasis-node/cmd/debug/storage/benchmark.go
+++ b/go/oasis-node/cmd/debug/storage/benchmark.go
@@ -159,6 +159,7 @@ func doBenchmark(cmd *cobra.Command, args []string) { // nolint: gocyclo
 					b.Fatalf("failed to Open(): %v", err)
 				}
 				newRoot.Hash = receiptBody.Roots[0]
+				newRoot.Type = receiptBody.RootTypes[0]
 				b.StartTimer()
 			}
 		})
--- a/go/oasis-node/cmd/debug/storage/benchmark.go
+++ b/go/oasis-node/cmd/debug/storage/benchmark.go
@@ -21,7 +21,9 @@ import (
 	memorySigner "github.com/oasisprotocol/oasis-core/go/common/crypto/signature/signers/memory"
 	"github.com/oasisprotocol/oasis-core/go/common/identity"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
+	genesisTests "github.com/oasisprotocol/oasis-core/go/genesis/tests"
 	cmdCommon "github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common"
+	commonFlags "github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common/flags"
 	storageAPI "github.com/oasisprotocol/oasis-core/go/storage/api"
 	"github.com/oasisprotocol/oasis-core/go/worker/storage"
 )
@@ -69,6 +71,7 @@ func doBenchmark(cmd *cobra.Command, args []string) { // nolint: gocyclo
 	}
 
 	// Create an identity.
+	genesisTests.SetTestChainContext()
 	ident, err := identity.LoadOrGenerate(dataDir, memorySigner.NewFactory(), false)
 	if err != nil {
 		logger.Error("failed to generate a new identity",
@@ -81,6 +84,10 @@ func doBenchmark(cmd *cobra.Command, args []string) { // nolint: gocyclo
 	viper.Set("storage.debug.insecure_skip_checks", true)
 
 	var ns common.Namespace
+
+	// Set some extra options to make the database cooperate with the specifics here.
+	viper.SetDefault(storage.CfgInsecureSkipChecks, true)
+	viper.SetDefault(commonFlags.CfgDebugDontBlameOasis, true)
 
 	storage, err := storage.NewLocalBackend(dataDir, ns, ident)
 	if err != nil {
--- a/go/roothash/api/block/header.go
+++ b/go/roothash/api/block/header.go
@@ -102,18 +102,21 @@ func (h *Header) EncodedHash() hash.Hash {
 }
 
 // StorageRoots returns the storage roots contained in this header.
-func (h *Header) StorageRoots() (roots []storage.Root) {
-	for _, rootHash := range []hash.Hash{
-		h.IORoot,
-		h.StateRoot,
-	} {
-		roots = append(roots, storage.Root{
+func (h *Header) StorageRoots() []storage.Root {
+	return []storage.Root{
+		{
 			Namespace: h.Namespace,
 			Version:   h.Round,
-			Hash:      rootHash,
-		})
+			Type:      storage.RootTypeIO,
+			Hash:      h.IORoot,
+		},
+		{
+			Namespace: h.Namespace,
+			Version:   h.Round,
+			Type:      storage.RootTypeState,
+			Hash:      h.StateRoot,
+		},
 	}
-	return
 }
 
 // RootsForStorageReceipt gets the merkle roots that must be part of
@@ -122,6 +125,16 @@ func (h *Header) RootsForStorageReceipt() []hash.Hash {
 	return []hash.Hash{
 		h.IORoot,
 		h.StateRoot,
+	}
+}
+
+// RootTypesForStorageReceipt gets the storage root type sequence for the roots
+// returned by RootsForStorageReceipt.
+func (h *Header) RootTypesForStorageReceipt() []storage.RootType {
+	// NOTE: Keep these in the same order as in RootsForStorageReceipt above!
+	return []storage.RootType{
+		storage.RootTypeIO,
+		storage.RootTypeState,
 	}
 }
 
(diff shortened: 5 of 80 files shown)


client/src/transaction/snapshot.rs
```diff
@@ -89,9 +90,10 @@ impl BlockSnapshot {
             .with_root(Root {
                 namespace: block.header.namespace,
                 version: block.header.round,
+                root_type: RootType::State as u8,
                 hash: block.header.state_root,
             })
-            .new(Box::new(read_syncer.clone()));
+            .new(Box::new(read_syncer.clone()), RootType::State as u8);
```
kostko, line 96: Seems redundant that this needs to be specified twice?

go/consensus/tendermint/abci/mux.go
```diff
@@ -995,7 +996,7 @@ func (mux *abciMux) ApplySnapshotChunk(req types.RequestApplySnapshotChunk) type
 
 	// Check if we are done with the restoration. In this case, finalize the root.
 	if done {
-		err = mux.state.storage.NodeDB().Finalize(mux.state.ctx, cp.Root.Version, []hash.Hash{cp.Root.Hash})
+		err = mux.state.storage.NodeDB().Finalize(mux.state.ctx, cp.Root.Version, []storageApi.Root{cp.Root})
```
kostko, line 999: In case you specify roots, why would you need to also specify the version as a separate argument? Finalize should just make sure that all the passed roots use exactly the same version.

go/consensus/tendermint/abci/prune_test.go
```diff
@@ -31,7 +32,7 @@ func TestPruneKeepN(t *testing.T) {
 		MaxCacheSize: 16 * 1024 * 1024,
 	})
 	require.NoError(err, "New")
-	tree := mkvs.New(nil, ndb)
+	tree := mkvs.New(nil, ndb, 0)
```
kostko, line 35: Shouldn't this use the provided constants? Also in other places.

go/storage/mkvs/overlay.go
```diff
@@ -24,10 +24,10 @@ type treeOverlay struct {
 // as the inner tree has its own cache and double caching makes less sense.
 //
 // The overlay is not safe for concurrent use.
-func NewOverlay(inner Tree) OverlayTree {
+func NewOverlay(inner Tree, rootType node.RootType) OverlayTree {
```
kostko, line 27: Shouldn't this get the type from `inner`?

go/worker/storage/committee/node.go
```diff
@@ -295,15 +297,25 @@ func NewNode(
...
 				if berr != nil {
 					return nil, berr
 				}
 
-				return []hash.Hash{
-					blk.Header.IORoot,
-					blk.Header.StateRoot,
+				return []storageApi.Root{
+					{
+						Namespace: commonNode.Runtime.ID(),
```
kostko, line 308: Why not `blk.Header.Namespace` to be consistent with below? And then you could add a common method to get the storage roots from the header -- actually there already is `blk.Header.StorageRoots()`, why don't you just use that?

go/worker/storage/committee/node.go
```diff
@@ -471,10 +483,19 @@ func (n *Node) ForceFinalize(ctx context.Context, round uint64) error {
...
+		Version:   round,
+		Type:      mkvsNode.RootTypeIO,
+		Hash:      block.Header.IORoot,
+	}
+	stateRoot := mkvsNode.Root{
+		Namespace: block.Header.Namespace,
+		Version:   round,
+		Type:      mkvsNode.RootTypeState,
+		Hash:      block.Header.StateRoot,
+	}
```
kostko, line 497: Could use `block.Header.StorageRoots()`?

runtime/src/storage/mkvs/tree/tree.rs
```diff
@@ -44,27 +44,30 @@ impl Options {
...
     pub(crate) cache: RefCell<Box<LRUCache>>,
+    pub(crate) root_type: u8,
     pub(crate) pending_write_log: BTreeMap<Key, PendingLogEntry>,
     pub(crate) lock: Arc<Mutex<isize>>,
 }
 
 impl Tree {
     /// Construct a new tree instance using the given read syncer and options struct.
-    pub fn new(read_syncer: Box<dyn ReadSync>, opts: &Options) -> Tree {
+    pub fn new(read_syncer: Box<dyn ReadSync>, root_type: u8, opts: &Options) -> Tree {
```
kostko, line 62: Should this make sure that in case a root is passed via options, the type matches?

go/roothash/api/block/header.go
```diff
@@ -103,16 +103,18 @@ func (h *Header) EncodedHash() hash.Hash {
...
+		Version:   h.Round,
+		Type:      storage.RootTypeIO,
+		Hash:      h.IORoot,
+	})
+	roots = append(roots, storage.Root{
+		Namespace: h.Namespace,
+		Version:   h.Round,
+		Type:      storage.RootTypeState,
+		Hash:      h.StateRoot,
+	})
```
kostko, line 117:
```suggestion
    return []storage.Root{
        {
            Namespace: h.Namespace,
            Version:   h.Round,
            Type:      storage.RootTypeIO,
            Hash:      h.IORoot,
        },
        {
            Namespace: h.Namespace,
            Version:   h.Round,
            Type:      storage.RootTypeState,
            Hash:      h.StateRoot,
        }
    }
```

go/storage/mkvs/db/badger/hash.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// Hash returns the hash portion of the typed hash.
+func (h *TypedHash) Hash() (rh hash.Hash) {
+	copy(rh[:], h[1:])
+	return
+}
+
+// TypedHashFromParts creates a new typed hash with the parts given.
+func TypedHashFromParts(typ node.RootType, hash hash.Hash) (h TypedHash) {
```
kostko, line 98: This probably shouldn't be exported.

go/storage/mkvs/db/badger/hash.go
```diff
@@ -0,0 +1,109 @@
...
+	h[0] = byte(typ)
+	copy(h[1:], hash[:])
+	return
+}
+
+// TypedHashFromRoot creates a new typed hash corresponding to the given storage root.
+func TypedHashFromRoot(root node.Root) (h TypedHash) {
```
kostko, line 105: This probably shouldn't be exported.

go/storage/mkvs/db/badger/hash.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/node"
+)
+
+var (
+	_ encoding.BinaryMarshaler   = (*TypedHash)(nil)
+	_ encoding.BinaryUnmarshaler = (*TypedHash)(nil)
+)
+
+// Size is the size of the typed hash in bytes.
+const Size = hash.Size + 1
```
kostko, line 20:
```suggestion
const typedHashSize = hash.Size + 1
```

go/storage/mkvs/node/node.go
```diff
@@ -51,19 +51,43 @@ var (
...
+
+// String returns the string representation of the storage root type.
+func (r RootType) String() string {
+	switch r {
+	case RootTypeIO:
+		return "io-root"
+	case RootTypeState:
+		return "state-root"
+	default:
+		return fmt.Sprintf("unknown-root-type-%d", r)
```
kostko, line 72:
```suggestion
		return fmt.Sprintf("[unknown root type: %d]", r)
```

kostko (commented):
Type should also be included in storage receipts (the `storage.ReceiptBody` structure).

go/worker/storage/committee/node.go
```diff
@@ -295,15 +297,25 @@ func NewNode(
...
+			GetRoots: func(ctx context.Context, version uint64) ([]storageApi.Root, error) {
 				blk, berr := commonNode.Runtime.History().GetBlock(ctx, version)
 				if berr != nil {
 					return nil, berr
 				}
 
-				return []hash.Hash{
-					blk.Header.IORoot,
-					blk.Header.StateRoot,
+				return []storageApi.Root{
```
kostko, line 306: Why not `return blk.Header.StorageRoots()`?

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -446,10 +445,10 @@ func (n *Node) syncCheckpoints() (*blockSummary, error) {
 				syncState.StateRoot = check.Root
 			}
```
kostko, line 447: Update `checkCheckpointUsable` to verify root types. Also check if there are any other places where types should be checked.

runtime/src/storage/mkvs/interop/rpc.rs
```diff
@@ -69,6 +69,7 @@ const CALL_TIMEOUT: Duration = Duration::from_secs(30);
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct ApplyRequest {
     pub namespace: Namespace,
+    pub root_type: u8,
```
kostko, line 72: Why doesn't this use `RootType`?

kostko (approved):

go/storage/mkvs/db/badger/migrate.go
```diff
@@ -0,0 +1,409 @@
...
+type v3UpdatedNode struct {
+	_ struct{} `cbor:",toarray"` // nolint
+
+	Removed bool
+	Hash    hash.Hash
+}
+
+type v4UpdatedNode = updatedNode
+
+type MigrationHelper interface {
```
kostko, line 63:
```suggestion
type V4MigrationHelper interface {
```

go/storage/mkvs/db/badger/migrate.go
```diff
@@ -0,0 +1,409 @@
...
+
+	opts := badger.DefaultOptions(cfg.DB)
+	opts = opts.WithLogger(cmnBadger.NewLogAdapter(db.logger))
+	opts = opts.WithSyncWrites(!cfg.NoFsync)
+	// Allow value log truncation if required (this is needed to recover the
+	// value log file which can get corrupted in crashes).
+	opts = opts.WithTruncate(true)
+	opts = opts.WithCompression(options.Snappy)
+	opts = opts.WithBlockCacheSize(cfg.MaxCacheSize)
+	opts = opts.WithDetectConflicts(false)
```
kostko, line 354: Should this part be moved to some common function? E.g. `configToBadgerOptions(cfg)` or similar?

kostko (commented):
Will the migration tool properly handle interruptions while migrating?

go/storage/mkvs/db/badger/check.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/dgraph-io/badger/v2"
+
+	"github.com/oasisprotocol/oasis-core/go/common/cbor"
+	"github.com/oasisprotocol/oasis-core/go/common/logging"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/db/api"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/node"
+)
+
+func CheckSanity(ctx context.Context, cfg *api.Config) error {
```
kostko, line 15: Add documentation comment.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -271,6 +289,12 @@ func (d *badgerNodeDB) GetNode(root node.Root, ptr *node.Pointer) (node.Node, er
 
 	tx := d.db.NewTransactionAt(versionToTs(root.Version), false)
 	defer tx.Discard()
+
+	// Check if the root actually exists.
+	if err := d.checkRoot(tx, root); err != nil {
```
kostko, line 294: Can you just quickly check what's the performance impact (since GetNode is in the hot path)? E.g. is it fine if we rely on Badger's memtables containing the root or would we benefit from a small LRU cache for roots?

go/storage/mkvs/db/badger/badger.go
```diff
@@ -1075,14 +1130,27 @@ func (s *badgerSubtree) PutNode(depth node.Depth, ptr *node.Pointer) error {
 	nodeKey := nodeKeyFmt.Encode(&h)
 	if s.batch.multipartNodes != nil {
 		if _, err = s.batch.readTxn.Get(nodeKey); err != nil && errors.Is(err, badger.ErrKeyNotFound) {
-			if err = s.batch.multipartNodes.Set(multipartRestoreNodeLogKeyFmt.Encode(&h), []byte{}); err != nil {
+			th := typedHashFromParts(node.RootTypeInvalid, h)
+			if err = s.batch.multipartNodes.Set(multipartRestoreNodeLogKeyFmt.Encode(&th), []byte{}); err != nil {
 				return err
 			}
+			if depth == 0 {
```
kostko, line 1137: Add some comments that this is special handling for the root node, also below.

go/storage/mkvs/db/badger/migrate.go
```diff
@@ -0,0 +1,693 @@
...
+	// All done, flush and clean up. The metadata blob will be removed
+	// in the defer handler.
+	v4.done = true
+	if err := v4.flush(true); err != nil {
+		return 0, err
+	}
+	return 4, nil
+}
+
+func Migrate(cfg *api.Config, helper MigrationHelper) (uint64, error) {
```
kostko, line 623: Add documentation comment.

go/storage/mkvs/db/badger/migrate.go
```diff
@@ -0,0 +1,693 @@
...
+
+	// Make sure that we can discard any deleted/invalid metadata.
+	db.db.SetDiscardTs(tsMetadata)
+
+	// Load metadata.
+	lastVersion, err := func() (uint64, error) {
+		tx := db.db.NewTransactionAt(tsMetadata, false)
+		defer tx.Discard()
+
+		var migMeta migrationCommonMeta
```
kostko, line 645: Add comments to describe what's going on here -- e.g., this is handling the case if migration was interrupted, right?

go/storage/mkvs/db/badger/migrate.go
```diff
@@ -0,0 +1,693 @@
...
+		item, rerr := tx.Get(migrationMetaKeyFmt.Encode())
+		if rerr == nil {
+			rerr = item.Value(func(data []byte) error {
+				return cbor.UnmarshalTrusted(data, &migMeta)
+			})
+			if rerr != nil {
+				return 0, rerr
```
kostko, line 652: Wrap error using `fmt.Errorf` to provide more context of what failed. Also elsewhere.

go/storage/mkvs/db/badger/badger.go
jberci: There doesn't seem to be an appreciable difference according to the storage benchmark debug command in the node.

go/oasis-node/cmd/debug/storage/benchmark.go
```diff
@@ -69,6 +70,7 @@ func doBenchmark(cmd *cobra.Command, args []string) { // nolint: gocyclo
 	}
 
 	// Create an identity.
+	signature.SetChainContext("this is a drill!")
```
kostko, line 73: You could use `SetTestChainContext` from `go/genesis/tests`.

go/oasis-node/cmd/debug/storage/benchmark.go
```diff
@@ -82,6 +84,10 @@ func doBenchmark(cmd *cobra.Command, args []string) { // nolint: gocyclo
 
 	var ns common.Namespace
 
+	// Set some extra options to make the database cooperate with the specifics here.
+	viper.SetDefault("worker.storage.debug.insecure_skip_checks", true)
+	viper.SetDefault("debug.dont_blame_oasis", true)
```
kostko, line 89: Import flag name constants instead of duplicating them here.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -486,24 +517,26 @@ func (d *badgerNodeDB) HasRoot(root node.Root) bool {
...
 }
 
-func (d *badgerNodeDB) Finalize(ctx context.Context, version uint64, roots []hash.Hash) error { // nolint: gocyclo
+func (d *badgerNodeDB) Finalize(ctx context.Context, roots []node.Root) error { // nolint: gocyclo
 	if d.readOnly {
 		return api.ErrReadOnly
 	}
 
+	if len(roots) == 0 {
+		return fmt.Errorf("mkvs/badger: need at least one root to finalize")
```
kostko, line 536: Add test for this.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -530,9 +563,12 @@ func (d *badgerNodeDB) Finalize(ctx context.Context, version uint64, roots []has
 
 	// Determine a set of finalized roots. Finalization is transitive, so if
 	// a parent root is finalized the child should be consider finalized too.
-	finalizedRoots := make(map[hash.Hash]bool)
-	for _, rootHash := range roots {
-		finalizedRoots[rootHash] = true
+	finalizedRoots := make(map[typedHash]bool)
+	for _, root := range roots {
+		if root.Version != version {
+			return fmt.Errorf("mkvs/badger: roots to finalize don't have matching versions")
```
kostko, line 569: Add test for this.

kostko (commented):
Add a test for multiple different root types having the same hash at the same version and making sure that everything behaves correctly in this case (e.g., access, pruning).

kostko (approved):

PR #3416, opened 2020-10-16
Add ability to set initial block height at genesis

abukosek:
This adds the ability to change the initial block height at genesis time to something other than 1.


go/oasis-test-runner/oasis/oasis.go
```diff
@@ -294,6 +295,9 @@ type NetworkCfg struct { // nolint: maligned
 	// Consensus are the network-wide consensus parameters.
 	Consensus consensusGenesis.Genesis `json:"consensus"`
 
+	// InitialHeight is the initial block height.
+	InitialHeight int64 `json:"initial_height"`
```
kostko, line 299:
```suggestion
	InitialHeight int64 `json:"initial_height,omitempty"`
```

kostko (approved):

Issue #2596, opened 2020-01-24
Add tooling/docs to easily reproduce Oasis Core builds


tjanez:
##### SUMMARY
We should add the necessary tooling and document how to reproduce Oasis Core builds (at the moment, the `oasis-node` binary).

This is also a prerequisite to be able to verify an Oasis Core build before signing it (#2461).

##### ISSUE TYPE
- Release process
- Documentation

##### ADDITIONAL INFORMATION
Even though Go has great support for reproducible builds (version 1.13 added the `-trimpath` flag to the `go build` command), there are still a lot of steps that need to be taken before one is able to reproduce an Oasis Core build:

1. The used Go compiler must be the same.
2. The external dependencies (currently [libseccomp](https://github.com/seccomp/libseccomp) for `oasis-node`) must be the same.
3. The git checkout from which the build is made, must be the same (we use git to infer the version based on previous tags and commits).

Probably the easiest way to achieve reproducible builds is to:
  - use a well-known Linux container as the basis of a reproducible build environment,
  - install a particular official Go release inside the container,
  - install distribution-provided release of libseccomp's headers inside the container,
  - bind-mount the git checkout into the container,
  - build a release,
  - prepare [tarballs in a reproducible way](https://reproducible-builds.org/docs/archives/).

kostko:
This should be expanded to include reproducible runtime builds (probably the same environment should be used for both) to ensure that you can reproduce the exact MRENCLAVE.

tjanez:
> Probably the easiest way to achieve reproducible builds is to:
>     * use a well-known Linux container as the basis of a reproducible build environment,
>     * install a particular official Go release inside the container,
>     * install distribution-provided release of libseccomp's headers inside the container,

After https://github.com/oasisprotocol/oasis-core/pull/2970, we will be rebuilding Docker images automatically, so it would make sense to create a minimal Docker image, e.g. `oasisprotocol/oasis-core-build` or `oasisprotocol/oasis-core-reproducible`, with just the necessary tools to be able to reproducibly build all release artifacts (without other development or testing stuff).

tjanez:
Some more thoughts on the implementation:
- create a minimal Docker image, e.g. `oasisprotocol/oasis-core-build` or `oasisprotocol/oasis-core-reproducible`, with just the necessary tools to be able to reproducibly build all release artifacts (without other development or testing stuff)
[...]

PR #3423, opened 2020-10-16
ADR 0006: Consensus Governance

kostko:
[Rendered](https://github.com/oasisprotocol/oasis-core/blob/kostko/docs/adr-consensus-governance/docs/adr/0006-consensus-governance.md)

### DIFF
--- a/docs/adr/0006-consensus-governance.md
+++ b/docs/adr/0006-consensus-governance.md
@@ -1,0 +1,478 @@
+# ADR 0006: Consensus Governance
+
+## Changelog
+
+- 2020-10-27: Voting period in epochs, min upgrade cancellation difference,
+  failed proposal state
+- 2020-10-16: Initial draft
+
+## Status
+
+Proposed
+
+## Context
+
+Currently the consensus layer does not contain any on-chain governance
+mechanism so any network upgrades need to be carefully coordinated off-chain.
+An on-chain governance mechanism would allow upgrades to be handled in a more
+controlled (and automatable) manner without introducing the risk of corrupting
+state.
+
+## Decision
+
+This proposal introduces a minimal on-chain governance mechanism where anyone
+can submit governance proposals and the validators can vote where one base unit
+of delegated stake counts as one vote.
+
+The high-level overview is as follows:
+
+- **A new governance API** is added to the consensus layer and its Tendermint
+  based implementation. It supports transactions for submitting proposals and
+  voting on proposals. It supports queries for listing current proposals and
+  votes for any given proposal.
+
+- **Two governance proposal kinds are supported**, a consensus layer upgrade
+  proposal (where the content is basically the existing upgrade descriptor) and
+  the cancellation of a pending upgrade.
+
+A proposal is created through a _submit proposal_ transaction and requires a
+minimum deposit (which is later refunded in case the proposal passes). Once a
+proposal is successfully submitted the voting period starts. Entities that are
+part of the validator set may cast votes for the proposal. After the voting
+period completes, the votes are tallied and the proposal either passes or is
+rejected.
+
+In case the proposal passes, the actions specified in the content of the propsal
+are executed. Currently the only actions are scheduling of an upgrade by
+publishing an upgrade descriptor or cancelling a previously passed upgrade.
+
+### State
+
+#### Staking
+
+This proposal adds the following consensus layer state in the staking module:
+
+- **Governance deposits account balance (`0x59`)**, similar to the common pool.
+
+#### Governance
+
+This proposal adds the following consensus layer state in the governance module:
+
+- **Next proposal identifier (`0x80`)**
+
+  The next proposal identifier is stored as a CBOR-serialized `uint64`.
+
+- **List of proposals (`0x81`)**
+
+  Each proposal is stored under a separate storage key with the following key
+  format:
+
+  ```
+  0x81 <proposal-id (uint64)>
+  ```
+
+  And CBOR-serialized value:
+
+  ```golang
+  // ProposalState is the state of the proposal.
+  type ProposalState uint8
+
+  const (
+      StateActive   ProposalState = 1
+      StatePassed   ProposalState = 2
+      StateRejected ProposalState = 3
+      StateFailed   ProposalState = 4
+  )
+
+  type Proposal struct {
+      // ID is the unique identifier of the proposal.
+      ID uint64 `json:"id"`
+      // Submitter is the address of the proposal submitter.
+      Submitter staking.Address `json:"submitter"`
+      // State is the state of the proposal.
+      State ProposalState `json:"state"`
+      // Deposit is the deposit attached to the proposal.
+      Deposit quantity.Quantity `json:"deposit"`
+
+      // Content is the content of the proposal.
+      Content ProposalContent `json:"content"`
+
+      // CreatedAt is the epoch at which the proposal was created.
+      CreatedAt beacon.EpochTime `json:"created_at"`
+      // ClosesAt is the epoch at which the proposal will close and votes will
+      // be tallied.
+      ClosesAt beacon.EpochTime `json:"closes_at"`
+
+      // Results are the final tallied results after the voting period has
+      // ended.
+      Results map[Vote]quantity.Quantity `json:"results,omitempty"`
+      // InvalidVotes is the number of invalid votes after tallying.
+      InvalidVotes uint64 `json:"invalid_votes,omitempty"`
+  }
+  ```
+
+- **List of active proposals (`0x82`)**
+
+  Each active proposal (one that has not yet closed) is stored under a separate
+  storage key with the following key format:
+
+  ```
+  0x82 <closes-at (beacon.EpochTime)> <proposal-id (uint64)>
+  ```
+
+  The value is empty as the proposal ID can be inferred from the key.
+
+- **List of votes (`0x83`)**
+
+  Each vote is stored under a separate storage key with the following key
+  format:
+
+  ```
+  0x83 <proposal-id (uint64)> <voter-address (staking.Address)>
+  ```
+
+  And CBOR-serialized value:
+
+  ```golang
+  // Vote is a governance vote.
+  type Vote uint8
+
+  const (
+      VoteYes     Vote = 1
+      VoteNo      Vote = 2
+      VoteAbstain Vote = 3
+  )
+  ```
+
+- **List of pending upgrades (`0x84`)**
+
+  Each pending upgrade is stored under a separate storage key with the following
+  key format:
+
+  ```
+  0x84 <upgrade-epoch (beacon.EpochTime)> <proposal-id (uint64)>
+  ```
+
+### Genesis Document
+
+The genesis document needs to be updated to include a `governance` field with
+any initial state (see [_State_]) and consensus parameters (see [_Consensus
+Parameters_]) for the governance service.
+
+[_State_]: #state
+[_Consensus Parameters_]: #consensus-parameters
+
+### Transaction Methods
+
+This proposal adds the following transaction methods in the governance module:
+
+#### Submit Proposal
+
+Proposal submission enables a new consensus layer governance proposal to be
+created.
+
+**Method name:**
+
+```
+gov.SubmitProposal
+```
+
+**Body:**
+
+```golang
+// ProposalContent is a consensus layer governance proposal content.
+type ProposalContent struct {
+    Upgrade       *UpgradeProposal       `json:"upgrade,omitempty"`
+    CancelUpgrade *CancelUpgradeProposal `json:"cancel_upgrade,omitempty"`
+}
+
+// UpgradeProposal is an upgrade proposal.
+type UpgradeProposal struct {
+    upgrade.Descriptor
+}
+
+// CancelUpgradeProposal is an upgrade cancellation proposal.
+type CancelUpgradeProposal struct {
+    // ProposalID is the identifier of the pending upgrade proposal.
+    ProposalID uint64 `json:"proposal_id"`
+}
+```
+
+**Fields:**
+
+- `upgrade` (optional) specifies an upgrade proposal.
+- `cancel_upgrade` (optional) specifies an upgrade cancellation proposal.
+
+Exactly one of the proposal kind fields needs to be non-nil, otherwise the
+proposal is considered malformed.
+
+Upon processing any proposal the following steps are first performed:
+
+- The account indicated by the signer is loaded.
+
+- If the account balance is less than `min_proposal_deposit`, the method call
+  fails with `ErrInsufficientBalance`.
+
+Upon processing an **`UpgradeProposal`** the following steps are then performed:
+
+- The upgrade descriptor is checked for basic internal validity. If the check
+  fails, the method call fails with `ErrInvalidArgument`.
+
+- The upgrade descriptor's `epoch` field is compared with the current epoch. If
+  the specified epoch is not at least `upgrade_min_epoch_diff` epochs ahead of
+  the current epoch, the method call fails with `ErrUpgradeTooSoon`.
+
+- The set of pending upgrades is checked to make sure that no upgrades are
+  currently pending. If there is an existing upgrade pending, the method call
+  fails with `ErrUpgradeAlreadyPending`.
+
+Upon processing a **`CancelUpgradeProposal`** the following steps are then
+performed:
+
+- The set of pending upgrades is checked to make sure that the given upgrade
+  proposal is currently pending to be executed. If there is no such upgrade, the
+  method call fails with `ErrNoSuchUpgrade`.
+
+- The upgrade descriptor's `epoch` field is compared with the current epoch. If
+  the specified epoch is not at least `upgrade_cancel_min_epoch_diff` epochs
+  ahead of the current epoch, the method call fails with `ErrUpgradeTooSoon`.
+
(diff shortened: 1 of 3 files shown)


docs/adr/0006-consensus-governance.md
```diff
@@ -0,0 +1,455 @@
...
+of delegated stake counts as one vote.
+
+The high-level overview is as follows:
+
+- **A new governance API** is added to the consensus layer and its Tendermint
+  based implementation. It supports transactions for submitting proposals and
+  voting on proposals. It supports queries for listing current proposals and
+  votes for any given proposal.
+
+- **Two governance prpoposal kinds are supported**, a consensus layer upgrade
```
tjanez, line 32:
```suggestion
- **Two governance proposal kinds are supported**, a consensus layer upgrade
```

docs/adr/0006-consensus-governance.md
```diff
@@ -0,0 +1,455 @@
...
+      // Deposit is the deposit attached to the proposal.
+      Deposit quantity.Quantity `json:"deposit"`
+
+      // Content is the content of the proposal.
+      Content ProposalContent `json:"content"`
+
+      // CreatedAt is the height at which the proposal was created.
+      CreatedAt int64 `json:"created_at"`
+      // ClosesAt is the height at which the proposal will expire.
+      ClosesAt int64 `json:"expires_at"`
```
tjanez, line 100: `ClosesAt` or `ExpiresAt` :slightly_smiling_face:?

docs/adr/0006-consensus-governance.md
```diff
@@ -0,0 +1,455 @@
...
+- `voting_period` (number of blocks) specifies the number of blocks after which
+  the voting for a proposal is closed and the votes are tallied.
+
+- `quorum` (uint8) specifies the minimum percentage of voting power that needs
+  to be cast on a proposal for the result to be valid.
+
+- `threshold` (uint8) specifies the minimum percentage of `VoteYes` votes in
+  order for a proposal to be accepted.
+
+- `upgrade_min_epochs_diff` (epochs) specifies the minimum number of epochs
```
ptrus, line 441: Should there be a similar parameter for minimum diff between current height & proposal expiry, so something like `min_proposal_duration`?

kostko: You mean the `voting_period` which is defined slightly above? Or something different?

kostko: So when you create a proposal you only submit the `ProposalContent`, you can't define your own closing time etc.

ptrus: Oh right, yeah didn't notice that. Makes sense :+1:

docs/adr/0006-consensus-governance.md
```diff
@@ -0,0 +1,459 @@
...
+#### Cancel Upgrade Proposal
+
+When a cancel upgrade proposal is executed, the proposal identified by
+`proposal_id` is looked up and removed from the list of pending upgrades. In
+case the pending upgrade does not exist anymore, no action is performed.
+
+### Consensus Parameters
+
+This proposal introduces the following new consensus parameters in the
+governance module:
```
jcueto, line 431: Here we can specify: "Since these are consensus parameters, they need to be specified in the genesis file and will be the same across all proposals."

This will help clarify the nature of the parameters.

docs/adr/0006-consensus-governance.md
```diff
@@ -0,0 +1,459 @@
...
+Currently the consensus layer does not contain any on-chain governance
+mechanism so any network upgrades need to be carefully coordinated off-chain.
+An on-chain governance mechanism would allow upgrades to be handled in a more
+controlled (and automatable) manner without introducing the risk of corrupting
+state.
+
+## Decision
+
+This proposal introduces a minimal on-chain governance mechanism where anyone
+can submit governance proposals and the validators can vote where one base unit
```
jcueto, line 22: Maybe we can explicitly point out that this includes both self-delegations and delegations from other users, since conceptually sometimes people think about delegation as primarily in the "delegations from other users" case. 

We can just add parentheses at the end: "(including both self-delegations and delegations from other users)."

docs/adr/0006-consensus-governance.md
```diff
@@ -0,0 +1,459 @@
...
+  the current epoch, the method call fails with `ErrUpgradeTooSoon`.
+
+- The set of pending upgrades is checked to make sure that no upgrades are
+  currently pending. If there is an existing upgrade pending, the method call
+  fails with `ErrUpgradeAlreadyPending`.
+
+Upon processing a **`CancelUpgradeProposal`** the following steps are then
+performed:
+
+- The set of pending upgrades is checked to make sure that the given upgrade
```
ptrus, line 229: Who can cancel a specific proposal, I assume only the one initially submitting the proposal right?

There is currently nothing preventing last-second canceling of a proposal? Should there be something like that? (e.g. `cancel_min_epochs_diff`). Alternatively should the `CancelUpgradeProposal` require voting as well?

Additionally, I think I'm missing some context on why even having the cancel proposal, what's a common scenario where it would be needed? (things that come to mind are late-discovered bugs in the proposed upgrade?)

docs/adr/0006-consensus-governance.md
```diff
@@ -0,0 +1,459 @@
...
+```golang
+// ProposalContent is a consensus layer governance proposal content.
+type ProposalContent struct {
+    Upgrade       *UpgradeProposal       `json:"upgrade,omitempty"`
+    CancelUpgrade *CancelUpgradeProposal `json:"cancel_upgrade,omitempty"`
+}
+
+// UpgradeProposal is an upgrade proposal.
+type UpgradeProposal struct {
+    upgrade.Descriptor
```
ptrus, line 188: Is https://github.com/oasisprotocol/oasis-core/issues/2596 a prerequisite for this ADR? Or maybe at least a "good to have", so voters can actually confirm the proposed upgrade identifier (hash) matches a specific source code. (i guess alternatively, although not ideal, we could use something like hash of the source code?)

docs/adr/0006-consensus-governance.md
```diff
@@ -0,0 +1,459 @@
...
+}
```
ptrus, line 189: (not directly related) should we change the type of the existing string `Method` field in the `upgrade.Descriptor` to an enum?

docs/adr/0006-consensus-governance.md
kostko: > Who can cancel a specific proposal, I assume only the one initially submitting the proposal right?

Anyone (that can fund a deposit) can propose an upgrade cancellation. It needs to go through the usual governance process.

> Additionally, I think I'm missing some context on why even having the cancel proposal, what's a common scenario where it would be needed? Things that come to mind are late-discovered bugs in the proposed upgrade? anything else?

Yes, that is exactly the use case for having an upgrade cancellation proposal. I don't think there are any other scenarios.

kostko: > Alternatively should the CancelUpgradeProposal require voting as well?

Any proposal, including the cancel upgrade proposal, requires the full governance process, so voting.

> There is currently nothing preventing last-second canceling of a proposal? Should there be something like that? (e.g. cancel_min_epochs_diff).

Yeah that's probably a good idea, I'll update the ADR.

docs/adr/0006-consensus-governance.md
kostko: Yes, that's right. So in theory the hash could also be a Git commit or a hash of a manifest containing further information.

docs/adr/0006-consensus-governance.md
kostko: It is already an enum, currently the only valid value is `UpgradeMethInternal` which corresponds to the string `"internal"`. Do you mean switching from `string` to something like `uint8`?

docs/adr/0006-consensus-governance.md
ptrus: Makes sense, thanks. Yeah disregard the comment about requiring voting, i see that this results in a proposal that goes through the governance process :+1: 

So I think there should probably be an additional check here that the resulting proposal end height (`current_height` + `voting_period`) is not after the start of the epoch in which the `Upgrade` is taking place?

docs/adr/0006-consensus-governance.md
ptrus: yes

docs/adr/0006-consensus-governance.md
kostko: > So I think there should probably be an additional check here that the resulting proposal end height (current_height + voting_period) is not after the start of the epoch in which the Upgrade is taking place?

Either that or a sanity check that `voting_period` must be less than any of the `*_min_epochs_diff`?

kostko: > Although with variable epoch lengtths that are coming i don't think this is possible?

Oh good point, so we could also change the voting period to be in epochs instead?

ptrus: Voting period in epochs makes sense to me.

kostko: Done.

docs/adr/0006-consensus-governance.md
kostko: Yeah we can do that.

docs/adr/0006-consensus-governance.md
```diff
@@ -0,0 +1,478 @@
...
+- The upgrade descriptor is checked for basic internal validity. If the check
+  fails, the method call fails with `ErrInvalidArgument`.
+
+- The upgrade descriptor's `epoch` field is compared with the current epoch. If
+  the specified epoch is not at least `upgrade_min_epochs_diff` epochs ahead of
+  the current epoch, the method call fails with `ErrUpgradeTooSoon`.
+
+- The set of pending upgrades is checked to make sure that no upgrades are
+  currently pending. If there is an existing upgrade pending, the method call
+  fails with `ErrUpgradeAlreadyPending`.
```
bennetyee, line 227: This seems a little odd:  we prevent new proposals when one is pending, but as long as active proposals have not been voted on and moved to pending state, additional proposals can be made.  Is the intent to prevent two pending proposals from becoming ... "in use" (i.e., approved and switch-over epoch has passed, so presumably all node operators have picked up the implementation or implemented the proposal themselves)?  What happens (if anything) if two or more active proposals go into pending state and their switch-over epochs are close to each other, e.g., within upgrade_min_epochs_diff of each other?

Not a critical issue, but just wanted clarification / trying to understand the intent.

kostko: > we prevent new proposals when one is pending, but as long as active proposals have not been voted on and moved to pending state, additional proposals can be made

I think this is just a misunderstanding, there are two things to an upgrade:
* First a governance upgrade proposal is created and it goes through the voting process.
* Only if the vote passes is an upgrade pending.

So this does not refer to other pending governance upgrade proposals but to already passed pending upgrades.

> What happens (if anything) if two or more active proposals go into pending state and their switch-over epochs are close to each other, e.g., within upgrade_min_epochs_diff of each other?

See the _Proposal Content Execution_ section below where there is another check when the passed governance upgrade proposal is executed. Only one upgrade can be pending and if two upgrade proposals are passed, the second one will fail to execute.

docs/adr/0006-consensus-governance.md
```diff
@@ -0,0 +1,478 @@
...
+
+This proposal adds the following consensus layer state in the staking module:
+
+- **Governance deposits account balance (`0x59`)**, similar to the common pool.
+
+#### Governance
+
+This proposal adds the following consensus layer state in the governance module:
+
+- **Next proposal identifier (`0x80`)**
```
ptrus, line 61: If we follow the existing convention, I think these identifiers will start at `0x90` (and the governance app id will be `0x09`) since `0x08` is in use by the new [improved beacon already](https://github.com/oasisprotocol/oasis-core/pull/3180/files).

docs/adr/0006-consensus-governance.md
```diff
@@ -0,0 +1,478 @@
...
+      Content ProposalContent `json:"content"`
+
+      // CreatedAt is the epoch at which the proposal was created.
+      CreatedAt beacon.EpochTime `json:"created_at"`
+      // ClosesAt is the epoch at which the proposal will close and votes will
+      // be tallied.
+      ClosesAt beacon.EpochTime `json:"closes_at"`
+
+      // Results are the final tallied results after the voting period has
+      // ended. Note that they take the
```
tjanez, line 107: Something is missing here...

docs/adr/0006-consensus-governance.md
```diff
@@ -0,0 +1,478 @@
...
+    ActiveProposals(ctx context.Context, height int64) ([]*Proposal, error)
+
+    // Proposal looks up a specific proposal.
+    Proposal(ctx context.Context, query *ProposalQuery) (*Proposal, error)
+
+    // Votes looks up votes for a specific proposal.
+    Votes(ctx context.Context, query *ProposalQuery) ([]*VoteEntry, error)
+
+    // PendingUpgrades returns a list of all pending upgrades.
+    PendingUpgrades(ctx context.Context, height int64) ([]*upgrade.Descriptor, error)
```
ptrus, line 331: This should probably return some new type that also contains the information at which epoch the upgrade is scheduled for?

docs/adr/0006-consensus-governance.md
```diff
@@ -0,0 +1,478 @@
...
+- If the account balance is less than `min_proposal_deposit`, the method call
+  fails with `ErrInsufficientBalance`.
+
+Upon processing an **`UpgradeProposal`** the following steps are then performed:
+
+- The upgrade descriptor is checked for basic internal validity. If the check
+  fails, the method call fails with `ErrInvalidArgument`.
+
+- The upgrade descriptor's `epoch` field is compared with the current epoch. If
+  the specified epoch is not at least `upgrade_min_epochs_diff` epochs ahead of
```
tjanez, line 222: Should we perhaps use `upgrade_min_epoch_diff`?

```suggestion
  the specified epoch is not at least `upgrade_min_epoch_diff` epochs ahead of
```

docs/adr/0006-consensus-governance.md
kostko: The upgrade descriptor has that information already?

docs/adr/0006-consensus-governance.md
```diff
@@ -0,0 +1,478 @@
...
+      VoteNo      Vote = 2
+      VoteAbstain Vote = 3
+  )
+  ```
+
+- **List of pending upgrades (`0x84`)**
+
+  Each pending upgrade is stored under a separate storage key with the following
+  key format:
+
```
ptrus, line 151: I assume a CBOR-serialized `UpgradeProposal` should be stored here

docs/adr/0006-consensus-governance.md
ptrus: Ah yeah, that's true :-)

docs/adr/0006-consensus-governance.md
kostko: No need, the idea is that you just need to get the ID and then you can fetch the proposal from `0x81 ...`.

docs/adr/0006-consensus-governance.md
```diff
@@ -0,0 +1,478 @@
...
+Upon processing a **`CancelUpgradeProposal`** the following steps are then
+performed:
+
+- The set of pending upgrades is checked to make sure that the given upgrade
+  proposal is currently pending to be executed. If there is no such upgrade, the
+  method call fails with `ErrNoSuchUpgrade`.
+
+- The upgrade descriptor's `epoch` field is compared with the current epoch. If
+  the specified epoch is not at least `upgrade_cancel_min_epochs_diff` epochs
```
tjanez, line 237: Similar as above, consider using `upgrade_cancel_min_epoch_diff`.

docs/adr/0006-consensus-governance.md
ptrus: Yeah makes sense :+1:

tjanez (approved):

docs/adr/0006-consensus-governance.md
kostko: Hm, so I looked at the beacon when deciding on this and it uses the prefix from the previous beacon/epochtime apps which seems like the right thing to do:
https://github.com/oasisprotocol/oasis-core/blob/ae0e6719ffa2cc8e67a83c25943321a89e957cc7/go/consensus/tendermint/apps/beacon/state/state.go#L16-L32

The app id for the beacon should be changed to `0x04` to be consistent.

ptrus: Ok yeah I also saw the comment in the beacon PR :+1:

ptrus (approved):

pro-wh (approved):
my summary:

this adds on-chain yes/no/abstain voting for "proposals." entities can vote, and their votes are counted proportional to their stake if they are a validator at the time the proposal is set to close. as an optimization for stable validator committees, they must also be validator at the time they vote. they can also change their vote leading up to the proposal close time.

this adds a set of proposal types for managing network upgrades. the on-chain representation of a proposed upgrade https://github.com/oasisprotocol/oasis-core/blob/v20.12.1/go/upgrade/api/api.go#L66 is mostly opaque, other than the epoch when it happens. this type and the mechanisms around it are already part of the codebase.

peterjgilbert (approved):

PR #3435, opened 2020-10-19
e2e/runtime/runtime-upgrade: Wait for old compute nodes to expire

### LINKED ISSUE #3404: Runtime upgrade E2E test should wait for old node expiration
Currently [the runtime upgrade E2E test](https://github.com/oasisprotocol/oasis-core/blob/e79b4979c20447f4b47ba4e7a35ae67e74647a0e/go/oasis-test-runner/scenario/e2e/runtime/runtime_upgrade.go) shuts down the old compute nodes after upgrade and starts the client, assuming that the new compute nodes will process the transactions:
https://github.com/oasisprotocol/oasis-core/blob/e79b4979c20447f4b47ba4e7a35ae67e74647a0e/go/oasis-test-runner/scenario/e2e/runtime/runtime_upgrade.go#L211-L220

The problem is that the committee scheduler can still schedule one of the old nodes, making the test flaky as it depends on what committee is elected. We should probably wait for the old nodes to expire before proceeding.

kostko:
Fixes #3404

### DIFF
--- a/.changelog/3404.internal.md
+++ b/.changelog/3404.internal.md
@@ -1,0 +1,1 @@
+e2e/runtime/runtime-upgrade: Wait for old compute nodes to expire
--- a/go/oasis-test-runner/oasis/oasis.go
+++ b/go/oasis-test-runner/oasis/oasis.go
@@ -187,11 +187,16 @@ func (n *Node) WaitReady(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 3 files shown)


ptrus (approved):

Yawning:
Cherry-picking this onto my branch appears to improve things.

Yawning (approved):

PR #3439, opened 2020-10-21
ADR-0007: Improved Random Beacon

Yawning:
tldr; Read the paper and the branch


kostko (commented):
Only some stylistic nits, otherwise I think it describes the the proposed implementation well.

kostko (commented):
Also don't forget to update the ADR index.

kostko (approved):

Issue #3414, opened 2020-10-15
Add storage byzantine E2E test cases


ptrus:
Add storage byzantine tests to test the failure-indicating commitments added in https://github.com/oasisprotocol/oasis-core/pull/3391

Scenarios:
- honestly respond to txnscheduler requests so that round starts, then reject -> should trigger discrepancy and round failure with `ErrMajorityFailure`
- honestly respond to txnscheduler and other executor worker request, but fail on runtime requests (todo: not sure if  there's a nice way to do this, maybe there's some methods that only runtime uses?) -> should trigger discrepancy and round failure with `ErrMajorityFailure`
- also add executor byzantine test that submits failure indicating commitments

Issue #3440, opened 2020-10-21
Additional validation for IO roots


kostko:
After #2053 the runtime storage nodes should perform additional validation when accepting updates via `Apply` for IO roots:
* The format of each key must be a valid [transaction key or a valid tag key](https://github.com/oasisprotocol/oasis-core/blob/6ada0fe410a9bc384a6b2f595c831c544bb6b5fc/go/runtime/transaction/transaction.go#L77-L86).
* The number of transaction inputs or outputs must be at most `runtime.txn_scheduler.max_batch_size`.
* The size of transaction inputs must be at most `runtime.txn_scheduler.max_batch_size_bytes`.

PR #3442, opened 2020-10-22
go/e2e/byzantine: Add storage support to byzantine node

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3414


go/oasis-node/cmd/debug/byzantine/byzantine.go
```diff
@@ -68,354 +113,127 @@ func activateCommonConfig(cmd *cobra.Command, args []string) {
...
-	}
-
-	ht := newHonestTendermint()
-	if err = ht.start(defaultIdentity, common.DataDir()); err != nil {
-		panic(fmt.Sprintf("honest Tendermint start failed: %+v", err))
-	}
-	defer func() {
-		if err1 := ht.stop(); err1 != nil {
-			panic(fmt.Sprintf("honest Tendermint stop failed: %+v", err1))
+	switch executorMode {
```
kostko, line 176: Add an explicit default case with a comment.

go/oasis-node/cmd/debug/byzantine/executor.go
```diff
@@ -71,7 +71,18 @@ ReceiveTransactions:
...
+		ctx,
+		defaultRuntimeID,
+		&p2p.Message{
+			GroupVersion:  groupVersion,
+			ProposedBatch: batch,
+		},
+	)
+}
+
+func (cbc *computeBatchContext) prepareTransactionBatch(ctx context.Context, clients []*storageClient, currentBlock *block.Block, batch []*api.Tx, identity *identity.Identity, corrupt bool) (*commitment.SignedProposedBatch, error) {
```
kostko, line 85: Break this into multiple lines.

go/oasis-node/cmd/debug/byzantine/executor.go
```diff
@@ -116,24 +127,20 @@ func (cbc *computeBatchContext) proposeTransactionBatch(ctx context.Context, gro
 		StorageSignatures: ioReceiptSignatures,
 		Header:            currentBlock.Header,
 	}
+	// Corrupt the proposed batch.
+	if corrupt {
+		dispatchMsg.IORoot = emptyRoot.Hash
```
kostko, line 132: So this corrupts it in a way that receipts are also invalidated and as such the signature verification will fail.

You could also corrupt it in other ways (at least until #3440 is implemented as then storage nodes would reject these):
* Storing garbage in the IO root while keeping the receipts valid. Nodes should already correctly reject such proposals.
* Putting more transactions in the IO root than allowed by `MaxBatchSize`/`MaxBatchSizeBytes`. Nodes should already correctly reject such proposals.

go/oasis-node/cmd/debug/byzantine/storage_node.go
```diff
@@ -0,0 +1,143 @@
...
+func (w *storageWorker) SyncIterate(ctx context.Context, request *syncer.IterateRequest) (*syncer.ProofResponse, error) {
+	if w.failReadRequests {
+		return nil, errByzantine
+	}
+
+	return w.backend.SyncIterate(ctx, request)
+}
+
+func (w *storageWorker) Apply(ctx context.Context, request *storage.ApplyRequest) ([]*storage.Receipt, error) {
+	return w.backend.Apply(ctx, request)
```
kostko, line 99: Should we also test with failing applies (this is used when creating the transaction batch)?

kostko (approved):

go/oasis-node/cmd/debug/byzantine/storage_node.go
ptrus: Added it.

Didn't include it before as nothing interesting really happens, as the proposer should just retry proposing batches until Apply succeeds, and this is something that's regularly tested in the daily e2e tests. But nevertheless doesn't hurt adding it.

go/oasis-node/cmd/debug/byzantine/executor.go
ptrus: Added a comment about how this corrupts the proposed batch.  Yeah probably doesn't make sense adding the mentioned corruptions since they'll soon be already rejected by the storage worker.

An interesting corruption would be something that passes all the checks but would only fail at runtime processing, which should results in all other nodes submitting failure indicating commitments -  but i don't think there's a way to do this (it'll always be caught before).

Issue #3412, opened 2020-10-15
Runtime client should retry checking blocks


ptrus:
Runtime client should periodically re-check blocks in case checking block for transaction fails (e.g. due to a no storage node being available):

https://github.com/oasisprotocol/oasis-core/blob/1e3a6fe489f9077864cff8e4c779f2099b24db6e/go/runtime/client/watcher.go#L69-L73

Discovered in the daily e2e tests.

kostko:
This is similar to #3388.

PR #3443, opened 2020-10-23
go/runtime/client: Recheck failed blocks and implement max tx age

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3412

- runtime client remembers failed roothash blocks and rechecks them on every new received block
- additionally transactions that do not succeed in 1500 blocks are considered failed and client drops them
  - as a future improvement should the client also support `SubmitTxNoWait` operation (opened https://github.com/oasisprotocol/oasis-core/issues/3444)

### DIFF
--- a/go/runtime/client/client.go
+++ b/go/runtime/client/client.go
@@ -6,6 +6,9 @@ import (
 	"fmt"
 	"sync"
 
+	flag "github.com/spf13/pflag"
+	"github.com/spf13/viper"
+
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
@@ -13,6 +16,7 @@ import (
 	consensus "github.com/oasisprotocol/oasis-core/go/consensus/api"
 	keymanagerAPI "github.com/oasisprotocol/oasis-core/go/keymanager/api"
 	keymanager "github.com/oasisprotocol/oasis-core/go/keymanager/client"
+	cmdFlags "github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common/flags"
 	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
 	"github.com/oasisprotocol/oasis-core/go/roothash/api/block"
 	"github.com/oasisprotocol/oasis-core/go/runtime/client/api"
@@ -25,9 +29,20 @@ import (
 	executor "github.com/oasisprotocol/oasis-core/go/worker/compute/executor/api"
 )
 
+const (
+	// CfgMaxTransactionAge is the number of consensus blocks after which
+	// submitted transactions will be considered expired.
+	CfgMaxTransactionAge = "runtime.client.max_transaction_age"
+
+	minMaxTransactionAge = 30
+)
+
 var (
 	_ api.RuntimeClient    = (*runtimeClient)(nil)
 	_ enclaverpc.Transport = (*runtimeClient)(nil)
+
+	// Flags has the flags used by the runtime client.
+	Flags = flag.NewFlagSet("", flag.ContinueOnError)
 )
 
 type clientCommon struct {
@@ -47,6 +62,8 @@ type runtimeClient struct {
 
 	watchers  map[common.Namespace]*blockWatcher
 	kmClients map[common.Namespace]*keymanager.Client
+
+	maxTransactionAge int64
 
 	logger *logging.Logger
 }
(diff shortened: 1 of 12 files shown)


kostko:
> should this be configurable?

Yeah probably.

> as a future improvement should the client also support SubmitTxNoWait operation (similar to the one in consensus layer)?

Good idea, open an issue.

Yawning (approved):

go/runtime/client/client.go
```diff
@@ -428,6 +449,11 @@ func New(
 	runtimeRegistry runtimeRegistry.Registry,
 	p2p *p2p.P2P,
 ) (api.RuntimeClient, error) {
+	maxTransactionAge := viper.GetInt64(CfgMaxTransactionAge)
+	if maxTransactionAge < minMaxTransactionAge && !cmdFlags.DebugDontBlameOasis() {
+		return nil, fmt.Errorf("max transaction age to low: %d", maxTransactionAge)
```
kostko, line 454:
```suggestion
		return nil, fmt.Errorf("max transaction age too low: %d", maxTransactionAge)
```
(also maybe mention what the minimum is)

go/runtime/client/client.go
```diff
@@ -436,9 +462,16 @@ func New(
...
+		watchers:          make(map[common.Namespace]*blockWatcher),
+		kmClients:         make(map[common.Namespace]*keymanager.Client),
+		maxTransactionAge: maxTransactionAge,
+		logger:            logging.GetLogger("runtime/client"),
 	}
 	return c, nil
 }
+
+func init() {
+	Flags.Int64(CfgMaxTransactionAge, 1500, "number of blocks after witch the submitted transaction will be considered expired")
```
kostko, line 474:
```suggestion
	Flags.Int64(CfgMaxTransactionAge, 1500, "number of consensus blocks after witch the submitted transaction will be considered expired")
```

go/runtime/client/watcher.go
```diff
@@ -103,64 +121,112 @@ func (w *blockWatcher) watch() {
...
-			// If this is the initial block or an epoch transition block,
-			// update latest known group version and resend all transactions.
-			if gotInitialCommittee && blk.Block.Header.HeaderType != block.EpochTransition {
-				continue
+			var failedBlocks []*block.Block
+			for _, b := range w.toBeChecked {
+				if err = w.checkBlock(b); err != nil {
+					w.Logger.Error("error checking block",
+						"err", err,
+						"block", b,
```
kostko, line 166: Should this maybe only log the round?

go/runtime/client/watcher.go
```diff
@@ -103,64 +121,112 @@ func (w *blockWatcher) watch() {
...
+					)
+					failedBlocks = append(failedBlocks, b)
+				}
 			}
+			w.toBeChecked = failedBlocks
```
kostko, line 171: Should there be a (configurable) limit on the size of this list? The questions is what should happen once that limit is reached.

ptrus: I'm not sure, if anything i think it might make more sense to have a limit on number of transactions being watched. I guess it wouldn't hurt at least logging a warning in case the `failedBlocks` is non-zero though.

kostko: But this doesn't have anything to do with the number of watched transactions, right? There could be a single transaction and 1000 blocks waiting to be re-checked? And yeah logging sounds like a good idea.

ptrus: Yeah it's not related. I just think that it's more realistic that the number of transaction will grow big, rather than number of blocks needing to be rechecked, since in case there's actually problem with the storage nodes (the only reason checking blocks can currently fail), there probably won't be any new roothash blocks anyway. Unless the storage nodes are specifically blocking client requests.

go/runtime/client/client.go
ptrus: fixed

go/runtime/client/client.go
ptrus: :+1:

go/runtime/client/watcher.go
ptrus: Yeah change to log `round` instead of the `block`.

go/runtime/client/watcher.go
kostko: Oh one thing that would be useful is to clear this list in case there are no more pending transactions? Because it doesn't make sense to re-check blocks if all transactions have either completed or expired.

go/runtime/client/client.go
```diff
@@ -25,9 +29,20 @@ import (
 	executor "github.com/oasisprotocol/oasis-core/go/worker/compute/executor/api"
 )
 
+const (
+	// CfgMaxTransactionAge is the number of consensus blocks after witch
```
Yawning, line 33: While I appreciate the seasonal spirit, `which`.

go/runtime/client/watcher.go
ptrus: agreed. Hm wondering if instead should just make `checkBlock` first check if there's any pending transactions and in case there aren't any just successfully return. Since it also doesn't make sense checking new blocks in case there aren't any pending transactions.

ptrus: I instead made `checkBlock` check if there's any pending transactions.

kostko (approved):

PR #3448, opened 2020-10-26
go/roothash: Refactor runtime messages

### LINKED ISSUE #3430: Refactor runtime messages
Runtime messages should be refactored, so that only a hash is provided in `ComputeResultsHeader` as specified in [ADR 0003](https://github.com/oasisprotocol/oasis-core/blob/master/docs/adr/0003-consensus-runtime-token-transfer.md).

kostko:
Fixes #3430 

TODO
* [x] Validation of per-runtime `MaxMessages` against global `MaxRuntimeMessages` roothash consensus parameter.
* [x] Validation of messages when submitted by transaction scheduler.
* [x] Emit message processing results.
* [x] Delivery of message results via RHP.
* [x] E2E test with a "ping/noop" message.
* [x] Use height of last successful round instead of height of last round when querying for message results.
* [x] Update runtime txsource workload to also emit messages.
* [x] Fix restore from genesis when events were emitted in the last round before dump.

### DIFF
--- a/go/consensus/tendermint/api/app.go
+++ b/go/consensus/tendermint/api/app.go
@@ -1,15 +1,39 @@
 package api
 
 import (
+	"errors"
+
 	tmabcitypes "github.com/tendermint/tendermint/abci/types"
 
 	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"
 	genesis "github.com/oasisprotocol/oasis-core/go/genesis/api"
 )
 
+// ErrNoSubscribers is the error returned when publishing a message that noone is subscribed to.
+var ErrNoSubscribers = errors.New("no subscribers to given message kind")
+
+// MessageSubscriber is a message subscriber interface.
+type MessageSubscriber interface {
+	// ExecuteMessage executes a given message.
+	ExecuteMessage(ctx *Context, kind, msg interface{}) error
+}
+
+// MessageDispatcher is a message dispatcher interface.
+type MessageDispatcher interface {
+	// Subscribe subscribes a given message subscriber to messages of a specific kind.
+	Subscribe(kind interface{}, ms MessageSubscriber)
+
+	// Publish publishes a message of a given kind by dispatching to all subscribers.
+	//
+	// In case there are no subscribers ErrNoSubscribers is returned.
+	Publish(ctx *Context, kind, msg interface{}) error
+}
+
 // Application is the interface implemented by multiplexed Oasis-specific
 // ABCI applications.
 type Application interface {
+	MessageSubscriber
+
 	// Name returns the name of the Application.
 	Name() string
 
@@ -37,7 +61,7 @@ type Application interface {
 
 	// OnRegister is the function that is called when the Application
 	// is registered with the multiplexer instance.
-	OnRegister(state ApplicationState)
+	OnRegister(ApplicationState, MessageDispatcher)
 
 	// OnCleanup is the function that is called when the ApplicationServer
 	// has been halted.
--- a/go/consensus/tendermint/api/app.go
+++ b/go/consensus/tendermint/api/app.go
@@ -70,13 +70,6 @@ type Application interface {
 	// ExecuteTx executes a transaction.
 	ExecuteTx(*Context, *transaction.Transaction) error
 
-	// ForeignExecuteTx delivers a transaction of another application for
-	// processing.
-	//
-	// This can be used to run post-tx hooks when dependencies exist
-	// between applications.
-	ForeignExecuteTx(*Context, Application, *transaction.Transaction) error
-
 	// InitChain initializes the blockchain with validators and other
 	// info from TendermintCore.
 	//
--- a/go/consensus/tendermint/api/app.go
+++ b/go/consensus/tendermint/api/app.go
@@ -29,6 +29,18 @@ type MessageDispatcher interface {
 	Publish(ctx *Context, kind, msg interface{}) error
 }
 
+// NoopMessageDispatcher is a no-op message dispatcher that performs no dispatch.
+type NoopMessageDispatcher struct{}
+
+// Implements MessageDispatcher.
+func (nd *NoopMessageDispatcher) Subscribe(interface{}, MessageSubscriber) {
+}
+
+// Implements MessageDispatcher.
+func (nd *NoopMessageDispatcher) Publish(*Context, interface{}, interface{}) error {
+	return nil
+}
+
 // Application is the interface implemented by multiplexed Oasis-specific
 // ABCI applications.
 type Application interface {
--- a/go/consensus/tendermint/apps/registry/api/api.go
+++ b/go/consensus/tendermint/apps/registry/api/api.go
@@ -1,0 +1,10 @@
+// Package api defines the registry application API for other applications.
+package api
+
+type messageKind uint8
+
+// MessageNewRuntimeRegistered is the message kind for new runtime registrations. The message is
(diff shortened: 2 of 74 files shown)


go/consensus/tendermint/api/app.go
```diff
@@ -1,15 +1,51 @@
...
+// MessageSubscriber is a message subscriber interface.
+type MessageSubscriber interface {
+	// ExecuteMessage executes a given message.
+	ExecuteMessage(ctx *Context, kind, msg interface{}) error
+}
+
+// MessageDispatcher is a message dispatcher interface.
+type MessageDispatcher interface {
+	// Subscribe subscribes a given message subscriber to messages of a specific kind.
+	Subscribe(kind interface{}, ms MessageSubscriber)
```
ptrus, line 24: Is it expected that message kinds will get more complex in future? 
If not, it might be worth considering a single type for all message kinds (instead of having per module types all defined the same way). Would also make the type signatures here more descriptive as `kinds` could be a concrete type and not `interface{}`

kostko: The interface is not there to allow more complex kinds but rather to allow each module to define its own kinds in its respective `api` packages without needing a central place to list all kinds.

abukosek (approved):

go/consensus/tendermint/api/app.go
```diff
@@ -1,15 +1,51 @@
...
+
+	// Publish publishes a message of a given kind by dispatching to all subscribers.
+	//
+	// In case there are no subscribers ErrNoSubscribers is returned.
+	Publish(ctx *Context, kind, msg interface{}) error
+}
+
+// NopMessageDispatcher is a no-op message dispatcher that performs no dispatch.
+type NopMessageDispatcher struct{}
```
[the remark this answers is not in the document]
kostko: Good idea, fixed.

go/roothash/api/api.go
```diff
@@ -166,7 +205,8 @@ type Event struct {
 
 	ExecutorCommitted            *ExecutorCommittedEvent            `json:"executor_committed,omitempty"`
 	ExecutionDiscrepancyDetected *ExecutionDiscrepancyDetectedEvent `json:"execution_discrepancy,omitempty"`
-	FinalizedEvent               *FinalizedEvent                    `json:"finalized,omitempty"`
+	Finalized                    *FinalizedEvent                    `json:"finalized,omitempty"`
```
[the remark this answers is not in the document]
kostko: Yeah good idea, added an explicit changelog fragment for this change.

go/consensus/tendermint/api/app.go
ptrus: Yeah that probably is indeed nicer :+1:

go/roothash/api/api.go
```diff
@@ -130,6 +138,25 @@ func NewRequestProposerTimeoutTx(nonce uint64, fee *transaction.Fee, runtimeID c
...
+	Suspended bool              `json:"suspended,omitempty"`
+
+	GenesisBlock *block.Block `json:"genesis_block"`
+
+	CurrentBlock       *block.Block `json:"current_block"`
+	CurrentBlockHeight int64        `json:"current_block_height"`
+
+	// LastNormalRound is the runtime round which was normally processed by the runtime. This is
+	// also the round that contains the message results for the last processed runtime messages.
+	LastNormalRound uint64 `json:"last_normal_round"`
```
ptrus, line 153: Maybe `LastNonEmpty{Round,Height}` would be more descriptive? Don't have any better suggestion, but i feel "Normal" is not the most descriptive.

go/registry/api/runtime.go
```diff
@@ -92,6 +92,9 @@ type ExecutorParameters struct {
 
 	// RoundTimeout is the round timeout in consensus blocks.
 	RoundTimeout int64 `json:"round_timeout"`
+
+	// MaxMessages is the maximum number of messages that can be emitted by the runtime.
```
ptrus, line 96: Maybe mention in the comment that this is the maximum number of messages _per round_.

go/roothash/api/commitment/executor.go
```diff
@@ -151,10 +164,17 @@ func (m *ComputeBody) ValidateBasic() error {
 		if header.StateRoot != nil {
 			return fmt.Errorf("failure indicating commitment includes StateRoot")
 		}
+		if header.MessagesHash != nil {
+			return fmt.Errorf("failure indicating commitment includes MessagesHash")
```
ptrus, line 168: Could also add cases for these two added failures to the `TestValidateBasic` test.

ptrus: Note that for other failure cases the tests currently exist in pool_test.go (i think i've put them there before adding the `ValidateBasic` function, and then forget moving those)
https://github.com/oasisprotocol/oasis-core/blob/64136dbb76df69f02d722a3b8ef919733589b51b/go/roothash/api/commitment/pool_test.go#L183-L193

Those should probably also be moved into `TestValidateBasic` and just a single "bad" failure indicating committeemen test could be left there.

go/roothash/api/block/message.go
```diff
@@ -1,6 +1,43 @@
...
+import (
+	"fmt"
+
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
+)
+
+// Message is a message that can be sent by a runtime.
 type Message struct {
-	// No valid messages are currently defined.
+	Noop *NoopMessage `json:"noop,omitempty"`
```
ptrus, line 11: The Noop message here just to ensure that runtime message propagation/processing works, while we don't have any other messages? Maybe add a comment explaining the need/usage of the Noop message.

ptrus (commented):
Some minor comments

go/roothash/api/api.go
kostko: The reason why I used "Normal" is that this is the header type (`HeaderType`) of "normal" (:grin:) blocks. I agree that it is not the best term.

go/registry/api/runtime.go
kostko: Good idea.

go/roothash/api/block/message.go
kostko: Yeah, I could remove these in #3511 even.

go/roothash/api/commitment/executor.go
kostko: Done, please take another look.

go/consensus/tendermint/apps/registry/api/api.go
```diff
@@ -0,0 +1,18 @@
...
+	// MessageNewRuntimeRegistered is the message kind for new runtime registrations. The message is
+	// the runtime descriptor of the runtime that has been registered.
+	//
+	// The message is not emitted for runtime descriptor updates.
+	MessageNewRuntimeRegistered = messageKind(0)
+
+	// MessageRuntimeUpdated is the message kind for runtime registration updates. The message is
+	// the runtime descriptor of the runtime that has been updated.
+	//
+	// The message is also emitted for new runtime registrations.
```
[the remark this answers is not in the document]
kostko: I added a separate `MessageRuntimeResumed` message for this case (currently doesn't do anything).

go/roothash/api/api.go
ptrus: Ah yeah that makes sense :-) Potentially could use `LastNormalBlock{Round,Height}`, but i'm also fine if it stays at it is.

ptrus (approved):
Looks good, thanks for the updates!

Issue #3409, opened 2020-10-15
Enable crash points in long-term tests


kostko:
We should enable crash points on some nodes in long-term tests.

PR #3449, opened 2020-10-26
Ptrus/feature/txsource crash points

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3409

TODO:
- [x] test a daily txsource run with crash points enabled: https://buildkite.com/oasisprotocol/oasis-core-long-tests/builds/288#e13c4fc6-83cd-4381-83be-cf1ba1fd283b
  - wtih current configuration there were 75 crashes in the above 4 hour test run, probably can reduce the probability a bit further :grin:
- [x] test run with longepochs & crash points enabled: https://buildkite.com/oasisprotocol/oasis-core-long-tests/builds/304#29a1b084-4444-4eae-83fd-62738a0f904c
  - reduced the prob. by a half from the previous run, now there's about ~10 crashes per hour

### DIFF
--- a/go/common/crash/crash.go
+++ b/go/common/crash/crash.go
@@ -6,6 +6,7 @@ package crash
 import (
 	"fmt"
 	"runtime"
+	"sync"
 	"time"
 
 	flag "github.com/spf13/pflag"
@@ -18,9 +19,14 @@ import (
 
 var testForceEnable bool
 
-// defaultCLIPrefix is the default CLI prefix used to configure crash points in
-// viper and cobra.
-const defaultCLIPrefix = "debug.crash"
+const (
+	// defaultCLIPrefix is the default CLI prefix used to configure crash points in
+	// viper and cobra.
+	defaultCLIPrefix = "debug.crash"
+
+	// CfgDefaultCrashPointProbability is the default crash point probability.
+	CfgDefaultCrashPointProbability = "debug.crash.default"
+)
 
 // RandomProvider interface that provides a Float64 random.
 type RandomProvider interface {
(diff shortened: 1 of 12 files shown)


go/common/crash/crash.go
```diff
@@ -54,14 +65,18 @@ func newDefaultRandomProvider() RandomProvider {
 }
 
 func defaultCrashMethod() {
-	runtime.Breakpoint()
+	// TODO: is there anything wrong with just using os.Exit here?
```
kostko, line 68: As long as this exits immediately without giving the program a way to cleanup, this is fine.

go/oasis-test-runner/oasis/compute.go
```diff
@@ -41,6 +43,8 @@ type ComputeCfg struct {
 
 	RuntimeProvisioner string
 
+	CrashPointsProbability float64
```
kostko, line 46: Should this move into `NodeCfg`?

go/oasis-test-runner/oasis/compute.go
```diff
@@ -26,6 +26,8 @@ type Compute struct { // nolint: maligned
 
 	runtimeProvisioner string
 
+	crashPointsProbability float64
```
kostko, line 29: Should this move into `Node`?

go/common/crash/crash.go
ptrus: In case of non-zero exit code it just does `syscall.Exit(code)`

Yawning: Just use `os.Exit(code)`, using the syscall package is frowned upon unless absolutely required.

kostko: Yeah in this case just remove the comment.

kostko (approved):

Issue #3453, opened 2020-10-28
Storage node should update access control policy on new storage node registrations


kostko:
Currently the storage node only updates its access control policy when either i) the committee changes or ii) a node within the currently elected committee changes (updates its registration). The access control policy for storage nodes includes allowing access to any storage node which is registered for the runtime, but the policy is not updated if none of the above happens (e.g., in case existing nodes don't update registrations and the committee doesn't change).

We should make it so that the storage node watches for storage node registration updates for the given runtime and updates the policy by calling `n.updateExternalServicePolicyLocked`.

This should be backwards compatible and backported to the 20.11.x stable branch.

PR #3455, opened 2020-10-28
Ptrus/fix/storage update policy

ptrus:
Fixes: #3453

Also changes to storage-sync test to ensure syncing works in the same epoch in which the node registers (also confirmed that this updated test fails on old code).

### DIFF
--- a/go/runtime/committee/nodes.go
+++ b/go/runtime/committee/nodes.go
@@ -39,6 +39,9 @@ type NodeDescriptorLookup interface {
 	// LookupTags looks up tags for a given node.
 	LookupTags(id signature.PublicKey) []string
 
+	// GetNodes returns current list of nodes.
+	GetNodes() []*node.Node
+
 	// WatchNodeUpdates subscribes to notifications about node descriptor updates.
 	//
 	// On subscription the current nodes will be sent immediately.
@@ -192,6 +195,17 @@ func (nw *nodeDescriptorWatcher) LookupByPeerID(id signature.PublicKey) *node.No
 	return nw.nodesByPeerID[id]
 }
 
+func (nw *nodeDescriptorWatcher) GetNodes() []*node.Node {
+	nw.RLock()
+	defer nw.RUnlock()
+
+	nodes := make([]*node.Node, 0, len(nw.nodes))
+	for _, v := range nw.nodes {
+		nodes = append(nodes, v)
+	}
+	return nodes
+}
+
 func (nw *nodeDescriptorWatcher) LookupTags(id signature.PublicKey) []string {
 	nw.RLock()
 	defer nw.RUnlock()
@@ -329,6 +343,17 @@ func (f *filteredNodeDescriptorLookup) LookupByPeerID(id signature.PublicKey) *n
 // Implements NodeDescriptorLookup.
 func (f *filteredNodeDescriptorLookup) LookupTags(id signature.PublicKey) []string {
 	return f.base.LookupTags(id)
+}
+
+// Implements NodeDescriptorLookup.
+func (f *filteredNodeDescriptorLookup) GetNodes() (filtered []*node.Node) {
+	for _, v := range f.base.GetNodes() {
+		tags := f.base.LookupTags(v.ID)
+		if f.filter(v, tags) {
+			filtered = append(filtered, v)
+		}
+	}
+	return
 }
 
 // Implements NodeDescriptorLookup.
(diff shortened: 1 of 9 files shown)


Yawning (approved):

go/worker/storage/committee/node.go
```diff
@@ -629,6 +630,43 @@ func (n *Node) worker() { // nolint: gocyclo
...
+	nodeCh, nodeSub, err := n.commonNode.Consensus.Registry().WatchNodes(n.ctx)
+	if err != nil {
+		n.logger.Error("worker/storage: failed to watch registry nodess",
+			"err", err,
+		)
+		return
+	}
+
+	// Watch for storage node registration updates for the runtime and update the policy.
+	go func() {
```
kostko, line 642: Should this, together with the subscription above, be moved to its own function and then just spawned here?

ptrus: yeah can do that

ptrus: done (also some other minor changes)

ptrus: Actually thinking a bit more about this, there's some more things that i'd like to fix, i'll move this back to a draft PR for now:
- avoid doing the query that queries ALL nodes while having the crossNode lock, instead the new watcher can maintain an up to date list of all registered storage nodes
https://github.com/oasisprotocol/oasis-core/blob/398e7186ccf243b0d07c8a0a5fb17dac38a5d213/go/worker/storage/committee/node.go#L394-L397
- this new code resulted in some unneeded calls to the `updateExternalServicePolicyLocked` since it's also called for all registered compute worker updates (while before it was just for compute nodes in the committee)
- There is also a race condition i think - it can happen now that the new watcher gets a (compute) committee node update notification before the node descriptor in the snapshot is updated (since we trigger `updateExternalServicePolicyLocked` from the new watcher, while we still query the `group.EpochSnapshot` for descriptors)

ptrus: @kostko  please take a look again, i think this is a more proper solution now

go/worker/storage/committee/node.go
```diff
@@ -627,6 +711,11 @@ func (n *Node) worker() { // nolint: gocyclo
 		return
 	}
 
+	// Start runtime node watcher.
+	go n.runtimeNodesWatcher()
+	// Start the quit watcher.
+	go n.watchQuit()
```
kostko, line 717: Isn't this something that should be spawned in `Start()`?

kostko (commented):
Cool, this looks nicer, some minor comments.

go/worker/storage/committee/node.go
ptrus: yeah agreed, makes more sense spawning there

go/runtime/committee/nodes.go
```diff
@@ -192,6 +195,16 @@ func (nw *nodeDescriptorWatcher) LookupByPeerID(id signature.PublicKey) *node.No
 	return nw.nodesByPeerID[id]
 }
 
+func (nw *nodeDescriptorWatcher) NodeList() (nodes []*node.Node) {
+	nw.RLock()
+	defer nw.RUnlock()
+
+	for _, v := range nw.nodes {
+		nodes = append(nodes, v)
+	}
```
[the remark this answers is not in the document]
ptrus: I don't think i can do that, since `nw.nodes` is a map of nodes.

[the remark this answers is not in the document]
kostko: Ah didn't see that, yeah in that case that won't work.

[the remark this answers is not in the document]
ptrus: done

go/worker/storage/committee/node.go
```diff
@@ -615,8 +588,120 @@ func (n *Node) flushSyncedState(summary *blockSummary) uint64 {
...
+		scheduler.KindComputeExecutor,
+		runtimeCommittee.WithAutomaticEpochTransitions(),
+	)
+	if err != nil {
+		n.logger.Error("worker/storage: failed to establish executor committee watcher",
+			"err", err,
+		)
+		return
+	}
+	committeeCh, committeeSub, err := committeeWatcher.Nodes().WatchNodeUpdates()
```
kostko, line 650: Could we use `n.Group.Nodes().WatchNodeUpdates()` instead? This would avoid spawning another committee watcher.

If you only want executor nodes, you could also filter it by doing something like:
```golang
committee.NewFilteredNodeLookup(nodes, committee.TagFilter(tagForCommittee(scheduler.KindExecutor)))
```
(would need to expose `tagForCommittee` though)

ptrus: good idea, done :+1:

kostko (approved):

PR #3456, opened 2020-10-30
go/common/crypto/signature: Add missing small-order points

kostko:


kostko:
Not needed as this will be fixed in our ed25519 library. We should remove the existing small order points from the blacklist when the library is updated.

Issue #3454, opened 2020-10-28
Runtime storage sync should use any storage node


kostko:
Currently storage node sync only uses nodes from the current storage committee. Instead, it should be also able to use (maybe with a lower priority) other storage nodes registered for the runtime to increase availability as certain runtime configurations (e.g., using a storage committee of size 1) make it more likely that the active committee does not have the required roots.

This should be backwards compatible and backported to the 20.11.x stable branch.

PR #3457, opened 2020-10-30
go/storage: Sync from all registered nodes for runtime

ptrus:
Fixes: #3454

Done:
- [x] refactor `committee.NodeDescriptorWatcher` into a more general `NodeDescriptorWatcher` that doesn't assumes committees
  - now in `runtime/nodes/node.go`
  - supports both `Versioned` (suitable for watching committees) and non-`Versioned`  mode (e.g. suitable for watching all registered nodes for a runtime)
- [x] refactor `committee.Client` into more generalized `nodes.Client` (that now supports the more general `NodeDescriptorWatcher`)
- [x] address TODOs some more cleanups / minor refactorings
- [x] add a test for syncing from a out-of-committee node

### DIFF
--- a/go/runtime/nodes/grpc/grpc.go
+++ b/go/runtime/nodes/grpc/grpc.go
@@ -1,4 +1,5 @@
-package committee
+// Package grpc provides nodes grpc connection utilities.
+package grpc
 
 import (
 	"context"
@@ -21,6 +22,7 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
 	"github.com/oasisprotocol/oasis-core/go/common/node"
 	"github.com/oasisprotocol/oasis-core/go/common/pubsub"
+	"github.com/oasisprotocol/oasis-core/go/runtime/nodes"
 )
 
 const (
@@ -31,6 +33,10 @@ const (
 	// Default.
 	grpcMinConnectTimeout = 20 * time.Second
 )
+
+// ErrNotVersionedWatcher is an error returned when versioned specific methods are called
+// for a non-versioned client.
+var ErrNotVersionedWatcher = fmt.Errorf("client watcher is not versioned")
 
 // NodeSelectionFeedback is feedback to the node selection policy.
 type NodeSelectionFeedback struct {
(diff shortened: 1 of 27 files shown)


go/runtime/nodes/grpc/grpc.go
```diff
@@ -435,61 +447,119 @@ func (cc *committeeClient) worker(ctx context.Context, ch <-chan *NodeUpdate, su
...
+						nc.deleteConnectionLocked(id)
 					}
 				case u.Freeze != nil:
-					// Committee has been frozen.
+					if !nc.nw.Versioned() {
+						nc.logger.Warn("versioned event for non-versioned node watcher",
+							"event", u,
+						)
+						return
+					}
```
kostko, line 478: Will this cause the client to silently stop functioning? This seems like a developer error, maybe it deserves a panic?

go/runtime/nodes/nodes.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/oasisprotocol/oasis-core/go/common/pubsub"
+)
+
+// NodeUpdate is a node update.
+type NodeUpdate struct {
+	Update      *node.Node
+	Delete      *signature.PublicKey
+	Reset       bool
+	Freeze      *VersionEvent
+	BumpVersion *VersionEvent
```
kostko, line 16: Maybe split into two groups, for versioned and non-versioned?

go/runtime/nodes/grpc/grpc.go
ptrus: It won't stop functioning since the `return` here is only to exit the closure (which is used so we can `defer` unlock).

But yeah this indeed is a developer error so I was also thinking about panicking here instead.

go/runtime/nodes/vresioned.go
```diff
@@ -1,4 +1,4 @@
-package committee
+package nodes
```
kostko, line 1: This file's name has a typo (`vresioned.go`), should be `versioned.go`.

go/runtime/nodes/vresioned.go
```diff
@@ -280,15 +247,24 @@ func (nw *nodeDescriptorWatcher) worker(ch <-chan *registry.NodeEvent, sub pubsu
...
+func (nw *versionedNodeDescriptorWatcher) Versioned() bool {
+	// This watcher supports versions.
+	return true
+}
+
+// NewBaseVersionedNodeDescriptorWatcher creates a new base versioned node descriptor watcher.
+//
+// The base watcher will only track nodes that will be explicitly marked to watch
+// via WatchNode/WatchNodeWithTags methods.
+// TODO: better name?
```
kostko, line 259: Is there a non-base version? Maybe just remove the "Base"?

go/worker/storage/committee/node.go
```diff
@@ -258,29 +263,53 @@ func NewNode(
...
 	// Create a new storage client that will be used for remote sync.
-	scl, err := client.New(
-		node.ctx,
-		commonNode.Runtime.ID(),
-		node.commonNode.Identity,
-		node.commonNode.Consensus.Scheduler(),
-		node.commonNode.Consensus.Registry(),
-		nil,
-		runtimeCommittee.WithFilter(runtimeCommittee.IgnoreNodeFilter(commonNode.Identity.NodeSigner.Public())),
+	// This storage client connects to all registered nodes for the committee.
```
kostko, line 276:
```suggestion
	// This storage client connects to all registered storage nodes for the runtime.
```

kostko (commented):
Nice work, mostly some nits, otherwise looks pretty good!

go/runtime/nodes/vresioned.go
ptrus: Yeah i agree that `VersionedNodeDescriptorWatcher` is better without `Base`. I thought about including something in the name that indicates that this watcher doesn't actually watch anything, and kinda serves as a "base" for a concrete watcher (e.g. the committee watcher). 
But now i also added the docstring that this mentions that this watcher only tracks nodes explicitly marked, which i think is good enough and can drop the `Base` from the name.

go/runtime/nodes/grpc/grpc.go
ptrus: I'm not sure if we should panic here or not. On one hand we probably should since this should never happen unless it's a development error, on the other hand if this error does happen, this will generally still work ok.

Probably should panic nonetheless.

go/runtime/nodes/grpc/grpc.go
```diff
@@ -435,61 +458,105 @@ func (cc *committeeClient) worker(ctx context.Context, ch <-chan *NodeUpdate, su
...
+					// Versioned node watcher - version freeze.
 					var nodes []signature.PublicKey
-					for id := range cc.conns {
+					for id := range nc.conns {
 						nodes = append(nodes, id)
 					}
-					cc.nodeSelectionPolicy.UpdateNodes(nodes)
+					nc.nodeSelectionPolicy.UpdateNodes(nodes)
```
kostko, line 495: Should this use `nc.updateNodeSelectionPolicyLocked`?

kostko (approved):

go/runtime/nodes/grpc/grpc.go
ptrus: I'll leave it as a warning for now

PR #3459, opened 2020-10-30
fixup! go/common/crypto: Bump ed25519 version

pro-wh:
Another bump, update key blacklisting test to use a valid public point


kostko:
Thanks, merged into #3458.

PR #3461, opened 2020-11-02
[BACKPORT/20.12.x] go/common/crypto: Bump ed25519 version

kostko:
Backport of #3458 and #3460.


pro-wh (approved):
We're bumping the consensus version. With this change, public key "validity" checks no longer check against a list of small order points. These are now considered valid, but with signatures for those keys always counted as invalid.

ptrus (approved):

PR #3467, opened 2020-11-03
changelog: Assemble changes for 20.12 release

tjanez:


kostko (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,74 @@ The format is inspired by [Keep a Changelog].
 
 <!-- TOWNCRIER -->
 
+## 20.12 (2020-11-03)
+
+| Protocol          | Version   |
+|:------------------|:---------:|
+| Consensus         | 1.1.0     |
```
ptrus, line 19: Wondering if there should be an additional big note somewhere that this release is NOT compatible with previous versions.

ptrus (approved):

CHANGELOG.md
tjanez: Did you mean to add something to the Change Log?

Or to our general docs when we post the upgrade instructions?

ptrus: I meant that maybe there should be a note in the change log as well (since it's a bit hard to notice currently, specially since we only bumped the minor version). But yeah I'm not totally convinced if it's needed - just a thought.

tjanez: Agreed. Let's bump the consensus' major version so we'll have 2.0.0 and it should be clear that this is incompatible with previous versions.

kostko (approved):

PR #3489, opened 2020-11-06
Update master branch for 20.12 release

tjanez:
Since the 20.12 release was done on the `stable/20.12.x` branch which was branched form the `stable/20.11.x` branch, we need to update the `master` branch for these changes to allow creating the 20.13 release on `master`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,76 @@ The format is inspired by [Keep a Changelog].
 
 <!-- TOWNCRIER -->
 
+## 20.12 (2020-11-03)
+
+| Protocol          | Version   |
+|:------------------|:---------:|
+| Consensus         | 2.0.0     |
+| Runtime Host      | 1.0.0     |
+| Runtime Committee | 1.0.0     |
+
+### Removals and Breaking Changes
+
+- go/common/crypto: Bump ed25519 version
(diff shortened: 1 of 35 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,207 @@ The format is inspired by [Keep a Changelog].
...
+  node registered mid-epoch.
+
+- rust: Bump futures to 0.3.7
+  ([#3460](https://github.com/oasisprotocol/oasis-core/issues/3460))
+
+  Fixes [RUSTSEC-2020-0059].
+
+  [RUSTSEC-2020-0059]: https://rustsec.org/advisories/RUSTSEC-2020-0059
+
+## 20.11.3 (2020-10-16)
```
kostko, line 85: The patch versions from 20.11.x should probably not be included in the changelog?

tjanez: I had similar thoughts at the beginning, but now I think it would be cleanest to just copy the entire Change Log changes from the 20.12 release, which include separate sections for 20.11.x releases.

Collapsing them under 20.12 now feels weird since 20.12 release wasn't done on the `master` branch and was actually branched off the `stable/20.11.x` branch.

kostko: Ok.

kostko (approved):

Issue #3431, opened 2020-10-19
Implement ADR 0003


kostko:
The implementation of [ADR 0003](https://github.com/oasisprotocol/oasis-core/blob/master/docs/adr/0003-consensus-runtime-token-transfer.md) is split into the following tasks:
* [x] [Support setting allowance](https://github.com/oasisprotocol/oasis-core/issues/3428)
* [x] [Add runtime accounts](https://github.com/oasisprotocol/oasis-core/issues/3429)
* [x] [Refactor runtime messages](https://github.com/oasisprotocol/oasis-core/issues/3430)
* [x] [Transfer between staking and runtime ledgers](https://github.com/oasisprotocol/oasis-core/issues/2077)
* [ ] [Basic consensus layer state accessors for Rust](https://github.com/oasisprotocol/oasis-core/issues/3564)

PR #3511, opened 2020-11-12
Add runtime accounts and staking runtime messages

### LINKED ISSUE #2077: Transfer between staking and runtime ledgers
As the staking ledger should be used exclusively for staking, we need to support transfer of tokens between the staking ledger (whcih lives in the BFT committee) and the ledger(s) maintained by the runtime(s). The runtime ledgers have runtime-specific semantics.

The transfer operation(s) would consist of two parts:
* Update of the staking ledger (can only be done by the BFT committee, e.g., the roothash ABCI application), which happens by updating the ABCI application state.
* Update of the runtime ledger (can only be done by the runtime as the ledger has runtime-specific semantics), which happens by updating storage.

Care must be taken to avoid any possibility of losing or duplicating funds (e.g., the operation(s) must be atomic).

Notes:
* The runtime probably needs some way to query staking balances. Care must be taken as the staking accounts can change while the runtime is executing.
* The runtime needs to signal the requested change in the staking account(s) to the BFT committee. Related to the first note, while the BFT committee can (and should) verify that there is enough balance in the staking account, the only thing it can do to prevent breaking atomicity is to abort the whole round (and thus many unrelated batches) as it has no way of reverting only specific updates to storage. Care must be taken to not turn this into a DoS vector.
* Perhaps there is a way to make these transfers multi-phase where the staking account is first locked and then transfers are performed knowing that the balance cannot change while the runtime is executing transactions. Again care must be taken to not turn any such locking scheme into a DoS vector.
* To prevent spamming only owners of staking accounts should be able to initiate such transfers. This means that the runtime needs to provide a valid authorization signature for the given staking account to the BFT committee (this signature would be provided by stakeholder, e.g., embedded in a regular transaction).

Probably out of scope:
* ~~With this feature in place, a badly written runtime can cause funds to appear from thin air and transfer them to the staking ledger which can potentially compromise the rest of the system. This probably cannot be avoided.~~ (see #2302)

### LINKED ISSUE #3429: Add runtime accounts
Runtime accounts should be added to the staking service as described by [ADR 0003](https://github.com/oasisprotocol/oasis-core/blob/master/docs/adr/0003-consensus-runtime-token-transfer.md).

kostko:
Fixes #3429 
Fixes #2077 

TODO
* [X] Tests for Rust Quantity and Address types
* [x] Gas accounting during proposer executor commit
* [x] Dispatch and process staking runtime messages
* [x] Update RHP as per ADR 0003
* [x] Consider moving `go/roothash/block/messages.go` to `go/roothash/messages`.
* [x] Consider moving consensus-layer stuff under `oasis_core_runtime::consensus::*` (currently it's directly under common). This includes address, registry, roothash, staking.
* [x] Maybe remove Noop message.
* [x] Update documentation.
* [x] Update test vectors for new transaction types (allow/withdraw).


runtime/src/common/quantity.rs
```diff
@@ -0,0 +1,75 @@
+//! An arbitrary precision unsigned integer.
+use std::fmt;
+
+use num_bigint::BigUint;
+use num_traits::{Num, Zero};
+use serde;
+
+/// An arbitrary precision unsigned integer.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash)]
+pub struct Quantity(BigUint);
```
ptrus, line 10: Should probably add some test cases for quantity serialization/deserialization to ensure it matches go side?

ptrus (commented):
checked first 4 commits (will continue with review later)

runtime/src/common/quantity.rs
kostko: Good idea, added.

go/oasis-node/cmd/debug/txsource/workload/oversized.go
```diff
@@ -10,37 +10,45 @@ import (
...
 	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"
 	consensusGenesis "github.com/oasisprotocol/oasis-core/go/consensus/genesis"
 	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
 )
 
-const (
-	// NameOversized is the name of the oversized workload.
-	NameOversized = "oversized"
+// NameOversized is the name of the oversized workload.
+const NameOversized = "oversized"
```
ptrus, line 20: I think this part

https://github.com/oasisprotocol/oasis-core/blob/b59d6b3734f45e63b6da497746c386269b51b2ab/go/oasis-node/cmd/debug/txsource/workload/oversized.go#L127-L131

should now be removed from the workload as the disable checktx flag was removed.

ptrus (commented):
checked up to: "go/roothash/api: Remove NoopMessage runtime message"

.changelog/3511.feature.md
```diff
@@ -0,0 +1,3 @@
+go/consensus/api: Add Registry/Staking/Scheduler to ClientBackend
```
tjanez, line 1: Nit: Change the title to:
```
go/consensus/api: Add `Registry()`/`Staking()`/`Scheduler()` to client backend
```

(Ideally, I would use `` `ClientBackend` ``, but then the title would be too long.)

tjanez (commented):
I review up to commit "go/consensus/api: Add Registry/Staking/Scheduler to ClientBackend"

tjanez (commented):
Some more comments. Reviewed up to commit "runtime: Move consensus layer structures to its own module"

go/runtime/host/protocol/types.go
```diff
@@ -119,6 +120,12 @@ type Error struct {
 type RuntimeInfoRequest struct {
 	// RuntimeID is the assigned runtime ID of the loaded runtime.
 	RuntimeID common.Namespace `json:"runtime_id"`
+
+	// ConsensusBackend is the consensus backend that is in use for the consensus layer.
```
ptrus, line 124:
```suggestion
	// ConsensusBackend is the consensus backend name that is in use for the consensus layer.
```

go/worker/common/committee/runtime_host.go
```diff
@@ -60,23 +63,35 @@ func (h *computeRuntimeHostHandler) Handle(ctx context.Context, body *protocol.B
...
+		var rs syncer.ReadSyncer
+		switch rq.Endpoint {
+		case protocol.HostStorageEndpointRuntime:
+			// Runtime storage.
+			rs = h.storage
+
+			// Prioritize nodes that signed the last storage receipts.
+			h.node.CrossNode.Lock()
+			blk := h.node.CurrentBlock
+			h.node.CrossNode.Unlock()
```
ptrus, line 75: Was already wrong before, but this lock should be held until after line 78. Should probably just change tihs part to:

```
h.node.CrossNode.Lock()
if blk := h.node.CurrentBlock; blk != nil {
...
}
h.node.CrossNode.Unlock()
```

go/worker/compute/executor/committee/node.go
```diff
@@ -569,16 +569,18 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
 func (n *Node) checkTx(ctx context.Context, tx []byte) error {
 	n.commonNode.CrossNode.Lock()
 	currentBlock := n.commonNode.CurrentBlock
+	currentConsensusBlock := n.commonNode.CurrentConsensusBlock
 	n.commonNode.CrossNode.Unlock()
```
ptrus, line 573: Same here, this lock should be held longer (my fault here probably :grin:).

tjanez (commented):
Overall, looks very good. I've only had a couple of smaller comments.

One thing that would also be useful would be to add the:

- `oasis-node stake accoutn gen_allow`
- `oasis-node stake accoutn gen_withdraw`

CLI commands that would generate the `staking.Allow` and `staking.Withdraw` transactions.

go/roothash/api/message/message.go
```diff
@@ -42,3 +47,25 @@ type NoopMessage struct {
...
+// StakingMessage is a runtime message that allows a runtime to perform staking operations.
+type StakingMessage struct {
+	cbor.Versioned
+
+	Transfer *staking.Transfer `json:"transfer,omitempty"`
+	Withdraw *staking.Withdraw `json:"withdraw,omitempty"`
+}
+
+// ValidateBasic performs basic validation of the runtime message.
+func (sm *StakingMessage) ValidateBasic() error {
```
ptrus, line 46: Should this also validate that both `Transfer` & `Withdraw` are not set? (since this is assumed by rest of the code).

go/worker/common/committee/runtime_host.go
kostko: Why would the lock need to be held there? The part after the unlock is only accessing `blk` (and not `h.node.CurrentBlock`) which has already been stored in a local variable and cannot change?

ptrus: Ah right, and also the `Block` itself won't ever change (other than be replaced with a different block) so indeed it's fine.

go/worker/compute/executor/committee/node.go
ptrus: not needed here either

go/roothash/api/message/message.go
kostko: Yeah, done.

docs/consensus/staking.md
```diff
@@ -384,8 +411,173 @@ The transaction signer implicitly specifies the escrow account.
...
+
+<!-- markdownlint-disable line-length -->
+[`NewWithdrawTx` function]:
+  https://pkg.go.dev/github.com/oasisprotocol/oasis-core/go/staking/api?tab=doc#NewWithdrawTx
+[`TransferEvent`]: #transfer-event
+<!-- markdownlint-enable line-length -->
+
 ## Events
 
+### Transfer Event
```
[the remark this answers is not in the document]
kostko: Added.

docs/consensus/staking.md
```diff
@@ -384,8 +411,173 @@ The transaction signer implicitly specifies the escrow account.
...
+    Beneficiary  Address           `json:"beneficiary"`
+    Allowance    quantity.Quantity `json:"allowance"`
+    Negative     bool              `json:"negative,omitempty"`
+    AmountChange quantity.Quantity `json:"amount_change"`
+}
+```
+
+**Fields:**
+
+* `allowance` contains the new total allowance.
```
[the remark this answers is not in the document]
kostko: Done.

kostko:
> One thing that would also be useful would be to add the:

Ah yes, I forgot about the CLI subcommands, added with tests, thanks!

runtime/src/consensus/mod.rs
```diff
@@ -0,0 +1,6 @@
+//! Consensus service interfaces.
```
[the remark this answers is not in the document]
kostko: Added.

go/oasis-node/cmd/debug/txsource/workload/oversized.go
kostko: Done.

go/oasis-node/cmd/debug/txsource/workload/oversized.go
```diff
@@ -98,34 +108,30 @@ func (oversized) Run(
 		if err != nil {
 			return fmt.Errorf("transaction.Sign: %w", err)
 		}
-		oversizedLogger.Debug("submitting oversized transaction",
+		o.Logger.Debug("submitting oversized transaction",
 			"payload_size", len(xfer.Data),
 		)
 
 		// Wait for a maximum of 5 seconds as submission may block forever in case the client node
 		// is skipping all CheckTx checks.
```
ptrus, line 116: Outdated comment

ptrus (approved):

runtime/src/protocol.rs
```diff
@@ -252,8 +252,19 @@ impl Protocol {
...
+                runtime_id,
+                consensus_backend,
+                consensus_protocol_version,
+            } => {
+                info!(self.logger, "Received host environment information";
+                    "runtime_id" => ?runtime_id,
+                    "consensus_backend" => consensus_backend,
+                    "consensus_protocol_version" => consensus_protocol_version,
+                );
+                // TODO: Verify consensus backend/protocol compatibility.
```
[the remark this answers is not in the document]
kostko: Will be part of the last item in #3431, I just filed #3564.

go/oasis-node/cmd/debug/txsource/workload/oversized.go
kostko: Fixed.

go/runtime/host/protocol/connection.go
```diff
@@ -114,6 +114,13 @@ type Connection interface {
 	InitGuest(ctx context.Context, conn net.Conn) error
 }
 
+// HostInfo contains the information about the host environment that is sent to the runtime during
+// connection initialization.
+type HostInfo struct {
+	ConsensusBackend         string
```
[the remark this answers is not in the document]
kostko: Added.

runtime/src/protocol.rs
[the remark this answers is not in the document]
tjanez: Thanks!

go/oasis-test-runner/scenario/e2e/stake_cli.go
```diff
@@ -569,6 +592,63 @@ func (sc *stakeCLIImpl) testAmendCommissionSchedule(childEnv *env.Env, cli *cli.
 	return nil
 }
 
+// testAllowWithdraw tests setting an allowance and withdrawing.
+func (sc *stakeCLIImpl) testAllowWithdraw(childEnv *env.Env, cli *cli.Helpers, src, beneficiary api.Address, beneficiaryEntityDir string) error {
+	var srcNonce uint64 = 5
```
tjanez, line 597: It would be nice to not hard-code the nonces. Something like:

```suggestion
	srcNonce, err := sc.getAccountNonce(childEnv, src)
	if err != nil {
		return fmt.Errorf("getAccountNonce for source account %s: %w", src, err)
	}
	beneficiaryNonce, err := sc.getAccountNonce(childEnv, beneficiary)
	if err != nil {
		return fmt.Errorf("getAccountNonce for beneficiary account %s: %w", beneficiary, err)
	}
```

go/oasis-test-runner/scenario/e2e/stake_cli.go
```diff
@@ -569,6 +592,63 @@ func (sc *stakeCLIImpl) testAmendCommissionSchedule(childEnv *env.Env, cli *cli.
...
+		return err
+	}
+
+	if err := cli.Consensus.SubmitTx(allowTxPath); err != nil {
+		return fmt.Errorf("failed to submit Allow tx: %w", err)
+	}
+
+	// Withdraw.
+	withdrawTxPath := filepath.Join(childEnv.Dir(), "stake_withdraw.json")
+	if err := sc.genWithdrawTx(childEnv, withdrawAmount, 0, src, beneficiaryEntityDir, withdrawTxPath); err != nil {
```
tjanez, line 615:
```suggestion
	if err := sc.genWithdrawTx(childEnv, withdrawAmount, beneficiaryNonce, src, beneficiaryEntityDir, withdrawTxPath); err != nil {
```

tjanez (approved):
Just a small nit, otherwise looks good!

go/consensus/tendermint/api/context.go
```diff
@@ -222,6 +225,17 @@ func (c *Context) WithCallerAddress(callerAddress staking.Address) *Context {
 	return child
 }
 
+// WithSimulation creates a child context in simulation mode.
```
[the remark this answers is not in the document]
tjanez: > Nit: Would it make sense to extract the context and simulation related changes from the "Add staking runtime messages" commit into a separate commit?

I don't think it would add much value. The PR is already an exemplar case on how to structure things into commits.

go/oasis-test-runner/scenario/e2e/stake_cli.go
kostko: Oh I guess this changed while rebasing on master as previousl there was no such thing :-)

tjanez: He he, indeed, I added this in #3559 :slightly_smiling_face:.

PR #3528, opened 2020-11-20
go/consensus/tendermint: Bump Tendermint Core to v0.34.0-oasis1

### LINKED ISSUE #3402: Include all protocol versions in Change Log version field
Currently we specify  Consensus, Runtime Host and Runtime Committee versions of a release in the Change Log. We should also include at least Tendermint (and ABCI) versions, as two releases could use same protocol versions but different Tendermint version resulting in incompatible releases. Probably doesn't hurt also including go toolchain version for completeness.

kostko:
Closes #3402.

### DIFF
--- a/go/common/version/version.go
+++ b/go/common/version/version.go
@@ -9,8 +9,6 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
-
-	"github.com/tendermint/tendermint/version"
 )
 
 // NOTE: This should be kept in sync with runtime/src/common/version.rs.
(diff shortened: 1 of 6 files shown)


go/common/version/version.go
```diff
@@ -93,12 +91,6 @@ var (
 	// https://github.com/tendermint/tendermint/blob/1635d1339c73ae6a82e062cd2dc7191b029efa14/state/validation.go#L21-L22.
 	TendermintAppVersion = ConsensusProtocol.MaskNonMajor().ToU64()
 
-	// Tendermint exposes the tendermint core version.
-	Tendermint = parseSemVerStr(version.TMCoreSemVer)
```
ptrus, line 97: This makes https://github.com/oasisprotocol/oasis-core/issues/3402 obsolete, feel free to close that issue

ptrus (approved):

go/common/version/version.go
kostko: Yeah the reason is that the Tendermint Core version is no longer included as it is added at build time.

Yawning (approved):

Issue #3502, opened 2020-11-10
Implement Consensus Governance backend (ADR 0006)


ptrus:
Implement consensus governance backend as described in ADR 0006.

PR #3536, opened 2020-11-25
Implement consensus governance backend

ptrus:
Closes: https://github.com/oasisprotocol/oasis-core/issues/3502

Adds initial consensus governance backend implementation as described in [ADR-0006](https://github.com/oasisprotocol/oasis-core/blob/master/docs/adr/0006-consensus-governance.md).

TODO: address remaining todo's, cleanups, add E2E test


go/consensus/tendermint/apps/governance/governance.go
```diff
@@ -0,0 +1,406 @@
...
+		}
+		validatorEntitiesEscrow[entityAddr] = escrow
+		if err := totalVotingStake.Add(escrow); err != nil {
+			return nil, nil, fmt.Errorf("failed to add to totalVotingStake: %w", err)
+		}
+	}
+	return totalVotingStake, validatorEntitiesEscrow, nil
+}
+
+func (app *governanceApplication) closeProposal(
```
kostko, line 216: Add a comment that this may modify the proposal in case an error happens.

kostko (commented):
Don't forget to also update the documentation with the new service.

(Review still in progress.)

go/common/keyformat/key_format.go
```diff
@@ -4,6 +4,8 @@ import (
 	"encoding"
 	"encoding/binary"
 	"fmt"
+
+	epochtime "github.com/oasisprotocol/oasis-core/go/epochtime/api"
```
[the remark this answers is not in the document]
ptrus: yeah i guess we can just use `uint64`s

go/consensus/tendermint/apps/governance/governance.go
```diff
@@ -0,0 +1,406 @@
...
+	// Check if a pending upgrade is scheduled for current epoch.
+	state := governanceState.NewMutableState(ctx.State())
+	pendingUpgrades, err := state.PendingUpgrades(ctx)
+	if err != nil {
+		return fmt.Errorf("tendermint/governance: couldn't get pending upgrades: %w", err)
+	}
+	var upgrade *upgrade.Descriptor
+	for _, pendingUpgrade := range pendingUpgrades {
+		if pendingUpgrade.Epoch == epoch {
+			upgrade = pendingUpgrade
```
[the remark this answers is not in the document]
ptrus: I think that's actually already implemented*, as i currently implemented it based on the ADR which states:

> The set of pending upgrades is checked to make sure that no upgrades are currently pending. If there is an existing upgrade pending the upgrade proposal fails.

However thinking about it more during implementation and now, i'm actually not sure if there's any reason to not allow multiple pending upgrades (other than limit to one pending upgrade in the same epoch).


(*technically there was a todo in the sanity check for it, at the time of the comment though)

go/common/keyformat/key_format.go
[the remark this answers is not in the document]
kostko: Yeah let's do that instead.

go/consensus/tendermint/apps/governance/governance.go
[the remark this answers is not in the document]
kostko: > However thinking about it more during implementation and now, i'm actually not sure if there's any reason to not allow multiple pending upgrades (other than limit to one pending upgrade on the same epoch).

Yeah we could just allow multiple upgrades as long as there is at least some time inbetween. Maybe relax the check when processing proposals to make sure that there is at least some "min upgrade difference" (in epochs) between upgrades?

go/genesis/genesis_test.go
```diff
@@ -695,39 +709,40 @@ func TestGenesisSanityCheck(t *testing.T) {
...
 
-	d = *testDoc
+	d = testDoc()
 	d.Staking.Ledger[stakingTests.DebugStateSrcAddress].General.Balance = *quantity.NewFromUint64(100)
 	require.Error(d.SanityCheck(), "invalid general balance should be rejected")
 
-	d = *testDoc
-	d.Staking.Ledger[stakingTests.DebugStateSrcAddress].Escrow.Active.Balance = *quantity.NewFromUint64(100)
-	require.Error(d.SanityCheck(), "invalid escrow active balance should be rejected")
+	// TODO: this doesn't actually fail, but it failed before because above mutation was propagated.
```
kostko, line 728: Hm why doesn't this fail, is there a bug in the sanity checker?

go/consensus/tendermint/apps/staking/state/state.go
```diff
@@ -424,8 +428,8 @@ func (s *ImmutableState) Slashing(ctx context.Context) (map[staking.SlashReason]
 	return params.Slashing, nil
 }
 
-func (s *ImmutableState) LastBlockFees(ctx context.Context) (*quantity.Quantity, error) {
-	value, err := s.is.Get(ctx, lastBlockFeesKeyFmt.Encode())
+func (s *ImmutableState) loadStoredBalance(ctx context.Context, key *keyformat.KeyFormat) (*quantity.Quantity, error) {
```
kostko, line 431: Should this also be used for `CommonPool` and `TotalSupply`?

go/consensus/tendermint/apps/governance/transactions.go
```diff
@@ -0,0 +1,282 @@
...
+	registryAPI "github.com/oasisprotocol/oasis-core/go/registry/api"
+	stakingAPI "github.com/oasisprotocol/oasis-core/go/staking/api"
+)
+
+func (app *governanceApplication) submitProposal(
+	ctx *api.Context,
+	state *governanceState.MutableState,
+	proposalContent *governance.ProposalContent,
+) error {
+	// TODO: could do more in CheckTx.
```
kostko, line 23: Yeah basic validation is stateless so it could be done in CheckTx.

go/consensus/tendermint/apps/governance/state/state.go
```diff
@@ -0,0 +1,358 @@
...
+		if !pendingUpgradesKeyFmt.Decode(it.Key(), &epoch, &proposalID) {
+			break
+		}
+		// Load the proposal.
+		proposal, err := s.Proposal(ctx, proposalID)
+		if err != nil {
+			return nil, err
+		}
+		if proposal.Content.Upgrade == nil {
+			// TODO: this should never happen, panic?
```
kostko, line 228: You could wrap the error in `api.UnavailableStateError(err)` which will trigger a panic.

go/consensus/tendermint/apps/governance/state/state.go
```diff
@@ -0,0 +1,358 @@
...
+// ActiveProposals looks up all active proposals.
+func (s *ImmutableState) ActiveProposals(ctx context.Context) ([]*governance.Proposal, error) {
+	it := s.is.NewIterator(ctx)
+	defer it.Close()
+
+	var proposals []*governance.Proposal
+	for it.Seek(activeProposalsKeyFmt.Encode()); it.Valid(); it.Next() {
+		var epoch epochtime.EpochTime
+		var proposalID uint64
+		// || !bytes.Equal(it.Value(), activeProposalValue)
```
kostko, line 114: What's this about?

kostko:
Also don't forget to add test vectors for the new transaction types (see `go/staking/gen_vectors` for an example).

go/governance/tests/tester.go
```diff
@@ -0,0 +1,484 @@
...
+	require.NoError(regErr, "WatchEvents")
+	defer regSub.Close()
+
+	// Advance to the epoch before the proposal closes.
+	// XXX: Without this additional epoch transition to the epoch before voting
+	// period ends, the validator node would not be registered and a consensus
+	// panic would be triggered. This can happen since unit tests are using the
+	// DebugStaticValidators option.
+	timeSource := consensus.EpochTime().(epochtime.SetableBackend)
+	err = timeSource.SetEpoch(ctx, testState.proposalCloseEpoch-1)
```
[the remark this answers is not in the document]
ptrus: Yeah agreed and makes sense (had to already do the same in the e2e test (not yet pushed to the PR))

go/consensus/tendermint/apps/governance/governance.go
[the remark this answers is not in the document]
ptrus: So now that i also looked at the existing `upgrade` module i think the initial limit in the ADR for a single pending upgrade comes from the fact that the `upgrade` module currently supports only a single scheduled upgrade.

So i either need to update the `upgrade` module to support multiple scheduled upgrades, or make "governance upgrade watcher that watches upgrades take into account that there can only be a single scheduled upgrade at a time.

go/governance/api/proposal.go
```diff
@@ -0,0 +1,241 @@
...
+// Otherwise the proposal is rejected.
+func (p *Proposal) CloseProposal(totalVotingStake quantity.Quantity, quorum, threshold uint8) error {
+	if p.State != StateActive {
+		return fmt.Errorf("%w: expected: %v, got: %v", errInvalidProposalState, StateActive, p.State)
+	}
+	if p.Results == nil {
+		return fmt.Errorf("%w: results not initialized", errInvalidProposalState)
+	}
+	if totalVotingStake.IsZero() {
+		return fmt.Errorf("%w: total voting stake is zero", errInvalidProposalState)
```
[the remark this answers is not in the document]
ptrus: fixed, thanks! :+1:

go/consensus/tendermint/apps/governance/state/state.go
ptrus: some outdated stuff :-)

go/consensus/tendermint/apps/staking/state/state.go
ptrus: good idea, will update

go/genesis/genesis_test.go
ptrus: looked into this now, it was just that the `100` escrow balance matched the initial (correct) genesis state :-) fixed by setting a different value.

go/common/keyformat/key_format.go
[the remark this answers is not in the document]
ptrus: done

go/upgrade/api/api.go
```diff
@@ -62,28 +62,120 @@ var (
...
+	case UpgradeMethodInternalName:
+		*m = UpgradeMethodInternal
+	default:
+		return fmt.Errorf("invalid upgrade method: %s", string(text))
+	}
+	return nil
+}
+
 // Descriptor describes an upgrade.
 type Descriptor struct {
```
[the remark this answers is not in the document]
ptrus: What would be the use case for it? In case we'd like to use it as an option instead of the current `Identifier` field (so we wouldn't need exact hash matches) that would require some more refactoring of the `upgrade` module (which currently assumes that `UpgradeInternal` method uses the hashes). Probably a new `UpgradeMethod` type?

go/consensus/tendermint/apps/governance/governance.go
ptrus: added a comment that the method modifies the passed proposal

go/upgrade/api/api.go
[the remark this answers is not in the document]
kostko: Yeah so I was thinking that the hash of a binary is not the best thing, for example in a world where you have different implementations. Of course it could be "hash of an externally-defined manifest" as well and that manifest would have the details. So maybe we can leave it as-is and use a different method which defines the identifier as "hash of a manifest".

go/consensus/tendermint/apps/governance/governance.go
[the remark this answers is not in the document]
ptrus: > Maybe relax the check when processing proposals to make sure that there is at least some "min upgrade difference" (in epochs) between upgrades?

Done, governance now supports multiple pending upgrades. The epoch diff between pending upgrades needs to be at least `upgrade_min_epoch_diff` (i think it made sense reusing this existing consensus parameter). 

Also updated the upgrade manager to support multiple pending ugprades.

go/consensus/tendermint/apps/governance/genesis.go
```diff
@@ -0,0 +1,95 @@
...
+			largestProposalID = proposal.ID
+		}
+		switch proposal.State {
+		case governance.StateActive:
+			if err = state.SetActiveProposal(ctx, proposal); err != nil {
+				return fmt.Errorf("tendermint/governance: failed to set active proposal: %w", err)
+			}
+		default:
+			if err = state.SetProposal(ctx, proposal); err != nil {
+				return fmt.Errorf("tendermint/governance: failed to set consensus parameters: %w", err)
```
kostko, line 41:
```suggestion
				return fmt.Errorf("tendermint/governance: failed to set proposal: %w", err)
```

go/staking/api/api.go
```diff
@@ -40,6 +40,12 @@ var (
 		signature.NewPublicKey("1abe11edfeeaccffffffffffffffffffffffffffffffffffffffffffffffffff"),
 	)
 
+	// GovernanceDepositsAddress is the governance deposits address.
+	// This address is reserved to prevent it from being accidentally used in the actual ledger.
+	GovernanceDepositsAddress = NewReservedAddress(
```
kostko, line 45: Update documentation to include the new special address.

go/consensus/tendermint/apps/supplementarysanity/checks.go
```diff
@@ -234,6 +245,51 @@ func checkKeyManager(ctx *abciAPI.Context, now epochtime.EpochTime) error {
...
+
+	params, err := st.ConsensusParameters(ctx)
+	if err != nil {
+		return fmt.Errorf("ConsensusParameters: %w", err)
+	}
+	err = params.SanityCheck()
+	if err != nil {
+		return fmt.Errorf("SanityCheck ConsensusParameters: %w", err)
+	}
+	// Sanity check proposals.
```
kostko, line 259: There should be a sanity check somewhere that the amount in the governance deposits address matches the total deposits of all active proposals.

go/control/api/api.go
```diff
@@ -68,6 +69,9 @@ type Status struct {
 
 	// Registration is the node's registration status.
 	Registration RegistrationStatus `json:"registration"`
+
+	// PendingUpgrades are the node's pending upgrades.
+	PendingUpgrades []*upgrade.PendingUpgrade `json:"pending_upgrades"`
```
kostko, line 74: :+1:

go/control/api/api.go
```diff
@@ -45,8 +45,9 @@ type NodeController interface {
 	// and shut down.
 	UpgradeBinary(ctx context.Context, descriptor *upgrade.Descriptor) error
 
-	// CancelUpgrade cancels a pending upgrade, unless it is already in progress.
-	CancelUpgrade(ctx context.Context) error
+	// CancelUpgrade cancels the specific pending upgrade, unless it is already in progress.
+	// TOOD: currently pending upgrade name uniqueness is not enforced.
+	CancelUpgrade(ctx context.Context, name string) error
```
kostko, line 50: Make sure to flag this as a breaking gRPC API change.

go/epochtime/api/api.go
```diff
@@ -19,6 +19,14 @@ type EpochTime uint64
 // EpochInvalid is the placeholder invalid epoch.
 const EpochInvalid EpochTime = 0xffffffffffffffff // ~50 quadrillion years away.
 
+// Diff returns the difference (in epochs) between two epochtimes.
```
kostko, line 22:
```suggestion
// Diff returns the absolute difference (in epochs) between two epochtimes.
```

go/governance/api/sanity_check.go
```diff
@@ -0,0 +1,146 @@
...
+	epochtime "github.com/oasisprotocol/oasis-core/go/epochtime/api"
+	upgrade "github.com/oasisprotocol/oasis-core/go/upgrade/api"
+)
+
+// SanityCheck performs a sanity check on the consensus parameters.
+func (p *ConsensusParameters) SanityCheck() error {
+	if !p.MinProposalDeposit.IsValid() {
+		return fmt.Errorf("min_proposal_deposit has invalid value")
+	}
+	// Quorum must be less than 100.
```
kostko, line 16: Or equal to? The check is fine, just the comment needs to be fixed. (Also below.)

go/oasis-test-runner/scenario/e2e/governance_upgrade.go
```diff
@@ -0,0 +1,379 @@
...
+
+var (
+	// GovernanceConsensusUpgrade is the governance consensus upgrade scenario.
+	GovernanceConsensusUpgrade scenario.Scenario = newGovernanceConsensusUpgradeImpl(true)
+	// GovernanceConsensusFailUpgrade is the governance consensus upgrade scenario where
+	// node should fail the upgrade.
+	GovernanceConsensusFailUpgrade scenario.Scenario = newGovernanceConsensusUpgradeImpl(false)
+	// TODO: cancel upgrade test case.
+	// TODO: governance state in genesis.
+)
```
kostko, line 34: Would be cool to also have a test with runtimes surviving a consensus layer upgrade.

kostko (commented):
Hm, maybe we should avoid using the hash of the binary to check for upgrades and use something like a software version instead? Otherwise people doing custom builds can easily fail the check?

ptrus:
> Hm, maybe we should avoid using the hash of the binary to check for upgrades and use something like a software version instead? Otherwise people doing custom builds can easily fail the check?

yeah agreed, probably best to open a follow-up issue and do it after this is merged, as this mostly means updates to the existing `upgrade` module (probably just changing the existing internal method to not use the hash). Shouldn't affect the governance code at all.

kostko:
> yeah agreed, probably best to open a follow-up issue and do it after this is merged

:+1:

go/oasis-test-runner/scenario/e2e/governance_upgrade.go
ptrus: Done, i moved the tests under `e2e/runtime/` and the tests now also exercise the runtime before and after the upgrade.

go/governance/api/sanity_check.go
ptrus: Fixed

go/governance/api/sanity_check.go
```diff
@@ -0,0 +1,118 @@
...
+		if p.CreatedAt > epoch {
+			return fmt.Errorf("proposal %v: with crated epoch in future", p.ID)
+		}
+		if !p.Submitter.IsValid() {
+			return fmt.Errorf("proposal %v: invalid proposal submitter", p.ID)
+		}
+		if err := p.Content.ValidateBasic(); err != nil {
+			return fmt.Errorf("proposal %v: basic validation failure: %w", p.ID, err)
+		}
+		if p.Content.Upgrade != nil {
```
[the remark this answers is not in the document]
ptrus: yeah makes sense, did so

go/consensus/tendermint/apps/supplementarysanity/checks.go
ptrus: good idea! added it to the sanity checks (it's checked both in genesis and in the supplementary checker)

go/governance/tests/tester.go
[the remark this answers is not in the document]
ptrus: Done. I fixed the `MustAdvanceEpoch` helper function to now increase the epochs one by one so this is now fixed generally across all tests (there might be some direct invocations to SetEpoch, but those can be fixed separately).

go/control/api/api.go
ptrus: noted in 3536.breaking.md

go/worker/upgrade/worker.go
```diff
@@ -0,0 +1,195 @@
...
+				)
+				return
+			}
+			w.logger.Debug("scheduled pending upgrades",
+				"gov_pending_upgrades", govUpgradeDescriptors,
+				"node_pending_upgrades", nodePendingUpgrades,
+			)
+
+			// Make sure all governance pending upgrades are scheduled.
+			for _, descriptor := range govUpgradeDescriptors {
```
ptrus, line 112: self todo: check why this is reported as not covered, although it's exercised in the governance-upgrade tests (confimed via logs)

ptrus: ah it's due to https://github.com/oasisprotocol/oasis-core/issues/3574 since nodes exit uncleanly for the upgrade (and after the upgrade is done, there's no other scheduled upgrade), should just defer the fix for that issue.

go/control/api/api.go
```diff
@@ -45,8 +45,9 @@ type NodeController interface {
 	// and shut down.
 	UpgradeBinary(ctx context.Context, descriptor *upgrade.Descriptor) error
 
-	// CancelUpgrade cancels a pending upgrade, unless it is already in progress.
-	CancelUpgrade(ctx context.Context) error
+	// CancelUpgrade cancels the specific pending upgrade, unless it is already in progress.
+	// TODO: currently pending upgrade name uniqueness is not enforced.
```
kostko, line 49: Should we file an issue for this?

go/epochtime/api/api.go
```diff
@@ -19,6 +19,14 @@ type EpochTime uint64
 // EpochInvalid is the placeholder invalid epoch.
 const EpochInvalid EpochTime = 0xffffffffffffffff // ~50 quadrillion years away.
 
+// Diff returns the absolute difference (in epochs) between two epochtimes.
+func (e EpochTime) Diff(other EpochTime) EpochTime {
```
kostko, line 23: Maybe rename to `AbsDiff` to make it more clear.

go/oasis-test-runner/scenario/e2e/e2e.go
```diff
@@ -3,9 +3,13 @@ package e2e
 
 import (
 	"bytes"
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	h "hash"
```
kostko, line 9:
```suggestion
	goHash "hash"
```

kostko (commented):
Cool, looks good!

@jberci could you review the parts related to upgrade descriptors?

kostko (approved):

jberci (approved):

go/control/api/api.go
ptrus: Added a note to https://github.com/oasisprotocol/oasis-core/issues/3579 will do it as part of that.

go/consensus/tendermint/apps/governance/transactions.go
```diff
@@ -0,0 +1,288 @@
...
+	// Validate proposal content basics.
+	if err = proposalContent.ValidateBasic(); err != nil {
+		ctx.Logger().Error("governance: malformed proposal content",
+			"content", proposalContent,
+			"err", err,
+		)
+		return governance.ErrInvalidArgument
+	}
+
+	if ctx.IsCheckOnly() {
```
ptrus, line 42: I moved the `IsCheckOnly()` check for both `submitProposal` and `castVote` methods.  Please take another quick look over this part @kostko 

Actually could probably move the check in `submitProposal` even further bellow probably, not sure how much does it make sense though.

kostko: I would move fetching consensus parameters below (as they are not used in CheckTx and avoidng stateful operations is a good thing), otherwise this looks good.

go/consensus/tendermint/apps/governance/transactions.go
```diff
@@ -0,0 +1,288 @@
...
+		ctx.Logger().Error("governance: vote for a non-active proposal",
+			"proposal_id", proposalVote.ID,
+			"state", proposal.State,
+			"proposal", proposal,
+			"vote", proposalVote,
+		)
+		return governance.ErrVotingIsClosed
+	}
+
+	if ctx.IsCheckOnly() {
```
kostko, line 265: This seems like it is doing a lot in CheckTx, maybe avoid it. If there's nothing stateless that can be checked in CheckTx, feel free to skip those checks.

go/consensus/tendermint/apps/governance/transactions.go
```diff
@@ -0,0 +1,288 @@
...
+		return nil
+	}
+
+	// Charge gas for this transaction.
+	if err = ctx.Gas().UseGas(1, governance.GasOpCastVote, params.GasCosts); err != nil {
```
kostko, line 270: I would move this further up (e.g., after fetching parameters), so gas is charged even when the checks fail.

go/consensus/tendermint/apps/governance/transactions.go
ptrus: ok makes sense yeah. Fixed.

go/consensus/tendermint/apps/governance/transactions.go
ptrus: :+1: was there before just moved it together with check-tx change. Reverted that change.

go/consensus/tendermint/apps/governance/transactions.go
ptrus: done

PR #3538, opened 2020-11-30
Fix storage sync after non-zero genesis

kostko:
* go/storage/mkvs/checkpoint: Add initial version parameter

  Previously if the local database contained a version earlier than the genesis
  version, the checkpointer would attempt to create a new checkpoint at that
  earlier version (and fail). Now the version is clamped at the initial version.
* go/worker/storage: Force checkpoint sync when replication is needed
  
  Previously a freshly initialized storage node with no genesis state would fall
  back to incremental sync even though there was no chance of that succeeding.

### DIFF
--- a/go/worker/storage/committee/node.go
+++ b/go/worker/storage/committee/node.go
@@ -8,6 +8,7 @@ import (
 	"math"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/eapache/channels"
 	"github.com/prometheus/client_golang/prometheus"
@@ -83,6 +84,8 @@ const (
 	RoundLatest = math.MaxUint64
 
 	defaultUndefinedRound = ^uint64(0)
+
+	checkpointSyncRetryDelay = 10 * time.Second
 )
 
 // outstandingMask records which storage roots still need to be synced or need to be retried.
@@ -206,6 +209,7 @@ type Node struct {
 
 	checkpointer           checkpoint.Checkpointer
 	checkpointSyncDisabled bool
+	checkpointSyncForced   bool
 
 	syncedLock  sync.RWMutex
 	syncedState watcherState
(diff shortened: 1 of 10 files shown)


Yawning (approved):

go/worker/storage/committee/node.go
```diff
@@ -871,15 +882,32 @@ func (n *Node) worker() { // nolint: gocyclo
...
-			// then deleted just before we request its chunks). One retry is enough.
-			n.logger.Info("first checkpoint sync failed, trying once more", "err", err)
+	CheckpointSyncRetry:
+		for {
 			summary, err = n.syncCheckpoints()
+			switch n.checkpointSyncForced {
+			case true:
+				// We have no other options but to perform a checkpoint sync as we are missing
+				// genesis state.
+				n.logger.Info("checkpoint sync required as we don't have genesis state, retrying")
```
ptrus, line 894: Should this also log the error returned from ` n.syncCheckpoints`? (which is currently not logged in case `n.checkpointSyncForced=true`).

Also won't this currently just loop forever in case `n.checkpointSyncForced` is true? I don't see anything that ever sets `n.checkpointSyncForced` to false? Should this be within `if err != nil {`?

kostko: Yeah it should be, I'm just adding a test for this.

go/worker/storage/committee/node.go
```diff
@@ -838,6 +849,12 @@ func (n *Node) worker() { // nolint: gocyclo
 		}
 	}
 
+	// Notify the checkpointer of the genesis round so it can be checkpointed.
+	if n.checkpointer != nil {
+		n.checkpointer.NotifyNewVersion(genesisBlock.Header.Round)
+		n.checkpointer.Flush()
```
ptrus, line 855: Could potentially also implement "flushing" within the `NotifyNewVersion` method on the checkpointer itself, and not expose it via a separate method. Unless there is a reason not to flush always when notifying of a new version.

ptrus (approved):

go/worker/storage/committee/node.go
kostko: Yeah the reason is to not do the checkpoint checks on each received block, but rather to batch the checks using a configurable timeout (because there is no sense to check this too often).

ptrus: makes sense :+1:

PR #3548, opened 2020-12-04
Modify runtime descriptors to support ADR 0004

### LINKED ISSUE #3450: Implement ADR 4 (part 1, runtime descriptors)
Runtime descriptors, runtime parameters, and the genesis document structures should be changed as described in https://github.com/oasisprotocol/oasis-core/blob/master/docs/adr/0004-runtime-governance.md.
Additional changes to support the new structures are required in the registry, staking, and scheduler.

Old descriptors are accepted and converted into the new format at genesis time, but all new runtime registrations must use the new descriptor format.

After this issue, as well as issues #3429 and #3430 are complete, the rest of the ADR (messages) can be implemented.  A separate issue will be made for that later.

abukosek:
Closes #3450.

TODO:
- [x] fix e2e tests.
- [x] fix unit tests.
- [x] remove `SignedRuntime` entirely.
- [x] double-check that `oasis-node debug fix-genesis` command does the right thing.
- [x] write a few changelog fragments.
- [x] when 3511 is merged, fix the TODO in `go/consensus/tendermint/apps/registry/transactions.go`.
- [x] update docs.
- [x] add `enable_runtime_governance_models` consensus parameter & tests for it.
- [x] add more tests.

In a separate PR:
- add `update_runtime` runtime message support & e2e tests.

### DIFF
--- a/docs/consensus/registry.md
+++ b/docs/consensus/registry.md
@@ -51,12 +51,25 @@ offline and having entities as separate resources enables that.
 
 A [runtime] is effectively a replicated application with shared state. The
 registry resource describes a runtime's operational parameters, including its
-identifier, kind, admission policy, committee scheduling, storage etc. For a
-full description of the runtime descriptor see [the `Runtime` structure].
+identifier, kind, admission policy, committee scheduling, storage, governance
+model, etc. For a full description of the runtime descriptor see
+[the `Runtime` structure].
 
-Currently only the owning entity is allowed to make any modifications to the
-runtime. There are plans to enable runtimes to update their own descriptors in
-the future to enable runtimes to be self-governing.
+<!-- markdownlint-disable no-space-in-emphasis -->
+The chosen governance model indicates how the runtime descriptor can be updated
+in the future.
+
+There are currently three supported governance models:
+
+* **Entity governance** where the runtime owner is the only one who can update
+  the runtime descriptor via `registry.RegisterRuntime` method calls.
+
+* **Runtime-defined governance** where the runtime itself is the only one who
+  can update the runtime descriptor by emitting a runtime message.
+
+* **Consensus layer governance** where only the consensus layer itself can
+  update the runtime descriptor through network governance.
+<!-- markdownlint-enable no-space-in-emphasis -->
 
 <!-- markdownlint-disable line-length -->
 [runtime]: ../runtime/index.md
@@ -209,16 +222,18 @@ runtime transaction can be generated using [`NewRegisterRuntimeTx`].
 registry.RegisterRuntime
 ```
 
-The body of a register runtime transaction must be a [`SignedRuntime`]
-structure, which is a [signed envelope] containing a [`Runtime`] descriptor.
+The body of a register runtime transaction must be a [`Runtime`] descriptor.
 The signer of the transaction MUST be the owning entity key.
 
-Registering a runtime may require sufficient stake in the owning entity's
-[escrow account].
+Registering a runtime may require sufficient stake in either the owning
+entity's (when entity governance is used) or the runtime's (when runtime
+governance is used) [escrow account].
+
+Changing the governance model from entity governance to runtime governance is
+allowed. Any other governance model changes are not allowed.
 
 <!-- markdownlint-disable line-length -->
 [`NewRegisterRuntimeTx`]: https://pkg.go.dev/github.com/oasisprotocol/oasis-core/go/registry/api?tab=doc#NewRegisterRuntimeTx
-[`SignedRuntime`]: https://pkg.go.dev/github.com/oasisprotocol/oasis-core/go/registry/api?tab=doc#SignedRuntime
 [`Runtime`]: https://pkg.go.dev/github.com/oasisprotocol/oasis-core/go/registry/api?tab=doc#Runtime
 <!-- markdownlint-enable line-length -->
 
--- a/go/consensus/tendermint/apps/registry/genesis.go
+++ b/go/consensus/tendermint/apps/registry/genesis.go
@@ -46,11 +46,11 @@ func (app *registryApplication) InitChain(ctx *abciAPI.Context, request types.Re
 	}
 	// Register runtimes. First key manager and then compute runtime(s).
 	for _, k := range []registry.RuntimeKind{registry.KindKeyManager, registry.KindCompute} {
-		for i, v := range st.Runtimes {
-			if v == nil {
+		for i, rt := range st.Runtimes {
+			if rt == nil {
 				return fmt.Errorf("registry: genesis runtime index %d is nil", i)
 			}
-			rt, err := registry.VerifyRegisterRuntimeArgs(&st.Parameters, ctx.Logger(), v, ctx.IsInitChain(), false)
+			err := registry.VerifyRuntime(&st.Parameters, ctx.Logger(), rt, ctx.IsInitChain(), false)
 			if err != nil {
 				return err
 			}
@@ -58,34 +58,30 @@ func (app *registryApplication) InitChain(ctx *abciAPI.Context, request types.Re
 				continue
 			}
 			ctx.Logger().Debug("InitChain: Registering genesis runtime",
-				"runtime_owner", v.Signature.PublicKey,
+				"runtime_id", rt.ID,
 			)
-			if err := app.registerRuntime(ctx, state, v); err != nil {
+			if err := app.registerRuntime(ctx, state, rt); err != nil {
 				ctx.Logger().Error("InitChain: failed to register runtime",
 					"err", err,
-					"runtime", v,
+					"runtime_id", rt.ID,
 				)
 				return fmt.Errorf("registry: genesis runtime registration failure: %w", err)
 			}
 		}
 	}
-	for i, v := range st.SuspendedRuntimes {
-		if v == nil {
+	for i, rt := range st.SuspendedRuntimes {
+		if rt == nil {
 			return fmt.Errorf("registry: genesis suspended runtime index %d is nil", i)
 		}
 		ctx.Logger().Debug("InitChain: Registering genesis suspended runtime",
-			"runtime_owner", v.Signature.PublicKey,
+			"runtime_id", rt.ID,
 		)
-		if err := app.registerRuntime(ctx, state, v); err != nil {
+		if err := app.registerRuntime(ctx, state, rt); err != nil {
 			ctx.Logger().Error("InitChain: failed to register runtime",
 				"err", err,
-				"runtime", v,
+				"runtime_id", rt.ID,
 			)
 			return fmt.Errorf("registry: genesis suspended runtime registration failure: %w", err)
-		}
-		var rt registry.Runtime
-		if err := cbor.Unmarshal(v.Blob, &rt); err != nil {
-			return fmt.Errorf("registry: malformed genesis suspended runtime: %w", err)
 		}
 		if err := state.SuspendRuntime(ctx, rt.ID); err != nil {
 			return fmt.Errorf("registry: failed to suspend runtime at genesis: %w", err)
@@ -130,7 +126,7 @@ func (rq *registryQuerier) Genesis(ctx context.Context) (*registry.Genesis, erro
 	if err != nil {
 		return nil, err
 	}
-	signedRuntimes, err := rq.state.SignedRuntimes(ctx)
+	runtimes, err := rq.state.Runtimes(ctx)
 	if err != nil {
 		return nil, err
 	}
@@ -177,7 +173,7 @@ func (rq *registryQuerier) Genesis(ctx context.Context) (*registry.Genesis, erro
 	gen := registry.Genesis{
 		Parameters:        *params,
 		Entities:          signedEntities,
-		Runtimes:          signedRuntimes,
+		Runtimes:          runtimes,
 		SuspendedRuntimes: suspendedRuntimes,
 		Nodes:             validatorNodes,
 		NodeStatuses:      nodeStatuses,
--- a/go/consensus/tendermint/apps/registry/state/state.go
+++ b/go/consensus/tendermint/apps/registry/state/state.go
@@ -33,10 +33,10 @@ var (
 	//
 	// Value is empty.
 	signedNodeByEntityKeyFmt = keyformat.New(0x12, keyformat.H(&signature.PublicKey{}), keyformat.H(&signature.PublicKey{}))
-	// signedRuntimeKeyFmt is the key format used for signed runtimes.
+	// runtimeKeyFmt is the key format used for runtimes.
 	//
-	// Value is CBOR-serialized signed runtime.
-	signedRuntimeKeyFmt = keyformat.New(0x13, keyformat.H(&common.Namespace{}))
+	// Value is CBOR-serialized runtime.
+	runtimeKeyFmt = keyformat.New(0x13, keyformat.H(&common.Namespace{}))
 	// nodeByConsAddressKeyFmt is the key format used for the consensus address to
 	// node public key mapping.
 	//
@@ -62,13 +62,13 @@ var (
 	keyMapKeyFmt = keyformat.New(0x17, keyformat.H(&signature.PublicKey{}))
 	// suspendedRuntimeKeyFmt is the key format used for suspended runtimes.
 	//
-	// Value is CBOR-serialized signed runtime.
+	// Value is CBOR-serialized runtime.
 	suspendedRuntimeKeyFmt = keyformat.New(0x18, keyformat.H(&common.Namespace{}))
-	// signedRuntimeByEntityKeyFmt is the key format used for signed runtime by entity
+	// runtimeByEntityKeyFmt is the key format used for runtime by entity
 	// index.
 	//
 	// Value is empty.
-	signedRuntimeByEntityKeyFmt = keyformat.New(0x19, keyformat.H(&signature.PublicKey{}), keyformat.H(&common.Namespace{}))
+	runtimeByEntityKeyFmt = keyformat.New(0x19, keyformat.H(&signature.PublicKey{}), keyformat.H(&common.Namespace{}))
 )
 
 // ImmutableState is the immutable registry state wrapper.
@@ -250,7 +250,7 @@ func (s *ImmutableState) SignedNodes(ctx context.Context) ([]*node.MultiSignedNo
 	return nodes, nil
 }
 
-func (s *ImmutableState) getSignedRuntime(ctx context.Context, keyFmt *keyformat.KeyFormat, id common.Namespace) (*registry.SignedRuntime, error) {
+func (s *ImmutableState) getRuntime(ctx context.Context, keyFmt *keyformat.KeyFormat, id common.Namespace) (*registry.Runtime, error) {
 	raw, err := s.is.Get(ctx, keyFmt.Encode(&id))
 	if err != nil {
 		return nil, abciAPI.UnavailableStateError(err)
@@ -259,20 +259,8 @@ func (s *ImmutableState) getSignedRuntime(ctx context.Context, keyFmt *keyformat
 		return nil, registry.ErrNoSuchRuntime
 	}
 
-	var signedRuntime registry.SignedRuntime
-	if err := cbor.Unmarshal(raw, &signedRuntime); err != nil {
-		return nil, abciAPI.UnavailableStateError(err)
-	}
-	return &signedRuntime, nil
-}
-
-func (s *ImmutableState) getRuntime(ctx context.Context, keyFmt *keyformat.KeyFormat, id common.Namespace) (*registry.Runtime, error) {
-	signedRuntime, err := s.getSignedRuntime(ctx, keyFmt, id)
-	if err != nil {
-		return nil, err
-	}
 	var runtime registry.Runtime
-	if err = cbor.Unmarshal(signedRuntime.Blob, &runtime); err != nil {
+	if err := cbor.Unmarshal(raw, &runtime); err != nil {
 		return nil, abciAPI.UnavailableStateError(err)
 	}
 	return &runtime, nil
@@ -283,7 +271,7 @@ func (s *ImmutableState) getRuntime(ctx context.Context, keyFmt *keyformat.KeyFo
 // This excludes any suspended runtimes, use SuspendedRuntime to query
 // suspended runtimes only.
 func (s *ImmutableState) Runtime(ctx context.Context, id common.Namespace) (*registry.Runtime, error) {
-	return s.getRuntime(ctx, signedRuntimeKeyFmt, id)
+	return s.getRuntime(ctx, runtimeKeyFmt, id)
 }
 
 // SuspendedRuntime looks up a suspended runtime by its identifier and
@@ -301,23 +289,10 @@ func (s *ImmutableState) AnyRuntime(ctx context.Context, id common.Namespace) (r
 	return
 }
 
-// SignedRuntime looks up a (signed) runtime by its identifier and returns it.
-//
-// This excludes any suspended runtimes, use SuspendedSignedRuntime to query
-// suspended runtimes only.
-func (s *ImmutableState) SignedRuntime(ctx context.Context, id common.Namespace) (*registry.SignedRuntime, error) {
-	return s.getSignedRuntime(ctx, signedRuntimeKeyFmt, id)
-}
-
-// SignedSuspendedRuntime looks up a (signed) suspended runtime by its identifier and returns it.
-func (s *ImmutableState) SignedSuspendedRuntime(ctx context.Context, id common.Namespace) (*registry.SignedRuntime, error) {
-	return s.getSignedRuntime(ctx, suspendedRuntimeKeyFmt, id)
-}
-
 func (s *ImmutableState) iterateRuntimes(
 	ctx context.Context,
 	keyFmt *keyformat.KeyFormat,
-	cb func(*registry.SignedRuntime) error,
+	cb func(*registry.Runtime) error,
 ) error {
 	it := s.is.NewIterator(ctx)
 	defer it.Close()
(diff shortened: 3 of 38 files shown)


go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -241,12 +242,71 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
 
 	// Check runtime's whitelist.
 	for _, rt := range paidRuntimes {
-		if rt.AdmissionPolicy.EntityWhitelist != nil && !rt.AdmissionPolicy.EntityWhitelist.Entities[newNode.EntityID] {
-			ctx.Logger().Error("RegisterNode: node's entity not in a runtime's whitelist",
-				"entity", newNode.EntityID,
-				"runtime", rt.ID,
-			)
-			return registry.ErrForbidden
+		if rt.AdmissionPolicy.EntityWhitelist != nil {
```
kostko, line 245: Maybe invert some checks (with continue statements) and get rid of some indentation.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -241,12 +242,71 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
...
+						n, grr := state.Node(ctx, nID)
+						if grr != nil {
+							ctx.Logger().Error("RegisterNode: failed to query node",
+								"err", grr,
+								"node", nID,
+							)
+							return grr
+						}
+
+						if n.HasRoles(role) {
```
kostko, line 297: This should only count nodes that are registering for the same runtime.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -241,12 +242,71 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
...
+						// No such role found in whitelist.
+						return registry.ErrForbidden
+					}
+					if maxNodes == 0 {
+						// No nodes of this type are allowed.
+						return registry.ErrForbidden
+					}
+
+					// Count existing nodes owned by entity.
+					ent, grr := state.Entity(ctx, newNode.EntityID)
```
kostko, line 277: Instead of querying the entity descriptor you should query the `nodeByEntityKeyFmt` (add state accessor, e.g. `GetEntityNodes`, as currently there is only `HasEntityNodes`). Also expired nodes (`n.IsExpired(epoch)`) should not count against the maximum as those are not eligible to be scheduled.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -241,12 +242,71 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
...
+					if grr != nil {
+						ctx.Logger().Error("RegisterNode: failed to query owning entity",
+							"err", grr,
+							"entity", newNode.EntityID,
+						)
+						return grr
+					}
+					// TODO: Do we need to handle entity-signed nodes too?
```
kostko, line 285: Should be automatically handled if you do it like the above.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -573,16 +638,38 @@ func (app *registryApplication) registerRuntime( // nolint: gocyclo
...
+	// Make sure that the entity or runtime has enough stake.
+	// Runtimes using the consensus layer governance model do not require stake.
+	if !params.DebugBypassStake && rt.GovernanceModel != registry.GovernanceConsensus {
 		claim := registry.StakeClaimForRuntime(rt.ID)
 		thresholds := registry.StakeThresholdsForRuntime(rt)
-		acctAddr := staking.NewAddress(rt.EntityID)
+
+		// Figure out which account to use based on the governance model.
+		var acctAddr staking.Address
+		if rt.GovernanceModel == registry.GovernanceEntity {
```
kostko, line 657: Convert to a switch statement.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -573,16 +638,38 @@ func (app *registryApplication) registerRuntime( // nolint: gocyclo
...
+			acctAddr = staking.NewAddress(rt.EntityID)
+		} else if rt.GovernanceModel == registry.GovernanceRuntime {
+			// TODO: I guess we need to wait for #3511 to do this properly :)
+			// We can then probably just use ctx.CallerAddress() in both cases.
```
kostko, line 661: :+1:

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -573,16 +638,38 @@ func (app *registryApplication) registerRuntime( // nolint: gocyclo
...
+			var a signature.PublicKey
+			copy(a[:], rt.ID[:])
+			acctAddr = staking.NewAddress(a)
+		} else {
+			// Basic validation should have caught this, but just in case...
+			ctx.Logger().Error("RegisterRuntime: unknown governance model.")
```
kostko, line 667:
```suggestion
			ctx.Logger().Error("RegisterRuntime: unknown governance model")
```

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -573,16 +638,38 @@ func (app *registryApplication) registerRuntime( // nolint: gocyclo
...
+			return registry.ErrForbidden
```
kostko, line 668: Should probably be `ErrInvalidArgument`.

go/consensus/tendermint/apps/scheduler/scheduler.go
```diff
@@ -407,7 +409,7 @@ func (app *schedulerApplication) electCommittee(
 		return nil
 	}
 
-	nrNodes, wantedNodes := len(nodeList), workerSize+backupSize
+	nrNodes, wantedNodes := len(nodeList), minSize
```
kostko, line 412: That's not what's meant by `MinPoolSize`. There should be an additional check above that fails in case `nrNodes < minSize` (e.g. there should be at least `MinPoolSize` eligible nodes).

go/consensus/tendermint/apps/scheduler/scheduler.go
```diff
@@ -362,7 +362,7 @@ func (app *schedulerApplication) electCommittee(
 		rngCtx       []byte
 		isSuitableFn func(*api.Context, *node.Node, *registry.Runtime) bool
 
-		workerSize, backupSize int
+		workerSize, backupSize, minSize int
```
kostko, line 365:
```suggestion
		workerSize, backupSize, minPoolSize int
```

go/genesis/genesis_test.go
```diff
@@ -178,13 +170,13 @@ func TestGenesisSanityCheck(t *testing.T) {
 		},
 		AdmissionPolicy: registry.RuntimeAdmissionPolicy{
 			EntityWhitelist: &registry.EntityWhitelistRuntimeAdmissionPolicy{
-				Entities: map[signature.PublicKey]bool{
-					validPK: true,
+				Entities: map[signature.PublicKey]registry.EntityWhitelistConfig{
+					validPK: {MaxNodes: make(map[node.RolesMask]uint16)},
```
kostko, line 174: I think a nil map should behave exactly the same so you could do:
```suggestion
					validPK: {},
```

go/oasis-test-runner/scenario/e2e/registry_cli.go
```diff
@@ -631,11 +632,12 @@ func (sc *registryCLIImpl) testRuntime(ctx context.Context, childEnv *env.Env, c
...
 			MaxApplyWriteLogEntries: 10,
 			MaxApplyOps:             11,
+			MinPoolSize:             9,
 		},
 		AdmissionPolicy: registry.RuntimeAdmissionPolicy{
 			EntityWhitelist: &registry.EntityWhitelistRuntimeAdmissionPolicy{
-				Entities: map[signature.PublicKey]bool{
-					testEntity.ID: true,
+				Entities: map[signature.PublicKey]registry.EntityWhitelistConfig{
+					testEntity.ID: {MaxNodes: make(map[node.RolesMask]uint16)},
```
kostko, line 640: Same as above.

go/registry/api/runtime.go
```diff
@@ -179,10 +187,13 @@ type StorageParameters struct {
...
 	// CheckpointChunkSize is the chunk size parameter for checkpoint creation.
 	CheckpointChunkSize uint64 `json:"checkpoint_chunk_size"`
+
+	// MinPoolSize is the minimum required candidate storage node pool size.
+	MinPoolSize uint64 `json:"min_pool_size"`
 }
 
 // ValidateBasic performs basic storage parameter validity checks.
-func (s *StorageParameters) ValidateBasic() error {
+func (s *StorageParameters) ValidateBasic(ver uint16) error {
```
kostko, line 196:
```suggestion
func (s *StorageParameters) ValidateBasic(version uint16) error {
```

go/registry/api/runtime.go
```diff
@@ -316,6 +341,56 @@ type Runtime struct { // nolint: maligned
...
+		return gmRuntime
+	case GovernanceConsensus:
+		return gmConsensus
+	default:
+		return "[unsupported runtime governance model]"
+	}
+}
+
+// FromString deserializes a string into a RuntimeGovernanceModel.
+func (gm *RuntimeGovernanceModel) FromString(str string) error {
```
kostko, line 381: Instead of this implement `MarshalText` and `UnmarshalText` so things like JSON serialization will work as expected.

go/registry/api/runtime.go
```diff
@@ -380,6 +455,11 @@ func (r *Runtime) ValidateBasic(strictVersion bool) error {
 	if err := r.Staking.ValidateBasic(r.Kind); err != nil {
 		return fmt.Errorf("bad staking parameters: %w", err)
 	}
+
+	if v > 1 && (r.GovernanceModel < 1 || r.GovernanceModel > GovernanceConsensus) {
```
kostko, line 459: Maybe define a constant like `GovernanceMax`.

kostko (commented):
Don't forget to update the documentation and add tests for the new features (whitelist changes, minimum pool size, submitting register runtime txes for consensus/runtime governance types).

kostko:
>  add update_runtime runtime message support & e2e tests (maybe in a separate PR?).

Yes, this should be a separate PR.

docs/consensus/registry.md
```diff
@@ -51,12 +51,21 @@ offline and having entities as separate resources enables that.
...
+The chosen governance model indicates how the runtime descriptor can be updated
+in the future.
+There are currently three supported governance models:
+    * **entity governance** -- the runtime owner is the only one who can update
+      the runtime descriptor via `registry.RegisterRuntime` method calls.
+    * **runtime-defined governance** -- the runtime itself is the only one who
+      can update the runtime descriptor by emitting a runtime message.
+    * **consensus layer governance** -- only the consensus layer itself can
+      update the runtime descriptor either through a network upgrade or some
+      other future means.
```
kostko, line 68:
```suggestion
The chosen governance model indicates how the runtime descriptor can be updated
in the future.

There are currently three supported governance models:

* **Entity governance** where the runtime owner is the only one who can update
  the runtime descriptor via `registry.RegisterRuntime` method calls.

* **Runtime-defined governance** where the runtime itself is the only one who
  can update the runtime descriptor by emitting a runtime message.

* **Consensus layer governance** where only the consensus layer itself can
  update the runtime descriptor through network governance.
```

go/consensus/tendermint/apps/registry/state/state.go
```diff
@@ -443,6 +375,46 @@ func (s *ImmutableState) NodeStatus(ctx context.Context, id signature.PublicKey)
...
+		if !signedNodeByEntityKeyFmt.Decode(it.Key(), &hEntityID, &hNodeID) || !hEntityID.Equal(&hID) {
+			break
+		}
+
+		rawSignedNode, err := s.is.Get(ctx, signedNodeKeyFmt.Encode(&hNodeID))
+		if err != nil {
+			return nil, abciAPI.UnavailableStateError(err)
+		}
+		if rawSignedNode == nil {
+			return nil, registry.ErrNoSuchNode
```
kostko, line 400: Doesn't this indicate state corruption as if the node is removed the `signedNodeByEntityKey` index should also be updated? In this case wrap it in `abciAPI.UnavailableStateError(...)`.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -241,13 +241,74 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
...
 			ctx.Logger().Error("RegisterNode: node's entity not in a runtime's whitelist",
 				"entity", newNode.EntityID,
 				"runtime", rt.ID,
 			)
 			return registry.ErrForbidden
 		}
+		if wcfg.MaxNodes == nil {
+			continue
+		}
+		if len(wcfg.MaxNodes) == 0 {
```
kostko, line 258: This should also cover the `nil` case so you don't need to have two checks.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -241,13 +241,74 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
...
+			continue
+		}
+
+		// Map is present and non-empty, check per-role restrictions
+		// on the maximum number of nodes per entity.
+
```
kostko, line 264: The below code needs some test coverage.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -542,13 +602,19 @@ func (app *registryApplication) registerRuntime( // nolint: gocyclo
...
 	// NOTE: If this is invoked during InitChain then there is no actual transaction
 	//       and thus no transaction signer so we must skip this check.
-	if !ctx.IsInitChain() && !sigRt.Signature.PublicKey.Equal(ctx.TxSigner()) {
+	if !ctx.IsInitChain() && !rt.EntityID.Equal(ctx.TxSigner()) {
 		return registry.ErrIncorrectTxSigner
 	}
 
+	// Consensus-layer governance runtimes can only be registered during genesis.
+	if !ctx.IsInitChain() && rt.GovernanceModel == registry.GovernanceConsensus {
+		return registry.ErrForbidden
```
kostko, line 615: Add test.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -573,16 +639,33 @@ func (app *registryApplication) registerRuntime( // nolint: gocyclo
 		if err != nil {
 			return err
 		}
+		// If the governance model is runtime-defined governance, the owner
+		// cannot perform any updates after the initial registration.
+		// Updates to such runtimes should be done via runtime messages.
+		if existingRt.GovernanceModel == registry.GovernanceRuntime {
+			return registry.ErrForbidden
```
kostko, line 646: Add test.

go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
```diff
@@ -137,14 +222,64 @@ func doFixGenesis(cmd *cobra.Command, args []string) {
...
+		KeyManager:      rt.KeyManager,
+		Staking:         rt.Staking,
+		TxnScheduler:    rt.TxnScheduler,
+		GovernanceModel: registry.GovernanceEntity,
+		Executor: registry.ExecutorParameters{
+			GroupSize:         rt.Executor.GroupSize,
+			GroupBackupSize:   rt.Executor.GroupBackupSize,
+			AllowedStragglers: rt.Executor.AllowedStragglers,
+			RoundTimeout:      rt.Executor.RoundTimeout,
+			MaxMessages:       32, // TODO: What should this be set to?
```
kostko, line 242: Good question, so setting this also requires configuring the global roothash consensus parameter (`MaxRuntimeMessages`). Either fixgenesis should set that as well (to the same value?) or we leave this at zero. Let's set this to 32 (both per-runtime and the global roothash consensus parameter) and then we can discuss what to set this to before the actual upgrade.

go/oasis-node/cmd/registry/runtime/runtime.go
```diff
@@ -405,17 +405,22 @@ func runtimeFromFlags() (*registry.Runtime, signature.Signer, error) { // nolint
...
-				return nil, nil, fmt.Errorf("entity whitelist runtime admission policy parse entity ID: %w", err)
+				return nil, fmt.Errorf("entity whitelist runtime admission policy parse entity ID: %w", err)
 			}
-			entities[e] = true
+			entities[e] = registry.EntityWhitelistConfig{
+				MaxNodes: make(map[node.RolesMask]uint16),
+			}
+			// TODO: Handle the clusterfuck of parsing the nested map from
+			// command-line arguments sometime later.  As it is now, it's
+			// configured as unlimited nodes of any role for the given entity.
```
kostko, line 423: Don't forget to file an issue.

go/registry/api/api.go
```diff
@@ -995,66 +984,53 @@ func exactlyOneTrue(conds ...bool) bool {
...
+		return fmt.Errorf("%w: test runtime not allowed", ErrInvalidArgument)
 	}
 
 	if err := rt.Genesis.SanityCheck(isGenesis); err != nil {
-		return nil, err
+		return err
+	}
+
+	// Make sure the specified runtime governance model is allowed.
+	if params.EnableRuntimeGovernanceModels == nil {
```
kostko, line 1019:
```suggestion
	if len(params.EnableRuntimeGovernanceModels) == 0 {
```

go/registry/api/api.go
```diff
@@ -995,66 +984,53 @@ func exactlyOneTrue(conds ...bool) bool {
...
+		// No runtime governance models are allowed.
+		return fmt.Errorf("%w: no runtime governance models are enabled", ErrForbidden)
+	}
+	if allowed, exists := params.EnableRuntimeGovernanceModels[rt.GovernanceModel]; !exists || !allowed {
```
kostko, line 1023: The default value in case a key doesn't exist is `false` so just do:
```suggestion
	if !params.EnableRuntimeGovernanceModels[rt.GovernanceModel] {
```

go/registry/api/api.go
```diff
@@ -1165,6 +1172,23 @@ func VerifyRuntimeUpdate(logger *logging.Logger, currentRt, newRt *Runtime) erro
...
+	// Check if governance model update is valid.
+	if currentRt.GovernanceModel != newRt.GovernanceModel {
+		// Transitioning from entity to runtime governance is allowed, but
+		// all other transitions are not.
+		if !(currentRt.GovernanceModel == GovernanceEntity && newRt.GovernanceModel == GovernanceRuntime) {
+			logger.Error("RegisterRuntime: invalid governance model transition",
+				"current_gm", currentRt.GovernanceModel.String(),
+				"new_gm", newRt.GovernanceModel.String(),
+			)
+			return ErrRuntimeUpdateNotAllowed
```
kostko, line 1184: Add test.

go/registry/api/runtime.go
```diff
@@ -110,6 +113,11 @@ func (e *ExecutorParameters) ValidateBasic() error {
 	if e.RoundTimeout < 5 {
 		return fmt.Errorf("round timeout too small")
 	}
+
+	if version > 1 && e.MinPoolSize < e.GroupSize+e.GroupBackupSize {
+		return fmt.Errorf("minimum pool size too small")
```
kostko, line 118: Add test.

go/registry/api/runtime.go
```diff
@@ -267,7 +292,7 @@ func (s *RuntimeStakingParameters) ValidateBasic(runtimeKind RuntimeKind) error
 const (
 	// LatestRuntimeDescriptorVersion is the latest entity descriptor version that should be used
 	// for all new descriptors. Using earlier versions may be rejected.
-	LatestRuntimeDescriptorVersion = 1
+	LatestRuntimeDescriptorVersion = 2
 
 	// Minimum and maximum descriptor versions that are allowed.
 	minRuntimeDescriptorVersion = 1
```
kostko, line 298: When would we support version 1 runtime descriptors? Isn't the idea that fixgenesis will convert them and so we don't need to support version 1 here (so no need for version checks above and below)?

go/registry/api/runtime.go
```diff
@@ -316,6 +341,61 @@ type Runtime struct { // nolint: maligned
...
+		return gmRuntime
+	case GovernanceConsensus:
+		return gmConsensus
+	default:
+		return "[unsupported runtime governance model]"
+	}
+}
+
+func (gm RuntimeGovernanceModel) MarshalText() ([]byte, error) {
+	return []byte(gm.String()), nil
```
kostko, line 383: This shouldn't just use `.String()` as that can in theory return the `[unsupported runtime governance model]`. It should instead return an error in this case.

go/staking/api/slashing.go
```diff
@@ -41,6 +41,10 @@ func (s SlashReason) MarshalText() ([]byte, error) {
 // UnmarshalText decodes a text slice into a SlashReason.
 func (s *SlashReason) UnmarshalText(text []byte) error {
 	switch string(text) {
+	// XXX: The "0" case is only for backward compatibility, so that the old
+	// genesis file loads -- remove this once mainnet is upgraded!
```
kostko, line 45: Wouldn't it be better if we handled this in fixgenesis?

go/consensus/tendermint/apps/registry/state/state.go
abukosek: Good catch, I'll add that.

go/consensus/tendermint/apps/registry/transactions.go
abukosek: Yeah, I'm adding that now :)

go/registry/api/runtime.go
abukosek: Oops, this is a leftover from before I wrote the fixgenesis code, I'll change the minimum version to 2.

go/staking/api/slashing.go
abukosek: No, because there are too many structures to duplicate and there's more chance of something going wrong with the conversion (and since this is in staking code it's better to be more careful).  A two-line fix like this is cleaner, IMO :)
I'll file an issue later, so we don't forget to remove this once the upgrade is complete.

abukosek:
Thanks for the comments!  This is now ready for re-review.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -542,13 +599,19 @@ func (app *registryApplication) registerRuntime( // nolint: gocyclo
 		return err
 	}
 
-	// Make sure the signer of the transaction matches the signer of the runtime.
+	// Make sure the signer of the transaction matches the signer of the
+	// entity that is controlling the runtime.
 	// NOTE: If this is invoked during InitChain then there is no actual transaction
 	//       and thus no transaction signer so we must skip this check.
-	if !ctx.IsInitChain() && !sigRt.Signature.PublicKey.Equal(ctx.TxSigner()) {
+	if !ctx.IsInitChain() && !rt.EntityID.Equal(ctx.TxSigner()) {
```
kostko, line 606: This actually needs to take into account the existing governance model. So this check is true for entity governance, but for runtime governance the check would look something like checking `ctx.CallerAddress()` for equality with `staking.NewRuntimeAddress(rt.ID)`.

So maybe put a switch here that decides based on the governance model. The switch could also incorporate the conditional below which returns `ErrForbidden` in case of consensus governance.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -241,13 +241,71 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
...
+		// only have a single role).
+		for _, role := range node.Roles() {
+			if !newNode.HasRoles(role) {
+				// Skip unset roles.
+				continue
+			}
+
+			maxNodes, exists := wcfg.MaxNodes[role]
+			if !exists {
+				// No such role found in whitelist.
```
kostko, line 272: Log something in these cases (below) that return `ErrForbidden`.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -573,16 +650,33 @@ func (app *registryApplication) registerRuntime( // nolint: gocyclo
 		if err != nil {
 			return err
 		}
+		// If the governance model is runtime-defined governance, the owner
+		// cannot perform any updates after the initial registration.
+		// Updates to such runtimes should be done via runtime messages.
+		if existingRt.GovernanceModel == registry.GovernanceRuntime {
```
kostko, line 656: If the runtime update message will also call `registerRuntime` then this check will prevent any updates from happening. In case the governance model is the same as in `rt` then this has already been checked in the switch above (checking against `ctx.CallerAddress()`). So the only remaining issue is if the governance model in `rt` would be different than in `existingRt` -- but that should be checked in `VerifyRuntimeUpdate`?

abukosek: You are right, I've fixed this now.  I've also moved the signer checks below this code, so we can do them based on the existing runtime governance model instead of the new one (otherwise it wouldn't be possible to transition from entity to runtime governance, since the transaction for doing that will still be signed with the entity's signer and not the runtime's) :)

go/consensus/tendermint/apps/registry/transactions.go
abukosek: Thanks, fixed! :)

go/registry/api/runtime.go
```diff
@@ -179,10 +187,13 @@ type StorageParameters struct {
...
 	// CheckpointChunkSize is the chunk size parameter for checkpoint creation.
 	CheckpointChunkSize uint64 `json:"checkpoint_chunk_size"`
+
+	// MinPoolSize is the minimum required candidate storage node pool size.
+	MinPoolSize uint64 `json:"min_pool_size"`
 }
 
 // ValidateBasic performs basic storage parameter validity checks.
-func (s *StorageParameters) ValidateBasic() error {
+func (s *StorageParameters) ValidateBasic(version uint16) error {
```
kostko, line 196: The version argument should no longer be needed as we only allow version 2 descriptors.

go/registry/api/runtime.go
```diff
@@ -214,6 +225,11 @@ func (s *StorageParameters) ValidateBasic() error {
 			return fmt.Errorf("storage CheckpointChunkSize parameter too small")
 		}
 	}
+
+	if version > 1 && s.MinPoolSize < s.GroupSize {
```
kostko, line 229:
```suggestion
	if s.MinPoolSize < s.GroupSize {
```

go/registry/api/runtime.go
```diff
@@ -380,6 +464,11 @@ func (r *Runtime) ValidateBasic(strictVersion bool) error {
 	if err := r.Staking.ValidateBasic(r.Kind); err != nil {
 		return fmt.Errorf("bad staking parameters: %w", err)
 	}
+
+	if v > 1 && (r.GovernanceModel < 1 || r.GovernanceModel > GovernanceMax) {
```
kostko, line 468:
```suggestion
	if r.GovernanceModel < 1 || r.GovernanceModel > GovernanceMax {
```

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -575,14 +638,67 @@ func (app *registryApplication) registerRuntime( // nolint: gocyclo
...
+		var rtToCheck *registry.Runtime
+		if existingRt != nil {
+			rtToCheck = existingRt
+		} else {
+			rtToCheck = rt
+		}
+
+		switch rtToCheck.GovernanceModel {
+		case registry.GovernanceEntity:
+			if !rt.EntityID.Equal(ctx.TxSigner()) {
```
kostko, line 660: I know the EntityID cannot change but this makes it more clear IMO:
```suggestion
			if !rtToCheck.EntityID.Equal(ctx.TxSigner()) {
```

go/consensus/tendermint/apps/registry/transactions.go
kostko: Looks good now, thanks!

docs/consensus/registry.md
```diff
@@ -51,12 +51,24 @@ offline and having entities as separate resources enables that.
...
-identifier, kind, admission policy, committee scheduling, storage etc. For a
-full description of the runtime descriptor see [the `Runtime` structure].
+identifier, kind, admission policy, committee scheduling, storage, governance
+model, etc. For a full description of the runtime descriptor see
+[the `Runtime` structure].
 
-Currently only the owning entity is allowed to make any modifications to the
-runtime. There are plans to enable runtimes to update their own descriptors in
-the future to enable runtimes to be self-governing.
+<!-- markdownlint-disable no-space-in-emphasis -->
```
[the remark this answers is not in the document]
abukosek: Yeah, it seems broken by default -- it complains about the spaces in emphasised phrases.
I've re-enabled it below now :)

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -241,13 +241,84 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
...
+			if grr != nil {
+				ctx.Logger().Error("RegisterNode: failed to query entity nodes",
+					"err", grr,
+					"entity", newNode.EntityID,
+				)
+				return grr
+			}
+			var curNodes uint16
+			for _, n := range nodes {
+				if n.IsExpired(uint64(epoch)) || n.GetRuntime(rt.ID) == nil {
```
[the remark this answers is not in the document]
abukosek: Good catch!  I've fixed it now.

go/consensus/tendermint/apps/registry/genesis.go
```diff
@@ -46,47 +46,43 @@ func (app *registryApplication) InitChain(ctx *abciAPI.Context, request types.Re
...
-					"runtime", v,
+					"runtime_id", rt.ID,
 				)
 				return fmt.Errorf("registry: genesis runtime registration failure: %w", err)
 			}
 		}
 	}
-	for i, v := range st.SuspendedRuntimes {
-		if v == nil {
+	for i, rt := range st.SuspendedRuntimes {
```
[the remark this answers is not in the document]
abukosek: Filed as https://github.com/oasisprotocol/oasis-core/issues/3619 :)

docs/consensus/registry.md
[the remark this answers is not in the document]
ptrus: thx!

kostko (approved):

go/oasis-node/cmd/registry/runtime/runtime.go
abukosek: Filed as https://github.com/oasisprotocol/oasis-core/issues/3626 :)

go/staking/api/slashing.go
abukosek: Filed as https://github.com/oasisprotocol/oasis-core/issues/3627 :)

abukosek:
Thanks!

PR #3559, opened 2020-12-09
go/oasis-node/cmd: Add oasis-node stake account nonce CLI command

tjanez:
It can be used to get a staking account's current nonce.

### DIFF
--- a/.changelog/3559.feature.md
+++ b/.changelog/3559.feature.md
@@ -1,0 +1,3 @@
+go/oasis-node/cmd: Add `oasis-node stake account nonce` CLI command
+
+It can be used to get a staking account's current nonce.
--- a/go/oasis-node/cmd/stake/account.go
+++ b/go/oasis-node/cmd/stake/account.go
@@ -58,7 +58,7 @@ var (
 
 	accountInfoCmd = &cobra.Command{
 		Use:   "info",
-		Short: "query account info",
+		Short: "get account info",
 		Run:   doAccountInfo,
 	}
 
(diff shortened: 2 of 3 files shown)


kostko (approved):

.changelog/3559.feature.md
```diff
@@ -0,0 +1,3 @@
+go/oasis-node/cmd: Add `oasis-node stake account nonce` CLI command
+
+It can be used to get a staking account's current nonce.
```
ptrus, line 3: Just wondering what's the use case for this, since the `stake account info` command already includes this in response? 

The only thing i can come up with is using this in some automation - which i think we actually want to discourage and instead defer to the grpc api.

tjanez: > Just wondering what's the use case for this, since the `stake account info` command already includes this in response?

The use case I have in mind is to have an elegant way of getting an account's nonce instead of having to parse the `stake account info` command's response which is not guaranteed to stay stable.

> The only thing i can come up with is using this in some automation - which i think we actually want to discourage and instead defer to the grpc api.

You guessed it :slightly_smiling_face:. I've had some occasions where it was the most convenient to just write a simple Bash script to generate, sign and submit a batch of staking transactions.

Something like a more proper Go program acting as a gRPC client and a CLI would be nice, but out of scope for these smallish projects.

So, I would say that we probably don't want to discourage all such usage of our CLI tools, just not relying on their output being stable and using them to actually write a backend, e.g. a gateway, a staking dashboard, ...

PR #3567, opened 2020-12-14
rust: ignore RUSTSEC-2020-0091 until dependency bumps

ptrus:


kostko:
I already bumped arc-swap in #3511 (see last commit).

PR #3569, opened 2020-12-15
go/runtime/scheduling: Consumer should pull from scheduler

kostko:
Previously the scheduler would push batches to the consumer (e.g. compute
node) which makes no sense as the consumer knows when it's ready to accept
new batches. This changes the model so that the consumer pulls batches from
the scheduler.

### DIFF
--- a/go/runtime/scheduling/simple/txpool/orderedmap/ordered_map.go
+++ b/go/runtime/scheduling/simple/txpool/orderedmap/ordered_map.go
@@ -188,12 +188,13 @@ func (q *orderedMap) UpdateConfig(cfg api.Config) error {
 		if current == nil {
 			break
 		}
-		if cfg.MaxPoolSize > uint64(newQueue.Len()) {
+		if uint64(newQueue.Len()) >= cfg.MaxPoolSize {
 			break
 		}
 		el := current.Value.(*pair)
 		txSize := uint64(len(el.Value))
 		if txSize > cfg.MaxBatchSizeBytes {
+			current = current.Prev()
 			continue
 		}
 		newQueue.PushFront(el)
--- a/go/worker/compute/executor/committee/node.go
+++ b/go/worker/compute/executor/committee/node.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/eapache/channels"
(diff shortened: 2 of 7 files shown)


go/worker/compute/executor/committee/node.go
```diff
@@ -745,53 +754,86 @@ func (n *Node) proposeTimeoutLocked() error {
...
+	if len(batch) == 0 {
+		return
+	}
+
+	// Check if we can actually propose a batch. If we are not a transaction scheduler, we may be
+	// able to propose a timeout.
+	switch {
+	case epoch.IsTransactionScheduler(lastHeader.Round):
+		// Continues bellow.
+	case epoch.IsExecutorWorker():
```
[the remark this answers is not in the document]
kostko: Yeah good point, will change it.

go/runtime/scheduling/simple/txpool/orderedmap/ordered_map.go
```diff
@@ -188,12 +188,13 @@ func (q *orderedMap) UpdateConfig(cfg api.Config) error {
...
 			break
 		}
-		if cfg.MaxPoolSize > uint64(newQueue.Len()) {
+		if uint64(newQueue.Len()) >= cfg.MaxPoolSize {
 			break
 		}
 		el := current.Value.(*pair)
 		txSize := uint64(len(el.Value))
 		if txSize > cfg.MaxBatchSizeBytes {
+			current = current.Prev()
```
ptrus, line 197: :facepalm: 
Can you please also add a simple unit test for update configuration to `scheduling/simple/txpool/tester.go`

kostko: Oh I see, there is one for the scheduler (this is the one that caught this) but not for the txpool. Will add it here as well.

ptrus (approved):
Great! i think this is indeed much better. Thanks for also noticing & fixing the update config stuff.

Looks good, just two minor comments.

go/runtime/scheduling/simple/txpool/orderedmap/ordered_map.go
kostko: Done.

Issue #3572, opened 2020-12-16
oasis-core-runtime: Remove deprecated functionality


kostko:
There is a bunch of functionality present in `oasis-core-runtime` which should be removed in favor of it being moved to runtimes (or the upcoming Runtime SDK) and the `oasis-core-runtime` only providing low-level building blocks:
* [x] `transaction::MethodDispatcher`
  No known external runtime uses it and the only use case for the `MethodDispatcher` is for unit tests.
* [x] `transaction::register_runtime_txn_methods` macro (related to MethodDispatcher).
* [x] `transaction::types::{TxnCall, TxnOutput, TxnCheckResult}` structures (related to MethodDispatcher).
* [x] `storage::StorageContext`
  Instead the storage context reference should be passed inside the `transaction::Context`. If a runtime needs to pass it using thread-local storage, it can implement something like `StorageContext` by itself.
  
  Known users: [oasis-ethwasi-runtime](https://github.com/oasislabs/oasis-ethwasi-runtime), [oasis-ssvm-runtime](https://github.com/second-state/oasis-ssvm-runtime)
* [x] The `runtime`, `tags` and `messages` fields should be removed (together with related methods) from `transaction::Context` in favor of the dispatcher returning tags/messages as part of execution results.
* [x] The `transaction` module in the `oasis-core-client`. This requires moving all E2E test clients to Go.

PR #3573, opened 2020-12-16
runtime: Various changes needed for the Runtime SDK

kostko:
Also see #3572.


runtime/src/dispatcher.rs
```diff
@@ -326,126 +473,45 @@ impl Dispatcher {
...
-            }
+        // Return and error response in case of failure.
+        if let Err(error) = results {
+            warn!(self.logger, "Dispatching batch error"; "err" => %error);
+            protocol
+                .send_response(
+                    id,
+                    Body::Error {
+                        module: "".to_owned(), // XXX: Error codes.
+                        code: 0,               // XXX: Error codes.
```
ptrus, line 510: Shouldn't at least a different code be used here, so that it doesn't correspond to `CodeNoError`?

go/runtime/client/api/api.go
```diff
@@ -122,6 +125,21 @@ type GetTxsRequest struct {
...
 
+// GetEventsRequest is a GetEvents request.
+type GetEventsRequest struct {
+	RuntimeID common.Namespace `json:"runtime_id"`
+	Round     uint64           `json:"round"`
+}
+
+// Event is an event emitted by a runtime.
+//
+// Key and value semantics are runtime-dependent.
```
ptrus, line 136: Should there be a note that these are basically the "runtime transaction tags"?

go/runtime/client/tests/tester.go
```diff
@@ -146,6 +146,13 @@ func testQuery(
 	// Check for values from TestNode/Client/SubmitTx
 	require.EqualValues(t, testInput, txns[0])
 
+	// Check events query (see mock worker for emitted events).
+	events, err := c.GetEvents(ctx, &api.GetEventsRequest{RuntimeID: runtimeID, Round: 2})
```
ptrus, line 150: Can you also update the txsource runtime workload to query the events?

runtime/src/dispatcher.rs
kostko: Good point, so currently this is not used anywhere but makes sense to change to an actual error code :-)

go/runtime/client/tests/tester.go
kostko: Added.

go/runtime/client/api/api.go
kostko: Added.

ptrus (approved):
thanks for the changes!

PR #3594, opened 2021-01-04
go/consensus/tendermint/roothash: Fix latest block reindexing

kostko:
Only commit the block in case it was not already committed during reindex. This
can happen when reindexing after a crash (e.g., the before_index crash point)
since the initial height at which the reindex happens does not necessarily
contain a round finalization event so reindexing up to height-1 doesn't help.

Discovered during long-term tests.


ptrus (approved):
nice find!

PR #3603, opened 2021-01-06
go/oasis-test-runner: governance workload

ptrus:
- Adds the governance workload. Governance workload does the following:
  - continuously submits proposals (both upgrade & cancel pending upgrade)
  - votes for proposals (makes sure some proposals get rejected and some pass) 
  - cancels any pending upgrades (via submitting a proposal and making sure it passes) `UpgradeCancelMinEpochDiff` before execution, to ensure no upgrades actually happen during the txsource test. In future we could/should probably also support this in the txsource test. 
- queries workload is updated with governance queries
- use multiple entities for txsrouce validators

TODO:
- [x] add governance queries to queries workload

### DIFF
--- a/go/oasis-node/cmd/debug/txsource/workload/governance.go
+++ b/go/oasis-node/cmd/debug/txsource/workload/governance.go
@@ -1,0 +1,462 @@
+package workload
+
+import (
+	"context"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"math/rand"
+	"time"
+
+	"google.golang.org/grpc"
+
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
+	memorySigner "github.com/oasisprotocol/oasis-core/go/common/crypto/signature/signers/memory"
+	consensus "github.com/oasisprotocol/oasis-core/go/consensus/api"
+	epochtime "github.com/oasisprotocol/oasis-core/go/epochtime/api"
+	governance "github.com/oasisprotocol/oasis-core/go/governance/api"
+	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
+	upgrade "github.com/oasisprotocol/oasis-core/go/upgrade/api"
+)
+
+// NameGovernance is the name of the governance workload.
+const NameGovernance = "governance"
+
+var errUnexpectedGovTxResult = fmt.Errorf("unexpected governance tx result")
+
+// Governance is the governance workload.
+var Governance = &governanceWorkload{
+	BaseWorkload: NewBaseWorkload(NameGovernance),
+}
+
+var numProposerAccounts = 10
+
+// How likely voters should vote YES for the proposal made by i'th proposer.
+var proposersVoteYesRate = []uint8{
+	100,
+	99,
+	98,
+	95,
+	92,
+	90,
+	80,
+	70,
+	50,
+	0,
+}
+
+type governanceWorkload struct {
+	BaseWorkload
+
+	ctx context.Context
+	rng *rand.Rand
+
+	currentEpoch epochtime.EpochTime
+	parameters   *governance.ConsensusParameters
+
+	consensus  consensus.ClientBackend
+	governance governance.Backend
+
+	proposerAccounts []*struct {
+		signer  signature.Signer
+		address staking.Address
+		nonce   uint64
+	}
+	ensureYesVote map[uint64]bool
+
+	validatorEntities []signature.Signer
+}
+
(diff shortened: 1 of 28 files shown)


go/oasis-node/cmd/debug/txsource/workload/workload.go
```diff
@@ -92,6 +94,15 @@ func (bw *BaseWorkload) GasPrice() uint64 {
 	return gasPrice.ToBigInt().Uint64()
 }
 
+// LoadNonce returns the current nonce for the provided address.
+func (bw *BaseWorkload) LoadNonce(ctx context.Context, addr staking.Address) (uint64, error) {
```
kostko, line 98: Note that there's also the `GetSignerNonce(ctx context.Context, req *GetSignerNonceRequest) (uint64, error)` method on the consensus backend which gets you the same thing so maybe you should use this instead of this method?

kostko (commented):
Just a very quick look, still need to check the actual governance workload :grin:

go/oasis-node/cmd/debug/txsource/workload/workload.go
ptrus: thanks for noticing! as you mentioned "offline" loading the nonce was indeed not needed :-)

go/oasis-node/cmd/debug/txsource/workload/governance.go
```diff
@@ -0,0 +1,479 @@
...
+		// Ensure epoch transition won't happen for the next 10 blocks. Otherwise
+		// if the epoch transition happens during the loop iteration, we can have
+		// outdated local state.
+		var epoch2 epochtime.EpochTime
+		var block *consensus.Block
+		block, err = cnsc.GetBlock(g.ctx, consensus.HeightLatest)
+		if err != nil {
+			return fmt.Errorf("querying block: %w", err)
+		}
+		epoch2, err = cnsc.GetEpoch(g.ctx, block.Height+10)
```
kostko, line 429: This will not work with the new beacon as `GetEpoch` there performs a query which will fail if you give it a future height. Also with the new beacon there is no way to know for sure when the next transition will happen.

go/oasis-node/cmd/debug/txsource/workload/governance.go
```diff
@@ -0,0 +1,479 @@
...
+
+		// Epoch transition - update the local state accordingly.
+		if epoch > g.currentEpoch {
+			if err = g.handleEpochTransition(epoch); err != nil {
+				return fmt.Errorf("handle epoch transition: %w", err)
+			}
+
+			// Make sure no pending upgrade will go through.
+			// XXX: this makes sure that any pending upgrades that are about to be executed are
+			// canceled. When txsource suite supports handling upgrades mid-run, remove this part.
```
kostko, line 446: Is there an issue for tracking this?

kostko (approved):

go/oasis-node/cmd/debug/txsource/workload/governance.go
ptrus: will open one

go/oasis-node/cmd/debug/txsource/workload/governance.go
ptrus: Hm yeah that's right. Ok i'll update the workload and turn things around a bit:
- first a request will be made (e.g. submit proposal, vote ...)
- after the request the workload will check if the request should have failed or succeed based on the state at that height

ptrus: actually simplified it a bit more even. In case the unexpected result happens (e.g. a proposal is successfully submitted when it shouldn't be - or otherwise) the workload checks if the epoch transition just happened, and in case it did it doesn't fail the whole workload and just emits an error.

kostko: Yeah that sounds like the right solution, thanks!

go/oasis-node/cmd/debug/txsource/workload/governance.go
ptrus: opened https://github.com/oasisprotocol/oasis-core/issues/3609

go/oasis-node/cmd/debug/txsource/workload/governance.go
ptrus: Np! thanks for noticing that this wouldn't work with the new beacon :grin: forgot about it.

Issue #3579, opened 2020-12-17
Support upgrades with different binary hashes


ptrus:
Currently the upgrade module supports a single upgrade type ("internal") which relies on binary hash to ensure correct node version for upgrade is used:
https://github.com/oasisprotocol/oasis-core/blob/5af6d91815f01b7a2d195932c8c10fd9faa7e343/go/upgrade/api/api.go#L71-L73

We should stop using the hash of the binary to ensure correct version is running and instead use something like a software version. Otherwise upgrade will be hard to coordinate with people doing custom builds. 

Can either create a new upgrade type or update (and rename?) the existing `"internal"` upgrade type to not rely on exact matching binaries - as that doesn't seem all that useful.

Should do after https://github.com/oasisprotocol/oasis-core/pull/3536 is merged, as that PR contains some significant changes to the upgrade module.

Also enforce unique upgrade descriptor names.

kostko:
> Can either update a new upgrade type or update (and rename?) the existing "internal" upgrade type to not rely on exact matching binaries - as that doesn't seem all that useful.

I would opt for reusing/renaming the existing upgrade type.

PR #3610, opened 2021-01-11
go/upgrade: use ProtocolVersions as internal upgrade identifier

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3579

TODO:
- [x] ~enforce upgrade descriptor name uniqueness~ actually instead of the proposed change, the `CancelUpgrade` method (the only place where this was problematic) was updated to require passing in the whole descriptor to be canceled.

NOTE: this also introduces a minor change the governance submit proposal transaction format (see governance/gen_vectors change).

### DIFF
--- a/go/governance/gen_vectors/main.go
+++ b/go/governance/gen_vectors/main.go
@@ -2,14 +2,15 @@
 package main
 
 import (
-	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"math"
 
+	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
 	"github.com/oasisprotocol/oasis-core/go/common/quantity"
+	"github.com/oasisprotocol/oasis-core/go/common/version"
 	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"
 	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction/testvectors"
 	epochtime "github.com/oasisprotocol/oasis-core/go/epochtime/api"
@@ -25,9 +26,6 @@ func main() {
 
 	var vectors []testvectors.TestVector
 
-	var emptyHash hash.Hash
-	emptyHash.Empty()
-
 	// Generate different gas fees.
 	for _, fee := range []*transaction.Fee{
 		{},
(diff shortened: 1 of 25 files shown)


go/governance/gen_vectors/main.go
```diff
@@ -40,14 +38,39 @@ func main() {
 			// Valid submit upgrade proposal transaction.
 			for _, epoch := range []uint64{0, 1000, 10_000_000} {
 				for _, name := range []string{"", "descriptor-name"} {
-					for _, identifier := range []string{hex.EncodeToString([]byte{}), hex.EncodeToString(emptyHash[:]), "000000000000000000000000000000000000000000000000000000000abcdefg"} {
+					for _, identifier := range []version.ProtocolVersions{
```
ptrus, line 41: cc @tjanez when this PR is finalized/merged we should communicate this change for the submit proposal tx format to Juan

go/upgrade/api/api.go
```diff
@@ -174,20 +157,16 @@ func (d Descriptor) ValidateBasic() error {
...
-		if err = descriptorHash.UnmarshalHex(d.Identifier); err != nil {
-			return fmt.Errorf("can't decode stored upgrade identifier: %w", err)
+		var descriptorVersion version.ProtocolVersions
+		if err := cbor.Unmarshal(d.Identifier, &descriptorVersion); err != nil {
+			return fmt.Errorf("can't decode descriptor upgrade identifier: %w", err)
 		}
 
-		if !OwnHash.Equal(&descriptorHash) {
-			return fmt.Errorf("binary hash missmatch: own: %d, required: %d", OwnHash, descriptorHash)
+		// TODO: allow patch version difference?
```
kostko, line 167: Yeah could ignore patch.

go/oasis-node/cmd/control/control.go
```diff
@@ -150,7 +151,7 @@ func doUpgradeBinary(cmd *cobra.Command, args []string) {
 	}
 
 	var desc upgrade.Descriptor
-	if err = json.Unmarshal(descriptorBytes, &desc); err != nil {
+	if err = cbor.Unmarshal(descriptorBytes, &desc); err != nil {
```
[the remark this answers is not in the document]
ptrus: yeah there was no reason for this. I reverted this to json so that cli commands still accept json.

go/upgrade/api/api.go
```diff
@@ -174,20 +157,16 @@ func (d Descriptor) ValidateBasic() error {
...
+		if ownVersion != descriptorVersion {
+			return fmt.Errorf("binary version missmatch: own: %d, required: %d", ownVersion, descriptorVersion)
```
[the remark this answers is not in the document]
ptrus: :+1: fixed throughout the code

go/upgrade/api/api.go
ptrus: Hm so i think we should either require exact match or ignore non-major. If we consider all such upgrades "breaking" changes then following the version scheme we would bump major version every-time in any case, so i think we can just match major versions? 

And yeah i think it's better to not do the exact version match as i imagine a scenario: upgrade is accepted and scheduled in X days, we need to do an urgent release before (e.g. to fix something critical). Hm now actually thinking about this, maybe we want to instead use the version as a `MinVersion` (e.g. the urgent fix could technically be breaking and we'd bump the major version as well)?

kostko: I think matching only the major version makes sense.

Besides the version the upgrade in question must also have the upgrade handler that is defined by the upgrade, so I guess "min version" could also work as long as the upgrade handler is also there.

ptrus: Yea agreed I also think major version is fine, will update the pr

ptrus: done

kostko (approved):

jberci (approved):

PR #3612, opened 2021-01-11
runtime/common: Implement Add, AddAssign, Sub, SubAssign for Quantity

kostko:


Yawning:
Should this have any specific (read: non-standard) underflow behavior?

runtime/src/common/quantity.rs
```diff
@@ -15,6 +18,36 @@ impl From<u64> for Quantity {
     }
 }
 
+impl<'a> Add<&'a Quantity> for Quantity {
+    type Output = Quantity;
+
+    fn add(mut self, other: &Quantity) -> Quantity {
```
ptrus, line 24: Maybe i'm missing something, but shouldn't `Add`  not update `self`?

kostko: So it is not `&mut self` but `mut self` (as required by the `Add` trait) which means that self is moved into the method. This means that there are no other references to it anywhere so it is safe to reuse it.

kostko:
> Should this have any specific (read: non-standard) underflow behavior?

Yeah it would be safer if we only provided `checked_sub`, will change it.

kostko:
Done, the use is now forced to be as follows:
```rust
        value = value
            .checked_sub(&amount.amount())
            .ok_or(Error::InsufficientBalance)?;
```
or
```rust
        total_supply = total_supply
            .checked_sub(&amount.amount())
            .expect("target account had enough balance so total supply should not underflow");
```

runtime/src/common/quantity.rs
ptrus: Ah i see. Thanks :+1:

ptrus (approved):

PR #3634, opened 2021-01-15
Add runtime message for updating the runtime descriptor

### LINKED ISSUE #3628: Implement ADR 4 (part 2, runtime messages)
As part of [ADR 0004], the `update_runtime` runtime message should be implemented to enable updating the runtime descriptor from the runtime itself.  Additionally, e2e tests should be written for this new functionality.

[ADR 0004]: https://github.com/oasisprotocol/oasis-core/blob/master/docs/adr/0004-runtime-governance.md

abukosek:
Closes #3628.

TODO:
- [x] add `update_runtime` runtime message.
- [x] write unit tests.
- [x] port the `Runtime` struct to Rust, ugh.
- [x] write e2e test.
- [x] debug e2e test.
- [x] add a test that tries to update a different runtime from another runtime (should fail).
- [x] clean-up code.
- [x] write changelog fragment.

### DIFF
--- a/go/consensus/tendermint/apps/keymanager/keymanager.go
+++ b/go/consensus/tendermint/apps/keymanager/keymanager.go
@@ -20,7 +20,6 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/keymanager/api"
 	keymanager "github.com/oasisprotocol/oasis-core/go/keymanager/api"
 	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
-	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
 )
 
 var emptyHashSha3 = sha3.Sum256(nil)
@@ -118,13 +117,22 @@ func (app *keymanagerApplication) onEpochChange(ctx *tmapi.Context, epoch beacon
 
 		// Suspend the runtime in case the registering entity no longer has enough stake to cover
 		// the entity and runtime deposits.
-		if !params.DebugBypassStake {
-			acctAddr := staking.NewAddress(rt.EntityID)
-			if err = stakeAcc.CheckStakeClaims(acctAddr); err != nil {
+		if !params.DebugBypassStake && rt.GovernanceModel != registry.GovernanceConsensus {
+			acctAddr := rt.StakingAddress()
+			if acctAddr == nil {
+				// This should never happen.
+				ctx.Logger().Error("unknown runtime governance model",
+					"rt_id", rt.ID,
+					"gov_model", rt.GovernanceModel,
+				)
+				return fmt.Errorf("unknown runtime governance model on runtime %s: %s", rt.ID, rt.GovernanceModel)
+			}
+
+			if err = stakeAcc.CheckStakeClaims(*acctAddr); err != nil {
 				ctx.Logger().Warn("insufficient stake for key manager runtime operation",
 					"err", err,
 					"entity", rt.EntityID,
-					"account", acctAddr,
+					"account", *acctAddr,
 				)
 
 				// Suspend runtime.
--- a/go/consensus/tendermint/apps/registry/transactions.go
+++ b/go/consensus/tendermint/apps/registry/transactions.go
@@ -462,9 +462,18 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
 	for _, rt := range paidRuntimes {
 		// Only resume a runtime if the entity has enough stake to avoid having the runtime be
 		// suspended again on the next epoch transition.
-		if !params.DebugBypassStake {
-			acctAddr := staking.NewAddress(rt.EntityID)
-			if err = stakeAcc.CheckStakeClaims(acctAddr); err != nil {
+		if !params.DebugBypassStake && rt.GovernanceModel != registry.GovernanceConsensus {
+			acctAddr := rt.StakingAddress()
+			if acctAddr == nil {
+				// This should never happen.
+				ctx.Logger().Error("unknown runtime governance model",
+					"rt_id", rt.ID,
+					"gov_model", rt.GovernanceModel,
+				)
+				return fmt.Errorf("unknown runtime governance model on runtime %s: %s", rt.ID, rt.GovernanceModel)
+			}
+
+			if err = stakeAcc.CheckStakeClaims(*acctAddr); err != nil {
 				continue
 			}
 		}
@@ -661,28 +670,25 @@ func (app *registryApplication) registerRuntime( // nolint: gocyclo
 			rtToCheck = rt
 		}
 
-		switch rtToCheck.GovernanceModel {
-		case registry.GovernanceEntity:
-			if !ctx.CallerAddress().Equal(staking.NewAddress(rtToCheck.EntityID)) {
-				// Transaction must be signed by controlling entity.
+		expectedAddr := rtToCheck.StakingAddress()
+		if expectedAddr == nil {
+			ctx.Logger().Error("RegisterRuntime: runtimes with consensus-layer governance can only be registered at genesis")
+			return registry.ErrForbidden
+		}
+
+		if !ctx.CallerAddress().Equal(*expectedAddr) {
+			switch rtToCheck.GovernanceModel {
+			case registry.GovernanceEntity:
 				ctx.Logger().Error("RegisterRuntime: transaction must be signed by controlling entity")
 				return registry.ErrIncorrectTxSigner
-			}
-		case registry.GovernanceRuntime:
-			if !ctx.CallerAddress().Equal(staking.NewRuntimeAddress(rtToCheck.ID)) {
-				// Caller must be the runtime itself.
+			case registry.GovernanceRuntime:
 				ctx.Logger().Error("RegisterRuntime: caller must be the runtime itself")
 				return registry.ErrForbidden
+			default:
+				// Basic validation should have caught this, but just in case...
+				ctx.Logger().Error("RegisterRuntime: invalid governance model")
+				return registry.ErrInvalidArgument
 			}
-		case registry.GovernanceConsensus:
-			// Consensus-layer governance runtimes can only be registered
-			// during genesis.
-			ctx.Logger().Error("RegisterRuntime: runtimes with consensus-layer governance")
-			return registry.ErrForbidden
-		default:
-			// Basic validation should have caught this, but just in case...
-			ctx.Logger().Error("RegisterRuntime: invalid governance model")
-			return registry.ErrInvalidArgument
 		}
 	}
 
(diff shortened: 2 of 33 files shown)


kostko (commented):
Make sure that there is a check which prevents runtimes from submitting update runtime messages for any runtime other than itself (e.g. validate the runtime ID).

Additionally the check in `registerRuntime` transaction processing for the `registry.GovernanceEntity` case should compare `ctx.CallerAddress()` against `staking.NewAddress(rtToCheck.EntityID)` instead of using `ctx.TxSigner()` (as it does currently).

go/registry/api/sanity_check.go
```diff
@@ -268,9 +268,11 @@ func SanityCheckStake(
 		generatedEscrows[addr].StakeAccumulator.AddClaimUnchecked(StakeClaimForNode(node.ID), StakeThresholdsForNode(node, nodeRts))
 	}
 	for _, rt := range runtimes {
-		// Add runtime stake claims.
-		addr := staking.NewAddress(rt.EntityID)
-		generatedEscrows[addr].StakeAccumulator.AddClaimUnchecked(StakeClaimForRuntime(rt.ID), StakeThresholdsForRuntime(rt))
+		if rt.GovernanceModel == GovernanceEntity {
```
kostko, line 271: For runtime governance the required stake should be in the runtime account. Is this checked correctly in all the places (registr runtime, committee scheduler, etc.)? I think some of these haven't been updated.

.changelog/3628.feature.md
```diff
@@ -0,0 +1,5 @@
+Add `update_runtime` runtime message
```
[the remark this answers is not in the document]
abukosek: Oops, fixed :)

go/registry/api/sanity_check.go
abukosek: Hmm, it's possible, I'll double-check.

go/roothash/api/message/message.go
```diff
@@ -57,3 +61,23 @@ func (sm *StakingMessage) ValidateBasic() error {
...
+	UpdateRuntime *registry.Runtime `json:"update_runtime,omitempty"`
+}
+
+// ValidateBasic performs basic validation of the runtime message.
+func (rm *RegistryMessage) ValidateBasic() error {
+	switch {
+	case rm.UpdateRuntime != nil:
+		// TODO: Should we validate the given runtime descriptor here?
+		// I think it's probably not necessary, since it will be validated in
+		// registerRuntime in the registry app, when it processes the message.
```
kostko, line 78: Probably no need.

runtime/src/consensus/registry.rs
```diff
@@ -4,27 +4,243 @@
...
+    },
+    consensus::staking,
     storage::mkvs::WriteLog,
 };
-use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize_repr, Deserialize_repr)]
+#[repr(u32)]
+pub enum RuntimeKind {
```
kostko, line 27: Add documentation comments for all of the new types.

runtime/src/consensus/staking.rs
```diff
@@ -21,3 +22,22 @@ pub struct Withdraw {
     pub from: Address,
     pub amount: Quantity,
 }
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize_repr, Deserialize_repr)]
+#[repr(i32)]
+pub enum ThresholdKind {
```
kostko, line 28: Add documentation comment.

runtime/src/consensus/registry.rs
```diff
@@ -4,27 +4,243 @@
...
+    #[serde(rename = "consensus-rpc")]
+    RoleConsensusRPC = 1 << 4,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct EntityWhitelistRuntimeAdmissionPolicy {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "entities")]
+    #[serde(default)]
+    pub entities: Option<BTreeMap<PublicKey, EntityWhitelistConfig>>,
```
ptrus, line 119: I think you could avoid the `Option` wrapper here and instead do:
```
#[serde(skip_serializing_if = "BTreeMap::is_empty")]
pub entities: BTreeMap<PublicKey, EntityWhitelistConfig>
```
and that should match the `go` serialization. Same for `max_nodes` bellow.

runtime/src/consensus/registry.rs
```diff
@@ -4,27 +4,243 @@
...
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct VersionInfo {
+    #[serde(default)]
+    pub version: Version,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(with = "serde_bytes")]
+    #[serde(default)]
+    pub tee: Option<Vec<u8>>,
```
ptrus, line 173: I think you could avoid the `Option` wrapper here and instead do:
```
#[serde(skip_serializing_if = "Vec::is_empty")]
pub tee: Vec<u8>
```
and that should match the `go` serialization.

ptrus (commented):
Looks good :+1: just some minor comments

runtime/src/consensus/registry.rs
abukosek: I think I tried something similar in the beginning, but it didn't work in all cases -- I'll give it another try to be sure.

runtime/src/consensus/registry.rs
kostko: I think the problem is in deserialization as it can be NULL.

abukosek: Yeah, sadly this doesn't work :/

go/consensus/tendermint/apps/keymanager/keymanager.go
```diff
@@ -118,8 +118,23 @@ func (app *keymanagerApplication) onEpochChange(ctx *tmapi.Context, epoch beacon
 
 		// Suspend the runtime in case the registering entity no longer has enough stake to cover
 		// the entity and runtime deposits.
-		if !params.DebugBypassStake {
-			acctAddr := staking.NewAddress(rt.EntityID)
+		if !params.DebugBypassStake && rt.GovernanceModel != registry.GovernanceConsensus {
```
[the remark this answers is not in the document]
abukosek: Done, looks much nicer now :)

runtime/src/consensus/registry.rs
```diff
@@ -4,27 +4,323 @@
...
+        quantity,
+        version::Version,
+    },
+    consensus::staking,
     storage::mkvs::WriteLog,
 };
-use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
+
+/// RuntimeKind represents the runtime functionality.
```
kostko, line 25: No need to include the type/field name in rust doc :-)

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -663,7 +672,7 @@ func (app *registryApplication) registerRuntime( // nolint: gocyclo
 
 		switch rtToCheck.GovernanceModel {
```
kostko, line 673: This could use the `rtToCheck.StakingAddress()` now?

runtime/src/consensus/registry.rs
abukosek: Oops, fixed :)

go/consensus/tendermint/apps/registry/transactions.go
abukosek: Fixed, but I've still left a switch in the error case, so we can have more relevant error messages.

abukosek:
Thanks for all the comments, this is now ready for re-review :)

runtime/src/consensus/registry.rs
ptrus: ah yeah, i was only thinking about serialization

go/registry/api/runtime.go
```diff
@@ -498,6 +498,22 @@ func (r *Runtime) IsCompute() bool {
 	return r.Kind == KindCompute
 }
 
+// StakingAddress returns the correct staking address for the runtime based
+// on its governance model or nil if there is no staking address under the
+// given governance model.
+func (r *Runtime) StakingAddress() *staking.Address {
+	var acctAddr staking.Address
```
ptrus, line 505: Minor: you can get rid of this variable and just return directly from the switch cases

ptrus (commented):
Thanks for the change. Looks good to me (will defer to Jernej for a final review)!

go/registry/api/runtime.go
abukosek: Compiler says no:
```
registry/api/runtime.go:508:10: cannot take the address of "github.com/oasisprotocol/oasis-core/go/staking/api".NewAddress(r.EntityID)
```
:D

ptrus: ah, nvm then :D

runtime/src/consensus/registry.rs
```diff
@@ -4,27 +4,324 @@
...
+        },
+        namespace::Namespace,
+        quantity,
+        version::Version,
+    },
+    consensus::staking,
     storage::mkvs::WriteLog,
 };
-use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
```
kostko, line 23: Move this to the top (before the serde imports) into a new group of imports (like we do elsewhere).

runtime/src/common/version.rs
```diff
@@ -1,12 +1,19 @@
+use num_traits::Zero;
```
kostko, line 1: This is in a strange place.

runtime/src/common/version.rs
```diff
@@ -1,12 +1,19 @@
...
 /// Protocol and runtime versioning.
```
kostko, line 2: This should be at the top and changed to:
```suggestion
//! Protocol and runtime versioning.
```

runtime/src/consensus/registry.rs
abukosek: Fixed :)  I thought that `make fmt` would sort the imports for me, but I guess we only have that on the Go side.

runtime/src/common/version.rs
abukosek: Fixed!  This comment already existed before, though.

runtime/src/common/version.rs
abukosek: I moved it a bit down before the serde stuff.

kostko (approved):

PR #3640, opened 2021-01-19
runtime node slashing (ADR 0005)

ptrus:
- [x] roothash/api: add evidence types
- [x] tendermint/roothash: add tx endpoint for submitting evidence of equivocation
- [x] roothash: add tests for the new evidence submission 
- [x] roothash/pool: slash non-majority for `SlashRuntimeIncorrectResults`
- [x] e2e tests
- [x] configurable amount of rewards should be awarded to nodes contributing evidence (or resolving discrepancies), add per runtime staking parameters
- [x] rewards to nodes should be escrowed (currently these are awarded into the general balance)

### DIFF
--- a/docs/adr/0005-runtime-compute-slashing.md
+++ b/docs/adr/0005-runtime-compute-slashing.md
@@ -175,13 +175,13 @@ Any slashing instructions related to freezing nodes are currently ignored.
 This proposal introduces/updates the following consensus state in the roothash
 module:
 
-- **List of past valid evidence (`0x23`)**
+- **List of past valid evidence (`0x24`)**
 
   A hash uniquely identifying the evidence is stored for each successfully
   processed evidence that has not yet expired using the following key format:
 
   ```
-  0x23 <H(runtime-id) (hash.Hash)> <round (uint64)> <evidence-hash (hash.Hash)>
+  0x24 <H(runtime-id) (hash.Hash)> <round (uint64)> <evidence-hash (hash.Hash)>
   ```
 
   The value is empty as we only need to detect duplicate evidence.
@@ -293,7 +293,7 @@ performed to verify evidence validity:
 - Public keys of signers of both commitments are compared. If they are not the
   same, the evidence is invalid.
 
-- Signatures of both proposed batches are compared. If either is invalid, the
+- Signatures of both proposed batches are validated. If either is invalid, the
   evidence is invalid.
 
 - Otherwise the evidence is valid.
--- a/docs/adr/0005-runtime-compute-slashing.md
+++ b/docs/adr/0005-runtime-compute-slashing.md
@@ -50,6 +50,14 @@ type RuntimeStakingParameters struct {
 
     // Slashing are the per-runtime misbehavior slashing parameters.
     Slashing map[staking.SlashReason]staking.Slash `json:"slashing,omitempty"`
+
+    // RewardSlashEquvocationRuntimePercent is the percentage of the reward obtained when slashing
+    // for equivocation that is transferred to the runtime's account.
+    RewardSlashEquvocationRuntimePercent uint8 `json:"reward_equivocation,omitempty"`
+
+    // RewardSlashBadResultsRuntimePercent is the percentage of the reward obtained when slashing
+    // for incorrect results that is transferred to the runtime's account.
+    RewardSlashBadResultsRuntimePercent uint8 `json:"reward_bad_results,omitempty"`
 }
 ```
 
--- a/go/consensus/tendermint/apps/roothash/roothash.go
+++ b/go/consensus/tendermint/apps/roothash/roothash.go
@@ -132,6 +132,20 @@ func (app *rootHashApplication) onCommitteeChanged(ctx *tmapi.Context, state *ro
 			return fmt.Errorf("failed to fetch runtime state: %w", err)
 		}
 
+		// Expire past evidence of runtime node misbehaviour.
+		if rtState.CurrentBlock != nil {
+			if round := rtState.CurrentBlock.Header.Round; round > params.MaxEvidenceAge {
+				ctx.Logger().Debug("removing expired runtime evidence",
+					"runtime", rt.ID,
+					"round", round,
+					"max_evidence_age", params.MaxEvidenceAge,
+				)
+				if err = state.RemoveExpiredEvidence(ctx, rt.ID, round-params.MaxEvidenceAge); err != nil {
+					return fmt.Errorf("failed to remove expired runtime evidence: %s %w", rt.ID, err)
+				}
+			}
+		}
+
 		// Since the runtime is in the list of active runtimes in the registry we
 		// can safely clear the suspended flag.
 		rtState.Suspended = false
@@ -359,6 +373,13 @@ func (app *rootHashApplication) ExecuteTx(ctx *tmapi.Context, tx *transaction.Tr
 		}
 
 		return app.executorProposerTimeout(ctx, state, &xc)
+	case roothash.MethodEvidence:
+		var ev roothash.Evidence
+		if err := cbor.Unmarshal(tx.Body, &ev); err != nil {
+			return err
+		}
+
+		return app.submitEvidence(ctx, state, &ev)
 	default:
 		return roothash.ErrInvalidArgument
 	}
--- a/go/consensus/tendermint/apps/roothash/roothash.go
+++ b/go/consensus/tendermint/apps/roothash/roothash.go
@@ -9,6 +9,7 @@ import (
 	beacon "github.com/oasisprotocol/oasis-core/go/beacon/api"
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
 	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"
 	tmapi "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/api"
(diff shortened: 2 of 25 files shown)


go/consensus/tendermint/apps/roothash/slashing.go
```diff
@@ -0,0 +1,89 @@
...
+		ctx.Logger().Error("failed to get runtime node by signature public key",
+			"public_key", pk,
+			"err", err,
+		)
+		// TODO: Should this be treated as an error?
+		return fmt.Errorf("node not found: %w", roothash.ErrInvalidEvidence)
+	}
+
+	// Retrieve the slash procedure for double signing.
+	st, err := stakeState.Slashing(ctx)
```
kostko, line 39: This should use the per-runtime slashing procedure.

ptrus: Yeah of course, thanks! Will also update the transactions test to ensure the expected amount was slashed (which would've found this).

go/consensus/tendermint/apps/roothash/transactions.go
```diff
@@ -267,3 +268,116 @@ func (app *rootHashApplication) executorCommit(
 
 	return nil
 }
+
+// TODO: evidence -> submitEvidence?
```
kostko, line 272: 👍

go/staking/api/slashing.go
```diff
@@ -11,18 +11,33 @@ import (
 type SlashReason uint8
 
 const (
-	// SlashConsensusEquivocation is slashing due to equivocation.
+	// SlashConsensusEquivocation is slashing due to double signing.
```
kostko, line 14: Hm?

ptrus: reverted this change (snuck in here after extracting these changes into the separate general changes PR)

go/consensus/tendermint/apps/roothash/state/state.go
```diff
@@ -144,6 +150,12 @@ func (s *ImmutableState) ConsensusParameters(ctx context.Context) (*roothash.Con
 	return &params, nil
 }
 
+// EvidenceHashExists returns true if the evidence hash for the runtime exists.
+func (s *ImmutableState) EvidenceHashExists(ctx context.Context, runtimeID common.Namespace, round uint64, hash *hash.Hash) (bool, error) {
```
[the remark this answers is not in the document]
ptrus: Yeah no need, fixed (also elsewhere)

go/consensus/tendermint/apps/roothash/slashing.go
```diff
@@ -0,0 +1,164 @@
...
+	// TODO: should probably be escrowed.
+	var rewardEntities []signature.PublicKey
+	for _, pk := range discrepancyResolvers {
+		node, err := regState.Node(ctx, pk)
+		if err != nil {
+			ctx.Logger().Error("failed to get runtime node by commitment signature public key",
+				"public_key", pk,
+				"err", err,
+			)
+			// Don't fail if node for whatever reason couldn't be found?
```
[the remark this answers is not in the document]
ptrus: Fixed the missing continue and changed this to error in case the error is not `ErrNodeNotFound` (also above).

go/consensus/tendermint/apps/roothash/roothash.go
```diff
@@ -501,6 +523,58 @@ func (app *rootHashApplication) tryFinalizeExecutorCommits(
...
+					// Worker nodes that submitted incorrect results get slashed.
+					workersIncorrectResults []signature.PublicKey
+					// Backup worker nodes that resolved the discrepancy get rewarded.
+					backupCorrectResults []signature.PublicKey
+				)
+				for _, n := range rtState.ExecutorPool.Committee.Members {
+					c, ok := commitments[n.PublicKey]
+					if !ok {
+						continue
+					}
```
kostko, line 543: This should also ignore commitments which indicate failure.

go/roothash/api/api.go
```diff
@@ -142,6 +156,150 @@ func NewRequestProposerTimeoutTx(nonce uint64, fee *transaction.Fee, runtimeID c
...
+
+// EquivocationExecutorEvidence is evidence of executor commitment equivocation.
+type EquivocationExecutorEvidence struct {
+	CommitA commitment.ExecutorCommitment `json:"commit_a"`
+	CommitB commitment.ExecutorCommitment `json:"commit_b"`
+}
+
+// ValidateBasic performs basic executor evidence validation checks.
+// TODO: maybe rename to: Validate(), might better indicate that this actually
+// checks if evidence is valid and is not just a basic check.
```
pro-wh, line 212: shall we? from reading the rest of this PR, we notably don't check that it's recent

go/roothash/api/api.go
```diff
@@ -142,6 +156,150 @@ func NewRequestProposerTimeoutTx(nonce uint64, fee *transaction.Fee, runtimeID c
...
+	if err != nil {
+		return fmt.Errorf("opening CommitA: %w", err)
+	}
+	b, err := ev.CommitB.Open()
+	if err != nil {
+		return fmt.Errorf("opening CommitB: %w", err)
+	}
+
+	if a.Body == nil {
+		return fmt.Errorf("CommitA: body empty")
```
pro-wh, line 224: is this possible? It looks like not. but if it becomes possible, which is more likely, that it's the evidence submitter's fault or that it's the signer's fault? here we return an error, which otherwise seems to tell the caller something happened that was the submitter's fault

go/consensus/tendermint/apps/roothash/slashing.go
```diff
@@ -43,11 +44,155 @@ func onEvidenceRuntimeEquivocation(
 		return fmt.Errorf("tendermint/roothash: nothing to slash from account %s", entityAddr)
 	}
 
-	// Move slashed amount to the runtime account.
-	// TODO: part of slashed amount (configurable) should be transferred to the transaction submitter.
+	// If the caller is a node, distribute slashed funds to the controlling entity instead of the
+	// caller directly.
+	rewardAddr := ctx.CallerAddress()
```
pro-wh, line 49: I wonder if people frontrun transactions like these, to take the reward for themselves

go/oasis-node/cmd/registry/runtime/runtime.go
```diff
@@ -539,6 +560,7 @@ func init() {
 
 	// Init Staking flags.
 	runtimeFlags.StringToString(CfgStakingThreshold, nil, "Additional staking threshold for this runtime (<kind>=<value>)")
+	runtimeFlags.StringToString(CfgStakingSlashing, nil, "Staking slashing parameter for this runtime (<kind>=<value>)")
```
pro-wh, line 563: `<kind>` -> `<reason>`?

go/roothash/api/commitment/pool.go
```diff
@@ -263,6 +263,8 @@ func (p *Pool) addOpenExecutorCommitment(
 		return ErrTxnSchedSigInvalid
 	}
 
+	// TODO: check for evidence of equivocation.
```
pro-wh, line 266: shall we file this?

pro-wh (approved):

docs/adr/0005-runtime-compute-slashing.md
```diff
@@ -175,13 +175,13 @@ Any slashing instructions related to freezing nodes are currently ignored.
 This proposal introduces/updates the following consensus state in the roothash
 module:
 
-- **List of past valid evidence (`0x23`)**
+- **List of past valid evidence (`0x24`)**
```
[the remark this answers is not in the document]
kostko: That was because when the ADR was written the new beacon wasn't in yet :)

go/consensus/tendermint/apps/roothash/slashing.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	// Slash runtime node entity
+	entityAddr := staking.NewAddress(node.EntityID)
+	totalSlashed, err := stakeState.SlashEscrow(ctx, entityAddr, penaltyAmount)
+	if err != nil {
+		return fmt.Errorf("tendermint/roothash: error slashing account %s: %w", entityAddr, err)
+	}
+	if totalSlashed.IsZero() {
+		return fmt.Errorf("tendermint/roothash: nothing to slash from account %s", entityAddr)
```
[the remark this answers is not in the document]
kostko: Nice catch, this shouldn't error.

go/roothash/api/api.go
kostko: If `Body` was `nil` this would actually indicate a bug in the `Open` logic. So this check indeed does not make sense and I'll remove it.

go/roothash/api/api.go
kostko: Yeah, I'll keep it as `ValidateBasic` and add a comment that says this excludes stateful checks for expiry.

go/consensus/tendermint/apps/roothash/slashing.go
kostko: Yeah that's a valid concern. Incentivizing evidence submission seems like it introduces more problems than it solves :-)

go/oasis-node/cmd/registry/runtime/runtime.go
kostko: :+1:

go/roothash/api/commitment/pool.go
kostko: Filed https://github.com/oasisprotocol/oasis-core/issues/3685.

go/consensus/tendermint/apps/roothash/transactions.go
```diff
@@ -267,3 +268,115 @@ func (app *rootHashApplication) executorCommit(
...
+	if err = ctx.Gas().UseGas(1, roothash.GasOpEvidence, params.GasCosts); err != nil {
+		return err
+	}
+
+	rtState, _, _, err := app.getRuntimeState(ctx, state, evidence.ID)
+	if err != nil {
+		return err
+	}
+
+	if len(rtState.Runtime.Staking.Slashing) == 0 {
```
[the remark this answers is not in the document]
kostko: But it could be nil and derefing a nil map is an error. In case of `Slashing` from the (global) staking consensus parameters, the `Slashing()` state accessor method returns a non-nil map in case the parameters are unset/nil but here this is not the case.

go/registry/api/runtime.go
```diff
@@ -263,10 +263,27 @@ type RuntimeStakingParameters struct {
 	// In case a node is registered for multiple runtimes, it will need to satisfy the maximum
 	// threshold of all the runtimes.
 	Thresholds map[staking.ThresholdKind]quantity.Quantity `json:"thresholds,omitempty"`
+
+	// Slashing are the per-runtime misbehavior slashing parameters.
+	Slashing map[staking.SlashReason]staking.Slash `json:"slashing,omitempty"`
+
+	// RewardSlashEquvocationRuntimePercent is the percentage of the reward obtained when slashing
```
ptrus, line 270: Should we update the ADR with these added params?

go/consensus/tendermint/apps/roothash/slashing.go
```diff
@@ -0,0 +1,204 @@
...
+	// Runtime account reward.
+	runtimeAccReward := totalSlashed.Clone()
+	if err := runtimeAccReward.Mul(quantity.NewFromUint64(runtimePercentage)); err != nil {
+		return fmt.Errorf("tendermint/roothash: runtimeAccReward.Mul: %w", err)
+	}
+	if err := runtimeAccReward.Quo(quantity.NewFromUint64(uint64(100))); err != nil {
+		return fmt.Errorf("tendermint/roothash: runtimeAccReward.Quo(100): %w", err)
+	}
+	runtimeAddr := staking.NewRuntimeAddress(runtimeID)
+	if _, err := stakeState.TransferFromCommon(ctx, runtimeAddr, runtimeAccReward, false); err != nil {
```
ptrus, line 168: Currently these rewards ignore the commission schedule and the whole reward is split between all delegators of the entity, is this fine or should the owning entity be able to take commission?

ptrus: (note: this comment applies to line 193, here we transfer to the runtime account anyway)

kostko: Hm, actually currently the delegators get nothing, it is treated the same as if the entity just increased its self-delegation. Do we want the delegators to get a cut as well?

ptrus: Oh right yeah i miss-checked that. I think that probably yes? (since if the the node is slashed, delegators do lose proportionately). But i'm not sure if it's currently worth the additional complexity. Maybe a good future improvement. I would at least add some notes that runtime slashing rewards are not rewarded to delegators.

kostko: Makes sense, I changed the code to take all delegators and commission into account.

go/registry/api/runtime.go
kostko: Updated.

go/consensus/tendermint/apps/staking/state/state.go
```diff
@@ -717,27 +719,90 @@ func (s *MutableState) TransferFromCommon(
 	if err != nil {
 		return false, fmt.Errorf("tendermint/staking: failed to transfer from common pool: %w", err)
 	}
+	if transferred.IsZero() {
+		// Common pool has been depleated, nothing to transfer.
+		return false, nil
+	}
+
+	// If escrow is requested, escrow the transferred stake immediately.
+	if escrow {
```
ptrus, line 728: Hm I think (not 100% sure, didn't check in detail) we can now also simplify the `AddRewards` method by calling `TransferFromCommon` from the accounts loop.

go/consensus/tendermint/apps/staking/state/state.go
```diff
@@ -703,12 +703,14 @@ func (s *MutableState) TransferFromCommon(
 	ctx *abciAPI.Context,
 	toAddr staking.Address,
 	amount *quantity.Quantity,
+	escrow bool,
```
ptrus, line 706: Maybe update documentation above with the added field and that it takes into account commission schedule.

ptrus (commented):
:heavy_check_mark: Thanks for finishing the PR (can't actually approve as i'm the author).

go/consensus/tendermint/apps/staking/state/state.go
kostko: Good idea, updated.

go/consensus/tendermint/apps/roothash/transactions.go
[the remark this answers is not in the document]
pro-wh: https://golang.org/ref/spec#Index_expressions language spec says reading from a nil map gives the zero value

PR #3646, opened 2021-01-21
slashing: rename DoubleSigning to ConsensusEquivocation

ptrus:
Extracted some general slashing changes from: https://github.com/oasisprotocol/oasis-core/pull/3640 (ADR-0005)

- rename `DoubleSigning`->`ConsensusEquivocation`
- change `SlashReason` type to `uint8`
- change `SlashEscrow` method to return slashed amount

### DIFF
--- a/go/consensus/tendermint/apps/staking/slashing.go
+++ b/go/consensus/tendermint/apps/staking/slashing.go
@@ -16,7 +16,7 @@ import (
 	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
 )
 
-func onEvidenceDoubleSign(
+func onEvidenceConsensusEquivocation(
 	ctx *abciAPI.Context,
 	addr tmcrypto.Address,
 	height int64,
@@ -66,7 +66,7 @@ func onEvidenceDoubleSign(
 		return err
 	}
 
-	penalty := st[staking.SlashDoubleSigning]
+	penalty := st[staking.SlashConsensusEquivocation]
 
 	// Freeze validator to prevent it being slashed again. This also prevents the
 	// validator from being scheduled in the next epoch.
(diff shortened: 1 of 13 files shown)


go/consensus/tendermint/apps/staking/state/state.go
```diff
@@ -624,55 +624,55 @@ func slashPool(dst *quantity.Quantity, p *staking.SharePool, amount, total *quan
...
 
 	// Compute the amount we need to slash each pool. The amount is split
 	// between the pools based on relative total balance.
 	total := from.Escrow.Active.Balance.Clone()
 	if err = total.Add(&from.Escrow.Debonding.Balance); err != nil {
-		return false, fmt.Errorf("tendermint/staking: compute total balance: %w", err)
+		return nil, fmt.Errorf("tendermint/staking: account total balance: %w", err)
+	}
+	// Nothing to slash.
+	if total.IsZero() {
```
ptrus, line 655: Note that before this check was bellow (after `slashPool`) calls.

However in case nothing could've been slashed (node has no stake) the `slashPool` call failed already (since it divides by `total` which would be zero), making the bellow check ineffective.

go/consensus/tendermint/apps/staking/slashing.go
```diff
@@ -96,6 +97,13 @@ func onEvidenceDoubleSign(
 		)
 		return err
 	}
+	if totalSlashed.IsZero() {
```
ptrus, line 100: Note:
Before the `stakeState.SlashEscrow` would fail in case nothing was slashed (see other comment about `slashPool` method). So this keeps the logic consistent with what it was before.

kostko (approved):

PR #3653, opened 2021-01-26
Client node support for hosting runtimes

kostko:
TODO
* [x] Move runtime host configuration to the runtime registry.
* [x] Support the client node hosting runtimes.
* [x] Support for provisioner remapping (a client node may use a non-TEE version of the runtime to perform local queries for public state against a runtime which runs in a TEE on executor nodes).
* [x] Describe all configuration changes.

### DIFF
--- a/go/runtime/client/host.go
+++ b/go/runtime/client/host.go
@@ -1,0 +1,174 @@
+package client
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/opentracing/opentracing-go"
+
+	"github.com/oasisprotocol/oasis-core/go/common/logging"
+	consensus "github.com/oasisprotocol/oasis-core/go/consensus/api"
+	"github.com/oasisprotocol/oasis-core/go/runtime/host"
+	"github.com/oasisprotocol/oasis-core/go/runtime/host/protocol"
+	runtimeRegistry "github.com/oasisprotocol/oasis-core/go/runtime/registry"
+	storage "github.com/oasisprotocol/oasis-core/go/storage/api"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/syncer"
+)
+
+var (
+	errMethodNotSupported   = errors.New("method not supported")
+	errEndpointNotSupported = errors.New("endpoint not supported")
+)
+
+type clientHost struct {
+	*runtimeRegistry.RuntimeHostNode
+
+	runtime   runtimeRegistry.Runtime
+	consensus consensus.Backend
+
+	stopCh chan struct{}
+	quitCh chan struct{}
+
+	logger *logging.Logger
+}
+
+// Start starts the client host.
+func (h *clientHost) Start() error {
+	go h.worker()
+	return nil
+}
+
+// Stop asks the client host to stop.
+func (h *clientHost) Stop() {
+	close(h.stopCh)
+}
+
+// Quit returns the channel which will signal when the client host has quit.
+func (h *clientHost) Quit() <-chan struct{} {
+	return h.quitCh
+}
+
+// Implements runtimeRegistry.RuntimeHostHandlerFactory.
+func (h *clientHost) GetRuntime() runtimeRegistry.Runtime {
+	return h.runtime
+}
+
+// Implements runtimeRegistry.RuntimeHostHandlerFactory.
+func (h *clientHost) NewRuntimeHostHandler() protocol.Handler {
+	return h
+}
+
+// Implements runtimeRegistry.RuntimeHostHandlerFactory.
+func (h *clientHost) NewNotifier(ctx context.Context, host host.Runtime) protocol.Notifier {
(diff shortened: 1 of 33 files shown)


go/runtime/client/host.go
```diff
@@ -0,0 +1,179 @@
...
+
+// Implements runtimeRegistry.RuntimeHostHandlerFactory.
+func (h *clientHost) NewNotifier(ctx context.Context, host host.Runtime) protocol.Notifier {
+	return nil
+}
+
+// Implements protocol.Handler.
+func (h *clientHost) Handle(ctx context.Context, body *protocol.Body) (*protocol.Body, error) {
+	// Storage.
+	if body.HostStorageSyncRequest != nil {
```
pro-wh, line 70: this outer one could be a switch too

go/runtime/client/host.go
```diff
@@ -0,0 +1,179 @@
...
+		return
+	}
+
+	// Start hosted runtime provisioning.
+	hrt, _, err := h.ProvisionHostedRuntime(context.Background())
+	switch {
+	case err == nil:
+	case errors.Is(err, runtimeRegistry.ErrRuntimeHostNotConfigured):
+		// Runtime host not configured, do nothing and wait to stop.
+		<-h.stopCh
```
pro-wh, line 139: how do we get in this situation?

go/runtime/host/helpers.go
```diff
@@ -0,0 +1,67 @@
...
+
+var (
+	ErrInvalidArgument = errors.New("runtime: invalid argument")
+	ErrInternal        = errors.New("runtime: internal error")
+	ErrCheckTxFailed   = errors.New("runtime: check tx failed")
+)
+
+// RichRuntime provides higher-level functions for talking with a runtime.
+type RichRuntime interface {
+	Runtime
```
pro-wh, line 25: what are the cases where we'll use Runtime instead of RichRuntime?

go/runtime/client/host.go
```diff
@@ -0,0 +1,179 @@
...
+	return h
+}
+
+// Implements runtimeRegistry.RuntimeHostHandlerFactory.
+func (h *clientHost) NewNotifier(ctx context.Context, host host.Runtime) protocol.Notifier {
+	return nil
+}
+
+// Implements protocol.Handler.
+func (h *clientHost) Handle(ctx context.Context, body *protocol.Body) (*protocol.Body, error) {
```
pro-wh, line 68: how does this end up getting used? I don't have any experience with RuntimeHostHandlerFactory.

go/runtime/client/host.go
kostko: So currently we create a new "client host" for each runtime regardless of whether the hosted runtime is actually configured (e.g., if the client node has specified the path to the runtime binary in the config). But thinking about it, this doesn't make much sense and we should just do the check early and not create this instance at all if the hosted runtime is not configured.

go/runtime/host/helpers.go
kostko: I was thinking about that as well. Probably `RichRuntime` would mostly be used and it could just be merged into `Runtime`. But we also wouldn't want each `Runtime` implementation to implement all of these wrappers.  Maybe we could just provide a `BaseRuntime`?

go/runtime/client/host.go
kostko: This gets called whenever the runtime calls a method on the host. On the client node we only support storage queries.

go/runtime/client/host.go
pro-wh: That new idea sounds good. What was the nature of "each runtime"? Every single runtime registered on the network, or is there some other list somewhere?

go/runtime/host/helpers.go
pro-wh: Okay yeah. I guess since we don't have plans to have a different implementation of CheckTx, all we have to gain that way is the code being more direct.

go/runtime/client/host.go
kostko: Oh, no so each runtime that is enabled in the configuration. So you configure two things:
* One is whether the client node will track the given runtime (e.g., blocks emitted by the runtime) and be able to submit transactions. To do that you just need to configure a runtime ID in the client node's configuration.
* The other is whether the client node can also host a runtime (e.g., has the runtime binary available locally). To do that you need to have the runtime binary available locally and be able to run it (host it).

kostko: Changed.

go/runtime/client/host.go
kostko: Good idea, changed.

go/runtime/host/helpers.go
pro-wh: feel free to split this out

pro-wh (approved):
thanks for the explanations. lg

PR #3656, opened 2021-01-27
ADR 0008: Key Derivation Process for Entity Keys

tjanez:


Esya:
I am currently implementing BIP32-ED25519 in https://github.com/oasisprotocol/oasis-sdk

One possible downside, is that as the paper mentions :

> Due to the way child
Ed25519 keys are derived, they will not have the 32-
byte master secret key. We note that the NaCl library [3]
provides the necessary interface to the extended private
key.

The ed25519 libraries will **need to support instanciation from an extended private key (64 bytes)** and not just from the 32-bytes seed in order to use the derived children keys.

kostko:
You now have two PRs open, should this one be closed?

tjanez:
> I am currently implementing BIP32-ED25519 in https://github.com/oasisprotocol/oasis-sdk
> 
> One possible downside, is that as the paper mentions :
> 
> > Due to the way child
> > Ed25519 keys are derived, they will not have the 32-
> > byte master secret key. We note that the NaCl library [3]
> > provides the necessary interface to the extended private
> > key.
> 
> The ed25519 libraries will **need to support instanciation from an extended private key (64 bytes)** and not just from the 32-bytes seed in order to use the derived children keys.

Yeah, this would be just one downside to using BIP32-Ed25519. We've switched to SLIP-0010 which avoids a spectrum of issues when trying to support non-hardened public parent key to public child key derivation with Ed25519 by limiting the derivation to only hardened private parent key to private child key derivation.

tjanez:
> You now have two PRs open, should this one be closed?

Yeah, sorry for the confusion, I'll remove the ADR 0008 spec from #3918. I think it will be easier to review/approve just the spec in this PR.

kostko (approved):

kostko (commented):
Don't forget to update the ADR index!

Yawning (approved):

pro-wh (approved):

peterjgilbert (approved):

PR #3662, opened 2021-01-29
runtime: Add support for runtime queries

kostko:
Based on #3653

### DIFF
--- a/runtime/src/dispatcher.rs
+++ b/runtime/src/dispatcher.rs
@@ -9,7 +9,7 @@ use std::{
     thread,
 };
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Result as AnyResult};
 use crossbeam::channel;
 use io_context::Context;
 use slog::Logger;
@@ -23,7 +23,9 @@ use crate::{
         },
         logger::get_logger,
     },
-    consensus::roothash::{self, Block, ComputeResultsHeader, COMPUTE_RESULTS_HEADER_CONTEXT},
+    consensus::roothash::{
+        self, Block, ComputeResultsHeader, Header, COMPUTE_RESULTS_HEADER_CONTEXT,
+    },
     enclave_rpc::{
         demux::Demux as RpcDemux,
         dispatcher::Dispatcher as RpcDispatcher,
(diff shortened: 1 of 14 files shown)


runtime/src/dispatcher.rs
```diff
@@ -45,11 +47,11 @@ use crate::{
         types::TxnBatch,
         Context as TxnContext,
     },
-    types::{Body, ComputedBatch, HostStorageEndpoint},
+    types::{Body, ComputedBatch, Error, HostStorageEndpoint},
 };
 
 /// Maximum amount of requests that can be in the dispatcher queue.
-const BACKLOG_SIZE: usize = 10;
+const BACKLOG_SIZE: usize = 1000;
```
pro-wh, line 54: could you add a summary of why there are so many changes to dispatcher.rs

runtime/src/protocol.rs
```diff
@@ -206,11 +206,7 @@ impl Protocol {
                         // is no need to do anything more.
                         return Ok(());
                     }
-                    Err(error) => Body::Error {
-                        module: "".to_owned(), // XXX: Error codes.
-                        code: 1,               // XXX: Error codes.
-                        message: format!("{}", error),
-                    },
+                    Err(error) => Body::Error(Error::new("dispatcher", 1, &format!("{}", error))),
```
pro-wh, line 209: could you add a note for developers on the change in style in protocol.rs

pro-wh (commented):
might it also make sense to expose CheckTx this way?

kostko:
> might it also make sense to expose CheckTx this way?

That's a very good idea.

runtime/src/protocol.rs
kostko: See above comment, will add a changelog entry for this part.

runtime/src/dispatcher.rs
kostko: Some minor refactoring to avoid repetition with sending responses to various requests back. Previously each handler had its own `send_response` but there's really no need for that.

kostko: That is just some minor refactoring to avoid repeating sending back responses from handlers. Previously each handler had its own `send_response` but there's really no need for that.

pro-wh: ah thanks

kostko:
@pro-wh Also exposed `CheckTx` now.

pro-wh (approved):
thanks for the changes, lg

PR #3672, opened 2021-02-02
go/consensus/tendermint/abci/mux: Reject early EstimateGas calls

Yawning:
Fixes 3669

### DIFF
--- a/go/consensus/tendermint/abci/mux.go
+++ b/go/consensus/tendermint/abci/mux.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"sort"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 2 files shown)


go/consensus/tendermint/abci/mux.go
```diff
@@ -574,6 +576,12 @@ func (mux *abciMux) EstimateGas(caller signature.PublicKey, tx *transaction.Tran
 		return 0, consensus.ErrInvalidArgument
 	}
 
+	// Certain modules, in particular the beacon require InitChain or BeginBlock
+	// to have completed before initialization is complete.
+	if mux.lastBeginBlock == blockHeightInvalid {
```
kostko, line 581: Does this require a mutex/atomic since `EstimateGas` can be called in parallel to `BeginBlock`?

Yawning: think that variable is updated at the top of begin block too, kind of annoying

kostko (approved):

PR #3676, opened 2021-02-03
go/beacon: Backport the beacon key generation/registration

### LINKED ISSUE #3674: Backport beacon key generation/registration
The beacon key generation and registration should be backported to the `stable/20.12.x` branch.

Yawning:
To prepare for the future migration to a PVSS based beacon scheme, every
node should generate and register a elliptic curve point in advance of
the migration.  This commit selectively backports the required logic for
a hopefully smooth transition.

Fixes #3674

### DIFF
--- a/go/common/node/node.go
+++ b/go/common/node/node.go
@@ -14,6 +14,7 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/pvss"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
 	"github.com/oasisprotocol/oasis-core/go/common/prettyprint"
 	"github.com/oasisprotocol/oasis-core/go/common/sgx/ias"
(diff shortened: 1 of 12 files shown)


go/registry/api/api.go
```diff
@@ -726,6 +730,24 @@ func VerifyRegisterNodeArgs( // nolint: gocyclo
 		return nil, nil, fmt.Errorf("%w: duplicate node TLS public key", ErrInvalidArgument)
 	}
 
+	if n.Beacon != nil {
```
kostko, line 733: This is not backwards compatible.

go/consensus/tendermint/apps/registry/state/state.go
```diff
@@ -608,6 +632,21 @@ func (s *MutableState) SetNode(ctx context.Context, existingNode, node *node.Nod
 		return abciAPI.UnavailableStateError(err)
 	}
 
+	// Beacon point.
```
kostko, line 635: This is not backwards compatible.

go/common/node/node.go
```diff
@@ -71,8 +72,9 @@ type Node struct { // nolint: maligned
 	// Beacon contains information for this node's participation
 	// in the random beacon protocol.
 	//
-	// NOTE: This is reserved for future use.
-	Beacon cbor.RawMessage `json:"beacon,omitempty"`
+	// TODO: This is optional for now, make mandatory once enough
+	// nodes provide this field.
+	Beacon *BeaconInfo `json:"beacon,omitempty"`
```
kostko, line 77: This probably needs to stay unchanged as well.

Yawning: The more I think about this, the more I'm vehemently against adding something that sticks on-chain data that can't be validated.

kostko: > The more I think about this, the more I'm vehemently against adding something that sticks on-chain data that can't be validated.

But that's already there. If nodes have invalid data in it at the time of the upgrade, they will be removed from genesis.

kostko (approved):

PR #3678, opened 2021-02-04
go/oasis-node/cmd/common/metrics: Fix label escaping

kostko:


go/oasis-node/cmd/common/metrics/metrics.go
```diff
@@ -272,7 +272,14 @@ func Enabled() bool {
 
 // EscapeLabelCharacters replaces invalid prometheus label name characters with "_".
 func EscapeLabelCharacters(l string) string {
-	return strings.Replace(l, ".", "_", -1)
+	return strings.Map(func(r rune) rune {
```
gw0, line 275: Looks better, but what do you think about an allowlist approach? Prometheus docs say that a user-provided label name must match `[a-zA-Z][a-zA-Z0-9_]*` (https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels). So, a regexp replace like `s/[^a-zA-Z0-9_]/_/g` is a rough quick fix to catch all invalid characters. To also fix the first character something like `s/^[^a-zA-Z]+//`.

gw0 (approved):
It fixes the issue at hand, but maybe an allowlist approach would be better?

Yawning (approved):

PR #3689, opened 2021-02-10
Add runtime scheduling constraints

### LINKED ISSUE #3644: Additional runtime eligibility constraints
We should add support for runtime eligibility constraints, separately for primary and backup nodes:
* Validator set membership.
* Maximum number of nodes per entity per role.

These constraints could be optionally enabled for a runtime and would not prevent registration but would determine what nodes are eligible for scheduling.

kostko:
Fixes #3644 

TODO
* [x] Actually implement the new constraints (validator set membership, max nodes per entity).
* [x] Fix runtime descriptor generation CLI to properly support constraints.
* [x] Basic validation of constraints structure.
* [x] Tests.
* [x] Cleanup.

### DIFF
--- a/go/consensus/tendermint/apps/roothash/roothash.go
+++ b/go/consensus/tendermint/apps/roothash/roothash.go
@@ -541,8 +541,34 @@ func (app *rootHashApplication) tryFinalizeExecutorCommits(
 ) error {
 	runtime := rtState.Runtime
 	round := rtState.CurrentBlock.Header.Round + 1
+	pool := rtState.ExecutorPool
 
-	commit, err := rtState.ExecutorPool.TryFinalize(ctx.BlockHeight(), runtime.Executor.RoundTimeout, forced, true)
+	commit, err := pool.TryFinalize(ctx.BlockHeight(), runtime.Executor.RoundTimeout, forced, true)
+	if err == commitment.ErrDiscrepancyDetected {
+		ctx.Logger().Warn("executor discrepancy detected",
+			"round", round,
+			logging.LogEvent, roothash.LogEventExecutionDiscrepancyDetected,
+		)
+
+		tagV := ValueExecutionDiscrepancyDetected{
+			ID: runtime.ID,
+			Event: roothash.ExecutionDiscrepancyDetectedEvent{
+				Timeout: forced,
+			},
+		}
+		ctx.EmitEvent(
+			tmapi.NewEventBuilder(app.Name()).
+				Attribute(KeyExecutionDiscrepancyDetected, cbor.Marshal(tagV)).
+				Attribute(KeyRuntimeID, ValueRuntimeID(runtime.ID)),
+		)
+
+		// We may also be able to already perform discrepancy resolution, check if this is possible
+		// by retrying finalization. We must make sure to not affect the computed timeout.
+		nextTimeout := pool.NextTimeout
+		commit, err = pool.TryFinalize(ctx.BlockHeight(), runtime.Executor.RoundTimeout, false, false)
+		pool.NextTimeout = nextTimeout
+	}
+
 	switch err {
 	case nil:
 		// Round has been finalized.
@@ -559,19 +585,19 @@ func (app *rootHashApplication) tryFinalizeExecutorCommits(
 		}
 
 		// If there was a discrepancy, slash nodes for incorrect results if configured.
-		if rtState.ExecutorPool.Discrepancy {
+		if pool.Discrepancy {
 			ctx.Logger().Debug("executor pool discrepancy",
 				"slashing", runtime.Staking.Slashing,
 			)
 			if penalty, ok := rtState.Runtime.Staking.Slashing[staking.SlashRuntimeIncorrectResults]; ok && !penalty.Amount.IsZero() {
-				commitments := rtState.ExecutorPool.ExecuteCommitments
+				commitments := pool.ExecuteCommitments
 				var (
 					// Worker nodes that submitted incorrect results get slashed.
 					workersIncorrectResults []signature.PublicKey
 					// Backup worker nodes that resolved the discrepancy get rewarded.
 					backupCorrectResults []signature.PublicKey
 				)
-				for _, n := range rtState.ExecutorPool.Committee.Members {
+				for _, n := range pool.Committee.Members {
 					c, ok := commitments[n.PublicKey]
 					if !ok || c.IsIndicatingFailure() {
 						continue
@@ -617,7 +643,7 @@ func (app *rootHashApplication) tryFinalizeExecutorCommits(
 		blk.Header.MessagesHash = *hdr.MessagesHash
 
 		// Timeout will be cleared by caller.
-		rtState.ExecutorPool.ResetCommitments(blk.Header.Round)
+		pool.ResetCommitments(blk.Header.Round)
 
 		// All good. Hook up the new block.
 		rtState.CurrentBlock = blk
@@ -644,23 +670,7 @@ func (app *rootHashApplication) tryFinalizeExecutorCommits(
 
 		return nil
 	case commitment.ErrDiscrepancyDetected:
-		// Discrepancy has been detected.
-		ctx.Logger().Warn("executor discrepancy detected",
-			"round", round,
-			logging.LogEvent, roothash.LogEventExecutionDiscrepancyDetected,
-		)
-
-		tagV := ValueExecutionDiscrepancyDetected{
-			ID: runtime.ID,
-			Event: roothash.ExecutionDiscrepancyDetectedEvent{
-				Timeout: forced,
-			},
-		}
-		ctx.EmitEvent(
-			tmapi.NewEventBuilder(app.Name()).
-				Attribute(KeyExecutionDiscrepancyDetected, cbor.Marshal(tagV)).
-				Attribute(KeyRuntimeID, ValueRuntimeID(runtime.ID)),
-		)
+		// This was already handled above, so it should not happen.
 		return nil
 	default:
 	}
--- a/go/consensus/tendermint/apps/scheduler/genesis.go
+++ b/go/consensus/tendermint/apps/scheduler/genesis.go
@@ -20,12 +20,7 @@ import (
 )
 
 func (app *schedulerApplication) InitChain(ctx *abciAPI.Context, req types.RequestInitChain, doc *genesis.Document) error {
-	baseEpoch, err := app.state.GetBaseEpoch()
-	if err != nil {
-		return fmt.Errorf("tendermint/scheduler: couldn't get base epoch: %w", err)
-	}
-	app.baseEpoch = baseEpoch
-
+	var err error
 	state := schedulerState.NewMutableState(ctx.State())
 	if err = state.SetConsensusParameters(ctx, &doc.Scheduler.Parameters); err != nil {
 		return fmt.Errorf("failed to set consensus parameters: %w", err)
--- a/go/consensus/tendermint/apps/scheduler/scheduler.go
+++ b/go/consensus/tendermint/apps/scheduler/scheduler.go
@@ -39,12 +39,13 @@ var (
 	RNGContextStorage    = []byte("EkS-ABCI-Storage")
 	RNGContextValidators = []byte("EkS-ABCI-Validators")
 	RNGContextEntities   = []byte("EkS-ABCI-Entities")
+
+	RNGContextRoleWorker       = []byte("Worker")
+	RNGContextRoleBackupWorker = []byte("Backup-Worker")
 )
 
 type schedulerApplication struct {
 	state api.ApplicationState
-
-	baseEpoch beacon.EpochTime
 }
 
 func (app *schedulerApplication) Name() string {
@@ -85,7 +86,12 @@ func (app *schedulerApplication) BeginBlock(ctx *api.Context, request types.Requ
 	if epochChanged || slashed {
 		// The 0th epoch will not have suitable entropy for elections, nor
 		// will it have useful node registrations.
-		if epoch == app.baseEpoch {
+		baseEpoch, err := app.state.GetBaseEpoch()
+		if err != nil {
+			return fmt.Errorf("tendermint/scheduler: cound't get base epoch: %w", err)
+		}
+
+		if epoch == baseEpoch {
 			ctx.Logger().Info("system in bootstrap period, skipping election",
 				"epoch", epoch,
 			)
@@ -163,8 +169,9 @@ func (app *schedulerApplication) BeginBlock(ctx *api.Context, request types.Requ
 
 		// Handle the validator election first, because no consensus is
 		// catastrophic, while failing to elect other committees is not.
+		var validatorEntities map[staking.Address]bool
 		if !params.DebugStaticValidators {
-			if err = app.electValidators(ctx, entropy, stakeAcc, entitiesEligibleForReward, nodes, params); err != nil {
+			if validatorEntities, err = app.electValidators(ctx, entropy, stakeAcc, entitiesEligibleForReward, nodes, params); err != nil {
 				// It is unclear what the behavior should be if the validator
 				// election fails.  The system can not ensure integrity, so
 				// presumably manual intervention is required...
@@ -177,7 +184,18 @@ func (app *schedulerApplication) BeginBlock(ctx *api.Context, request types.Requ
 			scheduler.KindStorage,
 		}
 		for _, kind := range kinds {
-			if err = app.electAllCommittees(ctx, request, epoch, entropy, stakeAcc, entitiesEligibleForReward, runtimes, committeeNodes, kind); err != nil {
+			if err = app.electAllCommittees(
+				ctx,
+				request,
+				epoch,
+				entropy,
+				stakeAcc,
+				entitiesEligibleForReward,
+				validatorEntities,
+				runtimes,
+				committeeNodes,
+				kind,
+			); err != nil {
 				return fmt.Errorf("tendermint/scheduler: couldn't elect %s committees: %w", kind, err)
 			}
 		}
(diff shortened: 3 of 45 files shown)


go/consensus/tendermint/apps/scheduler/scheduler.go
```diff
@@ -400,89 +429,129 @@ func (app *schedulerApplication) electCommittee(
...
+					continue
+				}
+				nodesPerEntity[n.EntityID]++
+			}
+
+			elected = append(elected, &scheduler.CommitteeNode{
+				Role:      role,
+				PublicKey: n.ID,
+			})
+			memberSet[n.ID] = true
```
pro-wh, line 535: the way this election algorithm works could make some electable configurations likely to fail. consider this exaggerated case:

- one big entity has 20 nodes x1, ..., x20
- 20 other entities each have one node a1, ..., t1
- the executor worker committee needs 19 nodes, no per-entity limit
- the executor backup committee needs 19 nodes, max 1 per entity

in this case, there are a few valid committee assignments, for example {x1, ..., x19} are the workers and {a1, ..., s1} are the backup.

in this algorithm, we first pick 19 nodes for the worker committee, where we'll likely take two or more nodes from {a1, ..., t1}, which leaves us with not enough to make the backup committee, and the backup committee election will fail.

there may be other effects other than election failure. I think in other configurations it could make 'later' elections likely to be choosing from pools that are effectively smaller than the configured min pool size.

go/genesis/genesis_test.go
```diff
@@ -225,11 +224,26 @@ func TestGenesisSanityCheck(t *testing.T) {
...
+				scheduler.RoleWorker: {
+					MinPoolSize: &registry.MinPoolSizeConstraint{
+						Limit: 1,
+					},
+				},
+			},
+			scheduler.KindStorage: {
+				scheduler.RoleWorker: {
+					MinPoolSize: &registry.MinPoolSizeConstraint{
+						Limit: 1,
```
pro-wh, line 242: should exercise the other constraint types too

go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
```diff
@@ -247,7 +246,27 @@ func convertRuntime(rt *oldRuntime) *registry.Runtime {
 			CheckpointInterval:      rt.Storage.CheckpointInterval,
 			CheckpointNumKept:       rt.Storage.CheckpointNumKept,
 			CheckpointChunkSize:     rt.Storage.CheckpointChunkSize,
-			MinPoolSize:             rt.Storage.GroupSize,
+		},
+		Constraints: map[scheduler.CommitteeKind]map[scheduler.Role]registry.SchedulingConstraints{
```
pro-wh, line 250: how do we work this file? is there a process where we 'reset' it after we use it to migrate to another version? or are we maintaining it to be a tool for translating from some specific old genesis version?

go/oasis-node/cmd/registry/runtime/runtime.go
```diff
@@ -436,6 +465,56 @@ func runtimeFromFlags() (*registry.Runtime, error) { // nolint: gocyclo
 		return nil, fmt.Errorf("invalid runtime admission policy")
 	}
 
+	// Constraints.
+	// TODO: Hack needed because viper's Sub doesn't work correctly.
```
pro-wh, line 469: are they tracking this issue?

go/oasis-test-runner/scenario/e2e/registry_cli.go
```diff
@@ -642,6 +641,31 @@ func (sc *registryCLIImpl) testRuntime(ctx context.Context, childEnv *env.Env, c
...
+					},
+				},
+			},
+			scheduler.KindStorage: {
+				scheduler.RoleWorker: {
+					MinPoolSize: &registry.MinPoolSizeConstraint{
+						Limit: 9,
+					},
+					MaxNodes: &registry.MaxNodesConstraint{
+						Limit: 1,
```
pro-wh, line 664: was this implicit before? or is this added

go/oasis-test-runner/scenario/e2e/runtime/multiple_runtimes.go
```diff
@@ -104,11 +104,26 @@ func (sc *multipleRuntimesImpl) Fixture() (*oasis.NetworkFixture, error) {
...
+						MinPoolSize: &registry.MinPoolSizeConstraint{
+							Limit: uint16(executorGroupSize),
+						},
+					},
+				},
+				scheduler.KindStorage: {
+					scheduler.RoleWorker: {
+						MinPoolSize: &registry.MinPoolSizeConstraint{
+							Limit: 1,
+						},
```
pro-wh, line 123: a similar MaxNodes is not added here

go/oasis-test-runner/scenario/e2e/runtime/txsource.go
```diff
@@ -418,25 +419,27 @@ func (sc *txSourceImpl) Fixture() (*oasis.NetworkFixture, error) {
 
 	// Storage committee.
 	f.Runtimes[1].Storage.GroupSize = uint64(sc.numStorageNodes)
-	f.Runtimes[1].Storage.MinPoolSize = f.Runtimes[1].Storage.GroupSize
+	f.Runtimes[1].Constraints[scheduler.KindStorage][scheduler.RoleWorker].MinPoolSize.Limit = uint16(sc.numStorageNodes)
```
pro-wh, line 422: kinda makes it awkward that we used uint64 for the group size :thinking:

go/registry/api/runtime.go
```diff
@@ -113,10 +111,6 @@ func (e *ExecutorParameters) ValidateBasic() error {
 		return fmt.Errorf("round timeout too small")
 	}
 
-	if e.MinPoolSize < e.GroupSize+e.GroupBackupSize {
-		return fmt.Errorf("minimum pool size too small")
```
pro-wh, line 117: do these checks get reinstated elsewhere?

go/consensus/tendermint/apps/scheduler/scheduler.go
```diff
@@ -367,31 +383,44 @@ func (app *schedulerApplication) electCommittee(
...
+	groupSizes := make(map[scheduler.Role]int)
 	switch kind {
 	case scheduler.KindComputeExecutor:
 		rngCtx = RNGContextExecutor
 		isSuitableFn = app.isSuitableExecutorWorker
-		workerSize = int(rt.Executor.GroupSize)
-		backupSize = int(rt.Executor.GroupBackupSize)
-		minPoolSize = int(rt.Executor.MinPoolSize)
+		groupSizes[scheduler.RoleWorker] = int(rt.Executor.GroupSize)
+		groupSizes[scheduler.RoleBackupWorker] = int(rt.Executor.GroupBackupSize)
```
pro-wh, line 398: this makes me wonder about moving the GroupSize and GroupBackupSize into the SchedulingConstraints, like this change does with MinPoolSize. are there other places where we like having it separate like this?

go/oasis-node/cmd/debug/txsource/workload/parallel.go
```diff
@@ -26,7 +26,7 @@ var Parallel = &parallel{
 }
 
 const (
-	parallelSendWaitTimeoutInterval = 30 * time.Second
+	parallelSendWaitTimeoutInterval = 60 * time.Second
```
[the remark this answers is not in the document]
kostko: Not related to this PR, I snuck this in due to long-term tests failing (in master) with some timeouts after validator restarts.

go/genesis/genesis_test.go
kostko: They are excersized elsewhere, see the newly added scheduler unit tests.

go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
kostko: This is specifically from 20.12 to master.

go/oasis-node/cmd/registry/runtime/runtime.go
kostko: Yeah there's a bunch of them. Also they are rewriting viper.

go/oasis-test-runner/scenario/e2e/registry_cli.go
kostko: This (MaxNodes) was not implicit. It is added here to test the CLI for generating runtime descriptors.

go/oasis-test-runner/scenario/e2e/runtime/multiple_runtimes.go
kostko: There was no MaxNodes before. But there was a MinPoolSize.

go/oasis-test-runner/scenario/e2e/runtime/txsource.go
kostko: Yes, I think it makes no sense. Should we just change the types of those as well since we can do breaking changes?

go/registry/api/runtime.go
kostko: No, we could add them back to catch nonsensical configs.

go/consensus/tendermint/apps/scheduler/scheduler.go
kostko: He he, I had the exact same thought as well. It would definitely make things nicer.

go/consensus/tendermint/apps/scheduler/genesis.go
```diff
@@ -20,12 +20,7 @@ import (
 )
 
 func (app *schedulerApplication) InitChain(ctx *abciAPI.Context, req types.RequestInitChain, doc *genesis.Document) error {
-	baseEpoch, err := app.state.GetBaseEpoch()
-	if err != nil {
-		return fmt.Errorf("tendermint/scheduler: couldn't get base epoch: %w", err)
-	}
-	app.baseEpoch = baseEpoch
-
+	var err error
```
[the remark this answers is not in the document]
kostko: Yeah the previous version actually had a subtle bug. The `baseEpoch` was only initialized during `InitChain`.

go/consensus/tendermint/apps/scheduler/scheduler.go
kostko: Hmm, yeah you're right. Let's think about how to fix this :-)

bennetyee: dumb question(s) from the peanut gallery:

- why are we picking executor backup committee in a way that's exclusive from the worker committee?  i.e., if a node is in the worker committee why should it not be in the backup?  from a load perspective, backup execution starts only after there is a dispute in the worker results, so shouldn't overlap (much) -- and once a dispute has been found arguably we could abandon the worker committee computation since it will not generate a result that will change what will happen.  as long as the choices for the backup and the worker committees are made independently it shouldn't matter [please check my logic on this].

- ideally the choice for the backup should be made using a new beacon value (yeah, i know we would rather not have to wait a whole new epoch for dispute resolution).  this is because the adversary who decides to try to compromise the computation should not know the identity of the nodes they need to compromise ahead of time.  is doing this feasible?  or, alternatively, have the random beacon generate two values, but delay the reveal until it is needed (need to trust the VSS parties to not reveal early) either for backup committee assignment or for the next worker committee, whichever occurs next.  (Presumably backup is needed rarely.)

kostko: > why are we picking executor backup committee in a way that's exclusive from the worker committee?

I don't think there's any particular reason for doing so. Overlap would actually result in less load as some of the workers would already have the result computed. I agree that as long as the choices are independent it may be fine.

> ideally the choice for the backup should be made using a new beacon value

I agree with the reasoning, but in that case discrepancy resolution would become much more expensive for the consensus layer and it would introduce a lot of additional complexity. Also how would that work with multiple runtimes? There would always need to be a beacon ready for any runtime?

ptrus: I think that with existing constrains: "max_nodes", "validator_set", it would work if you'd start selecting the group with stricter restrictions first? This works since the "validator_set" is basically just a stricter version of the "max_nodes: 1" constraint.  

But generally (if any new constraints are added, or if i missed something in the current constrains) this wouldn't work and would miss solutions.

kostko: > This works since the "validator_set" is basically just a stricter version of the "max_nodes: 1" constraint.

Yeah the problem is that the "validator_set" constraint is not just a stricter version of "max_nodes: 1" as it only requires the _entity_ (and not the node) being in the validator set, but does not limit the number of nodes per entity.

Allowing the two sets to overlap (as @bennetyee mentioned above) would avoid the problem but would require some further changes in executor nodes to support this configuration.

ptrus: > "validator_set" constraint is not just a stricter version of "max_nodes: 1" as it only requires the entity (and not the node) being in the validator set

Oh right.

bennetyee: choosing the backup from new randomness means that we force the adversary be adaptive, i.e., to choose targets of attacks flexibly, based on the new information (beacon value).  this matters when there are fast attacks, e.g., automated zero days.  when all attacks take time (longer than a random beacon epoch), then the adversary cannot be adaptive.

we already assume aspects of non-adaptive adversaries.  users can submit transaction proposals after the worker committee is selected, and the adversary knew the makeup of the committee before deciding whether to "try" to compromise the result.  since the makeup is known, there is no try, only do: a rational adversary would not make nodes under their control send wrong answers when they do not control the entire worker committee.

pro-wh: stricter-restrictions-first can still run into trouble when the min pool size has small or no margin over the total needed:

* one big entity has 5 nodes x1, ..., x5
* 20 other entities each have one node a1, ..., t1
* worker committee needs 13 nodes, limit 3 per entity
* backup committee needs 12 nodes, limit 2 per entity

we can work out that to elect committees, 2 of {x1, ..., x5} have to go to the backup committee and 3 have to go to the worker committee. but left up to random chance, we'd sometimes unknowingly elect 0 or 1 from {x1, ..., x5} for the backup committee and then be stuck when electing the worker committee.

kostko: I changed the logic to support overlapping primary/backup committees. Please take another look.

go/oasis-test-runner/scenario/e2e/runtime/txsource.go
kostko: Changed those to uint16 as well.

go/consensus/tendermint/apps/scheduler/genesis.go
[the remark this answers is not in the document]
pro-wh: oh goody

go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
pro-wh: ok

go/oasis-node/cmd/registry/runtime/runtime.go
pro-wh: lol ok gl to them

go/oasis-test-runner/scenario/e2e/registry_cli.go
pro-wh: ok

go/oasis-test-runner/scenario/e2e/runtime/txsource.go
pro-wh: sounds good

go/genesis/genesis_test.go
pro-wh: oh right

go/consensus/tendermint/apps/roothash/roothash.go
```diff
@@ -635,23 +661,7 @@ func (app *rootHashApplication) tryFinalizeExecutorCommits(
...
-			Event: roothash.ExecutionDiscrepancyDetectedEvent{
-				Timeout: forced,
-			},
-		}
-		ctx.EmitEvent(
-			tmapi.NewEventBuilder(app.Name()).
-				Attribute(KeyExecutionDiscrepancyDetected, cbor.Marshal(tagV)).
-				Attribute(KeyRuntimeID, ValueRuntimeID(runtime.ID)),
-		)
+		// This was already handled above, so it should not happen.
```
pro-wh, line 664: control shouldn't reach here, or just nothing left to do? we don't return early in the `if` added above

go/roothash/api/commitment/pool_test.go
```diff
@@ -1095,6 +1120,7 @@ func setupDiscrepancy(
 	sks []signature.Signer,
 	committee *scheduler.Committee,
 	nl NodeLookup,
+	enoughBackupCommits bool,
```
pro-wh, line 1123: :information_source: due to some overlap between the worker and backup committees, some tests for worker bad behavior also result in missing backup commits

pro-wh (approved):
'worker' and 'backup' roles are no longer exclusive with each other. that results in pretty extensive changes.

go/oasis-test-runner/scenario/e2e/runtime/byzantine.go
```diff
@@ -19,16 +19,17 @@ import (
 
 var (
```
ptrus, line 20: There is a `schedulerCheckScheduled` method in the byzantine worker that assumes distinct members in the `committee.Members` and should be updated for future proofness.

go/consensus/tendermint/apps/roothash/roothash.go
kostko: So the branch added above will overwrite `err` and the repeated `TryFinalize` cannot (currently) return the "discrepancy detected" error.

go/oasis-test-runner/scenario/e2e/runtime/byzantine.go
kostko: Good catch, updated.

PR #3696, opened 2021-02-11
go/worker/storage: Add StorageRPC role and change access policy

### LINKED ISSUE #3562: Change storage node access policy
The storage node access policy should be changed so that fetching checkpoints and diffs is allowed from any clients (currently only allowed from committee clients) and state access is only allowed from committee clients (currently allowed from any clients) or optionally from any clients if the node opts in to that (there should probably be a separate role like `RoleStorageRPC`).

This would also make it possible to make storage nodes register only after being synced even if #2039 is not yet implemented.

jberci:
Getting checkpoints and diffs is now allowed for any connecting node,
which eliminates a race condition and initialization difficulty in the
storage committee node startup.

State access is made gated but optionally public as before, depending on
command line parameters.

Closes #3562

### DIFF
--- a/go/oasis-test-runner/scenario/e2e/runtime/sentry.go
+++ b/go/oasis-test-runner/scenario/e2e/runtime/sentry.go
@@ -25,6 +25,7 @@ import (
 	storage "github.com/oasisprotocol/oasis-core/go/storage/api"
 	"github.com/oasisprotocol/oasis-core/go/storage/database"
 	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/checkpoint"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/syncer"
 )
 
 var (
@@ -38,6 +39,14 @@ var (
 	}
 
 	emptyGetCheckpointsReq = &checkpoint.GetCheckpointsRequest{Namespace: runtimeID}
+
+	emptySyncGetReq = &storage.GetRequest{
+		Tree: syncer.TreeID{
+			Root: storage.Root{
+				Namespace: runtimeID,
+			},
+		},
+	}
 )
 
 const sentryChecksContextTimeout = 30 * time.Second
@@ -139,6 +148,8 @@ func (s *sentryImpl) Fixture() (*oasis.NetworkFixture, error) {
 			// Disable cert rotation on one of the storage nodes so we can use
 			// its TLS certificates in the access control sanity checks.
 			DisableCertRotation: true,
+			// Also disable public RPC on one, so we can check access control.
+			DisablePublicRPC: true,
 		},
 		{
 			Backend:  database.BackendNameBadgerDB,
@@ -184,6 +195,7 @@ func (s *sentryImpl) Run(childEnv *env.Env) error { // nolint: gocyclo
 	sentry0, _, sentry0CtrlAddress, sentry0P2PPubkey := loadSentryNodeInfo(s.Net.Sentries()[0])
 	_, _, _, sentry1P2PPubkey := loadSentryNodeInfo(s.Net.Sentries()[1])
 	_, _, _, sentry2P2PPubkey := loadSentryNodeInfo(s.Net.Sentries()[2])
+	sentry3, sentry3Address, _, _ := loadSentryNodeInfo(s.Net.Sentries()[3])
 	sentry4, sentry4Address, _, _ := loadSentryNodeInfo(s.Net.Sentries()[4])
 
 	storage0 := s.Net.StorageWorkers()[0]
@@ -197,6 +209,10 @@ func (s *sentryImpl) Run(childEnv *env.Env) error { // nolint: gocyclo
 	}
 
 	storage1 := s.Net.StorageWorkers()[1]
+	storage1Identity, err := storage0.LoadIdentity()
+	if err != nil {
+		return fmt.Errorf("sentry: error loading storage node identity: %w", err)
+	}
 	storage1Address := storage1.GetClientAddress()
 	// Query for storage 1 TLS public keys.
 	storage1Ctrl, err := oasis.NewController(storage1.SocketPath())
@@ -321,8 +337,8 @@ func (s *sentryImpl) Run(childEnv *env.Env) error { // nolint: gocyclo
 	storage1Client := storage.NewStorageClient(conn)
 	_, err = storage1Client.GetCheckpoints(ctx, emptyGetCheckpointsReq)
 	s.Logger.Debug("storage1.GetCheckpoints with validator0 cert", "err", err)
-	if status.Code(err) != codes.PermissionDenied {
-		return errors.New("storage1 checkpoint endpoint should deny connection with validator0 certificate")
+	if err != nil {
+		return errors.New("storage1 checkpoint endpoint should allow connection with validator0 certificate")
 	}
 
 	// Check Storage-1 endpoint with Storage-0 client certificates.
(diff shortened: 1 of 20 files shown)


go/worker/storage/committee/node.go
```diff
@@ -665,21 +668,27 @@ func (n *Node) updateExternalServicePolicy(rtComputeNodes nodes.NodeDescriptorLo
...
-	if err != nil {
-		n.logger.Error("couldn't get nodes from registry", "err", err)
-	}
-	if len(nodes) > 0 {
-		// Only include storage nodes for our runtime.
-		var storageNodes []*node.Node
-		for _, nd := range nodes {
-			if nd.GetRuntime(n.commonNode.Runtime.ID()) != nil && nd.HasRoles(node.RoleStorageWorker) {
-				storageNodes = append(storageNodes, nd)
+	if n.rpcRoleProvider == nil {
```
kostko, line 671: Change this to a `case` and add comments for both branches (e.g. when is the branch taken).

go/worker/storage/committee/node.go
```diff
@@ -891,6 +881,18 @@ func (n *Node) worker() { // nolint: gocyclo
 	}
 	close(n.initCh)
 
+	// We are now ready to service requests.
```
kostko, line 884: It would be cool if this could wait until the storage is actually synced close enough with the current round (e.g., there could be no checkpoints or the node would still need to sync quite a few diffs from the last checkpoint).

go/worker/storage/committee/policy.go
```diff
@@ -20,16 +20,13 @@ var (
 	// sync-up.
 	storageNodesPolicy = &committee.AccessPolicy{
 		Actions: []accessctl.Action{
-			accessctl.Action(api.MethodGetDiff.FullName()),
-			accessctl.Action(api.MethodGetCheckpoints.FullName()),
-			accessctl.Action(api.MethodGetCheckpointChunk.FullName()),
+			accessctl.Action(api.MethodSyncGet.FullName()),
+			accessctl.Action(api.MethodSyncGetPrefixes.FullName()),
+			accessctl.Action(api.MethodSyncIterate.FullName()),
```
kostko, line 25: Why do other storage nodes need access to these methods?

go/worker/storage/init.go
```diff
@@ -23,6 +23,10 @@ const (
 	CfgWorkerEnabled      = "worker.storage.enabled"
 	cfgWorkerFetcherCount = "worker.storage.fetcher_count"
 
+	// CfgWorkerEnableAllStorageRPC enables storage state access for all nodes instead of just
+	// storage committee members.
+	CfgWorkerEnableAllStorageRPC = "worker.storage.enable_all_rpc"
```
kostko, line 28: Maybe change it to `worker.storage.public_rpc.enabled`.

go/worker/storage/committee/node.go
```diff
@@ -665,21 +674,32 @@ func (n *Node) updateExternalServicePolicy(rtComputeNodes nodes.NodeDescriptorLo
...
+	switch {
+	// If public storage RPC was enabled in the config, then the normally gated methods need to be allowed
+	// for everyone.
+	case n.rpcRoleProvider != nil:
+		for _, act := range storageNodesPolicy.Actions {
+			policy.AllowAll(act)
+		}
+
+	// If not configured otherwise, state access should be restricted to storage committee members.
+	default:
```
kostko, line 686: Why is this part not covered? Are we not testing with any storage nodes that don't expose public RPC? We should test those as well (e.g., configure a storage node without public RPC exposed). Maybe in cases where we have two storage nodes, set one to expose public RPC and the other to not expose it?

go/worker/storage/committee/node.go
```diff
@@ -750,6 +770,26 @@ func (n *Node) watchQuit() {
 	close(n.quitCh)
 }
 
+func (n *Node) nudgeAvailability(lastSynced, latest uint64) {
+	if lastSynced == n.undefinedRound || latest == n.undefinedRound {
+		return
+	}
+	if latest-lastSynced < maximumRoundDelayForAvailability {
```
kostko, line 777: Should we also stop being available (e.g., `n.roleProvider.SetUnavailable()`) if we fall too far behind (using a separate threshold to avoid oscillations around the boundary)?

go/storage/api/grpc.go
```diff
@@ -21,54 +21,45 @@ var (
...
 			WithNamespaceExtractor(func(ctx context.Context, req interface{}) (common.Namespace, error) {
-			r, ok := req.(*ApplyRequest)
+			r, ok := req.(*GetRequest)
 			if !ok {
 				return common.Namespace{}, errInvalidRequestType
 			}
-			return r.Namespace, nil
+			return r.Tree.Root.Namespace, nil
 		}).
 		WithAccessControl(func(ctx context.Context, req interface{}) (bool, error) {
```
ptrus, line 32:
```
func(ctx context.Context, req interface{}) (bool, error) {
  return true, nil
}
```
could be extracted into common/grpc package, so one could then do something like:
`.WithAccessControl(cmnGrpc.AccessControlAlways)` when there's no special logic for deciding on access control.

go/storage/client/init.go
```diff
@@ -55,25 +54,28 @@ func New(
...
 	namespace common.Namespace,
 	ident *identity.Identity,
-	schedulerBackend scheduler.Backend,
 	registryBackend registry.Backend,
 	runtime registry.RuntimeDescriptorProvider,
-	extraWatcherOpts ...committee.WatcherOption,
 ) (api.Backend, error) {
-	watcherOpts := append(extraWatcherOpts, committee.WithAutomaticEpochTransitions())
-	committeeWatcher, err := committee.NewWatcher(
+	nl, err := nodes.NewRuntimeNodeLookup(
```
ptrus, line 60: The method comment above `// New creates a new storage client that automatically follows a given runtime's storage committee.`
should change, since this now follows all registered storage nodes for the runtime with RoleStorageRPC enabled. While before it followed the active committee members.

Note that: `go/runtime/registry` also uses this method, did we double-check that anything using the runtime registry will/should work with this change? 
e.g. if there are no storage nodes with `RoleStorageRPC` enabled, the storage part of the `go/runtime/registry` won't work.

go/storage/client/init.go
```diff
@@ -55,25 +54,28 @@ func New(
...
+		return nil, fmt.Errorf("storage/client: failed to create runtime node watcher: %w", err)
 	}
 
-	return NewForNodes(ctx, ident, committeeWatcher.Nodes(), runtime)
+	publicStorageNl := nodes.NewFilteredNodeLookup(nl,
+		nodes.WithAllFilters(
+			// Ignore self.
+			nodes.IgnoreNodeFilter(ident.NodeSigner.Public()),
+			// Only storage nodes that opted into public RPC.
+			nodes.TagFilter(nodes.TagsForRoleMask(node.RoleStorageRPC)[0]),
```
ptrus, line 74: So if there's no nodes with `RoleStorageRPC` enabled, this won't select any nodes and the returned client will fail all the requests - even for the methods that don't have auth (getting checkpoints and diffs).

I think this filter should be skipped.

kostko: Note that the this default storage client is specifically used by a client node as the storage and compute workers use their own clients. So I think this is fine?

go/storage/client/init.go
kostko: > Note that: go/runtime/registry also uses this method, did we double-check that anything using the runtime registry will/should work with this change?

AFAIK only the runtime client uses this constructor as storage and compute workers have their own committee handling.

go/storage/client/init.go
ptrus: Yeah probably fine then. Should maybe rename the method to something that better indicates this is connecting to public storage nodes specifically.

kostko: Yeah good idea.

go/worker/storage/committee/node.go
jberci: Sounds sensible; added it.

go/oasis-test-runner/scenario/e2e/runtime/sentry.go
```diff
@@ -139,11 +148,14 @@ func (s *sentryImpl) Fixture() (*oasis.NetworkFixture, error) {
...
+			DisablePublicRPC: true,
 		},
 		{
-			Backend:  database.BackendNameBadgerDB,
-			Entity:   1,
-			Sentries: []int{4},
+			Backend:             database.BackendNameBadgerDB,
+			Entity:              1,
+			Sentries:            []int{4},
+			DisableCertRotation: true,
```
kostko, line 158: This disables cert rotation on both storage workers, so this part is not tested?

kostko (approved):

ptrus (approved):

Issue #3484, opened 2020-11-05
State of suspended runtimes is not correctly restored after a dump/restore


kostko:
Roothash state of suspended runtimes is not correctly restored after a dump/restore since the only place where we restore state is at genesis and not later when the runtime becomes unsuspended. This means that for any suspended runtimes the state will be reset.

This should be fixed together with a test for moving a suspended runtime across a dump/restore upgrade.

PR #3708, opened 2021-02-18
go/roothash/genesis: restore state from suspended runtimes

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3484

### DIFF
--- a/go/oasis-node/cmd/ias/auth_genesis.go
+++ b/go/oasis-node/cmd/ias/auth_genesis.go
@@ -50,8 +50,13 @@ func newGenesisAuthenticator() (iasProxy.Authenticator, error) {
 		logger:   logging.GetLogger("cmd/ias/proxy/auth/genesis"),
 		enclaves: newEnclaveStore(),
 	}
+	for _, rt := range doc.Registry.SuspendedRuntimes {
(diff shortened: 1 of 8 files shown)


go/oasis-node/cmd/ias/auth_genesis.go
```diff
@@ -50,6 +50,11 @@ func newGenesisAuthenticator() (iasProxy.Authenticator, error) {
 		logger:   logging.GetLogger("cmd/ias/proxy/auth/genesis"),
 		enclaves: newEnclaveStore(),
 	}
+	for _, rt := range doc.Registry.SuspendedRuntimes {
```
kostko, line 53: Nice catch! We should make sure that the `auth_registry` version works correctly for suspended runtimes as well.

ptrus: Yeah i'm running a test it with auth_registry now, will see if it works.

ptrus: Auth registry works correctly, i actually remember now that i fixed the `WatchRuntimes` awhile ago to include suspended runtimes :-)

kostko (approved):

Issue #3641, opened 2021-01-20
Optionally allow runtimes to perform Add/ReclaimEscrow


kostko:
We should optionally (based on a staking consensus parameter which defaults to false) allow runtimes to perform AddEscrow and ReclaimEscrow operations via runtime messages.

PR #3717, opened 2021-02-22
Add AddEscrow and ReclaimEscrow runtime messages

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3641

TODO:
- [x] add consensus parameter to disallow add/reclaim escrow messages
- [x] add add/reclaim escrow messages to runtime txsource workload
- [x] fix reclaim_escrow (it currently relies on account nonce updating (for debonding delegations), which doesn't happen on messages)

### DIFF
--- a/go/consensus/tendermint/api/context.go
+++ b/go/consensus/tendermint/api/context.go
@@ -57,13 +57,14 @@ func (m ContextMode) String() string {
 }
 
 // Context is the context of processing a transaction/block.
-type Context struct {
+type Context struct { // nolint: maligned
 	context.Context
 
 	parent *Context
 
-	mode        ContextMode
-	currentTime time.Time
+	mode               ContextMode
+	currentTime        time.Time
+	isMessageExecution bool
 
 	data          interface{}
 	events        []types.Event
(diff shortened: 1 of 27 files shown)


go/consensus/tendermint/apps/staking/staking.go
```diff
@@ -120,6 +120,12 @@ func (app *stakingApplication) ExecuteMessage(ctx *api.Context, kind, msg interf
 			return app.transfer(ctx, state, m.Transfer)
 		case m.Withdraw != nil:
 			return app.withdraw(ctx, state, m.Withdraw)
+		case m.AddEscrow != nil:
+			// TODO: probably nicer if `processRuntimeMessages` instead stores into context
```
ptrus, line 124: cc @kostko  what do you think?

kostko: Yeah I had the same thought while looking at the draft ;-)

ptrus: updated

go/consensus/tendermint/api/context.go
```diff
@@ -251,6 +253,16 @@ func (c *Context) IsSimulation() bool {
 	return c.mode == ContextSimulateTx
 }
 
+// SetIsMessageExecution sets the message execution flag.
+func (c *Context) SetIsMessageExecution(b bool) {
```
kostko, line 257: Maybe it would be better if we do it like `WithMessageExecution` (similar to WithCallerAddress/WithSimulation) so that it forces the caller to use a child context and not modify the flag by accident.

ptrus: Yeah agreed, will do.

kostko (approved):

PR #3719, opened 2021-02-22
go/oasis-test-runner: configurable sanitychecker

ptrus:
Daily longterm tests have been failing since validator-0 was falling behind the rest of the network due to the supplementary-sanity checker. This moves the supplementary-sanity checker to a new (client-1) node, which can fall behind without affecting rest of the network.

Also increase the timeout between governance workload iterations, as currently the amount of proposals was quickly pilling up over time (and both queries-workload and sanity-checker, go through all (even expired) proposal and votes).

TODO: 
- [x] run the longterm job to ensure the CI instances can handle the test suite with the additional node: https://buildkite.com/oasisprotocol/oasis-core-long-tests/builds/543

### DIFF
--- a/go/oasis-test-runner/oasis/validator.go
+++ b/go/oasis-test-runner/oasis/validator.go
@@ -82,6 +82,7 @@ func (val *Validator) startNode() error {
 	args := newArgBuilder().
 		debugDontBlameOasis().
 		debugAllowTestKeys().
+		debugEnableProfiling(val.Node.pprofPort).
 		workerCertificateRotation(true).
 		consensusValidator().
 		tendermintCoreAddress(val.consensusPort).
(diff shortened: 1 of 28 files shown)


ptrus:
@kostko This is ready for review. The last 2 failed runs were due to runtime staking messages being resubmitted (related to compute node being restarted at an unlucky time), causing the operation to re-apply (since the runtime doesn't use nonces). 

I think we should tackle those in a subsequent PR as this one got big enough and it's kinda hard to keep testing the changes here while the daily tests are also being run. This PR however should fix the permanent test failures that are currently happening on the daily tests runs.

go/oasis-test-runner/oasis/validator.go
```diff
@@ -146,8 +145,14 @@ func (net *Network) NewValidator(cfg *ValidatorCfg) (*Validator, error) {
 		consensusPort: net.nextNodePort,
 		clientPort:    net.nextNodePort + 1,
 	}
+	net.nextNodePort += 2
 	val.doStartNode = val.startNode
 
+	if cfg.EnableProfiling {
+		val.Node.pprofPort = net.nextNodePort
+		net.nextNodePort++
+	}
```
kostko, line 154: Would be nice if we could unify/cleanup this as it is the same for every `Node`, but this is also true for the `Node` constructor above so probably best for a separate issue/PR.

kostko (approved):
Nice work, thanks!

go/oasis-test-runner/oasis/validator.go
ptrus: :+1: I'll open an issue

PR #3720, opened 2021-02-22
go: Improve error message when delegating less than MinDelegationAmount

### LINKED ISSUE #3308: Display and log a better error in case escrow transaction does not pass MinDelegationAmount
Currently if an escrow transaction tries to delegate less than `MinDelegationAmount` the transaction fails with `staking.ErrInvalidArgument` and nothing is logged.

https://github.com/oasisprotocol/oasis-core/blob/6c6e70303b2cb18d72b6f4396ad8959fd2c3b5f4/go/consensus/tendermint/apps/staking/transactions.go#L178-L181

abukosek:
Closes #3308.

### DIFF
--- a/go/consensus/tendermint/apps/staking/transactions.go
+++ b/go/consensus/tendermint/apps/staking/transactions.go
@@ -192,7 +192,7 @@ func (app *stakingApplication) addEscrow(ctx *api.Context, state *stakingState.M
 
 	// Check if sender provided at least a minimum amount of stake.
 	if escrow.Amount.Cmp(&params.MinDelegationAmount) < 0 {
-		return staking.ErrInvalidArgument
+		return staking.ErrUnderMinDelegationAmount
 	}
 
 	fromAddr := ctx.CallerAddress()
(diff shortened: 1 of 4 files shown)


go/consensus/tendermint/apps/staking/transactions.go
```diff
@@ -192,7 +192,7 @@ func (app *stakingApplication) addEscrow(ctx *api.Context, state *stakingState.M
 
 	// Check if sender provided at least a minimum amount of stake.
 	if escrow.Amount.Cmp(&params.MinDelegationAmount) < 0 {
-		return staking.ErrInvalidArgument
+		return staking.ErrUnderMinDelegationAmount
```
kostko, line 195: Can we add a unit test that checks if this specific error is returned?

abukosek: Done! :)  I've also added a few extra tests for the other errors that can happen in `addEscrow`.

kostko (approved):

PR #3726, opened 2021-02-24
runtime: Provide sets of good and bad compute nodes in round results

kostko:


go/roothash/api/history.go
```diff
@@ -4,9 +4,24 @@ import (
...
 
+// RoundResults contains information about how a particular round was executed by the consensus
+// layer.
+type RoundResults struct {
+	// Messages are the results of executing emitted runtime messages.
+	Messages []*MessageEvent `json:"messages,omitempty"`
+
+	// GoodComputeNodes are the public keys of compute nodes that positively contributed to the
+	// round by replicating the computation correctly.
+	GoodComputeNodes []signature.PublicKey `json:"good_compute_nodes,omitempty"`
```
ptrus, line 19: Wondering if there's a better alternative naming to use for nodes that submitted correct and invalid results. I thought about it, but don't have a better suggestion though :D

kostko: Yeah, I was thinking about how to name this as well, but didn't come up with anything better. Tendermint uses the term "Byzantine Validators" for the bad ones.

runtime/src/consensus/roothash.rs
```diff
@@ -121,6 +121,23 @@ impl MessageEvent {
     }
 }
 
+/// Information about how a particular round was executed by the consensus layer.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct RoundResults {
```
ptrus, line 126: should we add a simple serialization test case for the `RoundResults` struct to match go side, similarly as we have for other structs here?

go/consensus/tendermint/apps/roothash/roothash.go
```diff
@@ -587,50 +591,41 @@ func (app *rootHashApplication) tryFinalizeExecutorCommits(
 			return fmt.Errorf("failed to process runtime messages: %w", err)
 		}
 
+		var (
```
ptrus, line 594: :+1:

ptrus (approved):

runtime/src/consensus/roothash.rs
kostko: Good idea, added.

Issue #3728, opened 2021-02-24
oasis-node debug fix-genesis CLI command doesn't configure beacon correctly


tjanez:
##### SUMMARY

The `oasis-node debug fix-genesis` CLI command doesn't configure beacon correctly:

Using the [2021-02-03 Testnet genesis file](https://github.com/oasisprotocol/testnet-artifacts/releases/download/2021-02-03/genesis.json), it fails with:

```
$ oasis-node debug fix-genesis -g genesis.json --genesis.new_file genesis-new.json
ts=2021-02-24T15:23:06.64897691Z level=error module=cmd/debug/fix-genesis caller=fixgenesis.go:191 msg="new genesis document sanity check failed" err="beacon: sanity check failed: unknown backend: ''"
```

The relevant part of the original `genesis.json`:

```json
  "beacon": {
    "params": {}
  },
```

The relevant part of the new `genesis-new.json`:

```json
  "beacon": {
    "base": 0,
    "params": {
      "backend": ""
    }
  },
```

##### ISSUE TYPE
- Bug Report

##### OASIS NODE VERSION

The latest `master` as of today: https://github.com/oasisprotocol/oasis-core/commit/c07c9877d7da5c17038193cd59dbde5bd8a096b6

```paste below
$ oasis-node --version
Software version: 20.12-gitc07c9877d
Consensus:
  Consensus protocol version: 3.0.0
Runtime:
  Host protocol version:      2.0.0
  Committee protocol version: 2.0.0
Go toolchain version: 1.15.6
```

kostko:
Yeah we should probably change `fixgenesis` to fill in some default parameters that will be used for the mainnet upgrade.

ptrus:
Also `Beacon.Base` should be transferred over from the original genesis.

Additionally `fixgenesis` should set some default governance parameters.

ptrus:
`consensus.parameters.max_evidence_size` should also probably be filled by `fixgenesis` (computed from existing "max_evidence_num")

PR #3731, opened 2021-02-25
go/cmd/fixgenesis: Add default beacon and governance parameters

ptrus:
Fixes: #3728

TODO:
- [x] see todo's in code
- [x] remove entities not passing stake claims (done, but blocked by: #3754)

### DIFF
--- a/go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
+++ b/go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
@@ -19,8 +19,10 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common/entity"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
 	"github.com/oasisprotocol/oasis-core/go/common/node"
+	"github.com/oasisprotocol/oasis-core/go/common/quantity"
 	consensus "github.com/oasisprotocol/oasis-core/go/consensus/genesis"
 	genesis "github.com/oasisprotocol/oasis-core/go/genesis/api"
+	governance "github.com/oasisprotocol/oasis-core/go/governance/api"
 	keymanager "github.com/oasisprotocol/oasis-core/go/keymanager/api"
 	cmdCommon "github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common"
 	"github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common/flags"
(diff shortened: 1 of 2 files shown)


go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
```diff
@@ -301,27 +324,57 @@ func updateGenesisDoc(oldDoc *oldDocument) (*genesis.Document, error) {
...
+	oldEpochInterval := oldDoc.Epochtime.Parameters.Interval
+	newDoc.Beacon.Base = beacon.EpochTime(oldDoc.Height / oldEpochInterval)
+	newDoc.Beacon.Parameters.Backend = beacon.BackendPVSS
+	newDoc.Beacon.Parameters.PVSSParameters = &beacon.PVSSParameters{
+		// TODO: These were taken from E2E tests.
+		CommitInterval:  oldEpochInterval / 2,
+		RevealInterval:  (oldEpochInterval / 2) - 4,
+		TransitionDelay: 4,
+		// TODO: sensible defaults? Compute based on old doc validator set?
+		Threshold:    15,
```
kostko, line 347: This could probably be 10.

go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
```diff
@@ -301,27 +324,57 @@ func updateGenesisDoc(oldDoc *oldDocument) (*genesis.Document, error) {
...
-		Consensus:  oldDoc.Consensus,
 		HaltEpoch:  oldDoc.HaltEpoch,
 		ExtraData:  oldDoc.ExtraData,
 	}
 
+	// Consensus.
+	newDoc.Consensus.Backend = oldDoc.Consensus.Backend
+	newDoc.Consensus.Parameters = oldDoc.Consensus.Parameters.Parameters
+	// TODO: different constant?
+	newDoc.Consensus.Parameters.MaxEvidenceSize = 100 * uint64(oldDoc.Consensus.Parameters.MaxEvidenceNum)
```
kostko, line 335: Probably should be 1024 to be safe.

go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
```diff
@@ -301,27 +324,57 @@ func updateGenesisDoc(oldDoc *oldDocument) (*genesis.Document, error) {
...
+		Threshold:    15,
+		Participants: 20,
+	}
+
+	// Roothash.
 	newDoc.RootHash.Parameters.MaxRuntimeMessages = 32
 
+	// Governance.
+	newDoc.Governance.Parameters = governance.ConsensusParameters{
+		MinProposalDeposit: *quantity.NewFromUint64(100000000000), // TODO: more?
```
kostko, line 356: Yeah this should probably be a bit more, like `10_000_000_000_000`?

go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
```diff
@@ -301,27 +324,57 @@ func updateGenesisDoc(oldDoc *oldDocument) (*genesis.Document, error) {
...
+		CommitInterval:  oldEpochInterval / 2,
+		RevealInterval:  (oldEpochInterval / 2) - 4,
+		TransitionDelay: 4,
+		// TODO: sensible defaults? Compute based on old doc validator set?
+		Threshold:    15,
+		Participants: 20,
+	}
+
+	// Roothash.
 	newDoc.RootHash.Parameters.MaxRuntimeMessages = 32
```
kostko, line 352: This should be increased to 256.

go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
ptrus: Just 1024 or `1024 * existing_MaxEvidenceNum`?

Yawning:
So, I'm moderately against automagically fixing up the beacon parameters because it is deployment specific, and ostensibly oasis-core is deployment agnostic.  Realistically there is one deployment that matters so if people are ok with a command in a supposedly generic codebase using values that are specific to one deployed instance of the codebase, then the values seem ok (though in that case the reveal interval could probably be shortened).

kostko:
Yeah that's always a problem. We could also move the mainnet specific tool to a separate repo.

> though in that case the reveal interval could probably be shortened

Any suggestions? So it needs to be enough for everyone to have a proposal slot.

tjanez:
Since this is a debug tool, I would vote for removing the hard failure in case the sanity checks don't pass:

https://github.com/oasisprotocol/oasis-core/blob/3c302bdad91ead313f4039169a4837c94ba60c8d/go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go#L189-L195

With something like:

```go
	// Validate the new genesis document.
	if err = newDoc.SanityCheck(); err != nil {
		logger.Warn("new genesis document sanity check failed",
			"err", err,
		)
	}
```

This will still warn the user but also give him a (potentially broken) new genesis file which he can fix manually later.

Or not at all if the failure is something like:
```
level=warn module=cmd/debug/fix-genesis caller=fixgenesis.go:214 msg="new genesis document sanity check failed" err="registry: sanity check failed: one or more unsafe debug flags set"
```
which is expected in certain use cases.

ptrus:
> Since this is a debug tool, I would vote for removing the hard failure in case the sanity checks don't pass:

Yeah good point, agreed. Will update to only warn.

kostko (approved):

PR #3732, opened 2021-02-25
go/worker/storage: Add validation for IO roots in Apply operations

### LINKED ISSUE #3440: Additional validation for IO roots

jberci:
Closes #3440

### DIFF
--- a/go/common/keyformat/key_format.go
+++ b/go/common/keyformat/key_format.go
@@ -182,6 +182,9 @@ func (k *KeyFormat) Encode(values ...interface{}) []byte {
 //
 // Returns false and doesn't modify the passed values if the key prefix
 // doesn't match.
+//
+// *NOTE:* If decoding fails for one of the values, previous values
+// will be modified.
 func (k *KeyFormat) Decode(data []byte, values ...interface{}) bool {
 	if data[0] != k.prefix {
 		return false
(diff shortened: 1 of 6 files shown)


go/runtime/transaction/transaction.go
```diff
@@ -85,6 +85,44 @@ var (
...
+			return fmt.Errorf("transaction: invalid key format")
+		}
+	}
+
+	if inputs > maxBatchSize || outputs > maxBatchSize {
+		return fmt.Errorf("transaction: too many inputs or outputs")
+	}
+	if inputSize > maxBatchSizeBytes {
+		return fmt.Errorf("transaction: input set size exceeds configuration")
+	}
```
kostko, line 121: It would probably be better if these checks were done while inside the loop to abort early.

go/runtime/transaction/transaction_test.go
```diff
@@ -181,3 +181,111 @@ func TestTransactionInvalidBatchOrder(t *testing.T) {
 	_, err = tree.GetInputBatch(ctx, 0, 0)
 	require.Error(t, err, "GetInputBatch should fail with inconsistent order")
 }
+
+func TestIOWriteLogValidation(t *testing.T) {
```
kostko, line 185: Could you change this function to do a table-based test instead? I think this would be more clear.

kostko (approved):

go/common/keyformat/key_format.go
```diff
@@ -182,6 +182,9 @@ func (k *KeyFormat) Encode(values ...interface{}) []byte {
 //
 // Returns false and doesn't modify the passed values if the key prefix
 // doesn't match.
+//
+// *NOTE:* If decoding fails for one of the values, then previous values
```
ptrus, line 186:
```suggestion
// *NOTE:* If decoding fails for one of the values, the previous values
```

PR #3737, opened 2021-02-26
Remove support for entity-signed node registrations

### LINKED ISSUE #3661: Remove support for entity-signed node registrations
We should remove support for entity-signed node registrations to reduce complexity. Note that such registrations are a debug-only feature and are not used in any real network.

abukosek:
Closes #3661.

TODO:
- [x] fix unit tests.
- [x] fix e2e tests.

### DIFF
--- a/go/common/entity/entity.go
+++ b/go/common/entity/entity.go
@@ -51,10 +51,6 @@ type Entity struct { // nolint: maligned
 	// will sign the descriptor with the node signing key rather than the
 	// entity signing key.
 	Nodes []signature.PublicKey `json:"nodes,omitempty"`
-
-	// AllowEntitySignedNodes is true iff nodes belonging to this entity
-	// may be signed with the entity signing key.
-	AllowEntitySignedNodes bool `json:"allow_entity_signed_nodes,omitempty"`
 }
 
 // ValidateBasic performs basic descriptor validity checks.
@@ -147,7 +143,6 @@ func GenerateWithSigner(baseDir string, signer signature.Signer, template *Entit
 	}
 	if template != nil {
 		ent.Nodes = template.Nodes
-		ent.AllowEntitySignedNodes = template.AllowEntitySignedNodes
 	}
 
 	if err := ent.Save(baseDir); err != nil {
@@ -223,5 +218,4 @@ func init() {
 
 	testEntity.Versioned = cbor.NewVersioned(LatestEntityDescriptorVersion)
 	testEntity.ID = testEntitySigner.Public()
-	testEntity.AllowEntitySignedNodes = true
 }
--- a/go/oasis-node/cmd/registry/node/node.go
+++ b/go/oasis-node/cmd/registry/node/node.go
@@ -18,7 +18,6 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
 	fileSigner "github.com/oasisprotocol/oasis-core/go/common/crypto/signature/signers/file"
-	"github.com/oasisprotocol/oasis-core/go/common/entity"
 	"github.com/oasisprotocol/oasis-core/go/common/identity"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
 	"github.com/oasisprotocol/oasis-core/go/common/node"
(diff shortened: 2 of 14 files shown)


go/registry/api/api.go
```diff
@@ -483,20 +483,12 @@ func VerifyRegisterNodeArgs( // nolint: gocyclo
...
-			if !params.DebugAllowEntitySignedNodeRegistration || (!entity.AllowEntitySignedNodes && (!isSanityCheck || isGenesis)) {
-				logger.Error("RegisterNode: registration signed by entity",
-					"signed_node", sigNode,
-					"node", n,
-				)
-				return nil, nil, fmt.Errorf("%w: registration signed by entity", ErrInvalidArgument)
-			}
-
-			expectedSigners = append(expectedSigners, entity.ID)
+			// Entity-signed node registrations are not allowed.
```
kostko, line 486: This should be changed to no longer check for entity signature at all. It should probably always use the second check.

abukosek:
Thanks for the comments, this is now ready for re-review :)

go/oasis-node/cmd/registry/node/node.go
```diff
@@ -124,26 +120,9 @@ func doInit(cmd *cobra.Command, args []string) { // nolint: gocyclo
...
-		if viper.GetBool(CfgSelfSigned) {
-			isSelfSigned = true
-		}
-		if !cmdFlags.DebugDontBlameOasis() && !isSelfSigned {
-			logger.Error("entity signed node descriptors are prohibited")
-			os.Exit(1)
-		}
+		logger.Error("entity-signed node descriptors are prohibited")
+		os.Exit(1)
 		defer entitySigner.Reset()
```
kostko, line 126: This should probably be removed as well as it is unreachable (and would crash if reached)?

go/oasis-node/cmd/registry/node/node.go
```diff
@@ -110,10 +109,7 @@ func doInit(cmd *cobra.Command, args []string) { // nolint: gocyclo
 	var (
 		entityID signature.PublicKey
 
-		entity       *entity.Entity
 		entitySigner signature.Signer
```
kostko, line 112: Why is the `entitySigner` still needed?

go/oasis-node/cmd/registry/node/node.go
abukosek: Good catch, fixed! :)

go/oasis-node/cmd/registry/node/node.go
```diff
@@ -124,27 +116,9 @@ func doInit(cmd *cobra.Command, args []string) { // nolint: gocyclo
...
-		isSelfSigned = !entity.AllowEntitySignedNodes
-		if viper.GetBool(CfgSelfSigned) {
-			isSelfSigned = true
-		}
-		if !cmdFlags.DebugDontBlameOasis() && !isSelfSigned {
-			logger.Error("entity signed node descriptors are prohibited")
-			os.Exit(1)
-		}
-		defer entitySigner.Reset()
+		logger.Error("entity-signed node descriptors are prohibited")
```
kostko, line 120: Maybe this should rather say that the "entity ID" argument is missing? As it is otherwise unclear what the error is referring to.

abukosek: Fixed!

go/oasis-node/cmd/registry/node/node.go
```diff
@@ -124,27 +116,9 @@ func doInit(cmd *cobra.Command, args []string) { // nolint: gocyclo
...
-		isSelfSigned = !entity.AllowEntitySignedNodes
-		if viper.GetBool(CfgSelfSigned) {
-			isSelfSigned = true
-		}
-		if !cmdFlags.DebugDontBlameOasis() && !isSelfSigned {
-			logger.Error("entity signed node descriptors are prohibited")
-			os.Exit(1)
-		}
-		defer entitySigner.Reset()
+		logger.Error("missing --node.entity_id command-line argument")
```
kostko, line 120: Minor nit, but it would be nicer if the check above was inverted and the above block was outdented (no `else` needed as the command would exit). E.g. fail if there's no entity ID specified, proceed otherwise.

abukosek: Done :)

kostko (approved):

go/common/entity/entity.go
```diff
@@ -51,10 +51,6 @@ type Entity struct { // nolint: maligned
 	// will sign the descriptor with the node signing key rather than the
 	// entity signing key.
 	Nodes []signature.PublicKey `json:"nodes,omitempty"`
-
-	// AllowEntitySignedNodes is true iff nodes belonging to this entity
-	// may be signed with the entity signing key.
-	AllowEntitySignedNodes bool `json:"allow_entity_signed_nodes,omitempty"`
```
ptrus, line 57: This removal actually breaks opening signed entity blobs that have explicitly set `allow_entity_signed_nodes:false`. There are some such entities on the mainnet. 

```
blob: [164 97 118 1 98 105 100 88 32 253 151 47 28 35 114 40 140 50 20 65 116 239 149 245 194 161 70 132 172 68 133 135 137 69 208 237 163 38 147 11 252 101 110 111 100 101 115 129 88 32 89 189 1 71 226 33 239 43 152 128 47 166 226 209 106 180 243 252 27 136 195 31 191 98 37 166 36 128 220 122 13 142 120 25 97 108 108 111 119 95 101 110 116 105 116 121 95 115 105 103 110 101 100 95 110 111 100 101 115 244]
deserialization into a map:
map[allow_entity_signed_nodes:false id:[253 151 47 28 35 114 40 140 50 20 65 116 239 149 245 194 161 70 132 172 68 133 135 137 69 208 237 163 38 147 11 252] nodes:[[89 189 1 71 226 33 239 43 152 128 47 166 226 209 106 180 243 252 27 136 195 31 191 98 37 166 36 128 220 122 13 142]] v:1]

deserialization into entity error:
msg="failed to fix genesis document" err="unable to open signed entity: cbor: found unknown field at map element index 3"
```

Since this is a breaking change, should probably increase descriptor version and support both versions. Note: no other special logic for old version should be needed, as on any real world network entity signed nodes are not allowed - so if set, the field will always be false.

kostko: Ouch, yeah we should bump the version and support both (version 1 only at genesis). @abukosek can you prepare a PR?

abukosek: Done: https://github.com/oasisprotocol/oasis-core/pull/3748 :)

PR #3744, opened 2021-03-02
runtime/common/quantity: Add Mul/Div implementations

kostko:


abukosek (approved):

runtime/src/common/quantity.rs
```diff
@@ -67,6 +67,117 @@ impl<'a> AddAssign<&'a Quantity> for Quantity {
...
+    }
+}
+
+impl MulAssign<u64> for Quantity {
+    fn mul_assign(&mut self, other: u64) {
+        self.0 *= other;
+    }
+}
+
+impl Div for Quantity {
```
ptrus, line 136: Should we consider instead implementing `num_traits::ops::checked::CheckedDiv` similar as with Sub?

ptrus (approved):

runtime/src/common/quantity.rs
kostko: Good idea, changed.

ptrus: cool! looks good :+1:

PR #3745, opened 2021-03-02
go/worker: Make compute and client node use local storage if available

### LINKED ISSUE #3251: Special case node in both compute/client and storage roles
When the local node is both a storage and compute (or client) node, all storage operations should be tried on the local node first as an optimization.

### LINKED ISSUE #3900: Persist rotating emphemeral TLS keys
By default the node tls keys (used for runtime gRPC communication) are ephemeral, rotated on epoch transitions and only stored in memory.
Since the keys are only stored in memory, on every node restart new set of keys are generated and the keys are not rotated nicely. Additionally, registry will contain wrong set of keys until the node updates its registry descriptor - which is not instantly after the restart.

There are also some edge cases where node will never update its keys in registry, e.g.: when a node graceful shutdown request is made and node restarts before it is shutdown.

Solution to all of the above is persisting the ephemeral keys in node local storage, so that ephemeral keys are persisted across restarts and only rotated on epoch transitions.

### LINKED ISSUE #3909: Flaky governance-upgrade test
The last part of the `governance-upgrade` test seems a bit flaky:

https://github.com/oasisprotocol/oasis-core/blob/05402d85a320a49ced5cebb78c157782ad227565/go/oasis-test-runner/scenario/e2e/runtime/governance_upgrade.go#L486-L497

Looking at logs sometimes this runtime p2p message from client-0 node is not propagated to the compute nodes. Waiting a bit before starting the client (via sleep) fixes the flakiness, but will debug a bit more to see why the p2p message from client is not propagated if started to soon.

jberci:
Closes #3251 

Also fixes #3909 by making the peer manager check for disconnected peers, and closes #3900.

### DIFF
--- a/go/common/crypto/tls/cert.go
+++ b/go/common/crypto/tls/cert.go
@@ -52,13 +52,7 @@ func LoadOrGenerate(certPath, keyPath, commonName string) (*tls.Certificate, err
 	return cert, nil
 }
 
-// Generate generates a new TLS certificate.
-func Generate(commonName string) (*tls.Certificate, error) {
-	pubKey, privKey, err := ed25519.GenerateKey(rand.Reader)
-	if err != nil {
-		return nil, fmt.Errorf("tls: failed to generate keypair: %w", err)
-	}
-
+func makeCertificate(commonName string, privKey ed25519.PrivateKey) (*tls.Certificate, error) {
 	// Tweak the template for the cert.
 	//
 	// TODO: The expiration period is probably too long, and could be reduced,
@@ -70,7 +64,7 @@ func Generate(commonName string) (*tls.Certificate, error) {
 	template.NotBefore = time.Now().Add(-1 * time.Hour)
 	template.NotAfter = time.Now().AddDate(1, 0, 0)
 
-	certDER, err := x509.CreateCertificate(rand.Reader, &template, &template, pubKey, privKey)
+	certDER, err := x509.CreateCertificate(rand.Reader, &template, &template, privKey.Public(), privKey)
 	if err != nil {
 		return nil, fmt.Errorf("tls: failed to create certificate: %w", err)
 	}
@@ -79,6 +73,15 @@ func Generate(commonName string) (*tls.Certificate, error) {
 		Certificate: [][]byte{certDER},
 		PrivateKey:  privKey,
 	}, nil
+}
+
+// Generate generates a new TLS certificate.
+func Generate(commonName string) (*tls.Certificate, error) {
+	_, privKey, err := ed25519.GenerateKey(rand.Reader)
+	if err != nil {
+		return nil, fmt.Errorf("tls: failed to generate keypair: %w", err)
+	}
+	return makeCertificate(commonName, privKey)
 }
 
 // Load loads a TLS certificate and private key.
@@ -94,15 +97,29 @@ func Load(certPath, keyPath string) (*tls.Certificate, error) {
 	return ImportPEM(certPEM, keyPEM)
 }
 
+// LoadFromKey loads a private key and regenerates the whole certificate from it.
+func LoadFromKey(keyPath, commonName string) (*tls.Certificate, error) {
+	keyPEM, err := ioutil.ReadFile(keyPath)
+	if err != nil {
+		return nil, err
+	}
+	key, err := ImportKeyPEM(keyPEM)
+	if err != nil {
+		return nil, err
+	}
+	ed25519Key, ok := key.(ed25519.PrivateKey)
+	if !ok {
+		return nil, fmt.Errorf("tls: private key not an Ed25519 private key")
+	}
+
+	return makeCertificate(commonName, ed25519Key)
+}
+
 // ImportPEM loads a TLS certificate and private key from in-memory PEM blobs.
 func ImportPEM(certPEM, keyPEM []byte) (*tls.Certificate, error) {
-	blk, _ := pem.Decode(keyPEM)
-	if blk == nil || blk.Type != keyPEMType {
-		return nil, fmt.Errorf("tls: failed to parse private key PEM")
-	}
-	key, err := x509.ParsePKCS8PrivateKey(blk.Bytes)
+	key, err := ImportKeyPEM(keyPEM)
 	if err != nil {
-		return nil, fmt.Errorf("tls: failed to parse private key: %w", err)
+		return nil, err
 	}
 
 	cert, err := ImportCertificatePEM(certPEM)
@@ -135,6 +152,19 @@ func ImportCertificatePEM(certPEM []byte) (*tls.Certificate, error) {
 	}, nil
 }
 
+// ImportKeyPEM loads a private key from an in-memory PEM blob.
+func ImportKeyPEM(keyPEM []byte) (interface{}, error) {
+	blk, _ := pem.Decode(keyPEM)
+	if blk == nil || blk.Type != keyPEMType {
+		return nil, fmt.Errorf("tls: failed to parse private key PEM")
+	}
+	key, err := x509.ParsePKCS8PrivateKey(blk.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("tls: failed to parse private key: %w", err)
+	}
+	return key, nil
+}
+
 // Save saves a TLS certificate and private key.
 func Save(certPath, keyPath string, cert *tls.Certificate) error {
 	certPEM, keyPEM, err := ExportPEM(cert)
@@ -150,6 +180,18 @@ func Save(certPath, keyPath string, cert *tls.Certificate) error {
 		return fmt.Errorf("tls: failed to write certificate: %w", err)
 	}
 
+	return nil
+}
+
+// SaveKey saves the private key from a certificate to a file.
+func SaveKey(keyPath string, cert *tls.Certificate) error {
+	_, keyPEM, err := ExportPEM(cert)
+	if err != nil {
+		return err
+	}
+	if err = ioutil.WriteFile(keyPath, keyPEM, 0o600); err != nil {
+		return fmt.Errorf("tls: failed to write private key: %w", err)
+	}
 	return nil
 }
 
--- a/go/common/identity/identity.go
+++ b/go/common/identity/identity.go
@@ -5,6 +5,8 @@ import (
 	"crypto/ed25519"
 	"crypto/rand"
 	"crypto/tls"
+	"fmt"
+	"os"
 	"path/filepath"
 	"sync"
 
@@ -34,6 +36,12 @@ const (
 
 	tlsKeyFilename  = "tls_identity.pem"
 	tlsCertFilename = "tls_identity_cert.pem"
+
+	tlsEphemeralKeyBaseFilename = "tls_ephemeral"
+
+	tlsEphemeralGenCurrent     = ""
+	tlsEphemeralGenNext        = "_next"
+	tlsEphemeralGenRotationNew = "_new_next"
 
 	// These are used for the sentry client connection to the sentry node and are never rotated.
 	tlsSentryClientKeyFilename  = "sentry_client_tls_identity.pem"
@@ -76,6 +84,8 @@ type Identity struct {
 	nextTLSCertificate *tls.Certificate
 	// tlsRotationNotifier is a notifier for certificate rotations.
 	tlsRotationNotifier *pubsub.Broker
+	// dataDir is the directory associated with this identity (for ephemeral key/cert files).
+	dataDir string
 }
 
 // WatchCertificateRotations subscribes to TLS certificate rotation notifications.
@@ -102,19 +112,42 @@ func (i *Identity) RotateCertificates() error {
 	defer i.Unlock()
 
 	if i.tlsCertificate != nil {
+		// Generate a new TLS certificate to be used in the next rotation.
+		newCert, err := tlsCert.Generate(CommonName)
+		if err != nil {
+			return err
+		}
+		newSigner := memory.NewFromRuntime(newCert.PrivateKey.(ed25519.PrivateKey))
+
+		// Save the newly generated ephemeral cert/key.
+		newKeyFile := ephemeralKeyPath(i.dataDir, tlsEphemeralGenRotationNew)
+		err = tlsCert.SaveKey(newKeyFile, newCert)
+		if err != nil {
+			return err
+		}
+
+		// Shuffle files around.
+		for _, names := range []struct {
+			oldGen string
+			newGen string
+		}{
+			{tlsEphemeralGenNext, tlsEphemeralGenCurrent},
+			{tlsEphemeralGenRotationNew, tlsEphemeralGenNext},
+		} {
+			oldKeyPath := ephemeralKeyPath(i.dataDir, names.oldGen)
+			newKeyPath := ephemeralKeyPath(i.dataDir, names.newGen)
+			if err = os.Rename(oldKeyPath, newKeyPath); err != nil {
+				return err
+			}
+		}
+
 		// Use the prepared certificate.
 		if i.nextTLSCertificate != nil {
 			i.tlsCertificate = i.nextTLSCertificate
 			i.tlsSigner = i.nextTLSSigner
 		}
-
-		// Generate a new TLS certificate to be used in the next rotation.
-		var err error
-		i.nextTLSCertificate, err = tlsCert.Generate(CommonName)
-		if err != nil {
-			return err
-		}
-		i.nextTLSSigner = memory.NewFromRuntime(i.nextTLSCertificate.PrivateKey.(ed25519.PrivateKey))
+		i.nextTLSCertificate = newCert
+		i.nextTLSSigner = newSigner
 
 		i.tlsRotationNotifier.Broadcast(struct{}{})
 	}
@@ -234,13 +267,13 @@ func doLoadOrGenerate(dataDir string, signerFactory signature.SignerFactory, sho
 		// Load successful, ensure that we won't ever rotate the certificates.
 		dnr = true
 	} else {
-		// Freshly generate TLS certificates.
-		cert, err = tlsCert.Generate(CommonName)
-		if err != nil {
-			return nil, err
-		}
-
 		if persistTLS {
+			// Freshly generate TLS certificates.
+			cert, err = tlsCert.Generate(CommonName)
+			if err != nil {
+				return nil, err
+			}
+
 			// Save generated TLS certificate to disk.
 			err = tlsCert.Save(tlsCertPath, tlsKeyPath, cert)
 			if err != nil {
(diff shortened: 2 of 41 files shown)


kostko (commented):
There's also a related comment in `group.go` (in `NewGroup`) that should be removed after this is done. Actually I guess the initial idea was to have that there instead of in the executor node specifically. Maybe it would make more sense to move it there?

go/oasis-test-runner/oasis/network.go
```diff
@@ -0,0 +1,1029 @@
...
+				)
+				return nil, fmt.Errorf("oasis/network: failed to add log watcher for %s: %w", name, err)
+			}
+		}
+
+		net.nodes = append(net.nodes, node)
+	}
+
+	if cfg != nil {
+		cfg.Into(node)
```
kostko, line 281: It seems this is performed twice (also above).

go/runtime/client/submitter.go
```diff
@@ -71,6 +75,17 @@ func (w *txSubmitter) checkBlock(blk *block.Block) error {
 		return nil
 	}
 
+	// If we have local storage, wait for it to sync up to this point.
+	if w.storageCommitteeNode != nil {
+		select {
+		case <-w.storageCommitteeNode.WaitForRound(blk.Header.Round):
```
kostko, line 81: I think this would be more general if it was handled in the storage backend wrapper exposed by the storage node (e.g., used with `RegisterStorage`) instead of directly here. Because there can be other uses of storage in the client when it is also a storage node.

E.g., create a simple wrapper in storage worker's `registerRuntime` that basically makes it wait for rounds for all operations (note only the backend that is wrapped is the one passed to `RegisterStorage`).

go/oasis-test-runner/scenario/e2e/runtime/multihost.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Assign some compute, client and storage workers to shared nodes.
+	f.ComputeWorkers[0].Name = "compute-storage-0"
+	f.StorageWorkers[0].Name = "compute-storage-0"
+
+	f.Clients[0].Name = "client-storage-0"
+	f.StorageWorkers[1].Name = "client-storage-0"
```
ptrus, line 38: maybe also test compute-storage-client node (could have two versions of the scenario)

go/oasis-test-runner/oasis/args.go
```diff
@@ -598,20 +656,60 @@ func (args *argBuilder) byzantineVersionFakeEnclaveID(rt *Runtime) *argBuilder {
...
+					ok = false
+					break
+				}
+			}
+			if ok {
+				output = append(output, arg.items...)
+				multiValued[leader] = append(multiValued[leader], arg.items)
+			}
+		}
+		if _, ok := shipped[leader]; !ok && !arg.multiValued {
```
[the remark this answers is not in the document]
ptrus: Ah although the way the "MultiHost" test is implemented it looks like the values will get overridden (even with different values).

go/worker/common/p2p/dispatch.go
```diff
@@ -177,14 +178,25 @@ func (h *topicHandler) dispatchMessage(peerID core.PeerID, m *queuedMsg, isIniti
...
+		switch {
+		case errors.Is(err, p2pError.ErrUnhandledMessage):
+			// Nothing to do here, continue to next handler.
+		case err != nil:
 			h.logger.Error("failed to handle message", "message", m.msg, "from", m.from, "err", err)
 			return err
 		}
 	}
+	if !handled {
+		return p2pError.EnsurePermanent(p2pError.ErrUnhandledMessage)
```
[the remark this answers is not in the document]
kostko: Oh that's a good point, haven't thought about retries. Yeah that makes sense.

go/oasis-test-runner/oasis/args.go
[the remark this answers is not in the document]
jberci: It's not really clear what to do here. Every feature needs to provide all options, even non-multivalued ones, since every feature needs to be able to fully configure a node by itself. On the other hand, it's perfectly legitimate to have multiple occurrences of a single value option in the argument list if there are multiple features per node, so some sort of merging is definitely needed.
One thing that could make sense would be to panic if their values differ.

go/oasis-test-runner/oasis/network.go
```diff
@@ -0,0 +1,954 @@
...
+
+		node = &Node{
+			Name:           name,
+			net:            net,
+			dir:            nodeDir,
+			assignedPorts:  map[string]uint16{},
+			hostedRuntimes: map[common.Namespace]*hostedRuntime{},
+		}
+
+		if cfg != nil { // XXX works for ias, seed.
```
kostko, line 269: What's with the comment?

go/oasis-test-runner/oasis/network.go
```diff
@@ -0,0 +1,954 @@
...
+			cfg.Into(node)
```
kostko, line 270: This seems to be performed twice (also below).

go/oasis-test-runner/oasis/oasis.go
```diff
@@ -144,6 +181,31 @@ func (n *Node) LoadIdentity() (*identity.Identity, error) {
...
+// Start starts the node.
+func (n *Node) Start() error {
+	args := newArgBuilder()
+	for _, f := range n.features {
+		if err := f.AddArgs(args); err != nil {
+			return fmt.Errorf("oasis/node: failed to add arguments for feature on node %s: %w", n.Name, err)
+		}
+		if cf, ok := f.(CustomStartFeature); ok {
+			if len(n.features) != 1 {
+				panic("oasis/node: trying to perform customized node start with multiple features attached")
```
kostko, line 193: I assume a fixture can still configure this? It resulting in a panic is probably not ideal and it may be better if this returned an error instead?

go/runtime/registry/registry.go
```diff
@@ -89,6 +93,12 @@ type Runtime interface {
 	// RegisterStorage sets the given local storage backend for the runtime.
 	RegisterStorage(storage storageAPI.Backend)
 
+	// AddRoles adds available node roles to the runtime.
+	AddRoles(roles node.RolesMask)
+
+	// HasRoles checks if the runtime has all of the roles specified.
```
kostko, line 99:
```suggestion
	// HasRoles checks if the node has all of the roles specified for this runtime.
```

go/oasis-test-runner/oasis/oasis.go
jberci: The idea here was that if a feature requires completely custom startup, it's likely to clash with everything else, since all the "regular" features probably also expect regular startup. In that sense, I don't think just returning an error makes sense. Alternatively, it could just let all features build up the args and then check if there's at most one custom feature - then panic if there's more, otherwise hand over startup to it at the end.

go/worker/storage/committee/node.go
```diff
@@ -402,9 +406,18 @@ func (n *Node) WaitForRound(round uint64, root *storageApi.Root) (<-chan uint64,
...
+	consensusRound := n.commonNode.CurrentBlock.Header.Round
+	n.commonNode.CrossNode.Unlock()
+
 	n.syncedLock.Lock()
 	defer n.syncedLock.Unlock()
 
+	if round > consensusRound+roundWaitConsensusOffset && !commonFlags.DebugDontBlameOasis() {
+		close(retCh)
+		return nil, storageApi.ErrVersionNotFound
+	}
```
kostko, line 419: This can probably be moved outside the `syncedLock` section.

go/oasis-test-runner/oasis/oasis.go
kostko: So the problem I have with panicking is that this could be configured through an external fixture (so user configuration) and then would cause the test runner to just crash which seems non-ideal? It would be better if it exited with a nicer error instead. But not a critical thing, we can start with a panic and improve usability later.

kostko (approved):

go/runtime/registry/registry.go
```diff
@@ -410,8 +435,39 @@ func (r *runtimeRegistry) NewUnmanagedRuntime(ctx context.Context, runtimeID com
...
+func (r *runtimeRegistry) AddRoles(roles node.RolesMask, runtimeID *common.Namespace) error {
+	r.RLock()
+	defer r.RUnlock()
+
+	if runtimeID != nil {
+		rt, ok := r.runtimes[*runtimeID]
+		if !ok {
+			return fmt.Errorf("runtime/registry: runtime %s is not supported", *runtimeID)
+		}
+		rt.AddRoles(roles)
```
ptrus, line 447: Might be nicer to `reutrn nil` here early, and then the `else` part can be one level higher.

ptrus (approved):

go/worker/common/p2p/dispatch.go
```diff
@@ -125,15 +125,14 @@ func (h *topicHandler) topicMessageValidator(ctx context.Context, unused core.Pe
...
 	defer func() {
-		if err == nil || !isInitial {
+		if retErr == nil || !isInitial {
 			return
 		}
-		if p2pError.IsPermanent(err) {
+		if p2pError.IsPermanent(retErr) {
 			h.logger.Error("failed to dispatch message in-line, not retrying",
-				"err", err,
+				"retErr", retErr,
```
kostko, line 135:
```suggestion
				"err", retErr,
```

go/worker/common/p2p/dispatch.go
```diff
@@ -151,7 +150,7 @@ func (h *topicHandler) dispatchMessage(peerID core.PeerID, m *queuedMsg, isIniti
 			}
 			if atomic.CompareAndSwapUint64(&h.numWorkers, numWorkers, numWorkers+1) {
 				h.logger.Error("failed to dispatch message in-line, retrying",
-					"err", err,
+					"retErr", retErr,
```
kostko, line 153:
```suggestion
					"err", retErr,
```

go/common/identity/identity.go
```diff
@@ -35,6 +37,13 @@ const (
 	tlsKeyFilename  = "tls_identity.pem"
 	tlsCertFilename = "tls_identity_cert.pem"
 
+	tlsEphemeralKeyBaseFilename  = "tls_ephemeral"
+	tlsEphemeralCertBaseFilename = "tls_ephemeral_cert"
+
+	tlsEphemeralGenCurrent     = ""
```
ptrus, line 43: Hm, could also just serialize the keys and store them in the localstorage badger db? I think it would simplify some of the shuffling as the rotation could be done in a transaction. 
_I feel_ it would also be a bit nicer as the keys would be a bit more hidden from the user (which i think makes sense, as these should be ephemeral).

Am also fine with this if it stays as its.

go/common/identity/identity.go
```diff
@@ -250,13 +287,33 @@ func doLoadOrGenerate(dataDir string, signerFactory signature.SignerFactory, sho
 			// Disable TLS rotation if we're persisting TLS certificates.
 			dnr = true
 		} else {
-			// Not persisting TLS certificate to disk, generate a new
-			// certificate to be used in the next rotation.
-			nextCert, err = tlsCert.Generate(CommonName)
+			// Current key; try loading, else generate, then save.
+			ephemeralCertPath, ephemeralKeyPath := ephemeralCertPaths(dataDir, tlsEphemeralGenCurrent)
+			cert, err = tlsCert.Load(ephemeralCertPath, ephemeralKeyPath)
```
ptrus, line 292: Should probably log something in case this is not a `os.IsNotExist(err)`

go/worker/common/p2p/dispatch.go
```diff
@@ -236,6 +245,57 @@ func (h *topicHandler) retryWorker(m *queuedMsg) {
 	}
 }
 
+func (h *topicHandler) tryPublishing(rawMsg []byte) error {
+	peers := len(h.topic.ListPeers())
```
ptrus, line 249: No need for a separate variable, only used in the line bellow.

go/common/identity/identity.go
kostko: Given that we are storing all the other keys in PEM files in the data directory I think it makes it more consistent to store these in the same way. It is true that these are ephemeral though.

go/common/identity/identity.go
```diff
@@ -35,6 +37,13 @@ const (
 	tlsKeyFilename  = "tls_identity.pem"
 	tlsCertFilename = "tls_identity_cert.pem"
 
+	tlsEphemeralKeyBaseFilename  = "tls_ephemeral"
+	tlsEphemeralCertBaseFilename = "tls_ephemeral_cert"
```
kostko, line 41: There is no need/reason to also store the certificate as we use public key pinning when verifying certificates, so only the public key really matters. Feel free to remove this part and regenerate the certificate.

go/worker/common/p2p/dispatch.go
```diff
@@ -62,11 +68,14 @@ type topicHandler struct {
 	p2p *P2P
 
 	topic       *pubsub.Topic
+	host        core.Host
 	cancelRelay pubsub.RelayCancelFunc
 	handlers    []Handler
 
 	numWorkers uint64
 
+	rawMsgQueue chan *rawMessage
```
ptrus, line 77: I think a more descriptive name like: `pendingQueue` (or something similar) would be better, as the type already tells that this is for `rawMsgs`.

go/worker/common/p2p/dispatch.go
```diff
@@ -236,6 +245,57 @@ func (h *topicHandler) retryWorker(m *queuedMsg) {
...
+			return nil
+		default:
+			return fmt.Errorf("worker/common/p2p: message queue overflow, libp2p still not initialized")
+		}
+	}
+
+	return h.topic.Publish(h.ctx, rawMsg)
+}
+
+func (h *topicHandler) publishWorker() {
```
ptrus, line 266: If you rename the channel name, then this could also be:

```suggestion
func (h *topicHandler) pendingMessagesWorker() {
```

With a docstring that this is handles messages when there's no connected peers.

go/worker/common/p2p/peermgmt.go
```diff
@@ -162,6 +222,36 @@ func (mgr *PeerManager) watchRegistryNodes(consensus consensus.Backend) {
 			if nodeEv.IsRegistration {
 				_ = mgr.UpdateNode(nodeEv.Node)
 			}
+		case <-ticker.C:
+			mgr.logger.Debug("ticker fired")
```
ptrus, line 226: This  log message is probably leftover from testing, doesn't seem that useful to keeping it.

go/worker/common/p2p/peermgmt.go
```diff
@@ -162,6 +222,36 @@ func (mgr *PeerManager) watchRegistryNodes(consensus consensus.Backend) {
...
+			func() {
+				mgr.Lock()
+				defer mgr.Unlock()
+
```
ptrus, line 230: Can stop early if no peers and then drop the check bellow.

```suggestion
if len(mgr.peers) == 0 {
  return
}
```

go/worker/common/p2p/init.go
```diff
@@ -18,6 +18,9 @@ const (
 	CfgP2PPeerOutboundQueueSize = "worker.p2p.peer_outbound_queue_size"
 	// CfgP2PValidateQueueSize sets the libp2p gossipsub buffer size of the validate queue.
 	CfgP2PValidateQueueSize = "worker.p2p.validate_queue_size"
+	// CfgP2PConnectednessLowWater sets the ratio of connected to unconnected peers at which
+	// the peer manager will try to reconnect to disconnected nodes.
+	CfgP2PConnectednessLowWater = "worker.p2p.connectedness_low_water"
```
ptrus, line 23: Would it be too much to always try to reconnect to disconnected nodes?  If anything, a "per peer" back-off would make more sense if we feel trying to reconnect every 5 secs is to excessive.

ptrus (approved):
Some minor comments.

Thanks for getting rid of some flakiness :grin:

go/worker/common/p2p/peermgmt.go
```diff
@@ -113,21 +155,36 @@ func (mgr *PeerManager) updateNodeLocked(node *node.Node, peerID core.PeerID) {
 	// way, then don't bother doing anything.
 	if oldNode := mgr.peers[peerID]; oldNode != nil {
 		if oldNode.addrHash.Equal(&addrHash) {
-			return
+			// But, if the peer isn't connected for some reason, still try to reconnect to it.
+			if mgr.host.Network().Connectedness(peerID) == network.Connected {
+				mgr.logger.Debug("updateNodeLocked addresses unchanged and peer still connected",
```
kostko, line 160: Since the log entry will include the line number this can be just:
```suggestion
				mgr.logger.Debug("addresses unchanged and peer still connected",
```

go/worker/common/p2p/peermgmt.go
```diff
@@ -113,21 +155,36 @@ func (mgr *PeerManager) updateNodeLocked(node *node.Node, peerID core.PeerID) {
...
 		}
 	}
 
 	// Cancel any existing connection attempts to the peer.
 	mgr.removePeerLocked(peerID)
 
+	// Also don't bother doing anything if the address list is empty.
+	// It's unlikely for connection attempts to no addresses to succeed.
+	if len(node.P2P.Addresses) == 0 {
+		mgr.logger.Debug("no addresses to connect to", "node_id", node.ID, "peerID", peerID)
```
kostko, line 175:
```suggestion
		mgr.logger.Debug("no addresses to connect to", "node_id", node.ID, "peer_id", peerID)
```

go/worker/common/p2p/peermgmt.go
```diff
@@ -162,6 +222,36 @@ func (mgr *PeerManager) watchRegistryNodes(consensus consensus.Backend) {
...
+				mgr.Lock()
+				defer mgr.Unlock()
+
+				connected := 0
+				for peerID := range mgr.peers {
+					if mgr.host.Network().Connectedness(peerID) == network.Connected {
+						connected++
+					}
+				}
+				mgr.logger.Debug("peer manager counted connected peers", "counted", connected)
```
kostko, line 237:
```suggestion
				mgr.logger.Debug("peer manager counted connected peers", "num_connected_peers", connected)
```

go/common/crypto/tls/cert.go
```diff
@@ -52,13 +52,7 @@ func LoadOrGenerate(certPath, keyPath, commonName string) (*tls.Certificate, err
...
 }
 
-// Generate generates a new TLS certificate.
-func Generate(commonName string) (*tls.Certificate, error) {
-	pubKey, privKey, err := ed25519.GenerateKey(rand.Reader)
-	if err != nil {
-		return nil, fmt.Errorf("tls: failed to generate keypair: %w", err)
-	}
-
+func makeCertificate(commonName string, pubKey, privKey interface{}) (*tls.Certificate, error) {
```
kostko, line 55: Should not take `interface{}` for the keys, we require Ed25519 keys during certificate verification, so would be good to make it type safe.

go/worker/common/p2p/init.go
jberci: I changed this to a combination of both: the new code takes care to preserve the connect backoff that was already there for cases where the connect worker is already running (e.g. if there's a disconnected peer and the addresses remain unchanged), and there's a periodic sweep that triggers reconnecting on disconnected peers (which also implicitly handles the corner case where the connect worker was exiting for some reason and the peer still isn't connected).

go/common/crypto/tls/cert.go
```diff
@@ -52,13 +53,7 @@ func LoadOrGenerate(certPath, keyPath, commonName string) (*tls.Certificate, err
...
 }
 
-// Generate generates a new TLS certificate.
-func Generate(commonName string) (*tls.Certificate, error) {
-	pubKey, privKey, err := ed25519.GenerateKey(rand.Reader)
-	if err != nil {
-		return nil, fmt.Errorf("tls: failed to generate keypair: %w", err)
-	}
-
+func makeCertificate(commonName string, pubKey crypto.PublicKey, privKey ed25519.PrivateKey) (*tls.Certificate, error) {
```
kostko, line 56: Why do you need both public and private keys here when you can do `privKey.Public()` to get the public part?

kostko (approved):

Issue #3444, opened 2020-10-23
Add non-blocking runtime SubmitTx variant


ptrus:
Similar as it was done for consensus (https://github.com/oasisprotocol/oasis-core/issues/3152), a `SubmitTxNoWait` runtime client method that publishes the runtime transaction would be useful.

PR #3746, opened 2021-03-02
go/runtime/client: add SubmitTxNoWait method

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3444

TODO: 
- [x] do some more updates in `client/watcher.go` (now `client/submitter.go`) it is not just a watcher anymore but also submits transactions. Also removed the implementation of `BackgroundService` trait in the submitter for some minor simplifications -   since this is a `runtime/client` package private struct only used by the client (which remains a `BackgroundService`) i think that's fine.

### DIFF
--- a/go/runtime/client/client.go
+++ b/go/runtime/client/client.go
@@ -28,7 +28,6 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/runtime/transaction"
 	storage "github.com/oasisprotocol/oasis-core/go/storage/api"
 	"github.com/oasisprotocol/oasis-core/go/worker/common/p2p"
-	executor "github.com/oasisprotocol/oasis-core/go/worker/compute/executor/api"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


go/runtime/client/client.go
```diff
@@ -158,27 +167,18 @@ func (c *runtimeClient) SubmitTx(ctx context.Context, request *api.SubmitTxReque
...
-			},
-			GroupVersion: resp.groupVersion,
-		})
+// Implements api.RuntimeClient.
+func (c *runtimeClient) SubmitTxNoWait(ctx context.Context, request *api.SubmitTxRequest) error {
+	_, err := c.submitTx(ctx, request)
+	if err != nil {
+		return err
 	}
+	return nil
```
kostko, line 181: Could just be `return err`?

kostko (commented):
Cool, also moves the retry/submission logic into the "watcher".

go/runtime/client/client.go
```diff
@@ -561,8 +550,8 @@ func (c *runtimeClient) Start() error {
 // Implements service.BackgroundService.
 func (c *runtimeClient) Stop() {
 	// Watchers.
-	for _, watcher := range c.watchers {
-		watcher.Stop()
+	for _, submitter := range c.txSubmitters {
```
kostko, line 553: Just noticed that this loop (and below) should probably be under a `c.Lock()` guard as in theory a request could come in right when this is shutting down?

kostko (approved):
Looks good to me, maybe @jberci can you take a quick look as well.

go/runtime/client/client.go
ptrus: nice find! yeah indeed should be, will update.

jberci (approved):

PR #3748, opened 2021-03-03
go/common/entity: Fix backward compatibility

abukosek:
Removing `AllowEntitySignedNodes` from the `Entity` descriptor was apparently not a good idea, since some entities on the mainnet had the field set.  This PR fixes that by adding the field back and enforcing that it's always set to `false`.

### DIFF
--- a/go/common/entity/entity.go
+++ b/go/common/entity/entity.go
@@ -32,7 +32,7 @@ var (
 const (
 	// LatestEntityDescriptorVersion is the latest entity descriptor version that should be used for
 	// all new descriptors. Using earlier versions may be rejected.
-	LatestEntityDescriptorVersion = 1
+	LatestEntityDescriptorVersion = 2
 
 	// Minimum and maximum descriptor versions that are allowed.
 	minEntityDescriptorVersion = 1
@@ -51,6 +51,12 @@ type Entity struct { // nolint: maligned
 	// will sign the descriptor with the node signing key rather than the
 	// entity signing key.
 	Nodes []signature.PublicKey `json:"nodes,omitempty"`
+
+	// AllowEntitySignedNodes must always be set to false in v1 descriptors
+	// and left unset in v2 descriptors.  It is kept only for backward
+	// compatibility when deserializing existing descriptors that might
+	// have had this set.
+	AllowEntitySignedNodes bool `json:"allow_entity_signed_nodes,omitempty"`
 }
 
 // ValidateBasic performs basic descriptor validity checks.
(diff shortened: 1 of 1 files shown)


ptrus (commented):
There's nothing enforcing that V2 descriptors do have `AllowEntitySignedNodes` unset. e.g. with this change, one can still submit a V2 descriptor cbor blob with explicitly set `AllowEntitySignedNodes: false` and it will be accepted, since `omitempty` is not strict in deserialization.

abukosek:
There's no nice way to check for this in Go, since the struct will always contain the field.  I think the proper way to fix this would be to add a pass to `fixgenesis` that removes `AllowEntitySignedNodes` from all the existing descriptors and then we can remove the field as well.

ptrus:
> add a pass to `fixgenesis` that removes `AllowEntitySignedNodes` from all the existing descriptors and then we can remove the field as well.

Hm i don't think we can do that, since the descriptors are stored signed in genesis. 

I _feel_ that the correct fix would be something like: omitting the field from the struct for V2 and having a custom deserializer for handling V1 descriptors by (ignoring the `AllowEntitySignedNodes` fields in case it's false) - and only allowing V1 in genesis. After V1 is not supported anymore, we can drop the custom deserializer.

abukosek:
Ah, I see we can't do that automatically, since the genesis file has `SignedEntity` structs, so while we could open and convert them into the new format, we can't re-sign them.

I think the best solution is to just bump the `LatestEntityDescriptorVersion` to `2` and `minEntityDescriptorVersion` to `2` and then tell the people who have entity descriptors with the `AllowEntitySignedNodes` set to regenerate them.  Any other ideas? (How many entity descriptors are affected anyways?)

abukosek:
Actually, even with the custom deserializer, we won't be able automatically convert the descriptors to v2, since they're stored as signed in the registry state as well (`consensus/tendermint/apps/registry/state/state.go`), so the entity's owner would still have to regenerate the descriptor at some point.

I guess we have two options:
1. Keep the struct with the removed `AllowEntitySignedNodes` field and just bump the version to 2.  Entity descriptors that have the field set need to be regenerated on the next upgrade.
2. Add the `AllowEntitySignedNodes` field back in, keep the version at 1, and just enforce that the field is always set to false.  Existing entity descriptors can stay as they are.

ptrus:
> Actually, even with the custom deserializer, we won't be able automatically convert the descriptors to v2, since they're stored as signed in the registry state as well

We wouldn't convert them, we'd support both versions (but not allow new registrations with non-latest version).

abukosek:
Closing this in favor of other solution mentioned on Slack :)

tjanez:
The revised implementation is in #3754.

Issue #3743, opened 2021-03-02
GetStatus endpoint should include the current epoch


kostko:
It would be useful if the `GetStatus` endpoint included the current epoch.

PR #3753, opened 2021-03-04
go/control/GetStatus: include the latest epoch

ptrus:
Fixes: #3743

### DIFF
--- a/go/consensus/tendermint/full/full.go
+++ b/go/consensus/tendermint/full/full.go
@@ -788,6 +788,13 @@ func (t *fullService) GetStatus(ctx context.Context) (*consensusAPI.Status, erro
 			status.LatestHash = latestBlk.Hash
 			status.LatestTime = latestBlk.Time
 			status.LatestStateRoot = latestBlk.StateRoot
+
+			var epoch beaconAPI.EpochTime
+			epoch, err = t.beacon.GetEpoch(ctx, status.LatestHeight)
+			if err != nil {
+				return nil, fmt.Errorf("failed to fetch epoch: %w", err)
+			}
+			status.LatestEpoch = epoch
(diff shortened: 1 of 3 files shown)


go/consensus/tendermint/full/full.go
```diff
@@ -794,6 +794,13 @@ func (t *fullService) GetStatus(ctx context.Context) (*consensusAPI.Status, erro
 			return nil, fmt.Errorf("failed to fetch current block: %w", err)
 		}
 
+		// Latest block epoch.
+		epoch, err := t.beacon.GetEpoch(ctx, status.LatestHeight)
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch epoch: %w", err)
```
kostko, line 800: Could this fail when there's no committed blocks? Maybe it should only query the epoch in case if successfully fetches the block above.

kostko: Nice, this was actually caught by the early query E2E test :-)

ptrus: :grin: oh yeah i thought we'll fail early above if there's no committed block, missed the exception.

go/consensus/tendermint/full/full.go
```diff
@@ -788,6 +788,11 @@ func (t *fullService) GetStatus(ctx context.Context) (*consensusAPI.Status, erro
 			status.LatestHash = latestBlk.Hash
 			status.LatestTime = latestBlk.Time
 			status.LatestStateRoot = latestBlk.StateRoot
+			epoch, err := t.beacon.GetEpoch(ctx, status.LatestHeight)
```
kostko, line 791: Maybe add an empty line above.

kostko (approved):

PR #3754, opened 2021-03-04
go/common/entity: Fix backward compatibility

abukosek:
This PR adds a custom deserializer to the `Entity` struct that can handle both v1 and v2 descriptors.

### DIFF
--- a/go/common/entity/entity.go
+++ b/go/common/entity/entity.go
@@ -32,7 +32,7 @@ var (
 const (
 	// LatestEntityDescriptorVersion is the latest entity descriptor version that should be used for
 	// all new descriptors. Using earlier versions may be rejected.
-	LatestEntityDescriptorVersion = 1
+	LatestEntityDescriptorVersion = 2
 
 	// Minimum and maximum descriptor versions that are allowed.
 	minEntityDescriptorVersion = 1
@@ -51,6 +51,46 @@ type Entity struct { // nolint: maligned
 	// will sign the descriptor with the node signing key rather than the
 	// entity signing key.
 	Nodes []signature.PublicKey `json:"nodes,omitempty"`
+}
+
+// UnmarshalCBOR is a custom deserializer that handles both v1 and v2 Entity
+// structures.  A v1 structure is converted to v2 seamlessly if the field
+// AllowEntitySignedNodes is false or missing, otherwise an error is returned.
+func (e *Entity) UnmarshalCBOR(data []byte) error {
+	// Determine Entity structure version.
+	v, err := cbor.GetVersion(data)
+	if err != nil {
+		return err
+	}
+	switch v {
+	case 1:
+		// Old version had an extra field that was used only for debugging/tests.
+		type EntityV1 struct { // nolint: maligned
+			cbor.Versioned
+			ID                     signature.PublicKey   `json:"id"`
+			Nodes                  []signature.PublicKey `json:"nodes,omitempty"`
+			AllowEntitySignedNodes bool                  `json:"allow_entity_signed_nodes,omitempty"`
+		}
+		var ev1 EntityV1
+		if err = cbor.Unmarshal(data, &ev1); err != nil {
+			return err
+		}
+		// Make sure that AllowEntitySignedNodes is not enabled.
+		if ev1.AllowEntitySignedNodes {
+			return fmt.Errorf("entity descriptor must have allow_entity_signed_nodes set to false")
+		}
+		// Convert into new format.
+		e.Versioned = cbor.NewVersioned(2)
+		e.ID = ev1.ID
+		e.Nodes = ev1.Nodes
+		return nil
+	case 2:
+		// New version, call the default unmarshaler.
+		type ev2 Entity
+		return cbor.Unmarshal(data, (*ev2)(e))
+	default:
+		return fmt.Errorf("invalid entity descriptor version: %v", v)
+	}
 }
 
 // ValidateBasic performs basic descriptor validity checks.
--- a/go/common/entity/entity_test.go
+++ b/go/common/entity/entity_test.go
@@ -1,0 +1,60 @@
+package entity
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/oasisprotocol/oasis-core/go/common/cbor"
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
+	memorySigner "github.com/oasisprotocol/oasis-core/go/common/crypto/signature/signers/memory"
+)
+
+func TestEntityDescriptorVersioning(t *testing.T) {
+	require := require.New(t)
+
+	type EntityV1 struct {
+		Entity
+		AllowEntitySignedNodes bool `json:"allow_entity_signed_nodes,omitempty"`
+	}
+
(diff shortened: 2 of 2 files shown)


go/common/entity/entity_test.go
```diff
@@ -0,0 +1,60 @@
...
+	var uv1t2 Entity
+	ev1.AllowEntitySignedNodes = false
+	require.NoError(cbor.Unmarshal(cbor.Marshal(ev1), &uv1t2), "v1 unmarshal with no AllowEntitySignedNodes field set to false should pass")
+	require.EqualValues(ev1.ID, uv1t2.ID)
+	require.EqualValues(ev1.Nodes, uv1t2.Nodes)
+	require.EqualValues(cbor.NewVersioned(2), uv1t2.Versioned)
+
+	var uv1t3 Entity
+	ev1.AllowEntitySignedNodes = true
+	require.Error(cbor.Unmarshal(cbor.Marshal(ev1), &uv1t3), "v1 unmarshal with no AllowEntitySignedNodes field set to true should fail")
```
kostko, line 44:
```suggestion
	require.Error(cbor.Unmarshal(cbor.Marshal(ev1), &uv1t3), "v1 unmarshal with AllowEntitySignedNodes field set to true should fail")
```

go/common/entity/entity_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	var uv1t1 Entity
+	require.NoError(cbor.Unmarshal(cbor.Marshal(ev1), &uv1t1), "v1 unmarshal with no AllowEntitySignedNodes field should pass")
+	require.EqualValues(ev1.ID, uv1t1.ID)
+	require.EqualValues(ev1.Nodes, uv1t1.Nodes)
+	require.EqualValues(cbor.NewVersioned(2), uv1t1.Versioned)
+
+	var uv1t2 Entity
+	ev1.AllowEntitySignedNodes = false
+	require.NoError(cbor.Unmarshal(cbor.Marshal(ev1), &uv1t2), "v1 unmarshal with no AllowEntitySignedNodes field set to false should pass")
```
kostko, line 37:
```suggestion
	require.NoError(cbor.Unmarshal(cbor.Marshal(ev1), &uv1t2), "v1 unmarshal with AllowEntitySignedNodes field set to false should pass")
```

go/common/entity/entity_test.go
```diff
@@ -0,0 +1,60 @@
...
+	require.Error(cbor.Unmarshal(cbor.Marshal(ev1), &uv1t3), "v1 unmarshal with no AllowEntitySignedNodes field set to true should fail")
+
+	k2 := memorySigner.NewTestSigner("test entity v2")
+	k2n1 := memorySigner.NewTestSigner("test entity v2 node 1")
+	k2n2 := memorySigner.NewTestSigner("test entity v2 node 2")
+	ev2 := Entity{
+		ID:    k2.Public(),
+		Nodes: []signature.PublicKey{k2n1.Public(), k2n2.Public()},
+	}
+	ev2.V = 2
```
kostko, line 53: Could use `Versioned: cbor.NewVersioned(2)` above.

go/common/entity/entity_test.go
```diff
@@ -0,0 +1,60 @@
...
+		AllowEntitySignedNodes bool `json:"allow_entity_signed_nodes,omitempty"`
+	}
+
+	k1 := memorySigner.NewTestSigner("test entity v1")
+	k1n1 := memorySigner.NewTestSigner("test entity v1 node 1")
+	k1n2 := memorySigner.NewTestSigner("test entity v1 node 2")
+	var ev1 EntityV1
+	ev1.ID = k1.Public()
+	ev1.Nodes = []signature.PublicKey{k1n1.Public(), k1n2.Public()}
+	ev1.V = 1
```
kostko, line 27: Could use `ev1.Versioned = cbor.NewVersioned(1)`. Also you could just do `ev1 := EntityV1{ ...` above.

kostko (commented):
Some minor comments, otherwise looks good!

go/common/entity/entity_test.go
abukosek: Compiler says no:
```
common/entity/entity_test.go:25:3: cannot use promoted field Entity.Versioned in struct literal of type EntityV1
common/entity/entity_test.go:26:3: cannot use promoted field Entity.ID in struct literal of type EntityV1
common/entity/entity_test.go:27:3: cannot use promoted field Entity.Nodes in struct literal of type EntityV1
```
:D

kostko: :disappointed:

kostko (approved):

go/common/entity/entity.go
```diff
@@ -53,6 +53,59 @@ type Entity struct { // nolint: maligned
...
+			return fmt.Errorf("entity descriptor must have allow_entity_signed_nodes set to false")
+		}
+		// Convert into new format.
+		e.Versioned = cbor.NewVersioned(2)
+		e.ID = ev1.ID
+		e.Nodes = ev1.Nodes
+		return nil
+	case 2:
+		// New version, call the default unmarshaler.
+		// Unfortunately, we can't just use the existing Entity struct, because
```
ptrus, line 89: I _think_ you could also do
```
case 2:
    type e2 Entity
    return cbor.Unmarshal(data, (*e2)(e))
```

abukosek: No, that doesn't work, see coment above -- it causes an infinite loop :)

ptrus: Hm, i think it should, unless i'm doing something weird:

```
root@f23e01d9066c:/code/go/common/entity# git diff origin/andrej/bugfix/entity-v2-again
diff --git a/go/common/entity/entity.go b/go/common/entity/entity.go
index c9d61a5c4..76aa5d430 100644
--- a/go/common/entity/entity.go
+++ b/go/common/entity/entity.go
@@ -85,22 +85,8 @@ func (e *Entity) UnmarshalCBOR(data []byte) error {
                e.Nodes = ev1.Nodes
                return nil
        case 2:
-               // New version, call the default unmarshaler.
-               // Unfortunately, we can't just use the existing Entity struct, because
-               // it would cause an infinite loop when calling Unmarshal.
-               type EntityV2 struct { // nolint: maligned
-                       cbor.Versioned
-                       ID    signature.PublicKey   `json:"id"`
-                       Nodes []signature.PublicKey `json:"nodes,omitempty"`
-               }
-               var ev2 EntityV2
-               if err = cbor.Unmarshal(data, &ev2); err != nil {
-                       return err
-               }
-               e.Versioned = ev2.Versioned
-               e.ID = ev2.ID
-               e.Nodes = ev2.Nodes
-               return nil
+               type e2 Entity
+               return cbor.Unmarshal(data, (*e2)(e))
        default:
                return fmt.Errorf("invalid entity descriptor version: %v", v)
        }


root@f23e01d9066c:/code/go/common/entity# go test
PASS
ok  	github.com/oasisprotocol/oasis-core/go/common/entity	0.004s

```

abukosek: OK, that's weird, it passes for me now as well :man_shrugging:  I'll update the code, thanks :)

ptrus: thanks!

ptrus (approved):

PR #3757, opened 2021-03-05
go/nodes/grpc: tweak backoff timeouts

ptrus:
Tweak some grpc timeouts. Before, the default storage commit timeout was less than `grpcBackoffMaxDelay` which made the storage commit request retries ineffective whenever the max delay was reached (e.g. first few requests after a storage node was restarted).

Discovered in longterm tests.

### DIFF
--- a/go/storage/client/client.go
+++ b/go/storage/client/client.go
@@ -110,6 +110,11 @@ func (b *storageClientBackend) writeWithClient(
 			op := func() error {
 				var rerr error
 				resp, rerr = fn(ctx, api.NewStorageClient(conn.ClientConn), conn.Node)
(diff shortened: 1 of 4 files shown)


go/storage/client/client.go
```diff
@@ -110,6 +110,12 @@ func (b *storageClientBackend) writeWithClient(
 			op := func() error {
 				var rerr error
 				resp, rerr = fn(ctx, api.NewStorageClient(conn.ClientConn), conn.Node)
+				b.logger.Debug(
+					"storage write request error",
```
kostko, line 114: Move this to the line above to be consistent with other places.

kostko (approved):

Issue #3645, opened 2021-01-21
Propose a new batch if there are message results


kostko:
The runtime executor (when designated as the tx scheduler) should always propose a new batch if there are message results that need to be processed even if there are no transactions. This makes sure the message results are processed in a timely manner.

PR #3771, opened 2021-03-10
go/worker/executor: Propose a new batch if there are message results

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3645

### DIFF
--- a/go/oasis-test-runner/scenario/e2e/runtime/runtime_message.go
+++ b/go/oasis-test-runner/scenario/e2e/runtime/runtime_message.go
@@ -1,0 +1,160 @@
+package runtime
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/env"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario"
+	"github.com/oasisprotocol/oasis-core/go/roothash/api/block"
+	"github.com/oasisprotocol/oasis-core/go/runtime/client/api"
+	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
+)
+
+var (
+	// RuntimeMessage is the runtime message scenario.
+	RuntimeMessage scenario.Scenario = newRuntimeMessage()
+
+	waitTimeout = 10 * time.Second
+)
+
+type runtimeMessageImpl struct {
+	runtimeImpl
+}
+
+func newRuntimeMessage() scenario.Scenario {
+	return &runtimeMessageImpl{
+		runtimeImpl: *newRuntimeImpl(
+			"runtime-message",
+			"",
+			nil,
+		),
+	}
+}
+
+func (sc *runtimeMessageImpl) Clone() scenario.Scenario {
+	return &runtimeMessageImpl{
+		runtimeImpl: *sc.runtimeImpl.Clone().(*runtimeImpl),
+	}
+}
+
+func (sc *runtimeMessageImpl) Fixture() (*oasis.NetworkFixture, error) {
+	f, err := sc.runtimeImpl.Fixture()
+	if err != nil {
+		return nil, err
+	}
+	// Use mock epoch to ensure no rounds due to epoch transition. This way we
+	// test batch proposals when there are no transactions but message results.
+	f.Network.SetMockEpoch()
+	return f, nil
+}
+
+func (sc *runtimeMessageImpl) Run(childEnv *env.Env) error {
+	if err := sc.Net.Start(); err != nil {
+		return err
+	}
+
+	fixture, err := sc.Fixture()
+	if err != nil {
+		return err
+	}
+
+	if err = sc.initialEpochTransitions(fixture); err != nil {
+		return err
+	}
+
+	ctx := context.Background()
+	c := sc.Net.ClientController().RuntimeClient
+
+	// We should be at round 2 after the epoch transitions.
+	waitCtx, cancel := context.WithTimeout(ctx, waitTimeout)
+	defer cancel()
+	if err = c.WaitBlockIndexed(waitCtx, &api.WaitBlockIndexedRequest{RuntimeID: runtimeID, Round: 2}); err != nil {
+		return err
+	}
+
+	// Save latest round.
+	sc.Logger.Debug("querying latest round")
+	round, err := c.GetBlock(ctx, &api.GetBlockRequest{RuntimeID: runtimeID, Round: api.RoundLatest})
+	if err != nil {
+		return err
+	}
+	latestRound := round.Header.Round
+	sc.Logger.Debug("latest runtime round", "round", latestRound)
+	if latestRound != 2 {
+		return fmt.Errorf("unexpected latest round, got: %d, expected: %d", latestRound, 2)
+	}
+
+	blkCh, sub, err := c.WatchBlocks(ctx, runtimeID)
+	if err != nil {
+		return err
+	}
+	defer sub.Close()
+
+	// Submit a consensus transfer transaction. This should result in two runtime
+	// rounds:
+	//   - in first round the consensus transfer transaction should be executed
+	//   - in the second round there should be no transactions, the round should
+	//     contain message results of the consensus transfer.
+	args := struct {
+		Transfer staking.Transfer `json:"transfer"`
+		Nonce    uint64           `json:"nonce"`
+	}{
+		Transfer: staking.Transfer{},
+		Nonce:    0,
+	}
+	sc.Logger.Debug("submitting consensus_transfer runtime transaction")
+	if _, err = sc.submitRuntimeTx(ctx, runtimeID, "consensus_transfer", args); err != nil {
+		return err
+	}
+
+	sc.Logger.Debug("watching runtime round transitions")
+	var reachedMsgRound bool
+	for {
+		select {
+		case blk := <-blkCh:
+			round := blk.Block.Header.Round
+			sc.Logger.Debug("round transition", "round", round, "header", blk.Block.Header)
+			switch {
+			case round <= latestRound:
+				// Skip old rounds.
+				continue
+			case round > latestRound+2:
+				// Only two rounds are expected.
+				return fmt.Errorf("unexpected runtime round: %d", round)
+			default:
+				if ht := blk.Block.Header.HeaderType; ht != block.Normal {
+					return fmt.Errorf("expected normal round, got: %d", ht)
+				}
+				txs, err := c.GetTxs(ctx, &api.GetTxsRequest{
+					RuntimeID: runtimeID,
+					Round:     round,
+					IORoot:    blk.Block.Header.IORoot,
+				})
+				if err != nil {
+					return err
+				}
+				switch round {
+				case latestRound + 1:
+					// Round with the submitted consensus_transfer transaction.
+					if len(txs) != 1 {
+						return fmt.Errorf("expected 1 transaction at round: %d, got: %d", round, len(txs))
+					}
+				case latestRound + 2:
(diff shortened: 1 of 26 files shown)


go/worker/compute/executor/committee/node.go
```diff
@@ -724,25 +724,49 @@ func (n *Node) proposeTimeoutLocked() error {
...
 		epoch := n.commonNode.Group.GetEpochSnapshot()
 
 		// If we are not an executor worker in this epoch, we don't need to do anything.
 		if !epoch.IsExecutorWorker() {
-			return nil, nil, errNotTxnScheduler
+			return nil, nil, nil, errNotTxnScheduler
 		}
-		return epoch, &header, nil
+
+		// Fetch message results emitted during the last normal round.
```
kostko, line 746: This should probably be factored out so it isn't duplicated in `startProcessingBatchLocked`.

ptrus: done

go/worker/compute/executor/committee/node.go
```diff
@@ -724,25 +724,49 @@ func (n *Node) proposeTimeoutLocked() error {
...
+		if err != nil {
+			n.logger.Error("failed to query round last normal round results",
+				"err", err,
+				"height", height,
+				"round", header.Round,
+			)
+			return nil, nil, nil, err
+		}
+
+		// TODO: propagate roundResults into the processing batch step, to avoid
```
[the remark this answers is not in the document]
ptrus: done

go/oasis-test-runner/scenario/e2e/runtime/runtime_message.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/env"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario"
+	"github.com/oasisprotocol/oasis-core/go/roothash/api/block"
+	"github.com/oasisprotocol/oasis-core/go/runtime/client/api"
+	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
+)
+
+var (
+	// RuntimeMessage is the runtime message.
```
kostko, line 17: He he :-)

go/worker/compute/executor/committee/node.go
```diff
@@ -723,26 +727,52 @@ func (n *Node) proposeTimeoutLocked() error {
 	return nil
 }
 
+func (n *Node) roundResults(height int64) (*roothash.RoundResults, error) {
```
kostko, line 730:
```suggestion
func (n *Node) getRoundResults(height int64) (*roothash.RoundResults, error) {
```

go/worker/compute/executor/committee/node.go
```diff
@@ -982,27 +1017,26 @@ func (n *Node) startProcessingBatchLocked(batch *unresolvedBatch) {
 	blk := n.commonNode.CurrentBlock
 	consensusBlk := n.commonNode.CurrentConsensusBlock
 	height := n.commonNode.CurrentBlockHeight
+	roundResults := lastRoundResults
 	go func() {
 		defer close(done)
 
-		// Fetch message results emitted during the last normal round.
-		state, err := n.commonNode.Consensus.RootHash().GetRuntimeState(ctx, blk.Header.Namespace, height)
+		rtDesc, err := n.commonNode.Runtime.RegistryDescriptor(n.ctx)
```
kostko, line 1024: Hm, so in case of changes, roothash has a snapshot of the descriptor that is valid for the current round. Could also fetch this at the epoch transition and cache it?

ptrus: yeah actually i see that we also already watch descriptor updates in the worker loop (in case we need to update scheduler parameters), can probably get it there as well?

go/oasis-test-runner/scenario/e2e/runtime/runtime_message.go
ptrus: :grimacing: updated

go/worker/compute/executor/committee/node.go
ptrus: Actually if we cached the runtime descriptor on epoch change we'd need to access the cached runtime only while having the CrossNode lock (or use a new lock for the descriptor - but this is what the common node runtime registry already does). Therefore i think it makes more sense to just query the common node runtime registry as we do now?  (note: in this specific case we already have the CrossNode lock before the goroutine, so it's not the best point, but there are other places in the executor where we query the runtime descriptor and we don't have the lock currently and would need it if we switched to caching the descriptor on epoch transitions).

Or is there any other benefit doing it the other way?

kostko: So my main concern is that the registry descriptor may be different than what is actually in effect for the current epoch. This is because on each epoch transition the roothash service makes a snapshot of the runtime descriptor that is valid in that epoch. If the descriptor is then updated mid-epoch it may contain different information.

kostko: Actually it would probably be better to update the `RegistryDescriptor` and related methods to return a snapshot at the start of an epoch?

ptrus: Ah i see. Yeah I was about to mention that in that case this is likely already problematic in some existing places where runtime descriptor is queried. I think updating the `RegistryDescriptor` is better. Need to just double-check if all existing places that use it will work as expected with the change.

ptrus: Yeah skimming over all places where we use either `RegistryDescriptor` or `WatchRegistryDescriptor` the correct (or intended) are he "currently active parameters" so querying the roothash is more appropriate. Will update.

ptrus: Actually there are few places that actually want the descriptor in registry (e.g. keymanager,storage,common committee when initializing - before the runtime is even initialized). So I added new methods to the runtime registry for fetching active descriptor (`ActiveDescriptor` `WatchActiveDescriptor`) and also kept the old ones.

go/runtime/registry/registry.go
```diff
@@ -221,33 +255,67 @@ func (r *runtime) stop() {
 	r.history.Close()
 }
 
-func (r *runtime) watchUpdates(ctx context.Context, ch <-chan *registry.Runtime, sub pubsub.ClosableSubscription) {
+func (r *runtime) watchUpdates(
+	ctx context.Context,
+	ch <-chan *roothash.AnnotatedBlock,
```
kostko, line 260: Instead of watching all blocks, this could also just watch epoch transitions as those are less frequent so there would be less wakeups.

go/worker/compute/executor/committee/node.go
```diff
@@ -982,27 +1017,26 @@ func (n *Node) startProcessingBatchLocked(batch *unresolvedBatch) {
 	blk := n.commonNode.CurrentBlock
 	consensusBlk := n.commonNode.CurrentConsensusBlock
 	height := n.commonNode.CurrentBlockHeight
+	roundResults := lastRoundResults
 	go func() {
 		defer close(done)
 
-		// Fetch message results emitted during the last normal round.
-		state, err := n.commonNode.Consensus.RootHash().GetRuntimeState(ctx, blk.Header.Namespace, height)
+		rtDesc, err := n.commonNode.Runtime.ActiveDescriptor(n.ctx)
```
kostko, line 1024: Hm actually I think even doing it like this is not enough as this doesn't guarantee that what is queried here is synced with `height`. Would be better to just do a `GetRuntimeState` query here as before.

go/runtime/registry/registry.go
```diff
@@ -411,22 +495,27 @@ func newRuntime(
...
-	// Start watching this runtime's descriptor.
-	ch, sub, err := consensus.Registry().WatchRuntimes(ctx)
+	ch, sub, err := consensus.Beacon().WatchEpochs(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("runtime/registry: failed to watch epochs %s: %w", id, err)
+	}
+	regCh, regSub, err := consensus.Registry().WatchRuntimes(ctx)
 	if err != nil {
-		return nil, fmt.Errorf("runtime/registry: failed to watch updates for runtime %s: %w", id, err)
+		return nil, fmt.Errorf("runtime/registry: failed to watch roothash blocks %s: %w", id, err)
```
kostko, line 504: This line probably needs to be reverted.

go/runtime/transaction/batch.go
```diff
@@ -2,8 +2,3 @@ package transaction
 
 // RawBatch is a list of opaque bytes.
 type RawBatch [][]byte
-
-// String returns a string representation of a batch.
-func (b RawBatch) String() string {
```
kostko, line 7: This could result in a lot of verbosity in logs.

go/worker/compute/executor/committee/node.go
```diff
@@ -753,14 +784,19 @@ func (n *Node) handleScheduleBatch(force bool) {
 
 	// Ask the scheduler to get us a scheduled batch.
 	batch := n.scheduler.GetBatch(force)
-	if len(batch) == 0 {
+	n.logger.Debug("got a batch",
+		"batch", batch,
```
kostko, line 788: This seems verbose.

go/runtime/transaction/batch.go
ptrus: It's not that bad, the transactions are displayed base64 encoded. The amount of logs is comparable to the amount of `received peer message` logs (where we log the message contents as well). Helps a lot with debugging though.

go/worker/compute/executor/committee/node.go
ptrus: Yeah this one is not as useful and very verbose. Removing it.

go/runtime/transaction/batch.go
kostko: Ok.

kostko (commented):
Hm, node tests seem to hang in `WorkerImplementationTests`, specifically in:
https://github.com/oasisprotocol/oasis-core/blob/20b756ab576233b30a35a19936034f0c8865a0ca/go/worker/compute/executor/tests/tester.go#L39-L40

ptrus:
> Hm, node tests seem to hang in WorkerImplementationTests, specifically in:

It was due to `syncCheckpoints` blocking on waiting for `ActiveDescriptor` causing the storage worker never to initialize. Added a timeout and a fallback to `RegistryDescriptor` as i think we'd still want to try syncing from checkpoints even if runtime is inactive (there might be registered storage nodes running). I also wen't over other places where `ActiveDescriptor` is used and made sure appropriate context is used (e.g. blocking for the duration of used context is fine/desired).

go/runtime/registry/registry.go
```diff
@@ -77,6 +79,13 @@ type Runtime interface {
 	// WatchRegistryDescriptor subscribes to registry descriptor updates.
 	WatchRegistryDescriptor() (<-chan *registry.Runtime, pubsub.ClosableSubscription, error)
 
+	// ActiveDescriptor waits for runtime to be initialized and then returns
```
kostko, line 82: Maybe add a comment here (and below) saying that this may block until the next epoch transition.

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -378,9 +378,19 @@ func (n *Node) syncCheckpoints() (*blockSummary, error) {
 	// for errors, driven by remainingRoots.
 	var syncState blockSummary
 
-	descriptor, err := n.commonNode.Runtime.RegistryDescriptor(n.ctx)
+	// Try getting the active descriptor first.
+	adCtx, cancel := context.WithTimeout(n.ctx, 2*time.Second)
```
kostko, line 382: Maybe make the timeout a constant defined at the top.

kostko (approved):

PR #3773, opened 2021-03-11
go/consensus: Move upgrade handling back to mux in BeginBlock/EndBlock

kostko:


go/upgrade/upgrade.go
```diff
@@ -58,13 +59,29 @@ func (u *upgradeManager) SubmitDescriptor(ctx context.Context, descriptor *api.D
...
 func (u *upgradeManager) PendingUpgrades(ctx context.Context) ([]*api.PendingUpgrade, error) {
 	u.Lock()
 	defer u.Unlock()
 
-	return u.pending, nil
+	return append([]*api.PendingUpgrade{}, u.pending...), nil
 }
 
+// Implements api.Backend.
+func (u *upgradeManager) HasPendingUpgradeAt(ctx context.Context, height int64) (bool, error) {
```
ptrus, line 71: Maybe add a note (or even return an error), that this should not be called with `consensus.HeightLatest` as the `pu.UpgradeHeight` will be `0` (=`upgade.InvalidUpgradeHeight`) in case the upgrade has not yet started.

ptrus (approved):

go/upgrade/upgrade.go
kostko: Nice catch, fixed, please take another look.

ptrus: looks good. thx

PR #3782, opened 2021-03-15
go/consensus: Store runtime state roots in a single key

kostko:
That makes it easier to construct proofs starting at the consensus state root
that prove what the state root of a specific runtime is. You can then chain
these proofs to prove something about runtime state.

### DIFF
--- a/go/consensus/tendermint/apps/roothash/state/state.go
+++ b/go/consensus/tendermint/apps/roothash/state/state.go
@@ -34,6 +34,14 @@ var (
 	//
 	// Key format is: 0x24 <H(runtime-id) (hash.Hash)> <round (uint64)> <evidence-hash (hash.Hash)>
 	evidenceKeyFmt = keyformat.New(0x24, keyformat.H(&common.Namespace{}), uint64(0), &hash.Hash{})
+	// stateRootKeyFmt is the key format used for runtime state roots.
+	//
+	// Value is the runtime's latest state root.
+	stateRootKeyFmt = keyformat.New(0x25, keyformat.H(&common.Namespace{}))
+	// ioRootKeyFmt is the key format used for runtime I/O roots.
+	//
+	// Value is the runtime's latest I/O root.
+	ioRootKeyFmt = keyformat.New(0x26, keyformat.H(&common.Namespace{}))
 
 	cborTrue = cbor.Marshal(true)
 )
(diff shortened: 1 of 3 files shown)


go/consensus/tendermint/apps/roothash/state/state.go
```diff
@@ -193,7 +197,16 @@ func NewMutableState(tree mkvs.KeyValueTree) *MutableState {
 
 // SetRuntimeState sets a runtime's roothash state.
 func (s *MutableState) SetRuntimeState(ctx context.Context, state *roothash.RuntimeState) error {
-	err := s.ms.Insert(ctx, runtimeKeyFmt.Encode(&state.Runtime.ID), cbor.Marshal(state))
+	if err := s.ms.Insert(ctx, runtimeKeyFmt.Encode(&state.Runtime.ID), cbor.Marshal(state)); err != nil {
+		return api.UnavailableStateError(err)
+	}
+
+	// Store just the current state root separately to make it easier to retrieve.
+	stateRoot, err := state.CurrentBlock.Header.StateRoot.MarshalBinary()
```
ptrus, line 205: Although not needed at the moment, maybe include a getter and a test for retrieving the state root.

kostko: Done.

go/consensus/tendermint/apps/roothash/state/state.go
```diff
@@ -193,8 +227,22 @@ func NewMutableState(tree mkvs.KeyValueTree) *MutableState {
 
 // SetRuntimeState sets a runtime's roothash state.
 func (s *MutableState) SetRuntimeState(ctx context.Context, state *roothash.RuntimeState) error {
-	err := s.ms.Insert(ctx, runtimeKeyFmt.Encode(&state.Runtime.ID), cbor.Marshal(state))
-	return api.UnavailableStateError(err)
+	if err := s.ms.Insert(ctx, runtimeKeyFmt.Encode(&state.Runtime.ID), cbor.Marshal(state)); err != nil {
+		return api.UnavailableStateError(err)
+	}
+
+	// Store just the current state and I/O roots separately to make them easier to retrieve when
```
ptrus, line 234:
```suggestion
	// Store the current state and I/O roots separately to make them easier to retrieve when
```

ptrus (approved):

Issue #3783, opened 2021-03-16
Add node status to GetStatus response


kostko:
We should add the current node status to the `control.GetStatus` endpoint response (probably into the `RegistrationStatus` structure). This allows one to quickly see whether the node is currently frozen.

PR #3785, opened 2021-03-16
go/control/api: Add node status to registration status response

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3783


go/worker/registration/worker.go
```diff
@@ -527,10 +527,20 @@ func (w *Worker) registrationStopped() {
...
 	*status = w.status
+	w.RUnlock()
+
+	if status == nil || status.Descriptor == nil {
+		return status, nil
+	}
+
+	ns, err := w.registry.GetNodeStatus(ctx, &registry.IDQuery{ID: status.Descriptor.ID, Height: consensus.HeightLatest})
+	if err != nil {
+		return nil, err
```
kostko, line 540: I guess this really shouldn't fail if registration succeeded.

kostko (approved):

PR #3790, opened 2021-03-17
go/roothash: Finish commitment processing early when possible

kostko:
When the set of current commitments already determines the fate of the process
(e.g., when there is a discrepancy or a majority of votes indicate success),
proceed with the process instead of waiting for the remaining commitments.

### DIFF
--- a/.changelog/3790.breaking.md
+++ b/.changelog/3790.breaking.md
@@ -1,0 +1,6 @@
+go/roothash: Finish commitment processing early when possible
+
+When the set of current commitments already determines the fate of the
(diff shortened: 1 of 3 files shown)


ptrus (commented):
Nice! The `ProcessCommitments` code looks correct. Be sure to add tests to pool_test for the new early processing cases (i know you already plan to do it :grimacing: )

kostko:
> Be sure to add tests to pool_test for the new early processing cases (i know you already plan to do it :grimacing: ) 

Yeah, test case is in now. Please take one more look.

ptrus (approved):
Looks good :+1:

PR #3806, opened 2021-03-22
oasis-net-runner: Fix runtime configuration in default fixture

### LINKED ISSUE #3794: oasis-net-runner: run a client node in --fixture.default.setup_runtimes=false
##### SUMMARY

Using `oasis-net-runner --fixture.default.setup_runtimes=false ...` starts a network with only one validator, as it says on the tin. Would it be good to start a non-validator node as well, similar to the client node in the yes-runimes case? You could then use this node to handle queries and stuff, the way you would normally prefer to use a non-validator node for that.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

go/oasis-net-runner

##### ADDITIONAL INFORMATION

for handling RPC interaction

### LINKED ISSUE #3805: oasis-net-runner: Default fixture doesn't configure any runtimes
#3719 introduced the `Runtimes` field for client fixtures so that each client node can be configured with only a subset of runtimes. The problem is that the default `oasis-net-runner` fixture was not updated to configure any runtimes so now no runtimes are configured.

kostko:
Fixes #3805 
Fixes #3794

### DIFF
--- a/.changelog/3794.feature.md
+++ b/.changelog/3794.feature.md
@@ -1,0 +1,1 @@
+oasis-net-runner: Configure a client node even without runtimes
--- a/.changelog/3805.bugfix.md
+++ b/.changelog/3805.bugfix.md
@@ -1,0 +1,1 @@
+oasis-net-runner: Configure a runtime in default fixture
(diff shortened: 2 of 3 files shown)


ptrus (approved):
Hm lets maybe open an issue for adding a simple e2e test using the default fixture

kostko:
Good idea, opened #3807.

Yawning (approved):

Issue #1815, opened 2019-06-13
storage/client: Use a more clever approach to choose the order in which to read from connected nodes


tjanez:
After #1792 is merged, the storage client will try to read from the connected nodes in random order:

```
... trimmed ...

rng := rand.New(mathrand.New(cryptorand.Reader))

var err error
var resp interface{}
for _, randIndex := range rng.Perm(n) {
	clientState := b.connectedNodesState.clientStates[randIndex]
	resp, err = fn(ctx, clientState.client)

.. trimmed ...
```

Consider using a more clever approach to select the order in which to read from the connected storage nodes.

One possible way would be to track per-node statistics and have a policy to prioritize based on that. This should be done in a manner that is reusable across many different protocols, e.g., instead of just storage gRPC requests it should also be informed by messages received via P2P etc. So a common "peer reputation tracking" mechanism.

kostko:
Obsoleted by #4459 which does peer scoring.

PR #3820, opened 2021-03-29
Fix storage worker

jberci:
Avoid database corruption in certain cases.

- [X] Root existence checking in Finalize
- [X] Detect Finalize failures in the worker and don't update syncing metadata
- [x] Handle Apply failures during sync (+byzantine test for it)
- [x] Sync heartbeat independent of incoming blocks

### DIFF
--- a/go/oasis-test-runner/scenario/e2e/runtime/byzantine.go
+++ b/go/oasis-test-runner/scenario/e2e/runtime/byzantine.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"time"
 
 	"github.com/oasisprotocol/oasis-core/go/common/quantity"
 	consensus "github.com/oasisprotocol/oasis-core/go/consensus/api"
@@ -14,7 +15,9 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario"
 	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario/e2e"
 	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
+	runtimeClient "github.com/oasisprotocol/oasis-core/go/runtime/client/api"
 	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs"
 )
 
 var (
@@ -213,6 +216,20 @@ var (
 			"--" + byzantine.CfgFailReadRequests,
 		},
 	)
+	// ByzantineStorageCorruptGetDiff is the byzantine storage node scenario that corrupts GetDiff
+	// responses.
+	ByzantineStorageCorruptGetDiff scenario.Scenario = newByzantineImpl(
+		"storage-corrupt-getdiff",
+		"storage",
+		// There should be no discrepancy or round failures.
+		nil,
+		oasis.ByzantineDefaultIdentitySeed,
+		nil,
+		[]string{
+			// Corrupt all GetDiff responses.
+			"--" + byzantine.CfgCorruptGetDiff,
+		},
+	)
 )
 
 type byzantineImpl struct {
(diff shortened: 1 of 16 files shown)


go/storage/api/context.go
```diff
@@ -45,3 +52,47 @@ func NodePriorityHintFromContext(ctx context.Context) []signature.PublicKey {
...
+
+// WithNodeBlacklist sets a storage blacklist key for any storage requests using this context.
+// The blacklist is initially empty (i.e. all nodes are acceptable).
+func WithNodeBlacklist(ctx context.Context) context.Context {
+	return context.WithValue(ctx, contextKeyNodeBlacklist, map[signature.PublicKey]struct{}{})
+}
+
+// BlacklistAddNode adds a node to the blacklist associated with the context. If there's no
+// associated blacklist, the function does nothing.
+func BlacklistAddNode(ctx context.Context, node *node.Node) {
```
kostko, line 64: Instead of doing it like this, can we make a blacklist interface that has a method for checking whether something is blacklisted? And then you can configure an instance using `WithNodeBlacklist`?

go/storage/api/context.go
```diff
@@ -45,3 +52,47 @@ func NodePriorityHintFromContext(ctx context.Context) []signature.PublicKey {
...
+}
+
+// NodeSelectionCallback is a callback used by the storage client to report connections used
+// for read requests.
+type NodeSelectionCallback = func(*node.Node)
+
+// WithNodeSelectionCallback sets a callback that will be called by the storage client on every read
+// request with the node that was eventually used to perform a successful request. If there was no
+// success, the callback isn't called.
+func WithNodeSelectionCallback(ctx context.Context, cb NodeSelectionCallback) context.Context {
```
kostko, line 89: Could this be an interface instead? Maybe shared with the blacklist interface (maybe the interface needs a different name like `NodeManager` and the manager together with context accessors could be defined in `go/runtime/nodes` to be reusable in other contexts as well)?

go/worker/storage/committee/node.go
```diff
@@ -1122,24 +1158,28 @@ mainLoop:
...
-			// Check if we're far enough to reasonably register as available.
-			n.nudgeAvailability(cachedLastRound, latestBlockRound)
-
-			// Notify the checkpointer that there is a new finalized round.
-			if n.checkpointer != nil {
-				n.checkpointer.NotifyNewVersion(finalized.Round)
+			// If finalization failed, things start falling apart.
+			// There's no point redoing it, since it's probably not a transient
+			// error, and cachedLastRound also can't be updated legitimately.
+			if finalized.err == nil {
```
kostko, line 1170: Maybe invert the match and `break` instead?

go/worker/storage/committee/node.go
```diff
@@ -1027,6 +1033,12 @@ mainLoop:
 						"old_root", lastDiff.prevRoot,
 						"new_root", lastDiff.thisRoot,
 					)
+					if errors.Is(err, storageApi.ErrExpectedRootMismatch) && lastDiff.srcNode != nil {
+						storageApi.BlacklistAddNode(n.ctx, lastDiff.srcNode)
+						n.logger.Info("node blacklisted due to bogus diff",
```
kostko, line 1038: Should be a warning.

kostko (approved):

go/oasis-test-runner/scenario/e2e/runtime/byzantine.go
```diff
@@ -207,12 +221,33 @@ var (
 		// There should be no discrepancy or round failures.
 		nil,
 		oasis.ByzantineDefaultIdentitySeed,
+		// Hack to work around the way the storage client selects nodes.
+		// It can happen that for this test, the client will keep selecting the byzantine
+		// node instead of the other storage nodes and so would never be able to fully sync.
+		// It can take up to two minutes for storage-0 to sync in this scenario with the
+		// current shuffling method in the storage client. See also #1815.
```
ptrus, line 228: Maybe also mention that this hack can be removed when #1815 is done

ptrus (approved):

Issue #3564, opened 2020-12-11
Basic consensus layer state accessors for Rust


kostko:
As of #3511 the runtimes are able to access consensus layer state. The missing part are the data structures required to access state from Rust, specifically:
* `consensus.LightBlock`
* `staking.state.ImmutableState` (partial)

This also requires the consensus layer version to be checked on runtime initialization and rejected in case it is incompatible. Later on we can consider adding additional Rust state representations as required.

PR #3823, opened 2021-03-30
Consensus layer state access from runtimes

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3564

TODO:
- [x] implement more of the staking state wrapper
- [x] add serialization tests for ported types
- [x] add more e2e tests (probably separate PR)
  - [x] rust state wrappers interop tests - https://github.com/oasisprotocol/oasis-core/pull/3833
  - [x] longtests workload - https://github.com/oasisprotocol/oasis-core/issues/3864


runtime/src/common/key_format.rs
```diff
@@ -74,6 +76,142 @@ pub trait KeyFormat {
...
+        )
+    }
+}
+
+/// Define a KeyFormat from KeyFromatAtom and a prefix.
+///
+/// # Examples
+///
+/// ```rust,ignore
+/// ker_format!(NewKeyFormatName, InnerType, 0x01);
```
kostko, line 187:
```suggestion
/// key_format!(NewKeyFormatName, InnerType, 0x01);
```

Also could we switch the order of the prefix and inner type (as this is how it is actually formatted)?

runtime/src/common/key_format.rs
```diff
@@ -74,6 +76,142 @@ pub trait KeyFormat {
...
+        }
+
+        (
+            T1::decode_atom(&data[0..T1::size()]),
+            T2::decode_atom(&data[T2::size()..]),
+        )
+    }
+}
+
+// TODO: probably could generalize this for N sized tuples using macros.
```
kostko, line 149: Would [`impl-trait-for-tuples`](https://crates.io/crates/impl-trait-for-tuples) help here?

runtime/src/consensus/state/staking.rs
```diff
@@ -0,0 +1,72 @@
...
+key_format!(CommonPoolKeyFmt, (), 0x52);
+key_format!(DelegationKeyFmt, (Address, Address), 0x53);
+key_format!(DebondingDelegationKeyFmt, (Address, Address, u64), 0x54);
+key_format!(DebondingQueueKeyFmt, (u64, Address, Address), 0x55);
+key_format!(ParametersKeyFmt, (), 0x56);
+key_format!(LastBlockFees, (), 0x57);
+key_format!(EpochSigningKeyFmt, (), 0x58);
+key_format!(GovernanceDepositsKeyFmt, (), 0x59);
+
+// TODO: implement the rest. Add some tests?
```
kostko, line 35: Would be cool to have interoperability tests for all the methods. But this can be part of a separate issue/PR.

runtime/src/consensus/state/staking.rs
```diff
@@ -0,0 +1,72 @@
+use std::collections::BTreeMap;
+
+use anyhow::{anyhow, Result};
```
kostko, line 3: Can we use `thiserror` instead and define the errors?

runtime/src/consensus/state/staking.rs
```diff
@@ -0,0 +1,72 @@
...
+key_format!(DebondingDelegationKeyFmt, (Address, Address, u64), 0x54);
+key_format!(DebondingQueueKeyFmt, (u64, Address, Address), 0x55);
+key_format!(ParametersKeyFmt, (), 0x56);
+key_format!(LastBlockFees, (), 0x57);
+key_format!(EpochSigningKeyFmt, (), 0x58);
+key_format!(GovernanceDepositsKeyFmt, (), 0x59);
+
+// TODO: implement the rest. Add some tests?
+impl<'a, T: ImmutableMKVS> ImmutableState<'a, T> {
+    pub fn account(self, ctx: Context, address: Address) -> Result<Account> {
```
kostko, line 37: Why do you consume `self` in these methods?

kostko (commented):
Make sure to also add a check to `protocol.rs` (where `RuntimeInfoRequest` is handled) to make sure that the Tendermint consensus backend is in use.

runtime/src/consensus/state/staking.rs
ptrus: Yeah shouldn't. Will update

runtime/src/consensus/state/staking.rs
ptrus: ok

runtime/src/common/key_format.rs
ptrus: Yeah probably need to look into it. Staking module needed just up to 3 so i was fine implementing those manually for now :grin:

runtime/src/consensus/state/staking.rs
kostko: > Agreed. This should utilize the existing protocol server right? Need to look a bit into it, but probably makes sense to do in a separate PR. However I wouldn't trust using these methods until we do have such tests grin

You mean the one used for storage interop tests? Yeah that could work but you would first need to import/load a database that has some consensus state in it.

> Additionally to that in the long run we should probably add a txsource workload that does queries via a runtime and directly - and compares the results.

Good idea, feel free to file issues.

runtime/src/consensus/state/staking.rs
ptrus: fixed

runtime/src/common/key_format.rs
ptrus: fixed

runtime/src/protocol.rs
```diff
@@ -286,10 +289,16 @@ impl Protocol {
...
-                    "consensus_protocol_version" => consensus_protocol_version,
+                    "consensus_protocol_version" => ?consensus_protocol_version,
                     "consensus_chain_context" => &consensus_chain_context,
                 );
-                // TODO: Verify consensus backend/protocol compatibility.
+
+                if BACKEND_NAME != &consensus_backend {
+                    return Err(ProtocolError::IncompatibleConsensusBackend.into());
+                }
+                if !consensus_protocol_version.compatible(&consensus_protocol_version) {
```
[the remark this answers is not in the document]
ptrus: :see_no_evil:

runtime/src/consensus/state/staking.rs
```diff
@@ -0,0 +1,209 @@
...
+        let mut result = Vec::new();
+        loop {
+            match it.next() {
+                Some((key, _)) => match AccountsKeyFmt::decode(&key) {
+                    Some(AccountsKeyFmt(address)) => result.push(address),
+                    None => break,
+                },
+                None => break,
+            }
+        }
```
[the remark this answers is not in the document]
ptrus: Nice i like the `map_while` one, changed. 
For `delegations`/`debonidng_delegations`  the semi-imperative version does seems nicer for building the result - but I did get rid of `loop` in favor of `while let`

runtime/src/consensus/state/staking.rs
ptrus: > You mean the one used for storage interop tests? Yeah that could work but you would first need to import/load a database that has some consensus state in it.

Yeah i have an outline of this working now. Added a flag to the interop protocol server to initialize some debug consensus state on the go side. Depending on when this PR get's approved i might include the tests in this PR, otherwise i'll open an isuse and a follow up PR.

runtime/src/protocol.rs
```diff
@@ -17,6 +17,7 @@ use thiserror::Error;
 
 use crate::{
     common::{cbor, logger::get_logger, namespace::Namespace, version::Version},
+    consensus::tendermint::BACKEND_NAME,
```
kostko, line 20: Maybe only import `tendermint` and refer to it as `tendermint::BACKEND_NAME` below. This makes it more clear which backend this is.

runtime/src/storage/mkvs/mod.rs
```diff
@@ -194,6 +194,35 @@ pub trait FallibleMKVS {
     fn commit(&mut self, ctx: Context, namespace: Namespace, version: u64) -> Result<Hash>;
 }
 
+// Immutable merkalized key value store.
```
kostko, line 197:
```suggestion
/// Immutable merkalized key value store.
```

go/runtime/host/protocol/types.go
```diff
@@ -278,6 +279,10 @@ type RuntimeKeyManagerPolicyUpdateRequest struct {
 
 // RuntimeQueryRequest is a runtime query request message body.
 type RuntimeQueryRequest struct {
+	// ConsensusBlock is the consensus light block at the last finalized round
+	// height (e.g., corresponding to .Block.Header.Round).
```
kostko, line 283:
```suggestion
	// height (e.g., corresponding to .Header.Round).
```

runtime/src/consensus/state/staking.rs
```diff
@@ -0,0 +1,198 @@
...
+        state::StateError,
+    },
+    key_format,
+    storage::mkvs::ImmutableMKVS,
+};
+
+/// Consensus staking state wrapper.
+pub struct ImmutableState<'a, T: ImmutableMKVS> {
+    pub mkvs: &'a T,
+}
```
kostko, line 25: Maybe add a constructor so you don't need to initialize the struct each time?

ptrus: :+1:

kostko (approved):

PR #3828, opened 2021-04-01
common/crypto/ed25519: Use curve25519-voi

Yawning:
* [x] Switch the import
 * [x] Use the new batch verification API
 * [x] Use expanded form public keys, under the assumption that the set of public keys is relatively static.
 * [x] Add a SLIP-0010 implementation (moved from the voi repo) 

~~This should eventually carve out the cache into somewhere else (maybe as part of voi, it seems generally useful), so that the tendermint side can also use it.  But what is there is ok as an illustration of how this will work.~~

### DIFF
--- a/go/common/crypto/signature/signature.go
+++ b/go/common/crypto/signature/signature.go
@@ -17,7 +17,8 @@ import (
 	"os"
 	"sync"
 
-	"github.com/oasisprotocol/ed25519"
+	"github.com/oasisprotocol/curve25519-voi/primitives/ed25519"
+	"github.com/oasisprotocol/curve25519-voi/primitives/ed25519/extra/cache"
 
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
@@ -66,7 +67,18 @@ var (
 	testPublicKeys        sync.Map
 	blacklistedPublicKeys sync.Map
 
-	defaultOptions = &ed25519.Options{}
+	defaultOptions = &ed25519.Options{
+		Verify: &ed25519.VerifyOptions{
+			AllowSmallOrderA:   false,
+			AllowSmallOrderR:   false,
+			AllowNonCanonicalA: true,
+			AllowNonCanonicalR: true,
+		},
+	}
+
+	cachingVerifier = cache.NewVerifier(
+		cache.NewLRUCache(4096), // Should be big enough?
+	)
 )
 
 // PublicKey is a public key used for signing.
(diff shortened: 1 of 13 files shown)


kostko (commented):
Looks good. Just making sure -- this is not consensus breaking and these verification parameters are compatible with what we had before?

Yawning:
> Looks good. Just making sure -- this is not consensus breaking and these verification parameters are compatible with what we had before?

Yes.

go/common/crypto/signature/signature.go
```diff
@@ -66,7 +67,18 @@ var (
...
+		Verify: &ed25519.VerifyOptions{
+			AllowSmallOrderA:   false,
+			AllowSmallOrderR:   false,
+			AllowNonCanonicalA: true,
+			AllowNonCanonicalR: true,
+		},
+	}
+
+	cachingVerifier = cache.NewVerifier(
+		cache.NewLRUCache(4096), // Should be big enough?
```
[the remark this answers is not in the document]
Yawning: Yes, but no one answered me when I asked what the maximum number of txns in a block can be.  This should be sized at "greater than or equal to the number of nodes in the system + maximum number of txes in a block".

Note: The tendermint integration will use it's own cache which can be more modest.

go/common/crypto/slip10/slip0010.go
```diff
@@ -0,0 +1,93 @@
...
+var curveConstant = []byte("ed25519 seed")
+
+// ChainCode is a SLIP-0010 chain code.
+type ChainCode [ChainCodeSize]byte
+
+// NewMasterKey derives a master key and chain code from a seed byte sequence.
+func NewMasterKey(seed []byte) (signature.Signer, ChainCode, error) {
+	// 1. Generate a seed byte sequence S of 512 bits according to BIP-0039.
+	if len(seed) != SeedSize {
+		return nil, ChainCode{}, fmt.Errorf("slip0010: invalid seed")
```
tjanez, line 32: Could you replace `slip0010` in this and the following 3 error messages with `slip10`?

go/common/crypto/slip10/slip0010.go
```diff
@@ -0,0 +1,93 @@
+// Package slip10 implements the SLIP-0010 private key derivation
```
tjanez, line 1: Nit: Rename `slip0010{,_test}.go` to `slip10{,_test}.go`

And also in the commit message (i.e. "common/crypto/slip10: ...").

go/common/crypto/slip10/slip0010.go
Yawning: I should have just left the package name as is.

go/common/crypto/signature/signature.go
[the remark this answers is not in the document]
kostko: The maximum number of txes in a block actually depends on consensus parameters and tx size. I guess this number makes sense for now.

kostko (approved):

PR #3832, opened 2021-04-02
storage: Drop the MKVS node version field

### LINKED ISSUE #3500: Consider dropping the MKVS node version field
Since the only MKVS backend uses Badger in managed mode to handle versioning, we should be able to drop the version field from the tree nodes. The only issue is that this is a breaking change to the tree format so it needs to be handled carefully in order for a migration to be possible.

kostko:
Fixes #3500


jberci (approved):

go/storage/mkvs/db/badger/migrate.go
```diff
@@ -665,15 +681,647 @@ func (v4 *v4Migrator) Migrate() (rversion uint64, rerr error) {
...
 
-	// All done, flush and clean up. The metadata blob will be removed
-	// in the defer handler.
+	// All done, flush and clean up.
 	v4.done = true
 	if err := v4.flush(true); err != nil {
 		return 0, err
 	}
 	return 4, nil
 }
```
ptrus, line 690: Can you please also update v4 migration to log something after its done migrating. iirc it doesn't explicitly log anything when everything is migrated.

go/storage/mkvs/db/badger/migrate.go
```diff
@@ -665,15 +681,647 @@ func (v4 *v4Migrator) Migrate() (rversion uint64, rerr error) {
...
+		newHash = n.GetHash()
+		newNode, _ = n.MarshalBinary()
+	default:
+		return nil, fmt.Errorf("unknown node type for node %s: 0x%02x", h, raw[0])
+	}
+
+	// Store node. We store the node as if it was created in the version we are migrating regardless
+	// of its original version. This is because we need to be purging any old versions.
+	entry := badger.NewEntry(v5NodeKeyFmt.Encode(&newHash), newNode)
+	err = v5.changeBatch.SetEntryAt(entry, item.Version())
```
ptrus, line 934: Shouldn't this set the entry at `version`  (according to the above comment).

kostko: No, that was the initial thought but the comment is now wrong since we actually also migrate all versions that come after the last finalized version.

go/storage/mkvs/db/badger/migrate.go
```diff
@@ -665,15 +681,647 @@ func (v4 *v4Migrator) Migrate() (rversion uint64, rerr error) {
...
+		return cbor.UnmarshalTrusted(data, &rootsMeta)
+	})
+	if err != nil {
+		return fmt.Errorf("error decoding roots metadata for version %d: %w", version, err)
+	}
+
+	maybeLoneRoots := make(map[typedHash]bool)
+	for rootHash, derivedRoots := range rootsMeta.Roots {
+		if len(derivedRoots) == 0 {
+			// Need to only set the flag iff the flag has not already been set
```
ptrus, line 1170: `rootsMeta.Roots` is `map[typedHash][]typedHash` right? 
So each key (`rootHash`) will only appear once so this check is unnecessary?

go/storage/mkvs/db/badger/migrate.go
kostko: Done.

go/storage/mkvs/db/badger/migrate.go
kostko: Hm that's a good point, so I copied this from the existing `Prune` implementation where it also doesn't seem to be necessary. Based on blame it seems that this originated from the time when there was no `Roots` map and this was inside a loop which iterated over Badger keys so uniqueness was not guaranteed. Will remove it in both places.

kostko: Fixed.

ptrus: Cool thanks!

ptrus (approved):

PR #3848, opened 2021-04-12
runtime: Support for block tags

ptrus:
Runtime: Add support for block tags (tags not emitted by a specific transaction).

Needed for: oasis-sdk#consensus-module (https://github.com/oasisprotocol/oasis-sdk/pull/92)

### DIFF
--- a/.changelog/3848.breaking.md
+++ b/.changelog/3848.breaking.md
@@ -1,0 +1,1 @@
+runtime: Add support for block tags
--- a/runtime/src/transaction/tree.rs
+++ b/runtime/src/transaction/tree.rs
@@ -74,6 +74,9 @@ struct TagKeyFormat {
     tx_hash: Hash,
 }
 
+/// Hash used for block emitted tags not tied to a specific transaction.
+pub const TAG_BLOCK_TX_HASH: Hash = Hash([0u8; 32]);
+
 impl KeyFormat for TagKeyFormat {
     fn prefix() -> u8 {
         'E' as u8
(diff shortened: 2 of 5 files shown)


runtime/src/transaction/tree.rs
```diff
@@ -74,6 +74,9 @@ struct TagKeyFormat {
     tx_hash: Hash,
 }
 
+/// Hash used for block emitted tags not tied to a specific transaction.
+pub const TAG_BLOCK_TX_HASH: Hash = Hash([0u8; 32]);
```
kostko, line 78: Don't forget to add this to the Go version as well.

.changelog/3848.breaking.md
```diff
@@ -0,0 +1 @@
+Runtime: Add support for block tags
```
kostko, line 1:
```suggestion
runtime: Add support for block tags
```

.changelog/3848.breaking.md
```diff
@@ -0,0 +1 @@
+runtime: Add support for block tags
```
kostko, line 1: Why is this breaking?

kostko (approved):

.changelog/3848.breaking.md
ptrus: Hm i was thinking to mark it as breaking as `ExecuteTxResult` (publicly exposed runtime api field) was changed. I guess it depends on what definition of "breaking" we'd like to use for these fragments :-)

kostko: Oh, yeah that's true, keep it as breaking.

PR #3830, opened 2021-04-01
Fix checkpoint abort handling

jberci:

PR #3863, opened 2021-04-15
Add test for storage sync after messy/incorrect cleanup

jberci:
Verified to fail if #3830 isn't applied.

### DIFF
--- a/go/oasis-test-runner/oasis/storage.go
+++ b/go/oasis-test-runner/oasis/storage.go
@@ -1,15 +1,18 @@
 package oasis
 
 import (
+	"context"
 	"crypto/ed25519"
 	"fmt"
 	"path/filepath"
 	"time"
 
+	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
 	"github.com/oasisprotocol/oasis-core/go/consensus/tendermint/crypto"
 	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
 	"github.com/oasisprotocol/oasis-core/go/storage/database"
+	workerStorage "github.com/oasisprotocol/oasis-core/go/worker/storage/api"
 )
 
 const storageIdentitySeedTemplate = "ekiden node storage %d"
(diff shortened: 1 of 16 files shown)


go/oasis-test-runner/oasis/storage.go
```diff
@@ -95,6 +98,24 @@ func (worker *Storage) DatabasePath() string {
 	return filepath.Join(worker.dir.String(), database.DefaultFileName(worker.backend))
 }
 
+// WaitForRoot waits until the node syncs up to the given root.
+func (worker *Storage) WaitForRound(ctx context.Context, runtimeID common.Namespace, round uint64) (uint64, error) {
+	ctrl, err := NewController(worker.SocketPath())
+	if err != nil {
+		return 0, err
+	}
+	client := workerStorage.NewStorageWorkerClient(ctrl.conn)
```
kostko, line 107: You could just add this to the controller.

go/worker/storage/committee/node.go
```diff
@@ -388,6 +394,29 @@ func (n *Node) getMetricLabels() prometheus.Labels {
...
 }
 
+func (n *Node) WaitForRound(round uint64, root *storageApi.Root) (<-chan uint64, error) {
+	retCh := make(chan uint64, 1)
+
+	if root != nil {
+		round = root.Version
+	}
+
+	n.syncedLock.RLock()
```
kostko, line 404: This probably needs to be upgraded to a write lock as you are modifying `roundWaiters` below.

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -174,16 +174,17 @@ func (n *Node) nodeWorker(
...
-		case errors.Is(result.err, checkpoint.ErrChunkCorrupted):
-			chunkReturnCh <- chunk
-			return result.err
-		case errors.Is(result.err, checkpoint.ErrChunkProofVerificationFailed):
-			errorCh <- checkpointStatusNext
-			return backoff.Permanent(result.err)
-		case result.err != nil:
-			errorCh <- checkpointStatusBail
-			return backoff.Permanent(result.err)
+			switch {
```
kostko, line 177: Should this be backported to `21.1.x`? It looks like a bugfix as the previous code would not skip a bad checkpoint?

kostko (approved):

go/worker/storage/committee/checkpoint_sync.go
jberci: The effect here is just to shorten the failure path. Without this fix, it would keep retrying up to `maxRetries` and then fail properly; with this, it stops immediately, as the original intention was.

PR #3891, opened 2021-04-27
Release process enhancements

tjanez:


docs/release-process.md
```diff
@@ -175,21 +165,21 @@ back-port some fixes (e.g. a security fix) and (backwards compatible) changes
...
+`stable/21.2.x`, and export it:
 
 ```bash
-RELEASE_BRANCH="stable/20.11.x"
+export RELEASE_BRANCH="stable/21.2.x"
 ```
 
 ### Back-port Changes
 
-Create a new branch, e.g. `<GITHUB-NAME>/${RELEASE_BRANCH}/backport-foo`, from
```
kostko, line 188: Hm I think we should still namespce the changes by user, similar as we do for regular branches.

tjanez: Yeah, this is a fair point. I was balancing the need for namespacing and having copy & paste Release Process docs.

My internal reasoning for not bothering with namespacing is that these branches (backport and changelog branches) should typically be short-lived and deleted automatically.

kostko (approved):

PR #3911, opened 2021-05-03
runtime: Add epoch to ExecuteTx/CheckTx/Query requests

### LINKED ISSUE #3892: Make current epoch number available in the runtime
The current epoch number should be made available to the runtime. The runtime could either fetch it from state or the host could include it as part of the batch execution request (this would have less overhead).

kostko:
Fixes #3892 

TODO
* [x] Tests.

### DIFF
--- a/go/runtime/client/client.go
+++ b/go/runtime/client/client.go
@@ -190,8 +190,12 @@ func (c *runtimeClient) CheckTx(ctx context.Context, request *api.CheckTxRequest
 	if err != nil {
 		return fmt.Errorf("client: failed to get light block at height %d: %w", rs.CurrentBlockHeight, err)
 	}
+	epoch, err := c.common.consensus.Beacon().GetEpoch(ctx, consensus.HeightLatest)
+	if err != nil {
+		return fmt.Errorf("client: failed to get current epoch: %w", err)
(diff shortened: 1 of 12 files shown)


go/runtime/client/client.go
```diff
@@ -190,8 +190,12 @@ func (c *runtimeClient) CheckTx(ctx context.Context, request *api.CheckTxRequest
 	if err != nil {
 		return fmt.Errorf("client: failed to get light block at height %d: %w", rs.CurrentBlockHeight, err)
 	}
+	epoch, err := c.common.consensus.Beacon().GetEpoch(ctx, consensus.HeightLatest)
```
ptrus, line 193: I think the above `LightBlock` query should also be made at  `consensus.HeightLatest` here.

ptrus (approved):

go/runtime/client/client.go
kostko: Hm, but during execution it happens at the last round height to ensure consistency. It would make sense for `CheckTx` to do the same?

ptrus: oh right, yeah it makes sense as it is then.

Issue #3862, opened 2021-04-15
Add support for obtaining consensus events from runtime


ptrus:
In order to make staking operations in runtimes easier to work with the runtime should be able see consensus events (such as `ReclaimEscrowEvent`) otherwise the runtime cannot really know when the submitted reclaim escrow actually occurred.

One possible solution is to include all (relevant to runtime?) consensus events since last round in `RuntimeExecute` requests.

PR #3940, opened 2021-05-12
Expose consensus events to runtimes

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3862 

Moves runtime block history indexer from the roothash service into the consensus service. Extends the indexer to also index runtime relevant events (currently only staking, should also include others). 

Roothash service:
- with the removal of the runtime block indexer from the service, the overhead of "tracking" all runtime events is probably not considerable, so in current version of the PR roothash tracks all events (instead of explicitly tracking and subscribing to "Watched/Tracked" runtime events). This change can be reverted.

Runtime event indexing:
- `CommitPendingConsensusEvents(height, events)` method is added to the runtime block indexer. For each consensus height the relevant events are indexed and stored in the history as pending events. On runtime block `Commit` pending events are moved and stored as the consensus events for the round. Pending events queue is cleared. 
- Executor node gets consensus events for the round by querying consensus events for all rounds since last successfully processed runtime round. (see: `executor/committee/node.go`)

TODO:
- [x] before, the runtime round blocks (via `roothash.WatchBlocks`) were emitted only after the round was indexed. Now this is not the case, so the history might not be up-to-date immediately after seeing the event. Should add a `WatchBlocks` method to the runtime history indexer?  
  - could remove the `roothash.WatchBlocks` method now (in favor of `History().WatchBlocks()`). The only issue is the keymanager, as it also needs to watch runtime blocks, but is not compatible with `RuntimeRegistry` (uses same file for localstorage, runtme registry currently assumes compute runtimes)  
- [ ] other relevant events (e.g. registry runtime events)
- [ ] runtime history unit tests
- [ ] e2e tests
- [ ] finish runtime side


go/roothash/api/commitment/executor.go
```diff
@@ -46,6 +46,9 @@ type ComputeResultsHeader struct {
 	IORoot       *hash.Hash `json:"io_root,omitempty"`
 	StateRoot    *hash.Hash `json:"state_root,omitempty"`
 	MessagesHash *hash.Hash `json:"messages_hash,omitempty"`
+	// TODO: consensus breaking change. Open a follow up issue to include this
+	// for a later release.
+	// EventsHash   *hash.Hash `json:"events_hash,omitempty"`
```
kostko, line 51: It would probably make sense to include this in `ProposedBatch` as well so that the workers can quickly check whether it matches with their local view before processing the batch.

go/runtime/history/api/api.go
```diff
@@ -0,0 +1,66 @@
...
+	GetBlock(ctx context.Context, round uint64) (*roothash.AnnotatedBlock, error)
+
+	// GetLatestBlock returns the block at latest round.
+	GetLatestBlock(ctx context.Context) (*roothash.AnnotatedBlock, error)
+
+	// GetRoundResults returns the round results for the given round.
+	GetRoundResults(ctx context.Context, round uint64) (*roothash.RoundResults, error)
+
+	// GetRoundEvents returns the events for the given round.
+	GetRoundEvents(ctx context.Context, round uint64) ([]*staking.Event, error)
```
kostko, line 59: Would probably be better if the indexed events were generalized? E.g. add an `IndexedEvent` structure which is basically an enum of the different events that can be indexed.

go/runtime/history/db.go
```diff
@@ -29,6 +30,14 @@ var (
...
 	// Value is CBOR-serialized roothash.RoundResults.
 	roundResultsKeyFmt = keyformat.New(0x03, uint64(0))
+	// stakingEventsKeyFmt is the round staking events index key format.
+	//
+	// Value is CBOR-serialized list of staking.Events.
+	stakingEventsKeyFmt = keyformat.New(0x04, uint64(0))
+	// pendingStakingEventsKeyFmt is the per height pending staking events index key format.
+	//
+	// Value is a CBOR-serialized list of staking.Events.
+	pendingStakingEventsKeyFmt = keyformat.New(0x05, int64(0))
```
kostko, line 40: Similar as above, should probably be generalized for arbitrary indexed events (probably different kinds of events don't need to be stored under different keys).

go/runtime/history/db.go
```diff
@@ -187,7 +196,7 @@ func (d *DB) commit(blk *roothash.AnnotatedBlock, roundResults *roothash.RoundRe
 			)
 		}
 
-		if blk.Block.Header.Round <= meta.LastRound && meta.LastConsensusHeight != 0 {
+		if blk.Block.Header.Round <= meta.LastRound && blk.Block.Header.Round != 0 {
```
kostko, line 199: A zero round could be valid?

go/runtime/history/db.go
```diff
@@ -202,6 +211,30 @@ func (d *DB) commit(blk *roothash.AnnotatedBlock, roundResults *roothash.RoundRe
...
+				return err
+			}
+			events = append(events, evs...)
+		}
+		// Remove pending events.
+		if err = tx.Delete(pendingStakingEventsKeyFmt.Encode()); err != nil {
+			return err
+		}
+
+		if err = tx.Set(stakingEventsKeyFmt.Encode(blk.Block.Header.Round), cbor.Marshal(events)); err != nil {
```
kostko, line 234: Hm this can get big, any reason why we don't keep them as separate items?

go/worker/compute/executor/committee/node.go
```diff
@@ -729,25 +730,40 @@ func (n *Node) proposeTimeoutLocked(roundCtx context.Context) error {
...
 		n.logger.Error("failed to query round last normal round results",
 			"err", err,
 			"height", height,
 		)
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
-	return state, roundResults, nil
+	// Load all events since last successful round.
```
kostko, line 751: This can be a lot of events in theory. Especially if the runtime is suspended for a while.

go/worker/compute/executor/committee/node.go
```diff
@@ -771,7 +787,7 @@ func (n *Node) handleScheduleBatch(force bool) {
 			return roundCtx, nil, nil, nil, errNotTxnScheduler
 		}
 
-		_, roundResults, err := n.getRtStateAndRoundResults(roundCtx, n.commonNode.CurrentBlockHeight)
+		_, roundResults, _, err := n.getRuntimeRoundData(roundCtx, n.commonNode.CurrentBlockHeight, n.commonNode.CurrentBlock.Header.Round)
```
kostko, line 790: Should also make sure to propose an empty block if there are any events and no transactions.

go/worker/compute/executor/committee/node.go
```diff
@@ -729,25 +730,40 @@ func (n *Node) proposeTimeoutLocked(roundCtx context.Context) error {
...
 			"height", height,
 		)
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
-	return state, roundResults, nil
+	// Load all events since last successful round.
+	var stakingEvents []*staking.Event
+	for r := state.LastNormalRound; r < currentRound; r++ {
```
kostko, line 753: Hm so round `r` will only include events that happened between last normal round and last round? I guess this makes sense as we need a consistent snapshot and until there's a round we don't have one, but we need to document it as this makes events slightly delayed.

go/consensus/tendermint/full/full.go
```diff
@@ -177,6 +177,8 @@ type fullService struct { // nolint: maligned
 	serviceClients   []api.ServiceClient
 	serviceClientsWg sync.WaitGroup
 
+	trackedRuntimes sync.Map
```
kostko, line 180: Any reason why you are using a `sync.Map` instead of a regular map and a lock? Would make it more type safe.

go/consensus/tendermint/full/runtimes.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+	lastNonSeenHeight := lastHeight + 1
+
+	switch {
+	case lastNonSeenHeight > indexHeight:
+		// Already indexed, skip.
+		logger.Debug("height already indexed, skipping", "last_height", lastHeight)
+		return nil
+	case lastNonSeenHeight == indexHeight:
+		// Nothing to reindex, just index the incoming height bellow.
```
kostko, line 92:
```suggestion
		// Nothing to reindex, just index the incoming height below.
```

go/consensus/tendermint/full/runtimes.go
```diff
@@ -0,0 +1,333 @@
...
+
+		crash.Here(crashPointBlockAfterConsensusEventsIndex)
+
+		var annBlk *api.AnnotatedBlock
+		var roundResults *api.RoundResults
+		annBlk, roundResults, err = t.getRuntimeRoothashEvents(runtimeID, height)
+		if err != nil {
+			return fmt.Errorf("failed to query roothash events: %w", err)
+		}
+		if annBlk == nil || roundResults == nil {
```
kostko, line 154: If either is nil? Shouldn't they be consistent? What if one is nil and the other one is not, will things get missed?

kostko (commented):
Hm, the common committee node (`go/worker/common/committee/node.go`) doesn't seem to be invoking the new history-backed `WatchBlocks`? This could cause problems as now it doesn't wait for any history reindexing?

go/worker/compute/executor/committee/node.go
ptrus: Yeah indeed. Not sure what the alternative is, limit max number of events? But in that case the runtime cannot really rely on getting all of the events?

kostko:
> could remove the roothash.WatchBlocks method now (in favor of History().WatchBlocks()). The only issue is the keymanager, as it also needs to watch runtime blocks, but is not compatible with RuntimeRegistry (uses same file for localstorage, runtme registry currently assumes compute runtimes)

So the keymanager runtimes don't have any roothash blocks of their own, but you probably mean the `clientRuntimeWatcher`? That is mainly needed for handling epoch transitions and node updates. Maybe we could avoid using those "unmanaged" nodes and watch for those explicitly?

go/worker/compute/executor/committee/node.go
ptrus: > I guess this makes sense as we need a consistent snapshot and until there's a round we don't have one

Yeah this, but also this way the consensus events match the consensus state the runtime gets. (`CurrentConsensusBlock` that is passed to the runtime is from the height at the previous round). So this makes the events and consensus state the runtime sees consistent. I think having these not in sync would be more confusing.

go/worker/compute/executor/committee/node.go
kostko: Maybe we could do it the same as for transactions -- support batching of events and have a max number of events per batch? The downside would be that this could delay runtimes being informed of consensus changes.

Unfortunately I don't see how this could work without consensus breaking changes as you would need to track which events were included in a batch (e.g., up to which height) and this information (together with the events hash) would need to be subject to discrepancy detection.

go/worker/compute/executor/committee/node.go
kostko: > So this makes the events and consensus state the runtime sees consistent. 

Yes that's right (we should add a test to make sure). I'm not saying that this should be changed, it should just be documented.

ptrus:
> Hm, the common committee node (go/worker/common/committee/node.go) doesn't seem to be invoking the new history-backed WatchBlocks? This could cause problems as now it doesn't wait for any history reindexing?

Yeah currently due to the `clientRuntimeWatcher` using that and not having any runtime supported there (so no history).

> So the keymanager runtimes don't have any roothash blocks of their own, but you probably mean the clientRuntimeWatcher?

Hm another option would be to handle the "no history" case in `common/committee/node.go` (don't watch for runtime blocks and mark as initialized)

go/worker/compute/executor/committee/node.go
ptrus: Hm yeah i think we might indeed need batching. Since the alternatives i can think of are: 
- drop events? - then every runtime needs to be aware that events can be dropped, at that point the runtime would just query consensus state instead of follow events
- share the (untrusted) index db store with the runtime, then the runtime would need to compute the events hash of the events used and provide it for dd. But this probably just moves the batching work to the runtime, not really solves anything?
- not limit - in case of a suspended runtime this can grow unlimitedly - making the runtime unable to be resumed again (due to taking to long to process all the events)

PR #3946, opened 2021-05-14
runtime/common/crypto/signature: Switch Ed25519 to our consensus flavor

Yawning:
We currently use semantics that are close to ed25519-dalek's strict,
but we require using the cofactored verification equation.  Implementing
verification ourselves lets us alter the behavior in the future as
we see fit, though "Just use Ristretto".

TODO:
 * [x] Add tests (blocked on #3947)
 * [x] Add one of those stupid changelog fragments.

### DIFF
--- a/.changelog/3946.internal.md
+++ b/.changelog/3946.internal.md
@@ -1,0 +1,1 @@
+runtime/common/crypto/signature: Switch Ed25519 to our consensus flavor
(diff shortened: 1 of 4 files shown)


Yawning:
You know what makes writing Rust even less fun than it already is?  Having to use CI as the compiler.

kostko (approved):

Issue #3760, opened 2021-03-06
Smarter transaction scheduler


kostko:
Runtimes can provide metadata to the scheduler via the `Meta` field in the CheckTx response.

It would be useful to standardize a few metadata items that the runtimes can emit and the scheduler understands while scheduling batches:

- Priority.
- Custom "weights" where for each "weight type" the runtime can define a maximum block weight. This could be gas or emitted runtime messages. The maximums could either be emitted by the runtime as part of the last ExecuteTx response or queried explicitly by the host via a new call (per round).

PR #3434, opened 2020-10-19
Ptrus/benchmark/runtime scheduler

ptrus:
Closes https://github.com/oasisprotocol/oasis-core/issues/3358

Done:
- [x] generalized incomming_queue into a `txpool` interface
- [x] adds `map`, `ordered_map` txpool implementations
- [x] adds benchmarks for simple `txpool` and `simple scheduler`
- [x] changes the implementation scheduler uses to use `ordered_map`
- [x] minor ["fix"](https://github.com/oasisprotocol/oasis-core/commit/e2695b48e9f5aa41c7d66e85cdc2b4bb56119b80) to the simple scheduler
- [x] do some more benchmarks & gather results
- [x] code-reviews (make sure to re-run benchmarks in case there's any faults in current impl.)
- [x] consider removing the incomming_queue & map txpool implementations before merging, as only the ordered queue will be used (left the commits in, so that the implementations can be easily brought back if ever needed in future)

Benchmark results:
```
Max Batch size 100, 10000 txs.

// Instant dispatch, this directly compares queue implementations.
BenchmarkSimpleSchedulerQueue/BenchmarkNoOpDispatch-8          76      299131505 ns/op   173677836 B/op      959513 allocs/op
BenchmarkSimpleSchedulerMap/BenchmarkNoOpDispatch-8          1278       19598526 ns/op     7015840 B/op       41988 allocs/op
BenchmarkSimpleSchedulerOrderedMap/BenchmarkNoOpDispatch-8    661       37891445 ns/op     8223897 B/op       62877 allocs/op

// Dispatch after some delay. This shows the overall impact of the queue implementation
// on the scheduler performance, in a more "real-world" scenario, where dispatch is not instant.
BenchmarkSimpleSchedulerQueue/BenchmarkDelayDispatch-8           4    5405653214 ns/op   188766716 B/op     1039317 allocs/op
BenchmarkSimpleSchedulerMap/BenchmarkDelayDispatch-8             5    5247941659 ns/op     6997904 B/op       41920 allocs/op
BenchmarkSimpleSchedulerOrderedMap/BenchmarkDelayDispatch-8      4    5304495489 ns/op     8130016 B/op       62004 allocs/op

Max Batch size 1000, 10000 txs.
BenchmarkSimpleSchedulerQueue/BenchmarkNoOpDispatch-8          534      45646138 ns/op    21148272 B/op      123215 allocs/op
[...]

PR #3965, opened 2021-05-24
runtime: add support for transaction priority and weights

ptrus:
Closes https://github.com/oasisprotocol/oasis-core/issues/3760

See also: https://github.com/oasisprotocol/oasis-sdk/pull/146

Adds support for runtime transaction priority and weights that can be set by runtime on `CheckTx`. Transaction are scheduled by priority (higher first). Transaction weights are used by scheduler when creating batches to ensure no batch passes the per-round weight limits. Supported limits are:
- `size` - number of transactions - per batch/round limit is defined in the registry
- `size_bytes` - size of transaction in bytes - per batch/round limit is defined in the registry
- `consensus_messages` - number of emitted consensus messages - per batch/round limit is defined in the registry
- custom runtime specific weights - runtime is queried for the per/batch round limits (TODO: open issue as a follow up - all currently used weights are obtained from the registry anyway)

Per-account sequence numbers can be implemented in a follow-up PR (TODO: open issue), as unless all runtimes are required to implement accounts, the scheduler needs to handle the base case (no per-account seq. numbers) as well.

TODO:
- [x] tests
- [x] open follow up issues (querying per-round custom weight limits from the runtime, and support for per-account sequence numbers)


go/common/version/version.go
```diff
@@ -95,7 +95,7 @@ var (
 	// the runtime.
 	//
 	// NOTE: This version must be synced with runtime/src/common/version.rs.
-	RuntimeHostProtocol = Version{Major: 3, Minor: 0, Patch: 0}
```
kostko, line 98: I believe we already bumped this to `3` in master (21.1 has `2`), so another bump is not needed.

runtime/src/common/quantity.rs
```diff
@@ -47,6 +49,14 @@ impl From<u64> for Quantity {
     }
 }
 
+impl TryFrom<&Quantity> for u64 {
+    type Error = anyhow::Error;
```
kostko, line 53: Would be better to use a more specific error, maybe one of the existing ones could be reused?

go/runtime/transaction/transaction.go
```diff
@@ -149,6 +149,86 @@ type outputArtifacts struct {
 	Output []byte
 }
 
+const (
+	// WeightConsensusMessages is the consensus messages weight key.
+	WeightConsensusMessages = "consensus_messages"
+	// WeightSizeBytes is the transaction byte size weight key.
+	WeightSizeBytes = "size_bytes"
+	// WeightSize is the transaction size weight key.
+	WeightSize = "size"
```
kostko, line 158: Maybe this should be called `count`?

ptrus: Yeah agreed

.changelog/3965.cfg.md
```diff
@@ -0,0 +1,5 @@
+go/worker/compute: remove `schedule_check_tx.enabled` flag
+
+Runtime executor workers now always call check-tx before scheduling runtime
+transactions. The option to disable check-tx for runtime transactions is
```
kostko, line 4:
```suggestion
Runtime executor workers now always call CheckTx before scheduling runtime
transactions. The option to disable CheckTx for runtime transactions is
```

go/runtime/scheduling/simple/txpool/tester.go
```diff
@@ -75,35 +88,43 @@ func testBasic(t *testing.T, pool api.TxPool) {
...
-	if pool.IsQueue() {
-		require.EqualValues(t, batch[0], []byte("hello world"))
-		for i := 0; i < 9; i++ {
-			require.EqualValues(t, batch[i+1], []byte(fmt.Sprintf("call %d", i)))
-		}
-		// Not a duplicate anymore.
-		err = pool.Add([]byte("hello world"))
-		require.NoError(t, err, "Add")
-		require.EqualValues(t, 42, pool.Size(), "Size")
+	/*
```
kostko, line 99: Should this be removed?

go/runtime/transaction/transaction.go
```diff
@@ -149,6 +149,86 @@ type outputArtifacts struct {
...
+	// in the CheckTx response.
+	priority uint64
+	// weights defines the transaction's runtime specific weights as specified
+	// in the CheckTx response.
+	weights map[string]uint64
+
+	hash hash.Hash
+}
+
+func (t *CheckedTransaction) String() string {
```
kostko, line 176: Add documentation comments (also below).

go/worker/compute/executor/committee/node.go
```diff
@@ -150,6 +149,9 @@ type Node struct { // nolint: maligned
 	// safely be used without holding the lock.
 	schedulerMutex sync.RWMutex
 	scheduler      schedulingAPI.Scheduler
+	// Per round weight limits that should be taken into account when scheduling
+	// transactions.
+	roundWeightLimits map[string]uint64
```
kostko, line 154: Why is this member variable needed? It seems to only be used during parameter update? It would be safer if this was removed.

runtime/src/types.rs
```diff
@@ -228,7 +230,28 @@ pub struct CheckTxResult {
...
 
     #[serde(rename = "meta")]
-    pub meta: Option<cbor::Value>,
+    pub meta: Option<CheckTxMetadata>,
+}
+
+/// CheckTx transaction metadata.
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct CheckTxMetadata {
+    #[serde(default)]
```
kostko, line 239: Probably should add the `#[serde(skip_serializing_if = "num_traits::Zero::is_zero")]` to be consistent.

go/runtime/scheduling/simple/txpool/priorityqueue/priority_queue.go
```diff
@@ -0,0 +1,218 @@
...
+	return nil
+}
+
+// Implements api.TxPool.
+func (q *priorityQueue) GetBatch(force bool) []*transaction.CheckedTransaction {
+	q.Lock()
+	defer q.Unlock()
+
+	// Check if a batch is ready.
+	// XXX: could also check if any other weight limits are reached.
```
kostko, line 77: Yeah this probably makes sense, but that would require keeping per-weight tallies for what is in the current pool, right? But that would generalize this nicely and wouldn't hardcode a `.Len()` here.

go/runtime/scheduling/simple/txpool/priorityqueue/priority_queue.go
```diff
@@ -0,0 +1,218 @@
...
+	for w := range q.weightLimits {
+		batchWeights[w] = 0
+	}
+	toRemove := []*item{}
+	q.priorityIndex.Ascend(func(i btree.Item) bool {
+		item := i.(*item)
+
+		// Check if the call fits into the batch.
+		// XXX: potentially there could be smaller transactions that would
+		// fit, which this will miss. Could do some lookahead.
```
kostko, line 93: Should we open a follow up issue for this?

go/runtime/scheduling/simple/txpool/priorityqueue/priority_queue.go
```diff
@@ -0,0 +1,218 @@
...
+		for w, val := range item.tx.Weights() {
+			if _, ok := batchWeights[w]; ok {
+				batchWeights[w] += val
+			}
+		}
+
+		return true
+	})
+
+	// Remove transaction discovered to be to big to even fit the batch.
```
kostko, line 121:
```suggestion
	// Remove transactions discovered to be too big to even fit the batch.
```

go/runtime/scheduling/simple/txpool/priorityqueue/priority_queue.go
```diff
@@ -0,0 +1,218 @@
...
+	// This can happen if weight limits changed, after the transaction was
```
kostko, line 122:
```suggestion
	// This can happen if weight limits changed after the transaction was
```

go/runtime/scheduling/simple/txpool/priorityqueue/priority_queue.go
```diff
@@ -0,0 +1,218 @@
...
+	defer q.Unlock()
+
+	return uint64(q.priorityIndex.Len())
+}
+
+// Implements api.TxPool.
+func (q *priorityQueue) UpdateConfig(cfg api.Config) error {
+	q.Lock()
+	defer q.Unlock()
+	q.maxTxPoolSize = cfg.MaxPoolSize
```
kostko, line 170: Add empty line above.

go/runtime/scheduling/simple/txpool/priorityqueue/priority_queue.go
```diff
@@ -0,0 +1,218 @@
...
+
+	return nil
+}
+
+// Implements api.TxPool.
+func (q *priorityQueue) Clear() {
+	q.Lock()
+	defer q.Unlock()
+
+	q.priorityIndex = btree.New(2)
```
kostko, line 183: Should probably do `q.priorityIndex.Clear` instead:
> This can be much faster than calling Delete on all elements, because that requires finding/removing each element in the tree and updating the tree accordingly. It also is somewhat faster than creating a new tree to replace the old one, because nodes from the old tree are reclaimed into the freelist for use by the new one, instead of being lost to the garbage collector.

kostko (commented):
Did you do any benchmarking to compare with the previous implementation?

ptrus:
> Did you do any benchmarking to compare with the previous implementation?

Yeah I did. I can dig up the results, but generally the adding/getting batches is a bit slower than previous `orderedmap` implementation. But nothing really drastic - and still better than the queue implementation we had initially - i compared with results here: https://github.com/oasisprotocol/oasis-core/pull/3434#issue-505995742). 

The drastic performance degradation is in the `RemoveBatch`, (somewhat expected) as we need to individually delete each element and there's no batch optimization.

go/worker/compute/executor/committee/node.go
ptrus: Yeah indeed not needed at this time (until there is support for custom runtime per-round weight limits). Fixed

go/runtime/scheduling/simple/txpool/priorityqueue/priority_queue.go
ptrus: Indeed, fixed.

go/runtime/scheduling/simple/txpool/priorityqueue/priority_queue.go
```diff
@@ -0,0 +1,234 @@
...
+			}
+		}
+
+		return true
+	})
+
+	// Remove transactions discovered to be too big to even fit the batch.
+	// This can happen if weight limits changed after the transaction was
+	// already set to be scheduled.
+	for _, tw := range toRemove {
```
kostko, line 134: This now needs to also update the pool weights.

ptrus: that's right, i should also add a test case for this

ptrus: fixed, thanks!

kostko (approved):

PR #3981, opened 2021-06-01
rust: specify ring version

ptrus:
Specify exact ring version in `Cargo.toml` as otherwise project depending on oasis-runtime (e.g. oasis-sdk), will need to manually lock `ring` to the exact version.

### DIFF
--- a/.changelog/3981.internal.md
+++ b/.changelog/3981.internal.md
@@ -1,0 +1,1 @@
+rust/Cargo.toml: specify exact ring depedency version
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.13.0"
+version = "0.15.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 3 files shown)


kostko (commented):
Makes sense. What's with all the updates in `Cargo.lock`?

ptrus:
I updated Cargo.lock (via `cargo update`) as a test to ensure `ring` version will stay at the specified version. 

I think now that `oasis-runtime` is also used as a library (in `oasis-sdk`) we should try keeping `Cargo.lock` up to date, since this are the version of dependencies that a crate depending on oasis-runtime will use (unless overridden). E.g. before this change a crate depending on oasis-runtime would get an incompatible ring version with the ring-sgx patch.

kostko (approved):

kostko:
Ok. While doing this, we should also update the crate versions to something other than `0.3.0-alpha`. Or maybe just remove the versions and expect downstream to use git tags?

ptrus:
Hm yeah i think we should just remove the version for now?

kostko:
Agreed, let's remove it.

PR #3985, opened 2021-06-01
go/worker/compute: Propose block on runtime genesis

### LINKED ISSUE #3894: Compute nodes should propose a block on runtime genesis
Currently the compute nodes will only propose a block when there are either transactions or message results. We should make them also propose a block on genesis (e.g., when `rtState.LastNormalRound` is equal to the genesis round) so that any migration code can run as soon as possible and not wait for the first transaction.

kostko:
Fixes #3894

### DIFF
--- a/go/worker/compute/executor/committee/node.go
+++ b/go/worker/compute/executor/committee/node.go
@@ -757,31 +757,38 @@ func (n *Node) getRtStateAndRoundResults(ctx context.Context, height int64) (*ro
 }
 
 func (n *Node) handleScheduleBatch(force bool) {
-	roundCtx, epoch, lastHeader, roundResults, err := func() (context.Context, *committee.EpochSnapshot, *block.Header, *roothash.RoundResults, error) {
+	roundCtx, epoch, lastHeader, rtState, roundResults, err := func() (
+		context.Context,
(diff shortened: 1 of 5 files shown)


Yawning (approved):

go/worker/compute/executor/committee/node.go
```diff
@@ -790,9 +797,18 @@ func (n *Node) handleScheduleBatch(force bool) {
...
 
-	// Ask the scheduler to get us a scheduled batch.
+	// Ask the scheduler to get a batch of transactions for us and see if we should be proposing
+	// a new batch to other nodes.
 	batch := n.scheduler.GetBatch(force)
-	if len(batch) == 0 && (!force || len(roundResults.Messages) == 0) {
+	switch {
+	case len(batch) > 0:
+		// We have some transactions, schedule batch.
+	case len(roundResults.Messages) > 0:
```
ptrus, line 806: I think that if the batch is not ready yet, we should not schedule a batch unless batch flush timeout is reached (`force==true`), even if there are messages?

kostko: Yeah makes sense, will change it back.

ptrus (approved):

Issue #3974, opened 2021-05-28
Runtime transaction scheduler query custom runtime batch weight limits


ptrus:
Runtime transaction scheduler (added in https://github.com/oasisprotocol/oasis-core/pull/3965) should support querying runtime for per-round weight limits. 
 > The limits could either be emitted by the runtime as part of the last ExecuteTx response or queried explicitly by the host via a new call (per round).

PR #3988, opened 2021-06-02
worker/executor: support custom runtime batch weight limits

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/3974

See also: https://github.com/oasisprotocol/oasis-sdk/pull/146

Executor worker now queries `"internal.BatchWeightLimits"` runtime method to receive custom batch weight limits on every round.

### DIFF
--- a/go/runtime/host/helpers.go
+++ b/go/runtime/host/helpers.go
@@ -44,6 +44,14 @@ type RichRuntime interface {
 		method string,
 		args cbor.RawMessage,
 	) (cbor.RawMessage, error)
+
+	// QueryBatchLimits requests the runtime to answer the batch limits query.
+	QueryBatchLimits(
+		ctx context.Context,
+		rb *block.Block,
+		lb *consensus.LightBlock,
+		epoch beacon.EpochTime,
+	) (map[transaction.Weight]uint64, error)
 }
 
 type richRuntime struct {
@@ -117,6 +125,25 @@ func (r *richRuntime) Query(
 		return nil, errors.WithContext(ErrInternal, "malformed runtime response")
 	}
 	return resp.RuntimeQueryResponse.Data, nil
+}
+
+// Implements RichRuntime.
+func (r *richRuntime) QueryBatchLimits(
+	ctx context.Context,
+	rb *block.Block,
+	lb *consensus.LightBlock,
+	epoch beacon.EpochTime,
+) (map[transaction.Weight]uint64, error) {
+	resp, err := r.Query(ctx, rb, lb, epoch, protocol.MethodQueryBatchWeightLimits, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	var weightLimits map[transaction.Weight]uint64
+	if err = cbor.Unmarshal(resp, &weightLimits); err != nil {
+		return nil, errors.WithContext(ErrInternal, fmt.Sprintf("malformed runtime response: %v", err))
+	}
+	return weightLimits, nil
 }
 
 // NewRichRuntime creates a new higher-level wrapper for a given runtime. It provides additional
--- a/go/runtime/host/mock/mock.go
+++ b/go/runtime/host/mock/mock.go
@@ -108,9 +108,18 @@ func (r *runtime) Call(ctx context.Context, body *protocol.Body) (*protocol.Body
 		}}, nil
 	case body.RuntimeQueryRequest != nil:
 		rq := body.RuntimeQueryRequest
-		return &protocol.Body{RuntimeQueryResponse: &protocol.RuntimeQueryResponse{
-			Data: cbor.Marshal(rq.Method + " world"),
-		}}, nil
+
+		switch rq.Method {
+		// Handle Batch Weight Limits request.
+		case protocol.MethodQueryBatchWeightLimits:
+			return &protocol.Body{RuntimeQueryResponse: &protocol.RuntimeQueryResponse{
+				Data: cbor.Marshal(map[transaction.Weight]uint64{}),
+			}}, nil
+		default:
+			return &protocol.Body{RuntimeQueryResponse: &protocol.RuntimeQueryResponse{
+				Data: cbor.Marshal(rq.Method + " world"),
+			}}, nil
+		}
 	default:
 		return nil, fmt.Errorf("(mock) method not supported")
 	}
--- a/go/runtime/host/protocol/connection.go
+++ b/go/runtime/host/protocol/connection.go
@@ -27,8 +27,8 @@ const (
 	connWriteTimeout = 5 * time.Second
 )
 
-// ErrNotReady is the error reported when the Runtime Host Protocol is not initialized.
 var (
+	// ErrNotReady is the error reported when the Runtime Host Protocol is not initialized.
 	ErrNotReady = errors.New(moduleName, 1, "rhp: not ready")
 
 	rhpLatency = prometheus.NewSummaryVec(
--- a/go/worker/compute/executor/committee/node.go
+++ b/go/worker/compute/executor/committee/node.go
@@ -574,6 +574,14 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
 	// Clear the potentially set "is proposing timeout" flag from the previous round.
 	n.proposingTimeout = false
 
+	if header.HeaderType != block.Normal {
+		// If last round was not successful, make sure we re-query the round weight limits
+		// before scheduling a batch as ExecuteTxResponse could have set invalid weights.
+		n.schedulerMutex.Lock()
+		n.limitsLastUpdate = header.Round - 1
+		n.schedulerMutex.Unlock()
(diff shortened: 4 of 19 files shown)


go/runtime/host/helpers.go
```diff
@@ -22,6 +22,11 @@ var (
 	ErrInternal = fmt.Errorf("runtime: internal error")
 )
 
+const (
+	// MethodQueryBatchWeightLimits is the name of the runtime batch weight limits query method.
+	MethodQueryBatchWeightLimits = "internal.BatchWeightLimits"
```
kostko, line 27: Should this be moved to `go/runtime/host/protocol` as it is kind of part of the RHP?

go/runtime/host/helpers.go
```diff
@@ -119,6 +132,42 @@ func (r *richRuntime) Query(
...
+	ctx context.Context,
+	rb *block.Block,
+	lb *consensus.LightBlock,
+	epoch beacon.EpochTime,
+) (map[transaction.Weight]uint64, error) {
+	if rb == nil {
+		return nil, ErrInvalidArgument
+	}
+
+	resp, err := r.Call(ctx, &protocol.Body{
```
kostko, line 146: Why not just use `r.Query`?

go/runtime/host/protocol/connection.go
```diff
@@ -27,10 +27,13 @@ const (
 	connWriteTimeout = 5 * time.Second
 )
 
-// ErrNotReady is the error reported when the Runtime Host Protocol is not initialized.
 var (
+	// ErrNotReady is the error reported when the Runtime Host Protocol is not initialized.
 	ErrNotReady = errors.New(moduleName, 1, "rhp: not ready")
 
+	// ErrQueryNotSupported is the runtime dispatcher query not supported error.
+	ErrQueryNotSupported = errors.New("dispatcher", 1, "query not supported")
```
kostko, line 35: Should this be in `types.go` instead (or a separate `errors.go`)? Maybe we should also namespace the modules somehow? E.g., by using `rhp/dispatcher` as the module name to avoid a clash with other modules?

ptrus: Yeah makes sense. I also just noticed that on the rust side we actually return error code `1` on all/most dispatcher errors, so i'll probably update this error to a different code. (and keep `1` as a "general error").

go/runtime/host/helpers.go
ptrus: Yeah planned to used that :-)

go/runtime/host/helpers.go
```diff
@@ -119,6 +127,34 @@ func (r *richRuntime) Query(
...
+		// Runtime does not implement custom batch limits.
+		return make(map[transaction.Weight]uint64), nil
+	case err != nil:
+		return nil, err
+	default:
+	}
+
+	var weightLimits map[transaction.Weight]uint64
+	if err = cbor.Unmarshal(resp, &weightLimits); err != nil {
+		return nil, errors.WithContext(fmt.Errorf("%v: %w", ErrInternal, err), "malformed runtime response")
```
kostko, line 153: Wow that's a weird construction, should you do something like this instead:
```suggestion
		return nil, errors.WithContext(ErrInternal, fmt.Sprintf("malformed runtime response: %v", err))
```
Maybe we could even make `WithContext` take variadic arguments so it would work without `fmt.Sprintf`.

go/runtime/host/helpers.go
```diff
@@ -119,6 +127,34 @@ func (r *richRuntime) Query(
...
+// Implements RichRuntime.
+func (r *richRuntime) QueryBatchLimits(
+	ctx context.Context,
+	rb *block.Block,
+	lb *consensus.LightBlock,
+	epoch beacon.EpochTime,
+) (map[transaction.Weight]uint64, error) {
+	if rb == nil {
+		return nil, ErrInvalidArgument
+	}
```
kostko, line 139: This is probably not needed as it is already done in `r.Query`.

go/runtime/host/mock/mock.go
```diff
@@ -108,6 +108,16 @@ func (r *runtime) Call(ctx context.Context, body *protocol.Body) (*protocol.Body
 		}}, nil
 	case body.RuntimeQueryRequest != nil:
 		rq := body.RuntimeQueryRequest
+
+		// Handle Batch Weight Limits request.
+		if rq.Method == protocol.MethodQueryBatchWeightLimits {
```
kostko, line 113: Could use a switch with the default being the existing return below.

go/worker/compute/executor/committee/node.go
```diff
@@ -797,6 +851,16 @@ func (n *Node) handleScheduleBatch(force bool) {
 		return
 	}
 
+	// Update per round scheduler parameters.
+	// XXX: could avoid querying here on every round if runtimes would return
+	// next round limits in last execute tx response.
```
kostko, line 856: To make things simpler they could also just return a flag whether the limits have changed and in case they have, we would query the new limits.

runtime/src/transaction/dispatcher.rs
```diff
@@ -59,7 +59,11 @@ pub trait Dispatcher {
         _args: cbor::Value,
     ) -> Result<cbor::Value, RuntimeError> {
         // Default implementation returns an error.
-        Err(RuntimeError::new("dispatcher", 1, "query not supported"))
+        Err(RuntimeError::new(
+            "rhp/dispatcher",
+            2,
```
kostko, line 64: Note that the SDK uses a different error code currently for "method not found".

runtime/src/types.rs
```diff
@@ -245,16 +249,61 @@ pub struct CheckTxMetadata {
     pub weights: Option<BTreeMap<CheckTxWeight, u64>>,
 }
 
+// https://github.com/serde-rs/serde/issues/1560#issuecomment-506915291
```
kostko, line 252: Maybe do as in the issue and put it in a private submodule `strings`.

runtime/src/transaction/dispatcher.rs
ptrus: Hm, but it's also a different module right, so it wouldn't work in any case? So should these common codes/runtime errors be defined somewhere in oasis-core-runtime (protocol?)?

Alternative for now would be to not specially treat "method not found" error here, which doesn't seem ideal

kostko: Yeah I meant module+code. I think defining it in oasis-core-runtime makes sense and then reusing it in the SDK? On the other hand, does it really matter? If there is an error while fetching weights, just log it and default to no weight updates?

ptrus: Yeah i guess it doesn't really matter that much for this case, we can skip this for now, could consider it in future to have some of the error codes defined as part oh the rhp.

go/worker/compute/executor/committee/node.go
ptrus: Would `ExecuteBatchResult` be the right place to add that flag?

kostko: Yes, that sounds right.

ptrus: Hm actually having a flag might actually not be simpler than returning the limits in `RuntimeExecuteTxBatchResponse` as we'd need to handle:
- worker failed to query limits initially (or never did the query since it wasn't the txnscheduler), but the scheduled transaction batch was limits so the scheduled batch was valid - even if `RuntimeExecuteTxBatchResponse` returned no changes to the limits, the worker needs to re-query these.
- runtimes now need to implement both the query to the limits and also track updates to the limits.

Just returning the limits map in `RuntimeExecuteTxBatchResponse` avoids these both cases.

go/worker/compute/executor/committee/node.go
```diff
@@ -662,6 +667,55 @@ func (n *Node) removeTxBatch(batch transaction.RawBatch) error {
...
+	defer n.schedulerMutex.Unlock()
+
+	if n.limitsLastUpdate != 0 && n.limitsLastUpdate >= blk.Header.Round {
+		// Already queried weights for this round.
+		return nil
+	}
+
+	rt := n.GetHostedRuntime()
+	if rt == nil {
+		return fmt.Errorf("updating runtime weight limtis while hosted runtime is not initialized")
```
kostko, line 681:
```suggestion
		return fmt.Errorf("updating runtime weight limits while hosted runtime is not initialized")
```

kostko (approved):

go/worker/compute/executor/committee/node.go
kostko: Ok, let's go with always querying the limits for now and then open an issue to add returning the limits map in execute response?

ptrus: Oh, i just pushed support for this :-) Check last commit.

PR #3999, opened 2021-06-04
 Make current beacon status available via the CLI

### LINKED ISSUE #3897: Make current beacon status available via the CLI
We should make the current beacon status available via the CLI. There is already an API for querying all the required things, but adding CLI is helpful for quick diagnostics.

Yawning:
* [x] Expose all the beacon things needed via gRPC
 * [x] Add the query commands
 * [x] Deal with getting bikeshedded over the output format
 * [x] Be terrorized by github actions and buildkite

Fixes #3897

### DIFF
--- a/go/oasis-node/cmd/debug/beacon/beacon.go
+++ b/go/oasis-node/cmd/debug/beacon/beacon.go
@@ -1,0 +1,102 @@
+// Package beacon implements the beacon introspection debug sub-commands.
+package beacon
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+	"google.golang.org/grpc"
+
+	beacon "github.com/oasisprotocol/oasis-core/go/beacon/api"
+	"github.com/oasisprotocol/oasis-core/go/common/logging"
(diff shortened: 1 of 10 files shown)


go/oasis-node/cmd/debug/beacon/beacon.go
```diff
@@ -0,0 +1,97 @@
...
+
+	b, err := client.GetBeacon(context.Background(), consensus.HeightLatest)
+	if err != nil {
+		logger.Error("failed to query beacon",
+			"err", err,
+		)
+		os.Exit(1)
+	}
+
+	fmt.Printf("beacon: %x", b)
```
kostko, line 67: I guess it would be nicer if this was part of the JSON output? You could define a structure that includes the current beacon and the PVSS state and dump that instead.

Yawning: The command is a debug one, if people are interested in programmatically interacting with the beacon they should use gRPC, not fork/exec.

kostko (approved):

go/oasis-node/cmd/debug/beacon/beacon.go
Yawning: I changed the output, against my better judgement.

PR #4012, opened 2021-06-09
Remove opentracing and jaeger support

### LINKED ISSUE #4011: Remove jager/opentracing support
As far as I know, this is entirely unused, and not actively being subjected to tests.  In an attempt to reduce dependency footprint, maintenance burden and technical debt, support should be removed entirely.

This probably alters the enclave RPC semantics, but is not consensus breaking.

Yawning:
No one uses this, and it's a maintenance burden.

 * [x] Go
 * [x] Rust
 * [x] Get terrorized by github actions.

Fixes #4011


pro-wh (commented):
cc @peterjgilbert, there's nothing coming up where we'd need a time breakdown for the transaction lifecycle, right?

kostko (approved):

PR #4021, opened 2021-06-11
runtime: Stop using webpki to verify the IAS cert chain

### LINKED ISSUE #2683: runtime: Get rid of ring and webpki
In an ideal world, we would not have a mess of Rust crypto libraries used by our system.  Having to depend on someone else's branch of the library (because upstream refuses to merge a trivial branch to support the SGX target) is suboptimal to begin with.  When said branch doesn't build with recent versions of the compiler, it's an even bigger issue (#2666).

This should be easy to do apart from where we use the libraries to validate AVR signatures.  I'm not sure what options there are for doing X.509 certificate validation, especially with RSA in the mix, but in the medium to long term we should switch if at all possible.

Yawning:
It's German for "The Ring The".

This removes the webpki (and consequently) ring dependency by switching to manually verifying the IAS AVR certificate chain, and AVR signature.  It makes a number of assumptions about how IAS does things (that are part of the API specification), so it is probably ok, though if Intel ever decides to change this, we will be rather sad.

The unfortunate thing about this is that I personally like ring/webpki as far as libraries go for the most part, but, having to carry around a fork just to get SGX support due to a trivial patch not being merged is not worth the hassle.

 * [x] Implement our own IAS AVR signature validation
 * [x] Make it more paranoid
 * [x] Add one of those changelog fragments

Fixes #2683

### DIFF
--- a/runtime/src/common/sgx/avr.rs
+++ b/runtime/src/common/sgx/avr.rs
@@ -5,13 +5,16 @@ use anyhow::{anyhow, Result};
 use base64;
 use byteorder::{LittleEndian, ReadBytesExt};
 use chrono::prelude::*;
-use pem::parse_many;
+use lazy_static::lazy_static;
+use oid_registry::{OID_PKCS1_RSAENCRYPTION, OID_PKCS1_SHA256WITHRSA};
 use percent_encoding;
+use rsa::{padding::PaddingScheme, Hash, PublicKey, RSAPublicKey};
 use serde::{Deserialize, Serialize};
 use serde_json;
 use sgx_isa::{AttributesFlags, Report};
+use sha2::{Digest, Sha256};
 use thiserror::Error;
-use webpki;
+use x509_parser::prelude::*;
 
 use crate::common::time::{insecure_posix_time, update_insecure_posix_time};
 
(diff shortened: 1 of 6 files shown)


runtime/src/common/sgx/avr.rs
```diff
@@ -342,47 +326,156 @@ fn validate_avr_signature(
...
+    // be used to verify the leaf certificate's signature.
+    //
+    // This could be more paranoid and check that the cert doesn't
+    // have trailing garbage, the usage is correct, etc, but we can
+    // take it as a matter of faith that it is well-formed since
+    // it is the same as the hard-coded one.
+    //
+    // TODO/perf: In theory this can be done once and only once, but
+    // the borrow checker thwarted my attempts to initialize a tuple
+    // containing a X509Certificate and Pem via lazy_static.
```
kostko, line 390: What was the issue here?

kostko (commented):
Only some minor nits, otherwise I think this looks good! :tada:

runtime/src/common/sgx/avr.rs
Yawning: The X509Certificate doesn't copy/move out of the Pem, so the borrow checker got mad.

runtime/src/common/sgx/avr.rs
```diff
@@ -342,47 +326,156 @@ fn validate_avr_signature(
...
+    if cert_ders[1] != *IAS_TRUST_ANCHOR {
+        return Err(anyhow!("AVR certificate chain trust anchor mismatch"));
+    }
+    let anchor = match parse_x509_certificate(&cert_ders[1]) {
+        Ok((_, cert)) => cert,
+        Err(_) => return Err(AVRError::MalformedCertificateDER.into()),
+    };
```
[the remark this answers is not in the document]
Yawning:
```
error[E0282]: type annotations needed
   --> runtime/src/common/sgx/avr.rs:397:10
    |
397 |         .map_err(|_| AVRError::MalformedCertificateDER.into())?;
    |          ^^^^^^^     ---------------------------------------- this method call resolves to `T`
    |          |
    |          cannot infer type for type parameter `F` declared on the associated function `map_err`
```

kostko (approved):

PR #4024, opened 2021-06-12
Enable development on macOS without docker

nhynes:
This PR:
* adds conditional compilation to the `oasis-core-runtime-loader`
* allows setting the provisioner through `oasis-net-runner` (to unconfined)
* allows running a net without a keymanager since `simple-keymanager` isn't fun to compile with musl

### DIFF
--- a/go/oasis-net-runner/fixtures/default.go
+++ b/go/oasis-net-runner/fixtures/default.go
@@ -113,6 +113,8 @@ func newDefaultFixture() (*oasis.NetworkFixture, error) {
 		RuntimeProvisioner: runtimeProvisioner,
 	}}
 
+	usingKeymanager := viper.IsSet(cfgKeymanagerBinary)
+
 	if viper.GetBool(cfgSetupRuntimes) {
 		fixture.Runtimes = []oasis.RuntimeFixture{
 			// Key manager runtime.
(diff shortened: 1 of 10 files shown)


go/oasis-net-runner/fixtures/default.go
```diff
@@ -215,6 +217,7 @@ func init() {
 	DefaultFixtureFlags.String(cfgNodeBinary, "oasis-node", "path to the oasis-node binary")
 	DefaultFixtureFlags.StringSlice(cfgRuntimeID, []string{"8000000000000000000000000000000000000000000000000000000000000000"}, "runtime ID")
 	DefaultFixtureFlags.StringSlice(cfgRuntimeBinary, []string{"simple-keyvalue"}, "path to the runtime binary")
+	DefaultFixtureFlags.String(cfgRuntimeProvisioner, "sandboxed", "the runtime provisioner: mock, unconfined, or sandbox")
```
kostko, line 220:
```suggestion
	DefaultFixtureFlags.String(cfgRuntimeProvisioner, "sandboxed", "the runtime provisioner: mock, unconfined, or sandboxed")
```

go/oasis-test-runner/oasis/network.go
```diff
@@ -136,6 +137,9 @@ type NetworkCfg struct { // nolint: maligned
 	// externally-provided.
 	UseShortGrpcSocketPaths bool `json:"-"`
 
+	// The runtime sandboxing provisioner. Can be `mock`, `unconfined` or `sandboxed`.
+	RuntimeProvisioner string `json:"-"`
```
kostko, line 141: The runtime provisioner can already be configured per compute worker, why the need for global configuration? You can just leave the default fixture part, so that it is possible to configure the provisioner without rolling a custom fixture. And maybe add the runtime provisioner config to the client/keymanager node fixtures as well.

kostko (approved):

PR #4059, opened 2021-06-21
go/runtime/host/sandbox: Prevent restart storm after 10m

kostko:
Defaults bite again.


Yawning (approved):

Yawning:
Is there anywhere else we are doing this, and if so should we have helpers so this doesn't happen again?

kostko:
> Is there anywhere else we are doing this, and if so should we have helpers so this doesn't happen again?

Good idea, added our own helper and made sure we use it everywhere.

PR #4061, opened 2021-06-22
go/worker/common: Move committee storage to Group

kostko:
Move committee storage to `Group` to make it available to all subworkers (instead of the executor worker having a specialized case) and change it so that it doesn't go through gRPC when communicating with the local node.


go/storage/client/client.go
```diff
@@ -48,6 +64,8 @@ type storageClientBackend struct {
 
 	nodesClient grpc.NodesClient
 	runtime     registry.RuntimeDescriptorProvider
+
+	backendOverrides map[signature.PublicKey]api.Backend
```
ptrus, line 68: Maybe add a note that this doesn't need to be behind a lock as long as it's only updated via options on initialization.

kostko: Done.

ptrus (approved):

PR #4071, opened 2021-06-23
go/consensus/tendermint/apps/governance: Ensure upgrade is in progress

### LINKED ISSUE #4070: Governance should check if upgrade was applied
The governance service currently performs an `EnsureCompatible` sanity check which fails in case someone would manually cancel the upgrade descriptor installed by governance. But the compatibility check only checks the protocol versions and not other properties like upgrade handlers.

It should instead check if the appropriate upgrade descriptor is installed and has been executed and fail in case this is not the case, stopping the node for upgrade.

kostko:
Fixes #4070

### DIFF
--- a/go/consensus/tendermint/apps/governance/governance.go
+++ b/go/consensus/tendermint/apps/governance/governance.go
@@ -149,12 +149,36 @@ func (app *governanceApplication) BeginBlock(ctx *api.Context, request types.Req
 		"upgrade", ud,
 	)
 
-	// Check if we are running the correct version for the upgrade, if not - trigger a consensus halt.
-	if err := ud.EnsureCompatible(); err != nil {
-		ctx.Logger().Error("not running updated binary",
-			"err", err,
-		)
-		return upgrade.ErrStopForUpgrade
+	// Check if the upgrade descriptor is installed in the node and has been executed.
+	if upgrader := ctx.AppState().Upgrader(); upgrader != nil {
+		switch pu, err := upgrader.GetUpgrade(ctx, ud); err {
+		case nil:
+			// Upgrade exists, make sure it is in the process of being applied.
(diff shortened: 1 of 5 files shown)


go/consensus/tendermint/apps/governance/governance.go
```diff
@@ -149,12 +149,36 @@ func (app *governanceApplication) BeginBlock(ctx *api.Context, request types.Req
...
+			if !pu.HasStage(upgrade.UpgradeStageStartup) {
+				ctx.Logger().Error("upgrade pending but not being applied",
+					"handler", ud.Handler,
+					"epoch", ud.Epoch,
+					"upgrade_height", pu.UpgradeHeight,
+					"last_completed_stage", pu.LastCompletedStage,
+				)
+				return upgrade.ErrStopForUpgrade
+			}
+		case upgrade.ErrUpgradeNotFound:
```
ptrus, line 166: This prevents locally canceling a consensus upgrade (2/3+ nodes locally canceling an upcoming pending upgrade), but i think this is fine as in case an already accepted upgrade needs to be reverted/aborted manual intervention is needed anyway so this can be achieved otherwise (e.g. stopping the nodes at the upgrade epoch and doing a "dump-restore" without the upgrade)

ptrus (approved):

go/consensus/tendermint/apps/governance/governance.go
kostko: Yeah, especially since there is also an option to cancel an upgrade via governance (should we consider reducing `upgrade_cancel_min_epoch_diff`?).

ptrus: Yeah that would make sense

Issue #2548, opened 2020-01-14
go/worker/txnscheduler: CheckTx should use a queue


abukosek:
PR #2502 implemented optional checking of transactions before queuing them in the transaction scheduler.  However, the way it's currently implemented, each check is done serially, which can slow-down execution if checking is enabled.

This should be optimized by using a separate queue for CheckTx requests and batching them, as suggested in https://github.com/oasislabs/oasis-core/pull/2502#discussion_r365741640.

samohyes:
So basiclly, right now, the txs are checked before added in txnscheduler. The checks including like whether there is enough balance in the accout or whether it's a successful contract call, right? The code looks complicated now.

abukosek:
The checks are disabled by default for now, so the behavior is the same as before.  If enabled, the checks depend on the runtime that's used.  In our oasis-runtime, the checks are done here: https://github.com/oasislabs/oasis-runtime/blob/9bc96c04d1e6503096aaefdfa6b999b26b972943/src/methods.rs#L26-L45

samohyes:
Thanks for the reply. 
As I can see the core check here in the CheckTx function is here 
https://github.com/oasislabs/oasis-core/blob/master/go/worker/txnscheduler/committee/node.go#L164
with the call 
`resp, err := workerHost.Call(ctx, checkRq)`

So according to what you said, this workerHost.Call will call the check in oasis-runtime(the link you provides), right?

abukosek:
Yes, the `workerHost.Call()` will make a call into the SGX enclave that's running the runtime and the method I linked before will be executed.  The rest of the new code is basically just error checking and serialization/deserialization.

PR #4131, opened 2021-07-12
go/worker/executor: batch CheckTx transactions

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/2548

TODO: 
- [ ] make sure in oasis-sdk that batch-wide limits are not applied to check-tx batches - as for check-tx the scheduler cannot know the weights up-front. so the runtime should not enforce batch-wide limits

### DIFF
--- a/go/worker/compute/executor/committee/batch.go
+++ b/go/worker/compute/executor/committee/batch.go
@@ -44,5 +44,6 @@ func (ub *unresolvedBatch) resolve(ctx context.Context, sb storage.Backend) (tra
 	if err != nil {
 		return nil, fmt.Errorf("failed to fetch inputs from storage: %w", err)
 	}
+	ub.batch = batch
 	return batch, nil
 }
--- a/go/worker/compute/executor/committee/node.go
+++ b/go/worker/compute/executor/committee/node.go
@@ -31,6 +31,7 @@ import (
 	runtimeRegistry "github.com/oasisprotocol/oasis-core/go/runtime/registry"
 	"github.com/oasisprotocol/oasis-core/go/runtime/scheduling"
 	schedulingAPI "github.com/oasisprotocol/oasis-core/go/runtime/scheduling/api"
+	"github.com/oasisprotocol/oasis-core/go/runtime/scheduling/simple/orderedmap"
 	"github.com/oasisprotocol/oasis-core/go/runtime/transaction"
 	scheduler "github.com/oasisprotocol/oasis-core/go/scheduler/api"
 	storage "github.com/oasisprotocol/oasis-core/go/storage/api"
@@ -52,9 +53,7 @@ var (
 
 	// Transaction scheduling errors.
 	errNoBlocks        = fmt.Errorf("executor: no blocks")
-	errNotReady        = fmt.Errorf("executor: runtime not ready")
 	errNotTxnScheduler = fmt.Errorf("executor: not transaction scheduler in this round")
-	errDuplicateTx     = p2pError.Permanent(p2pError.Relayable(fmt.Errorf("executor: duplicate transaction")))
 
 	// proposeTimeoutDelay is the duration to wait before submitting the propose timeout request.
 	proposeTimeoutDelay = 2 * time.Second
@@ -141,7 +140,9 @@ type Node struct { // nolint: maligned
 
 	lastScheduledCache    *lru.Cache
 	scheduleMaxTxPoolSize uint64
-	scheduleCh            *channels.RingChannel
+
+	checkTxCh    *channels.RingChannel
+	checkTxQueue *orderedmap.OrderedMap
 
 	// The scheduler mutex is here to protect the initialization
 	// of the scheduler variable and updates to scheduler parameters.
@@ -258,23 +259,25 @@ func (n *Node) HandlePeerMessage(ctx context.Context, message *p2p.Message, isOw
 			return true, nil
 		}
 
-		// Check transaction before queuing it.
-		tx, err := n.checkTx(ctx, rawTx)
-		if err != nil {
-			return true, err
+		// Skip recently scheduled transactions.
+		if n.lastScheduledCache != nil {
+			if _, b := n.lastScheduledCache.Get(hash.NewFromBytes(rawTx)); b {
+				n.logger.Debug("not scheduling duplicate transaction", "tx", rawTx)
+				return true, nil
+			}
 		}
-
-		n.logger.Debug("queuing transaction",
-			"transaction", tx,
+		// Queue transaction for checks.
+		n.logger.Debug("queuing transaction for check",
+			"tx", rawTx,
 		)
-		err = n.QueueTx(tx)
-		if err != nil {
+		if err := n.checkTxQueue.Add(rawTx); err != nil {
 			n.logger.Error("unable to queue transaction",
-				"tx", tx,
+				"tx", rawTx,
 				"err", err,
 			)
 			return true, err
 		}
+		n.checkTxCh.In() <- struct{}{}
 		return true, nil
 
 	case message.ProposedBatch != nil:
@@ -542,7 +545,7 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
 					"header_io_root", header.IORoot,
 					"proposed_io_root", state.proposedIORoot,
 					"header_type", header.HeaderType,
-					"batch", state.raw,
+					"batch_size", len(state.raw),
 				)
 				return
 			}
@@ -551,14 +554,14 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
 			batchProcessingTime.With(n.getMetricLabels()).Observe(time.Since(state.batchStartTime).Seconds())
 
 			n.logger.Debug("removing processed batch from queue",
-				"batch", state.raw,
+				"batch_size", len(state.raw),
 				"io_root", header.IORoot,
 			)
 			// Removed processed transactions from queue.
 			if err := n.removeTxBatch(state.raw); err != nil {
 				n.logger.Warn("failed removing processed batch from queue",
 					"err", err,
-					"batch", state.raw,
+					"batch_size", len(state.raw),
 				)
 			}
 		}()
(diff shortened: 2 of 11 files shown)


go/worker/compute/executor/committee/batch.go
```diff
@@ -44,5 +44,6 @@ func (ub *unresolvedBatch) resolve(ctx context.Context, sb storage.Backend) (tra
 	if err != nil {
 		return nil, fmt.Errorf("failed to fetch inputs from storage: %w", err)
 	}
+	ub.batch = batch
```
kostko, line 47: Nice catch!

go/worker/compute/executor/committee/node.go
```diff
@@ -141,8 +140,9 @@ type Node struct { // nolint: maligned
 
 	lastScheduledCache    *lru.Cache
 	scheduleMaxTxPoolSize uint64
-	scheduleCh            *channels.RingChannel
+	checkTxCh             *channels.RingChannel
 
+	checkTxQueue *orderedmap.OrderedMap
```
kostko, line 145: Group with above.

go/worker/compute/executor/committee/node.go
```diff
@@ -258,23 +258,18 @@ func (n *Node) HandlePeerMessage(ctx context.Context, message *p2p.Message, isOw
...
 
-		// Check transaction before queuing it.
-		tx, err := n.checkTx(ctx, rawTx)
-		if err != nil {
-			return true, err
-		}
-
-		n.logger.Debug("queuing transaction",
-			"transaction", tx,
+		// Queue transaction for check tx .
```
kostko, line 261:
```suggestion
		// Queue transaction for checks.
```

go/worker/compute/executor/committee/node.go
```diff
@@ -258,23 +258,18 @@ func (n *Node) HandlePeerMessage(ctx context.Context, message *p2p.Message, isOw
...
+		n.logger.Debug("queuing transaction for check",
+			"tx", rawTx,
 		)
-		err = n.QueueTx(tx)
-		if err != nil {
+		if err := n.checkTxQueue.Add(rawTx); err != nil {
```
kostko, line 265: One possible problem with this change is that nodes will now propagate bad transactions (e.g., ones that fail checks). Maybe this should block until the transaction in the batch passes the checks (or the context is cancelled) and propagate the correct error?

On the other hand this would delay transaction propagation until the whole batch is checked.

go/worker/compute/executor/committee/node.go
```diff
@@ -581,74 +576,81 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
 
-		n.scheduleCh.In() <- struct{}{}
+		n.checkTxCh.In() <- struct{}{}
 	}
 }
 
-// checkTx requests the runtime to check the validity of the given transaction.
-func (n *Node) checkTx(ctx context.Context, rawTx []byte) (*transaction.CheckedTransaction, error) {
+// checkTxs requests the runtime to check the validity of the given transaction batch.
+func (n *Node) checkTxs() {
```
kostko, line 584:
```suggestion
func (n *Node) checkTxBatch() {
```

go/worker/compute/executor/committee/node.go
```diff
@@ -581,74 +576,81 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
+	n.checkTxQueue.RemoveBatch(batch)
+
+	// Queue checked transactions for scheduling.
+	n.queueTxs(txs)
 }
 
-// QueueTx queues a runtime transaction for scheduling.
-func (n *Node) QueueTx(tx *transaction.CheckedTransaction) error {
+// queues a runtime transaction batch for scheduling.
+func (n *Node) queueTxs(txs []*transaction.CheckedTransaction) {
```
kostko, line 626:
```suggestion
func (n *Node) queueTxBatch(txs []*transaction.CheckedTransaction) {
```

go/worker/compute/executor/committee/node.go
```diff
@@ -581,74 +576,81 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
-func (n *Node) checkTx(ctx context.Context, rawTx []byte) (*transaction.CheckedTransaction, error) {
+// checkTxs requests the runtime to check the validity of the given transaction batch.
+func (n *Node) checkTxs() {
+	batch := n.checkTxQueue.GetBatch()
+	if len(batch) == 0 {
+		return
+	}
+	rt := n.GetHostedRuntime()
+	if rt == nil {
+		n.logger.Error("CheckTx: hosted runtime not initialized")
```
kostko, line 591: Should this rather be logged as debug?

go/worker/compute/executor/committee/node.go
```diff
@@ -581,74 +576,81 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
 }
 
-// QueueTx queues a runtime transaction for scheduling.
-func (n *Node) QueueTx(tx *transaction.CheckedTransaction) error {
+// queues a runtime transaction batch for scheduling.
+func (n *Node) queueTxs(txs []*transaction.CheckedTransaction) {
 	n.schedulerMutex.RLock()
 	defer n.schedulerMutex.RUnlock()
 
-	if n.scheduler == nil {
```
kostko, line 622: Is the scheduler nil check no longer needed?

go/worker/compute/executor/committee/node.go
```diff
@@ -581,74 +576,81 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
+	// Remove the checked transaction batch.
+	n.checkTxQueue.RemoveBatch(batch)
+
+	// Queue checked transactions for scheduling.
+	n.queueTxs(txs)
 }
 
-// QueueTx queues a runtime transaction for scheduling.
-func (n *Node) QueueTx(tx *transaction.CheckedTransaction) error {
+// queues a runtime transaction batch for scheduling.
```
kostko, line 625:
```suggestion
// queueTxBatch queues a runtime transaction batch for scheduling.
```

go/worker/compute/executor/committee/node.go
```diff
@@ -581,74 +576,81 @@ func (n *Node) HandleNewBlockLocked(blk *block.Block) {
...
-	if n.scheduler == nil {
-		return errNotReady
-	}
-
-	// Check if transaction was recently scheduled.
-	if n.lastScheduledCache != nil {
-		if _, b := n.lastScheduledCache.Get(tx.Hash()); b {
-			return errDuplicateTx
+	for _, tx := range txs {
+		// Skip recently scheduled transactions.
```
kostko, line 631: Maybe this check should be performed before even checking the batch (e.g. in `checkTxBatch`)?

go/worker/compute/executor/committee/node.go
```diff
@@ -1848,8 +1850,8 @@ func (n *Node) worker() {
 		case <-txnScheduleTicker.C:
 			// Force scheduling a batch if possible.
 			n.handleScheduleBatch(true)
-		case <-n.scheduleCh.Out():
-			// Regular scheduling of a batch.
```
kostko, line 1852: Add a comment back, e.g. "Check any queued transactions and attemt regular scheduling." or similar.

go/worker/compute/executor/committee/orderedmap/ordered_map.go
```diff
@@ -0,0 +1,194 @@
+// Package orderedmap implements a queue backed by an ordered map.
```
kostko, line 1: Should this instead go somewhere under `go/runtime/scheduling`?

go/worker/compute/executor/tests/tester.go
```diff
@@ -148,8 +164,4 @@ blockLoop:
 			t.Fatalf("failed to receive block")
 		}
 	}
-
-	// Requeuing same call should fail.
```
kostko, line 152: Is this tested somewhere else?

go/worker/compute/executor/committee/orderedmap/ordered_map.go
```diff
@@ -0,0 +1,194 @@
...
+)
+
+type pair struct {
+	Key   hash.Hash
+	Value []byte
+
+	element *list.Element
+}
+
+type OrderedMap struct {
```
kostko, line 22: Add documentation comment.

go/worker/compute/executor/committee/orderedmap/ordered_map.go
```diff
@@ -0,0 +1,194 @@
...
+
+import (
+	"container/list"
+	"fmt"
+	"sync"
+
+	"github.com/hashicorp/go-multierror"
+
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
+	"github.com/oasisprotocol/oasis-core/go/runtime/scheduling/simple/txpool/api"
```
kostko, line 12: So this is tied to the simple scheduler (due to the error type), maybe it should go into a package under `simple`? It seems similar to the txpool API but not quite the same.

go/worker/compute/executor/committee/orderedmap/ordered_map_test.go
```diff
@@ -0,0 +1,97 @@
+// Package tests is a collection of transactinon pool test cases.
```
kostko, line 1: This needs to be updated.

kostko (commented):
Is there any performance impact?

ptrus:
> Is there any performance impact?

Yeah the single-machine accounts CI benchmarks (from oasis-sdk CI) goes from ~35 to ~150, and the new bottleneck is serially fetching accounts from storage via host (https://github.com/oasisprotocol/oasis-sdk/issues/278).  Also hitting the dispatchers `BACKLOG_SIZE` limit with large number of concurrent transactions is gone.

go/worker/compute/executor/committee/node.go
ptrus: > One possible problem with this change is that nodes will now propagate bad transactions

Hm yeah, but that I think that this is already the case even now, for all nodes other than the active primary executor workers (which is usually smaller than the backup committee)?

go/worker/compute/executor/committee/node.go
ptrus: Yeah not needed anymore, actually the lock is also not needed anymore. 
Because now `queueTxs` is guaranteed to be called only after the scheduler is already initialized (since it's only called from the main worker loop). While before `QueueTx` could be called before the worker was initialized.

go/worker/compute/executor/committee/orderedmap/ordered_map.go
ptrus: Yeah good idea, i'll move it there.

go/worker/compute/executor/committee/node.go
ptrus: Yeah could do that, probably even before, in `HandlePeerMessage`. Not sure if we should also move the insert there? or only insert it here, once the transaction has passed the check?

Currently the transactions that fail CheckTx are missed by this cache - however a transaction could fail a CheckTx but pass it a future time. So probably better to keep the insertion here, only for transactions that pass the check.

go/worker/compute/executor/tests/tester.go
ptrus: Hm probably, but not explicitly/directly. I'll add the test back (just need to change the check a bit)

go/worker/compute/executor/committee/node.go
kostko: Yes, I think it makes sense to move the cache check into `HandlePeerMessage`, but insertion should be kept here after the transaction passes CheckTx for the reasons you mentioned above.

kostko (approved):

PR #4133, opened 2021-07-16
 runtime: Replace serde-cbor with oasis-cbor

kostko:
Includes two cleanup commits from #4101.

TODO
* [x] Use released version of oasis-cbor.


Yawning (approved):

runtime/src/types.rs
```diff
@@ -161,8 +190,8 @@ pub enum Body {
         header: Header,
         epoch: EpochTime,
         method: String,
-        #[serde(skip_serializing_if = "Option::is_none")]
-        args: Option<cbor::Value>,
+        #[cbor(optional)]
```
ptrus, line 193: nice :+1:

ptrus (approved):

PR #4155, opened 2021-07-28
storage/checkpoint_sync: add timeout for fetching checkpoints from nodes

ptrus:
Alternative fix to https://github.com/oasisprotocol/oasis-core/pull/4152 
Not sure which one we prefer


kostko (approved):
I think this one makes more sense as with the other one you can't know if the checkpoints are actually any good.

PR #4161, opened 2021-07-30
workload/runtime: ensure correct round for validating events

ptrus:
Ensure `validateEvents` queries events for the correct round.

Reason for failure in https://buildkite.com/oasisprotocol/oasis-core-long-tests/builds/852

### DIFF
--- a/.changelog/4161.internal.md
+++ b/.changelog/4161.internal.md
@@ -1,0 +1,1 @@
+workload/runtime: ensure correct round for validating events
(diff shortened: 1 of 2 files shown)


kostko (approved):
Maybe it would be good to have a SubmitTx variant that returns metadata (e.g. at which round the tx was included)? That should already be available, just needs to be exposed.

ptrus:
> Maybe it would be good to have a SubmitTx variant that returns metadata (e.g. at which round the tx was included)? That should already be available, just needs to be exposed.

Yeah it would, i'll open an issue

Issue #4162, opened 2021-07-30
Runtime client SubmitTx variant with metadata


ptrus:
It would be useful to have a runtime `SubmitTx` variant that returns metadata (e.g. at which round the tx was included). Currently one needs to manually watch the runtime blocks to obtain such info.

PR #4166, opened 2021-08-02
runtime/client: SubmitTx variant that returns metadata

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/4162

Not sure if there's any other potentially useful metadata that could be included

### DIFF
--- a/go/runtime/client/api/api.go
+++ b/go/runtime/client/api/api.go
@@ -47,6 +47,13 @@ type RuntimeClient interface {
 	// for transaction execution results.
 	SubmitTx(ctx context.Context, request *SubmitTxRequest) ([]byte, error)
 
+	// SubmitTxMeta submits a transaction to the runtime transaction scheduler and waits for
+	// transaction execution results.
+	//
+	// Response includes transaction metadata - e.g. round at which the transaction was included
+	// in a block.
+	SubmitTxMeta(ctx context.Context, request *SubmitTxRequest) (*SubmitTxMetaResponse, error)
+
 	// SubmitTxNoWait submits a transaction to the runtime transaction scheduler but does
 	// not wait for transaction execution.
 	SubmitTxNoWait(ctx context.Context, request *SubmitTxRequest) error
@@ -102,6 +109,13 @@ type RuntimeClientService interface {
 type SubmitTxRequest struct {
 	RuntimeID common.Namespace `json:"runtime_id"`
 	Data      []byte           `json:"data"`
+}
+
+// SubmitTxMetaResponse is the SubmitTxMeta response.
+type SubmitTxMetaResponse struct {
+	Round      uint64 `json:"round"`
+	Output     []byte `json:"data"`
+	BatchOrder uint32 `json:"batch_order"`
 }
 
 // CheckTxRequest is a CheckTx request.
--- a/go/runtime/client/api/api.go
+++ b/go/runtime/client/api/api.go
@@ -13,6 +13,7 @@ import (
 	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
 	"github.com/oasisprotocol/oasis-core/go/roothash/api/block"
 	enclaverpc "github.com/oasisprotocol/oasis-core/go/runtime/enclaverpc/api"
+	"github.com/oasisprotocol/oasis-core/go/runtime/host/protocol"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


go/runtime/client/api/api.go
```diff
@@ -104,6 +111,13 @@ type SubmitTxRequest struct {
 	Data      []byte           `json:"data"`
 }
 
+// SubmitTxMetaResponse is the SubmitTxMeta response.
+type SubmitTxMetaResponse struct {
+	Round      uint64
+	Output     []byte
+	BatchOrder uint32
```
kostko, line 118: Add explicit serialization annotations.

nhynes:
It'd be great if `SubmitTxMeta` could return the innermost error as `(code, module, message)` with the context (e.g., "happened during `checkTx`"). Right now, the actual runtime error gets stuffed into the `message` where it's not as useful. My workaround is

```diff
diff --git a/go/runtime/host/helpers.go b/go/runtime/host/helpers.go
index a70932f4b..baee1a80a 100644
--- a/go/runtime/host/helpers.go
+++ b/go/runtime/host/helpers.go
@@ -90,7 +90,7 @@ func (r *richRuntime) CheckTx(
 	// Interpret CheckTx result.
 	result := resp.RuntimeCheckTxBatchResponse.Results[0]
 	if !result.IsSuccess() {
-		return nil, errors.WithContext(ErrCheckTxFailed, result.Error.String())
+		return nil, errors.FromCode(result.Error.Module, result.Error.Code, result.Error.Message)
 	}
 
 	return result.ToCheckedTransaction(tx), nil
```

When prompted with "Why not make all errors application-layer, reserving the gRPC statuses for actual protocol errors?" Jernej responded
> **Jernej**
> the problem is that currently SubmitTx returns a byte slice which is the runtime-dependent result (which for the sdk includes error codes). for errors encountered during local CheckTx we would need to wrap them in some additional struct which would then result in three layers of errors: i) included in block but failed, error is in the runtime-dependent result; ii) failed local checktx, error is in the metadata wrapper struct; iii) some other error, error is in the grpc error

ptrus:
> It'd be great if SubmitTxMeta could return the innermost error as (code, module, message) with the context (e.g., "happened during checkTx"). Right now, the actual runtime error gets stuffed into the message where it's not as useful. My workaround is


I added `CheckTxError error` in the metadata response which contains the CheckTx error if it failed. Is this close to what you had in mind?
self-todo: add a test

go/runtime/client/api/api.go
```diff
@@ -104,6 +111,14 @@ type SubmitTxRequest struct {
 	Data      []byte           `json:"data"`
 }
 
+// SubmitTxMetaResponse is the SubmitTxMeta response.
+type SubmitTxMetaResponse struct {
+	Round        uint64 `json:"round"`
+	Output       []byte `json:"data"`
+	BatchOrder   uint32 `json:"batch_order"`
+	CheckTxError error  `json:"check_tx_error"`
```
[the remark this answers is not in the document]
ptrus: fixed

go/runtime/client/client.go
```diff
@@ -137,6 +148,15 @@ func (c *runtimeClient) submitTx(ctx context.Context, request *api.SubmitTxReque
 
 // Implements api.RuntimeClient.
 func (c *runtimeClient) SubmitTx(ctx context.Context, request *api.SubmitTxRequest) ([]byte, error) {
+	resp, err := c.SubmitTxMeta(ctx, request)
```
kostko, line 151: Does this now ignore errors during CheckTx?

ptrus: Yeah it did, fixed. Will update the mock runtime used in node tests to enable forcing check-tx errors so this can be tested.

nhynes:
> I added CheckTxError error in the metadata response which contains the CheckTx error if it failed. Is this close to what you had in mind?

I think that should work. We'll see :) Thanks!

go/runtime/host/mock/mock.go
```diff
@@ -20,6 +21,8 @@ import (
 type provisioner struct {
 }
 
+var CheckTxFailInput = []byte("checktx-mock-fail")
```
kostko, line 24: Add documentation comment.

go/runtime/host/mock/mock.go
```diff
@@ -95,12 +98,21 @@ func (r *runtime) Call(ctx context.Context, body *protocol.Body) (*protocol.Body
...
 
 		var results []protocol.CheckTxResult
-		for range rq.Inputs {
-			results = append(results, protocol.CheckTxResult{
-				Error: protocol.Error{
-					Code: errors.CodeNoError,
-				},
-			})
+		for _, input := range rq.Inputs {
+			if bytes.Equal(input, CheckTxFailInput) {
```
kostko, line 102: Let's use a switch in case we will later add more special inputs.

kostko (approved):

PR #4183, opened 2021-08-05
go/roothash/api/block: Use custom TimeStamp type for block's header

tjanez:
This enables prettier Oasis Node's `control status` CLI command's output for runtimes' `latest_time` field and matches the format of consensus' `latest_time` field.

### DIFF
--- a/go/roothash/api/block/header.go
+++ b/go/roothash/api/block/header.go
@@ -3,6 +3,7 @@ package block
 import (
 	"bytes"
 	"errors"
+	"time"
 
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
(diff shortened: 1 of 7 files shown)


go/roothash/api/block/header.go
```diff
@@ -17,6 +19,16 @@ var ErrInvalidVersion = errors.New("roothash: invalid version")
 // HeaderType is the type of header.
 type HeaderType uint8
 
+// TimestampType is a custom time stamp type that encodes like time.Time when
+// marshaling to JSON.
+type TimestampType uint64
```
kostko, line 24: This should just be `Timestamp`. The `HeaderType` above has `Type` because it's the "header type" ;)

go/roothash/api/block/header.go
```diff
@@ -17,6 +19,16 @@ var ErrInvalidVersion = errors.New("roothash: invalid version")
...
+
+// MarshalJSON encodes the time stamp to JSON by converting it from Unix time to
+// local time.
+func (tst TimestampType) MarshalJSON() ([]byte, error) {
```
kostko, line 28: What about unmarshalling? Add a test that makes sure this round-trips correctly.

go/roothash/api/block/header.go
tjanez: Oh, ok :slightly_smiling_face:. Fixed.

go/roothash/api/block/header.go
tjanez: Added unmarshaling and tests.

kostko (approved):

Issue #4192, opened 2021-08-08
Prevent startup when RLIMIT_NOFILE is too low


kostko:
The `oasis-node` process currently warns when the `RLIMIT_NOFILE` is set to a value less than or equal to 1024. Given that a too low limit can cause problems with BadgerDB we should probably fail startup if the limit is below something like 50k.

PR #4193, opened 2021-08-09
prevent startup when RLIMIT_NOFILE is too low

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/4192

### DIFF
--- a/go/oasis-node/cmd/common/common.go
+++ b/go/oasis-node/cmd/common/common.go
@@ -34,7 +34,9 @@ const (
 	CfgConfigFile = "config"
 	CfgDataDir    = "datadir"
 
-	badDefaultRlimit = 1024
+	// RequiredRLimit is the minimum required RLIMIT_NOFILE as too low of a
+	// limit can cause problems with BadgerDB.
+	RequiredRlimit = 50_000
 )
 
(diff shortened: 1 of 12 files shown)


Yawning (approved):

go/oasis-node/cmd/common/common.go
```diff
@@ -201,28 +203,22 @@ func initRlimit() error {
...
+		return fmt.Errorf("failed to query RLIMIT_NOFILE: %w", err)
 	}
 
-	if rlim.Cur <= badDefaultRlimit {
-		rootLog.Warn("the node user has a very low RLIMIT_NOFILE, consider increasing",
-			"cur", rlim.Cur,
-			"max", rlim.Max,
-		)
+	desiredLimit := viper.GetUint64(CfgDebugRlimit)
+	if flags.DebugDontBlameOasis() && desiredLimit != 0 {
```
Yawning, line 210: Should this be `desiredLimit > 0 && desiredLimit != rlim.Cur`?

ptrus: yeah i think that makes more sense, will update :+1:

PR #4198, opened 2021-08-11
runtime: Deprecate transaction::MethodDispatcher

Yawning:
By deprecate I mean, move into the simple-keyvalue runtime's main.rs,
preserving existing functionality in the worst possible way.

Part of #3572


.changelog/4198.internal.md
```diff
@@ -0,0 +1 @@
+runtime: Deprecate transaction::MethodDispatcher
```
kostko, line 1: Should probably be breaking and say remove instead of deprecate as it was already deprecated before.

tests/runtimes/simple-keyvalue/src/main.rs
```diff
@@ -26,19 +34,424 @@ use oasis_core_runtime::{
...
+/// # Examples
+///
+/// This macro should be invoked using a concrete API generated by `runtime_api`
+/// as follows:
+/// ```rust,ignore
+/// with_api! {
+///     register_runtime_txn_methods!(txn_dispatcher, api);
+/// }
+/// ```
+#[deprecated(note = "see oasis-core#3572")]
```
kostko, line 72: Probably can remove the deprecation attributes from this test-only version?

tests/runtimes/simple-keyvalue/src/main.rs
```diff
@@ -26,19 +34,424 @@ use oasis_core_runtime::{
...
     with_api, AddEscrow, Key, KeyValue, ReclaimEscrow, Transfer, UpdateRuntime, Withdraw,
 };
 
+// This is the old runtime method dispatcher that used to be part
+// of the oasis-core/runtime API.  It has been moved here since
+// new runtimes should use the SDK instead.
+//
+// Note: I'm too stupid to figure out how to make this into a separate
+// module, and I don't care enough to put more than "I tried it and
+// everything broke" amounts of effort into it.
```
kostko, line 59: Hm, what was the issue?

Yawning: Importing it by trying to abuse `#[path = "transaction.rs"]` from the upgraded version didnt' work because of the macro or something.

kostko (approved):

PR #4204, opened 2021-08-16
go/runtime/client: Fail SubmitTx early for unsupported runtimes

### LINKED ISSUE #4202: Runtime SubmitTx should fail for unsupported runtimes
Currently the `SubmitTx` family of methods will work even when specifying runtimes which are not configured as supported by the node. This will result in the transaction being submitted via the P2P network, but the client will be unable to see any results so the blocking variants will just block forever (or until a timeout).

The correct thing to do would be to fail SubmitTx early.

kostko:
Fixes #4202 

Make sure that the runtime is actually among the supported runtimes as
otherwise we will not be able to actually get any results back.

### DIFF
--- a/go/runtime/client/client.go
+++ b/go/runtime/client/client.go
@@ -85,6 +85,12 @@ func (c *runtimeClient) submitTx(ctx context.Context, request *api.SubmitTxReque
 		return nil, fmt.Errorf("client: cannot submit transaction, p2p disabled")
 	}
 
+	// Make sure that the runtime is actually among the supported runtimes for this node as
+	// otherwise we will not be able to actually get any results back.
+	if _, err := c.common.runtimeRegistry.GetRuntime(request.RuntimeID); err != nil {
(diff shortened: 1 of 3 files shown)


go/runtime/client/client.go
```diff
@@ -85,6 +85,12 @@ func (c *runtimeClient) submitTx(ctx context.Context, request *api.SubmitTxReque
 		return nil, fmt.Errorf("client: cannot submit transaction, p2p disabled")
 	}
 
+	// Make sure that the runtime is actually among the supported runtimes for this node as
+	// otherwise we will not be able to actually get any results back.
+	if _, err := c.common.runtimeRegistry.GetRuntime(request.RuntimeID); err != nil {
```
ptrus, line 90: Could add a test for this, maybe include it in one of the exiting e2e tests?

kostko: Added.

ptrus (approved):

Issue #1683, opened 2019-05-07
Do a better job about sanitizing sensitive data (eg: private keys)


Yawning:
There is no reason to leave key material on the heap/stack after use if said values are not going to be used again, and as much as possible such things should be cleared with a `memset_s` equivalent.

 * `go/common/crypto/signature.PrivateKey` needs a `Reset()` method, and said method should be called.  As far as I am aware, naively clearing the slice gets the desired behavior.

 * Rust: Apart from in the deoxysii-rust code, no attempt is made to sanitize data.  The "correct" way would be to use `zeroize` or similar, but our `impl_bytes!` macro derives `Copy`, so we can't have `Drop`.

Note: Some people disagree with this practice, or think that it is unnecessary.  The SGX developer documentation recommends/requires it.

Yawning:
I'm sort of convinced that this is mostly a lost cause in Go, though we can make a best-effort attempt at it.

pro-wh:
should we keep this open for the Go side?

PR #4242, opened 2021-09-02
keymanager-api-common: Implement Zeroize for secret types

kostko:
See #1683

TODO
* [x] Check if there are other places which need this.


pro-wh (approved):
lg, if `#[zeroize(drop)]` does what it sounds like

kostko:
> lg, if #[zeroize(drop)] does what it sounds like

It implements `Drop` that simply calls `self.zeroize()` :-)

PR #4243, opened 2021-09-02
go/common/node: Add custom text (un)marshaler for RolesMask type

tjanez:
This will result in easy to understand `"roles"` fields in various CLI commands that output JSON.

For example,  instead of `oasis-node control status` showing:

```
{
  ... trimmed ...
 
  "registration": {
    "last_registration": "2021-09-02T13:25:53+02:00",
    "descriptor": {

      ... trimmed ...

      "roles": 24
    },

  ... trimmed ...
}
```

  it will show:
  
```
{
  ... trimmed ...
 
  "registration": {
    "last_registration": "2021-09-02T13:25:53+02:00",
    "descriptor": {

      ... trimmed ...

      "roles": "validator,consensus-rpc"
    },

  ... trimmed ...
}
```

### DIFF
--- a/go/common/node/node.go
+++ b/go/common/node/node.go
@@ -24,6 +24,11 @@ import (
 )
 
 var (
+	// ErrInvalidRole is the error returned when a node role is invalid.
+	ErrInvalidRole = errors.New("node: invalid role")
+	// ErrDuplicateRole is the error returned when a node role is duplicated.
+	ErrDuplicateRole = errors.New("node: duplicate role")
+
 	// ErrInvalidTEEHardware is the error returned when a TEE hardware
 	// implementation is invalid.
 	ErrInvalidTEEHardware = errors.New("node: invalid TEE implementation")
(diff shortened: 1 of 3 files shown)


kostko (commented):
Should we reject duplicate roles when parsing?

go/common/node/node.go
```diff
@@ -109,6 +112,15 @@ const (
...
 	RoleReserved RolesMask = ((1 << 32) - 1) & ^((RoleStorageRPC << 1) - 1)
+
+	RoleComputeWorkerName = "compute"
+	RoleStorageWorkerName = "storage"
+	RoleKeyManagerName    = "key-manager"
+	RoleValidatorName     = "validator"
+	RoleConsensusRPCName  = "consensus-rpc"
+	RoleStorageRPCName    = "storage-rpc"
+
+	RolesMaskStringSep = ","
```
ptrus, line 123: This should probably be package private (lower case)

go/common/node/node.go
```diff
@@ -109,6 +112,15 @@ const (
 	// RoleReserved are all the bits of the Oasis node roles bitmask
 	// that are reserved and must not be used.
 	RoleReserved RolesMask = ((1 << 32) - 1) & ^((RoleStorageRPC << 1) - 1)
+
+	RoleComputeWorkerName = "compute"
```
ptrus, line 116: Add a doc comment for this variable block.

tjanez: Done.

tjanez:
> Should we reject duplicate roles when parsing?

Agreed. Done.

go/common/node/node.go
tjanez: I don't have strong preferences about this. Just noting that, for example, staking threshold names are also public:
https://github.com/oasisprotocol/oasis-core/blob/d1b69f5aac431eac05705cae0297936ad410a3eb/go/staking/api/api.go#L651-L657

ptrus: Yeah, I ment this only for the separator. I think others should be public.

tjanez: Oh, I didn't understand the comment was only referring to the separator :slightly_smiling_face:.
Yeah, I'll make the separator private. Thanks for catching this!

tjanez: Done.

kostko (approved):

ptrus (approved):

PR #4248, opened 2021-09-08
feature: Add VRF based committee elections

Yawning:
* [x] Bump the curve25519-voi import
 * [x] Implement TupleHash
 * [x] Make common/crypto/signature VRF aware
 * [x] Generate/Load the VRF signing key as part of the node identity
 * [x] Add the VRF signing key to the node descriptor
 * [x] Validate the VRF signing key contained in the node descriptor if present
 * [x] Add back deterministic epochs
 * [x] Consensus support for VRF submissions
 * [x] Nodes submit VRF proofs each epoch
 * [x] Use VRF proofs to select from an eligible entity's validator nodes
 * [x] Use VRF proofs + sortition to do elections
 * [x] Unbreak the tests
 * [x] Unfuck the byzantine node somehow
 * [x] Rip out the old PVSS code
 * [x] Rip out the old deterministic seed based election code.
 * [x] Add the relevant fixgenesis stuff

Testing:
 * [x] Switch the e2e tests to exercise the new VRF code
 * [x] Work around #4374
 * [x] Add byzantine VRF tests (not much to do here)

Documentation:
 * [x] Update/edit the ADR to match reality.

Post upgrade steps for the next breaking upgrade:
 * [ ] Remove the backward compatibility PVSS raw CBOR message field in the node descriptor.
 * [x] Enforce that a VRF signing key is contained in the node descriptor


go/beacon/api/vrf.go
```diff
@@ -0,0 +1,111 @@
...
+
+	Pi []byte `json:"pi"`
+}
+
+// MethodMetadata implements transaction.MethodMetadataProvider.
+func (txBody *VRFProve) MethodMetadata() transaction.MethodMetadata {
+	return transaction.MethodMetadata{
+		// XXX: This is critical but any node can submit one, is
+		// this safe to use?
+		Priority: transaction.MethodPriorityCritical,
```
kostko, line 69: If anyone can submit these then this would be unsafe. Critical priority should only be used for things that only proposers can submit.

go/staking/api/slashing.go
```diff
@@ -32,10 +32,13 @@ const (
 	// SlashConsensusEquivocationName is the string representation of SlashConsensusEquivocation.
 	SlashConsensusEquivocationName = "consensus-equivocation"
 	// SlashBeaconInvalidCommitName is the string representation of SlashBeaconInvalidCommit.
+	// Deprecated: This was for the PVSS beacon.
```
kostko, line 35: Probably we should update fixgenesis to remove these? And then we can also remove them here.

kostko (commented):
Initial review, still need to look at some stuff.

go/consensus/tendermint/apps/beacon/backend_vrf.go
```diff
@@ -0,0 +1,488 @@
...
+		return fmt.Errorf("beacon: failed to deserialize raw proof: %w", err)
+	}
+	ok, beta := proof.Verify(vrfState.Alpha)
+	if !ok {
+		return fmt.Errorf("beacon: failed to verify beta")
+	}
+
+	if oldPi := vrfState.Pi[node.ID]; oldPi == nil {
+		// Fresh proof, store pi.
+		vrfState.Pi[node.ID] = &proof
```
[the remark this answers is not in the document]
Yawning: Per discussion, we are leaving this as is for now.

go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
```diff
@@ -252,6 +348,39 @@ func computeStakeClaims(
...
+	if oldP.DebugMockBackend || oldP.InsecureParameters != nil || oldP.DebugDeterministic {
+		return beacon.Genesis{}, fmt.Errorf("existing document has debug/insecure beacon parameters set")
+	}
+	if oldP.Backend != "pvss" {
+		return beacon.Genesis{}, fmt.Errorf("existing document is not using the PVSS beacon")
+	}
+
+	// Derive "reasonable" parameters based off the existing parameters.
+	pvssP := oldP.PVSSParameters
+	alphaHighQualityThreshold := uint64(pvssP.Participants)
```
kostko, line 364: Is this sufficient or should it be increased?

Yawning: It's a tinfoil hattery parameter.  In theory, I am being over-cautious because there is no way for a single node to derive the other node's beta values (ie: the best that a single node can do even if granted control over alpha is to try to game it so that their beta gives them a statistical advantage in the sortition).

go/consensus/tendermint/apps/beacon/backend_vrf.go
```diff
@@ -0,0 +1,500 @@
...
+		// The beta is checked instead of a byte-for-byte comparison of Pi
+		// because proofs are not guaranteed to be deterministic (though
+		// the IETF draft happens to produce proofs that are).
+		oldBeta := oldPi.UnsafeToHash()
+		if !bytes.Equal(oldBeta, beta) {
+			return fmt.Errorf("beacon: node attempted to submit a different proof")
+		}
+	}
+
+	if ctx.IsCheckOnly() {
```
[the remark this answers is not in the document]
Yawning:
```
cpu: AMD Ryzen 5 5600X 6-Core Processor             
BenchmarkECVRF/Prove-12         	   15902	     76320 ns/op	    1078 B/op	      15 allocs/op
BenchmarkECVRF/ProofToHash-12   	  159141	      7691 ns/op	     288 B/op	       4 allocs/op
BenchmarkECVRF/Verify-12        	   13803	     85597 ns/op	    4486 B/op	      20 allocs/op
PASS
```

A bit more allocation heavy than Ed25519 verification, but "dirt cheap".

go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
kostko: Ok, makes sense, thanks!

go/consensus/tendermint/apps/beacon/backend_vrf.go
```diff
@@ -0,0 +1,488 @@
...
+
+	if oldPi := vrfState.Pi[node.ID]; oldPi == nil {
+		// Fresh proof, store pi.
+		vrfState.Pi[node.ID] = &proof
+		if err = state.SetVRFState(ctx, vrfState); err != nil {
+			return fmt.Errorf("beacon: failed to update state: %w", err)
+		}
+	} else {
+		// That's odd, the node already submitted a proof, ensure that the
+		// betas match, and if not, reject and consider slashing.
```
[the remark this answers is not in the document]
Yawning: The failure case is impossible assuming the VRF is specified/implemented correctly since the beta output is deterministic.  I could have the byzantine node re-submit the tx with the same proof though.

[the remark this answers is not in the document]
Yawning: This actually is kind of hard to test, because when using mock timekeeping the backend will immediately transition the epoch once it receives all outstanding VRF proofs, if a transition is queued.  Since I do not want to mess with timekeeping for test cases any more (and not having this workaround makes things extremely slow), I am going to punt on this for now because the logic is trivially correct.

kostko (approved):

go/staking/api/slashing.go
Yawning: Done.

PR #4250, opened 2021-09-09
runtime: Verify consensus layer state by using a light client

### LINKED ISSUE #3952: Runtime light client consensus verification
Now that we have all the infrastructure available, the runtime (when running outside the node's trust domain, e.g. as an enclave) should verify light client proofs of consensus layer state before proceeding to process runtime transactions. This should include the runtime confirming that its RAK/nonce is indeed registered in the consensus layer.

This requires the runtime to have a hardcoded trust anchor to the consensus layer which is probably a combination of the following:
* Consensus layer chain context.
* Trusted hash of a consensus layer header at a specific height.

One of the questions to explore in the design is how would the runtime handle a breaking consensus layer upgrade where the chain context changes.

kostko:
Fixes #3952 

TODO
* [x] Clean up E2E tests.
* [x] Cache last computed state root to avoid verification when the enclave itself knows it is correct (since it computed it locally).
* [x] Periodically notify runtime of new consensus layer blocks to reduce the amount of syncing needed on ExecuteTx.
* [x] Write a LRU light block store to avoid keeping all the blocks.
* [x] Persist sealed latest trusted consensus block in untrusted local storage.

### DIFF
--- a/runtime/src/consensus/tendermint/verifier.rs
+++ b/runtime/src/consensus/tendermint/verifier.rs
@@ -1,0 +1,699 @@
+//! Tendermint consensus layer verification logic.
+use std::{
+    collections::HashSet,
+    convert::{TryFrom, TryInto},
+    str::FromStr,
+    sync::Arc,
+    time::Duration,
+};
+
+use anyhow::anyhow;
+use crossbeam::channel;
+use io_context::Context;
+use sgx_isa::Keypolicy;
+use slog::{error, info};
+use tendermint::{
+    block::CommitSig,
+    vote::{SignedVote, ValidatorIndex, Vote},
+};
+use tendermint_light_client::{
+    builder::LightClientBuilder,
+    components::{self, verifier::ProdVerifier},
+    light_client,
+    operations::{ProdCommitValidator, ProdHasher, VotingPowerCalculator, VotingPowerTally},
+    predicates::{errors::VerificationError, ProdPredicates},
+    supervisor::Instance,
+    types::{
+        Commit, Hash as TMHash, LightBlock as TMLightBlock, PeerId, SignedHeader, Time,
+        TrustThreshold, ValidatorSet,
+    },
+};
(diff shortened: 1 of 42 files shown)


runtime/src/consensus/tendermint/verifier.rs
```diff
@@ -0,0 +1,686 @@
...
+        Ok(())
+    }
+
+    /// Start the verifier in a separate thread.
+    pub fn start(self) {
+        std::thread::spawn(move || {
+            let logger = get_logger("consensus/tendermint/verifier");
+            info!(logger, "Starting consensus verifier");
+
+            for retry in 1..=MAX_INITIALIZATION_RETRIES {
```
ptrus, line 294: Can you add a comment on why is this retrying needed? Should the thread `JoinHandle` be returned from the `start` method, so that the caller can catch this failing?

ptrus (approved):

runtime/src/consensus/tendermint/verifier.rs
kostko: Retry is needed because `start()` is called when processing the `RuntimeInfoRequest` which is part of the RHP initialization. The host will reject any requests made by the runtime until `RuntimeInfoRequest` is processed successfuly, which leads to this sometimes starting too quickly. I will add a comment.

The problem with a join handle is that it is not obvious where to block for the thread to finish as the protocol thread must not do that. Note that if this fails, then all further verifications will fail (due to the channel receiver being dropped) which will cause the dispatcher to panic and the runtime to abort.

kostko: Will also limit the restarts to specific error variants as once we're past creating an instance any failures should immediately be fatal.

kostko: Done.

ptrus: Thanks!

PR #4268, opened 2021-09-17
go/common/grpc: Configure max message size on the client

kostko:


nhynes:
where did the 100 -> 10 MiB change go?

kostko:
> where did the 100 -> 10 MiB change go?

Let's consider that separately.

ptrus (approved):

PR #4299, opened 2021-10-07
go/common: use entire error as error message

### LINKED ISSUE #4298: go: our own errors get sent out over grpc with blank message
##### SUMMARY

`errors.Code` seems to swallow the message when there's no context. when using a client other than our Go codebase, there's no facility to look up the module and code to get the message. in some cases such as through grpc-web, missing library support makes it impossible to get the module and code in the first place, and all you have is the message.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

go/common/errors

##### OASIS NODE VERSION

```paste below
21.3
```

##### OS / ENVIRONMENT

n/a

##### STEPS TO REPRODUCE

Trigger an error such as submitting a `staking.Allow` transaction when allowances are disabled (max allowances = 0).

##### ACTUAL RESULTS

GRPC error message blank

##### EXPECTED RESULTS

GRPC error message `staking: forbidden by policy`

pro-wh:
we had this design where in places that accepted an error message, we would use an empty string. or if it were a "WithContext" error, we'd use only the context. this change makes it so that we use the error message for the error message. I'm doing this to make Oasis software friendlier than "transaction submission failed: ."

also, there was this weird bug where a non-registered error would get repeated like `insufficient balance: insufficient balance` (`quantity.ErrInsufficientBalance`). that was due to some routines putting the message in both the message and the context in the non-registered error case. I'm fixing that to put it in the message only.

fixes #4298

### DIFF
--- a/.changelog/4298.feature.md
+++ b/.changelog/4298.feature.md
@@ -1,0 +1,1 @@
+go/common: use entire error as error message


kostko:
I see some tests need to be fixed to use `errors.Is`.

PR #4308, opened 2021-10-11
Remove separate storage nodes

### LINKED ISSUE #3905: Remove storage receipts from runtime genesis
While closing #1686 (in #2875) we forgot to remove some sanity checks from runtime genesis descriptor. We should actually just remove the `StorageReceipts` field as it makes no sense (as noted in https://github.com/oasisprotocol/oasis-core/issues/1686#issuecomment-566967712).

kostko:
This removes the separate storage committee and makes all compute nodes always imply storage as well. A notion of runtime mode is introduced which can be one of none, compute, keymanager, client and client-stateless. Storage receipts are removed and executor commitments are used as storage availability commitments as well. An additional Byzantine scenario is added where a faulty proposer generates bogus batches by including transactions that nobody else has.

The structure of proposals is changed so that executor workers don't need to download transactions again, but instead use the existing P2P transaction dispatch mechanism.

Also fixes #3905.

**Note that this change is consensus-breaking since it changes the committee scheduler and registry.**

TODO
* [x] Fix broken unit tests.
* [x] Reinstate unavailable storage failure code
* [x] Refactor executor commits such that roothash evidence does not need to include submitted messages.
* [x] Refactor signature scheme for proposals and executor commits.
* [x] ~~Explicit transaction requests for peers in case of missing transaction hashes.~~
* [x] Cleanup and further simplifications.
* [x] Think about more Byzantine scenarios.
* [x] Support for converting genesis dump from consensus v4.
* [x] Long-term tests.
* [x] Changelog entries.


go/storage/api/metrics.go
```diff
@@ -157,6 +115,14 @@ func (w *metricsWrapper) SyncIterate(ctx context.Context, request *IterateReques
 	return res, err
 }
 
+func (w *metricsWrapper) Apply(ctx context.Context, request *ApplyRequest) error {
```
Yawning, line 118: Do we no longer want metrics on this, even for local performance reasons?

go/worker/compute/executor/committee/node.go
```diff
@@ -792,7 +756,19 @@ func (n *Node) proposeTimeoutLocked(roundCtx context.Context) error {
 
 		// Make sure we are still in the right state/round.
 		n.commonNode.CrossNode.Lock()
-		if _, ok := n.state.(StateWaitingForBatch); !ok || round != n.commonNode.CurrentBlock.Header.Round {
+		// Make sure we are still in the right state.
+		var invalidState bool
+		switch n.state.(type) {
+		case StateWaitingForBatch:
```
Yawning, line 762: Merge these two cases (style), unless you think there will be a need to split them or add explicit coments.

Yawning (commented):
There appears to be some terminology inconsistency between compute and executor.  Should this be unified to the latter?

go/storage/api/metrics.go
kostko: Ooops, yeah not sure how this happened, added it back.

kostko:
> There appears to be some terminology inconsistency between compute and executor. Should this be unified to the latter?

Previously compute was a "super role" that a node would register for and the scheduler would then elect into subroles like executor, txnscheduler, merge but with the changed architecture compute basically equals executor (merge has been gone for a while and txnscheduler now rotates within the executor committee).

go/worker/compute/executor/committee/node.go
```diff
@@ -1225,8 +1205,27 @@ func (n *Node) startProcessingBatchLocked(batch *unresolvedBatch) {
...
+	if err != nil {
+		n.logger.Error("refusing to process bad batch", "err", err)
+		// TODO: We should indicate failure.
+		return
+	}
+	if resolvedBatch == nil {
+		// Some transactions are missing so we cannot start processing the batch just yet.
+		// Transition into StateWaitingForTxs and request transactions from peers.
+		n.logger.Debug("some transactions are missing", "num_missing", len(batch.missingTxs))
+		// TODO: Actually request transactions from peers.
```
Yawning, line 1222: Is this a "TODO when we get around to it", or a "TODO before merge"?

Yawning:
A gentle reminder to please triage the existing mountain of backlogged storage related issues after this is merged and update/close them as appropriate.

go/runtime/scheduling/simple/txpool/priorityqueue/priority_queue.go
```diff
@@ -149,6 +149,24 @@ func (q *priorityQueue) GetBatch(force bool) []*transaction.CheckedTransaction {
 	return batch
 }
 
+// Implements api.TxPool.
+func (q *priorityQueue) GetKnownBatch(batch []hash.Hash) ([]*transaction.CheckedTransaction, map[hash.Hash]int) {
+	q.Lock()
+	defer q.Unlock()
+
+	var result []*transaction.CheckedTransaction
```
ptrus, line 157: `result` could be pre-allocated since the size is known to be `len(batch)`

ptrus (commented):
Quickly checked, will review more

kostko:
I've pushed some updates that move the runtime client into its own worker so that it can reuse code (block processing loop, transaction pool) from the common worker.

go/runtime/scheduling/simple/txpool/priorityqueue/priority_queue.go
kostko: Done.

go/runtime/scheduling/simple/txpool/api/api.go
```diff
@@ -33,8 +33,20 @@ type TxPool interface {
...
 	GetBatch(force bool) []*transaction.CheckedTransaction
 
+	// GetKnownBatch gets a set of known transactions from the transaction pool.
+	//
+	// For any missing transactions nil will be returned in their place and the map of missing
+	// transactions will be populated accoordingly.
+	GetKnownBatch(batch []hash.Hash) ([]*transaction.CheckedTransaction, map[hash.Hash]int)
+
+	// GetTransactions returns the given number of transactions from the transaction pool without
+	// taking any batch limits into account.
```
ptrus, line 43: Maybe mention that this also ignores priority (weights), right?

go/runtime/scheduling/api/api.go
```diff
@@ -15,11 +15,23 @@ type Scheduler interface {
...
 	GetBatch(force bool) []*transaction.CheckedTransaction
 
+	// GetKnownBatch gets a set of known transactions from the transaction pool.
+	//
+	// For any missing transactions nil will be returned in their place and the map of missing
+	// transactions will be populated accoordingly.
+	GetKnownBatch(batch []hash.Hash) ([]*transaction.CheckedTransaction, map[hash.Hash]int)
+
+	// GetTransactions returns the given number of transactions from the transaction pool without
+	// taking any batch limits into account.
```
ptrus, line 30: Maybe mention that this also ignores priority (weights), right?

go/runtime/scheduling/simple/txpool/api/api.go
kostko: Good idea, added.

go/runtime/client/api/api.go
```diff
@@ -83,10 +82,9 @@ type RuntimeClient interface {
 	WatchBlocks(ctx context.Context, runtimeID common.Namespace) (<-chan *roothash.AnnotatedBlock, pubsub.ClosableSubscription, error)
 }
 
-// RuntimeClientService is the runtime client service interface.
-type RuntimeClientService interface {
-	RuntimeClient
-	service.BackgroundService
+type SubmitTxResult struct {
```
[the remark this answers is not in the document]
kostko: Added.

kostko:
I'm running some long-term tests to see if that finds any issues.

go/worker/compute/executor/committee/p2p.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+		return nil
+	default:
+		return p2pError.ErrUnhandledMessage
+	}
+}
+
+// HandlePeerTx implements NodeHooks.
+func (n *Node) HandlePeerTx(ctx context.Context, tx []byte) error {
+	if err := n.commonNode.TxPool.SubmitTxNoWait(ctx, tx, &txpool.TransactionMeta{Local: false}); err != nil {
```
Yawning, line 98: Since you aren't logging just return directly instead of having an `if`?

go/runtime/txpool/check_queue.go
```diff
@@ -0,0 +1,155 @@
...
+
+// RemoveBatch removes a batch of transactions from the queue.
+func (q *checkTxQueue) RemoveBatch(batch []*pendingTx) {
+	q.Lock()
+	defer q.Unlock()
+
+	for _, item := range batch {
+		if pair, ok := q.transactions[item.TxHash]; ok {
+			q.queue.Remove(pair.element)
+			delete(q.transactions, item.TxHash)
```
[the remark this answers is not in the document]
kostko: We don't, but yeah clearing it here makes sense.

go/runtime/txpool/txpool.go
```diff
@@ -0,0 +1,808 @@
...
+
+				// Sine publication failed, make sure we retry early.
+				t.republishCh.In() <- struct{}{}
+				publishTime = time.Time{}
+			}
+		}
+
+		if err := t.seenCache.Put(tx.Hash(), publishTime); err != nil {
+			// cache.Put can only error if capacity in bytes is used and the
+			// inserted value is too large. This should never happen in here.
```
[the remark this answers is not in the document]
kostko: Yeah this error just needs to be ignored (as it can actually never happen because of the LRU cache configuration). Updated the comment and removed logging.

go/runtime/txpool/txpool.go
```diff
@@ -0,0 +1,808 @@
...
+			delete(t.roundWeightLimits, w)
+		}
+	}
+	// Update batch weight limits.
+	for w, l := range limits {
+		t.roundWeightLimits[w] = l
+	}
+
+	if err := t.scheduler.UpdateParameters(t.scheduler.Name(), t.roundWeightLimits); err != nil {
+		return fmt.Errorf("updating scheduler parameters: %w", err)
```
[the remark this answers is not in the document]
kostko: This cannot actually fail, updated the API.

go/runtime/txpool/txpool.go
```diff
@@ -0,0 +1,808 @@
...
+			)
+			return fmt.Errorf("transaction scheduler algorithm update not supported")
+		}
+
+		// TODO: Remove the extra scheduler algorithm parameter.
+		if err := t.scheduler.UpdateParameters(t.scheduler.Name(), t.roundWeightLimits); err != nil {
+			t.logger.Error("error updating transaction scheduler parameters",
+				"err", err,
+			)
+			return fmt.Errorf("failed to update transaction scheduler parameters: %w", err)
```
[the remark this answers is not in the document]
kostko: This cannot actually fail, updated the API.

kostko:
Multiple runs of long-term tests seem to pass now, running one more on the latest commit.

kostko:
Latest long-term test run has been successful.

Yawning (approved):

PR #4327, opened 2021-10-26
ADR 0011: Incoming Runtime Messages

kostko:
[Rendered](https://github.com/oasisprotocol/oasis-core/blob/kostko/docs/adr-rt-incoming-msgs/docs/adr/0011-incoming-runtime-messages.md)

### DIFF
--- a/.changelog/4327.doc.md
+++ b/.changelog/4327.doc.md
@@ -1,0 +1,1 @@
+ADR 0011: Incoming Runtime Messages
--- a/docs/adr/0011-incoming-runtime-messages.md
+++ b/docs/adr/0011-incoming-runtime-messages.md
@@ -1,0 +1,390 @@
+# ADR 0011: Incoming Runtime Messages
+
+## Changelog
+
+- 2021-12-09: Introduce an explicit fee field, clarify token transfers
+- 2021-10-26: Initial draft
+
+## Status
+
+Accepted
+
+## Context
+
+There is currently a single mechanism through which the consensus layer and a
+runtime may interact in a consistent and secure manner. This is the mechanism
+of runtime messages that can be emitted by runtimes (see [ADR 3]) and allows
+the consensus layer to act on a runtime's behalf. This mechanism is currently
+used for _pulling_ tokens from consensus layer accounts that have previously
+set proper allowances and for updating the runtime descriptor when the runtime
+governance model (see [ADR 4]) is in effect.
+
+This ADR proposes to implement the reverse mechanism where anyone issuing a
+transaction at the consensus layer can queue arbitrary messages for processing
+by the runtime in its next round.
+
+[ADR 3]: 0003-consensus-runtime-token-transfer.md
+[ADR 4]: 0004-runtime-governance.md
+
+## Decision
+
+On a high level this proposal affects the following components:
+
+- A new transaction method `roothash.SubmitMsg` is added to the roothash
+  consensus service to queue a new message for the specific runtime.
+
+- Additional per-runtime state is added to the roothash service containing the
+  currently queued messages, sorted by arrival time.
+
+- During processing of a round the proposer may propose to pop any number of
+  messages and process them by pushing them to the runtime, similar as it does
+  for transaction batches. This is of course subject to discrepancy detection.
+
+- The runtime host protocol is updated to allow the host to push arbitrary
+  incoming messages in addition to the transaction batch.
+
+- The runtime descriptor is updated to include a field that specifies the
+  maximum size of the incoming message queue.
+
+### Incoming Message
+
+Each incoming message is represented as follows:
+
+```golang
+type IncomingMessage struct {
+    // ID is the unique identifier of the message.
+    ID uint64 `json:"id"`
+
+    // Caller is the address of the caller authenticated by the consensus layer.
+    Caller staking.Address `json:"caller"`
+
+    // Fee is the fee sent into the runtime as part of the message being sent.
+    // The fee is transferred before the message is processed by the runtime.
+    Fee quantity.Quantity `json:"fee,omitempty"`
+
+    // Tokens are any tokens sent into the runtime as part of the message being
+    // sent. The tokens are transferred before the message is processed by the
+    // runtime.
+    Tokens quantity.Quantity `json:"tokens,omitempty"`
+
+    // Data is arbitrary runtime-dependent data.
+    Data []byte `json:"data,omitempty"`
+}
+```
+
+### Executor Commitments
+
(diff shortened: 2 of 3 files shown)


docs/adr/0010-incoming-runtime-messages.md
```diff
@@ -0,0 +1,329 @@
...
+The transaction signer implicitly specifies the caller. Upon executing the
+submit message method the following actions are performed:
+
+- Gas is accounted for (new `submitmsg` gas operation).
+
+- The runtime descriptor for runtime `id` is retrieved. If the runtime does not
+  exist or is currently suspended the method fails with `ErrInvalidRuntime`.
+
+- The `executor.max_in_messages` field in the runtime descriptor is checked. If
+  it is equal to zero the method fails with `ErrIncomingMessageQueueFull`.
```
bennetyee, line 192: what is the gas cost of queuing a message?  the `max_in_messages` limit excessive queuing, but can be a DOS opportunity.  what can/should we do to prevent one runtime DOS'ing another?  enforce a min gas cost but allow recipient to increase? increase cost as queue gets longer?

docs/adr/0010-incoming-runtime-messages.md
```diff
@@ -0,0 +1,329 @@
...
+
+- The incoming queue metadata structure is fetched. If it doesn't yet exist it
+  is populated with zero values.
+
+- If the value of the `size` field in the metadata structure is equal to or
+  larger than the value of the `executor.max_in_messages` field in the runtime
+  descriptor the method fails with `ErrIncomingMessageQueueFull`.
```
bennetyee, line 199: what is the gas cost of queuing a message?  the `max_in_messages` limit excessive queuing, but can be a DOS opportunity.  what can/should we do to prevent one runtime DOS'ing another?  enforce a min gas cost but allow recipient to increase? increase cost as queue gets longer?

kostko: Currently the proposal has a fixed gas cost (see a few steps above) for queuing a message. But hm maybe it would make sense to dynamically increase the cost based on queue size.

kostko:
Any other thoughts about this ADR?

docs/adr/0010-incoming-runtime-messages.md
```diff
@@ -0,0 +1,329 @@
...
+are queried at the start of each round and are then passed to the runtime via
+the runtime host protocol.
+
+### Runtime Processing
+
+The proposal requires that messages are processed by the runtime in queue order
+(e.g. on each round `InMessagesCount` messages are poped from the queue). This
+simplifies the design but the runtimes need to carefully consider how much
+resources to allocate for executing messages (vs. regular transactions) in a
+round.
```
pro-wh, line 273: partially related to Bennet's idea of charging more gas for deeply queued messages, could we run into a situation where there are a lot of 'cheap' incoming messages that a runtime node perhaps prefers not to process, burying a more valuable incoming message?

incoming messages are arbitrary, and the consensus layer won't even check that they're meaningful. all we know is that the submitter used gas to enqueue it, but that gas fee would only go to the consensus nodes.

docs/adr/0010-incoming-runtime-messages.md
```diff
@@ -0,0 +1,329 @@
...
+
+```golang
+type IncomingMessage struct {
+    // Caller is the address of the caller authenticated by the consensus layer.
+    Caller staking.Address `json:"caller"`
+
+    // Tokens are any tokens sent into the runtime as part of the message being
+    // sent. The tokens are transferred before the message is processed by the
+    // runtime.
+    Tokens quantity.Quantity `json:"tokens,omitempty"`
```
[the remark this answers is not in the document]
kostko: So if the message is queued the amount is already in the runtime's account. Oh I see this is missing from the `roothash.SubmitMsg` processing description, will update.

docs/adr/0010-incoming-runtime-messages.md
```diff
@@ -0,0 +1,329 @@
...
+        in_msgs: Vec<IncomingMessage>, // Added argument.
+    ) -> Result<ExecuteBatchResult, RuntimeError>;
+}
+```
+
+### Executor Processing
+
+The executor processing pipeline is changed such that pending incoming messages
+are queried at the start of each round and are then passed to the runtime via
+the runtime host protocol.
```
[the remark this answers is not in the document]
kostko: We already query for round results which contains emitted message results and sets of good/bad compute entities.

docs/adr/0010-incoming-runtime-messages.md
kostko: The problem is with "a runtime node perhaps prefers not to process". Making it possible to cherry pick messages could lead to a situation where some messages are left in the queue forever. The runtime (e.g., not the node) can, however, decide to fail the message with an error instead of processing it (for example if it would exceed the resource limits or if the implied fee is too low).

Of course it needs to at least decode the message which incurs some costs.

We could also explicitly separate sent tokens into fee and tokens and the runtime could configure a min per-message fee (of course it would need to be independent of the actual message as the consensus layer should not try to decode messages).

Note that sending incoming messages should be more expensive than directly submitting transactions in the runtime.

pro-wh: ah yes, message fee > transaction fee should be enough to prevent spam

pro-wh (approved):

docs/adr/0010-incoming-runtime-messages.md
```diff
@@ -0,0 +1,329 @@
...
+  is populated with zero values.
+
+- If the value of the `size` field in the metadata structure is equal to or
+  larger than the value of the `executor.max_in_messages` field in the runtime
+  descriptor the method fails with `ErrIncomingMessageQueueFull`.
+
+- An `IncomingMessage` structure is generated based on the caller and method
+  body and the value of the `next_sequence_number` metadata field is used to
+  generate a proper key for storing it in the queue. The structure is inserted
+  into the queue.
```
Yawning, line 204: The queue is FIFO, and this work is being done by the tx processing logic right?  So the queue is always sorted, always ascending, and instead of doing an insert (walk the queue to find the appropriate insertion point), it always will be append?

The insertion not being an append is an invariant violation right?

Yawning (commented):
Apart from the DOS concerns that others have mentioned, I don't have any major issues with this.

docs/adr/0010-incoming-runtime-messages.md
pro-wh: yes to FIFO and append-only in this design

PR #4344, opened 2021-11-03
go/consensus: Add support for showing next block votes

### LINKED ISSUE #4241: Add consensus vote status subcommand
We should add a subcommand (and a corresponding gRPC API endpoint) that displays the current vote status of the consensus layer.

kostko:
Fixes #4241


kostko:
Included in #4400.

PR #4354, opened 2021-11-11
go/runtime/host/sgx: Autodetect SGX device name

### LINKED ISSUE #4333: Discover SGX driver device location
It seems that some later drivers use `/dev/sgx` instead of `/dev/isgx` but currently Oasis Core hardcodes the path as `/dev/isgx` so people need to create symlinks.

There should be some driver autodetection logic which would bind the supported device into the sandbox.

abukosek:
Closes #4333.

### DIFF
--- a/go/runtime/host/sgx/sgx.go
+++ b/go/runtime/host/sgx/sgx.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


go/runtime/host/sgx/sgx.go
```diff
@@ -145,6 +146,23 @@ func (s *sgxProvisioner) loadEnclaveBinaries(rtCfg host.Config) ([]byte, []byte,
 	return sgxs, sig, nil
 }
 
+func (s *sgxProvisioner) discoverSGXDevice() (string, error) {
+	// Different versions of Intel SGX drivers provide different names for
+	// the SGX device.  Autodetect which one actually exists.
+	sgxDevices := []string{"/dev/sgx", "/dev/isgx"}
```
kostko, line 152: There's also a `/dev/sgx/enclave` variant I think.

abukosek: Thanks, added!

go/runtime/host/sgx/sgx.go
```diff
@@ -145,6 +146,23 @@ func (s *sgxProvisioner) loadEnclaveBinaries(rtCfg host.Config) ([]byte, []byte,
 	return sgxs, sig, nil
 }
 
+func (s *sgxProvisioner) discoverSGXDevice() (string, error) {
+	// Different versions of Intel SGX drivers provide different names for
+	// the SGX device.  Autodetect which one actually exists.
+	sgxDevices := []string{"/dev/sgx", "/dev/sgx/enclave", "/dev/isgx"}
```
kostko, line 152: And maybe also `/dev/sgx_enclave` (see https://github.com/fortanix/rust-sgx/blob/master/intel-sgx/sgxs-loaders/src/isgx/mod.rs#L444).

kostko (approved):
Also please backport to the `stable/21.3.x` branch.

PR #4357, opened 2021-11-12
test-runner: Read runtime version from fixture

### LINKED ISSUE #4343: oasis-net-runner: support setting runtime registry descriptor version
Currently the [`oasis-net-runner.RuntimeCfg`](https://github.com/oasisprotocol/oasis-core/blob/a4ad6be7ec516485b0a2acd005c8f38b28ba0ff2/go/oasis-test-runner/oasis/runtime.go#L48-L74) doesn't support the [`Version`](https://github.com/oasisprotocol/oasis-core/blob/aa0e4ef63defdb1ceb3f2f32372d0bd156cae90d/go/registry/api/runtime.go#L348-L349) field. Therefore the `oasis-net-runner` cannot be used with oasis-sdk runtimes of versions `1.0.0` or above, as the major vesion in the registry descriptor needs to match the runtime version.

Similarly the [`RuntimeFixture`](https://github.com/oasisprotocol/oasis-core/blob/973cf4864d9ee127ec7d452760424f73df862451/go/oasis-test-runner/oasis/fixture.go#L223-L247)  should have the `Version` field added so that the runtime version can be defined in a fixture.

matevz:
Fixes #4343:
- adds `version` field of type `version.Version` to runtime fixture and config structs
- fixes converting versions from `uint64`, if component values are greater than 255

### DIFF
--- a/go/common/version/version.go
+++ b/go/common/version/version.go
@@ -47,15 +47,50 @@ func (v Version) ToU64() uint64 {
 // FromU64 returns the version from platform-dependent uint64.
 func FromU64(v uint64) Version {
 	return Version{
-		Major: uint16((v >> 32) & 0xff),
-		Minor: uint16((v >> 16) & 0xff),
-		Patch: uint16(v & 0xff),
+		Major: uint16((v >> 32) & 0xffff),
+		Minor: uint16((v >> 16) & 0xffff),
+		Patch: uint16(v & 0xffff),
 	}
 }
 
 // String returns the protocol version as a string.
 func (v Version) String() string {
 	return fmt.Sprintf("%d.%d.%d", v.Major, v.Minor, v.Patch)
+}
+
+// FromString parses version in semver format. e.g. "1.0.0"
+func FromString(s string) (Version, error) {
+	// Trim potential pre-release suffix.
+	s = strings.Split(s, "-")[0]
+	// Trim potential git commit.
+	s = strings.Split(s, "+")[0]
+	split := strings.SplitN(s, ".", 4)
+	if len(split) != 3 {
+		return Version{}, fmt.Errorf("version: failed to parse SemVer '%s': exactly three components are required", s)
+	}
+
+	var semVers []uint16 = []uint16{0, 0, 0}
+	for i, v := range split {
+		if i >= 3 {
+			break
+		}
+		ver, err := strconv.ParseUint(v, 10, 16)
+		if err != nil {
+			return Version{}, fmt.Errorf("version: failed to parse SemVer '%s': %w", s, err)
+		}
+		semVers[i] = uint16(ver)
+	}
+
+	return Version{Major: semVers[0], Minor: semVers[1], Patch: semVers[2]}, nil
+}
+
+// MustFromString parses version in semver format and panics, if there is an error.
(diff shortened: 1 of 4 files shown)


go/oasis-test-runner/oasis/fixture.go
```diff
@@ -225,6 +225,7 @@ type RuntimeFixture struct { // nolint: maligned
 	Kind       registry.RuntimeKind `json:"kind"`
 	Entity     int                  `json:"entity"`
 	Keymanager int                  `json:"keymanager"`
+	Version    uint64               `json:"version"`
```
kostko, line 228: Would be better if this was either a `version.Version` or a string which gets parsed into a `version.Version`.

matevz: done.

go/common/version/version.go
```diff
@@ -47,9 +47,9 @@ func (v Version) ToU64() uint64 {
 // FromU64 returns the version from platform-dependent uint64.
 func FromU64(v uint64) Version {
 	return Version{
-		Major: uint16((v >> 32) & 0xff),
-		Minor: uint16((v >> 16) & 0xff),
-		Patch: uint16(v & 0xff),
+		Major: uint16((v >> 32) & 0xffff),
```
kostko, line 50: Please add unit tests to make sure this conversion works for versions that use full uint16 components (e.g. >255).

go/common/version/version.go
```diff
@@ -111,7 +137,7 @@ var (
 	TendermintAppVersion = ConsensusProtocol.MaskNonMajor().ToU64()
 
 	// Toolchain is the version of the Go compiler/standard library.
-	Toolchain = parseSemVerStr(strings.TrimPrefix(runtime.Version(), "go"))
+	Toolchain, _ = FromString(strings.TrimPrefix(runtime.Version(), "go"))
```
kostko, line 140: This will hide errors. It should panic instead on error.

go/oasis-test-runner/oasis/runtime.go
```diff
@@ -36,6 +37,7 @@ type Runtime struct { // nolint: maligned
 	teeHardware node.TEEHardware
 	mrEnclaves  []*sgx.MrEnclave
 	mrSigner    *sgx.MrSigner
+	version     uint64
```
kostko, line 40: Where is this used?

go/oasis-test-runner/oasis/runtime.go
```diff
@@ -52,6 +54,7 @@ type RuntimeCfg struct { // nolint: maligned
 	Keymanager  *Runtime
 	TEEHardware node.TEEHardware
 	MrSigner    *sgx.MrSigner
+	Version     string
```
kostko, line 57: This could actually be `version.Version` and the parsing should happen in the fixture's `Create` function.

go/oasis-test-runner/oasis/runtime.go
matevz: Oops, this should go to `RuntimeCfg` struct. Removed.

tjanez (approved):

kostko (approved):

PR #4382, opened 2021-11-30
go/common/identity: Save re-generated sentry client and node's persistent TLS certificate

tjanez:
Sentry client TLS certificate is always re-generated from the private key when the Oasis Node starts.

Previously, the re-generated sentry client TLS certificate was not saved to disk, which caused confusion since the on-disk certificate file (i.e. `sentry_client_tls_identity_cert.pem`) had incorrect/outdated expiry date.

Also, save re-generated node's persistent TLS certificate. If a node's TLS certificate is persistent, it is always re-generated from the private key when the Oasis Node starts.
    
Previously, the re-generated node's persistent TLS certificate was not saved to disk, which caused confusion since the on-disk certificate file (i.e. `tls_identity_cert.pem`) had incorrect/outdated expiry date.

### DIFF
--- a/go/common/identity/identity.go
+++ b/go/common/identity/identity.go
@@ -269,67 +269,71 @@ func doLoadOrGenerate(dataDir string, signerFactory signature.SignerFactory, sho
 	var (
 		nextCert   *tls.Certificate
 		nextSigner signature.Signer
-		dnr        bool
+		dontRotate bool
 	)
 
-	// First, check if we can load the TLS certificate from disk.
+	// Load and re-generate node's persistent TLS certificate (if it exists).
+	// NOTE: This will reuse the node's persistent TLS private key (if it
+	// exists) and re-generate the TLS certificate with a validity of 1 year.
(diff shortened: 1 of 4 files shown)


go/common/identity/identity.go
```diff
@@ -320,24 +320,34 @@ func doLoadOrGenerate(dataDir string, signerFactory signature.SignerFactory, sho
...
-		// Load failed, generate fresh sentry client cert.
+		// Loading sentry client's private key failed, generate a new
+		// private key and the corresponding TLS certificate.
 		sentryClientCert, err = tlsCert.Generate(CommonName)
 		if err != nil {
 			return nil, err
 		}
-		// And save it to disk.
+		// And save them to disk.
 		err = tlsCert.Save(tlsSentryClientCertPath, tlsSentryClientKeyPath, sentryClientCert)
```
kostko, line 341: This should probably be removed now given that it now always happens below?

tjanez: Done.

kostko (approved):

PR #4383, opened 2021-11-30
[BACKPORT/21.3.x]  go/common/identity: Save re-generated sentry client TLS certificate

tjanez:
Backport of #4382.


tjanez:
I don't think a backport is necessary.

PR #4387, opened 2021-12-07
go/worker/storage: Checkpoint correct round on consensus checkpoint

kostko:


go/worker/storage/committee/node.go
```diff
@@ -858,34 +875,97 @@ func (n *Node) consensusCheckpointSyncer() {
 	}
 	defer sub.Close()
 
+	var (
+		versions []uint64
+		blkCh    <-chan *consensus.Block
+		blkSub   pubsub.ClosableSubscription
+	)
 	for {
 		select {
```
Yawning, line 884: This leaks blkCh/blkSub if the first two cases are taken.

kostko: Thanks, fixed!

Yawning (approved):

PR #4392, opened 2021-12-08
ADR-0012: Runtime message results

ptrus:
[Rendered](https://github.com/oasisprotocol/oasis-core/blob/ptrus/adr/runtime-message-results/docs/adr/0012-runtime-message-results.md)

### DIFF
--- a/docs/adr/0012-runtime-message-results.md
+++ b/docs/adr/0012-runtime-message-results.md
@@ -1,0 +1,168 @@
+# ADR 0012: Runtime Message Results
+
+## Changelog
+
+- 2021-12-04: Initial version
+
+- 2021-12-10: Extend the implementation section
+
+## Status
+
+Accepted
+
+## Context
+
+Currently, the results of emitted runtime messages are `MessageEvent`s, which
+only provide information whether the message execution was successful or not.
+For various use-cases additional information about message results would be
+useful.
+
+One of such is supporting staking by runtimes. Currently, a runtime can emit an
+`AddEscrow` message, but is unaware of the actual amount of shares it obtained
+as a result of the added escrow. For some use-cases (e.g. runtime staking user
+deposited funds) this information is crucial for accounting.
+
+Similarly, for `ReclaimEscrow`, the runtime doesn't have the direct information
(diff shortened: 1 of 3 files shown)


docs/adr/0012-runtime-message-results.md
```diff
@@ -0,0 +1,167 @@
...
+
+Implement support for arbitrary result data in `MessageEvent` runtime message
+results.
+
+## Implementation
+
+- Result field is added to `roothash.MessageEvent` struct:
+
+```golang
+// MessageEvent is a runtime message processed event.
```
Yawning, line 48: Are we going to impose a sensible limit on the size of each message, and the number of messages a runtime can publish at once?

kostko: This is already the case with the existing `MaxMessages` configuration option. There is no explicit limit on size of an individual message, but there is an upper limit which is the maximum transaction size. The messages however cannot be arbitrary, they are limited to specific actions (currently staking operations and runtime descriptor updates).

Yawning: Ok.

docs/adr/0012-runtime-message-results.md
```diff
@@ -0,0 +1,168 @@
...
+- `Publish` method of the `MessageDispatcher` interface is updated to include
+  the response:
+
+```golang
+// MessageDispatcher is a message dispatcher interface.
+type MessageDispatcher interface {
+        // Publish publishes a message of a given kind by dispatching to all subscribers.
+        //
+        // In case there are no subscribers ErrNoSubscribers is returned.
+        Publish(ctx *Context, kind, msg interface{}) (interface{}, error)
```
kostko, line 83: Should this instead be `[]interface{}` as there can be multiple subscribers? (Currently however there are not in this specific case used for runtime messages.)

And then the caller can decide what to do with the results (currently it would just take the first one).

ptrus: Yeah makes sense, updated.

kostko (approved):

Issue #4055, opened 2021-06-18
Add acceptable enclave quote statuses to SGX constraints


kostko:
The `CapabilityTEE` verification should support enforcement of enclave quote statuses by extending `sgx.Constraints` to include a field which specifies what is acceptable (could even be strict/non-strict similar to `OASIS_STRICT_AVR_VERIFY` in runtimes). Note that currently the runtimes can enforce this when establishing secure sessions which is sufficient for security purposes, but it would be useful if one could prevent non-compliant nodes from not even being scheduled. Otherwise this will result in those nodes triggering failures.

PR #4396, opened 2021-12-13
runtime: Enclave quote status related cleanups/improvements

Yawning:
*  [x] runtime: Change OASIS_STRICT_AVR_VERIFY to OASIS_UNSAFE_LAX_AVR_VERFIY
  * [x] Add acceptable enclave quote statuses to SGX constraints (#4055)

Yes I had to move the `Constraints` field definition to the node package, because it doesn't belong in our IAS support code, and there will be an import loop otherwise.


go/common/sgx/common.go
```diff
@@ -259,8 +259,10 @@ func (id EnclaveIdentity) String() string {
 	return hex.EncodeToString(id.MrEnclave[:]) + hex.EncodeToString(id.MrSigner[:])
 }
 
+/*
```
kostko, line 262: Just remove?

kostko (approved):

Issue #4212, opened 2021-08-24
Software version in node descriptor


kostko:
It would be useful if nodes would publish their software version in the node descriptor.

Issue #4090, opened 2021-06-28
Lower minimum round timeout


kostko:
Currently the minimum round timeout that can be configured is 5 consensus blocks. It should be changed such that it can be configured as low as 1 block.

PR #4400, opened 2021-12-16
go: Various small improvements from the backlog

Yawning:
* [x] Add consensus vote status subcommand #4241
 * [x] Software version in node descriptor #4212
 * [x] Lower minimum round timeout #4090

Fixes: #4241 #4212 #4090

### DIFF
--- a/go/common/node/node.go
+++ b/go/common/node/node.go
@@ -96,6 +96,9 @@ type Node struct { // nolint: maligned
 
 	// Roles is a bitmask representing the node roles.
 	Roles RolesMask `json:"roles"`
+
+	// SoftwareVersion is the node's oasis-node software version.
+	SoftwareVersion string `json:"software_version,omitempty"`
 }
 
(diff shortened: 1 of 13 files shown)


go/common/node/node.go
```diff
@@ -96,6 +96,9 @@ type Node struct { // nolint: maligned
 
 	// Roles is a bitmask representing the node roles.
 	Roles RolesMask `json:"roles"`
+
+	// SoftwareVersion is the node's oasis-node software version.
+	SoftwareVersion string `json:"software_version,omitempty"`
```
tjanez, line 101: Do we also need to bump the `LatestNodeDescriptorVersion` to the next version?

Yawning: Technically I should have done that when I added the VRF stuff, but, the storage node PR will do it.

tjanez: Ok, understood.

kostko (approved):

PR #4408, opened 2021-12-22
go/roothash: Persist last round results in consensus state

### LINKED ISSUE #4316: Consider persisting last round results in consensus state
Currently last round results require access to historic consensus layer state (e.g. the block at the `LastNormalHeight`) which can prevent executors that use consensus state sync from processing blocks until they observe the next normal block.

kostko:
Fixes #4316 
Based on #4308 

Previously last normal round results were only available as events at the height
where the runtime block was emitted. This caused problems when consensus state
sync was used as that block may not be available. Last rounds results are now
persisted in consensus state so they remain always available.

### DIFF
--- a/.changelog/4316.breaking.md
+++ b/.changelog/4316.breaking.md
@@ -1,0 +1,6 @@
+go/roothash: Persist last round results in consensus state
+
+Previously last normal round results were only available as events at the
+height where the runtime block was emitted. This caused problems when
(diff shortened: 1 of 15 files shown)


Yawning (approved):
This doesn't need fixgenesis handling right?

kostko:
No, the existing Oasis Core already emits the correct genesis dump.

PR #4414, opened 2022-01-03
go/oasis-test-runner: Try using abstract sockets

Yawning:


Yawning:
Not much gain in doing this for now, and this was more a proof of concept than anything.

PR #4428, opened 2022-01-12
go/governance: Replace quorum/threshold with stakeThreshold

### LINKED ISSUE #4309: Add new consensus governance voting mechanism specifying % of total stake needing to vote yes
##### SUMMARY
Currently, there are two [parameters controling consensus governance voting](https://docs.oasis.dev/oasis-core/high-level-components/index/services/governance#consensus-parameters):
- `quorum` (uint8: [0,100]) specifies the minimum percentage of voting power that needs to be cast on a proposal for the result to be valid.
- `threshold` (uint8: [0,100]) specifies the minimum percentage of `VoteYes` votes in order for a proposal to be accepted.

Consider replacing both of those parameters with a single parameter (i.e. a new threshold) that would specify % of total stake (i.e. voting power) that needs to cast `VoteYes` votes in order for a proposal to be accepted.

##### ISSUE TYPE
- Feature Idea

Yawning:
This replaces the old quorum + threshold based governance proposal
verification with a single unified threshold that the precentage of
yes votes in terms of total voting power (greater than or equal to).

The fixgenesis conversion logic will convert the existing parameters
to a stake threshold of 68%.

Fixes #4309.

### DIFF
--- a/go/governance/api/api.go
+++ b/go/governance/api/api.go
@@ -243,13 +243,10 @@ type ConsensusParameters struct {
 	// is closed and the votes are tallied.
 	VotingPeriod beacon.EpochTime `json:"voting_period,omitempty"`
 
-	// Quorum is he minimum percentage of voting power that needs to be cast on
-	// a proposal for the result to be valid.
-	Quorum uint8 `json:"quorum,omitempty"`
-
-	// Threshold is the minimum percentage of VoteYes votes in order for a
-	// proposal to be accepted.
-	Threshold uint8 `json:"threshold,omitempty"`
+	// StakeThreshold is the minimum percentage of VoteYes votes in terms
+	// of total voting power when the proposal expires in order for a
+	// proposal to be accepted.  This value has a lower bound of 67.
+	StakeThreshold uint8 `json:"stake_threshold,omitempty"`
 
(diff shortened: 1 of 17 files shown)


go/governance/api/api.go
```diff
@@ -243,13 +243,10 @@ type ConsensusParameters struct {
...
-	// Quorum is he minimum percentage of voting power that needs to be cast on
-	// a proposal for the result to be valid.
-	Quorum uint8 `json:"quorum,omitempty"`
-
-	// Threshold is the minimum percentage of VoteYes votes in order for a
-	// proposal to be accepted.
-	Threshold uint8 `json:"threshold,omitempty"`
+	// StakeThreshold is the minimum percentage of VoteYes votes in terms
+	// of total voting power when the proposal expires in order for a
+	// proposal to be accepted.  This value has a lower bound of 2/3.
```
pro-wh, line 248: lower bound of 67

pro-wh (changes requested):
lg

requesting changes for the `<` :arrow_right: `<=`

I guess at this point explicit no votes are just in case you change your mind and want to remove your yes vote

pro-wh (approved):

PR #4431, opened 2022-01-17
go/runtime/scheduling: Allow a higher priority tx even if queue full

kostko:


go/runtime/scheduling/simple/txpool/priorityqueue/priority_queue.go
```diff
@@ -138,15 +156,35 @@ func (q *priorityQueue) GetBatch(force bool) []*transaction.CheckedTransaction {
...
 		q.priorityIndex.Delete(item)
 		for k, v := range item.tx.Weights() {
 			q.poolWeights[k] -= v
 		}
 	}
 
-	return batch
+	// Update lowest priority.
+	if len(items) > 0 {
+		if lpi := q.priorityIndex.Max(); lpi != nil {
```
ptrus, line 175: Should we invert the `item.Less` function so that we can use `Min()` here, as that would make more sense (I think only `Max`->`Min` and `Ascend` -> `Descend`) will need update? 

I don't recall if there's any special reason why `Less` is currently inverted, I don't think it should impact the overall performance changing it.

kostko: Yeah makes sense, I also found this confusing.

kostko: Updated.

ptrus (approved):

Yawning (approved):

Issue #4436, opened 2022-01-19
Report runtime version in oasis-node control status for runtime client nodes


tjanez:
##### SUMMARY
It would be useful to know the version of the runtime an Oasis Node is running, even if the node is only running a runtime client.

Perhaps it could be part of `oasis-node control status`?

##### ISSUE TYPE
- Feature Idea

PR #4415, opened 2022-01-07
Incoming runtime messages (ADR 0011)

kostko:
This implements incoming runtime messages as per ADR 0011.

TODO
* [x] Add queries.
* [x] Update executor commitments.
* [x] Update runtime host protocol.
* [x] Update Rust runtime support library.
* [x] Update commit processing pipeline.
* [x] Update consensus roothash service to clear processed messages from queue.
* [x] Tests.

PR #4438, opened 2022-01-20
Add support for runtime schedule control

kostko:
Based on #4415.
Also adds functionality needed for #4436.

TODO
* [x] Cleanup.
* [x] More testing, including long-term tests.
* [x] Support for fetching more transactions from the host (so that we don't need to always dump the full txpool into the runtime to then just schedule a few of them).
* [x] Interoperability testing with old runtimes and old nodes in the same committee.
* [x] Forward-port some fixes from #4446.

### DIFF
--- a/go/runtime/host/protocol/types.go
+++ b/go/runtime/host/protocol/types.go
@@ -103,6 +103,8 @@ type Body struct {
 	HostLocalStorageSetResponse     *Empty                           `json:",omitempty"`
 	HostFetchConsensusBlockRequest  *HostFetchConsensusBlockRequest  `json:",omitempty"`
 	HostFetchConsensusBlockResponse *HostFetchConsensusBlockResponse `json:",omitempty"`
+	HostFetchTxBatchRequest         *HostFetchTxBatchRequest         `json:",omitempty"`
+	HostFetchTxBatchResponse        *HostFetchTxBatchResponse        `json:",omitempty"`
 }
 
 // Type returns the message type by determining the name of the first non-nil member.
@@ -151,13 +153,35 @@ type RuntimeInfoRequest struct {
 	LocalConfig map[string]interface{} `json:"local_config,omitempty"`
 }
 
-// RuntimeInfoResponse is a worker info response message body.
+// Features is a set of supported runtime features.
+type Features struct {
+	// ScheduleControl is the schedule control feature.
+	ScheduleControl *FeatureScheduleControl `json:"schedule_control,omitempty"`
+}
+
+// HasScheduleControl returns true when the runtime supports the schedule control feature.
+func (f *Features) HasScheduleControl() bool {
+	return f != nil && f.ScheduleControl != nil
+}
+
+// FeatureScheduleControl is a feature specifying that the runtime supports controlling the
+// scheduling of batches. This means that the scheduler should only take priority into account and
+// ignore weights, leaving it up to the runtime to decide which transactions to include.
+type FeatureScheduleControl struct {
+	// InitialBatchSize is the size of the initial batch of transactions.
+	InitialBatchSize uint32 `json:"initial_batch_size"`
+}
+
+// RuntimeInfoResponse is a runtime info response message body.
 type RuntimeInfoResponse struct {
-	// ProtocolVersion is the runtime protocol version supported by the worker.
+	// ProtocolVersion is the runtime protocol version supported by the runtime.
 	ProtocolVersion version.Version `json:"protocol_version"`
 
 	// RuntimeVersion is the version of the runtime.
 	RuntimeVersion version.Version `json:"runtime_version"`
+
+	// Features describe the features supported by the runtime.
+	Features *Features `json:"features,omitempty"`
 }
 
 // RuntimeCapabilityTEERakInitRequest is a worker RFC 0009 CapabilityTEE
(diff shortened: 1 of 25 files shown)


go/runtime/host/protocol/types.go
```diff
@@ -409,3 +462,20 @@ type HostFetchConsensusBlockRequest struct {
 type HostFetchConsensusBlockResponse struct {
 	Block consensus.LightBlock `json:"block"`
 }
+
+// HostFetchTxBatchRequest is a request to host to fetch a further batch of transactions. The host
+// will reply with the transactions that it hasn't sent yet.
+type HostFetchTxBatchRequest struct {
+	// Offset specifies the offset of the batch within the transaction pool.
+	//
+	// Note that this may miss transactions that have arrived after the last batch was fetched.
```
ptrus, line 471: Hm, offset doesn't seem all that useful as this might actually cause new low priority transactions to be prioritized (as high priority ones will be put in front of the queue and being missed by the offset). 

I guess it's up to the runtime to be aware of that and not use offset if it is not desired. But is there a specific case where offset would be useful?

kostko: Yes, the idea is to support streaming transactions from the queue so that we don't need to send the whole queue into the runtime only to have a few transactions actually scheduled. So the runtime sets an initial batch size to fetch and then can use this method to request additional transactions. One could add some sort of snapshotting feature to ensure consistent reads, but not sure if this will cause any problems in practice?

ptrus: Would in that case make more sense to use a fifo queue on the scheduler? And have runtime order these by priority? That way, using offset won't miss new transactions.

kostko: The problem in that case is that this would require sending the full queue into the runtime which this is trying to avoid. 

Missed transactions are not really a problem as they will be included in the next block (as they would be if we sent the complete queue immediately). One downside of having an integer offset is that this could result in the host sending transactions that the runtime has already seen (because new transactions were inserted above, pushing others down). 

Maybe it would be better if the offset was a transaction hash.

ptrus: Back to:

> Yes, the idea is to support streaming transactions from the queue so that we don't need to send the whole queue into the runtime only to have a few transactions actually scheduled. So the runtime sets an initial batch size to fetch and then can use this method to request additional transactions

So either:
- initial runtime request gets low number of txes -> there's not a lot of benefit from having `offset`
- initial runtime requests gets enough transactions to fill up the batch -> no need to subsequent requests (`offset` not needed)
- initial runtime requests gets a lot of transactions (but not enough to fill up the whole batch), so it tries to fetch new transactions by using `offset=<nr.of transactions>`:
  - if `limit` for the subsequent fetch is low (e.g. number of missing transactions to fill up the batch): most likely new some new high. prio transactions pushed the existing transactions, so the runtime will get a lot of duplicates and needing to do more subsequent fetches (using `transaction hash` for offset would fix this)
  - if `limit` for the subsequent is big (e.g. batch size): no benefit of using `offset` again, and it would be better to just re-fetch with `offset=0`

kostko: Changed offset to be a hash.

ptrus: Great, thanks!

go/runtime/scheduling/simple/txpool/priorityqueue/priority_queue.go
```diff
@@ -187,6 +187,58 @@ func (q *priorityQueue) removeTxsLocked(items []*item) {
 	}
 }
 
+// Implements api.TxPool.
+func (q *priorityQueue) GetPrioritizedBatch(offset *hash.Hash, limit uint32) []*transaction.CheckedTransaction {
+	q.Lock()
+	defer q.Unlock()
+
+	var (
+		batch       []*transaction.CheckedTransaction
```
ptrus, line 196: Might make sense pre-allocating batch here (with length 0)

go/runtime/scheduling/simple/txpool/priorityqueue/priority_queue.go
```diff
@@ -187,6 +192,57 @@ func (q *priorityQueue) removeTxsLocked(items []*item) {
...
+		// Add the tx to the batch.
+		batch = append(batch, item.tx)
+		if uint32(len(batch)) >= limit {
+			return false
+		}
+		return true
+	})
+	if offset != nil {
+		// Skip the offset item itself.
+		batch = batch[1:]
```
pro-wh, line 235: reposting from backport https://github.com/oasisprotocol/oasis-core/pull/4446#discussion_r793900480:

it's harder to tell that this is correct, compared to having an equality check inside the loop

- could the weight limits have changed, to make us skip the offset hash?
- could there have been duplicates of the offset hash?
- could something else have removed the offset hash?

some of these questions may have safe answers for how we use this data structure in the scheduler. it would be nice if we could make it safe in the general case

kostko: Fixed.

go/runtime/scheduling/simple/txpool/tester.go
```diff
@@ -87,9 +87,10 @@ func testBasic(t *testing.T, pool api.TxPool) {
 	require.EqualValues(t, 10, len(batch), "Batch size")
 	require.EqualValues(t, 51, pool.Size(), "Size")
 
-	hashes := make([]hash.Hash, len(batch))
-	for i, tx := range batch {
-		hashes[i] = tx.Hash()
+	hashes := make([]hash.Hash, 0, len(batch))
```
matevz, line 90: Nit: Since you're inserting duplicates, should this be
```suggestion
	hashes := make([]hash.Hash, 0, 2*len(batch))
```
Or just remove the capacity entirely.

Yawning (approved):
Looks fine, we're basically running a more involved version of this already.

PR #4441, opened 2022-01-20
[BACKPORT/21.3.x] deps: Switch to the upstream jsonrpc import

pro-wh:
of #4440

PR #4442, opened 2022-01-20
changelog: Assemble changes for 21.3.9 release

pro-wh:
pre-pushing this. gonna rebase after #4441 merges


pro-wh:
let's not make a new release for this after all. I've been informed that the dependency we're bumping is only used in tests

Issue #4402, opened 2021-12-16
Implement Runtime Message Results


ptrus:
Implement Runtime Message Results as described in [ADR-0012](https://github.com/oasisprotocol/oasis-core/blob/master/docs/adr/0012-runtime-message-results.md).

PR #4443, opened 2022-01-21
Runtime message results

ptrus:
Closes: #4402 

Implementation of runtime message results as described in [ADR 0012].

TODO:
- [x] finish the remaining todos
- [x] implement rust types for results
- [x] add checks to e2e test runtime
- [x] update the result types in the ADR

[ADR 0012]: docs/adr/0012-runtime-message-results.md


go/consensus/tendermint/abci/messages.go
```diff
@@ -21,16 +21,20 @@ func (md *messageDispatcher) Subscribe(kind interface{}, ms api.MessageSubscribe
 }
 
 // Implements api.MessageDispatcher.
-func (md *messageDispatcher) Publish(ctx *api.Context, kind, msg interface{}) error {
-	if len(md.subscriptions[kind]) == 0 {
-		return api.ErrNoSubscribers
+func (md *messageDispatcher) Publish(ctx *api.Context, kind, msg interface{}) ([]interface{}, error) {
```
pro-wh, line 24: now that I actually see the tests, the interface for this feels so idiosyncratic. for one, it's one of those rare functions where the first return value is meaningful when the error is non-nil.

from the code in this PR, the behavior is like this:

- the results slice always has one element per subscriber
- the error is a `multierror` with an element for each subscriber that errored, thus varying length
- the subscriber's interface treats its response as undefined when the error is non-nil
- for a subscriber that errors, its corresponding slot in the results is left as nil

I just wonder if we will really be able to write rigorous multi-subscriber programs with this. my concerns are:

1. it's not clear to what extent the code calling Publish is expected to know the composition and ordering of the subscribers. the previous code where the errors are piled together suggest the calling code doesn't know/doesn't care. but surely that knowledge must be increased now that we have responses that the calling code will be looking for.
2. it's not clear how the calling code should find the response it's looking for. by known index? or should it maybe loop through and find it by type?
3. some calling code that's less knowledgeable about the composition of subscribers won't be able to differentiate subscribers that succeed and return `nil` from subscribers that errored. at least not from looking at the results slice. not sure if this will come up.
4. for calling code that's more knowledgeable about the composition of subscribers, I think it will be hard to match up errors with what subscriber it came from. maybe those use cases will have to include more metadata in the errors to work around that.

do we have cases where there are multiple subscribers for a message? or if not, what are the intended use cases? how do/will we handle the situation where some subscribers fail and some succeed? is the zeroth result always the one that a runtime will want to see?

ptrus: > the error is a multierror with an element for each subscriber that errored, thus varying length

The error return type should probably be changed to be a list of errors, so then the caller knows which responses are non failed.

> do we have cases where there are multiple subscribers for a message? or if not, what are the intended use cases? 

No, no such cases. Not sure about intended use-cases.

> how do/will we handle the situation where some subscribers fail and some succeed

I guess it's up to the caller to decide.

> is the zeroth result always the one that a runtime will want to see?

For all existing cases, yes, as there's always a single subscriber. Maybe the could should propagate all results to the caller, and not just the first one (although this makes no difference at this point in time).

pro-wh: There's at least a use case for zero subscribers, so that we can support non-lock-step changes to paratimes and core. Maybe now is the time to consider switching to a zero-or-one subscriber model of message handlers?

pro-wh: error slice would be unambiguous I guess. potentially verbose retrofitting work though.

I see some places in this PR's callsite changes like this

```go
// Notify other interested applications about the resumed runtime.
if _, err = app.md.Publish(ctx, registryApi.MessageRuntimeResumed, rt); err != nil {
```

I guess we've been using this message dispatcher class for things other than runtime-sent messages too

kostko: > I guess we've been using this message dispatcher class for things other than runtime-sent messages too.

Yes, this is a general pubsub mechanism used by the consensus layer services.

go/consensus/tendermint/apps/staking/transactions_test.go
```diff
@@ -90,7 +90,7 @@ func TestReservedAddresses(t *testing.T) {
 	_ = staking.NewReservedAddress(testPK)
```
kostko, line 91: Update tests to include checks that return values are correct.

go/roothash/api/api.go
```diff
@@ -445,6 +445,9 @@ type MessageEvent struct {
 	Module string `json:"module,omitempty"`
 	Code   uint32 `json:"code,omitempty"`
 	Index  uint32 `json:"index,omitempty"`
+
+	// Result contains cbor encoded message execution result for successfully executed messages.
```
kostko, line 449:
```suggestion
	// Result contains CBOR-encoded message execution result for successfully executed messages.
```

go/roothash/api/api.go
```diff
@@ -445,6 +445,9 @@ type MessageEvent struct {
...
+	Result []byte `json:"result,omitempty"`
```
kostko, line 450: Should this be changed to `cbor.RawMessage`?

runtime/src/consensus/roothash.rs
```diff
@@ -119,6 +119,10 @@ pub struct MessageEvent {
     #[cbor(optional)]
     #[cbor(default)]
     pub index: u32,
+
+    #[cbor(optional)]
+    #[cbor(default)]
+    pub result: Vec<u8>,
```
kostko, line 125: In case we go with `cbor.RawMessage` in Go, this would be `cbor::Value`.

ptrus: Is the `Hash` trait implementation needed for `MessageEvent` and `RoundResults` structs? If yes, we'd need to manually define it then as `cbor::Value` does not implement `std::hash::Hash`. 
But if it's not needed, we could also just drop it.

kostko: Yeah probably not needed, just drop it.

go/consensus/tendermint/apps/staking/transactions_test.go
ptrus: Added

go/consensus/tendermint/abci/messages.go
ptrus: Ok agreed, i'll make `Publish` return a single reuslt then

ptrus: Updated. Please take a look, I'll update the ADR once this new approach is finalized.

pro-wh (approved):
new dispatcher interface looks good, thanks!

go/consensus/tendermint/abci/messages_test.go
```diff
@@ -58,45 +62,50 @@ func TestMessageDispatcher(t *testing.T) {
...
 	// Returning an error.
-	err = md.Publish(ctx, testMessageA, &errorMessage{})
+	res, err = md.Publish(ctx, testMessageA, &errorMessage{})
 	require.Error(err, "Publish")
+	require.EqualValues(nil, res, "correct publish message result")
 	require.True(errors.Is(err, errTest), "returned error should be the correct one")
 
-	// Multiple subscribers.
-	var ms2 testSubscriber
+	// Multiple subscribers. Multiple subscribers returning results on the same message is an invariant.
```
kostko, line 95:
```suggestion
	// Multiple subscribers. Multiple subscribers returning results on the same message is an invariant violation.
```

go/consensus/tendermint/apps/roothash/messages.go
```diff
@@ -30,19 +31,20 @@ func (app *rootHashApplication) processRuntimeMessages(
 		defer ctx.Close()
 	}
 
-	var results []*roothash.MessageEvent
+	var event []*roothash.MessageEvent
```
kostko, line 34:
```suggestion
	var events []*roothash.MessageEvent
```

go/consensus/tendermint/apps/staking/transactions_test.go
```diff
@@ -6,6 +6,7 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/oasisprotocol/oasis-core/go/beacon/api"
```
kostko, line 9:
```suggestion
	beacon "github.com/oasisprotocol/oasis-core/go/beacon/api"
```

kostko (approved):

go/consensus/tendermint/abci/messages.go
```diff
@@ -21,16 +23,31 @@ func (md *messageDispatcher) Subscribe(kind interface{}, ms api.MessageSubscribe
...
+		if resp, err := ms.ExecuteMessage(ctx, kind, msg); err != nil {
 			errs = multierror.Append(errs, err)
+		} else {
+			switch {
+			case resp != nil && result == nil:
+				// Non-nil result.
+				result = resp
+			case resp != nil && result != nil:
+				// Multiple non-nil results, this is unexpected and unsupported by the pub-sub interface at this time.
+				panic(fmt.Sprintf("unexpected result: got: %d, previous result: %d", resp, result))
```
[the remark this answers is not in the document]
ptrus: Agreed, but I think that panic is probably more suitable, as this is considered a programmer error at the moment.

Issue #4170, opened 2021-08-04
Setup rust code linting


ptrus:
Setup [`clippy`](https://github.com/rust-lang/rust-clippy) in CI to lint rust code. 

Will probably require code fixes since the rust code currently won't pass all of the lints. To avoid some unnecessary work should do this only after:
- https://github.com/oasisprotocol/oasis-core/issues/3572

Yawning:
We can also think about integrating more static analysis into our Rust (eg: https://github.com/rust-lang/miri).

PR #4455, opened 2022-01-28
rust: setup clippy lints and fix existing lint issues

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/4170


runtime/src/common/key_format.rs
```diff
@@ -70,9 +68,7 @@ pub trait KeyFormat {
         if data[0] != Self::prefix() {
             return None;
         }
-        if data.len() < 1 + Self::size() {
-            panic!("key format: malformed input");
-        }
+        assert!(data.len() > Self::size(), "key format: malformed input");
```
pro-wh, line 71: I'd somehow find this more understandable with `data.len() >= 1 + Self::size()`

runtime/src/common/key_format.rs
```diff
@@ -70,9 +68,7 @@ pub trait KeyFormat {
         if data[0] != Self::prefix() {
```
pro-wh, line 68: come to think of it, we access `[0]` before a length check? that could be risky

pro-wh (commented):
23/74

hold on let me empty out the linter

wordcloud

runtime/src/common/key_format.rs
ptrus: Nice find, added a check (and also an assert a bit below)

runtime/src/common/key_format.rs
pro-wh: we could make it more formal

```rust
const VERSION_PREFIX_SIZE: usize = the_size_of_a_u8;
data.len() >= VERSION_PREFIX_SIZE + Self:size()
```

runtime/src/storage/mkvs/cache/mod.rs
```diff
@@ -1,5 +1,170 @@
...
+    /// For e.g. the LRU cache, this may be a used timestamp.
+    fn get_cache_extra(&self) -> CacheExtra<Item>;
+    /// Set the item's caching hint.
+    fn set_cache_extra(&mut self, new_val: CacheExtra<Item>);
+    /// Return the size, in bytes, of the item when cached.
+    fn get_cached_size(&self) -> usize;
+}
+
+/// Callback type used for updating cache items after a commit.
+pub type CacheUpdater<C> = Box<dyn Fn(&mut C)>;
```
pro-wh, line 143: :information_source:  `-> ()` removed

runtime/src/storage/mkvs/sync/mod.rs
```diff
@@ -5,15 +5,86 @@ mod merge;
...
+use io_context::Context;
+
+use crate::{
+    common::crypto::hash::Hash,
+    storage::mkvs::{tree::*, Prefix},
+};
+
+/// Identifies a specific tree and a position within that tree.
+#[derive(Clone, Debug, Default, PartialEq, cbor::Encode, cbor::Decode)]
+pub struct TreeID {
```
pro-wh, line 28: same question about this, moved from sync.rs

pro-wh (commented):
50/74

runtime/src/common/key_format.rs
pro-wh: thanks!

runtime/src/common/sgx/avr.rs
```diff
@@ -100,12 +100,12 @@ lazy_static! {
     static ref IAS_TRUST_ANCHOR: Vec<u8> = {
         let pem = match parse_x509_pem(IAS_TRUST_ANCHOR_PEM.as_bytes()) {
             Ok((rem, pem)) => {
-                if !rem.is_empty() {
-                    panic!("anchor PEM has trailing garbage");
-                }
-                if pem.label != PEM_CERTIFICATE_LABEL {
-                    panic!("PEM does not contain a certificate: '{:?}'", pem.label);
-                }
+                assert!(rem.is_empty(), "anchor PEM has trailing garbage");
```
Yawning, line 103: Are asserts in rust unconditionally compiled?

pro-wh: ya I had to look this up, and `debug_assert!` is the conditional one

runtime/src/storage/mkvs/cache/cache.rs
```diff
@@ -1,166 +0,0 @@
...
-        root: Root,
-        ptr: NodePtrRef,
-        rs: &mut Box<dyn ReadSync>,
-    ) -> Result<Proof> {
-        (*self)(ctx, root, ptr, rs)
-    }
-}
-
-/// Cache interface for the in-mmory tree cache.
-pub trait Cache {
```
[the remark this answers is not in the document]
ptrus: `clippy::module_inception` because these were in `cache::cache` module, now these are in `cache` module directly

runtime/src/storage/mkvs/sync/mod.rs
ptrus: `clippy::module_inception` because these were in `sync::sync` module, now these are in `sync` module directly

runtime/src/common/key_format.rs
ptrus: Ah I see, I thought you generally preferred the `1+` version. Yeah I can make this case clearer.

runtime/src/consensus/tendermint/verifier.rs
```diff
@@ -640,20 +642,20 @@ impl VotingPowerCalculator for DomSepVotingPowerCalculator {
             let sign_bytes = signed_vote.sign_bytes();
             // Use Oasis Core domain separation scheme.
             let sign_bytes = Hash::digest_bytes_list(&[TENDERMINT_CONTEXT, &sign_bytes]);
-            if validator
+            let power = validator.power();
```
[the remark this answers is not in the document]
ptrus: The below: `if validator._.is_err(){ return Err()}` changed to `?`, but since the `validator` is moved into `map_err`  it complains if it is used in line 658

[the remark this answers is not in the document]
pro-wh: oh moved :thinking: 

whereas before, it was smart enough to know that if we didn't return, we wouldn't have moved `validator`

tragic, but thanks for the explanation

runtime/src/storage/mkvs/sync/mod.rs
pro-wh: thanks, didn't know about that

pro-wh (approved):
74/74

emptying out the linter again

wordcloud2

runtime/src/storage/mkvs/tree/iterator.rs
```diff
@@ -173,11 +173,7 @@ impl<'tree> TreeIterator<'tree> {
                     // case everything in this subtree will be larger so we need to take the first value.
                     // takeFirst bool
```
kostko, line 174: Looks like a leftover, we can remove this.

runtime/src/storage/mkvs/tree/tree.rs
```diff
@@ -1,162 +1 @@
-use std::{cell::RefCell, fmt, rc::Rc};
```
kostko, line 1: This file should be completely removed, it seems there's an empty line in there.

runtime/src/consensus/registry.rs
```diff
@@ -290,7 +290,7 @@ pub struct Runtime {
 }
 
 fn staking_params_are_empty(p: &RuntimeStakingParameters) -> bool {
-    return Option::is_none(&p.thresholds);
+    Option::is_none(&p.thresholds)
```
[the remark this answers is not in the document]
ptrus: Is the `skip_serializing_if = "staking_params_are_empty"` even needed still with the new cbor lib?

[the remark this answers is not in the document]
ptrus: I'll open an issue for this, and fix in a separate PR as suggested, as the fix should also contain tests.

kostko (approved):

PR #4456, opened 2022-02-01
 go/staking: add minimum transfer amount 

pro-wh:
cross reference: https://oasisprotocol.clickup.com/t/1zfn2hb

### DIFF
--- a/go/consensus/tendermint/apps/staking/transactions.go
+++ b/go/consensus/tendermint/apps/staking/transactions.go
@@ -131,6 +131,11 @@ func (app *stakingApplication) burn(ctx *api.Context, state *stakingState.Mutabl
 		return nil
 	}
 
+	// Check if sender provided at least a minimum amount.
+	if burn.Amount.Cmp(&params.MinTransferAmount) < 0 {
+		return staking.ErrUnderMinTransferAmount
+	}
+
 	fromAddr := ctx.CallerAddress()
 	if fromAddr.IsReserved() {
 		return staking.ErrForbidden
--- a/go/consensus/tendermint/apps/staking/transactions.go
+++ b/go/consensus/tendermint/apps/staking/transactions.go
@@ -39,6 +39,11 @@ func (app *stakingApplication) transfer(ctx *api.Context, state *stakingState.Mu
 		return nil, nil
 	}
 
+	// Check if sender provided at least a minimum amount.
+	if xfer.Amount.Cmp(&params.MinTransferAmount) < 0 {
+		return nil, staking.ErrUnderMinTransferAmount
+	}
+
 	fromAddr := ctx.CallerAddress()
 	if fromAddr.IsReserved() || !isTransferPermitted(params, fromAddr) {
 		return nil, staking.ErrForbidden
--- a/go/consensus/tendermint/apps/staking/transactions.go
+++ b/go/consensus/tendermint/apps/staking/transactions.go
@@ -607,6 +607,11 @@ func (app *stakingApplication) withdraw(
 		return nil, nil
 	}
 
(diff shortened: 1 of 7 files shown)


Yawning (commented):
I guess this needs fixgenesis support, otherwise it looks fine to me.

pro-wh:
hm, maybe I should perform the check on withdraw too, for symmetry

pro-wh:
adding to fixgenesis: the proposal is to set it to 1 base unit. this particular change is mostly to prevent zero-amount transfers from creating useless empty accounts.

go/consensus/tendermint/apps/staking/transactions.go
```diff
@@ -34,6 +34,11 @@ func (app *stakingApplication) transfer(ctx *api.Context, state *stakingState.Mu
 		return err
 	}
 
+	// Check if sender provided at least a minimum amount.
+	if xfer.Amount.Cmp(&params.MinTransferAmount) < 0 {
```
pro-wh, line 38: gonna move this to after the IsSimulation bail, to match AddEscrow

pro-wh:
rearranged checks: now compares with min transfer amount parameter after IsSimulation bail, similar to min delegation amount check.

added similar check to withdraw (same parameter)

Yawning:
Should this be applied to burn and allow as well?

.changelog/4456.feature.md
```diff
@@ -0,0 +1,4 @@
+go/staking: Add minimum transfer amount
```
kostko, line 1: This is a breaking change to the consensus protocol.

kostko:
> Should this be applied to burn and allow as well?

I don't think allow is problematic (as it only updates the account's allowance, a zero allowance is ignored and the total number of allowances per account is already limited), the more important thing is withdraw (which seems to be handled in this PR).

kostko:
For burn we could just ignore burning a zero amount as that results in a useless event to be emitted.

I guess an additional question is whether we would want to require a minimum balance to remain in an account, otherwise operations fail? Otherwise one can just do this:
* Transfer `min_transfer` to A.
* Transfer `min_transfer` from A to B.
* Transfer `min_transfer` from B to C.
* ...

pro-wh:
we should do the other transactions as well. it's not so much that allow, for example, won't really charge the allowances table, but moreso that we have to create an account and start tracking the nonce after a successful transaction

pro-wh:
let's have the minimum balance thing in a separate PR.

kostko:
> we should do the other transactions as well. it's not so much that allow, for example, won't really charge the allowances table, but moreso that we have to create an account and start tracking the nonce after a successful transaction

True, but this basically applies to any successfully authenticated transaction even though it doesn't deal with tokens (besides paying for fees). So this would basically imply a minimum account balance, because the check would need to be "you need to have at least X in your account before you can make any transaction"?

pro-wh:
added similar check (same parameter) to burn

adjusted ordering of checks in withdraw to check amount before consulting transfers disablement :shrug:

pro-wh:
let's leave allow and others as is and proceed to draft a minimum balance requirement \\:

kostko (approved):
I think this looks good. Just change that feature entry to breaking.

pro-wh:
summary of changes after review:

1. proposed value (via fixgenesis) set to 0.01 ROSE (10_000_000 base units)
2. test fixes
   a. transfer and withdraw result return
   b. TestReservedAddresses use non-nil tx body
3. staking tests GenesisState sets a non-empty MinTransferAmount, so we can exercise its serialization
4. updated TestGenesisChainContext

PR #4459, opened 2022-02-03
go/worker/storage: Transition storage sync to P2P

kostko:
TODO
* [x] Go over storage sync metrics and consider adding some more metrics.


go/worker/common/p2p/crypto.go
```diff
@@ -28,6 +28,10 @@ func (s *p2pSigner) Equals(other libp2pCrypto.Key) bool {
 }
 
 func (s *p2pSigner) Raw() ([]byte, error) {
+	if us, ok := s.signer.(signature.UnsafeSigner); ok {
```
Yawning, line 31: Instead of doing this, maybe generate and persist an appropriate amount of system entropy or something?

go/worker/common/p2p/rpc/client.go
```diff
@@ -0,0 +1,320 @@
...
+
+	// No peers could be reached to service this request.
+	c.logger.Debug("no peers could be reached to service request",
+		"method", method,
+	)
+
+	return nil, fmt.Errorf("call failed on all peers")
+}
+
+func (c *client) CallMulti(
```
Yawning, line 148: So this doesn't propagate errors?  Does it need to?

go/common/crypto/signature/signers/file/file_signer.go
```diff
@@ -207,6 +208,12 @@ func (s *Signer) Reset() {
 	}
 }
 
+// UnsafeBytes returns the byte representation of the private key.  This
+// MUST be removed for HSM support.
+func (s *Signer) UnsafeBytes() []byte {
```
Yawning, line 213: I really really really would rather avoid doing this.

Yawning:
The more I think about this, the more I am incredibly against exposing the actual private key to handle libp2p's garbage bullshit.  Instead I propose something like this:

```
// In the signature package.

// StaticEntropyProvider is the interface implemented by signers that support providing
// persistent static entropy that is entirely independent from a keypair.
type StaticEntropyProvider interface {
  // StaticEntropy returns PrivateKeySize bytes of cryptographic entropy that
  // is independent from the Signer's private key.  The value of this entropy
  // is constant for the lifespan of the signer's underlying key pair.
  StaticEntropy() []byte
}
```

When initializing the file signer, load (or generate) cryptographic entropy and persist it to `libp2p_is_dumb.pem` or something, and use that instead of exposing the private key.

kostko:
Agreed, that sounds like a better solution, especially given that the only use case for this is to generate the QUIC reset key.

Yawning:
> Agreed, that sounds like a better solution, especially given that the only use case for this is to generate the QUIC reset key.

We are pretty good about separating concerns and using different keys for basically everything, but deriving stuff like that off a signing key makes my tinfoil hat crinkle, even if the derivation procedure looks "ok".

go/worker/storage/p2p/protocol.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
+	storage "github.com/oasisprotocol/oasis-core/go/storage/api"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/checkpoint"
+)
+
+// Unique protocol identifier and version for the storage sync protocol.
+const (
+	StorageSyncProtocolID      = "storagesync"
+	StorageSyncProtocolVersion = "1.0.0"
```
ptrus, line 14: This could be of type `version.Version`

go/worker/common/p2p/rpc/client.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+
+	if rsp != nil {
+		return cbor.Unmarshal(rawRsp.Ok, rsp)
+	}
+	return nil
+}
+
+// NewClient creates a new RPC client for the given protocol.
+func NewClient(p2p P2P, runtimeID common.Namespace, protocolID, version string) Client {
```
ptrus, line 307: Change version type to `version.Version` (and also on the server) and in `NewRuntimeProtocolID` could then either use the full version or the major part only.

go/worker/common/p2p/rpc/client.go
```diff
@@ -0,0 +1,320 @@
...
+const (
+	RequestWriteDeadline = 5 * time.Second
+)
+
+// PeerFeedback is an interface for providing deferred peer feedback after an outcome is known.
+type PeerFeedback interface {
+	// RecordSuccess records a successful protocol interaction with the given peer.
+	RecordSuccess()
+
+	// RecordSuccess records an unsuccessful protocol interaction with the given peer.
```
ptrus, line 29:
```suggestion
	// RecordFailure records an unsuccessful protocol interaction with the given peer.
```

go/worker/common/p2p/rpc/peermgmt.go
```diff
@@ -0,0 +1,283 @@
...
+	// Peer is only added in case it supports the specified protocol.
+	AddPeer(peerID core.PeerID)
+
+	// RemovePeer unconditionally removes the peer from the peer manager.
+	RemovePeer(peerID core.PeerID)
+
+	// RecordSuccess records a successful protocol interaction with the given peer.
+	RecordSuccess(peerID core.PeerID, latency time.Duration)
+
+	// RecordSuccess records an unsuccessful protocol interaction with the given peer.
```
ptrus, line 38: RecordFailure

go/worker/common/p2p/rpc/peermgmt.go
```diff
@@ -0,0 +1,283 @@
...
+	avgRequestLatency time.Duration
+}
+
+func (ps *peerStats) getScore(avgRequestLatency time.Duration) float64 {
+	if ps.successes+ps.failures > 0 {
+		// We have some history for this peer.
+		failRateI := float64(ps.failures) / float64(ps.failures+ps.successes)
+		return float64(ps.avgRequestLatency) + failRateI*float64(avgRequestLatency)
+	} else {
+		const newPeerMul = 0.9
```
ptrus, line 63: Although this is an internal detail, maybe move this to the top of the file and name something like: `newPeerLatencyFactor` with a comment describing the use.

go/worker/common/p2p/rpc/peermgmt.go
```diff
@@ -0,0 +1,283 @@
...
+	successes         int
+	failures          int
+	avgRequestLatency time.Duration
+}
+
+func (ps *peerStats) getScore(avgRequestLatency time.Duration) float64 {
+	if ps.successes+ps.failures > 0 {
+		// We have some history for this peer.
+		failRateI := float64(ps.failures) / float64(ps.failures+ps.successes)
+		return float64(ps.avgRequestLatency) + failRateI*float64(avgRequestLatency)
```
ptrus, line 61: Should this be weighted a bit more to the `avgRequestLatency` when the history is small?

go/worker/common/p2p/rpc/peermgmt.go
```diff
@@ -0,0 +1,283 @@
...
+		return float64(avgRequestLatency) * newPeerMul
+	}
+}
+
+func (ps *peerStats) recordLatency(latency time.Duration) {
+	if ps.avgRequestLatency == 0 {
+		ps.avgRequestLatency = latency
+	} else {
+		// invAlpha = (N+1)/2
+		const invAlpha = 10
```
ptrus, line 73: Maybe move all the internal constants in the same section with documenting comments.

go/worker/common/p2p/rpc/peermgmt.go
kostko: Fixed.

go/worker/common/p2p/rpc/peermgmt.go
kostko: Done.

kostko:
@Yawning I just pushed a version that uses the static entropy provider, please take a look.

go/worker/common/p2p/rpc/peermgmt.go
```diff
@@ -0,0 +1,291 @@
...
+	successes         int
+	failures          int
+	avgRequestLatency time.Duration
+}
+
+func (ps *peerStats) getScore(avgRequestLatency time.Duration) float64 {
+	if ps.successes+ps.failures > 0 {
+		// We have some history for this peer.
+		failRate := float64(ps.failures) / float64(ps.failures+ps.successes)
+		return float64(ps.avgRequestLatency) + failRate*float64(avgRequestLatency)
```
ptrus, line 72: So a peer failing all requests quickly, will have a similar (slightly worse) score than an average node returning all correct results? Hm, I guess that does make sense, since if the peer is failing quickly that's not really all that bad, as not a lot of time is spent on it.

Maybe the penalty part for failing could have an additional constant factor? But I have no reason to suggest anything other than `1`.

go/worker/common/p2p/rpc/peermgmt.go
```diff
@@ -0,0 +1,291 @@
...
+	GetBestPeers() []core.PeerID
+}
+
+type peerStats struct {
+	successes         int
+	failures          int
+	avgRequestLatency time.Duration
+}
+
+func (ps *peerStats) getScore(avgRequestLatency time.Duration) float64 {
```
ptrus, line 68: Maybe a note that "lower is better"

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -470,10 +362,7 @@ func (n *Node) syncCheckpoints(genesisRound uint64) (*blockSummary, error) {
 			syncState.Roots = nil
 		}
 
-		// Compute how many nodes we will be asking for checkpoints.
-		groupSize := descriptor.Executor.GroupSize + descriptor.Executor.GroupBackupSize
-
-		status, err := n.handleCheckpoint(check, n.storageNodesGrpc, groupSize)
+		status, err := n.handleCheckpoint(check, 5) // TODO: cfg for parallelism
```
ptrus, line 365: Don't forget the todo.

Yawning:
> @Yawning I just pushed a version that uses the static entropy provider, please take a look.

That looks about like how I would do it.

go/worker/storage/committee/checkpoint_sync.go
kostko: Done, thanks!

go/worker/common/p2p/rpc/peermgmt.go
kostko: Done.

go/worker/common/p2p/rpc/client.go
```diff
@@ -0,0 +1,321 @@
...
+			// Ignore failed results.
+			if result.err != nil {
+				continue
+			}
+
+			rsps = append(rsps, result.rsp)
+			pfs = append(pfs, result.pf)
+		}
+	}
+	c.logger.Debug("gathered results", "num1", len(rsps), "num2", len(pfs))
```
ptrus, line 213: Maybe just `"num_results": len(rsps)`, since `rsps` and `pfs` will always be the same size?

ptrus (approved):

go/worker/common/p2p/rpc/client.go
kostko: Oops, this is a leftover from earlier debugging, will remove and check if there's anything similar left.

PR #4461, opened 2022-02-04
go/staking: add minimum balance

pro-wh:
writing down some basic thoughts on the idea of a minimum balance (general balance only)

cross reference https://oasisprotocol.clickup.com/t/1zfn2hb

still a lot of problems with this

### DIFF
--- a/go/consensus/tendermint/apps/staking/state/gas.go
+++ b/go/consensus/tendermint/apps/staking/state/gas.go
@@ -109,7 +109,7 @@ func AuthenticateAndPayFees(
 
 	// Transfer fee to per-block fee accumulator.
 	feeAcc := ctx.BlockContext().Get(feeAccumulatorKey{}).(*feeAccumulator)
-	if err := quantity.Move(&feeAcc.balance, &account.General.Balance, &fee.Amount); err != nil {
+	if err = quantity.Move(&feeAcc.balance, &account.General.Balance, &fee.Amount); err != nil {
 		return fmt.Errorf("staking: failed to pay fees: %w", err)
 	}
 
--- a/go/consensus/tendermint/apps/staking/state/gas.go
+++ b/go/consensus/tendermint/apps/staking/state/gas.go
@@ -69,15 +69,30 @@ func AuthenticateAndPayFees(
 		fee = &transaction.Fee{}
 	}
 
+	// Account must have enough to pay fee and maintain minimum balance.
+	needed := fee.Amount.Clone()
+	params, err := state.ConsensusParameters(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to fetch staking consensus parameters: %w", err)
+	}
+	if err = needed.Add(&params.MinTransactBalance); err != nil {
+		return fmt.Errorf("adding MinTransactBalance to fee: %w", err)
+	}
+
+	// Check against minimum balance plus fee.
+	if account.General.Balance.Cmp(needed) < 0 {
+		logger.Error("account balance too low",
+			"account_addr", addr,
+			"account_balance", account.General.Balance,
+			"min_transact_balance", params.MinTransactBalance,
+			"fee_amount", fee.Amount,
+		)
+		return staking.ErrBalanceTooLow
+	}
+
 	if ctx.IsCheckOnly() {
 		// Configure gas accountant on the context so that we can report gas wanted.
 		ctx.SetGasAccountant(abciAPI.NewGasAccountant(fee.Gas))
-
-		// Check that there is enough balance to pay fees. For the non-CheckTx case
-		// this happens during Move below.
-		if account.General.Balance.Cmp(&fee.Amount) < 0 {
-			return transaction.ErrInsufficientFeeBalance
-		}
 
 		// Check fee against minimum gas price if in CheckTx. Always accept own transactions.
 		// NOTE: This is non-deterministic as it is derived from the local validator
--- a/go/consensus/tendermint/apps/staking/state/state_test.go
+++ b/go/consensus/tendermint/apps/staking/state/state_test.go
@@ -664,6 +664,8 @@ func TestTransfer(t *testing.T) {
 	addr1 := staking.NewAddress(pk1)
 	pk2 := signature.NewPublicKey("bbbfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff")
 	addr2 := staking.NewAddress(pk2)
+	pk3 := signature.NewPublicKey("cccfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff")
+	addr3 := staking.NewAddress(pk3)
 
 	initialBalance := *quantity.NewFromUint64(200)
 
@@ -715,6 +717,12 @@ func TestTransfer(t *testing.T) {
 	require.Empty(ctx.GetEvents(), "no events should be emitted")
 
 	// Transfer.
+	ctxEB := appState.NewContext(abciAPI.ContextEndBlock, now)
+	defer ctxEB.Close()
+	err = s.SetConsensusParameters(ctxEB, &staking.ConsensusParameters{
+		MinTransactBalance: *quantity.NewFromUint64(50),
+	})
+	require.NoError(err, "SetConsensusParameters")
 	err = s.Transfer(ctx, addr1, addr2, quantity.NewFromUint64(50))
 	require.NoError(err, "Transfer")
 
@@ -732,4 +740,10 @@ func TestTransfer(t *testing.T) {
 	require.EqualValues(addr1, ev.From, "event should have the correct source address")
 	require.EqualValues(addr2, ev.To, "event should have the correct destination address")
 	require.EqualValues(*quantity.NewFromUint64(50), ev.Amount, "event should have the correct amount")
+
+	// Test min transact balance checks.
+	err = s.Transfer(ctx, addr1, addr2, quantity.NewFromUint64(101))
+	require.ErrorIs(err, staking.ErrBalanceTooLow, "Transfer leaving below min transact balance in source")
+	err = s.Transfer(ctx, addr1, addr3, quantity.NewFromUint64(49))
+	require.ErrorIs(err, staking.ErrBalanceTooLow, "Transfer giving below min transact balance to dest")
 }
--- a/go/consensus/tendermint/apps/staking/transactions.go
+++ b/go/consensus/tendermint/apps/staking/transactions.go
@@ -3,6 +3,7 @@ package staking
 import (
 	"fmt"
 
+	"github.com/oasisprotocol/oasis-core/go/common/errors"
 	"github.com/oasisprotocol/oasis-core/go/common/quantity"
 	"github.com/oasisprotocol/oasis-core/go/consensus/tendermint/api"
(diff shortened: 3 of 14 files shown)


go/consensus/tendermint/apps/staking/auth.go
```diff
@@ -62,6 +62,14 @@ func (app *stakingApplication) PostExecuteTx(ctx *api.Context, tx *transaction.T
 		return transaction.ErrInsufficientFeeBalance
 	}
 
+	params, err := state.ConsensusParameters(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to fetch staking consensus parameters: %w", err)
+	}
+	if account.General.Balance.Cmp(&params.MinTransactBalance) < 0 {
+		return staking.ErrUnderMinTransferAmount
```
pro-wh, line 70: adding this check at the end has two purposes:
1. prevent users from wasting accounts moving the minimum amount through many accounts
2. possibly be nice and prevent someone from accidentally locking themselves out by delegating everything

but there are problems
1. this is probably the wrong place to put it. we don't want to revert the fee payment because by now we've done the work. but paying the fee might be the thing that puts the account below the limit. could we determine this in general before doing the work? that's hard to know though, with a transfer additionally lowering the balance or a withdrawal increasing the balance.
2. other non-transaction things can bring you below the minimum, such as an allowed account withdrawing

Yawning: Should withdrawals that would bring the account below the minimum be rejected?

pro-wh: alright let's try this:

- no more postexecute check
- any action that reduce an account balance must pinky promise to check for minimum balance :skull: 
  - including fee payment
  - including transfer from an account by a transaction
  - including someone withdrawing from an account
  - including a paratime transferring the last of its balance, even though that doesn't leave a nonce behind
- certain infrastructural balances are exempt
  - common pool
  - fee accumulator

some things that this entails:

- in order to withdraw, you must already have min balance + fee already. doesn't matter if the withdrawal would have covered it
- a paratime can't transfer away the last of its balance either, even though this wouldn't leave a nonce behind
- you can still transfer a less-than-minimum amount to a new account. probably will have to combine this with the minimum transfer amount feature

pro-wh:
moving some more out-there ideas to #4466.

changes to this PR:
- authenticate check now checks after moving the fee in memory. if they don't have enough to meet the min balance after moving the fee, even the fee payment gets rolled back
- general post-execute (was written wrong anyway) check removed
- manually add some checks to staking app methods that are known to affect general balance: transfer/withdraw (from + to), add escrow, burn

thus:
- various other non-method places can still move funds to create less-than-minimum balances:
  - fee payment
  - debonding completion
- future work better be extra careful
- you can no longer transfer less than the min balance to a new account
- you can't withdraw without already having the min balance + fee

go/consensus/tendermint/apps/staking/auth.go
pro-wh: too complicated for this timeframe. going with the post-withdrawal check :+1:

pro-wh:
gonna try CI :raised_hands:

pro-wh:
hm ci is hanging on whatever comes after TestCrashingBackendDoNotInterfere

gonna see if I can reproduce that locally

pro-wh:
fixed a dos risk: there's now a single check for balance >= fee + min balance, shared between check tx and deliver tx. otherwise, an attacker may submit a high-gas-price transaction with fee <= balance < fee + min balance, which would get scheduled with high priority but always abort

pro-wh:
adding a way to mark some methods as exempt from the min balance check. the suggested use is for transactions routinely used by nodes, which (i) commonly have no balance and (ii) are known in the registry system

go/consensus/tendermint/apps/staking/state/gas.go
```diff
@@ -75,8 +76,13 @@ func AuthenticateAndPayFees(
 	if err != nil {
 		return fmt.Errorf("failed to fetch staking consensus parameters: %w", err)
 	}
-	if err = needed.Add(&params.MinTransactBalance); err != nil {
-		return fmt.Errorf("adding MinTransactBalance to fee: %w", err)
+	var minTransactBalanceExempt bool
+	if params.MinTransactBalanceExemptMethods[method] {
```
kostko, line 80: But if there's no registry check then someone can still submit these transactions (resulting in an account being created with a nonce) and they will fail but the account will remain?

go/staking/api/api.go
```diff
@@ -1103,6 +1103,7 @@ type ConsensusParameters struct { // nolint: maligned
 	GasCosts                          transaction.Costs                   `json:"gas_costs,omitempty"`
 	MinDelegationAmount               quantity.Quantity                   `json:"min_delegation"`
 	MinTransactBalance                quantity.Quantity                   `json:"min_transact_balance"`
+	MinTransactBalanceExemptMethods   map[transaction.MethodName]bool     `json:"min_transact_balance_exempt_methods"`
```
kostko, line 1106: Instead of these being parameters, should they rather be specified in code?

go/consensus/tendermint/apps/roothash/transactions.go
```diff
@@ -140,6 +136,23 @@ func (app *rootHashApplication) executorCommit(
 		return err
 	}
 
+	if ctx.IsCheckOnly() {
+		for i, commit := range cc.Commits {
+			if !commit.NodeID.Equal(ctx.TxSigner()) {
+				// we wouldn't be able to handle this...
```
pro-wh, line 142: I now realize we have the commit method that allows any transaction sender to send on behalf of any executor. so really that idea from before of "sender would have been registered anyway" doesn't even hold.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -185,6 +181,20 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
 		return err
 	}
 
+	if ctx.IsCheckOnly() {
+		// This method is exempt from the minimum balance requirement. We'll
+		// at least check that an entity allows this node though.
+		if !untrustedEntity.HasNode(untrustedNode.ID) {
```
pro-wh, line 187: I added this check, but an attacker could still repeatedly update their entity to allow new node accounts and then using RegisterNode to create new node accounts with nonces

go/consensus/tendermint/apps/roothash/transactions.go
kostko: Yes that is true, however currently the commits are not propagated otherwise so it will always be the node who submits them. We could however allow any node in the committee to submit commits (not just for itself).

Also note that this should then also be enforced during execution as now a bad proposer can bypass it.

go/consensus/tendermint/apps/staking/state/gas.go
```diff
@@ -24,6 +26,13 @@ type feeAccumulator struct {
 	balance quantity.Quantity
 }
 
+var MinTransactBalanceExemptMethods = map[transaction.MethodName]bool{
```
kostko, line 29: This should be in `go/consensus/api/transaction` or similar.

pro-wh:
we discussed more yesterday, and we'd like to extract the configurable minimum balance check, without the method exemptions

pro-wh:
in the process of adding a big e2e test

right now it passes, I think because the nodes are registered in the genesis state. if the test would run longer, they'd all fail to re-register

pro-wh:
in the process of adding a big e2e test

right now it passes, I think because the nodes are registered in the genesis state. if the test would run longer, they'd all fail to re-register

go/consensus/tendermint/apps/staking/state/gas.go
pro-wh: removing this from the pr, so gonna resolve this. let's keep it in mind if we later need exempt methods after all

go/consensus/tendermint/apps/roothash/transactions.go
pro-wh: removing this from the pr, so gonna resolve this. let's keep it in mind if we later need exempt methods after all

go/consensus/tendermint/apps/registry/transactions.go
pro-wh: we're removing this too

pro-wh:
tests added

go/consensus/tendermint/apps/staking/state/gas.go
```diff
@@ -94,8 +109,8 @@ func AuthenticateAndPayFees(
 
 	// Transfer fee to per-block fee accumulator.
 	feeAcc := ctx.BlockContext().Get(feeAccumulatorKey{}).(*feeAccumulator)
-	if err := quantity.Move(&feeAcc.balance, &account.General.Balance, &fee.Amount); err != nil {
-		return fmt.Errorf("staking: failed to pay fees: %w", err)
+	if err = quantity.Move(&feeAcc.balance, &account.General.Balance, &fee.Amount); err != nil {
+		return fmt.Errorf("staking: failed pay fees: %w", err)
```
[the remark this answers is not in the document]
pro-wh: shouldn't have

go/consensus/tendermint/apps/staking/transactions.go
```diff
@@ -84,6 +84,24 @@ func (app *stakingApplication) transfer(ctx *api.Context, state *stakingState.Mu
...
 		}
 
+		// Check against minimum balance.
+		if from.General.Balance.Cmp(&params.MinTransactBalance) < 0 {
+			ctx.Logger().Error("account balance too low",
+				"account_addr", fromAddr,
+				"account_balance", from.General.Balance,
+				"min_transact_balance", params.MinTransactBalance,
+			)
+			return nil, staking.ErrBalanceTooLow
```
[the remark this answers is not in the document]
pro-wh: yeah leme add some error wrapping (:

go/consensus/tendermint/apps/staking/state/state_test.go
```diff
@@ -715,6 +717,12 @@ func TestTransfer(t *testing.T) {
 	require.Empty(ctx.GetEvents(), "no events should be emitted")
 
 	// Transfer.
+	ctxEB := appState.NewContext(abciAPI.ContextEndBlock, now)
+	defer ctxEB.Close()
```
pro-wh, line 721: is this legal? to make another context while one is already active?

pro-wh: :+1: phew

go/consensus/tendermint/apps/staking/transactions.go
```diff
@@ -84,6 +84,24 @@ func (app *stakingApplication) transfer(ctx *api.Context, state *stakingState.Mu
...
 		}
 
+		// Check against minimum balance.
+		if from.General.Balance.Cmp(&params.MinTransactBalance) < 0 {
+			ctx.Logger().Error("after transfer source account balance too low",
+				"account_addr", fromAddr,
+				"account_balance", from.General.Balance,
+				"min_transact_balance", params.MinTransactBalance,
+			)
+			return nil, fmt.Errorf("after transfer source account: %w", staking.ErrBalanceTooLow)
```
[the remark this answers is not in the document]
pro-wh: I'll change to WithContext, thanks for the recommendation

kostko (approved):

PR #4427, opened 2022-01-11
Bump Tendermint

abukosek:
Closes #4249 and closes #4326.

TODO:
- [x] fix failing e2e tests
- [x] final polish
~- [ ] add support for tx prioritization~ -> will be done in a separate PR based on the existing 0.34
~- [ ] wait for tendermint to fix their stuff~ -> 0.35 release is broken beyond repair and will not be fixed by upstream

PR #4465, opened 2022-02-07
go/consensus: use typed attributes in all services

ptrus:


go/consensus/tendermint/api/api.go
```diff
@@ -117,7 +126,12 @@ func (bld *EventBuilder) Attribute(key, value []byte) *EventBuilder {
 // The typed attribute is automatically converted to a key/value pair where its EventKind is used
 // as the key and a CBOR-marshalled value is used as value.
 func (bld *EventBuilder) TypedAttribute(value TypedAttribute) *EventBuilder {
```
kostko, line 128: This method could automatically try to determine whether the passed attribute is also a `CustomTypedAttribute` to avoid an additional method.

go/registry/api/api.go
```diff
@@ -281,23 +281,51 @@ type EntityEvent struct {
 	IsRegistration bool           `json:"is_registration"`
 }
 
+// EventKind returns a string representation of this event's kind.
+func (e *EntityEvent) EventKind() string {
+	return "entity-event"
```
kostko, line 286: Let's use underscores everywhere.

kostko (commented):
Since we're here, Tendermint 0.35 requires all events to be actual strings and doesn't support binary data anymore (so values would need to be Base64-encoded). It also requires no dashes in attribute keys. Maybe we should make the change now to avoid breaking changes later (also see #4427)? cc @abukosek

ptrus:
> Maybe we should make the change now to avoid breaking changes later

Makes sense yes

go/registry/api/api.go
```diff
@@ -281,23 +282,64 @@ type EntityEvent struct {
 	IsRegistration bool           `json:"is_registration"`
 }
 
+// EventKind returns a string representation of this event's kind.
+func (e *EntityEvent) EventKind() string {
+	return "entity_event"
```
kostko, line 287: Why the `_event` suffix? Should probably remove it to be consitent with other places.

go/roothash/api/api.go
```diff
@@ -410,18 +412,59 @@ type ExecutorCommittedEvent struct {
 	Commit commitment.ExecutorCommitment `json:"commit"`
 }
 
+// EventKind returns a string representation of this event's kind.
+func (e *ExecutorCommittedEvent) EventKind() string {
+	return "executor_committ"
```
kostko, line 417:
```suggestion
	return "executor_commit"
```

kostko (approved):

PR #4466, opened 2022-02-08
staking: paranoid writeback changes

pro-wh:
1. eliminate some unnecessary Set*:
   - when short circuiting transfer to self, do not save the unmodified account
   - when starting to reclaim escrow, do not save the unaffected delegator account
2. rearrange Set* to lose tokens/shares, in case there's a way to make certain state Set* fail across validators. see below

transfer/withdraw saves from account first, then to account. if set from account succeeds and set to account fails, tokens disappear. fairly safe, "safe" referring to "less appealing to exploit"

reclaim escrow inconsistencies in partial failure, under new ordering:
1. set delegation succeeds, set escrow account fails -> delegator loses their shares, tokens remain in active escrow pool unreclaimable. fairly safe
2. set delegation succeeds, set escrow account succeeds, set debonding delegation fails -> tokens remain in debonding escrow pool, will never be released. fairly safe

add escrow seems fine as is, setting (1) delegator account, (2) escrow account, (3) delegation. token may disappear or get stuck in active escrow pool unreclaimable. fairly safe

do we want these changes? I can split these up if we want some and not others


go/consensus/tendermint/apps/staking/transactions.go
```diff
@@ -79,15 +79,14 @@ func (app *stakingApplication) transfer(ctx *api.Context, state *stakingState.Mu
 			return nil, err
 		}
 
+		if err = state.SetAccount(ctx, fromAddr, from); err != nil {
+			return nil, fmt.Errorf("failed to fetch account: %w", err)
```
Yawning, line 83: While you're here, probably change this?

go/consensus/tendermint/apps/staking/transactions.go
```diff
@@ -315,26 +314,14 @@ func (app *stakingApplication) reclaimEscrow(ctx *api.Context, state *stakingSta
...
+		return nil, staking.ErrForbidden
 	}
 
 	// Fetch escrow account.
-	//
-	// NOTE: Could be the same account, so make sure to not have two duplicate
-	//       copies of it and overwrite it later.
-	var from *staking.Account
-	if toAddr.Equal(reclaim.Account) {
-		from = to
```
Yawning, line 329: This was deliberate if somewhat wonky, but `to` is never mutated, so the removal appears to be ok.

Yawning (commented):
This "looks" ok, but I would feel better about it if our test cases covered the edge cases that lead to the weirdness in the first place.

kostko (commented):
The thinking behind the current setup is that these `Set` calls can only fail in case there is state corruption in which case all state updates are rolled back and consensus halts.

You could also use:
```golang
ctx = ctx.NewTransaction()
defer ctx.Close()
// ...
ctx.Commit()
```

pro-wh:
oh the transaction way would be nice. this being something that came up while thinking about adding a minimum balance check, it seems more correct to create a transaction and roll back if the minimum balance is not met

go/consensus/tendermint/apps/staking/transactions.go
pro-wh: right, to prevent creating multiple in-memory representations of the same record that can get out of sync. in the changed version, we no longer attempt to load the "to"/delegator account

pro-wh:
ok Set* rearrangements removed, NewTransaction + Commit added

pro-wh:
no, this is all wrong, we can't just start a transaction and keep using the same `state` object

PR #4470, opened 2022-02-09
Add support for node suspension due to runtime liveness failures

kostko:
TODO
* [x] Basic structures for tracking node liveness faults.
* [x] Update the scheduler service to ignore suspended nodes when electing committees.
* [x] Update the roothash service to track node liveness failure in the previous epoch.
* [x] Support for slashing.
* [x] Tests.

### DIFF
--- a/go/consensus/tendermint/apps/roothash/liveness.go
+++ b/go/consensus/tendermint/apps/roothash/liveness.go
@@ -1,0 +1,95 @@
+package roothash
+
+import (
+	"fmt"
+
+	beacon "github.com/oasisprotocol/oasis-core/go/beacon/api"
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
+	tmapi "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/api"
+	registryState "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/apps/registry/state"
+	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
+	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
+	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
+)
+
+// processLivenessStatistics checks the liveness statistics for the last epoch and penalizes any
+// nodes that didn't satisfy the liveness condition.
+func processLivenessStatistics(ctx *tmapi.Context, epoch beacon.EpochTime, rtState *roothash.RuntimeState) error {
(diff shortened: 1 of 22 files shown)


go/consensus/tendermint/apps/roothash/liveness.go
```diff
@@ -0,0 +1,94 @@
...
+		if fault.Failures >= maxFailures {
+			// Make sure to freeze forever if this would otherwise overflow.
+			if epoch > registry.FreezeForever-slashParams.FreezeInterval {
+				status.FreezeEndTime = registry.FreezeForever
+			} else {
+				status.FreezeEndTime = epoch + slashParams.FreezeInterval
+			}
+
+			// Slash if configured.
+			err = onRuntimeLivenessFailure(ctx, nodeID, &slashParams.Amount)
```
Yawning, line 82: This should be skipped if the penalty amount is zero to save quite a bit of work (Doing it in the callee is fine too).

go/consensus/tendermint/apps/roothash/liveness.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+// processLivenessStatistics checks the liveness statistics for the last epoch and penalizes any
+// nodes that didn't satisfy the liveness condition.
+func processLivenessStatistics(ctx *tmapi.Context, epoch beacon.EpochTime, rtState *roothash.RuntimeState) error {
+	if rtState.ExecutorPool == nil || rtState.LivenessStatistics == nil || rtState.Suspended {
+		return nil
+	}
+
+	// Skip evaluation if the number of total live rounds is below the set minimum.
```
Yawning, line 22: In order to catch pathological mis-configuration, this should either special case check for `TotalRounds == 0` or enforce `MinLiveRoundsForEvaluation >= 1`

go/consensus/tendermint/apps/roothash/liveness.go
kostko: Good idea, updated all of the roothash slashing functions to return early if the amount is zero.

go/consensus/tendermint/apps/roothash/liveness.go
kostko: Agreed, fixed.

Yawning (commented):
This needs fixgenesis support, because we presumably want to enable this behavior.

kostko:
My thinking was to leave it up to a runtime descriptor update.

Yawning (approved):

PR #4475, opened 2022-02-13
Document missing dependency (gcc-multilib) for Ubuntu setup

sbellem:
Problem encountered when trying to build `cipher-paratime` with SGX support.

See https://github.com/oasisprotocol/cipher-paratime/issues/19 for more details.

I don't know what the equivalent is for Fedora systems. I'll add it once I find out.


kostko (approved):

sbellem:
Thanks @tjanez!

PR #4477, opened 2022-02-14
go/runtime/txpool: Refactor runtime scheduling queue

kostko:
* Remove the `algorithm` field from the runtime descriptor (it is unused).
* Refactor the runtime txpool as it can be simplified now that there is just one "algorithm".


go/registry/api/runtime.go
```diff
@@ -115,9 +112,6 @@ func (e *ExecutorParameters) ValidateBasic() error {
 
 // TxnSchedulerParameters are parameters for the runtime transaction scheduler.
 type TxnSchedulerParameters struct {
```
ptrus, line 114: `fixgenesis` is not needed for this removal right, as the old field will just get skipped when deserializing the json, right?

kostko: That's right.

Yawning (approved):

PR #4480, opened 2022-02-16
runtime: Bump Rust dependencies

kostko:


runtime/src/consensus/tendermint/verifier.rs
```diff
@@ -590,7 +590,7 @@ struct InsecureClock;
 
 impl components::clock::Clock for InsecureClock {
     fn now(&self) -> Time {
-        Time(time::insecure_posix_system_time().into())
+        Time::from_unix_timestamp(time::insecure_posix_time(), 0).unwrap()
```
ptrus, line 593: Hehe, guess i can close https://github.com/oasisprotocol/oasis-core/pull/4480/files :grin:

kostko: LOL

runtime/Cargo.toml
```diff
@@ -51,8 +51,8 @@ num-bigint = "0.4"
 num-traits = "0.2.14"
 bech32 = "0.8.1"
```
ptrus, line 52: Tendermint bumps seem to be missing

kostko: Yeah because 0.23.1 implies 0.23.5 by default. Bad Tendermint, putting breaking changes into a patch release. Will update Cargo.toml as well.

kostko: Done.

ptrus (approved):

PR #4482, opened 2022-02-16
tendermint-rs: bump to 0.23.5

ptrus:
Based on: https://github.com/oasisprotocol/oasis-core/pull/4481


ptrus:
Will be updated in https://github.com/oasisprotocol/oasis-core/pull/4480/files

Issue #2987, opened 2020-06-09
Bump Rust toolchain to include LVI mitigations


kostko:
LVI mitigations just landed in Rust nightly (see https://github.com/rust-lang/rust/commit/fd4b177aabb9749dfb562c48e47379cea81dc277), so we need to bump the toolchain again.

PR #4491, opened 2022-02-17
runtime: allow SW_HARDENING_NEEDED IAS quote status

ptrus:
`SW_HARDENING_NEEDED` should be allowed since it only indicates that additional SW hardening might be needed for the platform.
Discovered this in the failing SGX E2E: due to INTEL-SA-00334 (which is overed by https://github.com/oasisprotocol/oasis-core/issues/2987)

### DIFF
--- a/go/common/node/node.go
+++ b/go/common/node/node.go
@@ -478,15 +478,15 @@ type SGXConstraints struct {
 	// AllowedQuoteStatuses are the allowed quote statuses for the node
 	// to be scheduled as a compute worker.
 	//
-	// Note: QuoteOK is ALWAYS allowed, and does not need to be specified.
+	// Note: QuoteOK and QuoteSwHardeningNeeded are ALWAYS allowed, and do not need to be specified.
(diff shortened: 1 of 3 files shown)


kostko (commented):
This also needs to be updated in the Go part, see `go/common/node/node.go` line 489.

go/common/node/node.go
```diff
@@ -478,15 +478,15 @@ type SGXConstraints struct {
 	// AllowedQuoteStatuses are the allowed quote statuses for the node
 	// to be scheduled as a compute worker.
 	//
-	// Note: QuoteOK is ALWAYS allowed, and does not need to be specified.
+	// Note: QuoteOK and QuoteSwHardeningNeeded are ALWAYS allowed, and does not need to be specified.
```
kostko, line 481:
```suggestion
	// Note: QuoteOK and QuoteSwHardeningNeeded are ALWAYS allowed, and do not need to be specified.
```

kostko (approved):

Issue #4460, opened 2022-02-04
rust: fix `staking_params_are_empty`


ptrus:
https://github.com/oasisprotocol/oasis-core/blob/c31ac9da4f8ca27aec130db1c85b2c71a738383a/runtime/src/consensus/registry.rs#L292-L294

^ should be updated with new fields. Also add a test for consistent runtime struct serialization. 

_Originally posted by @kostko in https://github.com/oasisprotocol/oasis-core/pull/4455#discussion_r799268412_

PR #4494, opened 2022-02-18
runtime/registry: fix `staking_params_are_empty`

ptrus:
Fixes: https://github.com/oasisprotocol/oasis-core/issues/4460


kostko (approved):
Nice!

PR #4509, opened 2022-02-22
backport 21.3.x - ci: show dirtiness after regenerating protos

pro-wh:
just wondering if what I'm seeing locally is the same as we see on CI

backport of #4598


kostko (approved):

pro-wh:
thanks for reviewing. I should re-open on master though if we want this

pro-wh:
weird, this time it didn't even trigger that dirty-after-generate error

pro-wh:
opened #4598 with a copy of this on master. also backported to 22.0.x in #4599, why not

PR #4513, opened 2022-02-24
go/common/dynlib: Embed dynlib

kostko:


Yawning:
For posterity's sake, the author of the package is ok with this (and suggested it).

Yawning (approved):

PR #4524, opened 2022-02-28
go/worker/common: Report locally supported runtime host versions

### LINKED ISSUE #4436: Report runtime version in oasis-node control status for runtime client nodes

kostko:
Fixes #4436

### DIFF
--- a/.changelog/4436.feature.md
+++ b/.changelog/4436.feature.md
@@ -1,0 +1,1 @@
+go/worker/common: Report locally supported runtime host versions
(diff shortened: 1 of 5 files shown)


tjanez (commented):
Could you add a change log fragment for liveliness as well and explain the output briefly?

tjanez (approved):

ptrus (approved):

PR #4557, opened 2022-03-09
runtime: Commit check transaction results for subsequent batches

kostko:


pro-wh (approved):
discussed on slack, we'd like to try this

ptrus (approved):

PR #4561, opened 2022-03-11
go/storage: reduce storage checkpointing

ptrus:
Should the consensus checkpoints also be disabled by default?


kostko:
We can probably leave consensus checkpoints as they are since consensus state is fairly small.

go/oasis-test-runner/oasis/args.go
```diff
@@ -466,6 +466,13 @@ func (args *argBuilder) workerStorageDebugDisableCheckpointSync(disable bool) *a
 	return args
 }
 
+func (args *argBuilder) workerStorageCheckpointSyncEnabled(enable bool) *argBuilder {
```
kostko, line 469: This should be renamed as it doesn't enable checkpoint sync but rather the creation of checkpoints.

go/storage/mkvs/checkpoint/checkpointer.go
```diff
@@ -285,17 +290,22 @@ func (c *checkpointer) worker(ctx context.Context) {
 		c.logger.Debug("storage checkpointer terminating")
 	}()
 
-	// Use a ticker to avoid checking for checkpoints too often.
-	ticker := time.NewTicker(c.cfg.CheckInterval)
-	defer ticker.Stop()
+	getCheckInterval := func() time.Duration {
```
kostko, line 293: This is the same as:
https://github.com/oasisprotocol/oasis-core/blob/a9e863a7cc07c9662bfced2a473c0e3b67473a83/go/consensus/tendermint/full/helpers.go#L9

Let's move it somewhere under `go/common` and use it in both places.

kostko (approved):

PR #4574, opened 2022-03-17
go/worker/registration: Add a random re-registration delay

Yawning:
This is hard coded to 5%, and will not be enabled if the mock backend is
in use so that the tests hopefully do not need alterations.

### DIFF
--- a/go/worker/registration/worker.go
+++ b/go/worker/registration/worker.go
@@ -3,6 +3,8 @@ package registration
 import (
 	"context"
 	"fmt"
+	"math"
+	"math/rand"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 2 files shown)


go/worker/registration/worker.go
```diff
@@ -236,11 +245,46 @@ func (w *Worker) registrationLoop() { // nolint: gocyclo
 		case <-w.consensus.Synced():
 		}
 		w.logger.Debug("consensus synced, entering registration loop")
+
+		beaconParameters, err := w.beacon.ConsensusParameters(w.ctx, consensus.HeightLatest)
+		if err != nil {
+			w.logger.Error("failed to query beacon parameters")
```
kostko, line 251: Would be good to also log the error.

go/worker/registration/worker.go
```diff
@@ -310,6 +356,16 @@ Loop:
 		case <-w.stopRegCh:
 			w.logger.Info("node deregistration and eventual shutdown requested")
 			return
+		case block := <-blockCh:
+			if block.Height != reregisterHeight {
```
kostko, line 360:
```suggestion
			if block.Height < reregisterHeight {
```

kostko (approved):

PR #4579, opened 2022-03-18
Don't clear tx pool on epoch transitions, add txsync protocol

kostko:


go/oasis-test-runner/scenario/e2e/runtime/runtime_client_kv_enc.go
```diff
@@ -150,37 +149,6 @@ func (cli *KeyValueEncTestClient) workload(ctx context.Context) error {
 		return fmt.Errorf("key still exists in database after removal: '%v'", resp)
 	}
 
-	// TODO: Someone that cares, that thinks this is still relevant
```
Yawning, line 153: I guess no one cared. :(

Yawning (approved):

go/oasis-test-runner/scenario/e2e/runtime/runtime_client_kv_enc.go
kostko: Yeah, it is not relevant as the way this is handled in e.g. the SDK is for the runtime to expose a query for it (and then the runtime can deal with EnclaveRPC). EnclaveRPC is no longer exposed via gRPC.

PR #4582, opened 2022-03-21
runtime/host/sgx: Log the advisory URLs on attestation failure

Yawning:


kostko (approved):

Yawning:
The commit and pr should read "advisory IDs", oops.

PR #4591, opened 2022-03-23
go/signature: apply options on registered contexts 

pro-wh:
when allowing unregistered signature contexts, still apply options for contexts that are registered


Yawning (approved):

kostko (approved):
Let's also backport this.

pro-wh:
gonna add a test to cover that unregistered malformed context branch while I'm here

pro-wh:
backports opened :+1:

PR #4593, opened 2022-03-23
go/registry: Remove support for DeprecatedBeacon

### LINKED ISSUE #4393: go/common/node: Remove the old PVSS public key placeholder
Once the network has transitioned to using VRF-driven elections, on the next major upgrade, remove the backward compatibility placeholder `DeprecatedBeacon` field from the node descriptor.

### LINKED ISSUE #4394: go/registry: Make submitting a VRF signing key mandatory
Once the network has transitioned to using VRF-driven elections, on the next major upgrade, mandate that node descriptors contain a valid VRF public key at genesis time.  Non-genesis registration tx-es currently already require a VRF public key.

Yawning:
The PVSS backend is no longer present in 22.x and so the field is now
removed, and even genesis registrations without a VRF signing key will
be rejected.

Fixes #4394
Fixes #4393

### DIFF
--- a/.changelog/4394.internal.md
+++ b/.changelog/4394.internal.md
@@ -1,0 +1,5 @@
+go/registry: Remove support for DeprecatedBeacon
+
+The PVSS backend is no longer present in 22.x and so the field is now
+removed, and even genesis registrations without a VRF signing key will
+be rejected.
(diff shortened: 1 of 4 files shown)


kostko (commented):
This change is consensus breaking as if someone does include the field, some nodes will accept it while others would reject it. I would rather avoid consensus breaking changes at this time.

Yawning:
Superseded by #4668

PR #4594, opened 2022-03-23
Split runtime/common/crypto into a crate

nhynes:
This PR simply moves the crypto utilities and some other dependencies to a crate so that they're easier to consume by downstream libraries (in this case, a Rust gateway). The alternatives are to copy the code verbatim; or else pull down altogether too many dependencies also used by the runtime.

The gateway just needs the ECDH box functions, so it's not awful to copy/re-implement those, but then it's harder to apply patches from this repo (but the code is pretty static and well tested, so maybe nothing will change).


Yawning (approved):
Sure, this is fine by me.

kostko (approved):

nhynes:
Thanks for the review. I was holding off on merging this because it wasn't 1000% clear that it'd be useful, but it turns out that it will be now. I'll also need to reboot the rust client PR in the oasis-sdk repo to fetch the runtime public key, so this week will be one of great closure :)

nhynes:
I ended up not using this because the crate functions invoke the global allocator, which isn't preferred. I copied the one function out, which should be fine.

PR #4596, opened 2022-03-23
go/oasis-node/cmd/control: Add runtime-stats

### LINKED ISSUE #4425: go/oasis-node: Integrate the `tools/runtime-stats` functionality
We have a nice tool for visualization of runtime stats in the tools repo.  Given that we have emerald and cipher running in production, and third party node operators, we should offer the ability to diagnose problems with executor node deployment.

While we are doing this, we should also expose "useful" metrics for executor status via prometheus (if not present already) to allow for automated monitoring.

Yawning:
This is a cleaned up version of the command in the tools repository.

Fixes #4425

### DIFF
--- a/.changelog/4596.feature.md
+++ b/.changelog/4596.feature.md
@@ -1,0 +1,3 @@
+go/oasis-node/cmd/control: Add runtime-stats
+
+This command allows querying historical runtime statistics.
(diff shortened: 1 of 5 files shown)


gw0:
If I see correctly, this PR only implements the `runtime-status` command to get on-demand statistics for all nodes for a runtime in a specific interval. Node operators are more concerned with the current health of its own node (not historical and not for other nodes).

It would probably be better to continuously monitor these runtime stats of all runtimes running on the current node. These could then be exposed in `oasis-node control status` and also as Prometheus metrics (mentioned in #4425).

kostko:
@gw0 Current health is already reported as part of the `control status` output (in 22.0) as the consensus layer is tracking liveness statistics. We will also be adding a Prometheus metric.

kostko (approved):

Issue #4587, opened 2022-03-22
Add metrics for number of nodes in runtime committees


gw0:
##### SUMMARY

We already expose a metric `tendermint_consensus_validators` for the current number of active validators. To monitor paratimes it would make sense to also expose metrics for the current number of compute and storage nodes in the paratime committees.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

- `go/worker/...`?

##### ADDITIONAL INFORMATION

Monitoring the health of paratime committees.

gw0:
You are right. I guess the proper way to monitor the health of runtimes is https://github.com/oasisprotocol/oasis-core/issues/4425.

PR #4610, opened 2022-03-28
go/worker: add executor livness metrics

ptrus:
Adds liveness (and some other useful) executor prometheus metrics.

TODO:
- [x] test locally (ensure metrics are emitted)


go/worker/common/committee/node.go
```diff
@@ -62,13 +66,54 @@ var (
...
+		[]string{"runtime"},
+	)
+	livenessTotalRounds = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "oasis_worker_liveness_total_rounds",
+			Help: "Number of total rounds in last epoch.",
+		},
+		[]string{"runtime"},
+	)
+	livenessLiveRounds = prometheus.NewGaugeVec(
```
kostko, line 97: Could we also add a percentage (live / total). Maybe it should report 100% if the node is not in the committee just to make alerting easier?

ptrus: Added

gw0:
Awesome, number of peers per runtime is actually what I was hoping for in https://github.com/oasisprotocol/oasis-core/issues/4587.

kostko (approved):
Nice! This should also be backported to 22.0.x.

go/worker/common/committee/node.go
```diff
@@ -651,6 +708,79 @@ func (n *Node) worker() {
...
+
+	n.logger.Debug("updating periodic worker node metrics")
+
+	epoch := n.Group.GetEpochSnapshot()
+	cmte := epoch.GetExecutorCommittee()
+	if cmte == nil {
+		return
+	}
+
+	executorCommitteePeers.With(labels).Set(float64(len(cmte.Peers)))
```
kostko, line 732: Actually this is not the right peers to look at. You probably want to see the list of P2P peers? That one should be in `n.P2P.Peers(runtimeID)`.

ptrus: updated

kostko (approved):

PR #4616, opened 2022-03-30
go/upgrade: Add empty upgrade handler

kostko:


go/upgrade/migrations/empty.go
```diff
@@ -0,0 +1,24 @@
+package migrations
+
+const (
+	// EmptyHandler is the name of the upgrade handler that doesn't perform any updates.
+	EmptyHandler = "empty"
```
ptrus, line 5: Maybe add a comment that this is useful for upgrades that bump protocol versions, but don't need any other migrations.

ptrus (approved):

go/upgrade/migrations/empty.go
kostko: Done.

tjanez (approved):

ptrus:
Could also add a simple E2E test for the upgrade handler:
https://github.com/oasisprotocol/oasis-core/blob/6a493b20d6aaaf9ad6636f8651a8896f93df0835/go/oasis-test-runner/scenario/e2e/upgrade.go#L140-L145

ptrus (approved):

PR #4617, opened 2022-03-30
go/oasis-node/cmd: Allow using non local gRPC connections

tjanez:
Require server-side TLS for non unix-socket gRPC connections.

### DIFF
--- a/.changelog/4617.feature.md
+++ b/.changelog/4617.feature.md
@@ -1,0 +1,3 @@
+go/oasis-node/cmd: Allow using non local gRPC connections
+
+Require server-side TLS for non unix-socket gRPC connections.
--- a/go/oasis-node/cmd/common/grpc/grpc.go
+++ b/go/oasis-node/cmd/common/grpc/grpc.go
@@ -7,11 +7,13 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
 	"github.com/spf13/viper"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 
 	cmnGrpc "github.com/oasisprotocol/oasis-core/go/common/grpc"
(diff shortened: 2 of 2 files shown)


go/oasis-test-runner/scenario/e2e/stake_cli.go
```diff
@@ -63,6 +63,9 @@ const (
 
 	// Testing escrow account public key (hex-encoded).
 	escrowPubkeyHex = "6ea5328f943ef6f66daaed74cb0e99c3b1c45f76307b425003dbc7cb3638ed35"
+
+	// Oasis Protocol Foundation's public gRPC endpoint.
+	publicGrpcAddr = "grpc.oasis.dev:443"
```
kostko, line 68: Should we disable this test by default? Otherwise this will make CI fail if the public endpoint is unavailable.

Yawning: oasis-core intentionally is network instance neutral.  I'm not sure this feature warrants changing that policy...

kostko: This is a test, not some default.

Yawning:
Ultimately I don't see the point of this.  The only time when this really is an issue is when voting on governance, and the current voting period which ends in a day should be the last governance vote where this tooling needs to be used in the first place.

tldr; use the sdk-cli.  If the sdk-cli doesn't provide what you need, go add to it.

go/oasis-test-runner/scenario/e2e/stake_cli.go
tjanez: I went ahead and removed the test since it would complicate things and is not really that crucial.

tjanez:
> Ultimately I don't see the point of this. The only time when this really is an issue is when voting on governance, and the current voting period which ends in a day should be the last governance vote where this tooling needs to be used in the first place.

I agree that the future of CLI tooling is to migrate to the new Oasis CLI we are developing in the Oasis SDK repo but I think this small change will still improve the UX until people migrate to that.

Another example where this would have impact is that people who don't have access to a synced node would like to use State Sync and [need the list of nodes that offer that](https://docs.oasis.dev/general/run-a-node/advanced/sync-node-using-state-sync#obtaining-addresses-of-oasis-nodes-publicly-exposed-grpc-endpoints).

With this change, they could use:
```
oasis-node registry node list -v -a grpc.oasis.dev:443 | \
  jq 'select(.roles | contains("consensus-rpc")) | .tls.addresses' 
 ```

kostko (approved):

PR #4640, opened 2022-04-04
go/worker/compute: Improve proposed transaction handling

kostko:


go/runtime/txpool/txpool.go
```diff
@@ -303,9 +359,10 @@ func (t *txPool) GetPrioritizedBatch(offset *hash.Hash, limit uint32) []*transac
 
 func (t *txPool) GetKnownBatch(batch []hash.Hash) ([]*transaction.CheckedTransaction, map[hash.Hash]int) {
 	t.schedulerLock.Lock()
-	defer t.schedulerLock.Unlock()
+	schedulerQueue := t.schedulerQueue
```
Yawning, line 362: This is spooky, I take it this is to avoid adding `GetKnownBatchLocked`?

go/runtime/txpool/txpool.go
```diff
@@ -78,6 +78,17 @@ type TransactionPool interface {
...
+	// SubmitProposedBatch adds the given (possibly new) transaction batch into the current
+	// proposal queue.
+	SubmitProposedBatch(batch [][]byte)
+
+	// PromoteProposedBatch promotes the specified transactions that are already in the transaction
+	// pool into the current proposal queue.
+	PromoteProposedBatch(batch []hash.Hash)
+
+	// ClearProposedBatch clears the proposal queue.
+	ClearProposedBatch()
```
Yawning, line 90: This is the only thing that clears the proposal queue.  `TransactionPool.Clear` does not.  Is this intentional?

kostko: `Clear` is not currently used anywhere, but yeah it makes sense that it should also clear the proposal queue. Will fix.

go/runtime/txpool/txpool.go
kostko: Fixed.

Yawning (approved):

ptrus (approved):

PR #4644, opened 2022-04-05
go/runtime/host/sgx: Don't trample over the actual error

Yawning:
This is largely cosmetic since we wrap a nil err (so the failure happens as it should), but might as well fix it.


pro-wh (approved):
ok

PR #4648, opened 2022-04-07
consensus/tendermint/db/badger: Fix WriteSync for large txns

abukosek:
If a transaction is too big for Badger, the correct way to handle this (according to Badger docs) is to commit the current contents of the transaction and start a new one instead of failing the entire transaction.
This PR adds proper handling for this situation.

### DIFF
--- a/go/consensus/tendermint/db/badger/badger.go
+++ b/go/consensus/tendermint/db/badger/badger.go
@@ -474,11 +474,31 @@ func (ba *badgerDBBatch) WriteSync() error {
 
 	for _, cmd := range ba.cmds {
 		if err := cmd.Execute(tx); err != nil {
-			ba.db.logger.Error("failed to execute command in Tx",
-				"err", err,
-				"cmd", cmd,
-			)
-			return err
+			if err == badger.ErrTxnTooBig {
+				// If a transaction is too big for Badger, the correct way
+				// to handle this (according to Badger docs) is to commit the
+				// current contents of the transaction and start a new one,
+				// instead of failing the entire transaction.
+				//
+				// See the following link for more information:
+				// https://dgraph.io/docs/badger/get-started/#read-write-transactions
+				if grr := tx.Commit(); grr != nil {
+					ba.db.logger.Error("failed to commit Tx",
+						"err", grr,
+					)
+					return grr
+				}
+				tx = ba.db.db.NewTransaction(true)
+				err = cmd.Execute(tx)
+			}
+
+			if err != nil {
+				ba.db.logger.Error("failed to execute command in Tx",
+					"err", err,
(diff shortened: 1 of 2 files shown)


go/consensus/tendermint/db/badger/badger.go
```diff
@@ -474,11 +474,24 @@ func (ba *badgerDBBatch) WriteSync() error {
 
 	for _, cmd := range ba.cmds {
 		if err := cmd.Execute(tx); err != nil {
-			ba.db.logger.Error("failed to execute command in Tx",
-				"err", err,
-				"cmd", cmd,
-			)
-			return err
+			if err == badger.ErrTxnTooBig {
```
ptrus, line 477: Since this looks a bit unconventional, maybe add a similar comment as the one in the issue description here in code as well. Together with a reference to the badger docs.

Yawning (approved):

go/consensus/tendermint/db/badger/badger.go
abukosek: Done :)

ptrus (approved):

go/consensus/tendermint/db/badger/badger.go
ptrus: Oh, but does this mean that we can end up with some "commands" being committed and some not? Since, this actually commits the old transaction and starts a new one.

Is this ok with the expected semantics of `WriteSync`?

abukosek: Good point.  There's some discussion on Badger's issue tracker here: https://github.com/dgraph-io/badger/issues/1242

It doesn't look like there's any other way to implement this.

So far, the only time I've encountered this error is when running Tendermint's 0.34->0.35 migration scripts, which consistently fail here because of `ErrTxnTooBig`.  It seems all normal transactions are small enough to go through with sync semantics unchanged, but really large ones can indeed commit half-way.  Given that we've never encountered this before, it looks like the only place where really large transactions are currently used is in the migration scripts and those are written in a way that handles partially-migrated databases, so it should be fine.

If we want to be extra safe, I can close this PR and change the Tendermint migration scripts to use `WriteBatch` instead, which does the same thing internally, except it doesn't have atomic semantics.

@kostko What do you think?

Yawning: If we're confident that we'll never see this issue apart from in the migration scripts, that's fine as well.

kostko: Yeah I think it is ok. Would be safer if there was an explicit flag (only enabled during migrations), but seeing how the DB config has this Tendermint's interface it would be rather clunky.

PR #4653, opened 2022-04-07
go/consensus/api/transactions: Fix OpenRawTransactions

Yawning:
Oops, im dumb.


pro-wh (approved):
https://github.com/oasisprotocol/oasis-core/blob/v22.1.2/go/common/crypto/signature/batch_verification.go#L43 cross reference confirmed

kostko (approved):

PR #4658, opened 2022-04-11
go/consensus/tendermint/apps/beacon: Simulate VRF gas earlier

Yawning:
We shouldn't need this, but just in case.


Yawning:
This isn't hugely important and can wait for a proper fix.

PR #4665, opened 2022-04-13
go/runtime/txpool: Limit outstanding transactions per sender

kostko:


go/runtime/txpool/check_queue.go
```diff
@@ -1,155 +1,85 @@
...
 
-// GetBatch gets a batch of transactions from the queue.
-func (q *checkTxQueue) GetBatch() []*pendingTx {
-	q.Lock()
-	defer q.Unlock()
+func (cq *checkTxQueue) retryBatch(pcts []*PendingCheckTransaction) {
+	cq.l.Lock()
+	defer cq.l.Unlock()
+
+	// NOTE: This is meant for retries so it ignores the size limit on purpose.
```
Yawning, line 37: Is there something in place to prevent unbounded growth?  I guess with the accounting in add, the queue size will be at most `2 *cq.maxSize`?

kostko: So this is only used for reinserting previously poped transactions (which are at most "batch size" long). In the worst case this could overflow the queue to be twice the max size (in case of check batches being that big -- which they are not by default) when the queue is full and batch checks are failing while new transactions are being queued.

go/runtime/txpool/txpool.go
```diff
@@ -92,29 +87,22 @@ type TransactionPool interface {
...
 	// Offset specifies the transaction hash that should serve as an offset when returning
 	// transactions from the pool. Transactions will be skipped until the given hash is encountered
 	// and only following transactions will be returned.
-	GetPrioritizedBatch(offset *hash.Hash, limit uint32) []*transaction.CheckedTransaction
+	GetPrioritizedBatch(offset *hash.Hash, limit uint32) []*Transaction
 
 	// GetKnownBatch gets a set of known transactions from the transaction pool.
 	//
 	// For any missing transactions nil will be returned in their place and the map of missing
 	// transactions will be populated accoordingly.
```
ptrus, line 100:
```suggestion
	// transactions will be populated accordingly.
```

ptrus (approved):

PR #4668, opened 2022-04-14
go/consensus/tendermint/apps/beacon: Breaking VRF fixes

### LINKED ISSUE #4393: go/common/node: Remove the old PVSS public key placeholder

### LINKED ISSUE #4394: go/registry: Make submitting a VRF signing key mandatory

### LINKED ISSUE #4667: go/consensus/tendermint/apps/beacon: Gas use happens too late
The VRF proof submission tx has gas use happen too late, and can in certain circumstances underestimate the amount of gas the tx will cost.  This is "ok" for now becaues such transactions would never have succeeded in the first place, but should be fixed.

### LINKED ISSUE #4838: common/crypto/signature: Use a newer version of ECVRF
When I implemented ECVRF in curve25519-voi, the IETF draft was less than version 11.  As of version 11, after leaving things largely unchanged for years, they broke proof backward compatibility.

* [x] The next time we bump the curve25519-voi import, switch to using the v10 backward compatible proof/verify routines. (#4839)
* [ ] Whenever we can jettison beacon state (so a dump-restore hard fork) switch to using the v13 style proofs.

Yawning:
This collects all the VRF related changes that are consensus breaking in one branch so I only have one thing to rebase when I can merge all of these.  Note that none of these fixes are security critical.

Fixes #4394
Fixes #4393
Fixes #4667
Fixes #4838


kostko:
Let's rebase now that master can break consensus.

kostko (approved):

PR #4669, opened 2022-04-14
go/control/status: Add fields for quick overview of node status

ptrus:
TODO:
- [x] not sure if current runtime worker statuses make sense, maybe some should be added as well (e.g. runtime suspended)

### DIFF
--- a/go/consensus/api/api.go
+++ b/go/consensus/api/api.go
@@ -4,6 +4,7 @@ package api
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"time"
 
@@ -202,8 +203,58 @@ type Vote struct {
 	VotingPower   uint64              `json:"voting_power"`
 }
 
+// StatusState is the concise status state of the consensus backend.
+type StatusState uint8
+
+var (
+	// StatusStateReady is the ready status state.
+	StatusStateReady StatusState = 0
+	// StatusStateSyncing is the syncing status state.
+	StatusStateSyncing StatusState = 1
+)
+
+// String returns a string representation of a status state.
+func (s StatusState) String() string {
+	switch s {
+	case StatusStateReady:
+		return "ready"
+	case StatusStateSyncing:
+		return "syncing"
+	default:
+		return "[invalid status state]"
+	}
+}
+
+// MarshalText encodes a StatusState into text form.
+func (s StatusState) MarshalText() ([]byte, error) {
+	switch s {
+	case StatusStateReady:
+		return []byte(StatusStateReady.String()), nil
+	case StatusStateSyncing:
+		return []byte(StatusStateSyncing.String()), nil
+	default:
+		return nil, fmt.Errorf("invalid StatusState: %d", s)
+	}
+}
+
+// UnmarshalText decodes a text slice into a StatusState.
+func (s *StatusState) UnmarshalText(text []byte) error {
+	switch string(text) {
+	case StatusStateReady.String():
+		*s = StatusStateReady
+	case StatusStateSyncing.String():
+		*s = StatusStateSyncing
+	default:
+		return fmt.Errorf("invalid StatusState: %s", string(text))
+	}
+	return nil
+}
+
 // Status is the current status overview.
 type Status struct { // nolint: maligned
+	// Status is an concise status of the consensus backend.
+	Status StatusState `json:"status"`
+
 	// Version is the version of the consensus protocol that the node is using.
 	Version version.Version `json:"version"`
 	// Backend is the consensus backend identifier.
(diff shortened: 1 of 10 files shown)


Yawning:
D: https://github.com/oasisprotocol/oasis-sdk/blob/main/cli/cmd/inspect/node_status.go

go/consensus/tendermint/full/full.go
```diff
@@ -774,6 +774,7 @@ func (t *fullService) GetStatus(ctx context.Context) (*consensusAPI.Status, erro
 	status.ChainContext = t.genesis.ChainContext()
 	status.GenesisHeight = t.genesis.Height
 	if t.started() {
+		status.Status = "Ready"
```
kostko, line 777: Shouldn't this check if `t.syncedCh` is closed? Because a node can be started and still syncing?

go/consensus/api/api.go
```diff
@@ -204,6 +204,9 @@ type Vote struct {
 
 // Status is the current status overview.
 type Status struct { // nolint: maligned
+	// Status is an concise status of the consensus backend.
+	Status string `json:"string"`
```
kostko, line 208: Probably should make its own type with some constants instead of this being directly a string.

go/worker/common/api/api.go
```diff
@@ -7,6 +7,9 @@ import (
 
 // Status is the common runtime worker status.
 type Status struct {
+	// Status is an concise status of the committee node.
+	Status string `json:"status"`
```
kostko, line 11: Probably should make its own type with some constants instead of this being directly a string.

go/worker/common/committee/node.go
```diff
@@ -621,6 +658,7 @@ func (n *Node) worker() {
 		)
 		return
 	}
+	atomic.StoreUint32(&n.hostedRuntimeProvisioned, 1)
```
kostko, line 661: The above doesn't block until successful provisioning, so this needs to be deferred until a "started" event is received from the runtime (in `handleRuntimeHostEvent`).

go/consensus/tendermint/full/full.go
ptrus: Yeah of course.

go/consensus/api/api.go
ptrus: Added

go/worker/common/api/api.go
ptrus: Added

go/worker/common/committee/node.go
ptrus: Updated

go/consensus/api/api.go
```diff
@@ -202,8 +203,58 @@ type Vote struct {
...
+	StatusStateReady StatusState = 0
+	// StatusStateSyncing is the syncing status state.
+	StatusStateSyncing StatusState = 1
+)
+
+// String returns a string representation of a status state.
+func (s StatusState) String() string {
+	switch s {
+	case StatusStateReady:
+		return "Ready"
```
kostko, line 220:
```suggestion
		return "ready"
```

go/consensus/api/api.go
```diff
@@ -202,8 +203,58 @@ type Vote struct {
...
+	case StatusStateSyncing:
+		return "Syncing"
```
kostko, line 222:
```suggestion
		return "syncing"
```

go/worker/common/api/api.go
```diff
@@ -1,12 +1,112 @@
...
+	StatusStateWaitingWorkersInit StatusState = 6
+	// StatusStateRuntimeSuspended is the runtime suspended status state.
+	StatusStateRuntimeSuspended StatusState = 7
+)
+
+// String returns a string representation of a status state.
+func (s StatusState) String() string {
+	switch s {
+	case StatusStateReady:
+		return "Ready"
```
kostko, line 36:
```suggestion
		return "ready"
```

go/worker/common/api/api.go
```diff
@@ -1,12 +1,112 @@
...
+	case StatusStateWaitingConsensusSync:
+		return "Waiting for consensus sync"
```
kostko, line 38:
```suggestion
		return "waiting for consensus sync"
```
(and so on)

kostko (commented):
Looks good, just some minor nits.

go/consensus/api/api.go
```diff
@@ -202,8 +203,58 @@ type Vote struct {
 	VotingPower   uint64              `json:"voting_power"`
 }
 
+// StatusState is the concise status state of the consensus backend.
+type StatusState int
```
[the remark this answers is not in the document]
ptrus: updated

kostko (approved):

PR #4681, opened 2022-04-22
go/runtime/txpool: transactions in incoming messages

pro-wh:
1. roothash incoming messages' data is no longer arbitrary bytes. it is now a runtime transaction
2. expand txpool to allow multiple tx queues. each queue can have its own way of organizing transactions. the existing "schedule queue" is now called the main queue
3. add a queue that exposes roothash incoming messages' transactions
4. while we're at it, add a queue for local transactions that we want to prioritize and preserve the order of. may come in handy later
5. add a mutex so that we don't recheck/republish/schedule at the same time


go/consensus/tendermint/apps/roothash/transactions.go
```diff
@@ -379,17 +381,24 @@ func (app *rootHashApplication) submitMsg(
...
+		Fee:         msg.Fee,
+		Tokens:      msg.Tokens,
+		Transaction: msg.Transaction,
+		Data:        msg.Data,
 	}
 	if err = state.SetIncomingMessageInQueue(ctx, rtState.Runtime.ID, inMsg); err != nil {
 		return err
 	}
 
+	if msg.Transaction != nil {
```
kostko, line 396: Huh? Why are you changing the consensus layer? You probably want to put this somewhere in `go/worker/compute` where it gets incoming messages, e.g.:
https://github.com/oasisprotocol/oasis-core/blob/9f399d01a3c17358a0bf278a3d6a5392deda20e5/go/worker/compute/executor/committee/node.go#L934-L938

pro-wh: ah yeah that's a good idea to do it for the entire batch. has to be a little before this exact point you link to though, because this point comes after the node fully prepares a [][]byte of transaction input data. we'll need to replicate the transactions in the txpool before that

pro-wh:
simplified map of executor worker

image

<details>
<summary>source</summary>
```dot
digraph {
    node [shape=none]
    "(block from consensus)" -> HandleNewBlockLocked -> maybeStartProcessingBatchLocked
    "(message with signed proposal)" -> queueBatchBlocking -> handleExternalBatchLocked -> maybeStartProcessingBatchLocked -> startProcessingBatchedLocked
    startProcessingBatchedLocked -> "(proposed site for import)"
    startProcessingBatchedLocked -> resolveBatchLocked -> "ub.resolve"
    g1 [label="goroutine"]
    resolveBatchLocked-> g1
    g1 -> requestMissingTransactions -> "(request through txsync)"
    g2 [label="goroutine"]
    startProcessingBatchedLocked -> g2
    g2 -> getRtStateAndRoundResults -> "(read from consensus)"
    g2 -> runtimeExecuteTxBatch -> "(entry into runtime)"
}
```
</details>

seems like this will need to start a goroutine as we have some locks held, and we'd need to go off and access consensus state. probably going to need a new state, in that case, to help coordinate the completion of "get transactions from roothash incoming messages" and "request the rest over txsync"

pro-wh:
updated, now a routine runs during requestMissingTransactions

uses existing `SubmitProposedBatch`--is this appropriate? seems to meet our need here:
- arranges for these transactions to undergo CheckTx
- marks them as in our possession, under txPool.proposedTxs

go/roothash/api/api.go
```diff
@@ -217,6 +217,9 @@ type SubmitMsg struct {
 	// Tokens are any tokens sent into the runtime as part of the message being sent. The tokens are
 	// transferred before the message is processed by the runtime.
 	Tokens quantity.Quantity `json:"tokens,omitempty"`
+	// Transaction is an optional transaction. `nil` means no transaction, while `tx = nil; &tx` is
+	// a transaction that is the empty byte string. Go, I swear.
+	Transaction *[]byte `json:"transaction,omitempty"`
```
kostko, line 222: This breaks consensus, why is this new field needed? Any message data is runtime-dependent and should go into `Data`. It's not like only some messages (with this transaction field set?) should be propagated into runtimes, all incoming messages should be.

pro-wh: not needed, although it saves us the overhead of adding an extra entry into the runtime to parse out the transactions.

it's also not this field alone that breaks consensus, is it? or would we not enter the other changes as breaking?

pro-wh: hm I guess that would only affect the runtime committee.

let me try having the .Data be a serialized structure that contains an optional transaction. we can have the node understand that, and still avoid an extra entry into the runtime

pro-wh:
updated this to repurpose `.Data` instead of adding a new `.Transaction` field. adapted the "incoming message data" structure from the runtime side https://github.com/oasisprotocol/oasis-sdk/pull/832.

pro-wh:
bleh I need to change this some more, likely to make it have a separate queue in the txpool struct

- right now it's really fiddly in the runtime, where we loop through the incoming messages and process transactions, but then we have to watch out for those transactions in the provided "batch" so that we don't accidentally process them a second time
- if we have a separate queue, we can avoid sending the incoming messages' transactions in the "batch" parameter

some other thoughts:
- might want to send the transactions in a separate parameter to the runtime, so that we don't have to parse again in the runtime
- in this current draft, we parse once outside the runtime and once inside. if those two ever disagree, it'll be somehow bad, in a way I haven't even worked out yet

pro-wh:
various thoughts from today:

requirements for separate incoming message embedded transactions queue

- far less stateful, we will always know the contents by taking messages from consensus roothash. doesn't need step-by-step maintenance of adding and removing finished transactions
- for txsync, counts as known. doesn't need to request it, able to provide it
- doesn't undergo checktx. any garbage, dupes, silly ordering, whatever get processed as is
- the runtime may be written to eliminate duplicates internally in schedule_and_execute_batch. but as always, the proposer may propose anything, not necessarily what the runtime implements
- the pool's data structure doesn't need order or allow duplicates; i.e., a set would work. the runtime will actually read from the incoming messages queue
- if a tx appears in this queue, we remove it from the main queue

decoding discrepancies between node and runtime

- maybe it's safer to let the runtime to do the only decoding
- we can still look at the transactions it made in the input write log. but how well would that be supported?
- if node sees transaction that runtime doesn't see, no worries. it just sits around
- if runtime sees transaction that node doesn't see, runtime has to halt because workers aren't able to obtain the transactions for a block over txsync

pro-wh:
image

mapping out how the codebase interacts with txPool

pro-wh:
design notes from today:

- txPool already has multiple stores of transactions. one is its scheduleQueue (field is named schedulerQueue), another is a proposedTxs map. we should be able to add another store, with transactions from incoming messages
- there's some duplication that I don't understand between the txPool's proposedTxs and the node's state's unresolvedBatch
- we can separate off a "main queue" that interacts with all of the existing features:
  - incoming transactions first undergo checktx
  - prioritization and indexing by scheduleQueue
  - automatic push via TxPublish
  - automatic recheck
  - available to pull via txSync
- we add a new concept around the "main queue" of a "schedulable queue" (to distinguish from other queues nearby in the system, such as the checktx queue)
- we add a new schedulable queue for transactions from incoming messages, with fewer features:
  - no checktx. any junk that won't work is already paid for from the consensus layer
  - no priority sorting, no indexing to delete extra txs from same account. one account can even submit multiple txs with increasing nonce and they should work in order. any other junk that won't work is already paid for from the consensus layer
  - no push via TxPublish. all nodes should have access from consensus
  - no recheck, as there was no check in the first place
  - possibly available to pull via txSync, as a nicety to other software. but for efficiency, oasis-node shan't request
- how do we keep the transactions in the queues disjoint?
  - when we load a tx into the incoming messages queue, something somewhere will remove that tx from the main queue
  - probably have an overarching index of all txs in all queues, in order to handle GetKnownBatch and txSync. can consult this before adding to main queue
- we can rename + redocument several existing methods to indicate that they only work on the main queue
  - seenCache: not sure
  - txPool.proposedTxs: collects from all queues
  - unresolvedBatch: not sure, seems like it ought to track txPool.proposedTxs and thus collect form all queues as well, but in some cases doesn't
  - republish worker: operates only on main queue
  - recheck worker: operates only on main queue
  - txPool.GetKnownBatch: operates on all queues
  - txSync: txs from all queues available, only used to fill in main queue
  - requestMissingTransactions -- startProcessingBatchLocked -- resolveBatchLocked circle of doom plus txPool.PromoteProposedBatch and txPool.SubmitProposedBatch:
    - requestMissingTransactions needs a name change. it does more than request. it maintains a lot of state outside of itself too
    - requestMissingTransactions uses txSync, which retrieves from all queues on other nodes
    - requestMissingTransactions currently sends txs to SubmitProposedBatch, I would guess, so that if this round fails and this node becomes proposer, it can check the txs for itself and propose them. we should be able to have it continue to do this, just filter out any that show up in the incoming messages queue, the same way SubmitTx shall have to
    - resolveBatchLocked dunno
    - startProcessingBatchLocked I think we won't have to change
    - PromoteProposedBatch I think we won't have to change
    - SubmitProposedBatch I think we won't have to change, but there's some duplicate work between it and PromoteProposedBatch
- to be determined: we currently have a txpool (lowercase p meaning the go module rather than the struct) Transaction type that's highly specialized for the main queue. should we come up with another type for less fully featured schedulable queues?

pro-wh:
design notes from today:

- new interfaces:
  - UsableTransactionSource - a place to retrieve transactions that are "good enough." "good enough" variously means checktx'd, came from incoming message, or came from our own node
    - GetSchedulingSuggestion() -> tx[] - return some number of transactions to give to the scheduler, up to the implementation
    - GetTxByHash(hash) -> tx? - return the specific tx, if it is in this queue. used for resolving a batch from hashes and serving txSync
    - HandleTxsUsed(hash[]) -> void - callback to indicate that the scheduler is done with a set of txs, by hash. for most implementations, remove it from internal storage. this method may need refinement to pass information about whether it was successfully executed + scheduled or failed
  - RecheckableTransactionStore - methods used for rechecking
    - TakeAll() -> tx[] - remove all txs and return them
    - OfferChecked(tx[]) -> void - add txs that are checked
  - RepublishableTransactionSource - a place to get transactions that we want to push
    - GetTxsToPublish(now) -> (tx[], next) - get txs that this queue wants to publish between last call and `now` given, as well as next time that it wants to publish any transactions. if nothing to publish in the future, return next=now+republishInterval so that republish worker checks again if new txs come in
  - no common interface for directly editing queue. those are all queue-specific. most changes can probably go through OfferChecked and HandleTxsUsed though
- new queues:
  - main queue: a priority queue, implements UsableTransactionSource and RecheckableTransactionStore
    - has internal index on sender, to implement existing behavior of replacing txs
    - did we used to republish everyone else's txs too? can maybe implement RepublishableTransactionStore too
  - local txs queue: implements UsableTransactionSource, RecheckableTransactionStore, RepublishableTransactionSource
    - no priority reordering, keep our own txs in order
    - scheduling suggestion is always everything
    - we can give txs from this queue to the runtime first
  - incoming messages txs queue: implements UsableTransactionStore
    - HandleTxsUsed does no removal, roothash module manages the set of pending incoming messages for us
    - scheduling suggestion is always empty, as the runtime will actually get transactions from incoming messages list instead of from tx batch parameter
- uses of queues by surrounding systems:
  - get initial batch to give to scheduler: UsableTransactionSource[], call GetSchedulingSuggestion, concatenate. pool will set it up in order of incoming messages txs queue, local txs queue, main queue
  - give scheduler more txs on request: some non-common interface going to the main queue only? maybe. forgot to put this on the map lol
  - checktx worker: RecheckableTransactionStore[] to offer into, checktx requests have an attached RecheckableTransactionStore reference for where to put checked tx
  - recheck worker: RecheckableTransactionStore[] to pull from, puts them in checktx queue with with the queue it got each tx from
  - republish worker: RepublishableTransactionSource[] to pull from. publishes any it gets, waits until earliest next-publish time. assuming newly added txs can't affect a next-publish time as it would want to be republished farthest in the future
- changes, including changes as a result of the above:
  - tx type is only the raw tx. metadata e.g. the current Transaction type will be internal to queues
  - previously a field tracked a tx's checkedness state, now it'll be managed by moving the tx from the main queue to the check tx queue
  - previously when rechecking, txs remain in the main queue while being rechecked. now they will be removed and not reinstated until checked
  - check tx queue will maintain metadata indicating which queue to put the finished tx in
    - in recheck, set it to which queue it got the tx from
    - isLocal flag will instead have the destination be the local txs queue
    - isDiscard flag will instead have the destination be nil
    - seems in one place we did local and discard, but the effect would have been the same as only discard. local currently means we publish after checking, but discard comes first and prevents it
  - seenCache had previously included transactions that are currently in queues. let's change it only to include txs that are known done (successful only)
    - deduping is now handled by queues, which was all in memory anyway
    - tracking when to publish what is now up to the queues
- dunno how to implement:
  - early flush if we have enough txs
    - we don't even have gas info on incoming message txs because we don't check them

go/roothash/api/message/incoming_message.go
```diff
@@ -26,10 +29,29 @@ type IncomingMessage struct {
 	// transferred before the message is processed by the runtime.
 	Tokens quantity.Quantity `json:"tokens,omitempty"`
 
-	// Data is arbitrary runtime-dependent data.
+	// Data is a serialized IncomingMessageData.
 	Data []byte `json:"data,omitempty"`
 }
 
+const LatestIncomingMessageDataVersion = 1
```
kostko, line 36: Why does Core require knowledge of the incoming message format? It should be completely up to the runtime? All incoming messages should be processed/propagated in the same way?

kostko (commented):
Would be cool if you started drafting some of this in code form to make it easier to follow. I think that going with multiple queues is the right approach.

Also it would be great if we didn't need to break the runtime host protocol for this change. What exactly will be changing in a breaking way? Can it be solved with features instead of breaking the protocol?

go/roothash/api/message/incoming_message.go
pro-wh: for executor workers, they must have all incoming message transactions ready to populate the proposed batch, all before entering the runtime for execution. in this design, we extract the transaction within core

pro-wh: hm maybe we don't even have to bump the runtime host protocol version. in the current design we're not even adding a new entry into the runtime to parse out the transactions, because we're exposing it to the node via this struct.

we can add a slot for completely opaque data next to `tx`, if that's desired. or we can add it later when runtimes want to start using such data

pro-wh:
there are indeed a lot of places where we relied on the cached tx hash. if many of these remain, let's use a much simplified {raw: byte[], hash: hash} Transaction struct

go/roothash/api/message/incoming_message.go
kostko: Hm, am I missing something? All incoming messages should be handled in the same way, regardless of what is in them or how the runtime interprets them. E.g. they would all be handled as "transactions", not only some subset of such messages. Otherwise this will make things more complicated.

go/runtime/txpool/local_queue.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+var (
+	_ UsableTransactionSource        = (*localQueue)(nil)
+	_ RecheckableTransactionStore    = (*localQueue)(nil)
+	_ RepublishableTransactionSource = (*localQueue)(nil)
+)
+
+// localQueue is a "front of the line" area for txs from our own node. We also keep these txs in order.
+type localQueue struct {
```
kostko, line 16: Probably needs synchronization?

go/runtime/txpool/queues.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"time"
+
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
+)
+
+// UsableTransactionSource is a place to retrieve txs that are "good enough." "Good enough" variously means CheckTx'd,
+// came from roothash incoming message, or came from our own node.
+type UsableTransactionSource interface {
+	// GetSchedulingSuggestion returns some number of txs to give to the scheduler as part of the initial
```
kostko, line 12: Shouldn't the number of transactions be a parameter (e.g. currently the initial batch size comes from the runtime's configuration)?

go/runtime/txpool/main_queue.go
```diff
@@ -0,0 +1,60 @@
...
+	for _, txMeta := range txMetas {
+		txs = append(txs, txMeta.tx)
+	}
+	return txs
+}
+
+func (mq *mainQueue) OfferChecked(txs [][]byte) {
+	for _, tx := range txs {
+		txMeta := newTransaction(tx, txStatusChecked)
+		if err := mq.inner.add(txMeta); err != nil {
```
kostko, line 53: I don't think this batch `OfferChecked` variant is ok as we need to be able to handle individual errors (at least in the current txpool implementation).

go/runtime/txpool/txpool.go
```diff
@@ -302,33 +312,52 @@ func (t *txPool) PromoteProposedBatch(batch []hash.Hash) {
...
-	pendingScheduleSize.With(t.getMetricLabels()).Set(float64(t.schedulerQueue.size()))
+func (t *txPool) GetSchedulingSuggestion() [][]byte {
+	var txs [][]byte
+	for _, q := range t.usableSources {
+		txs = append(txs, q.GetSchedulingSuggestion()...)
+	}
+	return txs
 }
 
-func (t *txPool) GetPrioritizedBatch(offset *hash.Hash, limit uint32) []*Transaction {
```
kostko, line 326: Where did support for iterating over the prioritized suggested batch via offset/limit go? This is needed as we don't want to send the entire transaction queue into the runtime.

go/runtime/txpool/txpool.go
```diff
@@ -231,38 +234,46 @@ func (t *txPool) SubmitTxNoWait(ctx context.Context, tx []byte, meta *Transactio
...
 
 func (t *txPool) addToCheckQueue(pct *PendingCheckTransaction) error {
+	// todo: hash used to be cached
+	h := hash.NewFromBytes(pct.tx)
 	t.logger.Debug("queuing transaction for check",
 		"tx", pct.tx,
-		"tx_hash", pct.hash,
-		"recheck", pct.isRecheck(),
+		"tx_hash", h,
+		"recheck", "dunno", // todo: can we live without this?
```
kostko, line 272: It is useful if we know at any point whether a given transaction has already been checked by the runtime. Not just for logging, but this can also be used as an optimization (later), for example it can hint the runtime that it does not need to check signatures on re-checks as they have already been verified in the initial check.

Why is this a problem?

go/runtime/txpool/txpool.go
```diff
@@ -231,38 +234,46 @@ func (t *txPool) SubmitTxNoWait(ctx context.Context, tx []byte, meta *Transactio
...
+		pct.dstQueue = t.localQueue
+	} else {
+		pct.dstQueue = t.mainQueue
 	}
 
 	return t.addToCheckQueue(pct)
 }
 
 func (t *txPool) addToCheckQueue(pct *PendingCheckTransaction) error {
+	// todo: hash used to be cached
```
kostko, line 267: Yeah it would be great if this was the case to avoid recomputing the tx hashes all over the place.

go/runtime/txpool/txpool.go
```diff
@@ -231,38 +234,46 @@ func (t *txPool) SubmitTxNoWait(ctx context.Context, tx []byte, meta *Transactio
 }
 
 func (t *txPool) submitTx(ctx context.Context, rawTx []byte, meta *TransactionMeta, notifyCh chan *protocol.CheckTxResult) error {
-	tx := newTransaction(rawTx, txStatusPendingCheck)
+	// todo: in the new design, we'll be submitting to the check queue first, which won't have "first seen time"
+	// metadata. seems like it could still work out, as checking is single threaded, so we can still preserve the
+	// original order when we finish checking and offer to the main queue
```
kostko, line 239: Sure if we are careful about ordering everywhere, including when checking a batch fails and batch is re-queued (e.g. `retryBatch` in the current check queue). Why can't we track first seen time?

kostko (commented):
I would suggest splitting this into two steps/PRs:
* Refactoring the transaction pool.
* Adding the RIM queue.

Also why are we going back to untagged transactions (e.g. ones without metadata) instead of having something like the current `Transaction` type which follows the transaction over its lifetime in the queues? The general one could be simplified (hash and first seen time?) and then queues can wrap with their own metadata if needed?

go/runtime/txpool/local_queue.go
pro-wh: I need to examine what goroutines interact with it

go/runtime/txpool/txpool.go
pro-wh: gonna merge this with the discussion about UsableTransactionSource.GetSchedulingSuggestion

go/runtime/txpool/queues.go
pro-wh: transferring comment from other thread:

> Where did support for iterating over the prioritized suggested batch via offset/limit go? This is needed as we don't want to send the entire transaction queue into the runtime.

should have those parameters probably

go/runtime/txpool/main_queue.go
pro-wh: ya the whole topic of how to notify the submitter of a transaction's fate is not settled yet. but for one thing, mainQueue is for txs from outside, and the submitter won't be some local listener we have. this will be more important in localQueue

go/runtime/txpool/txpool.go
pro-wh: let's preserve it for use in such future optimizations then. good that you bring up the idea, since I didn't see that in examining existing code

go/runtime/txpool/txpool.go
pro-wh: let's see how many uses remain after this draft is more complete. gonna be sad if it's only for logging. but still

go/runtime/txpool/txpool.go
pro-wh: would have to be tracked in the externally-facing tx structure, which isn't settled yet

pro-wh:
> I would suggest splitting this into two steps/PRs:
> 
>     * Refactoring the transaction pool.
> 
>     * Adding the RIM queue.
> 

adding RIM queue is too important to the design of the multiple queue system for us to do the refactor without it. may be able to do both and then remove the RIM queue in one PR though

> 
> Also why are we going back to untagged transactions (e.g. ones without metadata) instead of having something like the current `Transaction` type which follows the transaction over its lifetime in the queues? The general one could be simplified (hash and first seen time?) and then queues can wrap with their own metadata if needed?

not so much that we must go to untagged transactions, but we're experimenting with it because there were no strong uses of it outside of the queue noted while exploring the existing code and considering the goals. we may find some reasons to have tagged transactions during implementation

go/roothash/api/message/incoming_message.go
pro-wh: ya let's make that simplification, that way there's no parsing to be done, and thus no risk of different parsing

pro-wh:
changed: OfferChecked now gives singular tx. additional metadata and error return in the future probably

kostko:
> seems we currently all txs, even ones received from outside. whereas local txs we publish immediately after checktx, nonlocal ones we defer until the next republish wakeup. and then thereafter they all get republished every minute. if a republish attempt fails (although the current p2p implementation never reports any error), it'll be retried on the next wakeup

The interval is actually updated based on the P2P min republish interval (which I believe bumps it somewhere over 2 minutes currently).

Transactions received from outside don't need to be republished immediately (as they will be propagated further via gossipsub). The only reason why republish exists at all is because gossipsub is not reliable. Although since we added txsync maybe this is not really critical anymore.

pro-wh:
stuff from today:

adding some ideas about managing republishing, to space txs out uniformly throughout the republish interval. since we'll have two queues that want to republish, we could extract out this common interval management functionality.

pragmatically the two queues will push to the republisher rather than the republisher pulling, because the general design of these usable tx queues doesn't expose when txs are added or removed.

this would have been nice to have the check queue give to a usable tx queue, then the usable tx queue gives to the republish queue, then the republisher worker gets from the republish queue, eliminating the need for the check queue to republish directly when completing local txs. and the idea was for the local queue to use PushNow to insert things at the front of the republish queue, but this has the effect of reversing the transaction order. also of not really being immedate due to the uniform spacing taking giving a finite amount of time in between txs. might have to have a special code path for "publish right now but also put it at the end of the queue" after all.

and the whole handling of failed publish is missing. (although the p2p system doesn't currently support failed publish)

pro-wh:
ah let's continue to use the seenCache for republishing :shrug: ain't nobody asked us to even out the republish bandwidth

pro-wh:
notes from today:

- adding a modest amount of metadata to txs. they now have hash.
- undoing the republish new ideas. republishable sources will only give a list of all their transactions, and the republish worker will still consult the seen cache like before

pro-wh:
notes from today:

- pending check tx flags field is back, now holding a 'recheck' bit
- republish system is worked in: `OfferChecked` reaches out into seenCache to schedule the new tx as needed: immediately for localQueue, on next interval for mainQueue

go/runtime/txpool/txpool.go
pro-wh: ok we're back to caching this

go/runtime/txpool/txpool.go
pro-wh: pct.flags is back, with a bit for this recheck flag

pro-wh:
nvm managing seenCache from inside queue is lame. we don't know whether it's recheck. bumping the last publish time unconditionally will starve the republish worker because we recheck on a shorter interval. moved that back out the tx pool. aaaa

pro-wh:
squashing

---

today's notes:

image

<details>
<summary>source</summary>

```dot
digraph {
    node [shape=box]
    seenCache
    checkTxQueue
    proposedTxs
    queues
    blockInfo
    node [shape=parallelogram]
    "go requestMissingTransactions"
    "committee worker"
    checkWorker
    republishWorker
    recheckWorker
    flushWorker
    node [shape=house]
    CrossNode
    node [shape=diamond]
    any
    "runtime handler"
    "txsync handler"
    node [shape=none]
    any -> SubmitTx -> submitTx -> seenCache
    submitTx -> addToCheckQueue -> checkTxQueue
    any -> SubmitTxNoWait -> submitTx
    "go requestMissingTransactions" -> SubmitProposedBatch -> SubmitTxNoWait
    SubmitProposedBatch -> proposedTxs
    CrossNode -> PromoteProposedBatch -> GetKnownBatch
    PromoteProposedBatch -> proposedTxs
    CrossNode -> ClearProposedBatch -> proposedTxs
    CrossNode -> RemoveTxBatch -> "not implemented"
    "committee worker" -> RemoveTxBatch
    CrossNode -> GetSchedulingSuggestion -> queues
    "runtime handler" -> GetSchedulingSuggestion
    "txsync handler" -> GetKnownBatch -> queues
    GetKnownBatch -> proposedTxs
    CrossNode -> ProcessBlock -> blockInfo
    checkWorker -> checkTxBatch -> getCurrentBlockInfo -> blockInfo
    checkTxBatch -> checkTxQueue
    checkTxBatch -> queues
    checkTxBatch -> seenCache
    republishWorker -> queues
    republishWorker -> seenCache
    recheckWorker -> queues
    recheckWorker -> addToCheckQueue
    flushWorker -> "(channels only)"
}
```

</details>

analysis of what goroutines access what txpool state

pro-wh:
summary of places we've called RemoveTxBatch:

1. for a client to stop broadcasting a tx, when it sees the tx appear in a block
2. for a compute node to stop scheduling a tx, when it sees the tx appear in a block
3. for a compute node to stop scheduling a tx, when the tx no longer passes checktx

which sounds fine, I will try hooking these up to use `HandleTxsUsed`

pro-wh:
oh right, still have to paginate the GetSchedulingSuggestion

pro-wh:
notes from today:

- reinstated the initial batch "size," but it's now no longer the actual size. it's now called countHint, and it only affects how many we get from the main queue. anything from the local queue will be in addition (we always give all txs from the local queue). furthermore, the finished batch will also contain as many from the roothash incoming messages "queue" as allowed.
- split out a paginated GetSchedulingExtra (not good name :slightly_frowning_face: ) that for now goes directly to the main queue and uses the same old hash offset + limit parameters to handle requests from the runtime. in the future we may want to extend this to let the runtime be more choosy where it gets more txs from

pro-wh:
notes from today:

- added locking for the new queues
- common node, in `handleNewBlockLocked`, now gives incoming messages to the txpool, which puts them in to the roothash incoming messages tx queue

pro-wh:
notes from today:

- unit tests for new queues added
- initialization patched up enough not to cause compile errors, but still need to add code to construct queues properly

pro-wh:
notes from today:

- added constructors for queues, added initialization code in txpool
- adjustments to the rest of the codebase for changes to txpool

pro-wh:
notes from today: we're going with that simplified design where a roothash incoming message's data is entirely a runtime transaction. no extra structure around it

go/runtime/txpool/main_queue.go
pro-wh: added error return

go/runtime/txpool/queues.go
pro-wh: parameters added.

access to later "pages" of transactions added back too. for now it consults only the main queue. future work can add a request option to specify otherwise

go/runtime/txpool/local_queue.go
pro-wh: added a lock

go/roothash/api/message/incoming_message.go
pro-wh: thusly simplified

pro-wh:
gonna try the CI

pro-wh:
I just remembered, we some more logic to deduplicate transactions across queues.

- when adding a tx to check queue,
  - if present in a queue that's higher priority than destination (rim > local > main), discard immediately
    - tell requester we accepted it :shrug: 
- when adding a tx to a queue (e.g. after checking or when loading from roothash),
  - if present in a queue that's higher priority than destination (rim > local > main), discard
    - tell requester we accepted it :shrug: 
    - can happen e.g. if we start checking for local queue, then start checking for main queue before finishing the first check tx batch
  - if present in a queue that's lower priority than destination, remove tx from that lower queue

duplicates may exist within check queue unfortunately

pro-wh:
alright stuff's broken

pro-wh:
e2e failures in
- e2e/runtime/storage-sync-inconsistent
- e2e/runtime/sentry
- e2e/runtime/node-shutdown

pro-wh:
starting off with the e2e/runtime/sentry one, it just hangs here

> ...
level=debug ts=2022-06-21T23:31:36.932707459Z caller=network.go:454 module=oasis/e2e/runtime/sentry msg="starting node" name=client-0
level=info ts=2022-06-21T23:31:36.935245841Z caller=network.go:652 module=oasis/e2e/runtime/sentry msg="launching Oasis node" args="--datadir /tmp/oasis-test-runner1484540769/e2e/runtime/sentry/network/client-0 --log.level debug --log.format json --log.file /tmp/oasis-test-runner1484540769/e2e/runtime/sentry/network/client-0/node.log --genesis.file /tmp/oasis-test-runner1484540769/e2e/runtime/sentry/network/genesis.json --debug.dont_blame_oasis --debug.allow_root --debug.allow_test_keys --debug.rlimit 50000 --runtime.provisioner sandboxed --consensus.tendermint.abci.prune.strategy none --consensus.tendermint.core.listen_address tcp://0.0.0.0:20031 --consensus.tendermint.core.external_address tcp://127.0.0.1:20031 --grpc.log.debug --consensus.tendermint.p2p.seed 62819BBC1A0479D1903D2E6D9865A042DA8BB85E@127.0.0.1:20015 --worker.p2p.port 20032 --runtime.mode client-stateless --runtime.paths /tmp/oasis-test-runner1484540769/e2e/runtime/sentry/network/runtime-8000000000000000000000000000000000000000000000000000000000000000-1/bundle-0.orc --consensus.tendermint.debug.addr_book_lenient --consensus.tendermint.debug.allow_duplicate_ip --consensus.tendermint.upgrade.stop_delay 10s --debug.grpc.internal.socket_path /tmp/oasis-test-runner1484540769/internal-3278963799.sock"
level=info ts=2022-06-21T23:31:36.947469523Z caller=network.go:505 module=oasis/e2e/runtime/sentry msg="network started"
level=info ts=2022-06-21T23:31:36.947602734Z caller=runtime.go:524 module=scenario/e2e/runtime/sentry msg="ensuring client node is synced"
level=info ts=2022-06-21T23:32:17.221175051Z caller=runtime_client_kv.go:110 module=scenario/e2e/runtime/sentry msg="initializing simple key/value runtime!"

yup, gonna have to debug

pro-wh:
it's WaitHostedRuntime in checkWorker hanging

pro-wh:
notes from today:

deep in key manager land. something is calling `init_kdf` with empty req.policy

pro-wh:
notes from today:

- ya that key manager thing was unrelated. I was testing without some env vars that CI uses
- there's a point in time where the kv `insert` call comes in before the `get_runtime_id` call undergoes HandleTxsUsed, resulting in "txpool: replacement tx priority too low." but somehow it retries, thank goodness
- the insert eventually later runs too, judging by a HandleTxsUsed 

even so, the sentry e2e tests still doesn't pass. gotta look farther, I guess

pro-wh:
```
in master--
2022-06-28T11:38:55.724709635Z queuing get_runtime_id check
2022-06-28T11:38:55.724709635Z check start
2022-06-28T11:38:55.727534101Z check complete
2022-06-28T11:40:08.482472989Z epoch transition complete
2022-06-28T11:40:08.482537188Z state transition NotReady -> WaitingForBatch
2022-06-28T11:40:08.483045982Z consensus block received
2022-06-28T11:40:08.483108681Z queuing get_runtime_id recheck
2022-06-28T11:40:08.48318768 Z check start
(probably check completes? no log for all-recheck batches)
2022-06-28T11:40:08.483263879Z asking runtime to schedule batch
2022-06-28T11:40:08.491943375Z scheduled batch with get_runtime_id
2022-06-28T11:40:08.492549467Z p2p receives batch proposal

2022-06-28T11:40:14.119064041Z queuing insert check
2022-06-28T11:40:14.11910244 Z check start
2022-06-28T11:40:14.122107404Z check complete

2022-06-28T11:40:14.216255368Z consensus block received
2022-06-28T11:40:14.21683866 Z removing processed batch from queue
```

why wouldn't this above log about replacement tx not having higher priority? could there be something else removing it from the queue? or maybe I'm mistaking about that logging code being present in master?

```
in branch--
2022-06-29T00:19:27.634899185Z queuing get_runtime_id check
2022-06-29T00:19:27.634953985Z check start
2022-06-29T00:19:27.637924007Z check complete
2022-06-29T00:20:40.408953288Z epoch transition complete
2022-06-29T00:20:40.409010689Z state transition NotReady -> WaitingForBatch
2022-06-29T00:20:40.40912839 Z queuing get_runtime_id check (incorrect, should be recheck)
2022-06-29T00:20:40.40917049 Z check start
2022-06-29T00:20:40.412234512Z check complete (logged in error, we incorrectly didn't mark it as recheck)
2022-06-29T00:20:46.248823317Z scheduled batch with get_runtime_id
2022-06-29T00:20:46.243339977Z HandleTxsUsed (none)

2022-06-29T00:20:48.063626272Z queuing insert check
2022-06-29T00:20:48.063664073Z check start
2022-06-29T00:20:48.068149005Z check complete

2022-06-29T00:20:48.159254266Z consensus block received
2022-06-29T00:20:48.159656068Z removing processed batch from queue
```

pro-wh:
other weird thing: on the client node, the insert tx at some point on client-0 gets rejected for duplicate nonce

> 56:26.725 debug runtime/txpool                                                                                                                                           > check tx failed caller="txpool.go:535" recheck=true result={"error": {"module": "test", "code": 1, "message": "Duplicate nonce: 14785213797695930577"}} tx="o2RhcmdzomNrZXlpaGVsbG9fa2V5ZXZhbHVleFNoZWxsb192YWx1ZV9mcm9tXzgwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDA6MGVub25jZRvNL6GdE97A0WZtZXRob2RmaW5zZXJ0" tx_hash="c374e98ca7ba693102f9cee5349c0bd8562837f9873860ba739780b57d724162"

pro-wh:
ok it's the recheck timer lining up with the schedule batch trigger

```
c0 25:45.993 finished electing committees
c0 25:46.017 worker/common/committee/group epoch transition complete
c0 25:46.017 we are a transaction scheduler
c0 25:46.018 queuing transaction for check recheck
c0 25:46.026 asking runtime to schedule batch
c0 25:46.031 checked new transactions
c0 25:46.035 Transaction batch execution complete
```

we empty out the queues right when the node wants to schedule a batch

pro-wh:
okay there's two bad things going on with the uncontrolled multi-goroutine-ing.
1. at the beginning of the epoch, recheck happens at the same time the scheduler wants to schedule txs for a block. so they end up scheduling an empty block while all the txs are out getting checked. the next round though at least schedules them successfully
2. client sees tx finish, sends next tx that they want to run. but this arrives before our node sees the block and remove the first tx as finished. so it ends up getting rejected ("replacement tx priority too low"). and then they don't retry until like the next epoch

and this just goes back and forth, making everything take an entire epoch ):

pro-wh:
some experimental changes to see if they help (they don't):

- a "drain" mutex for coordinating "I want to do something with everything in the pool" type operations: (i) schedule; (ii) recheck; (iii) republish
- disabling the rule where you have to replace a tx with a higher priority tx

but this breaks things.

> 47:19.452 debug runtime/txpool                          > checkWorker -> WaitHostedRuntime caller="txpool.go:661"

that's now the last we ever hear of the txpool :skull:

pro-wh:
image

okay there's some amount of pubsub/chan/infinitechannel mayhem around the clear-out-used-txs <-> tell-client-about-finished-tx coordination, but luckily they're both done in a single goroutine

pro-wh:
ok fiddling with the order of handlers is not going to fix this. within the client node, I'm not even experiencing any issues. I added some logging, and it ends up like this:

> c0 15:14.051 spinach: client node HandleNewBlockLocked, sending to checkCh
> c0 15:14.051 spinach: client node worker, receiving from checkCh
> c0 15:14.056 spinach: client node checkBlock sending response to pendingTx ch
> c0 15:14.056 spinach: client node checkBlock calling HandleTxsUsed
> c0 15:14.056 spinach: service SubmitTxMeta received response
> c0 15:14.056 spinach: HandleTxsUsed done
> c0 15:14.083 queuing transaction for check c374e98ca7ba693102f9cee5349c0bd8562837f9873860ba739780b57d724162
> c0 15:14.094 Transaction batch check complete

although we kick off the "done" before we clear out the local tx pool, all the channels and checktx make sure the new txs arrive hundredths of a second after we clean out the old txs. seems racy, but we should be fine until after I stop working on this.

really it's the syncrhonization between client and executor node. in the logs for the same run,

> c0 15:14.037 finalized round
> (the logs above are chronologically here)
> c0 15:14.097 spinach: republishing, releasing drain lock
> e0 15:14.099 new message from peer tx c374e98ca7ba693102f9cee5349c0bd8562837f9873860ba739780b57d724162
> e0 15:14.102 Transaction batch check complete
> e0 15:14.104 spinach: ErrReplacementTxPriorityTooLow 23382c3162ba92c465445a28c88bfa20b6fdf75d172855f8b9ce0b4643590f23 <- c374e98ca7ba693102f9cee5349c0bd8562837f9873860ba739780b57d724162
> e0 15:14.157 finalized round

we're like tenths of a second apart on when they see the round finalized.

so on to the other ideas for preventing this:
- maybe allow replacement when the tx to be replaced is in the proposed batch and we're waiting for finalize? this logic will need a lot more cross-component access ):
- maybe have the client retry sooner than next epoch? like maybe a consensus block?

kostko:
> ErrReplacementTxPriorityTooLow

Just wondering how you encountered this error given that currently the runtime is not specifying a sender and so all transactions should be treated as having unique senders.

pro-wh:
oh, I hadn't carried over that piece of logic

pro-wh:
> > maybe have the client retry sooner than next epoch? like maybe a consensus block?
> 
> Where did you see that the client is waiting for an epoch to retry? I assume that by retry you mean republish -- this should not be tied to epochs but it uses wall-clock time. The problem may be that in tests epochs are quite short.

oh you're right, republish, as the client successfully accepts the transaction. this explains why it was taking a whole minute, whereas these local test networks have ~20 second epochs

pro-wh:
force pushing away extra logging and other local changes. will archive in #4881

go/runtime/txpool/main_queue.go
```diff
@@ -0,0 +1,105 @@
...
+		txMeta.priority = meta.Priority
+		txMeta.sender = string(meta.Sender)
+		txMeta.senderSeq = meta.SenderSeq
+	}
+
+	// If the sender is empty (e.g. because the runtime does not support specifying a sender), we
+	// treat each transaction as having a unique sender. This is to allow backwards compatibility.
+	if len(txMeta.sender) == 0 {
+		txMeta.sender = string(txMeta.hash[:])
+	}
```
pro-wh, line 90: reinstated

go/runtime/txpool/txpool.go
```diff
@@ -779,24 +820,52 @@ func (t *txPool) recheckWorker() {
...
-		txs := t.schedulerQueue.getAll()
+		t.recheck()
+	}
+}
 
-		if len(txs) == 0 {
-			continue
-		}
+func (t *txPool) recheck() {
+	t.drainLock.Lock()
```
kostko, line 828: So rechecking now blocks scheduling? Could this be a problem? I guess only if the current node is a scheduler/proposer and the recheck is taking long (timeout is 15 sec).

go/runtime/txpool/transaction.go
```diff
@@ -8,20 +8,12 @@ import (
...
 
-type txStatus uint8
-
-const (
-	txStatusPendingCheck = iota
-	txStatusChecked
-)
-
 // Transaction is a transaction in the transaction pool.
+// todo: move to main_queue.go
```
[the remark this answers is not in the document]
pro-wh: oop ya

go/runtime/txpool/queues.go
```diff
@@ -0,0 +1,39 @@
+package txpool
+
+import (
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
+	"github.com/oasisprotocol/oasis-core/go/runtime/host/protocol"
+)
+
+type TxQueueMeta struct {
```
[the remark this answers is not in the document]
pro-wh: it's rawer than Transaction, but less raw than the byte slice. notably it has a precomputed hash saved alongside. and it's under discussion whether to store more metadata like time when we first received it

go/runtime/txpool/transaction.go
```diff
@@ -87,8 +78,6 @@ func (tx *Transaction) SenderSeq() uint64 {
 
 // setChecked populates transaction data retrieved from checks.
 func (tx *Transaction) setChecked(meta *protocol.CheckTxMetadata) {
```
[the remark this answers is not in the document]
pro-wh: oh I should have offerChecked use this

go/runtime/txpool/queues.go
```diff
@@ -0,0 +1,39 @@
...
+
+// UsableTransactionSource is a place to retrieve txs that are "good enough." "Good enough" variously means CheckTx'd,
+// came from roothash incoming message, or came from our own node.
+type UsableTransactionSource interface {
+	// GetSchedulingSuggestion returns some number of txs to give to the scheduler as part of the initial
+	// batch.
+	GetSchedulingSuggestion(countHint uint32) []*TxQueueMeta
+	// GetTxByHash returns the specific tx, if it is in this queue. The bool is like `value, ok := txMap[key]`. Used
+	// for resolving a batch from hashes and serving txSync.
+	GetTxByHash(h hash.Hash) (*TxQueueMeta, bool)
```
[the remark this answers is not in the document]
pro-wh: this must have been from when it was []byte :facepalm:

go/runtime/txpool/main_queue.go
```diff
@@ -0,0 +1,105 @@
...
+		txs = append(txs, &TxQueueMeta{
+			Raw:  txMeta.tx,
+			Hash: txMeta.hash,
+		})
+	}
+	return txs
+}
+
+func (mq *mainQueue) OfferChecked(tx *TxQueueMeta, meta *protocol.CheckTxMetadata) error {
+	txMeta := newTransaction(tx.Raw)
```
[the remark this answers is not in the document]
pro-wh: gonna have it embed :+1:

go/runtime/txpool/txpool.go
```diff
@@ -308,29 +326,57 @@ func (t *txPool) PromoteProposedBatch(batch []hash.Hash) {
...
 func (t *txPool) ClearProposedBatch() {
 	t.proposedTxsLock.Lock()
 	defer t.proposedTxsLock.Unlock()
 
-	t.proposedTxs = make(map[hash.Hash]*Transaction)
+	t.proposedTxs = make(map[hash.Hash]*TxQueueMeta)
+}
+
+func (t *txPool) GetSchedulingSuggestion(countHint uint32) []*TxQueueMeta {
+	t.drainLock.Lock()
```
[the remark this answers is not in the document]
pro-wh: adding thanks

go/runtime/txpool/transaction.go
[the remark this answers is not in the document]
pro-wh: moved, renamed to MainQueueTransaction

go/runtime/txpool/transaction.go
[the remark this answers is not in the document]
pro-wh: replaced manual field setting with a call to this

go/runtime/txpool/queues.go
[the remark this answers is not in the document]
pro-wh: removed this extra bool return. now there's only the != nil check

go/runtime/txpool/main_queue.go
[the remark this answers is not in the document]
pro-wh: refactored thusly

go/runtime/txpool/txpool.go
```diff
@@ -85,31 +86,33 @@ type TransactionPool interface {
 	// ClearProposedBatch clears the proposal queue.
 	ClearProposedBatch()
 
-	// RemoveTxBatch removes a transaction batch from the transaction pool.
-	RemoveTxBatch(txs []hash.Hash)
+	// HandleTxsUsed uh
```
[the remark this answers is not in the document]
pro-wh: wrote docstrings

go/runtime/txpool/txpool.go
pro-wh: moving timestamp to TxQueueMeta. also taking this opportunity to give it a descriptive field name, `FirstSeen`

go/runtime/txpool/txpool.go
```diff
@@ -509,63 +546,59 @@ func (t *txPool) checkTxBatch(ctx context.Context, rr host.RichRuntime) {
...
 	}
 
-	if len(newTxs) == 0 {
+	if len(goodPcts) == 0 {
 		return
 	}
 
 	t.logger.Debug("checked new transactions",
-		"num_txs", len(newTxs),
+		"num_txs", numNewTxs,
```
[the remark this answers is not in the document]
pro-wh: adding

go/runtime/txpool/txpool.go
```diff
@@ -581,31 +614,34 @@ func (t *txPool) checkTxBatch(ctx context.Context, rr host.RichRuntime) {
...
-		// can error is if the capacity is in bytes and the value size is over capacity.
-		_ = t.seenCache.Put(tx.hash, publishTime)
+	if numNewTxs != 0 {
+		// Kick off publishing for any new txs.
+		t.republishCh.In() <- struct{}{}
 	}
 
 	// Notify subscribers that we have received new transactions.
-	t.checkTxNotifier.Broadcast(newTxs)
+	// todo: this now broadcasts rechecked transactions too
```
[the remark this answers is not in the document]
pro-wh: oh yeah I'll change this back. our only subscriber on this works to fill in the proposed transactions, which won't go wrong, but it would be a waste to have it do extra passes over txs that it already has

go/runtime/txpool/txpool.go
pro-wh: it does.

makes sense that it could be a problem if the queue gets so big. although the time to check the max queue size worth of transactions is less than 15 seconds right?

maybe if we need even longer queues, we could have some way to recheck them in batches and unblock the scheduling as soon as the highest priority ones are done

kostko:
Looks like some tests need updating.

pro-wh:
updated :sunglasses:

kostko:
I think the only major comment is the one about adding back metrics in all the places.

pro-wh:
oasis_txpool_pending_schedule_size now reports the size of the main queue (num txs)
new metrics oasis_txpool_local_queue_size and oasis_txpool_rim_queue_size to track the other two queues

go/runtime/txpool/txpool.go
```diff
@@ -581,31 +614,34 @@ func (t *txPool) checkTxBatch(ctx context.Context, rr host.RichRuntime) {
...
+	t.checkTxNotifier.Broadcast(goodPcts)
 	t.schedulerNotifier.Broadcast(false)
 
-	pendingScheduleSize.With(t.getMetricLabels()).Set(float64(t.PendingScheduleSize()))
+	// todo: metrics
+	// pendingScheduleSize.With(t.getMetricLabels()).Set(float64(t.PendingScheduleSize()))
```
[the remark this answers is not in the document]
pro-wh: reinstated

go/runtime/txpool/txpool.go
```diff
@@ -308,29 +326,57 @@ func (t *txPool) PromoteProposedBatch(batch []hash.Hash) {
...
 }
 
-func (t *txPool) GetPrioritizedBatch(offset *hash.Hash, limit uint32) []*Transaction {
-	return t.schedulerQueue.getPrioritizedBatch(offset, limit)
+func (t *txPool) HandleTxsUsed(hashes []hash.Hash) {
+	for _, q := range t.usableSources {
+		q.HandleTxsUsed(hashes)
+	}
+
+	// todo: metrics
```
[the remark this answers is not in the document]
pro-wh: reinstated

go/runtime/txpool/queues.go
```diff
@@ -0,0 +1,39 @@
+package txpool
+
+import (
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
+	"github.com/oasisprotocol/oasis-core/go/runtime/host/protocol"
+)
+
+type TxQueueMeta struct {
+	Raw  []byte
+	Hash hash.Hash
```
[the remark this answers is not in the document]
kostko: Do you still plan to do this? It would make for a cleaner API preventing accidental modification of these immutable fields.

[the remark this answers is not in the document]
pro-wh: ya sure I'll do that today

still have to be careful though, because most code that uses these are in the same package and has private member access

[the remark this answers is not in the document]
pro-wh: changed

go/runtime/txpool/main_queue.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// String returns a string representation of a transaction.
+func (tx *MainQueueTransaction) String() string {
+	return fmt.Sprintf("MainQueueTransaction{hash: %s, first_seen: %s, priority: %d}", tx.TxQueueMeta.Hash, tx.TxQueueMeta.FirstSeen, tx.priority)
+}
+
+// Raw returns the raw transaction data.
+func (tx *MainQueueTransaction) Raw() []byte {
+	return tx.TxQueueMeta.Raw()
```
kostko, line 43: Given that `TxQueueMeta` is embedded this is forwarded by default. Same with some below.

go/runtime/txpool/queues.go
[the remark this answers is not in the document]
kostko: Cool, just need to update the tests :-)

go/runtime/txpool/queues.go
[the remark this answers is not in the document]
kostko: Still missing documentation comment.

go/runtime/txpool/schedule_queue.go
```diff
@@ -28,20 +28,20 @@ func (tx priorityWrappedTx) Less(other btree.Item) bool {
 	}
 	// If transactions have same priority, sort by first seen time (earlier transactions are later
 	// in the queue as we are iterating over the queue in descending order).
-	return tx.time.After(tx2.time)
+	return tx.TxQueueMeta.FirstSeen().After(tx2.TxQueueMeta.FirstSeen())
```
kostko, line 31: This shouldn't require the explicit dereference as methods/fields on embedded structs are forwarded by default.

go/runtime/txpool/main_queue.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// Size returns the size (in bytes) of the raw transaction data.
+func (tx *MainQueueTransaction) Size() int {
```
kostko, line 47: Could also define this directly on `TxQueueMeta`.

go/runtime/txpool/main_queue.go
pro-wh: thanks, forwarding

go/runtime/txpool/main_queue.go
pro-wh: oh yeah I'll move that up

pro-wh: moved

go/runtime/txpool/schedule_queue.go
pro-wh: forwarded

go/runtime/txpool/main_queue.go
pro-wh: transferred doc comments too

go/runtime/txpool/queues.go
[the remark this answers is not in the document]
pro-wh: image
at long last

kostko (approved):

pro-wh:
thanks for reviewing!

PR #4686, opened 2022-04-22
go/worker/registration: add node status metrics

ptrus:


docs/oasis-node/metrics.md
```diff
@@ -92,6 +92,10 @@ oasis_worker_executor_liveness_live_rounds | Gauge | Number of live rounds in la
 oasis_worker_executor_liveness_total_rounds | Gauge | Number of total rounds in last epoch. | runtime | [worker/common/committee](https://github.com/oasisprotocol/oasis-core/tree/master/go/worker/common/committee/node.go)
 oasis_worker_failed_round_count | Counter | Number of failed roothash rounds. | runtime | [worker/common/committee](https://github.com/oasisprotocol/oasis-core/tree/master/go/worker/common/committee/node.go)
 oasis_worker_node_registered | Gauge | Is oasis node registered (binary). |  | [worker/registration](https://github.com/oasisprotocol/oasis-core/tree/master/go/worker/registration/worker.go)
+oasis_worker_node_registration_eligable | Gauge | Is oasis node eligable for registration (binary). |  | [worker/registration](https://github.com/oasisprotocol/oasis-core/tree/master/go/worker/registration/worker.go)
```
Yawning, line 95: `eligible`

go/worker/registration/worker.go
```diff
@@ -585,6 +618,81 @@ Loop:
...
+		case nodeStatus.ElectionEligibleAfter >= epoch:
+			workerNodeRegistrationEligible.Set(0)
+		default:
+			workerNodeRegistrationEligible.Set(1)
+		}
+
+		// Runtime metrics.
+		for id, faults := range nodeStatus.Faults {
+			rtId := id.String()
+			workerNodeStatusFaults.WithLabelValues(rtId).Set(float64(faults.Failures))
```
kostko, line 683: This will never reset when the counter goes back to zero and the entry in the `Faults` field is removed for that runtime. It should probably set these metrics for any configured runtimes that have no corresponding faults entry.

ptrus: Yeah makes sense. Updated

go/worker/registration/worker.go
```diff
@@ -585,6 +618,88 @@ Loop:
...
+		case nodeStatus.ElectionEligibleAfter == 0:
+			workerNodeRegistrationEligible.Set(0)
+		case nodeStatus.ElectionEligibleAfter >= epoch:
+			workerNodeRegistrationEligible.Set(0)
+		default:
+			workerNodeRegistrationEligible.Set(1)
+		}
+
+		// Runtime metrics.
+		for _, rt := range status.Descriptor.Runtimes {
```
kostko, line 681: Would probably be better to use `w.runtimeRegistry.Runtimes()` to avoid duplicates in case of multiple versions.

ptrus: agree

kostko (approved):

PR #4690, opened 2022-04-24
docker: Add oasisprotocol/aesmd Docker image

kostko:


docker/aesmd/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM ubuntu:20.04
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update -qq && apt-get install -qq curl lsb-release gpg
+
+RUN echo "deb https://download.01.org/intel-sgx/sgx_repo/ubuntu $(lsb_release -cs) main" > /etc/apt/sources.list.d/intel-sgx.list && \
+    curl -sSL "https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key" | apt-key add - && \
```
Yawning, line 8: Should this be included as an asset instead?  This shouldn't change all that much (and changes there probably shoul dhave manual review).

Yawning (changes requested):

docker/aesmd/Dockerfile
kostko: Good idea, done.

Yawning (approved):

PR #4702, opened 2022-04-26
go/oasis-node/cmd: unsafe-reset check datadir validity

ptrus:


go/oasis-node/cmd/node/unsafe_reset.go
```diff
@@ -67,6 +68,21 @@ func doUnsafeReset(cmd *cobra.Command, args []string) {
 		return
 	}
 
+	// Do a state dir sanity check.
+	glob := filepath.Join(dataDir, tendermintCommon.StateDir)
```
ptrus, line 72: Could maybe check for some other files/folders as well

ptrus: added

go/oasis-test-runner/scenario/e2e/runtime/dump_restore.go
```diff
@@ -165,16 +165,17 @@ func (sc *dumpRestoreImpl) Run(childEnv *env.Env) error {
...
 
+	if err = sc.DumpRestoreNetwork(childEnv, fixture, true, sc.mapGenesisDocumentFn); err != nil {
+		return err
+	}
+
 	// Completely reset state for one of the compute nodes so we can test initial sync.
 	sc.Logger.Info("completely resetting state for one of the compute nodes")
 	cli := cli.New(childEnv, sc.Net, sc.Logger)
-	if err = cli.UnsafeReset(sc.Net.ComputeWorkers()[1].DataDir(), false, false); err != nil {
+	if err = cli.UnsafeReset(sc.Net.ComputeWorkers()[1].DataDir(), false, false, true); err != nil {
```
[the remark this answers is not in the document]
ptrus: Because `unsafe-reset` is already run for all nodes in the `DumpRestoreNetwork`, so the tendermint directory will already be cleared on this node that we run unsafe-reset (without perserving any runtime dirs) again.

kostko (approved):

PR #4709, opened 2022-04-26
Emit runtime logs in oasis-node logs

### LINKED ISSUE #1819: Parse runtime logs and emit as oasis-node logs
Since the runtime now outputs structured JSON logs, we should make the worker host parse the log output on runtime's stdout and emit them via the oasis-node logger.

mitjat:
Resolves #1819 

Testing (besides unit tests): Ran an e2e scenario involving a nontrivial runtime, manually inspected logs (`less $(ls -dt --color=none /tmp/oasis-test-runner* | head -1)/e2e/runtime/runtime/network/compute-0/node.log`) to verify that runtime logs look as expected.

### DIFF
--- a/go/common/logging/logging.go
+++ b/go/common/logging/logging.go
@@ -213,6 +213,19 @@ func GetLoggerEx(module string, extraUnwind int) *Logger {
 	return backend.getLogger(module, extraUnwind)
 }
 
+// GetBaseLogger creates a new non-prefixed logger instance with the
+// specified module.
+//
+// The returned logger will not pre-include any log fields (aka prefixes)
+// except for the module name. Its level will be set in accordance
+// with the global config, which can be per-module.
+//
+// This may be called from any point, including before Initialize is
+// called, allowing for the construction of a package level Logger.
+func GetBaseLogger(module string) *Logger {
+	return backend.getBaseLogger(module)
+}
+
 // Initialize initializes the logging backend to write to the provided
 // Writer with the given format and log levels specified for each
 // module. If the requested module is not given, default level is
@@ -239,7 +252,6 @@ func Initialize(w io.Writer, format Format, defaultLvl Level, moduleLvls map[str
 	}
 
 	logger = level.NewFilter(logger, defaultLvl.toOption())
-	logger = log.With(logger, "ts", log.DefaultTimestampUTC)
 
 	backend.baseLogger = logger
 	backend.moduleLevels = moduleLvls
@@ -260,7 +272,7 @@ func Initialize(w io.Writer, format Format, defaultLvl Level, moduleLvls map[str
 
 	// libp2p/IPFS uses yet another logging library, that appears to be a
 	// wrapper around zap.
-	ipfsLogger := newZapCore(logger, "libp2p", 7)
+	ipfsLogger := newZapCore(log.With(logger, "ts", log.DefaultTimestampUTC), "libp2p", 7)
 	backend.setupLogLevelLocked(ipfsLogger.logger)
 
 	// Update the ipfs core logger.
@@ -323,19 +335,38 @@ func (b *logBackend) getLogger(module string, extraUnwind int) *Logger {
 		logger = &log.SwapLogger{}
 	}
 
-	var keyvals []interface{}
-	if module != "" {
-		keyvals = append(keyvals, []interface{}{
-			"module",
-			module,
-		}...)
+	prefixes := []interface{}{
+		"ts", log.DefaultTimestampUTC,
+		"caller", log.Caller(defaultUnwind + extraUnwind),
+		"module", module,
 	}
-	keyvals = append(keyvals, []interface{}{
-		"caller",
-		log.Caller(defaultUnwind + extraUnwind),
-	}...)
 	l := &Logger{
-		logger: log.WithPrefix(logger, keyvals...),
+		logger: log.WithPrefix(logger, prefixes...),
+		module: module,
+	}
+	b.setupLogLevelLocked(l)
+
+	if !b.initialized {
+		// Stash the logger so that it can be instantiated once logging
+		// is actually initialized.
+		sLog := logger.(*log.SwapLogger)
+		b.earlyLoggers = append(b.earlyLoggers, &earlyLogger{swapLogger: sLog, logger: l})
+	}
+
+	return l
+}
+
+func (b *logBackend) getBaseLogger(module string) *Logger {
+	b.Lock()
+	defer b.Unlock()
+
+	logger := b.baseLogger
+	if !b.initialized {
+		logger = &log.SwapLogger{}
+	}
+
+	l := &Logger{
+		logger: log.WithPrefix(logger, "module", module),
 		module: module,
 	}
 	b.setupLogLevelLocked(l)
--- a/go/runtime/host/helpers.go
+++ b/go/runtime/host/helpers.go
@@ -2,13 +2,10 @@ package host
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
-	"strings"
 
 	beacon "github.com/oasisprotocol/oasis-core/go/beacon/api"
 	"github.com/oasisprotocol/oasis-core/go/common/errors"
-	"github.com/oasisprotocol/oasis-core/go/common/logging"
 	consensus "github.com/oasisprotocol/oasis-core/go/consensus/api"
 	"github.com/oasisprotocol/oasis-core/go/roothash/api/block"
 	"github.com/oasisprotocol/oasis-core/go/runtime/host/protocol"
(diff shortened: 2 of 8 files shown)


go/common/logging/logging.go
```diff
@@ -323,19 +335,38 @@ func (b *logBackend) getLogger(module string, extraUnwind int) *Logger {
...
-	var keyvals []interface{}
-	if module != "" {
-		keyvals = append(keyvals, []interface{}{
-			"module",
-			module,
-		}...)
+	prefixes := []interface{}{
+		"ts", log.DefaultTimestampUTC,
+		"caller", log.Caller(defaultUnwind + extraUnwind),
+		"module", module,
```
mitjat, line 341: Slight simplification: We only used to set this for `module!=""`. Grepping shows this is currently always the case. I don't think that usage of `module==""` is encouraged, so if it ever does happen, it should be fine to explicitly draw attention to it with an empty string in the logs.

go/common/logging/logging.go
```diff
@@ -213,6 +213,19 @@ func GetLoggerEx(module string, extraUnwind int) *Logger {
...
 }
 
+// GetBaseLogger creates a new non-prefixed logger instance with the
+// specified module.
+//
+// The returned logger will not pre-include any log fields (aka prefixes)
+// except for the module name. Its level will be set in accordance
+// with the global config, which can be per-module.
+//
+// This may be called from any point, including before Initialize is
```
mitjat, line 223: I implemented this (and copy-pasted the doc paragraph) for consistency with GetLogger(), but I'm happy to simplify the code and panic if logging is not initialized and the method is called. It would work for the current use case, and no other use cases are foreseen.

go/runtime/host/sandbox/sandbox.go
```diff
@@ -525,15 +525,25 @@ func (r *sandboxedRuntime) manager() {
...
+				cfg.Logger,
+				"runtime_id", hostCfg.Bundle.Manifest.ID.Hex(),
+				"runtime_name", hostCfg.Bundle.Manifest.Name)
 			return process.Config{
 				Path: hostCfg.Bundle.Path,
 				Env: map[string]string{
 					"OASIS_WORKER_HOST": socketPath,
 				},
 				SandboxBinaryPath: cfg.SandboxBinaryPath,
+				Stdout:            logWrapper,
```
mitjat, line 545: The PR includes unit tests for the majority of parsing, but it does not cover this integration point (and the one in sgx). Not sure if it's worth adding any tests for this, and where best to add them. I could potentially add checks to [sandbox_test.go](https://github.com/oasisprotocol/oasis-core/blob/1e4123ac5b9293289376aa6191f3a999ca094644/go/runtime/host/sandbox/sandbox_test.go). It would need fixing first, though -- `OASIS_TEST_RUNTIME_HOST_BUNDLE_PATH` is never defined in CI (the older `OASIS_TEST_RUNTIME_HOST_RUNTIME_PATH` is, [here](https://github.com/oasisprotocol/oasis-core/blob/1e4123ac5b9293289376aa6191f3a999ca094644/.buildkite/go/test_and_coverage.sh)), so the test never runs.

Should I add tests anyway? I'm leaning towards "no; manual testing suffices". Is there a better place for these hypothetical tests? It would look a little weird to make the currently very lightweight sandbox_test.go be 80% about logging.

go/runtime/host/helpers.go
```diff
@@ -140,3 +143,118 @@ func (r *richRuntime) ConsensusSync(ctx context.Context, height uint64) error {
 func NewRichRuntime(rt Runtime) RichRuntime {
 	return &richRuntime{Runtime: rt}
 }
+
+// RuntimeLogger is a Writer that interprets data written to it as JSON-formatted
```
kostko, line 147: I would move these into a separate `logger.go`.

go/runtime/host/helpers.go
```diff
@@ -140,3 +143,118 @@ func (r *richRuntime) ConsensusSync(ctx context.Context, height uint64) error {
...
+		if w.buf[i] == '\n' {
+			w.processLogLine(w.buf[:i])
+			w.buf = w.buf[i+1:]
+			i = 0
+		}
+	}
+
+	// Prevent the buffer from growing indefinitely in case runtime logs
+	// don't contain newlines (e.g. because of unexpected log format).
+	if len(w.buf) > 10_000_000 {
```
kostko, line 189: Let's move this into a constant.

go/runtime/host/helpers.go
```diff
@@ -140,3 +143,118 @@ func (r *richRuntime) ConsensusSync(ctx context.Context, height uint64) error {
...
+	case "DEBG":
+		rtLogger.Debug(msg, kv...)
+	case "INFO":
+		rtLogger.Info(msg, kv...)
+	case "WARN":
+		rtLogger.Warn(msg, kv...)
+	case "ERRO":
+		rtLogger.Error(msg, kv...)
+	default:
+		w.logger.Warn("log line from runtime has no known error level set, using INFO", "log_line", string(line))
```
kostko, line 258: It says "using INFO" but it doesn't actually use it?

go/runtime/host/sandbox/sandbox.go
```diff
@@ -525,15 +525,25 @@ func (r *sandboxedRuntime) manager() {
...
+	if cfg.Logger == nil {
+		cfg.Logger = logging.GetLogger("runtime/host/sandbox")
+	}
 	// Use a default GetSandboxConfig if none was provided.
 	if cfg.GetSandboxConfig == nil {
 		cfg.GetSandboxConfig = func(hostCfg host.Config, socketPath, runtimeDir string) (process.Config, error) {
+			logWrapper := host.NewRuntimeLogWrapper(
+				cfg.Logger,
+				"runtime_id", hostCfg.Bundle.Manifest.ID.Hex(),
+				"runtime_name", hostCfg.Bundle.Manifest.Name)
```
kostko, line 538:
```suggestion
				"runtime_name", hostCfg.Bundle.Manifest.Name,
			)
```

go/runtime/host/sgx/sgx.go
```diff
@@ -185,6 +185,11 @@ func (s *sgxProvisioner) getSandboxConfig(rtCfg host.Config, socketPath, runtime
 	}
 	s.logger.Info("found SGX device", "path", sgxDev)
 
+	logWrapper := host.NewRuntimeLogWrapper(
+		s.logger,
+		"runtime_id", rtCfg.Bundle.Manifest.ID.Hex(),
+		"runtime_name", rtCfg.Bundle.Manifest.Name)
```
kostko, line 191:
```suggestion
		"runtime_name", rtCfg.Bundle.Manifest.Name,
	)
```

go/runtime/host/sgx/sgx.go
```diff
@@ -185,6 +185,11 @@ func (s *sgxProvisioner) getSandboxConfig(rtCfg host.Config, socketPath, runtime
 	}
 	s.logger.Info("found SGX device", "path", sgxDev)
 
+	logWrapper := host.NewRuntimeLogWrapper(
+		s.logger,
+		"runtime_id", rtCfg.Bundle.Manifest.ID.Hex(),
```
kostko, line 190:
```suggestion
		"runtime_id", rtCfg.Bundle.Manifest.ID,
```

go/runtime/host/sandbox/sandbox.go
```diff
@@ -525,15 +525,25 @@ func (r *sandboxedRuntime) manager() {
...
+	// Use a default Logger if none was provided.
+	if cfg.Logger == nil {
+		cfg.Logger = logging.GetLogger("runtime/host/sandbox")
+	}
 	// Use a default GetSandboxConfig if none was provided.
 	if cfg.GetSandboxConfig == nil {
 		cfg.GetSandboxConfig = func(hostCfg host.Config, socketPath, runtimeDir string) (process.Config, error) {
+			logWrapper := host.NewRuntimeLogWrapper(
+				cfg.Logger,
+				"runtime_id", hostCfg.Bundle.Manifest.ID.Hex(),
```
kostko, line 537:
```suggestion
				"runtime_id", hostCfg.Bundle.Manifest.ID,
```

go/runtime/host/sandbox/sandbox.go
kostko: Nice catch, can you open an issue for this? The test running script needs to be updated to generate a bundle and set the right env variable.

go/runtime/host/helpers.go
mitjat: :+1: done

go/runtime/host/helpers.go
mitjat: Thanks, fixed.

kostko:
Seems like some lints and unit tests are failing.

PR #4710, opened 2022-04-27
go/runtime/config: error if SGX loader required but not configured

ptrus:


go/runtime/registry/config.go
```diff
@@ -247,6 +249,11 @@ func newConfig(dataDir string, consensus consensus.Backend, ias ias.Endpoint) (*
 
 			switch sgxLoader := viper.GetString(CfgRuntimeSGXLoader); sgxLoader {
 			case "":
+				// Fail if SGX is required and provisioner is not configured.
+				if !forceNoSGX {
+					return nil, fmt.Errorf("required SGX provisioner not configured")
```
kostko, line 254: Maybe rather say that the SGX loader binary path is not configured to make it more clear for the node operators what they need to configure.

kostko (approved):

PR #4713, opened 2022-04-28
go/worker/common/p2p: Add support for persistent and blocked peers

abukosek:
This PR adds support for blocking p2p peers and for keeping persistent connections to peers.

The persistent peer functionality currently doesn't automatically connect to the given peers on startup, but will only keep the connection once it's already established.

### DIFF
--- a/go/worker/common/p2p/p2p.go
+++ b/go/worker/common/p2p/p2p.go
@@ -5,11 +5,13 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/libp2p/go-libp2p"
 	core "github.com/libp2p/go-libp2p-core"
+	"github.com/libp2p/go-libp2p-core/peer"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	pb "github.com/libp2p/go-libp2p-pubsub/pb"
 	"github.com/libp2p/go-libp2p/p2p/net/conngater"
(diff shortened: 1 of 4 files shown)


kostko (commented):
You could probably use [`pubsub.WithDirectPeers`](https://pkg.go.dev/github.com/libp2p/go-libp2p-pubsub#WithDirectPeers) to make the pubsub router maintain those connections. That option also supports specifying IP addresses, see [here](https://github.com/libp2p/go-libp2p-pubsub/blob/v0.6.1/gossipsub.go#L353) for how this is implemented.

go/worker/common/p2p/p2p.go
```diff
@@ -362,6 +365,73 @@ func New(identity *identity.Identity, consensus consensus.Backend) (*P2P, error)
...
+		var pk signature.PublicKey
+		if grr := pk.UnmarshalText([]byte(pubkey)); grr != nil {
+			return nil, fmt.Errorf("worker/common/p2p: malformed persistent peer address: cannot unmarshal P2P public key (%s): %w", pubkey, grr)
+		}
+		pid, grr := PublicKeyToPeerID(pk)
+		if grr != nil {
+			return nil, fmt.Errorf("worker/common/p2p: invalid persistent peer public key (%s): %w", pubkey, grr)
+		}
+
+		ipport := strings.Split(addr, ":")
```
kostko, line 406: Should probably use [`net.SplitHostPort`](https://pkg.go.dev/net#SplitHostPort).

kostko (approved):

go/worker/common/p2p/p2p.go
abukosek: Nice, thanks!

PR #4720, opened 2022-05-02
go/common/sgx/pcs: Initial types and quote parser

kostko:


go/common/sgx/pcs/certificates.go
```diff
@@ -0,0 +1,59 @@
...
+uzCBuDAfBgNVHSMEGDAWgBQiZQzWWp00ifODtJVSv1AbOScGrDBSBgNVHR8ESzBJ
+MEegRaBDhkFodHRwczovL2NlcnRpZmljYXRlcy50cnVzdGVkc2VydmljZXMuaW50
+ZWwuY29tL0ludGVsU0dYUm9vdENBLmRlcjAdBgNVHQ4EFgQUImUM1lqdNInzg7SV
+Ur9QGzknBqwwDgYDVR0PAQH/BAQDAgEGMBIGA1UdEwEB/wQIMAYBAf8CAQEwCgYI
+KoZIzj0EAwIDSQAwRgIhAOW/5QkR+S9CiSDcNoowLuPRLsWGf/Yi7GSX94BgwTwg
+AiEA4J0lrHoMs+Xo5o/sX6O9QWxHRAvZUGOdRQ7cvqRXaqI=
+-----END CERTIFICATE-----`
+
+// IntelTrustRoots are Intel's PCS signing root certificates.
+var IntelTrustRoots = x509.NewCertPool()
```
Yawning, line 27: Instead of having an init function, I sort of favor something like:
```
var IntelTrustRoots = func() *x509.CertPool {
  pool := x509.NewCertPool()
  if raw := []byte(pcsTrustRootCert); !pool.AppendCertsFromPEM(raw) {
    // This should fail, and is more informative that `false`
    _, _, err := CertFromPEM(raw)
    panic("sgx/pcs: invalid Intel trust root cert: " + err.Error())
  }
  return pool
}()
```

go/common/sgx/pcs/certificates.go
```diff
@@ -0,0 +1,52 @@
+package pcs
+
+import (
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+)
+
+const pcsTrustRootCert = `-----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
nhynes: I had a [build script](https://github.com/oasislabs/dcap-quote-verification/blob/master/build.rs#L32) fetch and inline this automatically. The API version went up by two since then, though.

go/common/sgx/pcs/quote.go
```diff
@@ -0,0 +1,584 @@
...
+	copy(qh.QEVendorID[:], data[12:])
+	copy(qh.UserData[:], data[28:])
+
+	qh.raw = data[:quoteHeaderLen]
+
+	return nil
+}
+
+// QEVendorID_Intel is the Quoting Enclave vendor ID for Intel.
+var QEVendorID_Intel = []byte{0x93, 0x9a, 0x72, 0x33, 0xf7, 0x9c, 0x4c, 0xa9, 0x94, 0x0a, 0x0d, 0xb3, 0x95, 0x7f, 0x06, 0x07}
```
nhynes, line 126:
```suggestion
var QEVendorID_Intel = []byte{0x93, 0x9a, 0x72, 0x33, 0xf7, 0x9c, 0x4c, 0xa9, 0x94, 0x0a, 0x0d, 0xb3, 0x95, 0x7f, 0x06, 0x07} // 939A7233F79C4CA9940A0DB3957F0607
```
which can be searched up to find the quote structure reference doc. I assume that people will be following that while they read this, so there's no need to link the spec, so this is just to help pattern matching.

go/common/sgx/pcs/quote.go
```diff
@@ -0,0 +1,584 @@
...
+	if len(data) < quoteSigEcdsaP256MinLen {
+		return fmt.Errorf("pcs/quote: invalid ECDSA-P256 quote signature length")
+	}
+
+	copy(qs.Signature[:], data[0:])
+	copy(qs.AttestationPublicKey[:], data[64:])
+	if err := qs.QEReport.UnmarshalBinary(data[128 : 128+reportBodyLen]); err != nil {
+		return err
+	}
+	copy(qs.QESignature[:], data[512:])
```
nhynes, line 200: no `offset += len(qs.QESignature)` or the like? this section is a little bit harder than the rest to follow.

nhynes (approved):
Looks like a DCAP quote parser and verifier to me! I'm surprised nobody else has written this yet.

go/common/sgx/pcs/certificates.go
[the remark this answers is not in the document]
Yawning: Certs changing (especially extremely long lived ones like Intel promises) should be viewed with suspicion, not automated away.

PR #4721, opened 2022-05-02
[BACKPORT/22.1.x] Multiple backports

kostko:
Backports of:
* #4693 (only dynlib fix)
* #4700
* #4707
* #4702
* #4710 
* #4715
* #4724  
* #4725
* #4726


.changelog/4707.internal.md
```diff
@@ -0,0 +1,4 @@
+ci: Build OpenSSL statically for Oasis Core releases
+
+This makes sure that oasis-core-loader is more portable between
```
Yawning, line 3: This seems questionable as hell, though OpenSSL is moderately better than it used to be.  Though still not something I would personally staticly link ever, because getting updates from the OS vendor is a good idea.

kostko: This is specifically for the oasis-core-runtime-loader (the SGXS loader) so I don't see any problems here. And it does make it easier to use the binary.

ptrus (approved):

.changelog/4707.internal.md
Yawning: I guess, I'm not thrilled by this though.

Yawning (approved):

Issue #3155, opened 2020-07-31
go/common/crypto/signature/signers: Sort out the RPC situation


Yawning:
We now have 2 flavors of RPC backed signers:
 * `signers/plugin` - A simplified `net/rpc` interface for use with `go-plugin`
 * `signers/remote` - A more fully fleshed out gRPC interface for the remote-signer API.

We could consider unifying these as `go-plugin` supports plugins that use gRPC as the transport (though I do not know how nicely it plays with our custom codec situation).

The rationale for using `net/rpc` for the plugin signer was purely one of expediency, and I have no strong attachment to the interface, beyond "it works, so why mess with it".  As the plugin package supports versioning I expect the transition to be fairly painless.

One side benefit of transitioning the plugin signer to gRPC would be the ability to write plugins in non-Go languages, but that also does not seem all that urgent.

nhynes:
I'd like to voice support for gRPC so that I can continue to not write Go.

Yawning:
Patches accepted (This isn't going to happen anywhere in the near future).

nhynes:
nvm. go-grpc doesn't seem to want to allow custom dialers. I could more easily modify the remote signer, but those changes (custom dialer + making TLS certs optional) will never get upstreamed. I guess I'll make a net/rpc->gRPC plugin...

PR #4733, opened 2022-05-05
common/crypto/signature/signers/remote: Support AF_LOCAL

Yawning:


Yawning:
My intention with this is to deprecate the plugin signer (#3155).  People that want to use a ledger device can use the SDK cli.

See also: https://github.com/oasisprotocol/oasis-sdk/pull/889

ptrus (approved):

PR #4741, opened 2022-05-06
runtime: Verify RAK in consensus state when serving requests

ptrus:


runtime/src/protocol.rs
```diff
@@ -140,6 +140,11 @@ impl Protocol {
         &self.config
     }
 
+    /// The runtime attestation key.
+    pub fn get_rak(&self) -> &RAK {
```
Yawning, line 144: Does this ever need the private component, or can this just return the public key?

runtime/src/consensus/registry.rs
```diff
@@ -18,6 +18,211 @@ use crate::{
...
+    pub consensus: ConsensusInfo,
+
+    /// Information for this node's participation in VRF based elections.
+    #[cbor(optional, default)]
+    pub vrf: Option<VRFInfo>,
+
+    /// Information for this node's participation in the old PVSS based random beacon protocol.
+    /// Deprecated.
+    #[cbor(optional, default)]
+    pub beacon: Option<Vec<u8>>,
```
Yawning, line 213: Sigh, there's a PR that gets rid of this on the Go side, but apparently "no breaking changes" or something, so the branch is rotting.

https://github.com/oasisprotocol/oasis-core/pull/4668

runtime/src/protocol.rs
ptrus: Can return just the public key. Good idea, will update.

runtime/src/consensus/tendermint/verifier.rs
```diff
@@ -307,6 +307,37 @@ impl Verifier {
             )));
         }
 
+        // Verify our own RAK is published in registry.
```
kostko, line 310: Only do this once per epoch. Also remember the found node ID and only lookup that entry in subsequent checks.

ptrus: good ideas, added.

runtime/src/consensus/registry.rs
```diff
@@ -18,6 +18,227 @@ use crate::{
...
```
kostko, line 213:
```suggestion
    pub beacon: Option<cbor::Value>,
```

runtime/src/consensus/registry.rs
```diff
@@ -18,6 +18,227 @@ use crate::{
...
+    pub roles: RolesMask,
+
+    /// Node's oasis-node software version.
+    #[cbor(optional, default)]
+    pub software_version: Option<String>,
+}
+
+impl Node {
+    /// Returns if the node has the provided RAK configured.
+    pub fn has_rak(&self, rak: &PublicKey) -> bool {
```
kostko, line 228: Should this also take a runtime ID argument? Just to make sure that the RAK is advertised for the expected runtime ID. There could also be a method like `get_runtime(id, version)` and then the `has_rak` would only exist on the `NodeRuntime` type. Or some combination of these.

runtime/src/consensus/tendermint/verifier.rs
```diff
@@ -307,6 +309,71 @@ impl Verifier {
...
+                        }
+                        if found_node.is_none() {
+                            return Err(Error::VerificationFailed(anyhow!(
+                                "own RAK: '{}' not found in registry state",
+                                rak
+                            )));
+                        }
+
+                        // Cache node ID to avoid re-scanning.
+                        cache.node_id = found_node
```
kostko, line 371:
```suggestion
                        cache.node_id = found_node;
```

kostko (commented):
Mostly looks good, I would just make sure that the RAK is advertised for the correct runtime ID/version combination.

runtime/src/consensus/registry.rs
ptrus: Good idea, added.

ptrus:
Going to do one more CI run with all SGX tests enabled on this branch, before merging.

runtime/src/consensus/registry.rs
```diff
@@ -18,6 +18,230 @@ use crate::{
...
+    pub roles: RolesMask,
+
+    /// Node's oasis-node software version.
+    #[cbor(optional, default)]
+    pub software_version: Option<String>,
+}
+
+impl Node {
+    /// Returns if the node has the provided RAK configured.
+    pub fn has_rak(&self, rak: &PublicKey, runtime_id: &Namespace) -> bool {
```
kostko, line 228: This should also take version as argument (as there can be multiple versions of the same runtime ID announced in the descriptor). And then caller would pass the version from `protocol.config.version` (should probably be set in the verifier constructor).

runtime/src/consensus/tendermint/verifier.rs
```diff
@@ -307,6 +309,71 @@ impl Verifier {
...
+                        let nodes = registry_state.nodes(Context::background()).map_err(|err| {
+                            Error::VerificationFailed(anyhow!(
+                                "failed to retrieve nodes from the registry: {}",
+                                err
+                            ))
+                        })?;
+                        let mut found_node: Option<PublicKey> = None;
+                        'outer: for node in nodes {
+                            if let Some(rts) = &node.runtimes {
+                                for rt in rts {
```
kostko, line 353: This is not checking the namespace (it should probably just use `node.has_rak` to ensure consistency).

ptrus: Oh right, I added `has_rak` to avoid duplication here, but then forgot to update this part of code :grin:

runtime/src/consensus/tendermint/verifier.rs
```diff
@@ -121,6 +122,7 @@ enum Command {
 /// Tendermint consensus layer verifier.
 pub struct Verifier {
     protocol: Arc<Protocol>,
+    protocol_version: Version,
```
kostko, line 125: I would rather call this `runtime_version` (as it is the version of the runtime binary).

kostko (approved):

PR #4744, opened 2022-05-09
go/common/node/address: use custom type instead of internal net.TCPAddr

ptrus:
Since `Address` is used in consensus state, it makes sense using an explicitly defined struct. The struct serialization is defined in a backwards compatible way.

### DIFF
--- a/.changelog/4744.internal.md
+++ b/.changelog/4744.internal.md
@@ -1,0 +1,1 @@
+go/common/node/address: use custom type instead of `net.TCPAddr`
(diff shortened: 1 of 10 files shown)


Yawning:
Crap, becasue we did this late, guess we can't even `omitempty` the `Zone` field huh.

ptrus:
Yeah not at the moment, but we can open a follow issue for the next breaking release to change these.

Yawning:
If anything, for the next breaking release we should start using `net/netip/AddrPort` (1.18-ism)

kostko (approved):

PR #4757, opened 2022-05-23
runtime: Add support for reporting EnclaveRPC peer feedback

kostko:
This makes EnclaveRPC more robust as the higher-level layer in the
runtime can trigger peer replacement on high level errors.

TODO
* [x] Also use the `KeyManagerClientWrapper` in the key manager worker.

### DIFF
--- a/go/worker/keymanager/handler.go
+++ b/go/worker/keymanager/handler.go
@@ -10,7 +10,7 @@ import (
 	runtimeKeymanager "github.com/oasisprotocol/oasis-core/go/runtime/keymanager/api"
 	"github.com/oasisprotocol/oasis-core/go/runtime/localstorage"
 	workerCommon "github.com/oasisprotocol/oasis-core/go/worker/common"
-	"github.com/oasisprotocol/oasis-core/go/worker/keymanager/p2p"
+	committeeCommon "github.com/oasisprotocol/oasis-core/go/worker/common/committee"
 )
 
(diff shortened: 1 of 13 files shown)


go/worker/keymanager/handler.go
```diff
@@ -48,16 +48,12 @@ func (h *hostHandler) Handle(ctx context.Context, body *protocol.Body) (*protoco
...
-			})
+			rsp, err := h.remoteClient.CallEnclave(ctx, body.HostRPCCallRequest.Request, body.HostRPCCallRequest.PeerFeedback)
 			if err != nil {
 				return nil, err
 			}
-			// TODO: Support reporting peer feedback from the enclave.
-			pf.RecordSuccess()
 			return &protocol.Body{HostRPCCallResponse: &protocol.HostRPCCallResponse{
-				Response: cbor.FixSliceForSerde(rsp.Data),
+				Response: cbor.FixSliceForSerde(rsp),
```
ptrus, line 56: Btw is this `FixSliceForSerde` still needed? `serde` is not used anymore for cbor serialization, right?

kostko: True, we could fix it in oasis-cbor, but currently it also doesn't allow deserialization into vectors from NULL, so it is still needed. Will open an issue for it though.

ptrus (approved):

Issue #4196, opened 2021-08-10
Adjust log severity for consensus protocol events


kostko:
Currently there are various log entries emitted from the consensus protocol under "warn" and "error" severities. The issue is that these entries may not refer to the local node so having such high severity may lead the node operator into a false belief that their node is at fault while the error actually occurred while processing an external transaction.

We should go over all the emitted log events and adjust the severity to be more appropriate.

mitjat:
Discussed offline: First step is to demote all of these logs to `debug`, then maybe re-up their level if they originate from the current node.

PR #4759, opened 2022-05-24
consensus: Reduce log spam

mitjat:
Closes https://github.com/oasisprotocol/oasis-core/issues/4196

For a manually selected subset of `Logger().Warn()` and `Logger().Error()` log statements, this PR downgrades them to `Debug()`.

I aimed to downgrade all those error logs that originate in a malformed transaction/message (= external to the node and network), and leave all those error logs that signal a problem with the node or the network itself.

In functions that deal with genesis/initstate, I erred on the side of not downgrading. This stuff doesn't get called often, and when it does, the stakes are higher.

### DIFF
--- a/go/consensus/tendermint/apps/keymanager/keymanager.go
+++ b/go/consensus/tendermint/apps/keymanager/keymanager.go
@@ -128,7 +128,7 @@ func (app *keymanagerApplication) onEpochChange(ctx *tmapi.Context, epoch beacon
 			}
 
 			if err = stakeAcc.CheckStakeClaims(*acctAddr); err != nil {
-				ctx.Logger().WarnQ("insufficient stake for key manager runtime operation",
+				ctx.Logger().Debug("insufficient stake for key manager runtime operation",
 					"err", err,
 					"entity", rt.EntityID,
 					"account", *acctAddr,
--- a/go/consensus/tendermint/apps/keymanager/keymanager.go
+++ b/go/consensus/tendermint/apps/keymanager/keymanager.go
@@ -128,7 +128,7 @@ func (app *keymanagerApplication) onEpochChange(ctx *tmapi.Context, epoch beacon
 			}
 
 			if err = stakeAcc.CheckStakeClaims(*acctAddr); err != nil {
-				ctx.Logger().Warn("insufficient stake for key manager runtime operation",
+				ctx.Logger().WarnQ("insufficient stake for key manager runtime operation",
 					"err", err,
 					"entity", rt.EntityID,
 					"account", *acctAddr,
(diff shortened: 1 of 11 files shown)


go/consensus/tendermint/apps/keymanager/keymanager.go
```diff
@@ -120,15 +120,15 @@ func (app *keymanagerApplication) onEpochChange(ctx *tmapi.Context, epoch beacon
 			acctAddr := rt.StakingAddress()
 			if acctAddr == nil {
 				// This should never happen.
-				ctx.Logger().Error("unknown runtime governance model",
+				ctx.Logger().Debug("unknown runtime governance model",
```
kostko, line 123: This results in a consensus halt, so it should stay as error.

go/consensus/tendermint/apps/roothash/roothash.go
```diff
@@ -157,15 +157,15 @@ func (app *rootHashApplication) onCommitteeChanged(ctx *tmapi.Context, state *ro
 			acctAddr := rt.StakingAddress()
 			if acctAddr == nil {
 				// This should never happen.
-				ctx.Logger().Error("unknown runtime governance model",
+				ctx.Logger().Debug("unknown runtime governance model",
```
kostko, line 160: This results in a consensus halt, so it should stay as error.

go/consensus/tendermint/apps/keymanager/keymanager.go
mitjat: Thank you. Reverted to `Error`; same for the identical log message in `roothash.go` (as noted by you below) and in `go/consensus/tendermint/apps/keymanager/keymanager.go` (which you didn't call out; LMK if that was intentional).

Yawning (commented):
I have mixed feelings about this.  While this is spammy, I would rather have the information when we need to debug failures rather than the information being lost due to insufficiently verbose logging.

mitjat:
@Yawning The motivation for this was that node operators are apparently (per @kostko; I haven't done support there) often confused by these error logs, thinking that it means something is wrong with their node. I imagine they can be especially misleading when something _is_ actually wrong with their node, but the log entry about the real reason is potentially buried between lots of other unrelated and unactionable `Error`-level entries.

The unactionable part alone is good reason these should be `Warn` at most.

I'm open to upgrading these to `Info` (or even `Warn`, though my feeling was that @kostko is pretty strongly against that). But the higher the level that we set, the more useful it would be to somehow tag these log lines to denote that the problem is not the node's fault. Maybe a field like `{'reason': 'Bad request or message that originated outside this node'}` ... although that's potentially not entirely true for registration messages etc.

kostko:
Yeah the problem is that this is mostly reporting things that have nothing to do with the local node, but is a side effect of processing other unrelated transactions. When a node itself submits a transaction and that fails, it is logged at that point (e.g. where submission happens) with error/warn verbosity anyway.

Also network-wide logs can be observed on other nodes that run with debug verbosity since all block processing is deterministic.

Yawning (approved):

PR #4772, opened 2022-06-01
add same unittest for derive_symmetric_key of client-sdk/web-ts for g…

xmzheng:
add same test logic for derive_symmetric_key for golang, rust, which should be identical to those in oasis-sdk/client-sdk/web-ts


runtime/Cargo.toml
```diff
@@ -56,6 +56,7 @@ oid-registry = "0.4.0"
 rsa = "0.5.0"
 base64-serde = "0.6.1"
 lru = "0.7.5"
+hex = "0.4"
```
kostko, line 59: This should be under `dev-dependencies` as it is only used in tests.

kostko (commented):
[Fix commit message](https://github.com/oasisprotocol/oasis-core/blob/master/CONTRIBUTING.md#git-commit-messages), squash into a single commit as the second commit is just a fixup, add [a changelog fragment](https://github.com/oasisprotocol/oasis-core/tree/master/.changelog#readme).

nhynes (approved):
if you could make a note of where you acquired the test vectors, this PR looks good to me

runtime/Cargo.toml
```diff
@@ -67,6 +67,7 @@ features = ["full"]
 
 [dev-dependencies]
 # For storage interoperability tests only.
+hex = "0.4"
```
kostko, line 70: This should be moved below as the above comment applies to the `jsonrpc` dependency.

kostko (commented):
Tests seem fine, please address the remaining commit/lint-related comments.

kostko:
@xmzheng ping on addressing the remaining issues and merging this.

PR #4775, opened 2022-06-02
go/archive-mode: disable runtime P2P if archive mode is used

ptrus:


go/oasis-node/cmd/node/node.go
```diff
@@ -273,16 +274,16 @@ func (n *Node) initRuntimeWorkers() error {
 	// Since the P2P layer does not have a separate Start method and starts
 	// listening immediately when created, make sure that we don't start it if
 	// it is not needed.
-	if n.RuntimeRegistry.Mode() != runtimeRegistry.RuntimeModeNone {
-		if genesisDoc.Registry.Parameters.DebugAllowUnroutableAddresses {
-			p2p.DebugForceAllowUnroutableAddresses()
-		}
```
kostko, line 279: Why was this removed?

go/worker/common/p2p/nop.go
```diff
@@ -0,0 +1,91 @@
...
+func (p *nopP2P) RegisterProtocolServer(srv rpc.Server) {
+}
+
+// Implements api.Service.
+func (p *nopP2P) GetMinRepublishInterval() time.Duration {
+	return time.Hour
+}
+
+// NewMock creates a new no-op mock P2P node.
+func NewMock() api.Service {
```
kostko, line 89: I would rather call it `NewNop` (or similar) as mock could imply it is only used in tests as it provides mock responses.

go/worker/common/p2p/nop.go
```diff
@@ -0,0 +1,91 @@
...
+func (*nopP2P) SetNodeImportance(kind api.ImportanceKind, runtimeID common.Namespace, p2pIDs map[signature.PublicKey]bool) {
+}
+
+// Implements api.Service.
+func (p *nopP2P) Cleanup() {
+}
+
+// Implements api.Service.
+func (p *nopP2P) Name() string {
+	return "mock p2p"
```
kostko, line 29:
```suggestion
	return "no-op p2p"
```

go/worker/common/p2p/nop.go
```diff
@@ -0,0 +1,91 @@
...
+	core "github.com/libp2p/go-libp2p-core"
+
+	"github.com/oasisprotocol/oasis-core/go/common"
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
+	"github.com/oasisprotocol/oasis-core/go/common/node"
+	"github.com/oasisprotocol/oasis-core/go/worker/common/p2p/api"
+	"github.com/oasisprotocol/oasis-core/go/worker/common/p2p/rpc"
+)
+
+// nopP2P is a no-op mock peer-to-peer node.
```
kostko, line 16:
```suggestion
// nopP2P is a no-op peer-to-peer node that does not propagate anything.
```

go/worker/common/p2p/rpc/client.go
```diff
@@ -215,6 +215,33 @@ type Client interface {
...
+
+// Implements Client.
+func (c *nopClient) Call(
+	ctx context.Context,
+	method string,
+	body, rsp interface{},
+	maxPeerResponseTime time.Duration,
+	opts ...CallOption,
+) (PeerFeedback, error) {
+	return nil, fmt.Errorf("unsupported: mock p2p")
```
kostko, line 230:
```suggestion
	return nil, fmt.Errorf("unsupported: p2p is disabled")
```

go/worker/common/p2p/rpc/client.go
```diff
@@ -215,6 +215,33 @@ type Client interface {
...
+// Implements Client.
+func (c *nopClient) CallMulti(
+	ctx context.Context,
+	method string,
+	body, rspTyp interface{},
+	maxPeerResponseTime time.Duration,
+	maxParallelRequests uint,
+	opts ...CallMultiOption,
+) ([]interface{}, []PeerFeedback, error) {
+	return nil, nil, fmt.Errorf("unsupported: mock p2p")
```
kostko, line 242:
```suggestion
	return nil, nil, fmt.Errorf("unsupported: p2p is disabled")
```

go/worker/common/p2p/rpc/client.go
```diff
@@ -215,6 +215,33 @@ type Client interface {
 	) ([]interface{}, []PeerFeedback, error)
 }
 
+type nopClient struct {
```
kostko, line 218: Maybe move the nop client and peer manager into a separate `nop.go`.

go/worker/common/p2p/rpc/peermgmt.go
```diff
@@ -325,6 +353,10 @@ func (mgr *peerManager) peerProtocolWatcher() {
 
 // NewPeerManager creates a new peer manager for the given protocol.
 func NewPeerManager(p2p P2P, protocolID protocol.ID, stickyPeers bool) PeerManager {
+	if p2p.GetHost() == nil {
+		// Mock P2P service, use the mock peer manager.
```
kostko, line 357:
```suggestion
		// No P2P service, use the no-op peer manager.
```

kostko (commented):
Nice work with abstracting out the P2P interface. Some minor comments.

go/oasis-node/cmd/node/node.go
ptrus: Unintended, reverted

go/worker/common/p2p/rpc/client.go
```diff
@@ -517,6 +517,10 @@ func NewClient(p2p P2P, runtimeID common.Namespace, protocolID string, version v
 		opt(&co)
 	}
 
+	if p2p.GetHost() == nil {
+		// Mock P2P service, use the mock client.
```
kostko, line 521:
```suggestion
		// No P2P service, use the no-op client.
```

kostko (approved):

PR #4806, opened 2022-06-16
runtime: Add support for PCS attestation

kostko:


go/common/sgx/pcs/quote.go
```diff
@@ -100,7 +113,25 @@ func (q *Quote) Verify(ts time.Time, tcb *TCBBundle) (*TCBLevel, error) {
 		return nil, fmt.Errorf("pcs/quote: disallowed debug/production enclave/mode combination")
 	}
 
-	return q.Signature.Verify(&q.Header, &q.ISVReport, ts, tcb)
+	if policy == nil {
```
Yawning, line 116: Hm, is it ok that the hardcoded default here would likely require a consensus breaking change to update?

go/common/sgx/pcs/tcb.go
```diff
@@ -158,35 +170,38 @@ type TCBInfo struct {
...
-		return fmt.Errorf("pcs/tcb: invalid issue date: %w", err)
+	if _, err = time.Parse(TimestampFormat, ti.NextUpdate); err != nil {
+		return fmt.Errorf("pcs/tcb: invalid next update date: %w", err)
 	}
 
 	if issueDate.After(ts) {
 		return fmt.Errorf("pcs/tcb: TCB info issue date in the future")
 	}
-	if nextUpdate.Before(ts) {
+	if ts.Sub(issueDate).Nanoseconds() > int64(policy.TCBValidityPeriod)*24*int64(time.Hour) {
```
Yawning, line 197: Comparing a difference expressed in ns, to an interval expressed in hours?

go/common/sgx/pcs/tcb.go
```diff
@@ -398,24 +447,27 @@ func (qe *QEIdentity) validate(ts time.Time) error {
...
-		return fmt.Errorf("pcs/tcb: invalid issue date: %w", err)
+	if _, err = time.Parse(TimestampFormat, qe.NextUpdate); err != nil {
+		return fmt.Errorf("pcs/tcb: invalid next update date: %w", err)
 	}
 
 	if issueDate.After(ts) {
 		return fmt.Errorf("pcs/tcb: QE identity issue date in the future")
 	}
-	if nextUpdate.Before(ts) {
+	if ts.Sub(issueDate).Nanoseconds() > int64(policy.TCBValidityPeriod)*24*int64(time.Hour) {
```
Yawning, line 463: Likewise, are the units here the same?  Doesn't look like it....

go/common/sgx/pcs/tcb.go
kostko: [`time.Hour`](https://pkg.go.dev/time#Hour) is the number of nanoseconds in an hour, so this should be fine?

go/common/sgx/pcs/quote.go
kostko: The alternative could be that we would have this default as part of the registry parameters (this PR introduces the TEE features parameter) that could be more easily updated through upgrade handlers. And only if those parameters were not set would this default be used (or this could just return an error if the policy is `nil`).

Agree that this would be better from an upgradability standpoint, but it would add some complexity. Let me look into how that would look like.

go/common/sgx/pcs/tcb.go
Yawning: Ok

go/common/sgx/pcs/quote.go
kostko: This has now been changed so that the default quote policy is part of the TEE features parameter.

pro-wh:
image
alright let's get reviewing

go/common/node/sgx_test.go
```diff
@@ -0,0 +1,131 @@
...
+	err = sa.ValidateBasic(&TEEFeatures{SGX: TEEFeaturesSGX{PCS: true}})
+	require.NoError(err, "ValidateBasic V1 SGX attestation")
+
+	enc := cbor.Marshal(sa)
+	require.EqualValues(enc, raw, "serialization should round-trip")
+}
+
+func FuzzSGXConstraints(f *testing.F) {
+	// Add some V0 constraints.
+	raw, _ := ioutil.ReadFile("testdata/sgx_constraints_v0.bin")
```
pro-wh, line 92: does testing.F work with require.noerror? if not can we have it panic here

go/common/node/sgx_test.go
```diff
@@ -0,0 +1,131 @@
...
+	f.Add(raw)
+	// TODO: Add some V1 constraints.
```
pro-wh, line 94: file issue?

go/common/sgx/pcs/pcs.go
```diff
@@ -31,12 +31,12 @@ type QuoteBundle struct {
 // Verify verifies the quote bundle.
 //
 // In case of successful verification it returns the TCB level.
```
pro-wh, line 33: update docstring?

go/common/sgx/pcs/quote.go
```diff
@@ -31,8 +31,21 @@ const (
 
 	// ppidDataLen is the PPID certification data length in bytes.
 	ppidDataLen = 404
+
+	// DefaultMinTCBEvaluationDataNumber is the default minimum TCB evaluation data number.
+	DefaultMinTCBEvaluationDataNumber = 12 // As of 2022-08-01.
```
pro-wh, line 36: what's this

go/common/sgx/pcs/tcb.go
```diff
@@ -36,37 +36,49 @@ type TCBBundle struct {
...
-) (*TCBLevel, error) {
+) error {
 	pk, err := bnd.getPublicKey(ts)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	tcbInfo, err := bnd.TCBInfo.open(ts, pk)
+	qeInfo, err := bnd.QEIdentity.open(ts, policy, pk)
```
pro-wh, line 53: :information_source: swapped :shrug: and added `policy` argument

go/common/sgx/pcs/quote.go
```diff
@@ -31,8 +31,21 @@ const (
...
 )
 
+// QuotePolicy is the quote validity policy.
+type QuotePolicy struct {
+	// TCBValidityPeriod is the validity (in days) of the TCB collateral.
+	TCBValidityPeriod uint16 `json:"tcb_validity_period"`
```
pro-wh, line 42: :information_source: previously we treated it as valid until the "next update" time

go/common/sgx/pcs/tcb.go
```diff
@@ -158,35 +170,38 @@ type TCBInfo struct {
...
+		issueDate time.Time
+		err       error
 	)
 	if issueDate, err = time.Parse(TimestampFormat, ti.IssueDate); err != nil {
 		return fmt.Errorf("pcs/tcb: invalid issue date: %w", err)
 	}
-	if nextUpdate, err = time.Parse(TimestampFormat, ti.NextUpdate); err != nil {
-		return fmt.Errorf("pcs/tcb: invalid issue date: %w", err)
+	if _, err = time.Parse(TimestampFormat, ti.NextUpdate); err != nil {
+		return fmt.Errorf("pcs/tcb: invalid next update date: %w", err)
```
pro-wh, line 191: so what was this then, if not something we should use to judge the validity period? should we even try to parse it, if we're not using it?

go/common/sgx/pcs/tcb.go
```diff
@@ -158,35 +170,38 @@ type TCBInfo struct {
 	FMSPC                   string     `json:"fmspc"`
 	PCEID                   string     `json:"pceId"`
 	TCBType                 int        `json:"tcbType"`
-	TCBEvaluationDataNumber int        `json:"tcbEvaluationDataNumber"`
+	TCBEvaluationDataNumber uint32     `json:"tcbEvaluationDataNumber"`
```
pro-wh, line 173: need citation, or I must suspect an adversary is trying to make `-1` sneak by the `< Min` check :laughing:

pro-wh (commented):
good stuff so far leading up to the actual PCS related work

go/common/node/sgx.go
```diff
@@ -0,0 +1,177 @@
...
+	default:
+		type scv1 SGXConstraints
+		return cbor.Marshal((*scv1)(sc)), nil
+	}
+}
+
+// ValidateBasic performs basic structure validity checks.
+func (sc *SGXConstraints) ValidateBasic(cfg *TEEFeatures) error {
+	if cfg == nil {
+		cfg = &TEEFeatures{}
```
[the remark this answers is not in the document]
Yawning: can also do:
```
// Maybe in global scope
var emptyFeatures TEEFeatures

cfg = &emptyFeatures
`

File: go/common/node/sgx.go [hunk 13]
```diff
@@ -0,0 +1,177 @@
...
+		return fmt.Errorf("invalid SGX constraints version: %d", v)
+	}
+}
+
+// MarshalCBOR is a custom serializer that handles different structure versions.
+func (sc *SGXConstraints) MarshalCBOR() ([]byte, error) {
+	switch sc.V {
+	case 0:
+		// Old version only supported the IAS-related constraints.
+		scv0 := sgxConstraintsV0{
```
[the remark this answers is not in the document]
kostko: The thing is that this needs to work with existing versions (e.g. V0) until PCS is enabled as it cannot break consensus. So support for both versions is needed. The Rust side handles this in a nicer way.

go/common/sgx/pcs/quote.go
kostko: What do you mean? It is the current TCB evaluation data number as of 2022-08-01. No reason to allow lower versions as in the future these can only increment with each TCB recovery.

> A monotonically increasing sequence number changed when Intel updates the content of the TCB evaluation data set: TCB Info, QE Idenity and QVE Identity. The tcbEvaluationDataNumber update is synchronized across TCB Info for all flavors of SGX CPUs (Family-Model-Stepping-Platform-CustomSKU) and QE/QVE Identity. This sequence number allows users to easily determine when a particular TCB Info/QE Idenity/QVE Identiy superseedes another TCB Info/QE Identity/QVE Identity (value: current TCB Recovery event number stored in the database).

go/common/sgx/pcs/tcb.go
kostko: We could skip parsing, I left it there to check whether the structure is well-formed (I don't expect it to ever be malformed as it is signed by Intel).

go/common/sgx/pcs/tcb.go
kostko: Citation for what? Unfortunately it is only defined as an `integer` in JSON. But in practice (current value is 12 and it is monotonically increasing) it can only be positive and this makes it easier to compare with the TCB evaluation data number set in the policy (which we control and enforce to be uint32 as we want explicit lengths in serializable structures).

go/common/node/sgx_test.go
kostko: Fixed.

go/common/node/sgx_test.go
kostko: Added V1 constraints instead :-)

go/common/sgx/pcs/pcs.go
kostko: Done.

runtime/src/consensus/tendermint/verifier.rs
```diff
@@ -203,6 +226,20 @@ impl Verifier {
...
     }
 
+    fn latest_consensus_state(&self, instance: &mut Instance) -> Result<ConsensusState, Error> {
+        let verified_block = instance
+            .light_client
+            .verify_to_highest(&mut instance.state)
+            .map_err(|err| Error::VerificationFailed(err.into()))?;
+        self.update_insecure_posix_time(&verified_block);
+
+        let state_root = state_root_from_header(&verified_block.signed_header);
```
peternose, line 236: Should we also update the cache, e.g. `verified_state_roots`?

kostko: The `verified_state_roots` are the runtime state roots while the above are consensus layer state roots (and `latest_state` is only dealing with consensus layer state roots). The light blocks that are fetched as part of `verify_to_highest` are already cached in the `LruStore` (see initialization in the `run` method), but the problem with `latest_state` is that it doesn't know what the latest height is, so it will in any case request this information from the node on each invocation.

But given that the node is notifying the runtime of new consensus layer blocks as they are finalized (which results in invoking the `sync` method of the verifier) we can record that and avoid the extra query on each `latest_state` request. I'll change it.

go/common/sgx/pcs/quote.go
pro-wh: oh thx for the description

go/common/sgx/pcs/tcb.go
pro-wh: https://json-schema.org/draft/2020-12/json-schema-validation.html#name-type this looks like json schema, and json schema doesn't constrain it to be nonnegative.

does go have a signed int32 type that we could use?

go/common/node/sgx.go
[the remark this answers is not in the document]
pro-wh: can we really enable PCS without breaking consensus? old nodes will still fail to parse the new ones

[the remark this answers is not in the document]
kostko: No, enabling is a breaking change but before it is enabled we must be able to parse V0. After we could disable it but we could also disable V0 later.

runtime/src/consensus/tendermint/verifier.rs
```diff
@@ -529,6 +533,23 @@ impl Verifier {
         Ok(state)
     }
 
+    fn update_insecure_posix_time(&self, verified_block: &TMLightBlock) {
+        // Update untrusted time if ahead. This makes sure that the enclave's sense of time is
+        // synced with consensus sense of time based on the fact that consensus time is harder to
+        // fake than host operating system time.
```
pro-wh, line 517: prior to this, were we using host operating system time? this commit doesn't remove any other calls to time::update_insecure_posix_time. and the API looks like it allows any caller to overwrite any previously set time. so will this harder-to-fake consensus time unpredictably battle easier-to-fake system time?

go/common/node/sgx.go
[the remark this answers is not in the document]
kostko: Note that marshal/unmarshal does not know whether PCS is enabled. Not supporting both makes the caller more messy.

pro-wh (commented):
commits after the pcs support commit also look ok

go/common/node/sgx.go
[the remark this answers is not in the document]
pro-wh: ah I see the migration handler added in a later commit

runtime/src/consensus/tendermint/verifier.rs
kostko: Previously (and still do with IAS quotes) we were also using IAS time which did battle with system time :) That is why enclaves aborted on some nodes with desynced clocks.

go/common/sgx/pcs/tcb.go
kostko: Yeah based on the JSON schema alone it can be negative, but given that the spec defines it as a monotonically increasing number and its latest value is 12, it will never be negative in practice and I believe it is fine if we reject negative values during parsing (which the JSON parser will do in case the destination type is an `uint32`).

Other `int` fields should maybe also be changed to more explicit types.

pro-wh: ok

pro-wh:
for the pcs stuff itself, do we have any design overview?

kostko:
> for the pcs stuff itself, do we have any design overview?

It is all part of Intel documentation:
* https://api.portal.trustedservices.intel.com/documentation
* https://api.trustedservices.intel.com/documents/Intel_SGX_PCK_Certificate_CRL_Spec-1.4.pdf
* https://www.intel.com/content/www/us/en/developer/articles/technical/quote-verification-attestation-with-intel-sgx-dcap.html
* https://www.intel.com/content/dam/develop/public/us/en/documents/intel-sgx-dcap-ecdsa-orientation.pdf
* https://www.intel.com/content/dam/develop/external/us/en/documents/intel-sgx-support-for-third-party-attestation-801017.pdf

pro-wh:
thanks, I'll read through

pro-wh (commented):
summary of the materials on PCS:

- PCS stands for "provisioning certification service," a component in Intel's DCAP line of work
- in the DCAP way of doing attestation, it's intended that the node obtain Intel's signature (regarding CPU authenticity) on its own, as opposed to the remote challenger asking IAS about a quote
- the node gets that first signature from PCS
- then, that signature from Intel certifies an Intel architectural enclave called the "provisioning certification enclave" and its "provisioning certification key" (PCK) which then signs for a custom quoting enclave and its "attestation key," which then the node then uses to produce quotes for specific application enclaves
- thus the proof in the DCAP way is an entire chain of signatures: Intel's root key in PCS -> PCK -> attestation key -> the enclave
- the overall workflow is very similar to how we would previously have nodes talk to IAS themselves (resulting in a proof we called an AVR), although this condenses multiple IAS interactions with a single PCS interaction when we want to generate proofs for multiple enclaves

and as for DCAP's stated goals:
- "eliminates runtime dependancies [sic] on external services" -> still have to contact PCS, so our use still depends on external services (unless we say node operators should do that on their own and it's not part of oasis-node? would that even make a practical difference?)
- "enables trust decisions to be made in-house" -> IAS previously gave various TCB info and we already had to make decisions in-house. this just seems more complicated, with more layers of non-application enclaves needing all sorts of comparisons

the actual implementations for generating and verifying these proofs were built in earlier PRs

extra questions:
- what do we use for the custom quoting enclave? an open source one from Intel?

pro-wh (approved):
entering this as approved in that I don't oppose this change, although I haven't looked at the implementation.

other reviewers, please continue to review.

authors, let me know if there's anything specific in the code that you want feedback on

kostko:
> still have to contact PCS, so our use still depends on external services (unless we say node operators should do that on their own and it's not part of oasis-node? would that even make a practical difference?)

It is slightly better in the sense that the PCS artifacts can be cached for a while while IAS responses cannot be (as everything is coupled together in AVRs). So a PCS outage would not immediately start causing availability problems (while IAS would) for nodes that have the artifacts cached. Note that currently we do not implement caching of TCB infos, but this can be done transparently in the future.

> what do we use for the custom quoting enclave? an open source one from Intel?

That's right. It is verified against the QE identity obtained from PCS.

pro-wh:
> the PCS artifacts can be cached for a while

nice!



> QE identity obtained from PCS

oh I see there's this endpoint https://api.portal.trustedservices.intel.com/documentation#pcs-qe-identity-v3

under the heading "... for ECDSA Attestation." there were aterials talking about how a custom quoting system could be built using ECDSA, so that must be an instantiation of the idea provided by Intel

PR #4808, opened 2022-06-20
net-runner: Fix regression introduced in #4564

matevz:
Do not add **any** key manager fixture entry, if no key manager is defined.

### DIFF
--- a/go/oasis-net-runner/fixtures/default.go
+++ b/go/oasis-net-runner/fixtures/default.go
@@ -115,9 +115,9 @@ func newDefaultFixture() (*oasis.NetworkFixture, error) {
 	usingKeymanager := len(viper.GetString(cfgKeymanagerBinary)) > 0
 
 	if viper.GetBool(cfgSetupRuntimes) {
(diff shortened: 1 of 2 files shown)


go/oasis-net-runner/fixtures/default.go
```diff
@@ -115,9 +115,9 @@ func newDefaultFixture() (*oasis.NetworkFixture, error) {
 	usingKeymanager := len(viper.GetString(cfgKeymanagerBinary)) > 0
 
 	if viper.GetBool(cfgSetupRuntimes) {
-		fixture.Runtimes = []oasis.RuntimeFixture{
+		if usingKeymanager {
 			// Key manager runtime.
-			{
+			keymanagerFixture := oasis.RuntimeFixture{
```
kostko, line 120: Since you don't use `keymanagerFixture` anywhere else, you could simplify by doing:
```suggestion
			fixture.Runtimes = append(fixture.Runtimes, oasis.RuntimeFixture{
```

kostko (approved):

PR #4832, opened 2022-06-30
go/runtime/registry: allow client nodes to run sgx runtimes

nhynes:
This PR allows client nodes to run SGX runtimes, which enables them to request keys from a (privately) peered key manager. The `RuntimeDebugForceELF` flag was replaced with a more general one. The former one doesn't seem to be used in this repo, so it should be okay.

### DIFF
--- a/go/runtime/registry/config.go
+++ b/go/runtime/registry/config.go
@@ -37,7 +37,7 @@ const (
 	CfgRuntimePaths = "runtime.paths"
 	// CfgSandboxBinary configures the runtime sandbox binary location.
 	CfgSandboxBinary = "runtime.sandbox.binary"
-	// CfgRuntimeEnv sets the runtime environment. Setting an environment that does not
+	// CfgRuntimeEnvironment sets the runtime environment. Setting an environment that does not
 	// agree with the runtime descriptor or system hardware will cause an error.
 	CfgRuntimeEnvironment = "runtime.environment"
 	// CfgRuntimeSGXLoader configures the runtime loader binary required for SGX runtimes.
--- a/go/runtime/registry/config.go
+++ b/go/runtime/registry/config.go
@@ -37,6 +37,9 @@ const (
 	CfgRuntimePaths = "runtime.paths"
 	// CfgSandboxBinary configures the runtime sandbox binary location.
 	CfgSandboxBinary = "runtime.sandbox.binary"
+	// CfgRuntimeEnv sets the runtime environment. Setting an environment that does not
+	// agree with the runtime descriptor or system hardware will cause an error.
+	CfgRuntimeEnvironment = "runtime.environment"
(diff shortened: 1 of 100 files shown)


go/runtime/registry/config.go
```diff
@@ -103,6 +103,17 @@ const (
...
 )
 
+const (
+	// RuntimeEnvironmentSGX specifies to run the runtime in SGX.
+	RuntimeEnvironmentSGX = "sgx"
+	// RuntimeEnvironmentELF specifies to run the runtime in the OS address space.
+	//
+	// Use of this runtime environment is only allowed if DebugDontBlameOasis flag is set.
+	RuntimeEnvironmentELF = "elf"
+	// RuntimeEnvironmentELF specifies to run the runtime in the most appropriate location.
```
kostko, line 113:
```suggestion
	// RuntimeEnvironmentAuto specifies to run the runtime in the most appropriate location.
```

go/runtime/registry/config.go
```diff
@@ -254,6 +267,9 @@ func newConfig(dataDir string, consensus consensus.Backend, ias ias.Endpoint) (*
 				if !forceNoSGX {
 					break
 				}
+				if runtimeEnv == CfgRuntimeEnvironment {
```
kostko, line 270: This seems wrong?

nhynes: my autocomplete sucks, but the Go type system sucks more

go/runtime/registry/config.go
nhynes: Oh, weird. I thought gofumpt catches this?

go/runtime/registry/config.go
kostko: That's the reason why the mode is typed :)

go/runtime/registry/config.go
kostko: Hm indeed, it usually does :thinking:

go/runtime/registry/config.go
```diff
@@ -37,6 +37,9 @@ const (
 	CfgRuntimePaths = "runtime.paths"
 	// CfgSandboxBinary configures the runtime sandbox binary location.
 	CfgSandboxBinary = "runtime.sandbox.binary"
+	// CfgRuntimeEnv sets the runtime environment. Setting an environment that does not
```
kostko, line 40: Looks like something is up with the linter, maybe since the update to Go 1.18?
```suggestion
	// CfgRuntimeEnvironment sets the runtime environment. Setting an environment that does not
```

nhynes: Okay, this is totally messed up. I'm going to fix the linter. Sorry for using so much `jernej-lint`. Next time won't have issues!

kostko (approved):

PR #4837, opened 2022-07-06
go: remove synced storage wrapper

ptrus:


go/runtime/registry/config.go
```diff
@@ -143,6 +143,15 @@ func (m RuntimeMode) IsClientOnly() bool {
 	return false
 }
 
+// HasLocalStorage returns true iff the mode is one that has local storage.
+func (m RuntimeMode) HasLocalStorage() bool {
```
kostko, line 147: This new method should also now be used in the storage worker constructor to make it consistent:
https://github.com/oasisprotocol/oasis-core/blob/d8e352bd7f1cca56b79bb21431da9bf97dbb22d2/go/worker/storage/worker.go#L45-L52

go/runtime/history/history.go
```diff
@@ -109,7 +124,14 @@ type runtimeHistory struct {
...
 
-	db *DB
+	db             *DB
+	blocksNotifier *pubsub.Broker
+
+	// Last storage synced round as reported by the storage backend (if enabled).
+	syncRoundLock          sync.RWMutex
+	lastStorageSyncedRound uint64
+
+	localStorageWorker bool
```
kostko, line 134:
```suggestion
	haveLocalStorageWorker bool
```

go/worker/storage/committee/node.go
```diff
@@ -569,11 +569,14 @@ func (n *Node) initGenesis(rt *registryApi.Runtime, genesisBlock *block.Block) e
...
+func (n *Node) flushSyncedState(summary *blockSummary) (uint64, error) {
 	n.syncedLock.Lock()
 	defer n.syncedLock.Unlock()
 
 	n.syncedState = *summary
+	if err := n.commonNode.Runtime.History().StorageSyncCheckpoint(n.ctx, n.syncedState.Round); err != nil {
+		return 0, err
+	}
 
 	// Wake up any round waiters.
```
kostko, line 581: Are these round waiters still needed now that synced storage is gone?

go/worker/compute/executor/committee/node.go
```diff
@@ -978,6 +978,33 @@ func (n *Node) startProcessingBatchLocked(batch *unresolvedBatch) {
...
+	blk := n.commonNode.CurrentBlock
+	blkCh, sub, err := n.commonNode.Runtime.History().WatchBlocks()
+	if err != nil {
+		n.logger.Error("failed to watch storage sync rounds", "err", err)
+		return
+	}
+	defer sub.Close()
+
+	n.logger.Debug("ensuring block round is synced to storage", "round", blk.Header.Round)
+OUTER:
```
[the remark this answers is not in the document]
ptrus: Good idea, moved

go/runtime/registry/config.go
ptrus: Updated

go/runtime/history/history.go
```diff
@@ -131,13 +153,61 @@ func (h *runtimeHistory) Commit(blk *roothash.AnnotatedBlock, roundResults *root
 	// Notify the pruner what the new round is.
 	h.pruneCh.In() <- blk.Block.Header.Round
 
+	// If no local storage worker, notify the block watcher that new block is committed,
+	// otherwise the storage-sync-checkpoint will do the notification.
+	if h.localStorageWorker {
+		return nil
+	}
+	h.blocksNotifier.Broadcast(blk)
```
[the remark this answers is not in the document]
ptrus: Added

go/runtime/history/history.go
```diff
@@ -131,13 +158,63 @@ func (h *runtimeHistory) Commit(blk *roothash.AnnotatedBlock, roundResults *root
...
+
 	return nil
 }
 
 func (h *runtimeHistory) ConsensusCheckpoint(height int64) error {
 	return h.db.consensusCheckpoint(height)
 }
 
+func (h *runtimeHistory) StorageSyncCheckpoint(ctx context.Context, round uint64) error {
+	h.logger.Info("storage sync checkpoint", "round", round, "have_local_storage", h.haveLocalStorageWorker)
```
kostko, line 176: Probably should be logged with debug verbosity.

go/runtime/history/history.go
```diff
@@ -131,13 +158,63 @@ func (h *runtimeHistory) Commit(blk *roothash.AnnotatedBlock, roundResults *root
...
+		return nil
+	default:
+		// Continue bellow.
+	}
+
+	annBlk, err := h.db.getBlock(round)
+	if err != nil {
+		return fmt.Errorf("runtime/history: storage sync block not found in history: %w", err)
+	}
+	h.logger.Info("storage sync checkpoint done", "round", round, "have_local_storage", h.haveLocalStorageWorker)
```
kostko, line 197: Debug verbosity.

go/runtime/txpool/txpool.go
```diff
@@ -449,6 +451,32 @@ func (t *txPool) checkTxBatch(ctx context.Context, rr host.RichRuntime) {
 	checkCtx, cancel := context.WithTimeout(ctx, checkTxTimeout)
 	defer cancel()
 
+	// Ensure block round is synced to storage.
```
kostko, line 454: How long can this block for? We should probably limit it.

An alternative would be doing checks against the latest block that we have synced, but this could be more out of date, not sure if this is a problem.

go/worker/compute/executor/committee/node.go
```diff
@@ -894,6 +894,31 @@ func (n *Node) runtimeExecuteTxBatch(
 	inputRoot hash.Hash,
 	inputs transaction.RawBatch,
 ) (*protocol.RuntimeExecuteTxBatchResponse, error) {
+	// Ensure block round is synced to storage.
```
kostko, line 897: This looks like a common operation (here and txpool), can it be extracted out?

go/runtime/history/history.go
```diff
@@ -122,7 +148,8 @@ func (h *runtimeHistory) RuntimeID() common.Namespace {
 	return h.runtimeID
 }
 
-func (h *runtimeHistory) Commit(blk *roothash.AnnotatedBlock, roundResults *roothash.RoundResults) error {
+func (h *runtimeHistory) Commit(blk *roothash.AnnotatedBlock, roundResults *roothash.RoundResults, notify bool) error {
+	h.logger.Info("committing block", "round", blk.Block.Header.Round, "notify", notify, "have_local_storage", h.haveLocalStorageWorker)
```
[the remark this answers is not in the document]
ptrus: Removed all of these added logs, as these were mainly used for debugging an e2e test issue in some previous iteration of the PR

go/runtime/txpool/txpool.go
ptrus: I think it should be generally pretty fast/not even needed? Maybe using the `checkCtx` (which uses `checkTxTimeout`) is a good limit for this (I missed this context before, that's why it uses the parent one). Can also add a shorter limit if we prefer.

I think that before this waiting was implicitly done by the synced storage, so I think it shouldn't be a big difference having this here?

go/worker/compute/executor/committee/node.go
ptrus: Extracted into a common method in the history interface.

go/runtime/txpool/txpool.go
kostko: Just using checkCtx sounds good.

go/runtime/history/history.go
```diff
@@ -131,13 +161,83 @@ func (h *runtimeHistory) Commit(blk *roothash.AnnotatedBlock, roundResults *root
...
+	h.syncRoundLock.Lock()
+	defer h.syncRoundLock.Unlock()
+	switch {
+	case round < h.lastStorageSyncedRound:
+		return fmt.Errorf("runtime/history: storage sync checkpoint at lower height (current: %d wanted: %d)", h.lastStorageSyncedRound, round)
+	case round == h.lastStorageSyncedRound:
+		// Nothing to do.
+		return nil
+	default:
+		// Continue bellow.
```
kostko, line 192:
```suggestion
		// Continue below.
```

go/runtime/history/history.go
```diff
@@ -131,13 +161,83 @@ func (h *runtimeHistory) Commit(blk *roothash.AnnotatedBlock, roundResults *root
...
+		case annBlk, ok := <-blkCh:
+			if !ok {
+				return 0, fmt.Errorf("runtime/history: watch blocks channel closed unexpectedly")
+			}
+
+			if annBlk.Block.Header.Round >= round {
+				return annBlk.Block.Header.Round, nil
+			}
+		case <-ctx.Done():
+			return 0, fmt.Errorf("runtime/history: wait round synced: %w", ctx.Err())
```
kostko, line 236: Why not just return `ctx.Err()`?

ptrus: Updated

kostko (approved):

PR #4839, opened 2022-07-08
go: Bump the curve25519-voi import

Yawning:
This also fixes the ECVRF version to v10 of the draft as that is what
is used in production right now.  When we have the opportunity to break
consensus, we SHOULD migrate to v13.


go/go.mod
```diff
@@ -41,7 +41,7 @@ require (
 	github.com/libp2p/go-libp2p-core v0.16.1
 	github.com/libp2p/go-libp2p-pubsub v0.7.0
 	github.com/multiformats/go-multiaddr v0.5.0
-	github.com/oasisprotocol/curve25519-voi v0.0.0-20211219162838-e9a669f65da9
+	github.com/oasisprotocol/curve25519-voi v0.0.0-20220708102147-0a8a51822cae
```
pro-wh, line 44: :information_source: cross reference: https://github.com/oasisprotocol/curve25519-voi/commit/0a8a51822caeb1af812cc6baec355c112bed1b99

pro-wh (approved):
as described

PR #4843, opened 2022-07-12
go/oasis-test-runner: support setting runtime state root and state

ptrus:
Adds support for setting runtime state root and initial runtime state.

Not needed by any existing tests, but useful for external tooling to initialize network with pre-existing runtime state.

### DIFF
--- a/go/oasis-net-runner/fixtures/default.go
+++ b/go/oasis-net-runner/fixtures/default.go
@@ -1,16 +1,20 @@
 package fixtures
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"math"
+	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/spf13/viper"
 
 	beacon "github.com/oasisprotocol/oasis-core/go/beacon/api"
 	"github.com/oasisprotocol/oasis-core/go/common"
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
 	"github.com/oasisprotocol/oasis-core/go/common/node"
 	"github.com/oasisprotocol/oasis-core/go/common/sgx"
 	"github.com/oasisprotocol/oasis-core/go/common/version"
@@ -19,6 +23,10 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis"
 	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
 	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
+	storage "github.com/oasisprotocol/oasis-core/go/storage/api"
+	"github.com/oasisprotocol/oasis-core/go/storage/database"
+	mkvsAPI "github.com/oasisprotocol/oasis-core/go/storage/mkvs/db/api"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/db/badger"
 )
 
 const (
@@ -32,6 +40,7 @@ const (
 	cfgRuntimeID               = "fixture.default.runtime.id"
 	cfgRuntimeBinary           = "fixture.default.runtime.binary"
 	cfgRuntimeVersion          = "fixture.default.runtime.version"
+	cfgRuntimeStatePath        = "fixture.default.runtime.state_path"
 	cfgRuntimeProvisioner      = "fixture.default.runtime.provisioner"
 	cfgRuntimeLoader           = "fixture.default.runtime.loader"
 	cfgSetupRuntimes           = "fixture.default.setup_runtimes"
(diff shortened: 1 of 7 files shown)


go/oasis-test-runner/oasis/compute.go
```diff
@@ -230,6 +233,22 @@ func (net *Network) NewCompute(cfg *ComputeCfg) (*Compute, error) {
...
+	// Initialize runtime state paths.
+	for i, path := range cfg.RuntimeStatePaths {
+		if path == "" {
+			continue
+		}
+		stateDir := registry.GetRuntimeStateDir(host.DataDir(), net.Runtimes()[i].descriptor.ID)
+		if err = os.MkdirAll(stateDir, os.ModePerm); err != nil {
+			return nil, fmt.Errorf("oasis/compute: failed creating runtime state dir: %w", err)
+		}
+		cmd := exec.Command("rsync", "-a", path, stateDir)
```
kostko, line 245: Is there a Go-native way to do this instead of having an implicit dependency on rsync?

Should maybe also add -S as these databases use sparse files? Although I'm not sure if this is also needed on the local filesystem.

ptrus: Here are some copy-dir implementations from the wild: [go/internal](https://github.com/golang/go/blob/b2b8872c876201eac2d0707276c6999ff3eb185e/src/cmd/go/internal/modcmd/vendor.go#L390-L422), [moby/moby]( https://github.com/moby/moby/blob/be628115bafce21680343cef30b7051bf8a98b71/daemon/graphdriver/copy/copy.go#L113-L256), [hasura/graphql](https://github.com/hasura/graphql-engine/blob/6532a658eb3dfae9dde251c6c9377f2268ecc8a8/cli/util/copy.go#L13-L115). 

Could add a similar helper function and stick it into some util/helper package.

go/oasis-net-runner/fixtures/default.go
```diff
@@ -221,6 +231,49 @@ func newDefaultFixture() (*oasis.NetworkFixture, error) {
...
+			if err != nil {
+				return nil, err
+			}
+
+			var stateRootHash hash.Hash
+			for _, r := range rts {
+				if r.Type == storage.RootTypeState {
+					stateRootHash = r.Hash
+				}
+			}
```
kostko, line 271: I would put this into a helper function (e.g. something like `getLatestStateRoot(path)`) to make this more clear.

go/oasis-test-runner/oasis/util.go
```diff
@@ -0,0 +1,112 @@
...
+	"path/filepath"
+)
+
+// from: https://github.com/hasura/graphql-engine/blob/6532a658eb3dfae9dde251c6c9377f2268ecc8a8/cli/util/copy.go#L13-L115
+// CopyFile copies the contents of the file named src to the file named
+// by dst. The file will be created if it does not already exist. If the
+// destination file exists, all it's contents will be replaced by the contents
+// of the source file. The file mode will be copied from the source and
+// the copied data is synced/flushed to stable storage.
+func CopyFile(src, dst string) (err error) {
```
kostko, line 17: These should probably be moved to the `go/common` package (there's already `Mkdir` in there).

go/oasis-test-runner/oasis/fixture.go
```diff
@@ -393,6 +396,8 @@ type ComputeWorkerFixture struct {
 
 	// Runtimes contains the indexes of the runtimes to enable.
 	Runtimes []int `json:"runtimes,omitempty"`
+	// RuntimeStatePaths are the paths to runtime state that will be copied to the compute worker node.
+	RuntimeStatePaths []string `json:"runtime_state_paths,omitempty"`
```
kostko, line 400: We should instead change the `RuntimeStatePaths` to be a map, similar to the `RuntimeConfig`, e.g. `map[int]string`. This would avoid the runtime alignment requirement.

go/oasis-net-runner/fixtures/default.go
```diff
@@ -221,12 +231,68 @@ func newDefaultFixture() (*oasis.NetworkFixture, error) {
...
 			}
 			fixture.Clients[0].Runtimes = append(fixture.Clients[0].Runtimes, rtIndex)
+
+			// Runtime state paths to use to initialize the runtime with.
+			runtimeStatePath := runtimeStatePaths[i]
+			// Set workers runtime state.
+			for j := range fixture.ComputeWorkers {
+				fixture.ComputeWorkers[j].RuntimeStatePaths[i] = runtimeStatePath
+			}
+			if runtimeStatePath == "" {
```
kostko, line 241: This can probably be moved before the above for loop now that the `RuntimeStatePaths` is a map.

kostko (approved):

go/oasis-net-runner/fixtures/default.go
ptrus: yep, fixed

PR #4877, opened 2022-08-05
go/worker/keymanager: Add key manager worker metrics

peternose:
Currently the key manager worker lacks metrics that would enable the operator to gain some useful insights into its operation. We should add the following metrics:
- Number of key manager policy updates
- Number of compute runtimes using the key manager
- Number of remote EnclaveRPC requests via P2P (separate per method)


kostko (approved):
Nice, looks good!

One thing that you need to add is a [changelog fragment](https://github.com/oasisprotocol/oasis-core/blob/master/.changelog/README.md) that describes what changes are there when making a release.

PR #4883, opened 2022-08-08
go: Add key manager worker status to node status

peternose:
**Task**
Currently the key manager worker does not report any status as part of the GetStatus node control gRPC API call. It should provide some useful information about the operation of the key manager node.

**Test**
Tested locally using default fixture, simple-keymanager and minimal-runtime:
- Key manager status is shown only for keymanager nodes and not for compute, client and validator nodes.
- Status changes to ready once consensus is synced and key manager is registered.
- Client runtimes and access list populates with a short delay.
- Private peers are shown if we set them using the worker.keymanager.private_peer_pub_keys flag.

### DIFF
--- a/go/worker/keymanager/api/api.go
+++ b/go/worker/keymanager/api/api.go
@@ -1,0 +1,100 @@
+package api
+
+import (
+	"fmt"
+
+	core "github.com/libp2p/go-libp2p-core"
+
+	"github.com/oasisprotocol/oasis-core/go/common"
+)
+
+// StatusState is the concise status state of the key manager worker.
+type StatusState uint8
+
+const (
+	// StatusStateReady is the ready status state.
+	StatusStateReady StatusState = 0
+	// StatusStateStarting is the starting status state.
+	StatusStateStarting StatusState = 1
+	// StatusStateStopped is the stopped status state.
+	StatusStateStopped StatusState = 2
+	// StatusStateDisabled is the disabled status state.
+	StatusStateDisabled StatusState = 3
+)
+
+// String returns a string representation of a status state.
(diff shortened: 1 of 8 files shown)


go/worker/keymanager/status.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	var rid *common.Namespace
+	if w.runtime != nil {
+		id := w.runtime.ID()
+		rid = &id
+	}
+
+	rts := make([]common.Namespace, 0, len(w.clientRuntimes))
+	for rt := range w.clientRuntimes {
```
kostko, line 45: Accessing this outside the `worker` goroutine will cause a potential data race as it is currently not protected by a lock. In this case it would probably be best to move the map constructor outside the worker and make sure to only access/manipulate it with the appropriate lock held.

go/worker/keymanager/api/api.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/oasisprotocol/oasis-core/go/common"
+)
+
+// StatusState is the concise status state of the key manager worker.
+type StatusState uint8
+
+const (
+	// StatusStateReady is the ready status state.
+	StatusStateReady StatusState = iota
```
kostko, line 16: Given that these are serialized please use explicit values instead of relying on `iota` to ensure the encoding remains stable.

kostko (commented):
One minor thing, otherwise looks good.

kostko:
Also do squash the two commits as the second one actually fixes an issue introduced by the first one :)

kostko (approved):

PR #4888, opened 2022-08-11
Add support for ephemeral keys in key manager

peternose:
Add API methods like the following:
- `get_ephemeral_key(epoch, key_pair_id)`
- `get_public_ephemeral_key(epoch, key_pair_id)`

Those would derive special ephemeral X25519 key pairs separate from the other runtime keys and where the epoch number is used as part of the derivation, but would only derive them if the epoch was not too far back in the past based on consensus layer state as seen by the key manager. Initially the epoch check may rely on untrusted consensus state provided by the key manager host node.

### DIFF
--- a/go/oasis-test-runner/scenario/e2e/runtime/keymanager_key_generation.go
+++ b/go/oasis-test-runner/scenario/e2e/runtime/keymanager_key_generation.go
@@ -1,0 +1,240 @@
+package runtime
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	beacon "github.com/oasisprotocol/oasis-core/go/beacon/api"
+	"github.com/oasisprotocol/oasis-core/go/common"
+	"github.com/oasisprotocol/oasis-core/go/common/cbor"
+	consensus "github.com/oasisprotocol/oasis-core/go/consensus/api"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/env"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario"
+)
+
+// KeymanagerKeyGeneration is the keymanager key generation scenario.
+//
+// It uses encryption and decryption transactions provided by the
+// simple key/value runtime to test whether the key manager client
+// can retrieve private and public ephemeral keys from the key manager
+// and if the latter generates those according to the specifications.
+var KeymanagerKeyGeneration scenario.Scenario = newKmKeyGenerationImpl()
+
+type kmKeyGenerationImpl struct {
+	runtimeImpl
+}
+
+func newKmKeyGenerationImpl() scenario.Scenario {
+	return &kmKeyGenerationImpl{
+		runtimeImpl: *newRuntimeImpl("keymanager-key-generation", BasicKVEncTestClient),
+	}
+}
+
+func (sc *kmKeyGenerationImpl) Fixture() (*oasis.NetworkFixture, error) {
+	return sc.runtimeImpl.Fixture()
+}
+
+func (sc *kmKeyGenerationImpl) Clone() scenario.Scenario {
+	return &kmKeyGenerationImpl{
+		runtimeImpl: *sc.runtimeImpl.Clone().(*runtimeImpl),
+	}
+}
+
+func (sc *kmKeyGenerationImpl) Run(childEnv *env.Env) error {
+	// Start the network, but no need to start the client. Just ensure it
+	// is synced.
+	ctx := context.Background()
+	if err := sc.runtimeImpl.startNetworkAndWaitForClientSync(ctx); err != nil {
+		return err
+	}
+
+	// Initialize the nonce DRBG.
+	rng, err := drbgFromSeed(
+		[]byte("oasis-core/oasis-test-runner/e2e/runtime/keymanager-key-generation"),
+		[]byte("keymanager-key-generation"),
+	)
+	if err != nil {
+		return err
+	}
+
+	// Data needed for encryption and decryption.
+	keyPairID := "key pair id"
+	plaintext := []byte("The quick brown fox jumps over the lazy dog")
+
+	epoch, err := sc.Net.Controller().Beacon.GetEpoch(ctx, consensus.HeightLatest)
+	if err != nil {
+		return fmt.Errorf("failed to get epoch at the latest height: %w", err)
+	}
+
+	// Encrypt plaintext using ephemeral public key for the current epoch.
+	// Successful encryption indicates that the key manager generated
+	// an ephemeral public key.
+	sc.Logger.Info("encrypting plaintext")
+	ciphertext, err := sc.submitKeyValueRuntimeEncryptTx(
+		ctx,
+		runtimeID,
+		rng.Uint64(),
+		epoch,
+		keyPairID,
+		plaintext,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to encrypt plaintext: %w", err)
+	}
+
+	// Decrypt ciphertext using ephemeral private key for the current epoch.
+	// Successful decryption indicates that the key manager generates
+	// matching public and private ephemeral keys.
+	sc.Logger.Info("decrypting ciphertext")
+	decrypted, err := sc.submitKeyValueRuntimeDecryptTx(
+		ctx,
+		runtimeID,
+		rng.Uint64(),
+		epoch,
+		keyPairID,
+		ciphertext,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to decrypt ciphertext: %w", err)
+	}
+	if !reflect.DeepEqual(decrypted, plaintext) {
+		return fmt.Errorf("decrypted ciphertext does match the plaintext (got: '%s', expected: '%s')", decrypted, plaintext)
+	}
+
+	// Decrypt ciphertext using ephemeral private key for the previous epoch.
+	// As ephemeral keys are derived from the epoch, the decryption should
+	// fail.
+	sc.Logger.Info("decrypting ciphertext with wrong epoch")
+	decrypted, err = sc.submitKeyValueRuntimeDecryptTx(
+		ctx,
+		runtimeID,
+		rng.Uint64(),
+		epoch-1,
+		keyPairID,
+		ciphertext,
+	)
+	if err == nil && reflect.DeepEqual(decrypted, plaintext) {
+		return fmt.Errorf("decryption with wrong epoch should fail or produce garbage")
+	}
+
+	// Decrypt ciphertext using ephemeral private key derived from wrong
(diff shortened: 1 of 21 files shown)


keymanager-api-common/src/api.rs
```diff
@@ -98,25 +99,34 @@ pub struct ReplicateResponse {
...
     }
+}
 
-    pub fn to_cache_key(&self) -> Vec<u8> {
-        let mut k = self.runtime_id.as_ref().to_vec();
-        k.extend_from_slice(self.key_pair_id.as_ref());
-        k
+/// Request runtime/key pair id tuple.
+#[derive(Clone, Default, cbor::Encode, cbor::Decode)]
+pub struct EphemeralKeyRequest {
```
kostko, line 120: This also needs to be runtime-scoped similar to regular keys as a key manager can serve multiple runtimes and they should have distinct keys.

keymanager-api-common/src/api.rs
```diff
@@ -98,25 +99,34 @@ pub struct ReplicateResponse {
 
 /// Request runtime/key pair id tuple.
 #[derive(Clone, Default, cbor::Encode, cbor::Decode)]
-pub struct RequestIds {
+pub struct RuntimeKeyRequest {
```
kostko, line 102: Maybe this should be something like `LongTermKeyRequest` to contrast with ephemeral ones? As all keys generated by the key manager are "runtime keys".

keymanager-api-common/src/api.rs
```diff
@@ -171,6 +181,14 @@ impl KeyPair {
     }
 }
 
+#[derive(PartialEq, Eq, Hash)]
+pub enum KeyPairType {
+    /// A runtime key.
+    Runtime = 0,
```
kostko, line 187: As above, should probably be `LongTerm` or similar.

keymanager-lib/src/methods.rs
```diff
@@ -30,6 +39,35 @@ pub fn get_public_key(req: &RequestIds, _ctx: &mut RpcContext) -> Result<Option<
     pk.map_or(Ok(None), |pk| Ok(Some(kdf.sign_public_key(pk)?)))
 }
 
+/// See `Kdf::get_or_create_keys`.
+pub fn get_or_create_ephemeral_keys(
+    req: &EphemeralKeyRequest,
+    ctx: &mut RpcContext,
+) -> Result<KeyPair> {
+    // Derive ephemeral keys only for epochs which are not too far back in the past.
```
kostko, line 47: Don't also forget about client authentication (possibly refactor a bit to avoid too much duplication) here as otherwise anyone could derive private ephemeral keys.

kostko (commented):
Some early comments on the current draft.

tests/runtimes/simple-keyvalue/src/methods.rs
```diff
@@ -350,6 +356,92 @@ impl Methods {
...
+                .get_public_ephemeral_key(io_ctx, key_pair_id, args.epoch);
+        let long_term_pk = tokio::runtime::Handle::current()
+            .block_on(result)
+            .map_err(|err| err.to_string())?
+            .ok_or("public ephemeral key not available")?;
+
+        // Generate ephemeral key.
+        let mut rng = OsRng {};
+        let ephemeral_sk = x25519_dalek::StaticSecret::new(&mut rng);
+        let ephemeral_pk = x25519_dalek::PublicKey::from(&ephemeral_sk);
```
kostko, line 383: This happens in a replicated manner across nodes and as such it will cause discrepancies if replication factor > 1 is used. Any randomness must be deterministic.

The only reason the E2E test passes is:
* In this test, the primary replication group size is 2 and the backup replication group size is 1. So even if this causes a discrepancy it is resolved by evaluating in the backup group which only has one member.
* The E2E test does not check whether any issues were detected. It needs to call `sc.Net.CheckLogWatchers()` at the end of the scenario.

go/oasis-test-runner/scenario/e2e/runtime/keymanager_key_generation.go
```diff
@@ -0,0 +1,236 @@
...
+func (sc *kmKeyGenerationImpl) Run(childEnv *env.Env) error {
+	ctx := context.Background()
+	if err := sc.runtimeImpl.startNetworkAndTestClient(ctx, childEnv); err != nil {
+		return err
+	}
+
+	// Wait for the client to exit.
+	if err := sc.waitTestClientOnly(); err != nil {
+		return err
+	}
```
kostko, line 53: Can remove the test client for this test case. Use just `sc.startNetworkAndWaitForClientSync` instead.

go/oasis-test-runner/scenario/e2e/runtime/keymanager_key_generation.go
```diff
@@ -0,0 +1,236 @@
...
+		rng.Uint64(),
+		epoch,
+		keyPairID,
+		ciphertext,
+	)
+	if err == nil {
+		return fmt.Errorf("decryption with invalid epoch should fail")
+	}
+
+	return nil
```
kostko, line 177:
```suggestion
	return sc.Net.CheckLogWatchers()
```

tests/runtimes/simple-keyvalue/src/methods.rs
```diff
@@ -350,6 +356,92 @@ impl Methods {
...
+            vec![],
+            &long_term_pk.key.0,
+            &ephemeral_sk.to_bytes(),
+        )
+        .map_err(|err| format!("failed to encrypt plaintext: {}", err))?;
+
+        // Encode ephemeral_pk || ciphertext to Base64.
+        let mut v = ephemeral_pk.as_bytes().to_vec();
+        v.extend(ciphertext);
+        let ciphertext = base64::encode(v);
```
kostko, line 398: Why base64 encoding? Everything is serialized into CBOR which is a binary format anyway.

tests/runtimes/simple-keyvalue/src/types.rs
```diff
@@ -40,6 +40,20 @@ pub struct KeyValue {
     pub value: String,
 }
 
+#[derive(Clone, Debug, Default, cbor::Encode, cbor::Decode)]
+pub struct Encrypt {
+    pub epoch: u64,
+    pub key_pair_id: String,
+    pub plaintext: String,
```
kostko, line 47: Would probably be easier to make these `Vec<u8>` (and `[]byte` on the Go side) instead of using base64.

tests/runtimes/simple-keyvalue/src/methods.rs
peternose: You are right. Now all makes sense as I was getting the following errors if I extended the client.
`"log assertion failed: round failure detected"`
`"log assertion failed: execution discrepancy detected"`

go/oasis-test-runner/scenario/e2e/runtime/keymanager_key_generation.go
peternose: Thought it was a good practice to always run the client just to be sure that everything works. But I guess we don't need this in this case as we are already testing it.

tests/runtimes/simple-keyvalue/src/methods.rs
peternose: Wanted to be consistent with plaintext's type, plus it is easier to log the ciphertext if you want/need to debug. Will change to `[]byte` everywhere, so that encryption is not limited to utf8.

kostko:
Given that #4806 has been merged, please rebase on master.

tests/runtimes/simple-keyvalue/src/methods.rs
```diff
@@ -350,6 +354,94 @@ impl Methods {
...
+        let result =
+            ctx.parent
+                .key_manager
+                .get_public_ephemeral_key(io_ctx, key_pair_id, args.epoch);
+        let long_term_pk = tokio::runtime::Handle::current()
+            .block_on(result)
+            .map_err(|err| err.to_string())?
+            .ok_or("public ephemeral key not available")?;
+
+        // Generate ephemeral key. Not secure, but good enough for testing purposes.
```
Yawning, line 379: I sure hope no one copy pastes this (or the casual nonce reuse) into actual code.

We have boilerplate code for all of this under runtime/src/common/crypto/mrae, why reinvent the wheel?

Yawning (commented):
My inner tinfoil hat says that the master secret used to derive ephemeral keys should be distinct and have nothing to do with the master secret used to derive the long term keys.

(And yes, I am aware that the current way of doing things is fine as long as KMAC is.)

tests/runtimes/simple-keyvalue/src/methods.rs
peternose: To use mrae I need a public key and a private key. In my case I want to encrypt a message using public key only. To do that I created an unsafe non-random private key and used it in mrae on line 384. So I think I'm not reinventing anything.

I constructed this unsafe encryption just to test if key manager can generate ephemeral keys. I can probably test this some other way also, e.g. by using signatures (but again the result should be deterministic) or by just returning a simple zero knowledge proof.

Yawning: Ok.

kostko (approved):

Yawning:
Should we generate test vectors for the KDF so we don't inadvertently break it? (Since it's deterministic this should be possible)

kostko:
Yeah that's a good idea. They should probably be added as Rust unit tests for the KDF.

keymanager-lib/src/kdf.rs
```diff
@@ -683,4 +693,133 @@ mod tests {
...
+                kdf_custom,
+                xof_custom,
+                sk: "70c077d8bd0c5c2be66d08fbacac4f1a996001883baaf62bd933e2f7de390e5f",
+                pk: "2d62567e9061a7cc64df3793fbee65c863862b8f0acc992e2a39e32757f2c743",
+            },
+        ];
+
+        // Values that don't effect key derivation.
+        let checksum = "100246b37912e916e71ff79982b2f62be892ddf1025cde84804f67e7f5713b75";
+        let runtime_id = "8000000000000000000000000000000000000000000000000000000000000000";
```
kostko, line 790: Runtime ID does affect key derivation so it should be part of the test vectors.

peternose: Runtime ID in kdf struct does not. However, runtime ID in the `LongTermKeyRequest` does at it effects `seed`. But here I only test if kdf returns the same values for the same inputs (seed, kdf, xof).
Another test could be that `LongTermKeyRequest` produces seeds that depend on the runtime ID. I do not test that but I do test in `requests_have_unique_customs` that customs are different which means that kmac and keys will be different.

Will add missing test that requests produce seeds that are dependent on their values (runtime id, key_pair_id, epoch).

kostko: Oh I see, this was refactored, yeah makes sense to add explicit tests for those requests as well.

kostko (approved):

Yawning (approved):

PR #4893, opened 2022-08-19
runtime: Add state_at to consensus Verifier trait

kostko:
* runtime: Add state_at to consensus Verifier trait
  
  This allows fetching historic consensus layer state in addition to the
  latest state.
* runtime: Refactor dispatcher initializer and expose consensus_verifier

### DIFF
--- a/runtime/src/consensus/tendermint/verifier.rs
+++ b/runtime/src/consensus/tendermint/verifier.rs
@@ -78,6 +78,21 @@ impl NopVerifier {
     pub fn new(protocol: Arc<Protocol>) -> Self {
         Self { protocol }
     }
+
+    fn fetch_light_block(&self, height: u64) -> Result<LightBlock, Error> {
+        let result = self
+            .protocol
+            .call_host(
+                Context::background(),
+                Body::HostFetchConsensusBlockRequest { height },
+            )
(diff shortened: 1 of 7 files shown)


abukosek (approved):

runtime/src/consensus/tendermint/verifier.rs
```diff
@@ -248,11 +251,19 @@ impl Verifier {
...
+        &self,
+        instance: &mut Instance,
+        height: u64,
+    ) -> Result<ConsensusState, Error> {
         let verified_block = instance
             .light_client
             .verify_to_target(height.try_into().unwrap(), &mut instance.state)
             .map_err(|err| Error::VerificationFailed(err.into()))?;
-        // No need to update time as this has already been done once for this height.
+        self.update_insecure_posix_time(&verified_block);
```
peternose, line 278: Should we update the cache also? 
`cache.update_verified_block(&verified_block);`

Seems like `verify_to_target`, `update_insecure_posix_time` and `update_verified_block` should always be called together, maybe in the future we refactor and group them.

kostko: Both good ideas, updated. And we can do the suggested refactor in a subsequent PR.

Yawning (approved):

PR #4903, opened 2022-08-24
runtime/consensus/verifier: Support trust root consensus layer upgrades

peternose:
Tested locally. Without changes compute workers are stuck with error messages `Consensus verifier failed to initialize`.
```
{"err":"builder: I/O error\n\nCaused by:\n   0: rpc error\n   1: server error: module: consensus code: 3 message: consensus: version not found\n, ...}
```
With changes compute workers report error message `Not enough trust to accept new chain context` when there is not enough trust to do the transition
```
{"level":"info","log_event":"consensus/tendermint/verifier/chain_context/no_trust","module":"runtime/consensus/tendermint/verifier","msg":"Not enough trust to accept new chain context","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","runtime_name":"test-runtime","tally":"VotingPowerTally { total: 3, tallied: 2, trust_threshold: TrustThresholdFraction { numerator: 2, denominator: 3 } }","ts":"2022-08-24T04:30:43.604572699Z"}
```
or `Failed to accept new chain context` if new light blocks are invalid (e.g. invalid time, height, hash, ...).
```
{"error":"NonIncreasingHeight(NonIncreasingHeightSubdetail { got: block::Height(46), expected: block::Height(51) })","level":"info","log_event":"consensus/tendermint/verifier/chain_context/failed","module":"runtime/consensus/tendermint/verifier","msg":"Failed to accept new chain context","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","runtime_name":"test-runtime","ts":"2022-08-24T04:29:42.405580653Z"}
```

### DIFF
--- a/.buildkite/scripts/test_e2e.sh
+++ b/.buildkite/scripts/test_e2e.sh
@@ -53,8 +53,8 @@ ${test_runner_binary} \
     --e2e/runtime.runtime.loader ${WORKDIR}/target/default/debug/oasis-core-runtime-loader \
     --e2e/runtime.tee_hardware ${OASIS_TEE_HARDWARE:-""} \
     --e2e/runtime.ias.mock=${ias_mock} \
-    --e2e/runtime/trust-root.runtime.source_dir ${WORKDIR}/tests/runtimes \
-    --e2e/runtime/trust-root.runtime.target_dir ${WORKDIR}/target \
+    --e2e/runtime/trust-root/{simple,change,change-fails}.runtime.source_dir=${WORKDIR}/tests/runtimes \
+    --e2e/runtime/trust-root/{simple,change,change-fails}.runtime.target_dir=${WORKDIR}/target \
     --remote-signer.binary ${WORKDIR}/go/oasis-remote-signer/oasis-remote-signer \
     --plugin-signer.name example \
     --plugin-signer.binary ${WORKDIR}/go/oasis-test-runner/scenario/pluginsigner/example_signer_plugin/example_signer_plugin \
--- a/go/oasis-test-runner/scenario/e2e/runtime/trust_root_change.go
+++ b/go/oasis-test-runner/scenario/e2e/runtime/trust_root_change.go
@@ -1,0 +1,417 @@
+package runtime
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+
+	"github.com/oasisprotocol/oasis-core/go/common/node"
+	genesis "github.com/oasisprotocol/oasis-core/go/genesis/api"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/env"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/log"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario/e2e"
+)
+
+const (
+	// LogEventTrustRootChangeNoTrust is the event emitted when a compute
+	// worker fails to initialize the verifier as there is not enough trust
+	// in the new light block.
+	LogEventTrustRootChangeNoTrust = "consensus/tendermint/verifier/chain_context/no_trust"
+
+	// LogEventTrustRootChangeFailed is the event emitted when a compute
+	// worker fails to initialize the verifier as the new light block is
+	// invalid, e.g. has lower height than the last known trusted block.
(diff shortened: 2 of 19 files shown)


go/oasis-test-runner/scenario/e2e/runtime/trust_root_change.go
```diff
@@ -0,0 +1,444 @@
...
+	for _, n := range sc.Net.ComputeWorkers() {
+		if err := n.Start(); err != nil {
+			return fmt.Errorf("failed to start node: %w", err)
+		}
+	}
+	for _, n := range sc.Net.ComputeWorkers() {
+		if err := n.WaitReady(ctx); err != nil {
+			if status, ok := status.FromError(err); ok {
+				switch status.Code() {
+				case codes.Canceled, codes.DeadlineExceeded:
```
kostko, line 267: How long will this wait in practice? The full 5 minutes? This could make the test overly long so it would be great if this could be reduced.

.buildkite/scripts/test_e2e.sh
```diff
@@ -53,8 +53,8 @@ ${test_runner_binary} \
     --e2e/runtime.runtime.loader ${WORKDIR}/target/default/debug/oasis-core-runtime-loader \
     --e2e/runtime.tee_hardware ${OASIS_TEE_HARDWARE:-""} \
     --e2e/runtime.ias.mock=${ias_mock} \
-    --e2e/runtime/trust-root.runtime.source_dir ${WORKDIR}/tests/runtimes \
-    --e2e/runtime/trust-root.runtime.target_dir ${WORKDIR}/target \
+    --e2e/runtime/trust-root/{simple,change,change-fails}.runtime.source_dir=${WORKDIR}/tests/runtimes \
+    --e2e/runtime/trust-root/{simple,change,change-fails}.runtime.target_dir=${WORKDIR}/target \
```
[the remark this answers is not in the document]
peternose: It's not, already tried.
`Error: unknown flag: --e2e/runtime/trust-root.runtime.source_dir`

runtime/src/consensus/tendermint/verifier.rs
```diff
@@ -765,14 +809,110 @@ impl Verifier {
...
+        // Chain context transition cannot be done directly from the embedded
+        // trust root as we don't have access to the matching trusted light
+        // block which validator set we need to verify blocks from the new chain.
+        let trusted_block = trusted_state
+            .trusted_block
+            .ok_or(Error::ChainContextTransitionFailed)?;
+
+        // Fetch the latest block and prepare untrusted state for verification.
+        let untrusted_block = io
+            .fetch_light_block(components::io::AtHeight::Highest)
```
kostko, line 846: It would be great if this could be done at genesis as the validator set could change between genesis and highest. Note that in case someone forgets to upgrade a node for a while the difference between genesis and highest could be potentially large. The question is how to determine genesis of the new network.

peternose: Changed. I used the following approach. First I ask the host for genesis height, then I fetch the genesis block and finally I verify if this block is the first one (`last_block_id` in the `header` should be `None`).

go/runtime/registry/host.go
```diff
@@ -242,6 +242,15 @@ func (h *runtimeHostHandler) Handle(ctx context.Context, body *protocol.Body) (*
 			Block: *lb,
 		}}, nil
 	}
+	if body.HostFetchGenesisHeightRequest != nil {
+		doc, err := h.consensus.GetGenesisDocument(ctx)
```
kostko, line 246: Could also introduce the `GetGenesisBlock` method to the consensus backend.

kostko (commented):
Looks nice, thanks! Just one minor thing that could maybe be changed.

go/runtime/host/protocol/types.go
```diff
@@ -478,6 +480,14 @@ type HostFetchConsensusBlockResponse struct {
 	Block consensus.LightBlock `json:"block"`
 }
 
+// HostFetchGenesisHeightRequest is a request to host to fetch the consensus genesis height.
+type HostFetchGenesisHeightRequest struct{}
```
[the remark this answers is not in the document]
peternose: I thought about both options and decided for my version for the following reasons:
- Round trip is not a problem as this call happens only after chain context changes (after upgrade, almost never). Furthermore, I think that tendermint implementation has genesis doc stored locally, so round-trip is only between host-runtime.
- There was no special height for genesis and I hate adding extra flags. Things would be simpler if 0 was the lowest block and -1 (max value) the latest one.

I can change to any version you prefer. Your call.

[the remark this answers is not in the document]
kostko: Ok, let's leave it as-is.

kostko (approved):

PR #4904, opened 2022-08-25
runtime: Support consensus event queries

kostko:
Based on #4893


runtime/src/consensus/tendermint/verifier.rs
```diff
@@ -282,6 +304,7 @@ impl Verifier {
         let state_root = state_root_from_header(&verified_block.signed_header);
         Ok(ConsensusState::from_protocol(
             self.protocol.clone(),
+            state_root.version + 1,
```
peternose, line 307: Should we move this to `from_protocol`?

kostko: No because it is specific to the Tendermint consensus backend so IMO it should be determined here. I agree that it is somewhat repetitive here and we could maybe move this into a function.

I've tried to explain this in the `ConsensusState` implementation:
```rust
pub struct ConsensusState {
    // An explicit height field is needed because the relationship between the underlying consensus
    // height and the corresponding state root is a consensus backend implementation detail.
    height: u64,
```

go/staking/api/api_test.go
```diff
@@ -506,3 +507,151 @@ func TestReclaimEscrowResultsSerialization(t *testing.T) {
...
+	require := require.New(t)
+
+	pk1 := signature.NewPublicKey("aaafffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff")
+	addr1 := NewAddress(pk1)
+	pk2 := signature.NewPublicKey("bbbfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff")
+	addr2 := NewAddress(pk2)
+	var txHash hash.Hash
+	txHash.Empty()
+
+	// NOTE: These cases should be synced with tests in runtime/src/consensus/staking.rs.
```
Yawning, line 521: Long term, instead of hardcoding crap like this, we should start sticking testvectors into JSON files that are used by both test cases.

go/runtime/host/logger.go
```diff
@@ -77,7 +79,8 @@ func (w RuntimeLogWrapper) processLogLine(line []byte) {
 	// Interpret line as JSON.
 	var m map[string]interface{}
 	if err := json.Unmarshal(line, &m); err != nil {
-		w.logger.Warn("non-JSON log line from runtime", "log_line", string(line), "err", err)
+		// If not valid JSON, forward line as normal log message with local timestamp.
```
Yawning, line 82: Should this get it's own changelog entry/commit?

Yawning (commented):
Doesn't this need to bump the runtime host protocol?

go/runtime/host/logger.go
kostko: Yes, it's already in its own commit, but I'll add a changelog entry and move to its own PR.

kostko:
> Doesn't this need to bump the runtime host protocol?

As this only adds methods, old runtimes should continue to work (which is desirable for an easy migration). It's just that newer runtimes would no longer work on old nodes.

kostko:
Ah yes, I'll bump the minor RHP version.

go/runtime/host/logger.go
kostko: Moved to #4924.

kostko:
Bumped.

Yawning (approved):

PR #4910, opened 2022-08-29
keymanager: Validate latest trust root height in key manager requests

peternose:
### Key manager requests should include latest consensus trust root
The key manager protocol should add the latest known trust root (or just the height) to (private key) EnclaveRPC requests. This would allow the key manager to detect whether the runtimes querying it have a fresh enough state (e.g. within 50 blocks) and reject operations if not.

Obviously the bad view can be either on the key manager side (Byzantine key manager operator) or the runtime side (Byzantine runtime operator). In case the key manager node is Byzantine the runtime can just switch to a different one (this already happens automatically when key manager requests fail).

This should be supported in a backwards compatible manner by making the new field optional.

### Test
Happy path is already covered with e2e tests. Unhappy path was tested locally.
- If key manager client is not up to date, `height` is not sent and after decoding takes a default value of `None`. Validation is skipped.
- If key manager client is up to date but `height` is lower than trust root's height, the client will receive an error, e.g. `private ephemeral key not available: call failed: height is not fresh`.
- If key manager is not up to date, the client will send `height` but it will not decode properly. The client will again receive an error, e.g. `private ephemeral key not available: call failed: unknown field`.

### DIFF
--- a/keymanager-lib/src/methods.rs
+++ b/keymanager-lib/src/methods.rs
@@ -9,12 +9,18 @@ use oasis_core_runtime::{
     enclave_rpc::Context as RpcContext,
 };
 
-/// Maximum age of ephemeral key in the number of epochs.
+/// Maximum age of an ephemeral key in the number of epochs.
 const MAX_EPHEMERAL_KEY_AGE: EpochTime = 10;
+/// Maximum age of a fresh height in the number of blocks.
+///
+/// A height is considered fresh if it is not more than specified amount
+/// of blocks lower than the height of the latest trust root.
+const MAX_FRESH_HEIGHT_AGE: u64 = 50;
 
 /// See `Kdf::get_or_create_keys`.
 pub fn get_or_create_keys(req: &LongTermKeyRequest, ctx: &mut RpcContext) -> Result<KeyPair> {
     authorize_private_key_generation(&req.runtime_id, ctx)?;
+    validate_height_freshness(req.height, ctx)?;
 
     Kdf::global().get_or_create_keys(req)
 }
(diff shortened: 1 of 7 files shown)


keymanager-lib/src/methods.rs
```diff
@@ -9,23 +9,30 @@ use oasis_core_runtime::{
...
 
 /// See `Kdf::get_public_key`.
 pub fn get_public_key(
     req: &LongTermKeyRequest,
-    _ctx: &mut RpcContext,
+    ctx: &mut RpcContext,
 ) -> Result<Option<SignedPublicKey>> {
     // No authentication or authorization.
     // Absolutely anyone is allowed to query public long-term keys.
+    validate_height_freshness(req.height, ctx)?;
```
kostko, line 35: Probably not needed for public key queries.

keymanager-lib/src/methods.rs
```diff
@@ -51,6 +59,7 @@ pub fn get_public_ephemeral_key(
     // No authentication or authorization.
     // Absolutely anyone is allowed to query public ephemeral keys.
     validate_epoch(req.epoch, ctx)?;
+    validate_height_freshness(req.height, ctx)?;
```
kostko, line 62: Probably not needed for public key queries.

kostko (commented):
Looks good, just remove the height checks from _public key_ queries as those have no authentication anyway.

keymanager-lib/src/methods.rs
peternose: I though it was needed, as even though public keys are available to anyone, the key manager would (sometimes) like to communicate only with runtimes that have a fresh enough state.

Took comment into account and removed validation in both public queries.

kostko (approved):

PR #4911, opened 2022-08-29
go: Add consensus transaction prioritization

abukosek:
This PR adds support for consensus transaction prioritization using the V1 tendermint mempool implementation.

Here are the base priorities for our consensus apps:
```
consensus/tendermint/apps/beacon/api.go:		AppPriority int64 = 100000
consensus/tendermint/apps/keymanager/api.go:		AppPriority int64 = 50000
consensus/tendermint/apps/registry/api.go:		AppPriority int64 = 50000
consensus/tendermint/apps/governance/api.go:		AppPriority int64 = 25000
consensus/tendermint/apps/roothash/api.go:		AppPriority int64 = 15000
consensus/tendermint/apps/staking/api.go:		AppPriority int64 = 1000
consensus/tendermint/apps/scheduler/api.go:		AppPriority int64 = 0
consensus/tendermint/apps/supplementarysanity/api.go:	AppPriority int64 = 0
```
The reason why the `scheduler` and `supplementarysanity` have a priority of 0 is that they have no transactions.
Node registrations have a slightly higher priority of 60000 instead of the base 50000 used for other registry transactions.

I also experimented a bit with bumping various priorities based on the transaction contents -- e.g. roothash executor commits and proposer timeouts based on runtime stake, but I don't think it's worth the extra complexity.
I suggest we first try the basic version as implemented in this PR and we can always add the extra priority fudging later, once we're sure this works as intended.

In case of errors, the returned priority is 0, which I think makes most sense, so people won't be able to submit faulty transactions to fill up the high-priority part of the queue.

### DIFF
--- a/go/consensus/tendermint/abci/mux.go
+++ b/go/consensus/tendermint/abci/mux.go
@@ -710,6 +710,7 @@ func (mux *abciMux) CheckTx(req types.RequestCheckTx) types.ResponseCheckTx {
 		Code:      types.CodeTypeOK,
 		GasWanted: int64(ctx.Gas().GasWanted()),
 		GasUsed:   int64(ctx.Gas().GasUsed()),
+		Priority:  ctx.GetPriority(),
 	}
 }
 
--- a/go/consensus/tendermint/api/context.go
+++ b/go/consensus/tendermint/api/context.go
@@ -72,6 +72,7 @@ type Context struct { // nolint: maligned
 	data          interface{}
 	events        []types.Event
 	gasAccountant GasAccountant
+	priority      int64
 
 	txSigner      signature.PublicKey
 	callerAddress staking.Address
@@ -108,6 +109,7 @@ func NewContext(
 		mode:          mode,
 		currentTime:   currentTime,
 		gasAccountant: gasAccountant,
+		priority:      0,
 		appState:      appState,
 		state:         state,
 		blockHeight:   blockHeight,
@@ -401,6 +403,18 @@ func (c *Context) InitialHeight() int64 {
 // BlockHeight returns the current block height.
 func (c *Context) BlockHeight() int64 {
 	return c.blockHeight
+}
+
+// SetPriority sets the current priority.
+// Higher number means higher priority.
+func (c *Context) SetPriority(p int64) {
+	c.priority = p
+}
+
+// GetPriority returns the current priority.
+// Higher number means higher priority.
+func (c *Context) GetPriority() int64 {
+	return c.priority
 }
 
 // BlockContext returns the current block context.
--- a/go/consensus/tendermint/apps/governance/governance.go
+++ b/go/consensus/tendermint/apps/governance/governance.go
@@ -65,6 +65,8 @@ func (app *governanceApplication) ExecuteTx(ctx *api.Context, tx *transaction.Tr
 		"tx", tx,
 	)
 
+	ctx.SetPriority(AppPriority)
+
 	switch tx.Method {
 	case governance.MethodSubmitProposal:
 		var proposalContent governance.ProposalContent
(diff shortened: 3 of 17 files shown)


kostko (commented):
Can we have some tests that these priorities work correctly?

It would probably be better if these priorities would be emitted similar to how gas stuff is done. E.g. there being some SetPriority in the context that one could query via GetPriority during CheckTx. Then priority could be configured as part of the regular execution flow instead of introducing another method.

go/consensus/tendermint/abci/mux.go
```diff
@@ -542,15 +542,15 @@ func (mux *abciMux) decodeTx(ctx *api.Context, rawTx []byte) (*transaction.Trans
 	return &tx, &sigTx, nil
 }
 
-func (mux *abciMux) processTx(ctx *api.Context, tx *transaction.Transaction, txSize int) error {
+func (mux *abciMux) processTx(ctx *api.Context, tx *transaction.Transaction, txSize int) (int64, error) {
```
kostko, line 545: I think it would be better if the caller extracted the priority from the context instead (similar to how it is for gas use).

go/consensus/tendermint/api/context.go
```diff
@@ -103,6 +105,7 @@ func NewContext(
 	blockHeight int64,
 	blockCtx *BlockContext,
 	initialHeight int64,
+	priority int64,
```
kostko, line 108: We could remove the constructor argument and assume if anyone wants to set priority then `SetPriority` would be called after construction.

go/consensus/tendermint/apps/governance/governance.go
```diff
@@ -71,12 +71,14 @@ func (app *governanceApplication) ExecuteTx(ctx *api.Context, tx *transaction.Tr
 		if err := cbor.Unmarshal(tx.Body, &proposalContent); err != nil {
 			return err
 		}
+		ctx.SetPriority(AppPriority)
```
kostko, line 74: Could just have a single set above given that this is the same for both cases? In case the transaction is later found to be invalid it will be discarded from the mempool anyway due to failing checks.

Also in other services/apps.

go/consensus/tendermint/full/full.go
```diff
@@ -570,6 +570,7 @@ func (t *fullService) lazyInit() error {
 	tenderConfig.Consensus.CreateEmptyBlocks = true
 	tenderConfig.Consensus.CreateEmptyBlocksInterval = emptyBlockInterval
 	tenderConfig.Consensus.DebugUnsafeReplayRecoverCorruptedWAL = viper.GetBool(CfgDebugUnsafeReplayRecoverCorruptedWAL) && cmflags.DebugDontBlameOasis()
+	tenderConfig.Mempool.Version = tmconfig.MempoolV1
```
kostko, line 573: What is the default?

abukosek: The default is V0: https://github.com/oasisprotocol/tendermint/blob/v0.34.21-oasis/config/config.go#L738

go/consensus/tendermint/abci/mux.go
abukosek: Oh, right, I forgot to change this when I moved priority handling into the context :)

go/consensus/tendermint/api/context.go
```diff
@@ -82,6 +82,8 @@ type Context struct { // nolint: maligned
 	blockCtx      *BlockContext
 	initialHeight int64
 
+	priority int64
```
kostko, line 85: Could be moved up into the gorup where the data, events and gasAccountant are.

kostko (approved):

PR #4916, opened 2022-08-30
registry: Add ProveFreshness consensus layer transaction

peternose:
### Add ProveFreshness consensus layer transaction

This is needed for `Client node TEE freshness verification` and should be implemented in the following manner.

We can add a (carefully parameter-gated) new transaction (e.g. `registry.ProveFreshness`) that accepts a fixed-size binary blob of 32 bytes and always succeeds without doing any processing or state changes. It should cost a reasonable amount of gas.

The enablement parameter should be part of the existing `tee_features` consensus parameter group and be a boolean called `freshness_proofs`. If it is set to true the new transaction is processed otherwise it is handled as a non-existing transaction (e.g. exactly as currently to avoid breaking consensus before the activation proposal passes).

The existing `consensus-tee-pcs` migration should be updated to enable the `freshness_proofs` feature and configure the default gas cost.

### Test
Tested locally (sub-command `gen_prove_freshness` will be added latter to the CLI repository).
```
oasis-node registry node gen_prove_freshness \
  --transaction.fee.gas 1000 \
  --transaction.fee.amount 0 \
  --transaction.file=tx.txt \
  --genesis.file ../genesis.json \
  --debug.dont_blame_oasis \
  --debug.allow_test_keys \
  --log.level=DEBUG \
  --assume_yes

oasis-node consensus submit_tx --transaction.file tx.txt
```
If proofs are not enabled, I get the following error.
```
level=error ts=2022-08-31T06:55:44.773530442Z caller=grpc.go:242 module=grpc/client msg="request failed" method=/oasis-core.Consensus/SubmitTx req_seq=1 rsp=null err="registry: invalid argument"
level=error ts=2022-08-31T06:55:44.773585673Z caller=consensus.go:129 module=cmd/consensus msg="failed to submit transaction" err="registry: invalid argument"
```

### DIFF
--- a/go/consensus/tendermint/apps/registry/registry.go
+++ b/go/consensus/tendermint/apps/registry/registry.go
@@ -94,24 +94,25 @@ func (app *registryApplication) ExecuteTx(ctx *api.Context, tx *transaction.Tran
 		if err := cbor.Unmarshal(tx.Body, &sigEnt); err != nil {
 			return err
 		}
-
 		return app.registerEntity(ctx, state, &sigEnt)
+
 	case registry.MethodDeregisterEntity:
 		return app.deregisterEntity(ctx, state)
+
 	case registry.MethodRegisterNode:
 		var sigNode node.MultiSignedNode
 		if err := cbor.Unmarshal(tx.Body, &sigNode); err != nil {
 			return err
 		}
-
 		return app.registerNode(ctx, state, &sigNode)
+
 	case registry.MethodUnfreezeNode:
 		var unfreeze registry.UnfreezeNode
 		if err := cbor.Unmarshal(tx.Body, &unfreeze); err != nil {
 			return err
 		}
-
 		return app.unfreezeNode(ctx, state, &unfreeze)
+
 	case registry.MethodRegisterRuntime:
 		var rt registry.Runtime
 		if err := cbor.Unmarshal(tx.Body, &rt); err != nil {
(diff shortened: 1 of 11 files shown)


go/consensus/tendermint/apps/registry/registry.go
```diff
@@ -121,6 +122,17 @@ func (app *registryApplication) ExecuteTx(ctx *api.Context, tx *transaction.Tran
 			return err
 		}
 		return nil
+
+	case registry.MethodProveFreshness:
+		var blob [32]byte
+		if err := cbor.Unmarshal(tx.Body, &blob); err != nil {
+			return err
```
kostko, line 129: To avoid this being consensus breaking, this must return `registry.ErrInvalidArgument` in case freshness proofs are not supported. Otherwise the resulting error module/code may be different, causing a consensus failure.

Looking back all of these above should actually return `registry.ErrInvalidArgument` during deserialization failures, but we can't change this now as it would be consensus-breaking.

peternose: Agree with both statements.

go/oasis-node/cmd/registry/node/node.go
```diff
@@ -333,6 +344,60 @@ func doIsRegistered(cmd *cobra.Command, args []string) {
...
+		if node.ID.Equal(nodeIdentity.NodeSigner.Public()) {
+			fmt.Println("node is registered")
+			os.Exit(0)
+		}
+	}
+	fmt.Println("node is not registered")
+	os.Exit(1)
+}
+
+func doGenProveFreshness(cmd *cobra.Command, args []string) {
```
kostko, line 370: Oh, we're trying to deprecate these, let's try not to add any new commands here. Instead those should belong to the CLI (will be moved into its own repo, currently here: https://github.com/oasisprotocol/oasis-sdk/tree/main/cli).

kostko (commented):
Mostly looks good, I would just avoid adding the CLI subcommands to oasis-node.

go/oasis-node/cmd/registry/node/node.go
peternose: Right, I forgot. Will remove cmd and  add it latter to the CLI repository.

kostko (approved):

PR #4922, opened 2022-09-05
runtime: Add client node TEE freshness verification

peternose:
### Add client node TEE freshness verification
Executor and key manager TEE nodes currently use node registration (which includes the randomly generated Runtime Attestation Key and nonce) in order to establish consensus layer view freshness.

But this approach doesn’t work for client nodes. Since client nodes do not currently register, there is no way for TEE runtimes running on them to verify consensus layer freshness. There is however a need for client nodes to be able to execute signed queries and requiring an up-to-date consensus layer view is crucial for verifying quotes and policies.

To do this the runtime just needs to generate a transaction containing a nonce and that transaction must be published in a block, with the host node providing the height at which the transaction was published, its index and Merkle proof of inclusion. The runtime can then use the light client to verify the specified block header and the provided Merkle proof using the verified transaction root hash.

The newly added `registry.ProveFreshness` transaction that accepts a fixed-size binary blob of 32 bytes and always succeeds without doing any processing should be used for this purpose. The runtime host protocol should be extended to support this new flow, initiated by the consensus verifier in the runtime (immediately after initializing the light client in case enabled in the static runtime config), e.g. adding the `HostProveFreshness{Request,Response}` messages.

Note that this check can be used very early on as it doesn’t require the node to be registered.

### Test
Unhappy path tested locally. Happy path already included in trust root tests.

### DIFF
--- a/go/consensus/api/submission.go
+++ b/go/consensus/api/submission.go
@@ -64,6 +64,13 @@ type SubmissionManager interface {
 	//
 	// It also automatically handles retries in case the nonce was incorrectly estimated.
 	SignAndSubmitTx(ctx context.Context, signer signature.Signer, tx *transaction.Transaction) error
+
+	// SignAndSubmitTxWithProof populates the nonce and fee fields in the transaction, signs
+	// the transaction with the passed signer, submits it to consensus backend and creates
+	// a proof of inclusion.
+	//
+	// It also automatically handles retries in case the nonce was incorrectly estimated.
+	SignAndSubmitTxWithProof(ctx context.Context, signer signature.Signer, tx *transaction.Transaction) (*transaction.SignedTransaction, *transaction.Proof, error)
 }
 
 type submissionManager struct {
@@ -124,7 +131,7 @@ func (m *submissionManager) EstimateGasAndSetFee(ctx context.Context, signer sig
 	return nil
 }
 
-func (m *submissionManager) signAndSubmitTx(ctx context.Context, signer signature.Signer, tx *transaction.Transaction) error {
+func (m *submissionManager) signAndSubmitTx(ctx context.Context, signer signature.Signer, tx *transaction.Transaction, withProof bool) (*transaction.SignedTransaction, *transaction.Proof, error) {
 	// Update transaction nonce.
 	var err error
 	signerAddr := staking.NewAddress(signer.Public())
@@ -134,14 +141,14 @@ func (m *submissionManager) signAndSubmitTx(ctx context.Context, signer signatur
 		if errors.Is(err, ErrNoCommittedBlocks) {
 			// No committed blocks available, retry submission.
 			m.logger.Debug("retrying transaction submission due to no committed blocks")
-			return err
+			return nil, nil, err
 		}
-		return backoff.Permanent(err)
+		return nil, nil, backoff.Permanent(err)
 	}
 
 	// Estimate the fee.
 	if err = m.EstimateGasAndSetFee(ctx, signer, tx); err != nil {
-		return fmt.Errorf("failed to estimate fee: %w", err)
+		return nil, nil, fmt.Errorf("failed to estimate fee: %w", err)
 	}
 
 	// Sign the transaction.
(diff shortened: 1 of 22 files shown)


runtime/src/consensus/tendermint/merkle.rs
```diff
@@ -0,0 +1,173 @@
+//! Merkle proofs used in Tendermint networks
+//!
+//! Rewritten to Rust from:
+//! https://github.com/tendermint/tendermint/blob/main/crypto/merkle/proof.go
```
kostko, line 4: So there is no Rust implementation of this scheme yet?

go/runtime/registry/handlers.go
```diff
@@ -0,0 +1,189 @@
...
+		raw = append(raw, tx.Raw())
+	}
+
+	return &protocol.HostFetchTxBatchResponse{
+		Batch: raw,
+	}, nil
+}
+
+func handlerHostProveFreshnessRequest(ctx context.Context, h *runtimeHostHandler, rq *protocol.HostProveFreshnessRequest) (*protocol.HostProveFreshnessResponse, error) {
+	// Subscribe before submit as we don't want to miss the block with the transaction.
```
kostko, line 125: This could also be part of the consensus backend interface, e.g. something like `SubmitTxWithProof`. Especially given that the `SubmitTx` method already subscribes to and waits for results:
https://github.com/oasisprotocol/oasis-core/blob/b64e8b5a08c27a31066b47e232b09087b8c51f70/go/consensus/tendermint/full/full.go#L247

This would also make it easier to abstract out the proof type (e.g. use `[]byte` as mentioned earlier).

runtime/src/consensus/tendermint/merkle.rs
peternose: It seems not. Was searching without any success.

runtime/src/consensus/tendermint/verifier.rs
```diff
@@ -795,6 +895,13 @@ impl Verifier {
             "latest_height" => cache.latest_known_height(),
         );
 
+        // Verify state freshness using the host. This step is required only for clients
+        // as executors and key managers verify freshness regularly using node registration
+        // (RAK with random nonces).
+        if self.protocol.get_rak().is_none() {
```
kostko, line 901: This won't work as the runtime will generate a RAK for Remote Attestation purposes in any case. It would be better if there was a configuration option that would enable this feature, e.g. in:
https://github.com/oasisprotocol/oasis-core/blob/b64e8b5a08c27a31066b47e232b09087b8c51f70/runtime/src/config.rs#L4-L18

runtime/src/consensus/transaction.rs
```diff
@@ -0,0 +1,37 @@
...
+    // Nonce is a nonce to prevent replay.
+    pub nonce: u64,
+    // Fee is an optional fee that the sender commits to pay to execute this
+    // transaction.
+    pub fee: Fee,
+
+    // Method is the method that should be called.
+    pub method: MethodName,
+    // Body is the method call body.
+    pub body: Vec<u8>,
```
kostko, line 17: The (more generally) correct type is `cbor::Value` not `Vec<u8>`. The `Vec<u8>` only works for this specific `registry.ProveFreshness` transaction as it takes a binary blob as body.

go/consensus/api/submission.go
```diff
@@ -63,7 +63,14 @@ type SubmissionManager interface {
 	// with the passed signer and submits it to consensus backend.
 	//
 	// It also automatically handles retries in case the nonce was incorrectly estimated.
-	SignAndSubmitTx(ctx context.Context, signer signature.Signer, tx *transaction.Transaction) error
+	SignAndSubmitTx(ctx context.Context, signer signature.Signer, tx *transaction.Transaction) (*transaction.SignedTransaction, error)
```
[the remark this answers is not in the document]
peternose: No, currently there is no need. I just wanted to be consistent with `SignAndSubmitTxWithProof`. It is actually interesting that we never need the signed transaction, neither the height at which it was published.
I can remove it if you like.

go/runtime/host/sandbox/sandbox.go
```diff
@@ -523,7 +517,13 @@ func (r *sandboxedRuntime) manager() {
 
 			// Notify subscribers that the runtime has stopped.
 			r.notifier.Broadcast(&host.Event{Stopped: &host.StoppedEvent{}})
-			continue
+		case <-time.After(time.Hour):
```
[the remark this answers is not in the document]
peternose: Agree.

go/consensus/api/submission.go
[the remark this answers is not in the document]
kostko: I would remove it.

kostko:
Rebase on master now that #4904 is merged.

go/consensus/api/submission.go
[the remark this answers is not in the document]
peternose: Removed and rebased.

runtime/src/consensus/tendermint/merkle.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /// Use the leaf_hash and inner_hashes to get the root merkle hash.
+    /// If the length of the inner_hashes slice isn't exactly correct, the result is None.
+    /// Recursive impl.
+    fn compute_hash_from_aunts(
+        index: i64,
+        total: i64,
+        leaf_hash: Hash,
+        inner_hashes: &Vec<Hash>,
```
kostko, line 88: Could this be changed to `&[Hash]` to avoid conversions to vec?

kostko (approved):

kostko (approved):

runtime/src/consensus/tendermint/merkle.rs
peternose: Yes, much better.

runtime/src/consensus/tendermint/merkle.rs
```diff
@@ -0,0 +1,173 @@
...
+    ) -> Option<Hash> {
+        if index >= total || index < 0 || total <= 0 {
+            return None;
+        }
+        match total {
+            0 => panic!("cannot call compute_hash_from_aunts() with 0 total"),
```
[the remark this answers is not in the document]
peternose: Agree, unreachable is better. However, I didn't want to remove this as I wanted to be consistent with GO version.

runtime/src/consensus/tendermint/merkle.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// Verify that the Proof proves the root hash.
+    /// Check index/total manually if needed.
+    pub fn verify(&self, root_hash: Hash, leaf: Hash) -> Result<()> {
+        if self.total < 0 {
+            return Err(anyhow!("proof total must be positive"));
+        }
+        if self.index < 0 {
+            return Err(anyhow!("proof index cannot be negative"));
+        }
+
```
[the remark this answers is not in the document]
peternose: Yes, but only if you call `ValidateBasic`. Will add this check, although I don't want to deviate from the GO implementation to much.

PR #4925, opened 2022-09-08
keymanager: Verify that policy was published in the consensus layer

peternose:
### Force on-chain publication of key manager policies
When the key manager policy is updated, the runtime should check whether this policy has actually been published in the consensus layer by using the light client verifier.

This will require adding the Rust implementation of the consensus state accessor for the key manager service. Then one can fetch the corresponding key manager Status and check that the correct policy has been published there.

### Test
Happy path is already tested in `trust-root/simple` scenario. Unhappy path tested locally, the key manager doesn't update policy and constantly throws the following error:
```{"caller":"worker.go:737","err":"worker/keymanager: failed to extract rpc response payload: rpc failure: 'policy hasn't been published'","level":"error","module":"worker/keymanager","msg":"failed to handle status update","ts":"2022-09-08T22:56:44.214990354Z"}```

### DIFF
--- a/keymanager-api-common/src/api.rs
+++ b/keymanager-api-common/src/api.rs
@@ -1,7 +1,4 @@
-use std::{
-    collections::{HashMap, HashSet},
-    default::Default,
-};
+use std::{collections::HashSet, default::Default, vec};
 
 use rand::{rngs::OsRng, Rng};
 use thiserror::Error;
@@ -10,11 +7,13 @@ use zeroize::Zeroize;
 
 use oasis_core_runtime::{
     common::{
-        crypto::signature::{PublicKey as OasisPublicKey, Signature, SignatureBundle},
+        crypto::signature::{PublicKey as OasisPublicKey, Signature},
         namespace::Namespace,
-        sgx::EnclaveIdentity,
     },
-    consensus::beacon::EpochTime,
+    consensus::{
+        beacon::EpochTime,
+        keymanager::{PolicySGX, SignedPolicySGX},
+    },
     impl_bytes,
 };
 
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: keymanager-lib/src/policy.rs [hunk 1]
```diff
@@ -64,23 +68,37 @@ impl Policy {
...
+        // Ensure the new policy was published in the consensus layer.
+        let state = ctx.consensus_verifier.latest_state()?;
+        let km_state = KeyManagerState::new(&state);
+        let published_policy = km_state
+            .status(Context::create_child(&ctx.io_ctx), new_policy.runtime_id)?
+            .ok_or(KeyManagerError::PolicyNotPublished)?
+            .policy
+            .ok_or(KeyManagerError::PolicyNotPublished)?;
+        let untrusted_policy: SignedPolicySGX = cbor::from_slice(raw_policy)?;
+        if !untrusted_policy.eq(&published_policy) {
```
kostko, line 89:
```suggestion
        if untrusted_policy != published_policy {
```

keymanager-lib/src/policy.rs
```diff
@@ -200,11 +218,11 @@ impl CachedPolicy {
     fn parse(raw: &[u8]) -> Result<Self> {
         // Parse out the signed policy.
         let untrusted_policy: SignedPolicySGX = cbor::from_slice(raw)?;
-        let policy = untrusted_policy.verify()?;
+        verify_and_trust_policy(&untrusted_policy)?;
```
kostko, line 221: After verification the policy is no longer untrusted so the previous setup seemed nicer. You could do `let policy = untrusted_policy.policy;`.

peternose: Had that but then I changed my mind, not sure why :smile:

keymanager-api-common/src/api.rs
```diff
@@ -305,6 +281,28 @@ impl Default for TrustedPolicySigners {
...
+    pub fn verify_and_trust(&self, policy: &SignedPolicySGX) -> Result<(), KeyManagerError> {
+        policy
+            .verify()
+            .map_err(|err| KeyManagerError::PolicyInvalid(err.into()))?;
+
+        self.trust(policy)
+    }
+
+    /// Verify that policy has enough signatures from trusted signers.
+    fn trust(&self, policy: &SignedPolicySGX) -> Result<(), KeyManagerError> {
```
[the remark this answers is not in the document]
kostko: This makes sense.

Maybe just rename to something else as `trust` makes it sound like this will make the policy trusted instead of verifying the threshold.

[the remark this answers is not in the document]
peternose: Renamed to `verify_trusted_signers`.

keymanager-api-common/src/lib.rs
```diff
@@ -32,35 +29,8 @@ pub fn set_trusted_policy_signers(signers: TrustedPolicySigners) -> bool {
...
-        let multisig_threshold = trusted_signers.threshold;
-        if signers.len() < multisig_threshold as usize {
-            return Err(KeyManagerError::PolicyInsufficientSignatures);
-        }
-
-        Ok(self.policy.clone())
-    }
+/// Verify that policy has valid signatures and that enough of them are from the global set
+/// of trusted policy signers.
+pub fn verify_policy_and_trusted_signers(policy: &SignedPolicySGX) -> Result<(), KeyManagerError> {
```
kostko, line 34: This (and maybe even the `verify` method) could even return the `policy.policy` so upstream callers can do something like `let policy = verify_policy_and_trusted_signers(untrusted_policy)?;` which seems to be common now.

kostko (approved):

keymanager-api-common/src/lib.rs
peternose: I applied your comment. Policy is now returned (just as it was before this PR).

I intentionally wanted to removed this for 4 reasons:
- The `verify` method had two responsibilities, as can be seen from the comment `/// Verify the signatures and return the PolicySGX`.
- I expect that `verify` returns `true/false` or `()/Err`, and not a policy. However, I expect `sign` to return a signed policy.
- Policy had to be cloned, to achieve that `Ok(self.policy.clone())`. Unnecessary step.
- Passing policy back and forth just doesn't feel right as it can always be accessed via `SignedPolicySGX`. It would be nice if this would give us some kind of protection, but it doesn't. For example, if policy was private and could be accessed only via getter function `verify`, then everyone would be forced to call `verify`.

The only reason for is to have more readable code `let policy = verify_policy...`.

kostko: There should be no need to clone a policy, you should be able to do something like:
```rust
pub fn verify<'a>(policy: &'a SignedPolicySGX) -> Result<&'a PolicySGX, Error> {
```
and then return a reference, e.g. `&policy.policy`.

> For example, if policy was private and could be accessed only via getter function verify, then everyone would be forced to call verify.

Yeah that would be ideal but not sure how it would work with deserialization.

kostko (approved):

PR #4930, opened 2022-09-12
keymanager: Refactor and reorganize runtime-related crates

peternose:
Currently there is a bunch of runtime-related crates, where the split is mostly for legacy reasons. We should consider how to consolidate these crates so that it makes them easier to maintain. The current crates are as follows:
- `oasis-core-client` (`client/`), could be moved into `oasis-core-runtime`.
- `oasis-core-keymanager-*` (`keymanager-*/`), some (APIs?) could be moved into oasis-core-runtime, some (KM implementation) could be consolidated under a single crate.


keymanager/src/api/requests.rs
```diff
@@ -0,0 +1,115 @@
+use std::default::Default;
```
kostko, line 1: This should be imported as part of the default prelude and so an explicit import should not be needed?

keymanager/src/crypto/types.rs
```diff
@@ -0,0 +1,110 @@
+use std::{default::Default, vec};
```
kostko, line 1: As above, neither should be needed due to std prelude being imported by default.

keymanager/src/policy/signers.rs
```diff
@@ -0,0 +1,145 @@
+use std::{collections::HashSet, default::Default};
```
kostko, line 1: No need to import `Default`.

kostko (approved):
Nice work!

PR #4932, opened 2022-09-12
go/common/dynlib: Handle Devuan in tests

Yawning:
Devuan puts libc in a somewhat odd place, and doesn't have enough backward compatibility symlinks, so the fallback test fails.

Resolve this by adding a special case for where Devuan sticks shared libraries.  Note that most sensible systems will only ever exercise the fallback path when running unit tests.

### DIFF
--- a/go/common/dynlib/cache.go
+++ b/go/common/dynlib/cache.go
@@ -537,6 +537,7 @@ func archDepsMusl() (elf.Machine, []string, error) {
 			"/lib64",
 			"/usr/lib64",
 			"/usr/lib/x86_64-linux-gnu",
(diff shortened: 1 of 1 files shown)


pro-wh (approved):
https://packages.debian.org/bullseye/amd64/libc6/filelist that's the normal place for it in debian too

Yawning:
> https://packages.debian.org/bullseye/amd64/libc6/filelist that's the normal place for it in debian too

It's a debian/ubuntu-ism (https://wiki.ubuntu.com/MultiarchSpec) who's rationale IMO is no longer relevant because running 32-bit code on a 64-bit architecture is largely not done anymore.

PR #4934, opened 2022-09-13
runtime/consensus/tendermint/verifier: Refactor verifier

peternose:
Refactor verifier to group verify/update time/update block operations
Also consider splitting the tendermint/verifier into multiple modules as it is getting quite large.


kostko (commented):
Seems strange that the store (LRU light block store) was left out of the verifier module as it is only used by the verifier. It should probably be moved there. But there is now also a different kind of store (the trust root store) in the `verifier::store` module.

kostko (approved):

PR #4938, opened 2022-09-20
go/governance: Add change parameters proposal

peternose:
We should add a governance proposal that allows changing parameters. The parameter change would be described by a CBOR document containing the following:
- Name of the consensus app/service that the parameter change applies to.
- Parameter change that can be applied to the app's `ConsensusParameters` structure.

To execute a parameter change proposal, the governance module would submit a message (using the existing message dispatch mechanism) to the given application containing the details of the proposal. Executing the parameter change would be up to the application and each application should carefully scope what is allowed to be changed.

This functionality should be parameter-gated behind a governance consensus parameter to be enabled in an upgrade handler.

### DIFF
--- a/go/beacon/api/api.go
+++ b/go/beacon/api/api.go
@@ -3,11 +3,9 @@ package api
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/oasisprotocol/oasis-core/go/common/errors"
 	"github.com/oasisprotocol/oasis-core/go/common/pubsub"
-	"github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common/flags"
 )
 
 const (
@@ -131,7 +129,7 @@ type ConsensusParameters struct {
 	// InsecureParameters are the beacon parameters for the insecure backend.
 	InsecureParameters *InsecureParameters `json:"insecure_parameters,omitempty"`
 
-	// VRFParamenters are the beacon parameters for the VRF backend.
+	// VRFParameters are the beacon parameters for the VRF backend.
 	VRFParameters *VRFParameters `json:"vrf_parameters,omitempty"`
 }
 
@@ -139,52 +137,6 @@ type ConsensusParameters struct {
 type InsecureParameters struct {
 	// Interval is the epoch interval (in blocks).
 	Interval int64 `json:"interval,omitempty"`
-}
-
-// SanityCheck does basic sanity checking on the genesis state.
-func (g *Genesis) SanityCheck() error {
-	switch g.Parameters.Backend {
-	case BackendInsecure:
-		params := g.Parameters.InsecureParameters
-		if params == nil {
-			return fmt.Errorf("beacon: sanity check failed: insecure backend not configured")
-		}
-
-		if params.Interval <= 0 && !g.Parameters.DebugMockBackend {
-			return fmt.Errorf("beacon: sanity check failed: epoch interval must be > 0")
-		}
-	case BackendVRF:
-		params := g.Parameters.VRFParameters
-		if params == nil {
-			return fmt.Errorf("beacon: sanity check failed: VRF backend not configured")
-		}
-
-		if params.AlphaHighQualityThreshold == 0 {
-			return fmt.Errorf("beacon: sanity check failed: alpha threshold must be > 0")
-		}
-		if params.Interval <= 0 {
-			return fmt.Errorf("beacon: sanity check failed: epoch interval must be > 0")
-		}
-		if params.ProofSubmissionDelay <= 0 {
-			return fmt.Errorf("beacon: sanity check failed: submission delay must be > 0")
-		}
-		if params.ProofSubmissionDelay >= params.Interval {
-			return fmt.Errorf("beacon: sanity check failed: submission delay must be < epoch interval")
-		}
-	default:
-		return fmt.Errorf("beacon: sanity check failed: unknown backend: '%s'", g.Parameters.Backend)
-	}
-
-	unsafeFlags := g.Parameters.DebugMockBackend
-	if unsafeFlags && !flags.DebugDontBlameOasis() {
-		return fmt.Errorf("beacon: sanity check failed: one or more unsafe debug flags set")
-	}
-
-	if g.Base == EpochInvalid {
-		return fmt.Errorf("beacon: sanity check failed: starting epoch is invalid")
-	}
-
-	return nil
 }
 
 // EpochEvent is the epoch event.
--- a/go/beacon/api/sanity_check.go
+++ b/go/beacon/api/sanity_check.go
@@ -1,0 +1,62 @@
+package api
+
+import (
+	"fmt"
+
+	"github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common/flags"
+)
+
+// SanityCheck does basic sanity checking on the genesis state.
+func (g *Genesis) SanityCheck() error {
+	if err := g.Parameters.SanityCheck(); err != nil {
+		return fmt.Errorf("beacon: sanity check failed: %w", err)
+	}
+
+	if g.Base == EpochInvalid {
+		return fmt.Errorf("beacon: sanity check failed: starting epoch is invalid")
+	}
+
+	return nil
+}
+
+// SanityCheck performs a sanity check on the consensus parameters.
+func (p *ConsensusParameters) SanityCheck() error {
+	switch p.Backend {
+	case BackendInsecure:
+		params := p.InsecureParameters
+		if params == nil {
+			return fmt.Errorf("insecure backend not configured")
+		}
+
+		if params.Interval <= 0 && !p.DebugMockBackend {
+			return fmt.Errorf("epoch interval must be > 0")
+		}
+	case BackendVRF:
+		params := p.VRFParameters
+		if params == nil {
+			return fmt.Errorf("VRF backend not configured")
+		}
+
+		if params.AlphaHighQualityThreshold == 0 {
+			return fmt.Errorf("alpha threshold must be > 0")
+		}
+		if params.Interval <= 0 {
+			return fmt.Errorf("epoch interval must be > 0")
+		}
+		if params.ProofSubmissionDelay <= 0 {
+			return fmt.Errorf("submission delay must be > 0")
+		}
+		if params.ProofSubmissionDelay >= params.Interval {
+			return fmt.Errorf("submission delay must be < epoch interval")
+		}
+	default:
+		return fmt.Errorf("unknown backend: '%s'", p.Backend)
+	}
+
+	unsafeFlags := p.DebugMockBackend
+	if unsafeFlags && !flags.DebugDontBlameOasis() {
+		return fmt.Errorf("one or more unsafe debug flags set")
+	}
+
+	return nil
+}
--- a/go/consensus/tendermint/apps/governance/governance.go
+++ b/go/consensus/tendermint/apps/governance/governance.go
@@ -10,6 +10,7 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common/quantity"
 	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"
 	"github.com/oasisprotocol/oasis-core/go/consensus/tendermint/api"
+	governanceApi "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/apps/governance/api"
 	governanceState "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/apps/governance/state"
 	registryapp "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/apps/registry"
 	registryState "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/apps/registry/state"
@@ -26,6 +27,7 @@ var _ api.Application = (*governanceApplication)(nil)
 
 type governanceApplication struct {
 	state api.ApplicationState
+	md    api.MessageDispatcher
 }
 
 func (app *governanceApplication) Name() string {
@@ -50,9 +52,12 @@ func (app *governanceApplication) Dependencies() []string {
 
 func (app *governanceApplication) OnRegister(state api.ApplicationState, md api.MessageDispatcher) {
 	app.state = state
+	app.md = md
 
 	// Subscribe to messages emitted by other apps.
 	md.Subscribe(api.MessageStateSyncCompleted, app)
+	md.Subscribe(governanceApi.MessageChangeParameters, app)
+	md.Subscribe(governanceApi.MessageValidateParameterChanges, app)
 }
 
 func (app *governanceApplication) OnCleanup() {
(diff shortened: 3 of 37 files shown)


go/consensus/tendermint/apps/governance/governance.go
```diff
@@ -263,6 +313,28 @@ func (app *governanceApplication) executeProposal(ctx *api.Context, state *gover
...
+		// disabled.
+		params, err := state.ConsensusParameters(ctx)
+		if err != nil {
+			ctx.Logger().Error("failed to query consensus parameters",
+				"err", err,
+			)
+			return governance.ErrInvalidArgument
+		}
+		if !params.EnableChangeParametersProposal {
+			ctx.Logger().Error("change parameters proposals are disabled")
```
kostko, line 327:
```suggestion
			ctx.Logger().Debug("change parameters proposals are disabled")
```

go/consensus/tendermint/apps/governance/governance.go
```diff
@@ -263,6 +313,28 @@ func (app *governanceApplication) executeProposal(ctx *api.Context, state *gover
...
+			return governance.ErrInvalidArgument
+		}
+
+		// Notify other interested applications about the change parameters proposal.
+		if _, err = app.md.Publish(ctx, governanceApi.MessageChangeParameters, proposal.Content.ChangeParameters); err != nil {
+			ctx.Logger().Error("failed to dispatch change parameters proposal message",
```
kostko, line 333:
```suggestion
			ctx.Logger().Debug("failed to dispatch change parameters proposal message",
```

go/consensus/tendermint/apps/governance/transactions.go
```diff
@@ -124,10 +125,32 @@ func (app *governanceApplication) submitProposal(
...
+		// Ensure upgrade descriptor is far enough in future so that cancellation is still allowed.
 		if upgrade.Descriptor.Epoch < params.UpgradeCancelMinEpochDiff+epoch {
 			return governance.ErrUpgradeTooSoon
 		}
+
+	case proposalContent.ChangeParameters != nil:
+		// To not violate the consensus, change parameters proposals should be ignored when
+		// disabled.
+		if !params.EnableChangeParametersProposal {
+			break
```
kostko, line 137: This is breaking, because it seems that the presence of the additional "change parameters" proposal is handled in the above `proposalContent.ValidateBasic()` and that will actually pass regardless of the value of `params.EnableChangeParametersProposal`. There needs to be a check earlier above (e.g. immediately after governance consensus parameters are fetched) to avoid problems.

Also this switch statement lacks a default branch which seems dangerous. Even though all validation is done earlier, I would add an explicit default that returns "invalid argument".

go/consensus/tendermint/apps/governance/transactions.go
```diff
@@ -124,10 +125,32 @@ func (app *governanceApplication) submitProposal(
...
+		}
+
+		// Notify other interested applications to validate the parameter changes.
+		var res interface{}
+		res, err = app.md.Publish(ctx, governanceApi.MessageValidateParameterChanges, proposalContent.ChangeParameters)
+		if err != nil {
+			ctx.Logger().Error("governance: failed to dispatch validate parameter changes message",
```
kostko, line 144:
```suggestion
			ctx.Logger().Debug("governance: failed to dispatch validate parameter changes message",
```

go/consensus/tendermint/apps/governance/transactions.go
```diff
@@ -124,10 +125,32 @@ func (app *governanceApplication) submitProposal(
...
+				"err", err,
+			)
+			return err
+		}
+		// At least one module should be interested in the proposed changes. If no one is,
+		// the proposal is rejected as not being supported.
+		if res == nil {
+			return fmt.Errorf("governance: change parameters proposal's changes not supported")
```
kostko, line 152: This should return `governance.ErrInvalidArgument`.

go/oasis-node/cmd/debug/fixgenesis/fixgenesis.go
```diff
@@ -341,6 +341,9 @@ type v4GovernanceConsensusParameters struct {
 	// UpgradeCancelMinEpochDiff is the minimum number of epochs between the current
 	// epoch and the proposed upgrade epoch for the upgrade cancellation proposal to be valid.
 	UpgradeCancelMinEpochDiff beacon.EpochTime `json:"upgrade_cancel_min_epoch_diff,omitempty"`
+
+	// EnableChangeParametersProposal is true iff change parameters proposals are allowed.
+	EnableChangeParametersProposal bool `json:"enable_change_parameters_proposal,omitempty"`
```
kostko, line 346: There should be no need to modify fixgenesis as `v4` refers to consensus layer v4 which did not have this field.

go/oasis-test-runner/scenario/e2e/runtime/governance_change_parameters.go
```diff
@@ -0,0 +1,136 @@
...
+		return nil, err
+	}
+
+	f.Network.DefaultLogWatcherHandlerFactories = append(
+		f.Network.DefaultLogWatcherHandlerFactories,
+		log.AssertJSONContains(
+			"msg",
+			"change parameters proposals are disabled",
+			"change parameters proposals should be disabled",
+		),
```
kostko, line 43: If this needs to be verified I would rather add an explicit log event instead of relying on messages having a certain format (e.g. the same as is done for other similar log events).

go/oasis-test-runner/scenario/e2e/runtime/governance_change_parameters.go
```diff
@@ -0,0 +1,136 @@
...
+	consensus "github.com/oasisprotocol/oasis-core/go/consensus/api"
+	governance "github.com/oasisprotocol/oasis-core/go/governance/api"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/env"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/log"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario/e2e"
+	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
+)
+
```
kostko, line 16: There is probably no need to spin up runtimes when testing these parameter change proposals as they don't really affect them (different to governance proposal for network upgrades)? Actually could we even test these during the "node unit test" given that it is really simple? I want to avoid adding too many e2e tests as each one takes a lot of time to set up the network, especially the runtime ones.

go/scheduler/api/api.go
```diff
@@ -288,6 +286,38 @@ type ConsensusParameters struct {
...
+// ConsensusParameterChanges are allowed scheduler consensus parameter changes.
+type ConsensusParameterChanges struct {
+	// MinValidators is the new minimum number of validators.
+	MinValidators *int `json:"min_validators"`
+
+	// MaxValidators is the new maximum number of validators.
+	MaxValidators *int `json:"max_validators"`
+
+	// MaxValidatorsPerEntity is the new maximum number of validators per entity.
+	MaxValidatorsPerEntity *int `json:"max_validators_per_entity"`
```
kostko, line 298: We should not allow changing this.

go/scheduler/api/api.go
```diff
@@ -288,6 +286,38 @@ type ConsensusParameters struct {
...
+
+	// RewardFactorEpochElectionAny is the new factor for a reward.
+	RewardFactorEpochElectionAny *quantity.Quantity `json:"reward_factor_epoch_election_any"`
```
kostko, line 301: We should not allow changing this.

go/oasis-node/cmd/debug/txsource/workload/governance.go
```diff
@@ -204,7 +204,7 @@ func (g *governanceWorkload) submitCancelUpgradeProposal(descriptor *upgrade.Des
 	}
 	var proposal *governance.Proposal
 	for _, p := range ps {
```
kostko, line 206: Could we update this txsource governance workload to also submit some fuzzed new proposal types?

go/beacon/api/api.go
```diff
@@ -131,62 +129,42 @@ type ConsensusParameters struct {
...
 }
 
-// InsecureParameters are the beacon parameters for the insecure backend.
-type InsecureParameters struct {
-	// Interval is the epoch interval (in blocks).
-	Interval int64 `json:"interval,omitempty"`
+// ConsensusParameterChanges are allowed beacon consensus parameter changes.
+type ConsensusParameterChanges struct {
+	// Backend is the new backend.
+	Backend *string `json:"backend"`
```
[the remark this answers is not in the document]
Yawning: This shouldn't be supported ever.

go/beacon/api/api.go
```diff
@@ -131,62 +129,42 @@ type ConsensusParameters struct {
...
+
+	// InsecureParameters are the new insecure parameters.
+	InsecureParameters **InsecureParameters `json:"insecure_parameters,omitempty"`
```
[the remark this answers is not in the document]
Yawning: This backend is *NEVER* going to be deployed, so the ability to mess with it should be omitted.

Yawning (commented):
I'm moderately concerned that changing parameters that never were intended to be changed like this will cause bad things to happen...

go/beacon/api/api.go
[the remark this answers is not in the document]
kostko: Agreed.

kostko:
> I'm moderately concerned that changing parameters that never were intended to be changed like this will cause bad things to happen...

Please flag any such parameters so they can be removed. E.g. I assume all of the beacon parameters are sensitive and I would be fine with just not supporting any of them to be changed like this.

go/beacon/api/api.go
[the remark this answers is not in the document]
peternose: With double pointers we can reset (set to `nil`) insecure parameters if we decide to change backend to VRF.
Removed double pointers and kept only VRF parameters.

go/consensus/tendermint/apps/beacon/beacon.go
```diff
@@ -70,7 +77,55 @@ func (app *beaconApplication) BeginBlock(ctx *api.Context, req types.RequestBegi
 }
 
 func (app *beaconApplication) ExecuteMessage(ctx *api.Context, kind, msg interface{}) (interface{}, error) {
-	return nil, fmt.Errorf("beacon: unexpected message")
+	switch kind {
+	case governanceApi.MessageValidateParameterChanges, governanceApi.MessageChangeParameters:
+		// Unmarshal changes and check if they should be applied to this module.
```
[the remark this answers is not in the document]
peternose: Done. Messages are now handled in `messages.go`.

go/oasis-test-runner/scenario/e2e/runtime/governance_change_parameters.go
peternose: Removed e2e test so this is not relevant anymore. Will remember for the next time.

go/oasis-test-runner/scenario/e2e/runtime/governance_change_parameters.go
peternose: Agree. Added unit tests for message handlers and extended node unit test.

go/oasis-node/cmd/debug/txsource/workload/governance.go
peternose: Added randomly generated change parameters proposals.

kostko (commented):
@Yawning should we just prohibit _any_ beacon parameters from being changed this way?

go/oasis-node/cmd/debug/txsource/workload/governance.go
```diff
@@ -277,6 +281,170 @@ OUTER:
 	}
 }
 
+func (g *governanceWorkload) doChangeParametersProposal() error { // nolint: gocyclo
```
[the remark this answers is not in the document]
peternose: Yes. If that doesn't hold for current tests, it might brake future tests or make them flaky. Not sure if there is a parameter which is safe to change. I can always keep the same parameter values, but we probably gain nothing with this.

go/beacon/api/sanity_check.go
```diff
@@ -0,0 +1,70 @@
...
+	if unsafeFlags && !flags.DebugDontBlameOasis() {
+		return fmt.Errorf("one or more unsafe debug flags set")
+	}
+
+	return nil
+}
+
+// SanityCheck performs a sanity check on the consensus parameter changes.
+func (c *ConsensusParameterChanges) SanityCheck() error {
+	if c.VRFParameters == nil {
```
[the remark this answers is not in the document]
peternose: Since we deploy only VRF and insecure backend is for testing purposes only, I think it makes no sense to check that. But I could accept current parameters and check that. An alternative would be to add a comment and once we add another backend that could be deployed we change this appropriately.

go/consensus/tendermint/apps/governance/transactions.go
```diff
@@ -43,6 +44,11 @@ func (app *governanceApplication) submitProposal(
 		return fmt.Errorf("failed to fetch consensus parameters: %w", err)
 	}
 
+	// To not violate the consensus, change parameters proposals should be ignored when disabled.
+	if proposalContent.ChangeParameters != nil && !params.EnableChangeParametersProposal {
+		return governance.ErrInvalidArgument
```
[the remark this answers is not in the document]
peternose: I'm aware that different errors are returned, just wasn't sure if it is breaking. Higher in the stack trace we return `module`, `code` and `message`, so I thought all are breaking or none.
```
Codespace: module,
Code:      code,
Log:       err.Error(),
```
Will fix and test with different versions.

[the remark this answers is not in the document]
kostko: Yeah `Log` is not consensus breaking.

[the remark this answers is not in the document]
peternose: Offline discussion revealed that the `module` (`codespace` in Tendermint) is currently not consensus breaking, only the `code` is. Since all unknown errors `errUnknownError` (`module=unknown, code=1`) have the same `code` as `ErrInvalidArgument` (`module=governance, code=1`) errors, current code "magically" works.
To avoid having different errors I changed `ExecuteTx` to return `ErrInvalidArgument` when the body of a transaction is invalid.

kostko (approved):

Yawning:
> @Yawning should we just prohibit _any_ beacon parameters from being changed this way?

I would.

kostko:
Ok, let's just remove support for changing beacon parameters.

peternose:
> Ok, let's just remove support for changing beacon parameters.

Removed and squashed commits.

kostko (approved):

PR #4954, opened 2022-09-23
docs: Reword Consensus -> consensus

matevz:
See oasisprotocol/docs#242


matevz:
Consensus Layer was used as a title on the image, ignoring.

PR #4957, opened 2022-09-27
Add Genesis file flag to oasis-net-runner CLI

aefhm:
**Why**

Add feature to aid https://github.com/oasisprotocol/oasis-indexer/issues/119.

**TODO**
- [ ] Ask for a release to obtain prebuilt binary

### DIFF
--- a/.changelog/4957.feature.md
+++ b/.changelog/4957.feature.md
@@ -1,0 +1,4 @@
+go/oasis-net-runner: Add Genesis file flag to oasis-net-runner CLI
+
+Network configuration accepts a Genesis file. This supports running a more
+deterministic local environment.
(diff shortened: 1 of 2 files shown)


kostko (commented):
Specifying a genesis file should be incompatible with many of the other configuration options as it will result in everything being overriden and those options will not have any effect and may result in broken configurations.

For example all of the following will be overriden so the manually specified genesis file must exactly match what is expected based on the configured fixture:
* Entities.
* Funding of entities.
* Runtimes.
* Node identities.
* Validator set.
* Initial height.
* Halt epoch.

Also note that you don't need to change the default fixture as the net runner supports specifying JSON fixtures. Generate a JSON fixture (you can generate one from defaults by using `oasis-net-runner dump-fixture ...` and then update as you want, e.g. by adding the genesis file) and then instead of using the `--fixture.default.*` options specify `--fixture.file myfile.json`.

aefhm:
Gotcha. I see, so the way is to nest `fixture.json` file with:

```
{
  ...
  "network": {
    ...
    "genesis_file": {...}
  }
}
```

in order to achieve a static/deterministic chaincontext. There is no need to modify the net-runner.

PR #4959, opened 2022-09-28
go/runtime/txpool: Add block propagation delay before tx publish

kostko:
After processing a new block locally, make sure that any transactions that are checked against that block are only propagated after waiting for the `newBlockPublishDelay` to increase probability that peers are up to date with the new block.

Obviously a better way would be to check the current round of each peer (e.g., via a p2p update protocol), but such filtering is unfortunately not possible to do with the current go-libp2p-pubsub implementation.

### DIFF
--- a/.changelog/4959.bugfix.md
+++ b/.changelog/4959.bugfix.md
@@ -1,0 +1,1 @@
+go/runtime/txpool: Add block propagation delay before tx publish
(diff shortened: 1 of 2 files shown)


pro-wh:
oh is this to handle txs that are valid in the new block but invalid in the old block?

kostko:
Yes.

Yawning (approved):

PR #4964, opened 2022-10-05
Peternose/cfg/seed node

peternose:
### Task
This is the first part of the _Seed node should also bootstrap libp2p peers_ task. Introduced changes are configuration breaking as from now on Oasis seed node has to be explicitly started with `--p2p.seed_node` flag or with the following config.
```
p2p:
  seed_node: true
```
A seed node can now support different services, which have to be enabled in the configuration (e.g. Tendermint seed node, Libp2p seed node).
```
# We enable Tendermint seed mode as before.
consensus:
  tendermint:
    mode: seed
    
# Enable libp2p seed node (will be done in the second part).
worker:
  p2p:
     # Config for host and discovery (pex,kad-dht,...)
```

Summary of changes:
- Moved `ModeSeed` away from the consensus layer as it doesn't belong there. This mode will be removed in the future if we decide to replace p2p stack but currently we keep the mode in the seed package so that the config stays the same.
- Changed seed service from the `Backend` to the `BackgroundService` as from now on it will be ran as a background service in the seed node.
- Created seed node which defers from the normal node as it doesn't have the consensus layer and all runtime workers.
- Merged controlled node and node controller, so that seed node can show its own status only.

### Test
Locally `oasis-net-runner` starts seed with `--p2p.seed_node` and validators with ` --p2p.seeds CGEm47kBve3q1QQWdDG3PfH5yLY1IIM6ozlPXZiA4KY=@127.0.0.1:20000`.

If we check seed node`s status we can see that only needed data is returned.
```
> oasis-node control status
{
  "software_version": "22.1-git5b6eea042+dirty",
  "identity": {
    "node": "AplzVKNeSAS4KiJnd+4l7o3HmE5bqUUwHWS0P7x7Iv8=",
    "p2p": "WoGCUWiHCZcV9dq/8AikQH3ffDueum2qRqybUQIUQI4=",
    "consensus": "qrXeEegG6dHSTsZaPbwofMvcv6Ealj/xAYXehD79ja8=",
    "tls": [
      "u7goIeofG3jn13QINl60fIEWN9Zs7RJdK2hLvwNYiQ0=",
      "HZrQEFtY9HqfrQ58q0F6uJJt2Zv4muGCyIZzdv7BXNk="
    ]
  },
  "seed": {
    "tendermint": {
      "chain_context": "5a1f9710fd9b7456a6bd074c5079b57e3cee930e0ee848e1c6db38e2a212e880",
      "addresses": [
        "WoGCUWiHCZcV9dq/8AikQH3ffDueum2qRqybUQIUQI4=@127.0.0.1:20000"
      ],
      "node_peers": [
        "eea664c592515f50689467371b7bc41a9b89c0db@127.0.0.1:20014",
        "cb5214e9b6e667e4b418535dfaf4bf65013f30d3@127.0.0.1:20005",
        "7fd9f39fb2e7e7b37913090e5b5b693859416576@127.0.0.1:20008",
        "ea8747bcef336063dd674fee48464d41613371ea@127.0.0.1:20011",
        "14c59affd30f571602c6b58016d0ca2bd7251f5f@127.0.0.1:20017"
      ]
    }
  }
}
```


go/control/api/api.go
```diff
@@ -156,6 +166,12 @@ type RuntimeStatus struct {
 	Storage *storageWorker.Status `json:"storage,omitempty"`
 }
 
+// SeedStatus is the status of the seed node and its services.
+type SeedStatus struct {
+	// Tendermint is the Tendermint seed node service status.
+	Tendermint *tmSeed.Status `json:"tendermint,omitempty"`
```
kostko, line 172: I would make this independent of the consensus layer backend, maybe even move the fields here directly. There is probably no need to split this up too much.

go/consensus/tendermint/full/full.go
```diff
@@ -600,6 +601,27 @@ func (t *fullService) lazyInit() error {
 		return err
 	}
 
+	// Convert addresses and public keys to Tendermint form. Converted IDs need to be lowercase
```
kostko, line 604: Probably no need for the "Converted IDs..." part of the comment as it is already explained in the function doc.

.changelog/4964.cfg.md
```diff
@@ -0,0 +1,8 @@
+go/consensus/tendermint: Refactor seed node
+
+Until now a seed node was a Tendermint backend-specific construct which
+supported only one service. To provide support for other services we had
+to refactor the seed node and move Tendermint seed construct to a service.
+Configuration of the latter stayed the same. However, to run a node as
+a seed node one has to specify `--mode seed` flag or add `mode: seed`
+setting to the config file.
```
kostko, line 8: After this change also the various addresses need to be changed? E.g. to use public keys instead of those tendermint addresses?

go/oasis-node/cmd/node/seed.go
```diff
@@ -0,0 +1,145 @@
...
+		)
+		return nil, err
+	}
+	node.svcMgr.Register(node.grpcInternal)
+
+	// Register the node as a node controller.
+	controlApi.RegisterService(node.grpcInternal.Server(), node)
+
+	// Initialize and start services.
+	if tmSeed.Enabled() {
```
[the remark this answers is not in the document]
peternose: I guess by twice you mean that the Oasis seed node is enabled once via `mode` and Tendermint seed node (a service inside Oasis seed node) is enabled via `consensus.tendermint.mode`.
The reason for that is in the description of the task, where it is written that the seed node should be enabled differently and Tendermint seed behavior should be enabled by the consensus layer.
We can make this default, so that Tendermint seed node service runs on every Oasis seed node.

go/control/api/api.go
kostko: This is true, but the existence of multiple internal services should be an internal detail as it may change in the future (e.g. migrate towards not having this Tendermint-specific service).

kostko (approved):

PR #4981, opened 2022-10-13
go/p2p: Bootstrap libp2p peers

peternose:
### Task

Add libp2p seed node which can bootstrap libp2p peers.

Notes:
- Rate limiter was not added. Peer store is simple without peer scoring, advanced peer selection, etc. This can/should be added in separate PRs.

### Configuration

Libp2p seed node can be configured using the following settings:
```
# Host
p2p.port

# Connection manager
p2p.max_num_peers
p2p.peer_grace_period
p2p.persistent_peers

# Connection gater
p2p.blocked_peers

# Discovery
p2p.discovery.bootstrap.enable
p2p.discovery.bootstrap.retention_period
```

Peers still configure seeds using the following setting:
```
p2p.seeds
```
As we currently run 2 seed nodes on one address but on different ports, 2 seeds need to be configured.
```
p2p.seeds:
 - pubkey@IP:consensus-port
 - pubkey@IP:libp2p-port
```

### Test
Tested locally. If we disable registry discovery (e.g. comment `connectRegisteredPeers` in `PeerManager`), the peer nodes still connect.

```
"runtimes": {
  "8000000000000000000000000000000000000000000000000000000000000000": {
    "committee": {
      "peers": [
          "/ip4/192.168.0.2/tcp/20017/p2p/12D3KooWP2DNWFz7ivdpENeQirCkPTwCJ762EKChdKHCNEuAxUVC",
          "/ip4/192.168.0.2/tcp/20014/p2p/12D3KooWGe86FFdDn3vZtEupwTM55LjsL2WVGfM1jtrhEaGdgh7X",
          "/ip4/192.168.0.2/tcp/20019/p2p/12D3KooWD8gjDP1pxqhpjy2ZkPguWVKSMBu26mJXwridnoAynhxM"
      ],
    },
  },
}
```
Seed node now shows addresses and peers from both seed nodes (Tendermint, libp2p).
```
➜  seed-0 oasis-node control status 
{
  "seed": {
    "addresses": [
      // Tendermint address.
      "mbqm2wnNxSgsm8OSyCT+fYo1JuQU1tJWU9wVz8vv1Cs=@127.0.0.1:20000",
      // Libp2p address.
      "{12D3KooWALMXUq9y2BwY6VgEdJyVDcmWrdQsJ6LJth7eoCcRhheN: [/ip4/192.168.64.107/tcp/20001 /ip4/127.0.0.1/tcp/20001]}"
    ],
    "node_peers": [
      // Tendermint peers.
      "4533c1467e52281edcc866176cb0cc2e968880c5@127.0.0.1:20006",
      "45a2a3e75e00a2a10d3dfba5e0b6ca8a5dd91848@127.0.0.1:20015",
      "9ee8d0a7ff2c20db17f25f39a17a548e6582af4e@127.0.0.1:20009",
      "b9cec91ce2fa349f318f7a28e2fc739d79c05a5e@127.0.0.1:20012",
      "762d91e3e0ddd029d6342d18f9f6690cf4755274@127.0.0.1:20018",
      "75fc1446e282ca226958c47f68620318a78e73c5@127.0.0.1:20002",
      "0566341e8963c12a73f730b8872f30dffad27d7c@127.0.0.1:20004",
      // Libp2p peers.
      "/ip4/127.0.0.1/tcp/20008/p2p/12D3KooWCv1cFd8k1MhxaKU2LKuCanPLx8sYym6bz2qQWoPkhVsW",
      "/ip4/127.0.0.1/tcp/20011/p2p/12D3KooWEDUkJKh4xRnjrfA6iNQZCKCDvnrLrb6tZfhV9Kkm9gG4",
      "/ip4/127.0.0.1/tcp/20014/p2p/12D3KooWGe86FFdDn3vZtEupwTM55LjsL2WVGfM1jtrhEaGdgh7X",
      "/ip4/127.0.0.1/tcp/20017/p2p/12D3KooWP2DNWFz7ivdpENeQirCkPTwCJ762EKChdKHCNEuAxUVC",
      "/ip4/127.0.0.1/tcp/20019/p2p/12D3KooWD8gjDP1pxqhpjy2ZkPguWVKSMBu26mJXwridnoAynhxM"
    ]
  }
}
```

### DIFF
--- a/.changelog/4981.feature.md
+++ b/.changelog/4981.feature.md
@@ -1,0 +1,9 @@
+go/p2p: Bootstrap libp2p peers using seed nodes
+
+Seed nodes can now bootstrap libp2p peers. Bootstrapping can be enabled or
+disabled using a new configuration setting named
+`p2p.discovery.bootstrap.enable`
+which can be used on seed nodes as well as non-seed nodes (e.g. clients,
+key managers). The latter can also configure how frequently peers are fetched
+from the seed nodes with
+`p2p.discovery.bootstrap.retention_period`.
--- a/go/p2p/discovery/bootstrap/client.go
+++ b/go/p2p/discovery/bootstrap/client.go
@@ -1,0 +1,258 @@
+package bootstrap
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"sync"
+	"time"
+
+	"github.com/cenkalti/backoff/v4"
+	"github.com/libp2p/go-libp2p/core/discovery"
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/peerstore"
+
+	cmnBackoff "github.com/oasisprotocol/oasis-core/go/common/backoff"
+	"github.com/oasisprotocol/oasis-core/go/common/logging"
+	"github.com/oasisprotocol/oasis-core/go/p2p/rpc"
+)
+
+const (
+	// defaultRetentionPeriod is the default peer retention period.
+	defaultRetentionPeriod = time.Hour
+
+	// discoveryBackOffInitialInterval is the initial time interval for the exponential back-off
+	// used between failed discoveries.
+	discoveryBackOffInitialInterval = time.Minute
+
+	// discoveryBackOffMaxInterval is the maximum time interval for the exponential back-off
+	// used between failed discoveries.
+	discoveryBackOffMaxInterval = time.Hour
+
+	// maliciousSeedBackOff is the time interval during which no peers will be returned if the seed
+	// has behaved maliciously.
+	maliciousSeedBackOff = 24 * time.Hour
+)
+
+// ClientOptions are client's peer discovery options.
+type ClientOptions struct {
+	retentionPeriod time.Duration
+}
+
+// ClientOption is a client peer discovery option setter.
+type ClientOption func(opts *ClientOptions)
+
+// WithRetentionPeriod configures peer retention period.
+func WithRetentionPeriod(retentionPeriod time.Duration) ClientOption {
+	return func(opts *ClientOptions) {
+		opts.retentionPeriod = retentionPeriod
+	}
+}
+
+// DefaultClientOptions returns the default client options.
+func DefaultClientOptions() *ClientOptions {
+	return &ClientOptions{
+		retentionPeriod: defaultRetentionPeriod,
+	}
+}
+
+type discoveryCache struct {
+	peers   []peer.AddrInfo
+	expires time.Time
+	limit   int
+}
+
+// client is an implementation of a peer discovery that fetches peers from the seed node and
+// advertises services.
+type client struct {
+	logger *logging.Logger
+
+	host host.Host
+	rc   rpc.Client
+
+	seed            peer.AddrInfo
+	retentionPeriod time.Duration
+
+	mu            sync.Mutex
+	nextDiscovery time.Time
+	backoff       *backoff.ExponentialBackOff
+	cache         map[string]*discoveryCache
+}
+
+// NewClient creates a new bootstrap protocol client.
+func NewClient(h host.Host, seed peer.AddrInfo, opts ...ClientOption) discovery.Discovery {
+	l := logging.GetLogger("p2p/discovery/bootstrap").With(
+		"seed_id", seed.ID,
+		"seed_addrs", seed.Addrs,
+	)
+
+	cos := DefaultClientOptions()
+	for _, opt := range opts {
+		opt(cos)
+	}
+
+	// Add the seed to the peer store so that libp2p knows how to contact it.
+	h.Peerstore().AddAddrs(seed.ID, seed.Addrs, peerstore.PermanentAddrTTL)
+
+	rc := rpc.NewClient(h, ProtocolID())
+
+	bo := cmnBackoff.NewExponentialBackOff()
+	bo.InitialInterval = discoveryBackOffInitialInterval
+	bo.MaxInterval = discoveryBackOffMaxInterval
+	bo.Reset()
+
+	return &client{
+		logger:          l,
+		host:            h,
+		rc:              rc,
+		seed:            seed,
+		retentionPeriod: cos.retentionPeriod,
+		nextDiscovery:   time.Now(),
+		backoff:         bo,
+		cache:           make(map[string]*discoveryCache),
+	}
+}
+
+// Advertise implements discovery.Advertiser and discovery.Discovery.
+func (c *client) Advertise(ctx context.Context, ns string, opts ...discovery.Option) (time.Duration, error) {
+	req := AdvertiseRequest{
+		Namespace: ns,
+	}
+	var res AdvertiseResponse
+
+	pf, err := c.rc.Call(ctx, c.seed.ID, MethodAdvertise, req, &res)
+	if err != nil {
+		pf.RecordFailure()
+
+		c.logger.Error("failed to advertise",
+			"err", err,
+		)
+		return 0, fmt.Errorf("failed to advertise: %w", err)
+	}
+
+	pf.RecordSuccess()
+
+	return res.TTL, nil
+}
+
+// FindPeers implements discovery.Discoverer and discovery.Discovery.
+func (c *client) FindPeers(ctx context.Context, ns string, opts ...discovery.Option) (<-chan peer.AddrInfo, error) {
+	// Check how many peers we need to fetch.
+	var options discovery.Options
+	err := options.Apply(opts...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to apply options: %w", err)
+	}
+	limit := options.Limit
+
+	// Fetch peers and flush them down the channel.
+	ch := make(chan peer.AddrInfo)
+	go func() {
+		defer close(ch)
+		peers := c.fetchPeers(ctx, ns, limit)
+		sendPeers(ctx, ch, peers, limit)
+	}()
+
+	return ch, nil
+}
+
+func (c *client) fetchPeers(ctx context.Context, ns string, limit int) []peer.AddrInfo {
+	// Allow only one fetch request at a time. No need for parallelism here as speed is not
+	// crucial and the number of requests should be low anyway.
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	cache, ok := c.cache[ns]
+	if !ok {
+		cache = &discoveryCache{}
+		c.cache[ns] = cache
+	}
+
+	now := time.Now()
+	if c.nextDiscovery.After(now) {
+		return cache.peers
+	}
+	if cache.expires.After(now) && (cache.limit == 0 || cache.limit >= limit && limit != 0) {
+		return cache.peers
+	}
+
+	peers, pf, err := func() ([]peer.AddrInfo, rpc.PeerFeedback, error) {
+		req := DiscoverRequest{
+			Namespace: ns,
+			Limit:     limit,
+		}
+		var rsp DiscoverResponse
+
+		pf, err := c.rc.Call(ctx, c.seed.ID, MethodDiscover, req, &rsp)
+		if err != nil {
+			c.logger.Error("failed to call seed node",
+				"err", err,
+			)
+			return nil, pf, err
+		}
+
+		if limit != 0 && len(rsp.Peers) > limit {
+			c.logger.Debug("seed node returned too many peers",
+				"limit", limit,
+				"received", len(rsp.Peers),
+			)
+			return nil, pf, ErrMaliciousSeed
+		}
+
+		peers := make([]peer.AddrInfo, len(rsp.Peers))
+		for i, json := range rsp.Peers {
+			err := peers[i].UnmarshalJSON(json)
+			if err != nil {
+				c.logger.Debug("failed to unmarshal addr info",
+					"err", err,
+				)
+				return nil, pf, ErrMaliciousSeed
+			}
+		}
+
+		return peers, pf, nil
+	}()
+
+	switch err {
+	case nil:
+		// All good. Start serving new peers.
+		cache.peers = peers
+		cache.limit = limit
+		cache.expires = now.Add(c.retentionPeriod)
+
+		c.backoff.Reset()
+		pf.RecordSuccess()
+
+	case ErrMaliciousSeed:
+		// Return no peers while the seed is marked as malicious.
+		cache.peers = nil
+		cache.limit = 0
+		cache.expires = now.Add(maliciousSeedBackOff)
+
+		pf.RecordFailure()
+		pf.RecordBadPeer()
+	default:
+		// Keep serving peers from the cache if something went wrong.
+		c.nextDiscovery = now.Add(c.backoff.NextBackOff())
+
+		pf.RecordFailure()
+	}
+
+	return cache.peers
+}
+
+func sendPeers(ctx context.Context, peerCh chan<- peer.AddrInfo, peers []peer.AddrInfo, limit int) {
+	if limit == 0 || limit > len(peers) {
+		limit = len(peers)
+	}
+
+	order := rand.Perm(len(peers))[:limit]
+	for _, i := range order {
+		select {
+		case peerCh <- peers[i]:
+		case <-ctx.Done():
+			return
+		}
+	}
+}
--- a/go/p2p/discovery/bootstrap/messages.go
+++ b/go/p2p/discovery/bootstrap/messages.go
@@ -1,0 +1,26 @@
+package bootstrap
+
+import (
+	"time"
+)
+
+// AdvertiseRequest is a message requesting seed to register the peer.
+type AdvertiseRequest struct {
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// AdvertiseResponse is a message response with a deadline for the next registration.
+type AdvertiseResponse struct {
+	TTL time.Duration `json:"ttl,omitempty"`
+}
+
+// DiscoverRequest is a message requesting peers.
+type DiscoverRequest struct {
+	Namespace string `json:"namespace,omitempty"`
+	Limit     int    `json:"limit,omitempty"`
+}
+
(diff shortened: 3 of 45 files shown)


.changelog/4981.feature.md
```diff
@@ -0,0 +1,27 @@
...
+Seed nodes can now bootstrap libp2p peers. Bootstrapping can be enabled or
+disabled using a new configuration setting named
+`worker.p2p.discovery.bootstrap.enable`
+which can be used on seed nodes as well as non-seed nodes (e.g. clients,
+key managers).
+
+To run a seed node with custom configuration that supports bootstrapping
+one needs to configure the libp2p host first:
+
+- `worker.p2p.port` # host
```
kostko, line 12: Should we move this directly under `p2p` at some point?

go/worker/common/p2p/p2p.go
```diff
@@ -315,112 +328,24 @@ func messageIdFn(pmsg *pb.Message) string { // nolint: revive
...
 
+	// Initialize bootstrapping.
+	var bootstrapDiscovery *bootstrap.Discovery
+	if cfg.EnableBootstrap {
+		bootstrapDiscovery = bootstrap.NewDiscovery(host, cfg.Seeds)
+	}
+
 	// Initialize the gossipsub router.
 	ctx, ctxCancel := context.WithCancel(context.Background())
 	pubsub, err := pubsub.NewGossipSub(
```
kostko, line 351: Note that there is also a pubsub option [`pubsub.WithDiscovery`](https://pkg.go.dev/github.com/libp2p/go-libp2p-pubsub#WithDiscovery) (see implementation in [`discovery.go`](https://github.com/libp2p/go-libp2p-pubsub/blob/master/discovery.go)) which seems like you could pass it the the discovery implementation and it would handle some of the stuff you implemented here?

Also while looking at what is available in go-libp2p, there is this ["backoff connector"](https://github.com/libp2p/go-libp2p/blob/master/p2p/discovery/backoff/backoffconnector.go) which seems like it could be useful and possibly be shared (there's also [`pubsub.WithDiscoverConnector`](https://github.com/libp2p/go-libp2p-pubsub/blob/v0.8.1/discovery.go#L343)) between both bootstrap discovery and registry-based discovery (currently implemented in the peer manager)? Maybe there are some caveats so you should check first if that is something useful.

.changelog/4981.feature.md
peternose: Of course. Move most of the stuff to `p2p`, keep worker related things there, move registry discovery under discoveries, ...

go/worker/common/p2p/p2p.go
peternose: I'm aware about pubsub discovery, but I prefer to add this in a separate PR for 2 reasons. 
- Every topic in pubsub has its own namespace (network in Tendermint) so without a middleware the seed cannot support more chains with different chain context.
- Rate at which pubsub receives new peers through a channel needs to be throttled (comment inside the code) and maybe also limited (to not disconnect all peers from one topic).

Agree, we should merge peers from all discoveries (bootstrap, PEX, registry) and use the backoff. Thought about doing that after we implement all discoveries (if we are planning to implement PEX also). However, I think the registry peers should have priority as those should be available most of the time.

Libp2p also offers dht discovery using kad-dht routing. We can also use that instead of our own PEX implementation, or both of them. Just not sure how to grade peers from so many sources :smile:

kostko: Ok, sounds good.

>  However, I think the registry peers should have priority as those should be available most of the time.

Agreed, they should have priority.

> Libp2p also offers dht discovery using kad-dht routing.

Yeah maybe let's not add that just yet.

kostko:
> Seed node now shows addresses and peers from both seed nodes (Tendermint, libp2p).

Can we have a consistent peer output (e.g. reuse the multiaddr string format as is used for committee peers)?

> Tested locally. If we disable registry discovery (e.g. hack connectWorker in PeerManager to not connect at all), the peer nodes still connect.

Can you also test how this works between consensus-only nodes (e.g., between validators that don't have any runtimes configured)?

peternose:
> Can we have a consistent peer output (e.g. reuse the multiaddr string format as is used for committee peers)?

First answer would be no, as I don't see Tendermint exposing peer's public key. They use IDs to verify public key gained with `secretConn.RemotePubKey()`, but I don't see this being exposed anywhere.

> Can you also test how this works between consensus-only nodes (e.g., between validators that don't have any runtimes configured)?

Consensus nodes use only Tendermint's p2p network, our libp2p worker is not even started. As they do not participate, there should be nothing to test. 
I could only check how Tendermint reacts when he gets an invalid seed address (same ip but different ports). Locally had no problems. I did check libp2p and that one has no problems as seed addresses are merged into one multiaddr and the library then uses the right port.

kostko:
> First answer would be no, as I don't see Tendermint exposing peer's public key. They use IDs to verify public keys gained with secretConn.RemotePubKey(), but I don't see this being exposed anywhere.

No, I mean for the libp2p peers, Tendermint peers can stay as they are. Currently it seems like libp2p peers are represented as a dict in some places (seed peers) and a multiaddr string in others (committee peers).

kostko:
> Consensus nodes use only Tendermint's p2p network, our libp2p worker is not even started. As they do not participate, there should be nothing to test.

Ok then we need to do this later (e.g. always start the libp2p worker). Because the idea is that we will start using our libp2p stack for some consensus services (e.g. state sync) and slowly get rid of Tendermint P2P.

peternose:
> No, I mean for the libp2p peers, Tendermint peers can stay as they are. Currently it seems like libp2p peers are represented as a dict in some places (seed peers) and a multiaddr string in others (committee peers).

Ok, will change.

go/p2p/discovery/bootstrap/client.go
```diff
@@ -0,0 +1,162 @@
...
+		)
+
+		return 0, fmt.Errorf("failed to advertise: %w", err)
+	}
+
+	return res.TTL, nil
+}
+
+// FindPeers implements discovery.Discoverer and discovery.Discovery.
+func (c *client) FindPeers(ctx context.Context, ns string, opts ...discovery.Option) (<-chan peer.AddrInfo, error) {
```
kostko, line 98: Is it ok for `FindPeers` to block until the peers are found? Or would it be better if it returned immediately but peers would be sent over the channel? Didn't check what the caller expects, but if it should be sync it seems weird that they would use a channel for the return value.

go/p2p/discovery/peerstore/peerstore.go
```diff
@@ -0,0 +1,403 @@
...
+		startOne:      cmSync.NewOne(),
+	}
+}
+
+// Add inserts peer address into the store under the given namespace.
+func (s *Store) Add(ns string, info peer.AddrInfo) time.Duration {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	peers, ok := s.registrations[ns]
```
kostko, line 87: What if a peer is submitting advertise requests and randomizing the namespace? It seems like this could cause the store to fill up unboundedly? Should there be a max cap of namespaces per peer? And a max cap of total advertisements to avoid memory exhaustion?

Also there should probably be a rate limit of advertisements per peer (probably a useful thing to have in general for our P2P RPC framework so that various methods could configure rate limits), but this is probably best done in a subsequent task.

go/p2p/discovery/bootstrap/client.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// NewClient creates a new bootstrap protocol client.
+func NewClient(h host.Host, seed peer.AddrInfo, opts ...ClientOption) discovery.Discovery {
+	l := logging.GetLogger("p2p/discovery/bootstrap").With(
+		"seed_id", seed.ID,
+		"seed_addrs", seed.Addrs,
+	)
+
+	do := ClientOptions{}
```
ptrus, line 57: Might make sense to define a default `ClientOptions` object, with some sane defaults (non-zero default `retentionPeriod`).

go/p2p/p2p.go
```diff
@@ -374,110 +377,16 @@ func messageIdFn(pmsg *pb.Message) string { // nolint: revive
...
-		libp2p.ListenAddrs(sourceMultiAddr),
-		libp2p.Identity(api.SignerToPrivKey(identity.P2PSigner)),
-		libp2p.ConnectionManager(cm),
-		libp2p.ConnectionGater(cg),
-	)
+	cfg.HostConfig.Signer = identity.P2PSigner
+	host, cg, err := NewHost(&cfg.HostConfig)
 	if err != nil {
-		return nil, fmt.Errorf("p2p: failed to initialize libp2p host: %w", err)
+		return nil, fmt.Errorf("worker/common/p2p: failed to initialize libp2p host: %w", err)
```
ptrus, line 389: I think the errors should now be prefixed only with `p2p:`

go/p2p/discovery/bootstrap/messages.go
```diff
@@ -0,0 +1,24 @@
+package bootstrap
+
+import "time"
+
+// AdvertiseRequest is a message requesting seed to register the peer.
+type AdvertiseRequest struct {
+	Namespace string `json:"namespace,omitempty"`
```
ptrus, line 7: Would it make sense for the protocol to support advertising multiple namespaces with a single request?

I'm not 100% sure, but it might make some parts of the code a bit simpler (for example, no need to track the combination of both per-namespace and per-seed backoffs in downstream code in `discovery.go`). I think it would also reduce the number of requests being sent (while slightly increasing the size of requests).

Just a thought, didn't think this fully through.

go/p2p/discovery/bootstrap/client.go
```diff
@@ -0,0 +1,162 @@
...
+		Limit:     limit,
+	}
+	var rsp DiscoverResponse
+
+	_, err := c.rc.Call(ctx, c.seed.ID, MethodDiscover, req, &rsp)
+	if err != nil {
+		return nil, fmt.Errorf("failed to call seed node: %w", err)
+	}
+
+	peers := make([]peer.AddrInfo, len(rsp.Peers))
```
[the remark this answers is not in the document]
peternose: Not sure about punishing the node as we will probably have only few seeds. If this was PEX, then I would punish/block the peer. And yes, I should stop processing them although it feels like I'm throwing away the data.

go/p2p/discovery/bootstrap/client.go
peternose: No, it should be async. Will fix.

go/p2p/discovery/peerstore/peerstore.go
peternose: Rate limiting solves most of the problems, as cleanup/gc should be done periodically. Can add the cap now, and add rate limiting in a separate task.

go/p2p/discovery/bootstrap/server.go
```diff
@@ -0,0 +1,97 @@
...
+	limit := MaxPeers
+	if req.Limit > 0 && req.Limit < MaxPeers {
+		limit = req.Limit
+	}
+
+	peers := s.store.NamespacePeers(req.Namespace, limit)
+
+	encoded := make([][]byte, 0, len(peers))
+	for _, info := range peers {
+		json, err := info.MarshalJSON()
```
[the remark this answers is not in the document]
peternose: `AddrInfo` contains `Multiaddr` which is an interface, so I cannot use CBOR as it panics `panic: cbor: cannot unmarshal byte string into Go struct field peer.AddrInfo.Addrs of type multiaddr.Multiaddr`.

go/p2p/discovery/bootstrap/messages.go
peternose: > Would it make sense for the protocol to support advertising multiple namespaces with a single request?

Yes, it would. However, I wanted to follow interfaces suggested by libp2p [discovery](https://github.com/libp2p/go-libp2p/blob/v0.23.2/core/discovery/discovery.go#L12) package which make this impossible. I decided to still follow them as there should be only few requests per hour. Maybe I should just abandon that.

go/p2p/discovery/bootstrap/client.go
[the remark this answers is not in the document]
kostko: This would be a protocol violation, no? If you make a request and say give me at most 10 peers, but the node is giving you 1000 peers, then there is either a bug or someone is not following the protocol which is to respect the limit field. If you want more then you should set a higher limit in which case it is fine.

go/p2p/discovery/peerstore/peerstore.go
kostko: Yeah, for now just add the cap and we'll do rate limiting later.

go/p2p/discovery/bootstrap/client.go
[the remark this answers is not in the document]
peternose: Yes, it's protocol violation. Will just punish it as blocking could left us with no seeds.

go/p2p/discovery/bootstrap/server.go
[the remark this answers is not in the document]
ptrus: In that case, maybe the binary marshaller/unmarshaller makes more sense to use.

go/p2p/rpc/rpc.go
```diff
@@ -21,16 +22,25 @@ type P2P interface {
 	Host() core.Host
 }
 
-// contextKeyPeerID is the context key used for storing the peer ID.
-type contextKeyPeerID struct{}
+// contextKeyPeerID is the context key used for storing the peer addr info.
```
ptrus, line 25:
```suggestion
// contextKeyPeerAddrInfo is the context key used for storing the peer addr info.
```

go/p2p/discovery/bootstrap/client.go
```diff
@@ -0,0 +1,194 @@
...
+		return nil, fmt.Errorf("failed to apply options: %w", err)
+	}
+	limit := options.Limit
+
+	// Lock also RPCs to ensure only one request per retention period.
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if c.timestamps[ns].Add(c.retentionPeriod).Before(time.Now()) {
+		peers, err := c.findPeersLocked(ctx, ns, limit)
```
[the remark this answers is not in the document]
kostko: Yeah this was my thinking with the earlier comment about this whole thing not being async. The part that needs to be async are these remote discovery calls not just sorting the peers.

go/p2p/discovery/bootstrap/server.go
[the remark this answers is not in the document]
peternose: If `AddrInfo` had one. `ID` and `multiaddr.Multiaddr` have it, so I could define my own marshaller. But is it worth it?

go/p2p/discovery/peerstore/peerstore.go
```diff
@@ -0,0 +1,502 @@
...
+	}()
+
+	go func() {
+		defer wg.Done()
+		s.startCleanups(ctx)
+	}()
+}
+
+// startBackups periodically backups peers to the common store.
+func (s *Store) startBackups(ctx context.Context) {
```
ptrus, line 416: The backup part is very similar to the `peermgmt/backup.go`. Could the common backup functionality be extracted and reused?

go/p2p/discovery/bootstrap/client.go
```diff
@@ -0,0 +1,194 @@
...
+	_, err := c.rc.Call(ctx, c.seed.ID, MethodDiscover, req, &rsp)
+	if err != nil {
+		return nil, fmt.Errorf("failed to call seed node: %w", err)
+	}
+
+	if limit != 0 && len(rsp.Peers) > limit {
+		c.logger.Debug("seed node returned too many peers",
+			"limit", limit,
+			"received", len(rsp.Peers),
+		)
```
kostko, line 178: Should probably also `pf.RecordFailure()`?

go/p2p/discovery/bootstrap/client.go
```diff
@@ -0,0 +1,194 @@
...
+		return nil, ErrMaliciousSeed
+	}
+
+	peers := make([]peer.AddrInfo, len(rsp.Peers))
+	for i, p := range rsp.Peers {
+		err := peers[i].UnmarshalJSON(p)
+		if err != nil {
+			c.logger.Debug("failed to unmarshal addr info",
+				"err", err,
+			)
```
kostko, line 188: Should probably also `pf.RecordFailure()`?

go/p2p/discovery/bootstrap/client.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+func (c *client) findPeersLocked(ctx context.Context, ns string, limit int) ([]peer.AddrInfo, error) {
+	req := DiscoverRequest{
+		Namespace: ns,
+		Limit:     limit,
+	}
+	var rsp DiscoverResponse
+
+	_, err := c.rc.Call(ctx, c.seed.ID, MethodDiscover, req, &rsp)
```
kostko, line 169: Should store `pf` here and use it below (record success at the end, record failure if there are issues with the response).

peternose:
Addressed the comments:
- Made discovery async. As peer channel is returned before seed responds, I had to move logic for detecting malicious seeds to the client. Same goes for backoffs. Malicious seeds are now blocked for 1 day.
- Added `BinaryAddrInfo`, but I still prefer using json marshaller as code is more clean.
- Extracted backup logic. Libp2p and discovery peerstores now use `backup` and  `scheduler` packages for backups and cleanups.
- The code ignored peer feedbacks because p2p rpc client doesn't have any listeners. And I'm not sure if we should add one as latency here is not that important. We want to connect to all seeds and not only the fastest ones. Furthermore, before using feedbacks we should probably modify them and bind them to methods, as e.g. advertisement can have lower latency than discovery.
- Changed `peerstoreBackup` to backup all peers and not only the connected ones.

kostko (commented):
> Added BinaryAddrInfo, but I still prefer using json marshaller as code is more clean.

Yeah this seems quite messy and it also requires us to keep up with any upstream changes?

kostko (approved):

ptrus (approved):
Looks good, left a couple of minor comments

PR #4995, opened 2022-10-24
go/oasis-node: Start libp2p node always

peternose:
Always start libp2p node and not only when runtimes are configured. This will include the validator nodes into the libp2p network and make them available for consensus-related services.


Yawning (approved):
Sure, this seems fine.  I know it's orthogonal to this, but do we have a story for p2p + sentry nodes?

kostko (approved):

PR #5000, opened 2022-10-26
go/p2p/rpc: add support for consensus-wide libp2p protocols

ptrus:
This will be needed for adding a consensus-wide light-blocks libp2p protocol.

### DIFF
--- a/go/p2p/rpc/client.go
+++ b/go/p2p/rpc/client.go
@@ -12,11 +12,9 @@ import (
 	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/protocol"
 
-	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/errors"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
-	"github.com/oasisprotocol/oasis-core/go/common/version"
 	"github.com/oasisprotocol/oasis-core/go/common/workerpool"
 )
 
@@ -220,7 +218,6 @@ type client struct {
 
 	host       core.Host
 	protocolID protocol.ID
-	runtimeID  common.Namespace
 
 	opts *ClientOptions
 
@@ -509,9 +506,7 @@ func (c *client) sendRequestAndDecodeResponse(
 }
 
 // NewClient creates a new RPC client for the given protocol.
-func NewClient(p2p P2P, runtimeID common.Namespace, protocolID string, version version.Version, opts ...ClientOption) Client {
-	pid := NewRuntimeProtocolID(runtimeID, protocolID, version)
-
+func NewClient(p2p P2P, protocolID protocol.ID, opts ...ClientOption) Client {
 	var co ClientOptions
 	for _, opt := range opts {
 		opt(&co)
@@ -522,14 +517,10 @@ func NewClient(p2p P2P, runtimeID common.Namespace, protocolID string, version v
 		return &nopClient{&nopPeerManager{}}
 	}
 	return &client{
-		PeerManager: NewPeerManager(p2p, pid, co.stickyPeers),
+		PeerManager: NewPeerManager(p2p, protocolID, co.stickyPeers),
 		host:        p2p.GetHost(),
-		protocolID:  pid,
-		runtimeID:   runtimeID,
+		protocolID:  protocolID,
 		opts:        &co,
-		logger: logging.GetLogger("worker/common/p2p/rpc/client").With(
-			"protocol", protocolID,
-			"runtime_id", runtimeID,
-		),
+		logger:      logging.GetLogger("p2p/rpc/client").With("protocol", protocolID),
 	}
 }
--- a/go/p2p/rpc/rpc.go
+++ b/go/p2p/rpc/rpc.go
@@ -23,6 +23,11 @@ type P2P interface {
 	GetHost() core.Host
 }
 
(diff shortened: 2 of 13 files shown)


go/p2p/rpc/rpc.go
```diff
@@ -23,6 +23,11 @@ type P2P interface {
 	GetHost() core.Host
 }
 
+// NewProtocolID generates a protocol identifier for a consensus P2P protocol.
+func NewProtocolID(protocolID string, version version.Version) protocol.ID {
+	return protocol.ID(fmt.Sprintf("/oasis/%s/consensus/%s", protocolID, version.MaskNonMajor()))
```
ptrus, line 28: Runtime protocol ID is:
`/oasis/[protocol_id]/[runtime_id]/[version]`

The currently proposed consensus protocol ID is:
`/oasis/[protocol_id]/consensus/[version]`

---

This also made me think, should we include chain context in the protocol identifiers? (also for runtime protocols)

peternose: It would be nice if mainnet and testnet protocol IDs would differ so that we can use the same discovery techniques for both of them (seed node, pex, ..). Otherwise, we will have to prefix them somewhere, e.g. in a discovery wrapper. 

The protocol ID should be unique, so no need for `consensus` part.

Topics are named `[chain_context]/[comittee_version]/[runtime_id]/[kind]` and `pubsub` prefixes this with `floodsub:` to get the protocol ID or namespace for that topic. I think that in the next release we can change the names, so maybe use something like this:
- Runtime: `/oasis/[chain_context]/[protocol_id]/[runtime_id]/[version]` 
- Consensus:  `/oasis/[chain_context]/[protocol_id]/[version]`
- Topic: `/oasis/[chain_context]/[kind]/[runtime_id]/[comittee_version]`
Topic kinds and protocol IDs should be unique.

go/p2p/rpc/client.go
```diff
@@ -509,8 +508,19 @@ func (c *client) sendRequestAndDecodeResponse(
 }
 
 // NewClient creates a new RPC client for the given protocol.
-func NewClient(p2p P2P, runtimeID common.Namespace, protocolID string, version version.Version, opts ...ClientOption) Client {
-	pid := NewRuntimeProtocolID(runtimeID, protocolID, version)
+func NewClient(p2p P2P, runtimeID *common.Namespace, protocolID string, version version.Version, opts ...ClientOption) Client {
```
peternose, line 511: My plan was to change this when I get to my RPC taks. I would change this to `NewClient(p2p P2P, p core.ProtocolID, ...)` and let the caller decide between `NewProtocolID` and `NewRuntimeProtocolID`. This way you don't need to work with pointers.

I think that the `runtimeID` is never used so this can be removed. Unless I'm missing something.

kostko: Yeah I think it makes sense to just take a `core.ProtocolID` here and make this up to the caller.

go/p2p/rpc/rpc.go
kostko: Yeah we can afford protocol-breaking changes in master, so we can change these to include the chain context. The above scheme makes sense, I'll open a task so we don't forget about it.

go/p2p/rpc/client.go
ptrus: Agreed, updated.

go/p2p/rpc/rpc.go
ptrus: Ok, i'll drop the "consensus" part for now, and we can update to include chain-context in a followup PR.

go/p2p/rpc/client.go
```diff
@@ -522,14 +517,10 @@ func NewClient(p2p P2P, runtimeID common.Namespace, protocolID string, version v
...
 		host:        p2p.GetHost(),
-		protocolID:  pid,
-		runtimeID:   runtimeID,
+		protocolID:  protocolID,
 		opts:        &co,
-		logger: logging.GetLogger("worker/common/p2p/rpc/client").With(
-			"protocol", protocolID,
-			"runtime_id", runtimeID,
-		),
+		logger:      logging.GetLogger("worker/common/p2p/rpc/client").With("protocol", protocolID),
```
kostko, line 524:
```suggestion
		logger:      logging.GetLogger("p2p/rpc/client").With("protocol", protocolID),
```

go/p2p/rpc/server.go
```diff
@@ -106,16 +103,10 @@ func (s *server) HandleStream(stream network.Stream) {
...
 	return &server{
 		Service:    srv,
-		runtimeID:  runtimeID,
-		protocolID: pid,
-		logger: logging.GetLogger("worker/common/p2p/rpc/server").With(
-			"protocol", protocolID,
-			"runtime_id", runtimeID,
-		),
+		protocolID: protocolID,
+		logger:     logging.GetLogger("worker/common/p2p/rpc/server").With("protocol", protocolID),
```
kostko, line 110:
```suggestion
		logger:     logging.GetLogger("p2p/rpc/server").With("protocol", protocolID),
```

kostko (approved):

go/p2p/rpc/rpc.go
```diff
@@ -23,6 +23,11 @@ type P2P interface {
 	GetHost() core.Host
 }
 
+// NewProtocolID generates a protocol identifier for a consensus P2P protocol.
+func NewProtocolID(protocolID string, version version.Version) protocol.ID {
```
peternose, line 27: These two methods could be moved to `go/p2p/api/protocols.go`, but not sure.

peternose (approved):

PR #5002, opened 2022-10-28
go/p2p: Improve peer manager's peer handling

peternose:
Peer manager currently connects to all nodes in the registry which has multiple drawbacks:
- Number of connections should be controlled. On startup we currently connect to all validators (and reconnect if needed), that is 120 connections of 130 available by default.
- Connections get pruned if there are a lot of nodes in the registry. This can lead to unbalanced and uncontrolled protocol clusters, and can in extreme situations cut all node's connections to a protocol. Key managers and committees are protected with a tag, but others are not.
- Manager should connect only to peers which support protocols/topics the node is interested in.
- No support for peer discovery.

This should be changed so that peers are connected depending on the supported protocols and the number of peers in a protocol/topic should be regulated in a controlled fashion.

```
➜  compute-0 oasis-node control status
{
  "p2p": {
    "pub_key": "SMmEfIZU4HNlqtCepTEBtPC1pvN3LkE1IYDJj/ewve0=",
    "peer_id": "12D3KooWEiVkENN2J74dTijP8exfexvE3sJF6oJ4eGrHU857Zeex",
    "addresses": [
      "192.168.0.1:20012",
      "127.0.0.1:20012"
    ],
    "num_peers": 4,
    "num_connections": 6,
    "protocols": {
      "/oasis/keymanager/c000000000000000ffffffffffffffffffffffffffffffffffffffffffffffff/1.0.0": 1,
      "/oasis/storagesync/8000000000000000000000000000000000000000000000000000000000000000/1.0.0": 2,
      "/oasis/txsync/8000000000000000000000000000000000000000000000000000000000000000/1.0.0": 3
    },
    "topics": {
      "e48f1b7e23673c6eb1de959f6f2489930aa1fbf8f38b9f0990b7665d654fa314/4/8000000000000000000000000000000000000000000000000000000000000000/committee": 2,
      "e48f1b7e23673c6eb1de959f6f2489930aa1fbf8f38b9f0990b7665d654fa314/4/8000000000000000000000000000000000000000000000000000000000000000/tx": 3
    }
  }
}
```


go/p2p/peermgmt.go
```diff
@@ -3,423 +3,1240 @@ package p2p
...
+		}
 	}
+
+	if n.HasRoles(node.RoleStorageRPC) {
+		for _, rt := range n.Runtimes {
+			pid := rpc.NewRuntimeProtocolID(rt.ID, pub.StoragePubProtocolID, pub.StoragePubProtocolVersion)
+			m.addPeerToProtocol(info.ID, pid)
+		}
+		for _, rt := range n.Runtimes {
+			pid := rpc.NewRuntimeProtocolID(rt.ID, storageSync.StorageSyncProtocolID, storageSync.StorageSyncProtocolVersion)
```
kostko, line 1016: This should be moved up to compute nodes (all compute nodes are also storage nodes).

go/p2p/api/api.go
```diff
@@ -105,49 +111,32 @@ type Handler interface {
...
-	pubKey, err := publicKeyToPubKey(pk)
-	if err != nil {
-		return "", err
-	}
-
-	id, err := peer.IDFromPublicKey(pubKey)
-	if err != nil {
-		return "", err
-	}
+type PeerManager interface {
```
ptrus, line 114: Please add doc comment

go/p2p/api/api.go
```diff
@@ -105,49 +111,32 @@ type Handler interface {
...
+	// PeerRegistry returns the peer registry.
+	PeerRegistry() PeerRegistry
 
-	return id, nil
+	// PeerTagger returns the peer tagger.
+	PeerTagger() PeerTagger
 }
 
-const peerTagImportancePrefix = "oasis-core/importance"
+type PeerRegistry interface {
```
ptrus, line 122: Please add doc comment

go/p2p/api/api.go
```diff
@@ -105,49 +111,32 @@ type Handler interface {
...
 
-// TagValue returns the connection manager tag value associated with the given importance kind.
-func (ik ImportanceKind) TagValue() int {
-	switch ik {
-	case ImportantNodeCompute, ImportantNodeKeyManager:
-		return 1000
-	default:
-		panic(fmt.Errorf("unsupported tag: %d", ik))
-	}
+type PeerTagger interface {
```
ptrus, line 137: Please add doc comment

go/p2p/api/convert.go
```diff
@@ -0,0 +1,49 @@
...
+		id, err := PublicKeyToPeerID(pk)
+		if err != nil {
+			return nil, err
+		}
+		ids = append(ids, id)
+	}
+	return ids, nil
+}
+
+func NewTopicIDForRuntime(chainContext string, runtimeID common.Namespace, kind TopicKind) string {
```
ptrus, line 42: Please add doc comment

go/p2p/peermgmt/registry.go
```diff
@@ -0,0 +1,277 @@
...
+
+// Implements api.RegistryManager.
+func (r *peerRegistry) TopicPeers(topic string) []*peer.AddrInfo {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	tp := r.topicPeers[topic]
+	peers := make([]*peer.AddrInfo, 0, len(tp))
+	for _, peer := range tp {
+		peers = append(peers, peer)
```
ptrus, line 89: I think this should append a copy of `AddrInfo` and not the pointer directly.  Otherwise, `TopicPeers` is not thread-safe and the caller can also modify the value in the original `topicPeers`.

go/p2p/peermgmt/registry.go
```diff
@@ -0,0 +1,277 @@
...
+
+// Implements api.RegistryManager.
+func (r *peerRegistry) ProtocolPeers(p core.ProtocolID) []*peer.AddrInfo {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	pp := r.protocolPeers[p]
+	peers := make([]*peer.AddrInfo, 0, len(pp))
+	for _, peer := range pp {
+		peers = append(peers, peer)
```
ptrus, line 75: I think this should append a copy of `AddrInfo` and not the pointer directly.  Otherwise,  `ProtocolPeers` is not thread-safe and the caller can also modify the value in the original `protocolPeers`.

ptrus (commented):
Started review, will continue later

go/p2p/peermgmt/registry.go
```diff
@@ -0,0 +1,277 @@
...
+func (r *peerRegistry) inspectNode(n *node.Node) ([]core.ProtocolID, []string) {
+	// Protocol versions cannot be found in the registry so we assume all peers speak the latest
+	// protocols. The peer manager should handle this problem and block or negatively grade peers
+	// that are not up-to-date.
+	protocols := make([]core.ProtocolID, 0)
+	topics := make([]string, 0)
+
+	if n.HasRoles(node.RoleKeyManager) {
+		for _, rt := range n.Runtimes {
+			pid := rpc.NewRuntimeProtocolID(rt.ID, kmp2p.KeyManagerProtocolID, kmp2p.KeyManagerProtocolVersion)
```
kostko, line 225: Seems a bit backwards that the general P2P stack (in `go/p2p`) needs to import all of the higher-level P2P protocols for this discovery purpose? Would it be better to have some sort of global registry that would be populated during `init()` inside these higher-level protocols (e.g. the higher-level protocol would import this package and invoke some function to add a handler).

The handler would get passed a node and would return the protocols/topics (or, maybe better, would add protocols/topics to some existing map that would be passed).

go/p2p/peermgmt/registry.go
peternose: Copy won't work, as one can always mess with the addrs slice inside `AddrInfo`. As I don't want to copy these things, I decided to lowercase the method and keep it private. The package should treat it as read-only, as it already does.

go/p2p/peermgmt/peermgr.go
```diff
@@ -0,0 +1,375 @@
...
+	go func() {
+		defer wg.Done()
+		m.monitorPeers(ctx)
+	}()
+
+	<-m.stopCh
+}
+
+func (m *PeerManager) monitorPeers(ctx context.Context) {
+	var wg sync.WaitGroup
```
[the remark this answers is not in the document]
peternose: Leftovers.

go/p2p/peermgmt/registry.go
peternose: Agree totally. Fixed. I prefer not to pass maps as the handler could manipulate it or produce some kind of data races.

go/p2p/peermgmt/backup.go
```diff
@@ -0,0 +1,182 @@
...
+			return nil, err
+		}
+		infos[i] = &info
+	}
+
+	return infos, nil
+}
+
+// Backup periodically backups peers to the store.
+func (b *peerstoreBackup) Backup(ctx context.Context) {
```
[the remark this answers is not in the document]
peternose: > I would suggest renaming this to somehing like `StartBackups` (or just `Start`).

Agree, I chose `Run` as `Start` would expect `Stop`and to avoid stuttering. 

> a guard against spawning multiple backup routines

Agree, added a waiting line.

> start a backup worker in a separate gorutine

Yes, could start a backup worker but in this case I decided that all peer manager's workers, handlers, watchers, ... use a blocking `Run(ctx)` instead of a non-blocking pair `Start`/`Stop`. Not sure what is better.

> backup worker would automatically be started in `newPeerstoreBackup`

I would never do that (unless I'm forced) for multiple reasons:
- Constructor should set initial values if non-zero values don't suffice and nothing more.
- I want to have control of all go-routines that are started so that I can do a graceful shutdown.
- I don't want to hide go routines. If I spawn them everywhere it will be hard to figure out how to stop them.
- Harder to write tests.
- It's ugly, see `libp2p` library how `host` starts immediately without our control.

go/p2p/peermgmt/connector.go
```diff
@@ -0,0 +1,232 @@
...
+
+func (b *peerBackOff) reset() {
+	b.backoff.Reset()
+	b.next = time.Now()
+}
+
+type peerConnector struct {
+	logger *logging.Logger
+
+	mu sync.Mutex
```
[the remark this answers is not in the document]
peternose: Yes, it protects only those two maps (but actually the whole struct, as only those two fields are writable). In this case I prefer having a default name here (e.g. `l`, `mu`, `mx`, embedding ...). I moved the lock lower, though.

go/p2p/txsync/protocol.go
```diff
@@ -1,8 +1,13 @@
 package txsync
```
kostko, line 1: Unrelated to this PR, but I just noticed that `txsync` is defined in `go/p2p`, but it would probably make more sense for it to be moved under `go/worker/common/p2p` (similar to e.g. storage and key manager)?

go/p2p/peermgmt/backup.go
```diff
@@ -0,0 +1,194 @@
...
+
+	return infos, nil
+}
+
+// run periodically backups peers to the store. If a backup is already in progress, this method will
+// wait for it to finish.
+func (b *peerstoreBackup) run(ctx context.Context) {
+	// Allow to run only one backup at a time.
+	select {
+	case b.runCh <- struct{}{}:
```
kostko, line 143: This seems a bit weird given that `run` does not seem to return until context is cancelled which only happens during node shutdown? I think it makes more sense to make this a no-op if already running?

Also it seems weird to make this based on a context as this seems more like a service with a start/stop?

peternose: Changed to no-op if already running and switched to start/stop approach.

kostko (approved):

ptrus (approved):

PR #5007, opened 2022-10-29
go/p2p/rpc: Refactor rpc calls

peternose:
Peer selection is very tightly coupled with RPC client as peer manager is constructed inside client's constructor. We also miss simple RPC calls which contact exactly one peer.

### Test
Memory leak was check with the following code, which prints 1 and 101, meaning that 100 go routines leaked.

```
package main

import (
	"fmt"
	"runtime"
	"time"

	"github.com/eapache/channels"
)

func main() {
	fmt.Println(runtime.NumGoroutine())
	for i := 0; i < 100; i++ {
		leak()
	}
	time.Sleep(time.Minute)
	fmt.Println(runtime.NumGoroutine())
}

func leak() {
	n := 10
	var chs []channels.SimpleOutChannel

	for i := 0; i < n; i++ {
		ch := channels.NewNativeChannel(channels.BufferCap(1))
		defer close(ch)

		chs = append(chs, ch)
		ch <- i
	}

	ch := channels.NewNativeChannel(channels.None)
	channels.Multiplex(ch, chs...)

	for v := range ch.Out() {
		if v.(int) < n/2 {
			break
		}
	}
}
```

### DIFF
--- a/go/p2p/rpc/client.go
+++ b/go/p2p/rpc/client.go
@@ -23,10 +23,15 @@ import (
 const (
 	// RequestWriteDeadline is the maximum amount of time that can be spent on writing a request.
 	RequestWriteDeadline = 5 * time.Second
+	// RequestReadDeadline is the maximum amount of time that can be spent on reading a request.
+	RequestReadDeadline = 5 * time.Second
 	// DefaultCallRetryInterval is the default call retry interval for calls which explicitly enable
 	// retries by setting the WithMaxRetries option to a non-zero value. It can be overridden by
 	// using the WithRetryInterval call option.
 	DefaultCallRetryInterval = 1 * time.Second
+	// DefaultParallelRequests is the default number of parallel requests that can be mande
+	// when calling multiple peers.
+	DefaultParallelRequests = 5
 )
 
 // PeerFeedback is an interface for providing deferred peer feedback after an outcome is known.
@@ -93,15 +98,17 @@ type ValidationFunc func(pf PeerFeedback) error
 
 // CallOptions are per-call options.
 type CallOptions struct {
-	retryInterval time.Duration
-	maxRetries    uint64
-	validationFn  ValidationFunc
+	maxPeerResponseTime time.Duration
+	retryInterval       time.Duration
+	maxRetries          uint64
+	validationFn        ValidationFunc
 }
 
 // NewCallOptions creates options using default and given values.
 func NewCallOptions(opts ...CallOption) *CallOptions {
 	co := CallOptions{
-		retryInterval: DefaultCallRetryInterval,
+		maxPeerResponseTime: RequestReadDeadline,
+		retryInterval:       DefaultCallRetryInterval,
 	}
 	for _, opt := range opts {
 		opt(&co)
@@ -111,6 +118,13 @@ func NewCallOptions(opts ...CallOption) *CallOptions {
 
 // CallOption is a per-call option setter.
 type CallOption func(opts *CallOptions)
+
+// WithMaxPeerResponseTime configures the maximum response time for a peer.
+func WithMaxPeerResponseTime(d time.Duration) CallOption {
+	return func(opts *CallOptions) {
+		opts.maxPeerResponseTime = d
+	}
+}
 
 // WithMaxRetries configures the maximum number of retries to use for the call.
 func WithMaxRetries(maxRetries uint64) CallOption {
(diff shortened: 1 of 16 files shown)


go/p2p/rpc/client.go
```diff
@@ -374,37 +415,38 @@ func (c *client) CallMulti(
...
-	}
+			pf, err := c.timeCall(peerCtx, peer, &request, rsp, co.maxPeerResponseTime)
 
-	if len(resultChs) == 0 {
-		return nil, nil, nil
+			resultCh <- result{rsp, pf, err}
+		})
 	}
-	resultCh := channels.NewNativeChannel(channels.None)
-	channels.Multiplex(resultCh, resultChs...)
```
kostko, line 386: Hm, the below construct is not equivalent as it waits in order so it will always wait for the first peer to complete etc. The `channels.Multiplex` basically does one large `select` over all of the channels, so it will process results as they come in, regardless of whether the last peer in the list actually produced results first. This is important  when the aggregation function asks to break early and not wait for all of the peers.

peternose: Not sure if this is true as the channel is buffered/non blocking. Tested and I get a random order `[1 7 3 13 5]`.

```
package main

import (
	"context"
	"fmt"
	"math/rand"
	"time"

	"github.com/oasisprotocol/oasis-core/go/common/workerpool"
)

func main() {
	peers := []int{}
	for i := 0; i < 100; i++ {
		peers = append(peers, i)
	}

	pool := workerpool.New("p2p/rpc")
	pool.Resize(10)
	defer pool.Stop()

	peerCtx, cancel := context.WithCancel(context.Background())
	defer cancel()

	type result struct {
		rsp int
		err error
	}

	resultCh := make(chan result, len(peers))

	for _, peer := range peers {
		peer := peer

		pool.Submit(func() {
			select {
			case <-peerCtx.Done():
				return
			default:
			}

			var err error
			if peer%2 == 0 {
				err = fmt.Errorf("even peers fail")
			}

			time.Sleep(time.Duration(rand.Intn(1000)) * time.Millisecond)

			resultCh <- result{peer, err}
		})
	}

	var rsps []int

loop:
	for i := 0; i < len(peers); i++ {
		select {
		case result := <-resultCh:
			if result.err != nil {
				break
			}

			rsps = append(rsps, result.rsp)

			if len(rsps) == 5 {
				break loop // agregate function
			}

		case <-peerCtx.Done():
			break loop
		}
	}

	fmt.Printf("%+v\n", rsps)
}
```

kostko: Imagine the following scenario:
* 10 peers in total.
* First 5 peers take 30 seconds to complete the request.
* Last 5 peers take 1 second to complete the request.
* Aggregation function signals to break after getting response from 5 peers.

In this case `CallMulti` should not wait 30 seconds to return as it would have the 5 results after roughly 1 second.

Based on how `Multiplex` is implemented (a `select` over the result channels of all peers at once) it should do the right thing while the new implementation would wait for result from the first peer, then the second, etc.

kostko: Ah sorry for some reason I thought the result channel was per peer, you're right this looks ok.

go/p2p/rpc/client.go
```diff
@@ -168,101 +157,168 @@ type AggregateFunc func(rsp interface{}, pf PeerFeedback) bool
 
 // CallMultiOptions are per-multicall options
```
kostko, line 158:
```suggestion
// CallMultiOptions are per-multicall options.
```

go/p2p/rpc/client.go
```diff
@@ -168,101 +157,168 @@ type AggregateFunc func(rsp interface{}, pf PeerFeedback) bool
...
+	// when routing calls.
+	CallOne(
+		ctx context.Context,
+		peers []peer.ID,
+		method string,
+		body, rsp interface{},
+		opts ...CallOption,
+	) (PeerFeedback, error)
+
+	// CallMulti routes the given RPC method call to multiple peers in the list in
```
kostko, line 244:
```suggestion
	// CallMulti routes the given RPC method call to multiple (possibly all) peers in the list in
```

go/p2p/rpc/client.go
```diff
@@ -168,101 +157,168 @@ type AggregateFunc func(rsp interface{}, pf PeerFeedback) bool
...
-
-	// Call attempts to route the given RPC method call to one of the peers that supports the
-	// protocol based on past experience with the peers.
+	// Call attempts to route the given RPC method call to the given peer. It's up to the caller
+	// to provide only connected peers that support the protocol.
 	//
 	// On success it returns a PeerFeedback instance that should be used by the caller to provide
 	// deferred feedback on whether the peer is any good or not. This will help guide later choices
 	// when routing calls.
 	Call(
```
kostko, line 221: What is the difference between `Call` and `CallOne` with only a single peer specified? Should we just get rid of this one?

kostko (commented):
Can you also backport the memory leak and peer scoring fixes?

go/p2p/rpc/client.go
peternose: I wanted to get rid of `validationFn` and didn't want to have in the logs that calls to all peers failed. But now that I look back it is better to have only one method. 
Actually, I still prefer having two methods, I will just change that `Call` calls `CallOne`.

kostko (approved):

PR #5008, opened 2022-10-29
go/common/workerpool: Fix memory leak when workerpool is stopped early

peternose:
When workerpool si stopped, the job channel might still contain jobs which haven't been processed. Therefore, the channel never closes and leaves one go routine hanging.

### Test
Tested with the following code. Before the code printed 1/101, afterwards 1/1.

```
package main

import (
	"fmt"
	"runtime"
	"time"

	"github.com/oasisprotocol/oasis-core/go/common/workerpool"
)

func main() {
	fmt.Println(runtime.NumGoroutine())
	for i := 0; i < 100; i++ {
		leak()
	}
	time.Sleep(time.Minute)
	fmt.Println(runtime.NumGoroutine())
}

func leak() {
	pool := workerpool.New("p2p/rpc")
	pool.Resize(10)
	defer pool.Stop()

	for i := 0; i < 100; i++ {
		pool.Submit(func() {
			time.Sleep(time.Millisecond)
		})
	}
}
```

### DIFF
--- a/.changelog/5008.bugfix.md
+++ b/.changelog/5008.bugfix.md
@@ -1,0 +1,5 @@
+go/common/workerpool: Fix memory leak when workerpool is stopped early
+
+When workerpool si stopped, the job channel might still contain jobs which
+haven't been processed. Therefore, the channel never closes and leaves one
+go routine hanging.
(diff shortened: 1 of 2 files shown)


kostko (approved):

kostko:
Please also backport.

PR #5011, opened 2022-10-31
go tendermint staking: reduce DelegationsTo scanning

pro-wh:
the escrow address is the first part of the key so we can seek to it


kostko (approved):

kostko:
Also consider backporting.

Yawning (approved):

PR #5022, opened 2022-11-08
go/staking: reduce DebondingDelegationsFor scanning

ptrus:
`DebondingDelegationsFor` keys are ordered by `delegatorAddr`. Once past it, it doesn't need to scan further (analogous to `DelegationsTo` fixed in: https://github.com/oasisprotocol/oasis-core/pull/5011).

Seek query is already (correctly) prefixed with `delegatorAddr`.


pro-wh (approved):
agreed

PR #5034, opened 2022-11-10
go/governance: Support delegator votes

ptrus:

PR #5050, opened 2022-11-14
go/scheduler: include entity IDs in consensus validators state

ptrus:
#5034 will need an efficient way to obtain a list of active validator entities


go/scheduler/api/api.go
```diff
@@ -208,6 +208,9 @@ type Validator struct {
 	// ID is the validator Oasis node identifier.
 	ID signature.PublicKey `json:"id"`
 
+	// EntityID is the validator entity Oasis node identifier.
```
kostko, line 211:
```suggestion
	// EntityID is the validator entity identifier.
```

kostko (approved):

PR #5055, opened 2022-11-15
go: always return `ErrInvalidArgument` on tx deserialization failure

ptrus:
Currently, some modules are returning non-specific errros on tx deserialization failures which results in them being assigned to module "unknown", code 1. It would be more correct to always return the corresponding ErrInvalidArgument error for all deserialization failures.

### DIFF
--- a/go/keymanager/api/api.go
+++ b/go/keymanager/api/api.go
@@ -27,9 +27,12 @@ const (
 )
 
 var (
+	// ErrInvalidArgument is the error returned on malformed arguments.
+	ErrInvalidArgument = errors.New(ModuleName, 1, "keymanager: invalid argument")
+
 	// ErrNoSuchStatus is the error returned when a key manager status does not
 	// exist.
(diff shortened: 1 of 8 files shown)


go/keymanager/api/api.go
```diff
@@ -27,9 +27,12 @@ const (
 )
 
 var (
+	// ErrInvalidArgument is the error returned on malformed arguments.
+	ErrInvalidArgument = errors.New(ModuleName, 1, "keymanager: invalid argument")
+
 	// ErrNoSuchStatus is the error returned when a key manager status does not
 	// exist.
-	ErrNoSuchStatus = errors.New(ModuleName, 1, "keymanager: no such status")
+	ErrNoSuchStatus = errors.New(ModuleName, 2, "keymanager: no such status")
```
ptrus, line 35: Adding `ErrInvalidArgument` with not `1` as the error code would be a breaking change anyway, so I updated this one to `2`. That way code `1` is consistently used for `ErrInvalidArgument` errors.

kostko (approved):

PR #4571, opened 2022-03-16
Add archive mode support

ptrus:
Closes https://github.com/oasisprotocol/oasis-core/issues/4539

TODO:
- [x] include runtimes in the archive_api e2e test
- [x] also test starting a node in archive mode after reaching halt or upgrade epoch
- [x] deduplicate some of the duplicated code between archive and full node (`archive.go` and `full.go`)
- [x] update instructions to enable archive mode to the changelog fragment
- [x] ~wait for https://github.com/oasisprotocol/oasis-core/pull/4427 to be merged and rebase~

PR #5070, opened 2022-11-22
Config rework

abukosek:
This is still in heavy WIP.

- [X] Pass 1: Convert configuration cmdline flags into structs, make a global config struct, use that everywhere, add loading from YAML config file.
- [x] Pass 2: Make the test runner and other non-node parts use the new config structs.
- [x] Pass 3: Make everything work.


kostko (commented):
Instead of using a global config, can we pass it explicitly?

abukosek:
> Instead of using a global config, can we pass it explicitly?

Hmm, that seems like a lot of extra work.  I'll first finish the current design and then we can discuss changing it.

A lot of modules need access to config options outside of their module, so we'd have to pass the entire global config struct to all modules anyway.  And currently, only a few modules have some kind of init function where we could pass the config in explicitly, so all the others would have to be refactored to include something like that.  I think the current global struct design is cleaner (load once, use anywhere).

gw0:
*Idea:* In order to still support "re-usable" config files (after the removal of cmdline flags) it would be great to support some variable substitution before the config file gets parsed. Could be as trivial as replacing e.g. `${EXTERNAL_IP}` or `{{ EXTERNAL_IP }}` with value of env variable `EXTERNAL_IP` or the value read from a `.env` file. Or a bit more powerful, like in `docker-compose` (https://docs.docker.com/compose/environment-variables/).

abukosek:
> _Idea:_ In order to still support "re-usable" config files (after the removal of cmdline flags) it would be great to support some variable substitution before the config file gets parsed. Could be as trivial as replacing e.g. `${EXTERNAL_IP}` or `{{ EXTERNAL_IP }}` with value of env variable `EXTERNAL_IP` or the value read from a `.env` file. Or a bit more powerful, like in `docker-compose` (https://docs.docker.com/compose/environment-variables/).

I don't think that the `yaml` package we're using to parse the config file supports this.  This looks like something that should be done with external templating tools and not in oasis-core :)

ptrus:
I think being able to set/override config values via environment variables should be possible, unless there is some hard reasoning not to support it.

One option would be using something like [koanf](https://github.com/knadh/koanf) (we use it in https://github.com/oasisprotocol/emerald-web3-gateway). It makes it really easy to use a `Config` struct (as defined in this PR) and automatically support config fields to be overridden via env variables, see example: https://github.com/oasisprotocol/emerald-web3-gateway/blob/a60756a8f1fb05cecd1d95bf31940a32812c9d8a/conf/config.go#L226-L255  (this is technically also possible to do with viper, but it has some issues like: https://github.com/spf13/viper/issues/368 so I'd suggest avoiding it).

gw0:
> I don't think that the `yaml` package we're using to parse the config file supports this. This looks like something that should be done with external templating tools and not in oasis-core :)

It does not, but setting env variables is convenient and adding support (https://github.com/a8m/envsubst) is as simple as:
```go
import "github.com/a8m/envsubst"
data, err := envsubst.ReadFile("config.yaml")
err = yaml.Unmarshal(data, ...)
```

abukosek:
Alright, I'll add that once I'm done with the test runner :)

kostko (commented):
Some initial comments.

go/common/config/config.go
```diff
@@ -0,0 +1,143 @@
...
+
+// GlobalConfig holds the global configuration options.
+var GlobalConfig Config
+
+// Config is the top-level configuration structure.
+type Config struct {
+	// Oasis node mode (validator, non-validator, compute, keymanager, etc.).
+	Mode string `yaml:"mode"`
+	// Node is a consensus validator.
+	ConsensusValidator bool `yaml:"consensus_validator"`
```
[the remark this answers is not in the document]
abukosek: No idea why we have a separate option for this, I just translated the flag that was already defined before: :man_shrugging:
https://github.com/oasisprotocol/oasis-core/blob/9e12c44f88a62198d25535c489964a2192f8e98d/go/oasis-node/cmd/common/flags/flags.go#L22-L23

go/common/config/util.go
```diff
@@ -0,0 +1,53 @@
...
+	"strings"
+	"unicode"
+
+	"github.com/spf13/cast"
+)
+
+// ParseSizeInBytes converts strings like 1GB or 12 mb into an unsigned integer number of bytes.
+// Note: This function was shamelessly lifted from viper:
+// https://github.com/spf13/viper/blob/master/util.go
+func ParseSizeInBytes(sizeStr string) uint {
```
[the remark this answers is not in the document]
abukosek: Maybe, I'll have a look.

go/consensus/tendermint/config/config.go
```diff
@@ -0,0 +1,170 @@
...
+	ListenAddress string `yaml:"listen_address"`
+
+	// Max number of inbound peers.
+	P2PMaxNumInboundPeers int `yaml:"p2p_max_num_inbound_peers"`
+	// Max number of outbound peers (excluding persistent peers).
+	P2PMaxNumOutboundPeers int `yaml:"p2p_max_num_outbound_peers"`
+	// Rate at which P2P packets can be sent (bytes/s).
+	P2PSendRate int64 `yaml:"p2p_send_rate"`
+	// Rate at which P2P packets can be received (bytes/s).
+	P2PRecvRate int64 `yaml:"p2p_recv_rate"`
```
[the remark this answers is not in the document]
abukosek: Yeah, I have a comment about that in the P2P code...  These are just translated from the original code.  I'll change them to unsigned.

go/consensus/tendermint/config/config.go
```diff
@@ -0,0 +1,170 @@
...
+type Config struct {
+	// Tendermint mode (full, archive).
+	Mode string `yaml:"mode"`
+
+	// Tendermint address advertised to other nodes.
+	ExternalAddress string `yaml:"external_address"`
+	// Tendermint listen address.
+	ListenAddress string `yaml:"listen_address"`
+
+	// Max number of inbound peers.
```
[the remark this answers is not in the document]
abukosek: Oops, I flattened these a bit, will move them into separate structs (also elsewhere).

go/oasis-node/cmd/node/run.go
```diff
@@ -39,7 +39,7 @@ const (
 )
 
 func Mode() string {
```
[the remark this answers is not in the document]
abukosek: Good idea, I'll do that :)

go/consensus/tendermint/config/config.go
```diff
@@ -0,0 +1,170 @@
...
+
+import (
+	"fmt"
+	"time"
+)
+
+// Config is the Tendermint configuration structure.
+type Config struct {
+	// Tendermint mode (full, archive).
+	Mode string `yaml:"mode"`
```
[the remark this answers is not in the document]
abukosek: Yes, after I move the type definitions for these modes into the config package, I'll change this into a proper type.

go/consensus/tendermint/config/config.go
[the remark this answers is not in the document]
abukosek: Oh, I see why these are signed... because tendermint also uses signed types for these :man_facepalming:
We just pass them directly, so in this case I guess it's better to leave them signed in this config and just check that they're >= 0, like I've done in `Validate()`.

go/common/config/util.go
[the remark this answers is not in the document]
abukosek: It doesn't look like the YAML package we're using supports that.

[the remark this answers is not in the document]
kostko: Hm, could you define a custom type that wraps uint64 and defines a custom unmarshaller that handles this?

go/consensus/api/mode/mode.go
```diff
@@ -0,0 +1,51 @@
...
+import "fmt"
+
+// Mode is the consensus node mode.
+type Mode string
+
+const (
+	// ModeFull is the name of the full node consensus mode.
+	ModeFull Mode = "full"
+	// ModeArchive is the name of the archive node consensus mode.
+	ModeArchive Mode = "archive"
```
kostko, line 13: Since we are here, why is this needed in addition to the global node mode? Could it be removed?

go/oasis-node/cmd/common/config/config.go
```diff
@@ -0,0 +1,47 @@
...
+			Level:  make(map[string]string),
+		},
+		Debug: DebugConfig{
+			AllowRoot: false,
+			Rlimit:    0,
+		},
+	}
+}
+
+type DebugConfig struct {
```
kostko, line 33: Add missing doc comments.

kostko (commented):
Can you also make an overview of what configuration options have moved?

go/runtime/registry/config/config.go
```diff
@@ -0,0 +1,222 @@
+// Package config implements global configuration options.
+package config
+
+import (
+	"fmt"
+	"time"
+)
+
+// RuntimeMode is the runtime mode.
+type RuntimeMode string
```
kostko, line 10: This should now be removed in favor of the top-level node mode which already covers all the needed options.

go/runtime/registry/config/config.go
```diff
@@ -0,0 +1,222 @@
...
+)
+
+// Config is the runtime registry configuration structure.
+type Config struct {
+	// Runtime mode (none, compute, keymanager, client, client-stateless).
+	Mode RuntimeMode `yaml:"mode"`
+
+	// Runtime provisioner to use (mock, unconfined, sandboxed).
+	Provisioner RuntimeProvisioner `yaml:"provisioner"`
+	// Paths to runtime resources (format: <path>,<path>,...).
```
kostko, line 135:
```suggestion
	// Paths to runtime bundles.
```

go/worker/common/config/config.go
```diff
@@ -0,0 +1,43 @@
+// Package config implements global configuration options.
+package config
+
+// Config is the common worker configuration structure.
+type Config struct {
+	// Port to use for incoming gRPC client connections.
```
kostko, line 6: Is this still used anywhere? AFAIK everything has migrated to P2P?

go/runtime/registry/registry.go
```diff
@@ -41,7 +43,7 @@ var ErrRuntimeHostNotConfigured = errors.New("runtime/registry: runtime host not
 // Registry is the running node's runtime registry interface.
 type Registry interface {
 	// Mode returns the configured behavior of runtime workers on this node.
-	Mode() RuntimeMode
+	Mode() regConfig.RuntimeMode
```
kostko, line 46: This should be removed in favor of referencing the node configuration.

go/runtime/registry/registry.go
```diff
@@ -70,7 +72,7 @@ type Runtime interface {
 	ID() common.Namespace
 
 	// Mode returns the configured behavior of runtime workers on this node.
-	Mode() RuntimeMode
+	Mode() regConfig.RuntimeMode
```
kostko, line 75: This should be removed in favor of referencing the node configuration.

go/runtime/registry/config.go
```diff
@@ -67,95 +37,10 @@ const (
...
-		return true
-	}
-	return false
-}
-
 // RuntimeConfig is the node runtime configuration.
 type RuntimeConfig struct {
 	// Mode is the runtime mode for this node.
-	Mode RuntimeMode
+	Mode regConfig.RuntimeMode
```
kostko, line 43: This is probably no longer needed now?

go/config/config.go
```diff
@@ -0,0 +1,168 @@
...
+	// ModeValidator is the name of the validator node mode.
+	ModeValidator NodeMode = "validator"
+	// ModeNonValidator is the name of the non-validator node mode.
+	ModeNonValidator NodeMode = "non-validator"
+	// ModeCompute is the name of the compute node mode.
+	ModeCompute NodeMode = "compute"
+	// ModeKeyManager is the name of the key manager node mode.
+	ModeKeyManager NodeMode = "keymanager"
+	// ModeClient is the name of the client node mode.
+	ModeClient NodeMode = "client"
```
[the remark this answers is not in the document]
abukosek: No idea, we should ask the person who added this mode :)
To me it looks like we could remove it.

go/worker/common/config/config.go
abukosek: It is still referenced in `GetNodeAddresses()` in `worker/common/config.go`, but that function does not appear to be used anywhere, so I think it's safe to remove.

go/config/config.go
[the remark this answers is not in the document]
kostko: Yeah let's remove it. cc @peternose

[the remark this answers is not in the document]
kostko: Oh to be clear, I would remove `non-validator` and leave `client`.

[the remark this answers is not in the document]
abukosek: Yeah, I was also thinking about removing `non-validator` :)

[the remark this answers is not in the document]
peternose: I agree to remove `non-validator`.

go/consensus/api/mode/mode.go
abukosek: I've removed this, but it appears that the archive API test wants to have a node that's both a compute and an archive node...  What is the use case for this?  Do we need to keep this functionality or is it OK to remove?

kostko: Huh, which test wants that? It doesn't make any sense.

kostko: I only see a compute node that is _turned into_ an archive node. So it is not both at the same time.

abukosek: The test code seems to have a notion of a "consensus archive node" and a "runtime archive node" -- but an archive node as defined in the main node mode cannot have runtimes.  So somebody added this notion of a "runtime archive node", but why?

kostko: Every archive node should be a runtime archive node if it has any runtimes defined. cc @ptrus

ptrus: I think that this duplication was done in: https://github.com/oasisprotocol/oasis-core/pull/4964

Since originally, I see only the mode defined under consensus (https://github.com/oasisprotocol/oasis-core/pull/4571).

Feel free to assign me on the issue for fixing/cleaning this up if noone else will.

kostko: If it's not too bad we should try to fix it in this PR.

abukosek: It's OK, I've already fixed this in this PR :)

go/oasis-test-runner/oasis/oasis.go
```diff
@@ -179,15 +190,80 @@ func (n *Node) LoadIdentity() (*identity.Identity, error) {
 	return identity.Load(n.dir.String(), factory)
 }
 
+func (n *Node) AddSeedNodesToConfig() {
```
kostko, line 193: Add documentation comments.

go/oasis-test-runner/scenario/e2e/registry_cli.go
```diff
@@ -442,7 +458,7 @@ func (sc *registryCLIImpl) initNode(childEnv *env.Env, ent *entity.Entity, entDi
 			"--" + cmdRegNode.CfgNodeRuntimeID, testNode.Runtimes[0].ID.String(),
 			"--" + cmdSigner.CfgSigner, fileSigner.SignerName,
 			"--" + cmdSigner.CfgCLISignerDir, entDir,
-			"--" + cmdCommon.CfgDataDir, dataDir,
+			"--" + cmdCommon.CfgConfigFile, cfgFile,
```
kostko, line 461: Why is a config file needed for `registry node init`?

go/oasis-test-runner/scenario/e2e/registry_cli.go
```diff
@@ -338,12 +341,12 @@ func (sc *registryCLIImpl) listNodes(childEnv *env.Env) ([]signature.PublicKey,
...
 
 // isRegistered checks if the given node is registered.
-func (sc *registryCLIImpl) isRegistered(childEnv *env.Env, nodeName, nodeDataDir string) error {
+func (sc *registryCLIImpl) isRegistered(childEnv *env.Env, nodeName, nodeCfgFile string) error {
 	sc.Logger.Info(fmt.Sprintf("checking if node %s is registered", nodeName))
 	args := []string{
 		"registry", "node", "is-registered",
 		"--" + grpc.CfgAddress, "unix:" + sc.Net.Validators()[0].SocketPath(),
-		"--" + cmdCommon.CfgDataDir, nodeDataDir,
+		"--" + cmdCommon.CfgConfigFile, nodeCfgFile,
```
kostko, line 349: Why is a config file needed for `registry node is-registered`?

go/p2p/config/config.go
```diff
@@ -0,0 +1,112 @@
+// Package config implements global configuration options.
+package config
+
+import "time"
+
+// Config is the P2P configuration structure.
+type Config struct {
+	// Port to use for incoming P2P connections.
+	HostPort uint16 `yaml:"port"`
```
kostko, line 9: Why not just `Port`?

go/p2p/host.go
```diff
@@ -102,6 +102,7 @@ func (cfg *HostConfig) Load() error {
 
 // ConnManagerConfig describes a set of settings for a connection manager.
 type ConnManagerConfig struct {
+	// XXX: MinPeers in main config is uint32, but int here -- investigate.
```
kostko, line 105: Does this need investigation?

go/p2p/p2p.go
```diff
@@ -488,6 +488,7 @@ func (cfg *Config) Load() error {
 
 // GossipSubConfig describes a set of settings for a gossip pubsub.
 type GossipSubConfig struct {
+	// XXX: Main config has int64, but here just int -- investigate.
```
kostko, line 491: Does this need investigation?

go/runtime/registry/config/config.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+// RuntimeProvisioner is the runtime provisioner.
+type RuntimeProvisioner string
+
+const (
+	// RuntimeProvisionerMock is the name of the mock runtime provisioner.
+	//
+	// Use of this provisioner is only allowed if DebugDontBlameOasis flag is set.
+	RuntimeProvisionerMock = "mock"
```
kostko, line 16: FYI, this is not actually using the `RuntimeProvisioner` type (also below).

go/runtime/registry/config/config.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	return nil
+}
+
+// RuntimeEnvironment is the runtime environment.
+type RuntimeEnvironment string
+
+const (
+	// RuntimeEnvironmentSGX specifies to run the runtime in SGX.
+	RuntimeEnvironmentSGX = "sgx"
```
kostko, line 51: FYI, this is not actually using the `RuntimeEnvironment` type (also below).

go/runtime/registry/config/config.go
```diff
@@ -0,0 +1,142 @@
...
+type HistoryPrunerConfig struct {
+	// History pruner strategy.
+	Strategy string `yaml:"strategy"`
+	// History pruning interval.
+	Interval time.Duration `yaml:"interval"`
+	// Number of last rounds to keep.
+	NumKept uint64 `yaml:"num_kept"`
+}
+
+func (c *Config) Validate() error {
```
kostko, line 92: Add documentation comments (also elsewhere).

go/runtime/registry/registry.go
```diff
@@ -599,12 +584,12 @@ func New(ctx context.Context, dataDir string, consensus consensus.Backend, ias i
...
 
-	switch cfg.Mode {
-	case RuntimeModeNone:
+	switch config.GlobalConfig.Mode {
+	case config.ModeValidator, config.ModeSeed:
 		r.logger.Info("runtime support is disabled")
 	default:
 		r.logger.Info("runtime support is enabled",
-			"mode", cfg.Mode,
+			"mode", config.GlobalConfig.Mode,
```
kostko, line 592: Given that this is now global, it should probably be logged during node startup and removed from here (this entire `switch`).

go/worker/common/config.go
```diff
@@ -75,32 +32,19 @@ func NewConfig() (*Config, error) {
 	}
 
 	cfg := Config{
-		ClientPort:      uint16(viper.GetInt(CfgClientPort)),
 		SentryAddresses: sentryAddresses,
 		TxPool: txpool.Config{
```
kostko, line 36: Should we just use `txpool.Config` directly in the config (could move it in a subpackage if needed)? It doesn't make much sense to have this copying here now that we are using config structures anyway?

go/config/config.go
```diff
@@ -0,0 +1,184 @@
...
+
+// Config is the top-level configuration structure.
+type Config struct {
+	// Oasis node mode (validator, non-validator, compute, keymanager, etc.).
+	Mode NodeMode `yaml:"mode"`
+
+	// Node is a consensus validator.
+	// This additional option exists because it is currently possible to
+	// run a node that is simultaneously a validator and a compute node.
+	ConsensusValidator bool `yaml:"consensus_validator"`
```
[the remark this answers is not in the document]
abukosek: Do we anticipate other options under the `consensus` hierarchy in the future?  I flattened this because I think it's ugly to have just a single option in a struct (the option itself might as well have been named just `validator`).
I can add an extra struct just for this, but since this is a nice opportunity to make the config options nicer, maybe we can keep it as it is now?

go/oasis-test-runner/scenario/e2e/registry_cli.go
abukosek: It needs the data dir from it and I figured it would be nicer for the users if they didn't have to retype the data dir on the command line, but instead just pass their config file (same for `registry node is-registered`).

go/p2p/config/config.go
abukosek: `Port` makes more sense, `HostPort` is what the original author named it.

go/p2p/host.go
abukosek: Yes (same for other type mismatches in the new p2p stack).

I imagine that the underlying library we're using has the signed types -- in this case, we should also change the config to use signed types and check that they don't overflow, since it's safer than blindly casting an uint to an int.

go/runtime/registry/config/config.go
abukosek: Yeah, this is a copy&paste of the existing registry code -- I'll fix it :)

go/worker/common/config.go
abukosek: Yeah, that makes sense.

go/p2p/host.go
abukosek: I've modified this as described above.

go/worker/common/config.go
abukosek: I've changed txpool to use the structure defined in the config.

.changelog/5070.cfg.md
```diff
@@ -0,0 +1,105 @@
...
+
+- `consensus.tendermint.mode` in favor of using the global node mode (`mode`).
+
+- `runtime.mode` in favor of using the global node mode (`mode`).
+
+- `worker.client.port` as it is no longer needed.
+
+- `worker.registration.force_register` as it is deprecated.
+
+The following options have been renamed:
```
kostko, line 22: Given that there have been a lot of changes, can we also make a migration subcommand? So that we can include that in the upgrade instructions.

.changelog/5070.cfg.md
```diff
@@ -0,0 +1,105 @@
...
+
+- `ias.debug.skip_verify` to `ias.debug_skip_verify`.
+
+- `runtime.sandbox.binary` to `runtime.sandbox_binary`.
+
+- `runtime.sgx.loader` to `runtime.sgx_loader`.
+
+- `runtime.history.pruner.*` to `runtime.history_pruner.*`.
+
+- `worker.sentry.addresses` to `workers.sentry_addresses`.
```
kostko, line 83: I would change `worker.*` to `runtime.*`.

.changelog/5070.cfg.md
```diff
@@ -0,0 +1,105 @@
...
+
+- `debug.rlimit` to `common.debug.rlimit`.
+
+- `debug.allow_root` to `common.debug.allow_root`.
+
+- `pprof.bind` to `pprof.bind_address`.
+
+- `consensus.validator` to `consensus_validator`.
+
+- `consensus.tendermint.*` to `tendermint.*`.
```
kostko, line 41: Could we just drop this `tendermint` thing (especially since it will be renamed upstream) from the config and have everything under `consensus`?

.changelog/5070.cfg.md
```diff
@@ -0,0 +1,105 @@
...
+- `log.level` to `common.log.level`.  The log level is now a map of module
+to log level.  Use the `default` module to map the default log level.
+
+- `debug.rlimit` to `common.debug.rlimit`.
+
+- `debug.allow_root` to `common.debug.allow_root`.
+
+- `pprof.bind` to `pprof.bind_address`.
+
+- `consensus.validator` to `consensus_validator`.
```
kostko, line 39: Change this back.

go/config/config.go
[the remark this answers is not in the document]
kostko: See my other comment, I think we should put everything under `consensus` and drop the `tendermint` thing.

.changelog/5070.cfg.md
abukosek: As agreed on Slack, we're going to make this in a separate PR later.

.changelog/5070.cfg.md
```diff
@@ -0,0 +1,103 @@
...
+
+- `ias.debug.skip_verify` to `ias.debug_skip_verify`.
+
+- `runtime.sandbox.binary` to `runtime.sandbox_binary`.
+
+- `runtime.sgx.loader` to `runtime.sgx_loader`.
+
+- `runtime.history.pruner.*` to `runtime.history_pruner.*`.
+
+- `worker.sentry.addresses` to `runtime.sentry_addresses`.
```
kostko, line 81: Why is this not moved together with other sentry changes? I think it should just go under `sentry.addresses`?

.changelog/5070.cfg.md
```diff
@@ -0,0 +1,103 @@
...
+- `worker.registration.*` to `registration.*`.
+
+- `worker.sentry.*` to `sentry.*`.
+
+- `worker.sentry.control.authorized_pubkey` to
+`sentry.control.authorized_pubkeys`.
+
+- `worker.storage.*` to `storage.*`.
+
+- `worker.storage.public_rpc.enabled` to `worker.storage.public_rpc_enabled`.
```
kostko, line 98:
```suggestion
- `worker.storage.public_rpc.enabled` to `storage.public_rpc_enabled`.
```

.changelog/5070.cfg.md
```diff
@@ -0,0 +1,103 @@
...
+
+- `worker.storage.checkpoint_sync.disabled` to
+`worker.storage.checkpoint_sync_disabled`.
```
kostko, line 101:
```suggestion
`storage.checkpoint_sync_disabled`.
```

.changelog/5070.cfg.md
```diff
@@ -0,0 +1,103 @@
...
+
+- `worker.storage.checkpointer.*` to `storage.checkpointer.*`.
```
kostko, line 103: This is probably covered by `worker.storage.*` to `storage.*`?

.changelog/5070.cfg.md
```diff
@@ -0,0 +1,103 @@
...
+- `consensus.tendermint.checkpointer.*` to `consensus.abci.checkpointer.*`.
+
+- `consensus.tendermint.sentry.upstream_address` to
+`consensus.sentry_upstream_addresses`.
+
+- `consensus.tendermint.upgrade.stop_delay` to
+`consensus.upgrade_stop_delay`.
+
+- `consensus.tendermint.supplementarysanity.*` to
+`consensus.supplementary_sanity.`.
```
kostko, line 63:
```suggestion
`consensus.supplementary_sanity.*`.
```

go/worker/registration/worker.go
```diff
@@ -1128,15 +1112,9 @@ func New(
...
-	if viper.GetBool(CfgRegistrationForceRegister) {
-		if err = SetForcedDeregister(w.store, false); err != nil {
-			return nil, fmt.Errorf("failed to clear persisted forced-deregister: %w", err)
-		}
-		storedDeregister = false
-	}
 	w.storedDeregister = storedDeregister
 
-	if flags.ConsensusValidator() {
+	if config.GlobalConfig.Consensus.Validator {
```
kostko, line 1117: This should also work if you set the global mode to validator?

go/consensus/tendermint/config/config.go
```diff
@@ -0,0 +1,228 @@
...
+	ABCI ABCIConfig `yaml:"abci,omitempty"`
+
+	// Consensus state sync configuration.
+	StateSync StateSyncConfig `yaml:"state_sync,omitempty"`
+
+	// Supplementary sanity checks configuration.
+	SupplementarySanity SupplementarySanityConfig `yaml:"supplementary_sanity,omitempty"`
+
+	// Enable tendermint debug logs (very verbose).
+	LogDebug bool `yaml:"log_debug,omitempty"`
```
kostko, line 48: Should this go under the debug options?

go/oasis-node/cmd/node/node.go
```diff
@@ -453,6 +454,7 @@ func NewNode() (node *Node, err error) { // nolint: gocyclo
 	// binary is from the logs.
 	logger.Info("Starting oasis-node",
 		"Version", version.SoftwareVersion,
+		"Mode", config.GlobalConfig.Mode,
```
kostko, line 457:
```suggestion
		"version", version.SoftwareVersion,
		"mode", config.GlobalConfig.Mode,
```

go/runtime/txpool/config/config.go
```diff
@@ -0,0 +1,18 @@
+// Package config implements the txpool configuration options.
+package config
+
+import "time"
+
+// TxPoolConfig is the common worker tx pool configuration structure.
+type TxPoolConfig struct {
```
kostko, line 7: Why not just `Config`?

go/runtime/txpool/config/config.go
```diff
@@ -0,0 +1,18 @@
+// Package config implements the txpool configuration options.
+package config
+
+import "time"
+
+// TxPoolConfig is the common worker tx pool configuration structure.
```
kostko, line 6:
```suggestion
// TxPoolConfig is the runtime transaction pool configuration structure.
```

go/worker/registration/worker.go
abukosek: The previous code only checked the flag:
https://github.com/oasisprotocol/oasis-core/blob/master/go/oasis-node/cmd/common/flags/flags.go#L75-L77

go/consensus/tendermint/config/config.go
abukosek: Hmm, it kind of makes sense to leave it here because it only sets the Tendermint log level to debug and isn't some unsupported debug flag.

go/worker/registration/worker.go
kostko: Yeah this was wrong, we should make sure that if you set the global mode to validator your node will actually become a validator :-)

kostko (approved):

PR #5075, opened 2022-11-29
runtime/src/enclave_rpc: Support insecure key manager RPC requests

peternose:
Create a new simpler, session-less, RPC request/response protocol to be used when dealing with public keys.

### DIFF
--- a/go/runtime/enclaverpc/api/api.go
+++ b/go/runtime/enclaverpc/api/api.go
@@ -1,6 +1,16 @@
 // Package api defines the EnclaveRPC interface.
 package api
 
+// Kind is the RPC call kind.
+type Kind uint8
+
+// Supported RPC call kinds.
+const (
+	KindNoiseSession  Kind = 0
+	KindInsecureQuery Kind = 1
+	KindLocalQuery    Kind = 2
+)
+
 // Frame is an EnclaveRPC frame.
 //
 // It is the Go analog of the Rust RPC frame defined in runtime/src/enclave_rpc/types.rs.
--- a/go/runtime/host/protocol/types.go
+++ b/go/runtime/host/protocol/types.go
@@ -238,6 +238,8 @@ type RuntimeCapabilityTEERakQuoteResponse struct {
 type RuntimeRPCCallRequest struct {
 	// Request.
 	Request []byte `json:"request"`
+	// Kind is the type of RPC call.
+	Kind enclaverpc.Kind `json:"kind,omitempty"`
 }
 
 // RuntimeRPCCallResponse is a worker RPC call response message body.
@@ -442,8 +444,9 @@ type RuntimeConsensusSyncRequest struct {
 
 // HostRPCCallRequest is a host RPC call request message body.
 type HostRPCCallRequest struct {
-	Endpoint string `json:"endpoint"`
-	Request  []byte `json:"request"`
+	Endpoint string          `json:"endpoint"`
+	Request  []byte          `json:"request"`
+	Kind     enclaverpc.Kind `json:"kind,omitempty"`
 
 	// PeerFeedback contains optional peer feedback for the last RPC call under the given endpoint.
 	//
--- a/go/worker/keymanager/worker.go
+++ b/go/worker/keymanager/worker.go
@@ -144,36 +144,44 @@ func (w *Worker) Initialized() <-chan struct{} {
 	return w.initCh
 }
 
-func (w *Worker) CallEnclave(ctx context.Context, data []byte) ([]byte, error) {
-	// Handle access control as only peers on the access list can call this method.
-	peerID, ok := rpc.PeerIDFromContext(ctx)
-	if !ok {
-		return nil, fmt.Errorf("not authorized")
-	}
+func (w *Worker) CallEnclave(ctx context.Context, data []byte, kind enclaverpc.Kind) ([]byte, error) {
+	switch kind {
+	case enclaverpc.KindNoiseSession:
+		// Handle access control as only peers on the access list can call this method.
+		peerID, ok := rpc.PeerIDFromContext(ctx)
+		if !ok {
+			return nil, fmt.Errorf("not authorized")
+		}
 
-	// Peek into the frame data to extract the method.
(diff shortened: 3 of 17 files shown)


Yawning:
You know, looking over this, it occurs to me that while the primary reason why our domain separated ed25519 is non-standard is because of Ledger etc.  There are places where such concerns are irrelevant so we could use ed25519ctx (or sr25519 for that matter).

Probably not for this PR but something to consider in the long term maybe.

go/runtime/enclaverpc/api/api.go
```diff
@@ -10,6 +14,55 @@ type Frame struct {
...
+}
+
+// ResponseBundle is an RPC response enriched with server information.
+type ResponseBundle struct {
+	Response   []byte      `json:"response"`
+	SenderInfo *SenderInfo `json:"sender_info"`
+}
+
+// InitRequest is a key manager initialization request.
+type InitRequest struct {
```
kostko, line 60: Key manager protocol types shouldn't be in the general EnclaveRPC package.

go/worker/keymanager/worker.go
```diff
@@ -221,33 +220,26 @@ func (w *Worker) updateStatus(status *api.Status, runtimeStatus *runtimeStatus)
...
-		Args   InitRequest `json:"args"`
-	}
-
 	var policy []byte
 	if status.Policy != nil {
 		policy = cbor.Marshal(status.Policy)
 	}
 
-	call := InitCall{
+	request := enclaverpc.Request{
```
kostko, line 228: Hm, so this was a local call before as one must ensure that this method can only be called locally.

keymanager/src/runtime/methods.rs
```diff
@@ -62,14 +63,18 @@ pub fn get_or_create_ephemeral_keys(
...
     // No authentication or authorization.
     // Absolutely anyone is allowed to query public ephemeral keys.
     validate_epoch(req.epoch, ctx)?;
 
     let kdf = Kdf::global();
     let pk = kdf.get_public_key(req)?;
-    pk.map_or(Ok(None), |pk| Ok(Some(kdf.sign_public_key(pk)?)))
+
+    let req = cbor::to_vec((*req).clone());
+    let sig = kdf.sign_public_key(pk, req)?;
```
bennetyee, line 75: drive by here.

that we're signing req along w/ pk is good but should there be more context, such as cacheability and cache expiry?  esp with ephemeral keys but probably also relevant to `get_public_key` above, once a key is no longer valid the holder of the corresponding private key could disclose it (esp if the protocols using the keypair is forward secure), but a cacheable signed statement that's unscoped saying the key is good would make that dangerous.

keymanager/src/crypto/kdf.rs
```diff
@@ -419,30 +429,21 @@ impl Kdf {
...
             .as_ref()
             .ok_or(KeyManagerError::NotInitialized)?;
-        let signature = signer.sign(&PUBLIC_KEY_CONTEXT, &body)?;
 
-        Ok(SignedPublicKey {
-            key,
-            checksum,
-            signature,
-        })
+        SignedPublicKey::new(key, checksum, req, signer)
```
bennetyee, line 446: So `req: Vec<u8>` is a sort of cryptographic context to indicate what the signature "means", but it's not perfect. In particular, `req` is a serialized request, presumably by oasis-cbor, and the serialized struct obeys canonical cbor like rules with struct member names used as keys in a map, and keys are sorted.  this means that two structs with different names/purposes but with identical members (names and types, but possibly occuring in a different order) will serialize into the same "type" of byte layout, even though rust will treat the two structs as different types -- the encoded bytes of one struct's value could be decoded as an instance of the other struct without oasis-cbor generating an error.  so if ever there were two requests that use same-cbor-layout structs, then the replies could be confused.  it's unlikely to occur, but it's a potential gotcha.

peternose:
After discussion we had last week I'm splitting this PR into two parts. In the first part only support for insecure RPC calls will be added. The latter will be used in the second part to make public key requests and to verify signatures.

Tested locally:
- If request for private key uses an insecure call, an error is returned.
```
level=error ts=2022-12-13T10:20:16.576441842Z caller=root.go:434 module=test-runner msg="failed to run scenario" err="root: failed to run scenario: failed to decrypt ciphertext: failed to submit decrypt tx to runtime: runtime tx failed: private ephemeral key not available: call failed: method call is not secure: \"get_or_create_ephemeral_keys\"" scenario=e2e/runtime/keymanager-key-generation run_id=0
```
- If request for public key uses an insecure or secure call, things work (insecure call should not be used as the signature is not verified; will be done in another PR).
- If the key manager client and the key manager are not upgraded, the default channel is a secure one.

keymanager/src/runtime/methods.rs
peternose: I think we already have an implicit cache expiry. To verify the signature one needs RAK which should be verified with a quote and a policy. And quotes have expiration date already set.

go/runtime/host/protocol/types.go
```diff
@@ -234,10 +234,22 @@ type RuntimeCapabilityTEERakQuoteResponse struct {
 	Signature signature.RawSignature `json:"signature"`
 }
 
+// RPCKind is the RPC call kind.
```
kostko, line 237: Should this rather be defined in `go/runtime/enclaverpc/api` and then used in all the places where `kind` is defined?

runtime/src/enclave_rpc/dispatcher.rs
```diff
@@ -96,11 +100,20 @@ where
 pub struct Method {
     /// Method dispatcher.
     dispatcher: Box<dyn MethodHandlerDispatch + Send + Sync>,
+    /// Specifies whether the method has to be called over a secure channel.
+    is_secure: bool,
+    /// Specifies whether the method has to be called locally.
+    is_local: bool,
```
kostko, line 106: Should we rather put these fields into the `MethodDescriptor`? This would also make the method registrations more clear as the parameters would be named.

kostko (approved):

PR #5085, opened 2022-12-08
go/consensus: Implement libp2p backed light client sync protocol

ptrus:
Depends on: https://github.com/oasisprotocol/tendermint/pull/192

TODO: 
- [x] finalize tendermint PR (waiting for comments in this PR here, if we decide to do any more updates to `MalevolentProvider`)
- [x] update tendermint dependency tag once https://github.com/oasisprotocol/tendermint/pull/192 is merged and tagged


go/consensus/api/api.go
```diff
@@ -385,6 +386,24 @@ type Backend interface {
 	//
 	// This may be nil in case checkpoints are disabled.
 	Checkpointer() checkpoint.Checkpointer
+
+	// RegisterP2PBackend registers the P2P backend used for light client state sync.
+	RegisterP2PBackend(rpc.P2P) error
+
+	// LightClient returns the libp2p backend light client (if configured.).
+	LightClient() (LightClient, error)
```
kostko, line 394: I noticed that we store the light client instance in the `Node` and then pass it around. Does it make more sense for the light client to be decoupled from the consensus backend (I guess one reason the consensus backend needs it is for state sync)? Or in case it is coupled should all the places just always access it via `.Consensus.LightClient()`?

go/consensus/tendermint/light/client.go
```diff
@@ -228,6 +122,7 @@ func NewClient(ctx context.Context, cfg ClientConfig) (Client, error) {
 		tmlightdb.New(tmdb.NewMemDB(), ""), // TODO: Make the database configurable.
```
kostko, line 122: While we're here, given that we will be using the light client more, should we make this persistent (e.g. in the tendermint part of the data directory, where the location is passed via `cfg.LightDbPath` or similar)? I guess it makes sense in case there will only ever be a single light client instance per node (which probably makes sense).

go/consensus/tendermint/full/full.go
```diff
@@ -785,6 +757,37 @@ func (t *fullService) lazyInit() error { // nolint: gocyclo
...
+			// Create new state sync state provider.
+			cfg := lightAPI.ClientConfig{
+				GenesisDocument: tmGenDoc,
+				TrustOptions: tmlight.TrustOptions{
+					Period: viper.GetDuration(CfgConsensusStateSyncTrustPeriod),
+					Height: int64(viper.GetUint64(CfgConsensusStateSyncTrustHeight)),
+					Hash:   tenderConfig.StateSync.TrustHashBytes(),
+				},
+			}
+			if stateProvider, t.lightClient, err = newStateProvider(t.ctx, t.genesis.ChainContext(), cfg, t.p2p); err != nil {
```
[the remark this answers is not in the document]
ptrus: Should probably be moved out, but I didn't yet start modifying the configuration settings. The trust settings configurations will be refactored to something like: `consensus.tendermint.light_client.trust_hash`. And the light client can then be instantiated separately and passed into the state sync if state sync is enabled (via the existing "consensus.tendermint.state_sync.enabled" flag).

go/consensus/tendermint/light/p2p/p2p.go
```diff
@@ -0,0 +1,319 @@
...
+func (lp *lightClientProvider) Initialized() <-chan struct{} {
+	return lp.initCh
+}
+
+// MalevolentProvider implements api.Provider.
+func (lp *lightClientProvider) MalevolentProvider() {
+	lp.RLock()
+	defer lp.RUnlock()
+
+	// Ignore feedback right after peer was switched, as it could be a delayed
```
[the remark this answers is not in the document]
ptrus: The problem is that the peer can be switched for other reasons (light client exposes multiple methods and each could report a bad peer feedback which would then cause the peer to be switched internally). But using peer-feedback works fine because the feedback is reported for the actual underlying peer which served the response. 

For the light client, it's problematic because it is not aware of the underlying peer, it only knows that the provider misbehaved. Since we are already modifying our TM fork, we could instead make it aware of the peers. It would look something like this:
- the `MalevolentProvider` would be updated to: `MalevolentProvider(peer core.PeerID)` (or maybe even `string` type)
- the `LightBlock` & `ReportEvidence` methods would be updated to include the `peerID` in the response

The light client could then report the exact peer that provided the malicious response.

go/consensus/tendermint/light/client.go
ptrus: Yeah I think this makes sense.

go/consensus/api/api.go
ptrus: > Or in case it is coupled should all the places just always access it via .Consensus.LightClient()?

Yeah, if it stays coupled, this should be done.

> Does it make more sense for the light client to be decoupled from the consensus backend

Not sure, maybe. I'll try it out and see how it looks.

go/consensus/tendermint/full/full.go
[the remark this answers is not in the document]
ptrus: Hm, that's true.

So if state sync is configured, the light-client would need to be initialized before the consensus node, with trust settings from the configuration, and if state sync is disabled (meaning this is a full node), the light-client should be initialized once the full node is synced, from the trusted root obtained from the node.
Does this make sense?

Probably doesn't make sense for state-sync to use a separate internal light-client? Although that would simplify the above to: always wait for consensus to be synced, then initialize a light client with a trusted latest block. Which would then also nicely lead to decoupling the light-client out of the consensus backend.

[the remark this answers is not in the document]
kostko: > meaning this is a full node

In both cases this can be a full node, one just doesn't have all the history (e.g. starts from the state sync snapshot) and the other syncs from genesis.

> the light-client should be initialized once the full node is synced, from the trusted root obtained from the node

Yeah it should probably use the lowest and highest blocks after full node consensus sync known by the node and store them as trusted.

go/consensus/tendermint/light/p2p/p2p.go
```diff
@@ -0,0 +1,319 @@
...
+	// feedback of the previous peer.
+	if time.Now().Before(lp.peerSelectionTime.Add(freshPeerFeedbackTimeout)) {
+		return
+	}
+
+	lp.p2pMgr.RecordBadPeer(*lp.peerID)
```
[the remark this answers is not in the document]
ptrus: Recording the peer as bad will cause for it to be dropped and replaced (in the `worker` loop).

go/consensus/tendermint/full/full.go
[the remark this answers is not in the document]
ptrus: > Yeah it should probably use the lowest and highest blocks after full node consensus sync known by the node and store them as trusted.

Would it make sense that the light-client also follows full node blocks and periodically stores the latest block as trusted? (e.g. every 100 blocks)?

go/consensus/api/api.go
ptrus: I decoupled the light-client from the consensus backend, please take another look.

go/consensus/tendermint/light/client.go
ptrus: As discussed, there's now two light clients. The "internal" one, only used for state sync, still uses the in-memory store.
The Light client service (a light client that is exposed to other backends as well) is persistent and is initialized with trusted blocks from the locally synced tendermint backend.

go/oasis-node/cmd/node/node.go
```diff
@@ -398,21 +391,6 @@ func (n *Node) startRuntimeWorkers() error {
 		return err
 	}
 
-	// Start the public consensus services worker.
```
ptrus, line 401: External gRPC server, TLS address in registry and similar are not needed anymore after this change. Due to the amount of changes, it will be removed in a followup PR.

go/consensus/tendermint/full/full.go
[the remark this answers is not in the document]
kostko: Hm maybe once per epoch, given that the validator set will not change inside an epoch.

go/consensus/tendermint/light/p2p/p2p.go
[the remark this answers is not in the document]
kostko: Ah, the loop watches the peer manager, ok.

go/consensus/tendermint/full/full.go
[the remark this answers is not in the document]
ptrus: Added once per epoch.

go/consensus/tendermint/full/full.go
```diff
@@ -785,6 +771,37 @@ func (t *fullService) lazyInit() error { // nolint: gocyclo
 			}
 		}()
 
+		// Configure state sync if enabled.
+		var stateProvider tmstatesync.StateProvider
+		if viper.GetBool(CfgConsensusStateSyncEnabled) {
+			t.Logger.Info("state sync enabled")
+
+			if t.p2p == nil {
```
kostko, line 779: This doesn't seem to hold the lock while accessing `t.p2p`.

go/consensus/tendermint/light/p2p/p2p.go
```diff
@@ -0,0 +1,324 @@
...
+		ctx:          ctx,
+		chainID:      chainID,
+		peerRegistry: make(map[core.PeerID]bool),
+		p2pMgr:       mgr,
+		rc:           rc,
+	}
+}
+
+type lightClientProvider struct {
+	sync.RWMutex
```
kostko, line 87: Also use a separate member for this mutex.

go/consensus/tendermint/light/p2p/p2p.go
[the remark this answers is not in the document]
kostko: > The light client could then report the exact peer that provided the malicious response.

Yeah I think that would be a better approach. I would use the `string` type for peer IDs to make it agnostic of what this is used for.

go/consensus/tendermint/light/service.go
```diff
@@ -0,0 +1,337 @@
...
+	"github.com/oasisprotocol/oasis-core/go/consensus/tendermint/common"
+	"github.com/oasisprotocol/oasis-core/go/consensus/tendermint/db"
+	"github.com/oasisprotocol/oasis-core/go/consensus/tendermint/light/api"
+	p2pLight "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/light/p2p"
+	genesisAPI "github.com/oasisprotocol/oasis-core/go/genesis/api"
+	"github.com/oasisprotocol/oasis-core/go/p2p/rpc"
+)
+
+const (
+	dbName = "tendermint/tendermint-light"
```
kostko, line 35:
```suggestion
	dbName = "tendermint/light"
```

go/consensus/tendermint/light/service.go
```diff
@@ -0,0 +1,337 @@
...
+	// Wait for consensus to be synced.
+	select {
+	case <-c.stopCh:
+		return
+	case <-c.ctx.Done():
+		return
+	case <-c.consensus.Synced():
+	}
+
+	// Store latest light block into the trusted store.
```
kostko, line 170: It should also try to store the lowest available light block.

go/consensus/tendermint/light/service.go
```diff
@@ -0,0 +1,337 @@
...
+		c.logger.Error("failed to initialize light client store with a trusted block", "err", err)
+		return
+	}
+
+	chainCtx, err := c.consensus.GetChainContext(c.ctx)
+	if err != nil {
+		c.logger.Error("failed to obtain chain context", "err", err)
+		return
+	}
+	tmGenDoc, err := tmapi.GetTendermintGenesisDocument(c.genesis)
```
kostko, line 191: This seems expensive just to get the Tendermint's chain context. Maybe expose something like `tmapi.TendermintChainID(chainContext string)` and call that instead on `chainCtx`?

go/consensus/tendermint/light/service.go
```diff
@@ -0,0 +1,337 @@
...
+		return
+	}
+	c.lc = &lightClient{tmc: tmc}
+	close(c.initCh)
+	initOk = true
+
+	// Watch epochs and insert new trusted blocks on every epoch transition.
+	ch, sub, err := c.consensus.Beacon().WatchEpochs(c.ctx)
+	if err != nil {
+		c.logger.Error("failed to obtain watch epochs", "err", err)
```
kostko, line 226:
```suggestion
		c.logger.Error("failed to watch epochs", "err", err)
```

go/consensus/tendermint/light/service.go
```diff
@@ -0,0 +1,337 @@
...
+		viper.GetDuration(CfgConsensusLightClientTrustPeriod),
+		providers[0],  // Primary provider.
+		providers[1:], // Witnesses.
+		c.store,
+		tmlight.MaxRetryAttempts(lcMaxRetryAttempts),
+		tmlight.Logger(common.NewLogAdapter(!viper.GetBool(common.CfgLogDebug))),
+		tmlight.DisableProviderRemoval(),
+	)
+	if err != nil {
+		c.logger.Error("failed to obtain initialize tendermint light client", "err", err)
```
kostko, line 216:
```suggestion
		c.logger.Error("failed to initialize tendermint light client", "err", err)
```

go/consensus/tendermint/light/service.go
```diff
@@ -0,0 +1,337 @@
...
+	return tmLb, nil
+}
+
+func (c *client) worker() {
+	if !c.enabled {
+		return
+	}
+	var initOk bool
+	defer func() {
+		if !initOk {
```
kostko, line 156: Is there a reason why this doesn't always close the quit channel, but it only does so on failures? Will this result in the service hanging when stopped regularly?

go/consensus/tendermint/light/service.go
```diff
@@ -0,0 +1,337 @@
...
+			if err = c.store.SaveLightBlock(tmLb); err != nil {
+				c.logger.Error("failed to initialize light client store with a trusted block", "err", err)
+				continue
+			}
+		}
+	}
+}
+
+// GetLightBlock implements api.Client.
+func (c *client) GetLightBlock(ctx context.Context, height int64) (*consensus.LightBlock, rpc.PeerFeedback, error) {
```
kostko, line 257: Given that this particular light client service also has a full node available, should it always first try to fetch the block from the consensus backend? This way the callers don't need to worry about querying the consensus backend first and can just always invoke the light client.

kostko (commented):
I think the overall design looks good now. Some minor comments.

go/consensus/tendermint/light/service.go
ptrus: Added.

go/consensus/tendermint/light/service.go
ptrus: It was wrong. I refactored the init/stop/quit channels. Should be better now.

go/consensus/tendermint/light/p2p/p2p.go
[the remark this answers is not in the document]
ptrus: Updated, see: https://github.com/oasisprotocol/tendermint/pull/192
Instead of modifying `LightBlock`, I added `LightBlockWithPeerID` to the Tendermint `Provider` interface, as `LightBlock` is used quite a bit throughout the code.

Also see updates in this PR.

PR #5088, opened 2022-12-08
go/registry: Add RuntimeSuspendedEvent and WatchUpdates method

ptrus:


go/consensus/tendermint/apps/roothash/roothash.go
```diff
@@ -177,6 +179,7 @@ func (app *rootHashApplication) onCommitteeChanged(ctx *tmapi.Context, state *ro
 			}
 		}
 		if (empty || !sufficientStake) && !params.DebugDoNotSuspendRuntimes {
+			ctx.EmitEvent(api.NewEventBuilder(registryApp.AppName).TypedAttribute(&registry.RuntimeSuspendedEvent{RuntimeID: rt.ID}))
```
kostko, line 182: Would it be better if we put this into `regState.SuspendRuntime` so that new call sites always emit this event?

kostko: For example, note that the key manager service also suspends runtimes and this is currently missing here.

ptrus: Maybe yes. Although emitting events directly from the state wrapper packages doesn't seem nice either. But it does avoid doing this from multiple call sites, so it's probably worth it.

kostko: True, it's currently somewhat strange. But we are doing it for some things, e.g. in the staking service.

ptrus: Updated

kostko (approved):

PR #5100, opened 2022-12-13
go/p2p-light-client followup: remove unused gRPC/TLS parts

ptrus:
https://github.com/oasisprotocol/oasis-core/pull/5085 made a bunch of gRPC/TLS/consensus-rpc code obsolete. This follow-up PR cleans up unused parts.

The last remaining external grpc endpoint is the sentry control endpoint. Node's behind sentries connect to this endpoint to obtain sentry addresses, to use them for registration.

TODO:
- [x] remove consensus-rpc role
- [x] figure out where to store the exported mainnet genesis file


go/registry/api/api_test.go
```diff
@@ -137,104 +137,6 @@ func TestVerifyRegisterNodeArgs(t *testing.T) {
...
-					ID: nodeP2PSigner.Public(),
-				},
-				VRF: &node.VRFInfo{
-					ID: nodeVRFSigner.Public(),
-				},
-				Roles:      node.RoleConsensusRPC,
-				Expiration: 11,
-			},
-			ErrInvalidArgument,
-			"invalid consensus RPC node (missing P2P address)",
```
kostko, line 176: Instead of removing this case, should we just use a different role? All nodes should now include P2P addresses anyway.

runtime/src/consensus/registry.rs
```diff
@@ -59,8 +59,8 @@ pub struct TLSInfo {
     /// Public key that will be used for establishing connections after certificate rotation (if enabled).
     pub next_pub_key: PublicKey,
 
-    /// List of addresses at which the node can be reached.
-    pub addresses: Option<Vec<TLSAddress>>,
+    #[cbor(rename = "addresses", optional)]
+    pub _deprecated_addresses: Option<Vec<TLSAddress>>,
```
kostko, line 63: I guess v3 descriptors should deserialize fine with v2 because the addresses field will just be set to `None`?

kostko (commented):
> The last remaining external grpc endpoint is the sentry control endpoint. Node's behind sentries connect to this endpoint to obtain sentry addresses, to use them for registration.

Wild thought for a different PR, but could this be moved over to use P2P (e.g. add the sentry as a persistent peer and talk to it using the "sentry" protocol, only allowing configured keys)?

runtime/src/consensus/registry.rs
ptrus: Yeah.  This is/should be also  covered with tests, e.g.: 
```
qWF2A2JpZFggAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABjcDJwomJpZFggAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABpYWRkcmVzc2Vz9mN0bHOiZ3B1Yl9rZXlYIAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAbG5leHRfcHViX2tleVggAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABlcm9sZXMAaHJ1bnRpbWVz9mljb25zZW5zdXOiYmlkWCAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAGlhZGRyZXNzZXP2aWVudGl0eV9pZFggAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABqZXhwaXJhdGlvbgA
```
is one of the V3 descriptors that is tested to deserialize correctly (both on go and rust sides).

`test_consistent_node` tests various V3 descriptors to deserialize correctly.

ptrus:
> Wild thought for a different PR, but could this be moved over to use P2P (e.g. add the sentry as a persistent peer and talk to it using the "sentry" protocol, only allowing configured keys)?

Probably doable yes. Could be looked into at a later time. Not sure how beneficial it would be to spend time on this, though probably a very low priority. Maybe one day we'll be able to just get rid of sentries in their current design :grin:

Also, the sentry gRPC is quite self-contained, and I was already able to remove the majority of the complex gRPC code (policies and proxy) so am quite satisfied with that.

go/registry/api/api_test.go
ptrus: Makes sense, done.

kostko (approved):

PR #5101, opened 2022-12-14
keymanager/src/client: Fetch public keys using insecure RPC requests

peternose:
### Task
Fetch and verify public long-term/ephemeral runtime keys through an insecure channel.
        
### Test
Key manager upgrade tested locally with e2e test. Upgrade works as we allow empty runtime signing keys. An alternative would be to be strict here, but then one key manager would not be included in the node list because of the `Runtime signing key mismatch for runtime` error. Not sure if this is a problem though, as the new manager will be added to the list as soon as the old one is deregistered.

### DIFF
--- a/go/runtime/host/protocol/types.go
+++ b/go/runtime/host/protocol/types.go
@@ -16,6 +16,7 @@ import (
 	consensus "github.com/oasisprotocol/oasis-core/go/consensus/api"
 	consensusTx "github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"
 	consensusResults "github.com/oasisprotocol/oasis-core/go/consensus/api/transaction/results"
+	keymanager "github.com/oasisprotocol/oasis-core/go/keymanager/api"
 	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
 	"github.com/oasisprotocol/oasis-core/go/roothash/api/block"
 	"github.com/oasisprotocol/oasis-core/go/roothash/api/commitment"
@@ -89,6 +90,8 @@ type Body struct {
 	RuntimeExecuteTxBatchResponse              *RuntimeExecuteTxBatchResponse             `json:",omitempty"`
 	RuntimeAbortRequest                        *Empty                                     `json:",omitempty"`
 	RuntimeAbortResponse                       *Empty                                     `json:",omitempty"`
+	RuntimeKeyManagerStatusUpdateRequest       *RuntimeKeyManagerStatusUpdateRequest      `json:",omitempty"`
+	RuntimeKeyManagerStatusUpdateResponse      *Empty                                     `json:",omitempty"`
 	RuntimeKeyManagerPolicyUpdateRequest       *RuntimeKeyManagerPolicyUpdateRequest      `json:",omitempty"`
 	RuntimeKeyManagerPolicyUpdateResponse      *Empty                                     `json:",omitempty"`
 	RuntimeKeyManagerQuotePolicyUpdateRequest  *RuntimeKeyManagerQuotePolicyUpdateRequest `json:",omitempty"`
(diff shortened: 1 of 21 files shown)


keymanager/src/crypto/types.rs
```diff
@@ -96,13 +117,271 @@ pub struct InputKeyPair {
...
+
 /// Signed public key.
 #[derive(Clone, Debug, Default, PartialEq, Eq, cbor::Encode, cbor::Decode)]
 pub struct SignedPublicKey {
     /// Public key.
     pub key: PublicKey,
     /// Checksum of the key manager state.
     pub checksum: Vec<u8>,
-    /// Sign(sk, (key || checksum)) from the key manager.
+    /// Sign(sk, (key || checksum || runtime id || key pair id || epoch || expiration date) from
```
kostko, line 140:
```suggestion
    /// Sign(sk, (key || checksum || runtime id || key pair id || epoch || expiration epoch)) from
```

keymanager/src/crypto/types.rs
```diff
@@ -96,13 +117,271 @@ pub struct InputKeyPair {
...
+        key_pair_id: KeyPairId,
+        epoch: Option<EpochTime>,
+    ) -> Vec<u8> {
+        let mut body = key.as_ref().to_vec();
+        body.extend_from_slice(checksum);
+        body.extend_from_slice(runtime_id.as_ref());
+        body.extend_from_slice(key_pair_id.as_ref());
+        if let Some(epoch) = epoch {
+            body.extend_from_slice(&epoch.to_be_bytes());
+            body.extend_from_slice(&MAX_SIGNED_EPHEMERAL_PUBLIC_KEY_AGE.to_be_bytes());
```
kostko, line 211: This seems like it makes the scheme harder to upgrade if you want to actually change the value of the constant as old clients would deem responses from new runtimes invalid. Any downside to including the expiration field in the structure itself?

go/runtime/host/protocol/types.go
```diff
@@ -402,12 +408,17 @@ type RuntimeExecuteTxBatchResponse struct {
 	Deprecated1 cbor.RawMessage `json:"batch_weight_limits,omitempty"`
 }
 
-// RuntimeKeyManagerPolicyUpdateRequest is a runtime key manager policy request message body.
+// RuntimeKeyManagerStatusUpdateRequest is a runtime key manager status update request message body.
+type RuntimeKeyManagerStatusUpdateRequest struct {
```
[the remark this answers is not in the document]
peternose: Yes. I think that's better than having 2 methods, one for policy and one for the key.

keymanager/src/crypto/types.rs
peternose: Adding new field to the `SignedPublicKey` breaks old key manager clients as they will fail to decode cbor encoded RPC responses (extra field).
- We can make expiration field optional, disable cache expiration in the next core release and enable it after key manager clients are upgraded (probably in the next key manager release).
- We can set the expiration field only when public key request is made over an insecure channel (upgraded clients) and permanently enable after the next release.

The second solution would give us cache expiration immediately after the clients/runtimes upgrade.

kostko: Yeah the second option sounds ok. Also speaking of versioning, should we introduce some way of versioning the protocol to make this easier in the future (could be done in a separate PR)?

keymanager/src/crypto/types.rs
```diff
@@ -96,13 +116,308 @@ pub struct InputKeyPair {
...
+
 /// Signed public key.
 #[derive(Clone, Debug, Default, PartialEq, Eq, cbor::Encode, cbor::Decode)]
 pub struct SignedPublicKey {
     /// Public key.
     pub key: PublicKey,
     /// Checksum of the key manager state.
     pub checksum: Vec<u8>,
-    /// Sign(sk, (key || checksum)) from the key manager.
+    /// Sign(sk, (key || checksum || runtime id || key pair id || epoch || expiration epoch) from
```
kostko, line 139:
```suggestion
    /// Sign(sk, (key || checksum || runtime id || key pair id || epoch || expiration epoch)) from
```

kostko (approved):

PR #5113, opened 2023-01-03
go/sgx/ias: Add support for blacklisting GIDs

ptrus:


runtime/src/common/sgx/ias.rs
```diff
@@ -55,6 +55,8 @@ enum AVRError {
     InvalidSignature,
     #[error("IAS quotes are disabled by policy")]
     Disabled,
+    #[error("Blacklisted IAS quote GID")]
```
kostko, line 58:
```suggestion
    #[error("blacklisted IAS quote GID")]
```

kostko (approved):

PR #5114, opened 2023-01-04
go/registry: Support changing a runtime's owner

ptrus:
This existing code ensures that the new entity actually exists and has the required stake:
https://github.com/oasisprotocol/oasis-core/blob/2680a6e97f0ebf2ffa892c72cbaf7289e85f3938/go/consensus/tendermint/apps/registry/transactions.go#L749-L775

However, in the existing registry node tests `params.DebugBypassStake` is used, so this cannot really be tested there. That's why I added a separate unit test suite for register runtime (with is currently only testing the updating entity owner behavior).

### DIFF
--- a/go/consensus/tendermint/apps/registry/transactions.go
+++ b/go/consensus/tendermint/apps/registry/transactions.go
@@ -715,6 +715,8 @@ func (app *registryApplication) registerRuntime( // nolint: gocyclo
 		// based on the existing governance model and not the new one,
 		// otherwise it would be impossible to transition from entity to
 		// runtime governance, for example.
+		// Similarly, if we're updating the controlling entity, we should check the signer
+		// based on the existing EntityID and not the new one.
 		var rtToCheck *registry.Runtime
 		if existingRt != nil {
 			rtToCheck = existingRt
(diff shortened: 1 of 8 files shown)


kostko (commented):
Hm, so one thing to take care is removing the stake claim for the old entity which currently doesn't seem to happen?

Also this should be caught by the sanity checks in tests, so this needs to be tested somewhere where `DebugBypassStake` is not set.

ptrus:
> Hm, so one thing to take care is removing the stake claim for the old entity which currently doesn't seem to happen?

Fixed.

> Also this should be caught by the sanity checks in tests, so this needs to be tested somewhere where DebugBypassStake is not set.

Added this to the txsoruce test (workload/registration). I also tested this without the fix for removing stake claim and confirmed that sanity checker detects it.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -770,6 +772,18 @@ func (app *registryApplication) registerRuntime( // nolint: gocyclo
 			)
 			return nil, err
 		}
+
+		// If owner was updated remove stake claim from previous owner.
```
kostko, line 776: Actually, this should also happen when changing the governance model to runtime as the `acctAddr` changed? I guess this transition is never tested in E2E tests as it should otherwise cause the sanity check to fail? Also why isn't the above just using `rt.StakingAddress()` to compute `acctAddr`?

ptrus: > Actually, this should also happen when changing the governance model to runtime as the acctAddr changed

Added

> Also why isn't the above just using rt.StakingAddress() to compute acctAddr

Refactored this whole section to make it cleaner.

> I guess this transition is never tested in E2E tests

Yeah never tested. I'll look if there's an easy way to add it.

ptrus: Added a test for this to `runtime-dynamic` E2E test.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -746,30 +748,31 @@ func (app *registryApplication) registerRuntime( // nolint: gocyclo
...
 				"runtime", rt.ID,
-				"account", acctAddr,
+				"account", rtAddress,
 			)
 			return nil, err
 		}
+
+		// If owner was updated remove stake claim from previous owner.
+		if existingRt != nil && *existingRt.StakingAddress() != *rtAddress {
+			if err = stakingState.RemoveStakeClaim(ctx, *existingRt.StakingAddress(), claim); err != nil {
```
kostko, line 767: Given this is now doing multiple operations related to stake claims, it would be better to do the updates inside a transaction similar to how it is done in `registerNode`:
```golang
	// Start a new transaction and rollback in case we fail.
	ctx = ctx.NewTransaction()
	defer ctx.Close()

	// ...

	ctx.Commit()
```

go/oasis-node/cmd/debug/txsource/workload/registration.go
```diff
@@ -290,13 +303,22 @@ func (r *registration) Run( // nolint: gocyclo
...
 
+	iteration := 0
+	var loopCtx context.Context
+	var cancel context.CancelFunc
 	for {
+		if cancel != nil {
+			cancel()
+		}
+		loopCtx, cancel = context.WithTimeout(ctx, registryIterationTimeout)
+		defer cancel()
```
kostko, line 314: Add empty line below.

go/oasis-node/cmd/debug/txsource/workload/registration.go
```diff
@@ -325,6 +347,30 @@ func (r *registration) Run( // nolint: gocyclo
 			"node", selectedNode.nodeDesc,
 		)
 
+		// Periodically re-register the runtime with a new owner.
+		if iteration&20 == 0 {
```
kostko, line 351: Maybe move this "owner change interval" to a constant?

go/consensus/tendermint/apps/registry/transactions.go
ptrus: Added.

kostko (approved):

PR #5117, opened 2023-01-05
tendermint: GetEvents: return events in order

mitjat:
Currently consensus events (e.g. in various GetEvents RPCs) are returned in the following order:
- BeginBlock
- EndBlock
- DeliverTx

This PR switches all GetEvents RPCs to the more natural order:
- BeginBlock
- DeliverTx
- EndBlock

The motivation comes from oasis-indexer, where processing events in the order they were returned before this PR can lead to violated assumptions (e.g. "no account can ever have negative balance").

### DIFF
--- a/go/consensus/tendermint/roothash/roothash.go
+++ b/go/consensus/tendermint/roothash/roothash.go
@@ -288,14 +288,8 @@ func (sc *serviceClient) getEvents(ctx context.Context, height int64, txns [][]b
 	}
 
 	var events []*api.Event
-	// Decode events from block results.
+	// Decode events from block results (at the beginning of the block).
 	blockEvs, err := EventsFromTendermint(nil, results.Height, results.BeginBlockEvents)
-	if err != nil {
-		return nil, err
-	}
-	events = append(events, blockEvs...)
-
-	blockEvs, err = EventsFromTendermint(nil, results.Height, results.EndBlockEvents)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


go/consensus/tendermint/roothash/roothash.go
```diff
@@ -418,10 +419,11 @@ func (sc *serviceClient) reindexBlocks(currentHeight int64, bh api.BlockHistory)
 		}
 
 		// Index block.
-		tmEvents := append(results.BeginBlockEvents, results.EndBlockEvents...)
+		tmEvents := results.BeginBlockEvents
 		for _, txResults := range results.TxsResults {
 			tmEvents = append(tmEvents, txResults.Events...)
 		}
+		tmEvents = append(tmEvents, results.EndBlockEvents...)
```
mitjat, line 426: Looks unrelated, but might as well change this too, for consistency.
This covers all the occurrences of `EndBlockEvents` in the codebase.

kostko (approved):

.changelog/5117.breaking.md
```diff
@@ -0,0 +1,4 @@
+go/tendermint: Change order of events returned from GetEvents()
```
ptrus, line 1: I guess under a strict definition this could be considered a breaking change as it modifies the order in API responses. However, this is not a consensus breaking change (no consensus issues in case some nodes apply this change)  and could be considered a bugfix. That would be useful if we want to backport this to `22.2.x` branch (the release process forbids breaking changes in point releases). 

Since oasis-indexer relies on this change, I would recommend backporting this to `stable/22.2.x` and making a new release, as otherwise the indexer backed nodes would need a custom-built version of oasis-node.

ptrus (approved):

.changelog/5117.breaking.md
mitjat: Thanks!

I meanwhile already fixed the indexer so it does some heuristic reordering of events. So it does not depend on this change. I was hoping that once this PR goes in, I could remove the reordering on the indexer side, but I guess we'll also have to process archive data from Cobalt sooner or later, and I'm not sure it's worth backporting this change all the way there.

Still, it's nice to have flexibility. Good to know this is not consensus-breaking. Relabeling as bugfix.

PR #5121, opened 2023-01-10
crypto/x25519: Add type-safe X25519 private/public key types

peternose:
Waiting for https://github.com/oasisprotocol/curve25519-voi/pull/120 to be merged.


runtime/src/common/crypto/x25519.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+}
+
+impl cbor::Encode for PrivateKey {
+    fn into_cbor_value(self) -> cbor::Value {
+        cbor::to_value(self.0.to_bytes())
+    }
+}
+
+impl cbor::Decode for PrivateKey {
```
kostko, line 51: Add a `try_default` implementation that returns the default value, something like:
```rust
fn try_default() -> Result<Self, DecodeError> {
    Ok(Default::default())
}
```
Same for the public key type.

kostko (approved):

PR #5123, opened 2023-01-10
go/worker/client: Better handle latest round queries with verification

kostko:
When a query is requesting to be executed against the latest round and the runtime reports a consensus verifier error, use an earlier round instead as the latest round may not yet be verifiable by the light client as it needs to wait for the validator signatures.

### DIFF
--- a/go/worker/client/committee/node.go
+++ b/go/worker/client/committee/node.go
@@ -11,6 +11,7 @@ import (
 
 	cmnBackoff "github.com/oasisprotocol/oasis-core/go/common/backoff"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
+	"github.com/oasisprotocol/oasis-core/go/common/errors"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
 	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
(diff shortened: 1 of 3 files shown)


go/worker/client/committee/node.go
```diff
@@ -149,22 +149,42 @@ func (n *Node) Query(ctx context.Context, round uint64, method string, args []by
...
+		if err != nil {
+			return nil, fmt.Errorf("client: failed to get epoch at height %d: %w", annBlk.Height, err)
+		}
+
+		return hrt.Query(ctx, annBlk.Block, lb, epoch, maxMessages, method, args)
 	}
 
-	return hrt.Query(ctx, annBlk.Block, lb, epoch, maxMessages, method, args)
+	data, err := queryFn(round)
+	if protocol.IsConsensusVerifierError(err) {
```
ptrus, line 174: Could this handle a more specific error, e.g. a specific code? (I assume `VerificationFailed`)? Or are there multiple different errors that can happen in case it is failing due to finalized header not being available.

ptrus (approved):

go/worker/client/committee/node.go
kostko: Agreed, updated to use a more specific error, take a look.

ptrus (approved):

PR #5127, opened 2023-01-11
go/common/node: Cleanup VRF configuration as it is mandatory

kostko:
Simplify types by removing the ability for a node's VRF to be nil. Note that in practice (e.g. in all existing deployments) it was already mandatory as registering a node without it set would always fail.


ptrus (approved):

go/common/node/node.go
```diff
@@ -145,19 +145,22 @@ type nodeV2 struct { // nolint: maligned
...
 func (nv2 *nodeV2) ToV3() *Node {
-	return &Node{
+	nv3 := &Node{
 		Versioned:       cbor.NewVersioned(3),
 		ID:              nv2.ID,
 		EntityID:        nv2.EntityID,
 		Expiration:      nv2.Expiration,
 		P2P:             nv2.P2P,
 		Consensus:       nv2.Consensus,
-		VRF:             nv2.VRF,
```
peternose, line 155: Not sure if zero value VRF could be a problem here. We could also blacklist zero value signing key.

peternose (approved):

go/common/node/node.go
kostko: Verification will fail for small-order keys and you need to sign your node registration. So using an all-zero signing key will not get you far.

go/consensus/tendermint/apps/registry/transactions.go
```diff
@@ -462,20 +462,7 @@ func (app *registryApplication) registerNode( // nolint: gocyclo
...
-					return true
-				}
-				if existingNode.VRF != nil && newNode.VRF == nil {
-					return true
-				}
-				return !existingNode.VRF.ID.Equal(newNode.VRF.ID)
-			}()
-
-			if statusDirty = vrfChanged; statusDirty {
+			if !existingNode.VRF.ID.Equal(newNode.VRF.ID) {
```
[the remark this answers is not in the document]
kostko: Done.

PR #5128, opened 2023-01-11
runtime: Refactor consensus verifier predicates

kostko:


runtime/src/consensus/tendermint/verifier/predicates.rs
```diff
@@ -0,0 +1,146 @@
...
+            "epoch seems to have moved backwards"
+        )));
+    }
+
+    // If round has advanced make sure that consensus height has also advanced as a round can
+    // only be finalized in a subsequent consensus block. This is to avoid a situation where
+    // one would keep feeding the same consensus block for subsequent rounds.
+    //
+    // NOTE: This needs to happen before the call to verify_consensus_only which updates the
+    //       cache.last_verified_height field.
```
peternose, line 67: Note could be removed or moved as `cache.last_verified_height` is not updated in the `verify_consensus_only` anymore.

runtime/src/consensus/tendermint/verifier/mod.rs
```diff
@@ -339,48 +336,18 @@ impl Verifier {
...
 
         // Verify the consensus layer block first to obtain an authoritative state root.
-        let consensus_block = self.verify_consensus_only(cache, instance, consensus_block)?;
-        let state_root = consensus_block.get_state_root();
-        let state = ConsensusState::from_protocol(
-            self.protocol.clone(),
-            state_root.version + 1,
-            state_root,
-        );
+        let height = consensus_block.height;
```
peternose, line 345: Can we trust this height? Before these changes the height was taken from the trusted header. Maybe validation should be added to `verify_consensus_only`.
```
// Validate height.
if height != consensus_block.height {
     return Err(Error::VerificationFailed(anyhow!("height mismatch")));
}
```

[nit] No need to introduce a new variable.

runtime/src/consensus/tendermint/verifier/mod.rs
```diff
@@ -469,37 +394,14 @@ impl Verifier {
...
-        // Only does forward verification and fails if height is lower than the last trust height.
-        let height = untrusted_header.header().height.value();
-        let verified_block = self.verify_to_target(height, cache, instance)?;
-
-        // Validate passed consensus block.
-        if untrusted_header.header() != verified_block.signed_header.header() {
-            return Err(Error::VerificationFailed(anyhow!("header mismatch")));
-        }
+        let (consensus_block, state) =
+            self.verify_consensus_only(cache, instance, consensus_block)?;
```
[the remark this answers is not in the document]
kostko: No, the query variant explicitly did not have this.

runtime/src/consensus/tendermint/verifier/mod.rs
kostko: Updated.

peternose (approved):

ptrus (approved):

PR #5140, opened 2023-01-16
runtime/src/rak: Refactor RAK and add REK test key

peternose:
Adding REK test key to be used in (generate entropy) runs/tests on non-SGX platforms.


runtime/src/protocol.rs
```diff
@@ -470,7 +467,7 @@ impl Protocol {
 
         #[cfg(target_env = "sgx")]
         {
-            if self.rak.quote().is_none() {
+            if self.identity.quote().is_none() {
```
kostko, line 470: Could simplify this whole section to `self.identity.quote().ok_or(ProtocolError::...)?;`. And maybe similar above.

kostko (approved):

kostko (approved):

PR #5149, opened 2023-01-18
runtime: Simplify the LRU-backed consensus verifier light store

kostko:


runtime/src/consensus/tendermint/verifier/store/lru.rs
```diff
@@ -94,12 +85,22 @@ impl LightStore for LruStore {
             store.lowest_height = Some(height);
```
peternose, line 85: Lowest height should be renamed to lowest trusted and promoted always except when the newly added is trusted+lower. In that case the update is needed.

Maybe we simplify this by adding `trust_root_height` to the constructor and always promote it, even if it was somehow removed. We need to revert `lowest` method in that case to previous version.
```
// Promote trust root to prevent it from being evicted.
if let Some(trust_root_height) = store.trust_root_height {
    store.blocks.promote(&trust_root_height);
}
```

runtime/src/consensus/tendermint/verifier/store/lru.rs
```diff
@@ -114,40 +115,44 @@ impl LightStore for LruStore {
...
-        let store = inner.store(status);
+        let mut store = self.inner();
 
-        store
-            .lowest_height
-            .and_then(|lowest_height| store.blocks.get(&lowest_height).cloned())
+        store.lowest_height.and_then(|lowest_height| {
+            store
+                .blocks
+                .get(&lowest_height)
```
peternose, line 141: This will return lowest block for trusted status only (or whatever is the status of the lowest block). For others, this will return none.
I would revert this to previous version (copy-paste-edit from `highest`).

peternose (changes requested):

peternose (approved):

ptrus (approved):

PR #5159, opened 2023-02-01
go/consensus/tendermint/apps/keymanager: Fix committee construction

peternose:
Previously, a node was added to the key manager committee if the node's first registered key manager runtime passed validation and matched the key manager status. From now on, all supported versions of the key manager runtime must pass this check.


kostko (commented):
Looks good, just add some unit tests that would fail before and pass with the fix.

kostko (approved):

PR #5161, opened 2023-02-02
go/worker/compute: Do not drop valid proposals

kostko:
Previously valid proposals could be dropped instead of being forwarded via the P2P gossip when the local node's consensus view was slightly behind even though the proposal was valid. With smaller committees and certain topologies this could result in some nodes not getting the proposals.

### DIFF
--- a/go/worker/compute/executor/committee/node.go
+++ b/go/worker/compute/executor/committee/node.go
@@ -37,10 +37,7 @@ import (
 var (
 	errSeenNewerBlock     = fmt.Errorf("executor: seen newer block")
 	errRuntimeAborted     = fmt.Errorf("executor: runtime aborted batch processing")
-	errIncompatibleHeader = p2pError.Permanent(fmt.Errorf("executor: incompatible header"))
 	errBatchTooLarge      = p2pError.Permanent(fmt.Errorf("executor: batch too large"))
-	errIncorrectRole      = fmt.Errorf("executor: incorrect role")
-	errIncorrectState     = fmt.Errorf("executor: incorrect state")
 	errMsgFromNonTxnSched = fmt.Errorf("executor: received txn scheduler dispatch msg from non-txn scheduler")
 
 	// proposeTimeoutDelay is the duration to wait before submitting the propose timeout request.
@@ -117,6 +114,7 @@ type Node struct { // nolint: maligned
 	// Guarded by .commonNode.CrossNode.
 	proposingTimeout bool
 	missingTxsCancel context.CancelFunc
+	pendingProposals *pendingProposals
 
 	commonNode   *committee.Node
 	commonCfg    commonWorker.Config
@@ -206,12 +204,13 @@ func (n *Node) getMetricLabels() prometheus.Labels {
 	}
 }
 
-func (n *Node) queueBatchBlocking(ctx context.Context, proposal *commitment.Proposal) error {
+func (n *Node) processProposal(ctx context.Context, proposal *commitment.Proposal) error {
 	rt, err := n.commonNode.Runtime.ActiveDescriptor(ctx)
 	if err != nil {
 		n.logger.Warn("failed to fetch active runtime descriptor",
 			"err", err,
 		)
+		// Do not forward the proposal further as we are unable to validate it.
 		return p2pError.Permanent(err)
 	}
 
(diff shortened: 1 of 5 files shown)


go/worker/compute/executor/committee/node.go
```diff
@@ -1314,24 +1314,23 @@ func (n *Node) HandleNewEventLocked(ev *roothash.Event) {
...
 	}
 
 	epoch := n.commonNode.Group.GetEpochSnapshot()
 
-	// We can only receive external batches if we are an executor member.
+	// Ignore proposals if we are not an executor committee member, but still forward them.
 	if !epoch.IsExecutorMember() {
-		n.logger.Error("got external batch while in incorrect role")
-		return errIncorrectRole
+		return nil // Forward proposal.
```
ptrus, line 1326: Before, in these cases where a non-permanent p2p error was returned the dispatch was retried right? 

Is it ok that there is now no retry? (e.g. if a node is lagging behind, it could be able to process this message in a followup retry?).

kostko: No, I already changed this back in my local branch but it turns out that actually this retry process is currently implemented incorrectly as the retry could time out just before the proposal would become valid. I'm fixing it locally and will push a new version soon.

kostko: Just pushed the new version, handling of proposal retries is now done in the executor as it knows when it makes sense to retry (e.g. after processing a new block update).

ptrus: :+1: the new version makes sense to me

go/worker/compute/executor/committee/proposals.go
```diff
@@ -0,0 +1,82 @@
...
+	round := batch.proposal.Header.Round
+
+	// Drop any past proposals.
+	if round <= currentRound {
+		return p2pError.Permanent(fmt.Errorf("proposal round is in the past")) // Do not forward.
+	}
+
+	n.pendingProposals.q.ReplaceOrInsert(batch)
+
+	// In case of overflows, remove the round that is the most in the future.
```
peternose, line 45: We should probably drop past proposals first and if we still have too many proposal, then drop max.

```
for b, ok := n.pendingProposals.q.Min(); ok; {
	if b.proposal.Header.Round >= currentRound {
		break
	}
	n.pendingProposals.q.DeleteMin()
}
```

go/worker/compute/executor/committee/proposals.go
```diff
@@ -0,0 +1,82 @@
...
+			return
+		}
+
+		// Fast skip proposals that are no longer valid.
+		if batch.proposal.Header.Round <= currentRound {
+			continue
+		}
+
+		// Skip proposals that fail to be handled.
+		if err := n.handleProposalLocked(batch); err != nil {
```
peternose, line 75: Currently this is not a problem, but in the future we can fall into a loop here as the batch is re-added to the queue. Btw, this error never happens.

go/worker/compute/executor/committee/proposals.go
```diff
@@ -0,0 +1,82 @@
...
+
+	// In case of overflows, remove the round that is the most in the future.
+	if n.pendingProposals.q.Len() >= maxPendingProposals {
+		n.pendingProposals.q.DeleteMax()
+	}
+
+	return nil
+}
+
+// handlePendingProposalsLocked attempts to handle any pending proposals.
```
peternose, line 53: Maybe we should mention that we handle only one proposal, not all of them.

go/worker/compute/executor/committee/proposals.go
kostko: Yes currently the error never happens because we do the quick check above.

go/worker/compute/executor/committee/proposals.go
kostko: Sounds good. The check probably needs to be `>` though as an equal round is invalid (proposal is always for the next round).

go/worker/compute/executor/committee/proposals.go
kostko: I see, this should be removed and it doesn't make sense to ever handle more than one valid proposal (even if handling would fail).

kostko: Fixed.

peternose (approved):

PR #5166, opened 2023-02-06
go/keymanager/api: Move key manager gas costs

peternose:
Moving key manager gas costs from registry to key manager.

### DIFF
--- a/go/keymanager/api/api.go
+++ b/go/keymanager/api/api.go
@@ -53,6 +53,19 @@ var (
 	initResponseContext = signature.NewContext("oasis-core/keymanager: init response")
 )
 
+const (
+	// GasOpUpdatePolicy is the gas operation identifier for policy updates
+	// costs.
+	GasOpUpdatePolicy transaction.Op = "update_policy"
+)
+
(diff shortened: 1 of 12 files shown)


go/keymanager/api/api.go
```diff
@@ -53,6 +53,19 @@ var (
 	initResponseContext = signature.NewContext("oasis-core/keymanager: init response")
 )
 
+const (
+	// GasOpUpdateKeyManager is the gas operation identifier for key manager
+	// policy updates costs.
+	GasOpUpdateKeyManager transaction.Op = "update_keymanager"
```
kostko, line 59: This should probably be called `GasOpUpdatePolicy`.

kostko (approved):

go/keymanager/api/api.go
peternose: Should I also rename `"update_keymanager"` to `"update_policy"` and then remove `"update_keymanager"` in the fix genesis?

kostko: Yeah I think that makes the most sense.

kostko (approved):

PR #5168, opened 2023-02-06
go/consensus/supplementarysanity: Fix checks for legacy validators

ptrus:
Before, when upgrading from 22.2.x the supplementary-sanity checks would fail for validator nodes that did not yet re-register with added P2P address.

(Unrelated change) update `minNodeDescriptorVersion` to 3, because V2 descriptors are automatically migrated to V3 in deserialization, so no need to allow V2 descriptors.

### DIFF
--- a/go/registry/api/api.go
+++ b/go/registry/api/api.go
@@ -706,9 +706,10 @@ func VerifyRegisterNodeArgs( // nolint: gocyclo
 	}
 	expectedSigners = append(expectedSigners, n.P2P.ID)
 	p2pAddressRequired := n.HasRoles(P2PAddressRequiredRoles)
-	switch isGenesis {
+	switch isGenesis || isSanityCheck {
 	case true:
 		// Allow legacy descriptor with optional p2p address for validator.
+		// XXX: Remove this after 23.0.x.
 		if n.HasRoles(node.RoleValidator) {
 			p2pAddressRequired = false
 		}
--- a/.changelog/5168.internal.md
+++ b/.changelog/5168.internal.md
@@ -1,0 +1,1 @@
+go/consensus/supplementarysanity: Fix checks for legacy validators
--- a/go/common/node/node.go
+++ b/go/common/node/node.go
@@ -62,7 +62,7 @@ const (
 	LatestNodeDescriptorVersion = 3
 
 	// Minimum and maximum descriptor versions that are allowed.
-	minNodeDescriptorVersion = 2
+	minNodeDescriptorVersion = 3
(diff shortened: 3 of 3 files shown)


go/registry/api/api.go
```diff
@@ -709,6 +709,7 @@ func VerifyRegisterNodeArgs( // nolint: gocyclo
 	switch isGenesis {
 	case true:
 		// Allow legacy descriptor with optional p2p address for validator.
+		// XXX: Remove this after 22.3.x.
```
kostko, line 712:
```suggestion
		// XXX: Remove this after 23.0.x.
```

go/consensus/tendermint/apps/supplementarysanity/checks.go
```diff
@@ -67,7 +67,16 @@ func checkRegistry(ctx *abciAPI.Context, now beacon.EpochTime) error {
...
 	}
-	_, err = registry.SanityCheckNodes(logger, params, signedNodes, seenEntities, runtimeLookup, false, now, ctx.Now(), uint64(ctx.BlockHeight()))
+
+	baseEpoch, err := ctx.AppState().GetBaseEpoch()
+	if err != nil {
+		return fmt.Errorf("GetBaseEpoch: %w", err)
+	}
+
+	// If legacy nodes registered at genesis can be present, use isGenesis=true version of the sanity check
+	// to allow non-migrated nodes (for 22.3.x these are validator nodes with no p2p address).
```
kostko, line 77:
```suggestion
	// to allow non-migrated nodes (for 23.0.x these are validator nodes with no p2p address).
```

kostko (approved):

peternose:
These changes also force that the public key has to be in the entity's node list for the first `MaxNodeExpiration` epochs (see [546](https://github.com/oasisprotocol/oasis-core/blob/ptrus/fix/validator-p2p-check/go/registry/api/api.go#L546)). Not sure if this is a problem.

kostko:
Note that this is only for supplementary sanity checks which are disabled by default, but yeah this seems like an unintended consequence as it could cause the sanity checks to incorrectly fail. Maybe it would be better to use the `isSanityCheck` flag in `VerifyRegisterNodeArgs`, e.g.:
```golang
	switch isGenesis || isSanityCheck {
	case true:
		// Allow legacy descriptor with optional p2p address for validator.
```

ptrus:
Changed to use the `isSanityCheck` solution. Tried to avoid this before, since it's not the "correct fix", but it does make the logic much simpler, and it's also only a temporary special case anyway.

peternose (approved):

PR #5176, opened 2023-02-10
Multiple minor fixes

kostko:
* tests/workload/registration: Increase iteration timeout
* go/config: Move genesis file path to configuration file

### DIFF
--- a/go/oasis-test-runner/oasis/network.go
+++ b/go/oasis-test-runner/oasis/network.go
@@ -623,10 +623,10 @@ func (net *Network) startOasisNode(
 	cfg.Common.Log.Level["default"] = "debug"
 	cfg.Common.Log.Format = "json"
 	cfg.Common.Log.File = nodeLogPath(node.dir)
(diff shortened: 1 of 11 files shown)


abukosek (approved):

ptrus (approved):

go/oasis-test-runner/oasis/network.go
```diff
@@ -641,6 +641,8 @@ func (net *Network) startOasisNode(
 		cfg.Consensus.UpgradeStopDelay = 10 * time.Second
 
 		extraArgs = extraArgs.debugAllowDebugEnclaves()
+	} else {
+		baseArgs = append(baseArgs, "--genesis.file", net.GenesisPath())
```
peternose, line 645: Not sure why we need this and why only in `else` statement. We already defined the file with `cfg.Genesis.File`.

kostko: It is for some non-node debug subcommands that do their own things and should actually be moved into separate binaries or removed.

PR #5182, opened 2023-02-14
docs: Fix development setup instructions

kostko:
Add missing clang dependency, update SGX tooling installation instructions (nightly no longer required to install tools).

### DIFF
--- a/docs/development-setup/prerequisites.md
+++ b/docs/development-setup/prerequisites.md
@@ -11,7 +11,7 @@ Core:
 * System packages:
   * [Bubblewrap] (at least version 0.3.3).
   * [GCC] (including C++ subpackage).
-  * [Clang] development package. (If you want to build a version after v22.1.9)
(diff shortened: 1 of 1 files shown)


docs/development-setup/prerequisites.md
```diff
@@ -90,24 +90,15 @@ Core:
...
-  cargo +nightly install sgxs-tools
+  cargo install fortanix-sgx-tools
+  cargo install sgxs-tools
   ```
   <!-- markdownlint-enable line-length -->
 
-  _NOTE: These utilities must be compiled with a nightly version of the Rust
-  toolchain since they use the `#![feature]` macro._
-
 * Oasis Core's Rust toolchain version with Fortanix SGX target.
```
peternose, line 102: Do we need to remove/update this section also, as nightly is not needed anymore?

peternose (approved):

PR #5187, opened 2023-02-20
go/runtime: Also re-attest based on MaxAttestationAge

kostko:


go/runtime/host/sgx/sgx.go
```diff
@@ -259,28 +259,22 @@ func (s *sgxProvisioner) getSandboxConfig(rtCfg host.Config, socketPath, runtime
...
 }
 
-func (s *sgxProvisioner) hostInitializer(
-	ctx context.Context,
-	rt host.Runtime,
-	version version.Version,
-	p process.Process,
-	conn protocol.Connection,
-) (*host.StartedEvent, error) {
+func (s *sgxProvisioner) hostInitializer(ctx context.Context, hc *sandbox.HostInitializerContext) (*host.StartedEvent, error) {
```
peternose, line 262: I would maybe rename struct to `xxxParams`, `xxxOptions` or something similar. Don't prefer having two contexts or hiding values inside them.

peternose (approved):

PR #5192, opened 2023-02-21
tests: Multiple txsource fixes

kostko:


ptrus (approved):

go/oasis-node/cmd/debug/txsource/workload/queries.go
```diff
@@ -872,6 +872,14 @@ func (q *queries) Run(
 	for {
 		loopCtx, cancel := context.WithTimeout(ctx, queriesIterationTimeout)
 
+		// Ensure the node appears synced before doing queries. Given that nodes can be killed and
+		// held stopped for a while, they need to sync once they come back.
+		if isSynced, _ := q.control.IsSynced(loopCtx); !isSynced {
+			_ = q.control.WaitSync(loopCtx)
+			time.Sleep(1 * time.Second)
+			continue
```
peternose, line 880: It's a good practice to always cancel context.
We are stealing time from query timeout when syncing, not sure if a problem.
If the node never syncs, we never gracefully exit the loop.
I think one line could be removed.
```
if err = q.control.WaitSync(loopCtx); err != nil {
	time.Sleep(time.Second)
	continue
}
``

PR #5193, opened 2023-02-21
[STABLE/22.2.x] oasis-net-runner: add some useful configuration options

ptrus:
Adds support for disabling debug test entity, configuring epochtime interval and governance parameters to the `oasis-net-runner`.

These are useful/needed for implementing: `22.2.x` -> `master` E2E network upgrade tests.

### DIFF
--- a/go/oasis-test-runner/oasis/entity.go
+++ b/go/oasis-test-runner/oasis/entity.go
@@ -18,7 +18,6 @@ import (
 const entityIdentitySeedTemplate = "oasis entity %d"
 
 var entityArgsDebugTest = []string{
-	"--" + flags.CfgDebugDontBlameOasis,
 	"--" + flags.CfgDebugTestEntity,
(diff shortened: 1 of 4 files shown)


go/oasis-test-runner/oasis/entity.go
```diff
@@ -18,7 +18,6 @@ import (
 const entityIdentitySeedTemplate = "oasis entity %d"
 
 var entityArgsDebugTest = []string{
-	"--" + flags.CfgDebugDontBlameOasis,
```
ptrus, line 21: This flag is now set in `oasis-test-runner/oasis/network.go` because it always needs to be enabled for test-runner networks, regardless if it contains the debug test entity or not.

kostko (approved):
Looks good, do we want the same support in master as well?

ptrus:
> Looks good, do we want the same support in master as well?

Yeah indeed, I will prepare a PR for master as well.

PR #5194, opened 2023-02-22
tests: Add timeout for node sync in queries txsource workload

kostko:


go/oasis-node/cmd/debug/txsource/workload/queries.go
```diff
@@ -874,13 +876,15 @@ func (q *queries) Run(
...
-			_ = q.control.WaitSync(loopCtx)
-			time.Sleep(1 * time.Second)
-			continue
+		switch isSynced, _ := q.control.IsSynced(loopCtx); isSynced {
+		case true:
+			// Node is synced, proceed with queries.
+			err = q.doQueries(loopCtx, rng)
+		case false:
+			// Node is not synced, wait for it to sync (at most for queriesIterationTimeout).
+			err = q.control.WaitSync(loopCtx)
```
peternose, line 885:
```suggestion
			err = q.control.WaitSync(loopCtx)
			q.iteration--
```

peternose: If context expires (or any other error occurs), the error will be handled below and the function will return.

peternose (changes requested):

go/oasis-node/cmd/debug/txsource/workload/queries.go
kostko: Yes this is on purpose as the node should not keep syncing forever.

kostko: I thought about tweaking the iteration count and thought that it shouldn't matter that much, but sure it is better hygiene :)

peternose (approved):

go/oasis-node/cmd/debug/txsource/workload/queries.go
kostko: I changed it around so the iteration counter is only updated in the normal case.

PR #5196, opened 2023-02-23
keymanager/src/runtime: Support master secret rotations

peternose:
**Master secret proposal requirements:**
- A new master secret can be proposed after the rotation period expires and for the next generation only.
	- The first master secret (generation 0) can be proposed immediately and even if the rotation interval is set to 0.
	- If the rotation interval is set to 0, rotations are disabled and secrets cannot be proposed anymore. To enable them again, update the rotation interval in the policy.
- Only nodes from the committee can propose secrets.
- Only one master secret can be proposed per epoch.
	- If accepted, the next secret can be proposed after the rotation interval expires.
	- If not accepted, the next secret can be proposed in the next epoch.

**Master secret rotation requirements:**
- Can happen on every epoch if the following conditions are met:
	- A master secret was proposed for the next generation and the current epoch.
		- The rotation period is not verified here as it is already checked when the secret is proposed.
		- Optionally, we can add this check to cover the case when the policy changes after the secret is proposed.
	- The majority of the enclaves have replicated the proposed secret and announced that in their init status (`next_checksum` field is set).

**Setup:**
- The key manager is initialized with an empty checksum and no nodes.
- Every node needs to register with an empty checksum to be included in the key manager committee.

**Protocol (key manager):**
- Try to rotate the master secret every epoch as part of the key manager status generation.
	- Fetch the latest master secret proposal or abort.
	- Verify the generation number and epoch of the proposal or abort.
	- Count how many nodes have stored the proposal locally.
		- Compare the checksum of the proposal to the `next_checksum` field in the init response.
	- Accept the proposal if the majority of the nodes have replicated the proposal.
		- Increment the generation number by 1.
		- Update the last rotation epoch.
		- Update the checksum.
	- Broadcast the new status.
		- If the generation number has advanced, the enclaves will try to apply the proposal they stored locally.

**Protocol (node):**
- Listen to key manager status updates and master secret publication events.
	- Rotation period and next generation can be computed from the key manager status.
- Wait until the rotation period expires.
- Select a random block in the current epoch.
- When the selected block arrives, generate and propose a master secret for the next generation:
	- Generate a random master secret (in enclave).
	- Encrypt the secret to all enclaves from the key manager committee (in enclave).
	- Compute the next checksum (in enclave).
		- The checksum is computed using hash chains, i.e., `KDF(KDF(...KDF(runtime_id, secret_0), secret_N), proposal)`.
	- Publish a proposal for the next master secret.
		- The transaction contains all ciphertexts, the next checksum, and the next epoch.
- When a new master secret is proposed (either by us or some other node), store it locally:
	- Send the encrypted proposal to the enclave.
	- Decrypt the ciphertext or abort if the secret was not encrypted with the enclave's REK key (in enclave).
	- Encrypt and store the proposal locally (in enclave).
	- Compute the next checksum (in enclave).
	- Initialize the enclave again and register with the updated `next checksum` field.
	- Postpone the master secret generation to the next epoch.
		- If the current proposal is not replicated by the majority of the enclaves, propose another secret in the next epoch.
- When the key manager status is updated, try to apply the proposal if the generation number has advanced:
	- Send the key manager status to the enclave.
	- If the generation number is up-to-date, abort (in enclave).
	- Load the proposal for the next master secret (in enclave):
		- If the proposal is not found, replicate it from another enclave (in enclave).
	- Compute and verify the checksum (in enclave).
		- If the checksum matches, accept the proposal and store it locally (in enclave).
		- Otherwise, abort.
	- Initialize the enclave again and register with updated `checksum` field and empty `next_checksum` field.
	- Postpone the master secret generation after the rotation period expires.


keymanager/src/crypto/kdf.rs
```diff
@@ -288,60 +331,195 @@ impl Kdf {
...
+
+            let next_generation = inner.get_next_generation();
+            let curr_checksum = inner.checksum.clone().unwrap_or(runtime_id.0.to_vec());
+            (next_generation, curr_checksum)
+        };
+
+        // On startup replicate ephemeral secrets.
+        if next_generation == 0 {
+            let last = epoch + 1;
+            for epoch in (0..=last).rev() {
```
kostko, line 392: This seems unbounded, is this safe? I assume the idea is that sooner or later replication will fail and then we will stop. But if it never fails, we should still limit the number of epochs here.

keymanager/src/crypto/kdf.rs
```diff
@@ -288,60 +331,195 @@ impl Kdf {
...
+            // Fetched values are untrusted and need to be verified.
+            let vs = master_secret_fetcher(generation)?;
+            let (secret, prev_checksum) = match vs.checksum.is_empty() {
+                true => (vs.secret, runtime_id.0.to_vec()),
+                false => (vs.secret, vs.checksum),
+            };
+            let next_checksum = Self::checksum_master_secret(&secret, &prev_checksum);
+
+            if next_checksum != last_checksum {
+                return Err(KeyManagerError::MasterSecretChecksumMismatch.into());
```
kostko, line 472: Do I understand right that one bad node could result in the replication process aborting even if another node may have the correct secret? Maybe this validation should be done as part of the node selection logic in the master secret fetcher?

keymanager/src/crypto/kdf.rs
```diff
@@ -674,46 +801,92 @@ impl Kdf {
...
+    ///
+    /// If a proposal already exists, it will be overwritten.
+    fn store_master_secret_proposal(storage: &dyn KeyValue, secret: &Secret) {
+        // Using the same key for all proposals will override the previous one.
+        let key = MASTER_SECRET_PROPOSAL_STORAGE_KEY.to_vec();
 
-        // Encrypt the generation.
+        // Encrypt the master secret.
+        // Additional data has to be different from the one used when storing verified master
+        // secrets so that the attacker cannot replace secrets with rejected proposals.
```
kostko, line 841: Ah, we should add explicit domain separation contexts to `new_d2` similar to how we have it in the `common/sgx/seal` implementation. Currently everything here is just using `MASTER_SECRET_SEAL_CONTEXT` (because previously there were no other data to seal) which should be avoided.

We should also change the seal contexts to follow the new context scheme (they depend on MRENCLAVE anyway so no backward compatibility is needed).

kostko (commented):
Would be good to explicitly test the scenario where master secret replication fails (e.g. not enough nodes confirm to have replicated the master secret) and then succeeds in the next rotation, making sure that nodes properly handle potential reverts.

Are there any backwards compatibility concerns with regards to running old (22.2.x) enclaves with new consensus state due to added fields? Specifically for the case where a transition to the new version is in progress.

keymanager/src/crypto/kdf.rs
pro-wh: could make a more general context and use the additional data to ensure sealed stuff isn't misused

pro-wh (commented):
just looking so far. how does the overall protocol work? what was this about key manager nodes replicating the proposed new key? why is that done on the Go side, i.e. outside the enclave?

keymanager/src/crypto/kdf.rs
kostko: > could make a more general context and use the additional data to ensure sealed stuff isn't misused

Yeah this is how it is currently done in this PR, but it seems less explicit.

go/consensus/tendermint/apps/keymanager/keymanager.go
```diff
@@ -306,15 +333,14 @@ nextNode:
 				// The first version gets to be the source of truth.
 				isInitialized = true
 				isSecure = initResponse.IsSecure
-				checksum = initResponse.Checksum
```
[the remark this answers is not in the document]
peternose: Added a comment stating that the first registrations must be done with an empty checksum.

Initialization is unrelated to rotations, except for the fact that an uninitialized key manager cannot perform rotations as the committee is empty. So we can have an initialized key manager without any secrets.

go/worker/keymanager/worker.go
```diff
@@ -809,11 +902,50 @@ func (w *Worker) randomBlockHeight(epoch beacon.EpochTime, percentile int64) (in
...
+	nextGen := w.kmStatus.NextGeneration()
+	if nextGen != 0 {
+		// Disable rotation if the policy is not set.
+		var rotationInterval beacon.EpochTime
+		if w.kmStatus.Policy != nil {
+			rotationInterval = w.kmStatus.Policy.Policy.MasterSecretRotationInterval
+		}
+
+		// Secrets are allowed to be generated at most one epoch before the rotation.
+		nextEpoch = w.kmStatus.RotationEpoch + rotationInterval - 1
```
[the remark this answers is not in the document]
peternose: Agree. To make the code a bit clearer, I decided to rather merge this with the `if` clause below.

go/worker/keymanager/worker.go
```diff
@@ -809,11 +902,50 @@ func (w *Worker) randomBlockHeight(epoch beacon.EpochTime, percentile int64) (in
...
+
+		// Rotation not allowed.
+		if rotationInterval == 0 {
+			nextEpoch = math.MaxUint64
+		}
```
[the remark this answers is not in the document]
peternose: Cannot move to the `else` section as the policy can also set the rotation interval to 0 (rotations disabled).
Cannot initialize to `math.MaxUint64` because if no secrets have been proposed yet, `nextEpoch` should be 0 (meaning we should generate the secret immediately).

keymanager/src/crypto/kdf.rs
```diff
@@ -288,60 +331,195 @@ impl Kdf {
...
+    /// initialized again.
     ///
-    /// If given checksum and generation don't match internal state,
-    /// the state is reset and an error is returned.
-    pub fn status(
+    /// WARNINGS:
+    /// - Once master secrets have been persisted to disk, it is intended that manual
+    /// intervention by the operator is required to remove/alter them.
+    /// - The first initialization can take a very long time, especially if all generations
+    /// of the master secret must be replicated from other enclaves.
```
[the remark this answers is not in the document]
peternose: Tested, see [ADR 0022](https://github.com/oasisprotocol/adrs/pull/23/files).

keymanager/src/crypto/kdf.rs
peternose: A single bad secret can abort initialization. However, if the node is restarted or initialization is retried, replication will resume where it left off because all verified secrets had been stored locally.
Agree, that we could use another node to replicate the remaining secrets. But in that case a bad node would get away with it as these errors would silently be ignored.
Took the comment into account and changed the fetcher. Now we silently ignore bad secrets.

keymanager/src/crypto/kdf.rs
peternose: Fixed, added/renamed contexts for master secrets and master secret proposals.

go/consensus/tendermint/apps/keymanager/transactions.go
```diff
@@ -111,19 +109,144 @@ func (app *keymanagerApplication) updatePolicy(
...
+	// Reject if the signer is not in the key manager committee.
+	kmStatus, err := state.Status(ctx, kmRt.ID)
+	if err != nil {
+		return err
+	}
+	if !slices.Contains(kmStatus.Nodes, ctx.TxSigner()) {
+		return fmt.Errorf("keymanager: master secret can be published only by the key manager committee")
+	}
+
+	// Reject if the master secret has been proposed in this epoch.
```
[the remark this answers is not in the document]
peternose: This check is required as only one master secret is allowed to be proposed in an epoch, if rotation is allowed.

See [ADR 0022](https://github.com/oasisprotocol/adrs/pull/23/files) for protocol description.

go/consensus/tendermint/apps/keymanager/transactions.go
```diff
@@ -111,19 +109,144 @@ func (app *keymanagerApplication) updatePolicy(
...
+		return fmt.Errorf("keymanager: master secret can be proposed once per epoch")
+	}
+
+	// Reject if rotation is not allowed.
+	if err = kmStatus.VerifyRotationEpoch(secret.Secret.Epoch); err != nil {
+		return fmt.Errorf("keymanager: master secret rotation not allowed: %w", err)
+	}
+
+	// Verify the secret. Master secrets can be published for the next epoch and for
+	// the next generation only.
```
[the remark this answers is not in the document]
peternose: Decided to keep as it is, as rotation interval check is fast.

keymanager/src/crypto/kdf.rs
```diff
@@ -674,46 +801,92 @@ impl Kdf {
...
+        let (ciphertext, nonce) = unpack_encrypted_secret_nonce(&ciphertext)
             .expect("persisted state is corrupted, invalid size");
 
-        // Decrypt the persisted generation.
+        // Decrypt the persisted master secret proposal.
         let d2 = Self::new_d2();
-        let plaintext = d2
-            .open(&nonce, ciphertext.to_vec(), runtime_id)
-            .expect("persisted state is corrupted");
+        let plaintext = match d2.open(&nonce, ciphertext.to_vec(), vec![]) {
```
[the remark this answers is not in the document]
peternose: Changed, now proposals have a separate context and now the additional data is not empty but equal to generation number of the proposal.

peternose:
> Would be good to explicitly test the scenario where master secret replication fails.

Added new test.
 
> Are there any backwards compatibility concerns with regards to running old (22.2.x) enclaves with new consensus state due to added fields? Specifically for the case where a transition to the new version is in progress.

I tested this once (before comments were taken into account) and secret replication worked. Will test that again with cross-version tests.

There should be no compatibility concerns, as new requests (load, generate master secret, ...) should fail when sent to an old enclave and discarded after few retries.

kostko:
> > Could log an error in case of verification failures instead of silently ignoring?
> 
> I thought that we are not logging anything in the `oasis-core-keymanager` project for security reasons, e.g. to not accidentally log secrets. Should I add the logging anyway? I guess I will have to update the secret fetcher and expose the ID of a node that returned the last secret.

That is true. Ok, maybe leave it for now as it should never happen (famous last words).

go/consensus/tendermint/apps/keymanager/transactions.go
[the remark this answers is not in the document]
pro-wh: oh right

go/consensus/tendermint/apps/keymanager/transactions.go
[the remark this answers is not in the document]
pro-wh: ok

go/consensus/tendermint/apps/keymanager/keymanager.go
```diff
@@ -351,19 +391,34 @@ nextNode:
...
 		status.RSK = RSK
-
 		status.Nodes = append(status.Nodes, n.ID)
 	}
 
+	// Accept the proposal if the majority of the nodes have replicated
+	// the proposal for the next master secret.
+	if numNodes := len(status.Nodes); numNodes > 0 && nextChecksum != nil {
+		percent := len(updatedNodes) * 100 / numNodes
+		if percent > minProposalReplicationPercent {
```
pro-wh, line 407: the consensus app has subtly off-by-one logic, it rejects if percent < 66 when deciding to allow a proposal to be published. `>=` would be consistent here

pro-wh (commented):
I've looked through the ADR and matched it up with this implementation. some thoughts on the two together:

- naming clarifications for anyone else reading through:
  - in the tendermint app, the "master secret," a new storage slot in the state, is only the _proposed_ master secret. that's why it's immediately accepted in the publish master secret transaction. for the actual master secret in use, look in the "status."
  - so don't confuse "state" and "status" when you're reading this. "status" is the good thing. "state" is the mechanism that stores the status and other things.
  - verbs used on master secrets:
    - "generate" refers to an enclave sampling an unused master secret at random. it gives it to its node. this one's pretty intuitive
    - "publish" refers to a node giving a proposal to the consensus system
    - "load" refers to a node giving a master secret to its enclave. I need to look more into this
    - "add" is something within the kdf.rs file. I need to look more into this.
    - "rotate" is the action that actually cuts over to another master secret.

there's a check in the consensus layer that a proposal is encrypted for at least 66% of the committee. it can't verify that the payloads are correctly encrypted though, so it's more of a thing to catch unlucky timing for honest nodes. make sure we don't rely on this for security, as a malicious node could enter junk data to pretend that it has encrypted it for many nodes.

I believe there's a confirmation steps where the committee confirms it's able to decrypt the proposal (I haven't walked through the code for this though), so I think we're fine on integrity.

as for availability, it looks like a malicious node could DoS by proposing junk, at the beginning of every epoch. then honest nodes would try to wait for a random time in the epoch so they typically wouldn't compete. is there a mechanism to prevent a node from proposing on every epoch?

the committee proceeds with rotation as long as a majority successfully processes the proposal. is there anything to fill in the secret for the minority that somehow failed? e.g. if they got restarted but hadn't registered their new REK before another node generated a new proposal

kostko:
> there's a check in the consensus layer that a proposal is encrypted for at least 66% of the committee. it can't verify that the payloads are correctly encrypted though, so it's more of a thing to catch unlucky timing for honest nodes. make sure we don't rely on this for security, as a malicious node could enter junk data to pretend that it has encrypted it for many nodes.

That is right, it is only used as a sanity check, then other nodes need to actually confirm that replication/decryption was successful (by updating its status). Note that "malicious nodes" would actually need to be malicious/compromised enclaves. I mentioned this in one of the earlier comments, if one wanted a better guard against compromised enclaves one would need to include a ZKP that the encryption was performed correctly so that the consensus layer could verify.

> I believe there's a confirmation steps where the committee confirms it's able to decrypt the proposal (I haven't walked through the code for this though), so I think we're fine on integrity.

Yes that is right, enough enclaves need to actually confirm that replication/decryption was successful (by updating its status) before the master secret will actually be activated.

> as for availability, it looks like a malicious node could DoS by proposing junk, at the beginning of every epoch. then honest nodes would try to wait for a random time in the epoch so they typically wouldn't compete. is there a mechanism to prevent a node from proposing on every epoch?

This is true and there is currently no such mechanism. Note that it would require a compromised enclave to propose bad encryptions (as the enclaves are actually retrieving REKs from the consensus layer state, encrypting to them and signing the entire proposal with their RAK).

An alternative would be to collect multiple proposals and then randomly select one based on VRF (at epoch boundary?). This would however delay the rotation, but we could maybe start collecting proposals one epoch earlier?

> the committee proceeds with rotation as long as a majority successfully processes the proposal. is there anything to fill in the secret for the minority that somehow failed? e.g. if they got restarted but hadn't registered their new REK before another node generated a new proposal

Yes, the usual master secret replication mechanism that is performed on key manager runtime bootstrapping. There is now also an E2E test for this mechanism as per my earlier comments.

pro-wh:
ah yeah it would have to compromise the enclave. maybe that's fine then

keymanager/src/secrets/interface.rs
```diff
@@ -0,0 +1,17 @@
...
+use crate::crypto::{Secret, VerifiableSecret};
+
+/// Interface for providing master and ephemeral secrets.
+pub trait SecretProvider {
+    /// Returns an iterator that provides access to all the replicas of the master secret
+    /// for the given generation.
+    fn master_secret_iter(
+        &self,
+        generation: u64,
+    ) -> Box<dyn Iterator<Item = VerifiableSecret> + '_>;
```
kostko, line 12: Do these need to be boxed? You could use (and similar below):
```suggestion
    ) -> impl Iterator<Item = VerifiableSecret> + '_;
```

kostko (approved):

keymanager/src/crypto/kdf.rs
```diff
@@ -739,30 +935,35 @@ impl Kdf {
...
-        seal_key.zeroize();
-
-        d2
+    fn derive_signing_key(runtime_id: &Namespace, secret: &Secret) -> signature::PrivateKey {
+        let sec = Inner::derive_secret(secret, &RUNTIME_SIGNING_KEY_CUSTOM, runtime_id.as_ref());
+        signature::PrivateKey::from_bytes(sec.0.to_vec())
     }
 }
 
+fn new_d2(key_policy: Keypolicy, context: &[u8]) -> DeoxysII {
```
kostko, line 959: Should we just export the `new_d2` from `common::sgx::seal` as it seems identical? Or maybe with a slight rename to be more verbose what it is.

keymanager/src/secrets/interface.rs
peternose: I used my solution to avoid the following error:
```
`impl Trait` only allowed in function and inherent method return types, not in trait method return
see issue #91611 <https://github.com/rust-lang/rust/issues/91611> for more information
add `#![feature(return_position_impl_trait_in_trait)]` to the crate attributes to enable
``

kostko: Ah right, not yet stable in traits.

peternose:
Rebased and fixed conflicts. Upgrade test will be added in another PR after merge.

PR #5198, opened 2023-02-24
keymanager: Add support for master secret generations

peternose:
Master secret rotation, 1st part:
- Refactor key manager's init method to be able to support multiple generations of the master secret.

### DIFF
--- a/keymanager/src/api/requests.rs
+++ b/keymanager/src/api/requests.rs
@@ -1,6 +1,10 @@
+use std::sync::Arc;
+
+use anyhow::Result;
+
 use oasis_core_runtime::{
     common::{
-        crypto::signature::{self, Signature},
+        crypto::signature::{self, Signature, Signer},
         namespace::Namespace,
     },
     consensus::{beacon::EpochTime, keymanager::SignedEncryptedEphemeralSecret},
@@ -8,9 +12,14 @@ use oasis_core_runtime::{
 
 use crate::crypto::{KeyPairId, Secret};
 
+/// Context used for the init response signature.
+const INIT_RESPONSE_CONTEXT: &[u8] = b"oasis-core/keymanager: init response";
+
 /// Key manager initialization request.
 #[derive(Clone, Default, cbor::Encode, cbor::Decode)]
 pub struct InitRequest {
+    /// Generation of the latest master secret.
+    pub generation: u64,
     /// Checksum for validating replication.
     pub checksum: Vec<u8>,
     /// Policy for queries/replication.
@@ -41,11 +50,30 @@ pub struct SignedInitResponse {
     pub signature: Signature,
 }
 
+impl SignedInitResponse {
+    /// Create a new signed init response.
+    pub fn new(
+        init_response: InitResponse,
+        signer: &Arc<dyn Signer>,
+    ) -> Result<SignedInitResponse> {
+        let body = cbor::to_vec(init_response.clone());
+        let signature = signer.sign(INIT_RESPONSE_CONTEXT, &body)?;
+
+        Ok(SignedInitResponse {
+            init_response,
+            signature,
+        })
+    }
+}
+
 /// Key manager master secret replication request.
 #[derive(Clone, Default, cbor::Encode, cbor::Decode)]
 pub struct ReplicateMasterSecretRequest {
     /// Latest trust root height.
     pub height: Option<u64>,
+    /// Generation.
+    #[cbor(optional)]
+    pub generation: u64,
 }
 
 /// Key manager master secret replication response.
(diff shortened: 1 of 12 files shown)


keymanager/src/crypto/kdf.rs
```diff
@@ -629,44 +802,86 @@ mod tests {
...
+                    longterm_keys: LruCache::new(NonZeroUsize::new(1).unwrap()),
+                    ephemeral_keys: LruCache::new(NonZeroUsize::new(1).unwrap()),
                 }),
             }
         }
     }
 
+    /// Untrusted key/value store which stores arbitrary binary key/value pairs in memory.
+    pub struct InMemoryKeyValue {
+        cache: Mutex<HashMap<Vec<u8>, Vec<u8>>>,
```
kostko, line 840: Maybe just call it `store` instead of `cache` as it doesn't expire items.

keymanager/src/crypto/kdf.rs
```diff
@@ -998,41 +1262,54 @@ mod tests {
...
-                        v.epoch,
-                        Secret(
-                            v.ephemeral_secret
-                                .from_hex::<Vec<u8>>()
-                                .unwrap()
-                                .try_into()
-                                .unwrap(),
-                        ),
-                    )]),
+                    generation: Some(0),
```
kostko, line 1290: Shouldn't this use `v.generation`?

keymanager/src/runtime/methods.rs
```diff
@@ -54,35 +57,84 @@ const MAX_EPHEMERAL_KEY_AGE: EpochTime = 10;
...
     let policy_checksum = Policy::global().init(ctx, &req.policy)?;
-    Kdf::global().init(req, ctx, policy_checksum)
+
+    let kdf = Kdf::global();
+    kdf.set_runtime_id(runtime_id)?;
+
+    // Load master secrets if not already loaded. It is intended that manual intervention
+    // by the operator is required to remove/alter persisted secrets.
+    kdf.load_master_secrets(storage, &runtime_id)?;
+
```
kostko, line 74: Eventually we should verify the request against consensus layer state.

keymanager/src/runtime/methods.rs
```diff
@@ -54,35 +57,84 @@ const MAX_EPHEMERAL_KEY_AGE: EpochTime = 10;
...
+    let next_generation = kdf.next_generation();
+    if !req.checksum.is_empty() && next_generation <= req.generation {
+        let nodes = key_manager_nodes(ctx, runtime_id)?;
+        for generation in next_generation..=req.generation {
+            let secret = fetch_master_secret(ctx, generation, &nodes)?;
+            kdf.add_master_secret(storage, &runtime_id, secret, generation)?;
+        }
+    }
+
+    // TODO: Allow enclaves to generate master secrets for all generations.
```
kostko, line 85: Why? To support starting with a non-zero generation?

keymanager/src/api/requests.rs
```diff
@@ -1,16 +1,25 @@
...
 use crate::crypto::{KeyPairId, Secret};
 
+/// Context used for the init response signature.
+const INIT_RESPONSE_CONTEXT: &[u8] = b"oasis-core/keymanager: init response";
+
 /// Key manager initialization request.
 #[derive(Clone, Default, cbor::Encode, cbor::Decode)]
 pub struct InitRequest {
+    /// Generation of the latest master secret.
+    pub generation: u64,
```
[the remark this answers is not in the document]
peternose: Have this change (with omitempty) in the last part. Wanted to update the key manager enclave only in this PR.

keymanager/src/api/requests.rs
```diff
@@ -104,16 +132,8 @@ pub struct LongTermKeyRequest {
...
-impl LongTermKeyRequest {
-    pub fn new(height: Option<u64>, runtime_id: Namespace, key_pair_id: KeyPairId) -> Self {
-        Self {
-            height,
-            runtime_id,
-            key_pair_id,
-        }
-    }
+    /// Generation.
+    pub generation: u64,
```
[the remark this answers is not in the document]
peternose: Doesn't need to be optional as the new key manager will never ask the old one for keys, and key manager clients should  upgrade and send non-empty generations after the old key manager is shut down.
We can use optional to save some bandwidth, though.

keymanager/src/crypto/kdf.rs
```diff
@@ -234,220 +288,156 @@ impl Kdf {
...
     ) -> Result<KeyPair> {
         // Construct a seed that must be unique for every key request.
-        // Long-term keys: seed = runtime_id || key_pair_id
         // Ephemeral keys: seed = runtime_id || key_pair_id || epoch
         let mut seed = runtime_id.as_ref().to_vec();
         seed.extend_from_slice(key_pair_id.as_ref());
-        if let Some(epoch) = epoch {
-            seed.extend_from_slice(epoch.to_be_bytes().as_ref());
-        }
+        seed.extend_from_slice(epoch.to_be_bytes().as_ref()); // TODO: Remove once we transition to ephemeral secrets (how?)
```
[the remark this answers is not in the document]
peternose: Yes, this TODO will be handled in that PR (comes after rotations).

keymanager/src/runtime/methods.rs
peternose: Ok, will do it in another PR after rotations.

keymanager/src/runtime/methods.rs
peternose: No, Maybe the comment is not clear. What I meant was that this part is legacy code which should be removed (in the last part) and replaced with the new process: 
- generate and propose the next master secret 
- decrypt and store proposed secret
- if accepted, transition to the next secret

kostko (approved):

PR #5199, opened 2023-02-24
go/keymanager/api: Add signed encrypted master secret

peternose:
Master secret rotation, 2nd part:
- Add signed encrypted master secret.

### DIFF
--- a/go/keymanager/api/secret.go
+++ b/go/keymanager/api/secret.go
@@ -15,6 +15,9 @@ import (
 // committee to which the secret needs to be encrypted.
 const minEnclavesPercent = 66
 
+// EncryptedMasterSecretSignatureContext is the context used to sign encrypted key manager master secrets.
+var EncryptedMasterSecretSignatureContext = signature.NewContext("oasis-core/keymanager: encrypted master secret")
+
 // EncryptedEphemeralSecretSignatureContext is the context used to sign encrypted key manager ephemeral secrets.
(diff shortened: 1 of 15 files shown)


go/keymanager/api/secret.go
```diff
@@ -42,38 +106,47 @@ func (s *EncryptedEphemeralSecret) SanityCheck(epoch beacon.EpochTime, reks map[
...
+	// Secret is the encrypted ephemeral secret.
 	Secret EncryptedEphemeralSecret `json:"secret"`
 
 	// Signature is a signature of the ephemeral secret.
 	Signature signature.RawSignature `json:"signature"`
 }
 
 // Verify sanity checks the encrypted ephemeral secret and verifies its signature.
-func (s *SignedEncryptedEphemeralSecret) Verify(epoch beacon.EpochTime, reks map[x25519.PublicKey]struct{}, rak signature.PublicKey) error {
+func (s *SignedEncryptedEphemeralSecret) Verify(epoch beacon.EpochTime, reks map[x25519.PublicKey]struct{}, rak *signature.PublicKey) error {
```
kostko, line 149: Why this change?

peternose: I plan to introduce a new (helper) function which fetches `rak` key from node's capabilities. The key will be passed to the `Verifiy` function meaning that the array will be copied at least twice. I thought that was too much and decided to use pointers instead, especially as `rak`s are read-only. Should I revert changes?

kostko (approved):

PR #5200, opened 2023-02-24
Move halt epoch from genesis to node-local configuration

kostko:


go/consensus/tendermint/config/config.go
```diff
@@ -30,6 +30,9 @@ type Config struct {
 	// Transaction submission configuration.
 	Submission SubmissionConfig `yaml:"submission,omitempty"`
 
+	// Epoch at which to force-shutdown the node (in epochs).
```
peternose, line 33: Not sure if we need to mention (e.g. in parentheses) that zero disarms shutdown (same for height).

go/consensus/tendermint/abci/state.go
```diff
@@ -65,7 +65,7 @@ type applicationState struct { // nolint: maligned
 	timeSource beacon.Backend
 
 	haltMode        bool
-	haltEpochHeight beacon.EpochTime
+	haltEpoch       beacon.EpochTime
 	haltBlockHeight uint64
```
peternose, line 69: Should we also rename this to `HaltHeight`?

peternose (approved):

PR #5204, opened 2023-02-27
keymanager/src/runtime: Verify and modify init request

peternose:
The init request was never verified against the consensus layer state and, therefore, was not trustworthy. To make this request more informative and easily verifiable against consensus, it was extended to include all key manager status fields.

### DIFF
--- a/runtime/src/types.rs
+++ b/runtime/src/types.rs
@@ -338,6 +338,9 @@ pub struct Features {
     /// A feature specifying that the runtime supports updating key manager's status.
     #[cbor(optional)]
     pub key_manager_status_updates: bool,
+    /// A feature specifying that the runtime supports rotating key manager's master secret.
+    #[cbor(optional)]
+    pub key_manager_master_secret_rotation: bool,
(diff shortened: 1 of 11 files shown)


runtime/src/types.rs
```diff
@@ -346,6 +349,7 @@ impl Default for Features {
             schedule_control: None,
             key_manager_quote_policy_updates: true,
             key_manager_status_updates: true,
+            key_manager_master_secret_rotation: false,
```
kostko, line 352: I assume this is `false` by default because it only makes sense for key manager runtimes? Otherwise it would make sense to have it default to `true` or introduce a different way of querying for key manager features (e.g. a separate local RPC for that) -- not sure if the complexity is worth it though.

peternose: > I assume this is `false` by default because it only makes sense for key manager runtimes?

True. As key manager is a runtime, I thought that I could use these features.

kostko (approved):

PR #5220, opened 2023-03-13
runtime: Refactor Tokio runtime constructor

kostko:


runtime/src/init.rs
```diff
@@ -26,8 +27,11 @@ pub fn start_runtime(initializer: Box<dyn Initializer>, config: Config) {
     // Initialize runtime identity with runtime attestation key and runtime encryption key.
     let identity = Arc::new(Identity::new());
 
+    // Initialize the async Tokio runtime.
+    let tokio_runtime = new_tokio_runtime();
```
peternose, line 31: Maybe cloning outside of the constructor, like identity is.
```suggestion
    let tokio_runtime = new_tokio_runtime().handle();

    // Initialize the dispatcher.
    let dispatcher = Dispatcher::new(tokio_runtime.clone(), initializer, identity.clone());
```

peternose (approved):

PR #5222, opened 2023-03-14
docs: Migrate oasis-node CLI docs to this repo

matevz:
Migrates oasis-node CLI docs from https://github.com/oasisprotocol/docs/tree/5ad98d118f8da3fc0c1fe50964e731f848084501/docs/general/manage-tokens/advanced/oasis-cli-tools to this repository in line of oasisprotocol/docs#394.


kostko:
The oasis-node CLI will be deprecated and eventually removed, I don't think we should move the docs here.

PR #5233, opened 2023-03-21
go/oasis-node: Fix graceful termination

peternose:
This should speed up the runtime integration tests by more than half a minute (5s for every termination).

### DIFF
--- a/go/oasis-node/cmd/common/pprof/pprof.go
+++ b/go/oasis-node/cmd/common/pprof/pprof.go
@@ -2,7 +2,6 @@
 package pprof
 
 import (
-	"context"
 	"fmt"
 	"net"
 	"net/http"
@@ -24,9 +23,6 @@ type pprofService struct {
 
 	listener net.Listener
 	server   *http.Server
-
-	ctx   context.Context
-	errCh chan error
 }
 
 // DumpHeapToFile writes the current process heap to given file with unique suffix.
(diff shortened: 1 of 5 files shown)


go/oasis-node/cmd/common/pprof/pprof.go
```diff
@@ -24,9 +23,6 @@ type pprofService struct {
 
 	listener net.Listener
 	server   *http.Server
-
-	ctx   context.Context
```
peternose, line 28: Never used.

go/oasis-node/cmd/common/pprof/pprof.go
```diff
@@ -79,7 +75,11 @@ func (p *pprofService) Start() error {
 
 	go func() {
 		if err := p.server.Serve(p.listener); err != nil {
-			p.errCh <- err
```
peternose, line 82: If we tried to stop the service, the server was closed, this channel blocked and the service was never stopped.

go/worker/storage/committee/node.go
```diff
@@ -437,10 +440,15 @@ func (n *Node) finalize(summary *blockSummary) {
 		)
 	}
 
-	n.finalizeCh <- finalizeResult{
```
peternose, line 440: When ctx was canceled, this blocked and so did `fetcherGroup.Wait()`.

go/oasis-test-runner/env/env.go
```diff
@@ -192,8 +192,11 @@ func (env *Env) Cleanup() {
 	}
 
 	// Tear down this environment's commands.
-	for _, v := range env.cleanupCmds {
-		v.termOrKill()
+	for _, m := range env.cleanupCmds {
+		go m.termOrKill()
```
peternose, line 196: Are we allowed to terminate these services concurrently?

kostko: This looks fine as it is just sending the signals in parallel.

kostko (approved):

PR #5237, opened 2023-03-22
go/oasis-node/cmd/config: Add the migrate subcommand

abukosek:
A new `migrate` subcommand is added to the node's `config`
command.  This subcommand can be used to automatically migrate
the old YAML config file into the new format introduced in
commit 2a132b3.

The subcommand logs the various changes it makes and warns the
user if a config option is no longer supported, etc.
At the end, any unknown sections of the input config file are
printed to the terminal to give the user a chance to review
them and make manual changes if appropriate.

### DIFF
--- a/go/oasis-node/cmd/config/migrate/migrate.go
+++ b/go/oasis-node/cmd/config/migrate/migrate.go
@@ -1,0 +1,618 @@
+// Package migrate implements the migrate command.
+// nolint: gocyclo,revive,govet,goconst
+package migrate
+
+import (
+	"fmt"
+	"os"
+	"runtime"
+	"sync"
+
+	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
+	"github.com/spf13/viper"
+	"gopkg.in/yaml.v3"
+
+	"github.com/oasisprotocol/oasis-core/go/common/logging"
+	"github.com/oasisprotocol/oasis-core/go/config"
+	cmdCommon "github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common"
+)
+
+const (
+	cfgIn  = "in"
+	cfgOut = "out"
+)
+
+var (
+	migrateCmd = &cobra.Command{
+		Use:   "migrate",
+		Short: "migrate YAML config file into new format",
+		Run:   doMigrateConfig,
+	}
+
+	migrateFlags = flag.NewFlagSet("", flag.ContinueOnError)
+
+	logger = logging.GetLogger("cmd/config/migrate")
+
+	// This is a bit of a kludge to allow running the command from unit tests
+	// multiple times without getting a "logging: already initialized" error.
+	initOnce sync.Once
+)
+
+func pruneEmptyMaps(m map[string]interface{}) {
+	for k, v := range m {
+		if vMap, isMap := v.(map[string]interface{}); isMap {
+			pruneEmptyMaps(vMap)
+			if len(vMap) == 0 {
+				delete(m, k)
+			}
+		}
+	}
+}
+
+func doMigrateConfig(cmd *cobra.Command, args []string) {
+	initOnce.Do(func() {
+		config.GlobalConfig.Common.Log.Level["default"] = "info"
+		if err := cmdCommon.Init(); err != nil {
+			cmdCommon.EarlyLogAndExit(err)
+		}
+	})
+
+	// Perform some sanity checks on the input and output file names.
+	cfgInFileName := viper.GetString(cfgIn)
+	cfgOutFileName := viper.GetString(cfgOut)
+
+	if len(cfgInFileName) == 0 {
+		logger.Error("input file name missing, use the --in flag to specify it")
+		os.Exit(1)
+	}
+	if len(cfgOutFileName) == 0 {
+		logger.Error("output file name missing, use the --out flag to specify it")
+		os.Exit(1)
+	}
+	if cfgInFileName == cfgOutFileName {
+		logger.Error("input and output file names must be different")
+		os.Exit(1)
+	}
+
+	// Start with a blank config that we will populate as we convert the old one.
+	//
+	// NOTE: We don't use config.DefaultConfig() here on purpose, since the
+	// input config file might contain some environment variable substitutions
+	// and would fail to parse as the appropriate type from the structs when
+	// saving it later.
+	// Additionally, we don't want to write any settings that weren't also
+	// present in the old config file to keep the new config file pretty.
+	newCfg := make(map[string]interface{})
+
+	logger.Info("loading input config file", "file_name", cfgInFileName)
+
+	// Load input config file.
+	oldCfgRaw, err := os.ReadFile(cfgInFileName)
+	if err != nil {
+		logger.Error("failed to read input file", "file_name", cfgInFileName, "err", err)
+		os.Exit(1)
+	}
+
+	// Parse old config into map.
+	oldCfg := make(map[string]interface{})
+	err = yaml.Unmarshal(oldCfgRaw, &oldCfg)
+	if err != nil {
+		logger.Error("failed to parse input file", "file_name", cfgInFileName, "err", err)
+		os.Exit(1)
+	}
+
+	logger.Info("input config file loaded successfully", "file_name", cfgInFileName)
+
+	// Helper for making sub-maps.
+	mkSubMap := func(root map[string]interface{}, name string) {
+		if _, ok := root[name]; !ok {
+			root[name] = make(map[string]interface{})
+		}
+	}
+
+	// Helper for casting into maps (we're going to be doing that a lot).
+	m := func(i interface{}) map[string]interface{} {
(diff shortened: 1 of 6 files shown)


kostko (commented):
I would not put this under `debug`.

Also can you add a `cfg` changelog entry that describes how to use this command?

kostko (commented):
Can you also add some quick unit tests for these migrations? E.g. using some dummy configs derived from real ones.

go/oasis-node/cmd/config/migrate/migrate.go
```diff
@@ -0,0 +1,617 @@
...
+		Use:   "migrate",
+		Short: "migrate YAML config file into new format",
+		Run:   doMigrateConfig,
+	}
+
+	migrateFlags = flag.NewFlagSet("", flag.ContinueOnError)
+
+	logger = logging.GetLogger("cmd/config/migrate")
+
+	// This is a bit of a kludge to allow running the command from unit tests
```
kostko, line 36: Would probably be nicer to use a `sync.Once`?

go/oasis-node/cmd/config/migrate/migrate.go
```diff
@@ -0,0 +1,617 @@
...
+
+	// Check for options that are only available on the command-line.
+	if _, ok = oldCfg["debug"]; ok {
+		logger.Warn("note that some debug.* options are from now on only available on the command-line")
+	}
+	if _, ok = oldCfg["grpc"]; ok {
+		logger.Warn("note that grpc.* options are from now on only available on the command-line")
+	}
+	if _, ok = oldCfg["metrics"]; ok {
+		logger.Warn("note that metrics.* options are from now on only available on the command-line")
```
kostko, line 561: Oh, didn't notice that before, why is this the case, doesn't it make sense to have them in the config file?

go/oasis-node/cmd/config/migrate/migrate.go
```diff
@@ -0,0 +1,617 @@
...
+	}
+	if _, ok = oldCfg["tracing"]; ok {
+		logger.Warn("note that tracing.* options are from now on only available on the command-line")
```
kostko, line 564: Which tracing options do we support?

go/oasis-node/cmd/config/migrate/migrate.go
```diff
@@ -0,0 +1,617 @@
...
+						for pk, pv := range m(v) {
+							nk := pk
+							if pk == "persistent_peer" {
+								logger.Info("consensus.tendermint.p2p.persistent_peer is now consensus.p2p.persistent_peers")
+								nk = "persistent_peers"
+							} else if pk == "unconditional_peer" {
+								logger.Info("consensus.tendermint.p2p.unconditional_peer is now consensus.p2p.unconditional_peers")
+								nk = "unconditional_peers"
+							} else if pk == "seed" {
+								logger.Info("consensus.tendermint.p2p.seed is now p2p.seeds")
```
kostko, line 219: Note that the previous version was a bit different -- it took a Tendermint P2P address where it now takes a P2P public key before the `@` part. Unfortunately, these cannot be migrated that easily as the hash function is one-way.

Either we need to support both (in `ConsensusAddressesToTendermint`) or we need to throw an error/warning here (something that will not get lost with all of the other info statements) that this will not work directly.

kostko: Agreed.

go/oasis-node/cmd/config/migrate/migrate.go
abukosek: I guess we could move them into the config in a separate PR...

go/oasis-node/cmd/config/migrate/migrate.go
abukosek: Hmm, you're right, we don't :D  I added this based on some real-world config files I received to test this command with.  I guess the authors of those kept this section in even though it hasn't had any effects.

go/oasis-node/cmd/config/migrate/migrate.go
abukosek: Hmm, good catch, I will add an error here.

abukosek: I've made it emit an error here, as well as leave the `tendermint.p2p.seed` section unmigrated, so it gets printed at the end of the process, which is more noticeable to the user.

peternose (commented):
I think that most of the code could be simplified with a map `old.config.value => new.config.value` and a function which creates all maps on the path, copies values from one config to the other and prints a comment, but not sure.

go/oasis-node/cmd/config/migrate/migrate.go
```diff
@@ -0,0 +1,617 @@
...
+			}
+			delete(m(runtime), "mode")
+		}
+
+		if sandbox, ok := m(runtime)["sandbox"]; ok {
+			if binary, ok := m(sandbox)["binary"]; ok {
+				logger.Info("runtime.sandbox.binary is now runtime.sandbox_binary")
+				mkSubMap(newCfg, "runtime")
+				m(newCfg["runtime"])["sandbox_binary"] = binary
+				delete(m(runtime), "sandbox")
```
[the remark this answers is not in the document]
abukosek: Yes, these sections only have one value and after the migration a pruning step is done which deletes empty sections anyways, so deleting only the value here would have the same effect on a valid config file.

I will change it to delete just the value in case anyone has invalid options in the same section of their config file.

go/oasis-node/cmd/config/migrate/migrate.go
```diff
@@ -0,0 +1,617 @@
...
+						m(newCfg["consensus"])["supplementary_sanity"] = v
+						delete(m(tendermint), k)
+					} else if k == "p2p" {
+						mkSubMap(m(newCfg["consensus"]), "p2p")
+						for pk, pv := range m(v) {
+							nk := pk
+							if pk == "persistent_peer" {
+								logger.Info("consensus.tendermint.p2p.persistent_peer is now consensus.p2p.persistent_peers")
+								nk = "persistent_peers"
+							} else if pk == "unconditional_peer" {
```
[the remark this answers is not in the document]
abukosek: OK, I will also add this option (not sure if we already released a version of the Oasis Node with your new P2P changes? -- probably best to support both `unconditional_peer` and `unconditional_peer_ids`).

go/oasis-node/cmd/config/migrate/migrate.go
```diff
@@ -0,0 +1,610 @@
...
+
+	if len(cfgInFileName) == 0 {
+		logger.Error("input file name missing, use the --in flag to specify it")
+		os.Exit(1)
+	}
+	if len(cfgOutFileName) == 0 {
+		logger.Error("output file name missing, use the --out flag to specify it")
+		os.Exit(1)
+	}
+	if cfgInFileName == cfgOutFileName {
```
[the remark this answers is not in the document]
abukosek: Yeah, this is by design, it's just a safeguard against fatfingering :)

abukosek:
Thank you for your comments!

I also tried to come up with a way to represent and apply the changes in a nicer, more organized way, but the actual changes are too chaotic to do that, so unfortunately this mess of if/else statements is what works best :D

This command will only be useful in the next release.  Afterwards, we can remove it or replace its structure with something nicer if necessary.
I don't think overengineering it at this point is worth the time.

go/oasis-node/cmd/config/migrate/migrate.go
```diff
@@ -0,0 +1,607 @@
...
+	if err = os.WriteFile(cfgOutFileName, newCfgRaw, 0o600); err != nil {
+		logger.Error("failed to write converted config file", "file_name", cfgOutFileName, "err", err)
+		os.Exit(1)
+	}
+
+	// Validate new config.
+	newCfgStruct := config.DefaultConfig()
+	err = yaml.Unmarshal(newCfgRaw, &newCfgStruct)
+	if err != nil {
+		logger.Error("failed to parse config file after migration (this might be normal if you're using environment variable substitutions in your original config file)", "err", err)
```
peternose, line 583: This error message is a bit confusing for me, as I expect the problem is in the original file and not in the converted one. I hope the users won't have the same problem. 

Options:
- Add few extra lines, one printing that we are doing a migration, one that migration was completed, file saved, another that we are validating the migrated file.
- Or split migration/validation, e.g. create a new command and print the command for validation at the end. Maybe even add a warning that the file is not validated. New command could be useful latter to verify already migrated config files.
- Keep as it is.

abukosek: Config validation is done by default now when then node starts up and loads the config file, so I don't think a separate command is needed for that.

I will add a few more info logs to let the user know what's happening.

peternose (approved):
Looks good.

One last thing. I would maybe change that the mode is always present in the config file, so that you know the node's type as soon as you see the config file, without need to know the default value.

PR #5244, opened 2023-04-04
Config migration improvements

abukosek:
- [X] Always print the node mode at the top of the migrated config file.
- [X] Automatically migrate known P2P addresses from the old format to the new (for now implemented only for the mainnet and testnet seed nodes).
- [X] Migrate other P2P addresses as well, but replace the old ID with `INSERT_P2P_PUBKEY_HERE` to make it more obvious to the user that manual changes are required.
- [X] Instruct the user to run `oasis-node control status` to get the P2P public keys for new addresses.
- [x] Add `metrics.*` settings to the config structure.
- [x] Maybe figure out a way to print parts of the YAML config file that don't get deserialized into any structures instead of silently ignoring them when loading the config.

### DIFF
--- a/go/oasis-node/cmd/common/metrics/config/config.go
+++ b/go/oasis-node/cmd/common/metrics/config/config.go
@@ -1,0 +1,61 @@
+// Package config implements global metrics configuration options.
+package config
+
+import (
+	"fmt"
+	"time"
+)
+
+// Config is the metrics configuration structure.
+type Config struct {
+	// Metrics mode (none, pull, push).
+	Mode string `yaml:"mode"`
+	// Metrics pull address.
+	Address string `yaml:"address"`
+
+	// Metrics push job name (debug-only).
+	JobName string `yaml:"job_name,omitempty"`
+	// Metrics push instance labels (debug-only).
+	Labels map[string]string `yaml:"labels,omitempty"`
+	// Metrics push interval (debug-only).
+	Interval time.Duration `yaml:"interval,omitempty"`
+}
+
+// Validate validates the configuration settings.
+func (c *Config) Validate() error {
+	switch c.Mode {
+	case "none":
+	case "pull":
+		if len(c.Address) == 0 {
+			return fmt.Errorf("missing address in pull mode")
+		}
+	case "push":
+		if len(c.Address) == 0 {
+			return fmt.Errorf("missing address in push mode")
+		}
+		if len(c.JobName) == 0 {
+			return fmt.Errorf("missing job_name in push mode")
+		}
+		if len(c.Labels) == 0 {
+			return fmt.Errorf("missing labels in push mode")
+		}
+		if c.Interval == 0 {
+			return fmt.Errorf("missing interval in push mode")
+		}
+	default:
+		return fmt.Errorf("unknown metrics mode: %s", c.Mode)
+	}
+
+	return nil
+}
+
+// DefaultConfig returns the default configuration settings.
+func DefaultConfig() Config {
+	return Config{
+		Mode:     "none",
+		Address:  "127.0.0.1:3000",
+		JobName:  "",
+		Labels:   map[string]string{},
+		Interval: 5 * time.Second,
+	}
+}
--- a/go/oasis-node/cmd/common/metrics/metrics.go
+++ b/go/oasis-node/cmd/common/metrics/metrics.go
@@ -15,21 +15,15 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/client_golang/prometheus/push"
 	flag "github.com/spf13/pflag"
-	"github.com/spf13/viper"
 
 	"github.com/oasisprotocol/oasis-core/go/common/service"
 	"github.com/oasisprotocol/oasis-core/go/common/version"
+	"github.com/oasisprotocol/oasis-core/go/config"
 	"github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common/flags"
 	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/env"
 )
 
 const (
-	CfgMetricsMode     = "metrics.mode"
-	CfgMetricsAddr     = "metrics.address"
-	CfgMetricsLabels   = "metrics.labels"
-	CfgMetricsJobName  = "metrics.job_name"
-	CfgMetricsInterval = "metrics.interval"
-
 	MetricUp = "oasis_up"
 
 	MetricsJobTestRunner = "oasis-test-runner"
@@ -46,9 +40,6 @@ const (
 )
 
 var (
-	// Flags has the flags used by the metrics service.
-	Flags = flag.NewFlagSet("", flag.ContinueOnError)
-
 	UpGauge = prometheus.NewGauge(
 		prometheus.GaugeOpts{
 			Name: MetricUp,
@@ -119,7 +110,7 @@ func (s *pullService) Cleanup() {
 }
 
 func newPullService(ctx context.Context) (service.BackgroundService, error) {
-	addr := viper.GetString(CfgMetricsAddr)
+	addr := config.GlobalConfig.Metrics.Address
 
 	svc := *service.NewBaseBackgroundService("metrics")
 
@@ -139,7 +130,7 @@ func newPullService(ctx context.Context) (service.BackgroundService, error) {
 		ln:                    ln,
 		s:                     &http.Server{Handler: promhttp.Handler(), ReadTimeout: 5 * time.Second},
 		errCh:                 make(chan error),
-		rsvc:                  newResourceService(viper.GetDuration(CfgMetricsInterval)),
+		rsvc:                  newResourceService(config.GlobalConfig.Metrics.Interval),
 	}, nil
 }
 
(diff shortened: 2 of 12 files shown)


go/oasis-test-runner/oasis/network.go
```diff
@@ -657,9 +655,6 @@ func (net *Network) startOasisNode(
 		cfg.Consensus.StateSync.TrustHeight = node.consensusStateSync.TrustHeight
 		cfg.Consensus.StateSync.TrustHash = node.consensusStateSync.TrustHash
 	}
-	if viper.IsSet(metrics.CfgMetricsAddr) {
-		extraArgs = extraArgs.appendNodeMetrics(node)
```
kostko, line 661: How does this work for tests now? During tests we want to be able to collect metrics from all nodes and also the test runner itself.

For example, see [the daily txsource test](https://github.com/oasisprotocol/oasis-core/blob/master/.buildkite/scripts/daily_txsource.sh).

go/oasis-node/cmd/config/migrate/migrate.go
```diff
@@ -50,6 +54,86 @@ func pruneEmptyMaps(m map[string]interface{}) {
...
+		var p2pNew string
+
+		if knownNew, isKnown := known[p2pOld]; isKnown {
+			// New address is known, use it.
+			p2pNew = knownNew
+		} else {
+			// New address is not known, make sure the user replaces the ID.
+			s := strings.Split(p2pOld, "@")
+			addr := s[len(s)-1]
+			p2pNew = fmt.Sprintf("INSERT_P2P_PUBKEY_HERE@%s", addr)
```
peternose, line 86: Using vars we can end up with values that are not expected. But I guess this is still ok, as we don't crash.
Ideally, we would migrate raw configs (fix values) and templates (keep values, show warnings) differently, but that is probably too much.
```
INSERT_P2P_PUBKEY_HERE@{{ p2p_address }}
INSERT_P2P_PUBKEY_HERE@${P2P_ADDRESS}
```

go/oasis-test-runner/oasis/network.go
abukosek: Oh, damn, that's actually used...  I'll fix the test-runner to take cmdline args again and convert them into the config struct internally.

go/oasis-node/cmd/config/migrate/migrate.go
abukosek: I mean, it's not a 100% hands-off automatic process -- you are expected to check the resulting migrated config and make any changes if required :P

go/oasis-node/cmd/common/metrics/config/config.go
```diff
@@ -0,0 +1,35 @@
...
+	JobName string `yaml:"job_name,omitempty"`
+	// Metrics push instance labels (debug-only).
+	Labels map[string]string `yaml:"labels,omitempty"`
+	// Metrics push interval (debug-only).
+	Interval time.Duration `yaml:"interval,omitempty"`
+}
+
+// Validate validates the configuration settings.
+func (c *Config) Validate() error {
+	return nil
```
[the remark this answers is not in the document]
abukosek: I've added some validation here, but I don't want to make it too strict because then if someone has all the fields set and wants to temporarily disable metrics, the validation shouldn't complain about the (now) extra fields...

go/oasis-node/cmd/common/metrics/metrics.go
```diff
@@ -237,20 +228,20 @@ func (s *pushService) initPusher(isReinit bool) {
...
+		labels:                config.GlobalConfig.Metrics.Labels,
+		interval:              config.GlobalConfig.Metrics.Interval,
+		rsvc:                  newResourceService(config.GlobalConfig.Metrics.Interval),
 		stopCh:                make(chan struct{}),
 		quitCh:                make(chan struct{}),
 	}
 
 	if svc.jobName == "" {
-		return nil, fmt.Errorf("metrics: %s required for push mode", CfgMetricsJobName)
+		return nil, fmt.Errorf("metrics: job_name required for push mode")
```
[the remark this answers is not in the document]
abukosek: Added to config validation.
But I think the existing text here is better, because it actually tells you which field is missing.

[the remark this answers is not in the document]
peternose: > existing text here is better

Maybe you're right. I just thought that generalization is a bit better because then one can rename `job_name` to `jobname` without any extra work.

go/oasis-node/cmd/common/metrics/config/config.go
```diff
@@ -0,0 +1,61 @@
...
+	Labels map[string]string `yaml:"labels,omitempty"`
+	// Metrics push interval (debug-only).
+	Interval time.Duration `yaml:"interval,omitempty"`
+}
+
+// Validate validates the configuration settings.
+func (c *Config) Validate() error {
+	switch c.Mode {
+	case "none":
+	case "pull":
```
peternose, line 28: Could use constants here, e.g. by moving them to a separate package `constants` to avoid import cycles.

go/oasis-node/cmd/common/metrics/config/config.go
```diff
@@ -0,0 +1,61 @@
...
+		if len(c.Address) == 0 {
```
peternose, line 29: Could use `_, err = net.ResolveTCPAddr` or something similar, if that is not too strict.

peternose (approved):

go/oasis-node/cmd/common/metrics/config/config.go
abukosek: The address could also be a UNIX socket or something else and the machine that's running the migration command might not be the machine that will actually be running the node :)

I don't want to force a specific type of address here -- the validation for this can happen once the code actually tries to connect to the address.  This is also similar to what's done in other config modules, the really basic stuff can be verified in the config module, but anything more complex should be done in the code that uses it instead.

go/oasis-test-runner/oasis/network.go
```diff
@@ -45,6 +45,11 @@ import (
 	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
 )
 
+const (
+	cfgMetricsAddr     = "metrics.address"
```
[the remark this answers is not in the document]
abukosek: Perhaps, but I think this is out of scope for this PR.  The metrics code is a mess and requires a larger overhaul.

go/oasis-test-runner/oasis/network.go
```diff
@@ -657,8 +662,12 @@ func (net *Network) startOasisNode(
...
 		cfg.Consensus.StateSync.TrustHash = node.consensusStateSync.TrustHash
 	}
-	if viper.IsSet(metrics.CfgMetricsAddr) {
-		extraArgs = extraArgs.appendNodeMetrics(node)
+	if viper.IsSet(cfgMetricsAddr) {
+		cfg.Metrics.Mode = metrics.MetricsModePush
+		cfg.Metrics.Address = viper.GetString(cfgMetricsAddr)
+		cfg.Metrics.Interval = viper.GetDuration(cfgMetricsInterval)
+		cfg.Metrics.JobName = node.Name
+		cfg.Metrics.Labels = metrics.GetDefaultPushLabels(node.net.env.ScenarioInfo())
```
kostko, line 670:
```suggestion
		cfg.Metrics.Labels = metrics.GetDefaultPushLabels(net.env.ScenarioInfo())
```

go/oasis-test-runner/oasis/network.go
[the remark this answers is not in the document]
kostko: Yeah this does seem a bit hackish and the proposal from @peternose seems like it would be a quick fix, e.g.:
* Adding a `MetricsCfg` structure to `NetworkCfg` (it can just have these two fields, e.g. address and interval).
* Just populate it from viper in `doScenario` (in `cmd/root.go`) where you have these already available.
* And then below in `startOasisNode` it can be just configured from `net` instead of using `viper`.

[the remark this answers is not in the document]
abukosek: Ah, you're right, this wasn't as bad as I've imagined -- it's fixed now :)

go/oasis-test-runner/cmd/root.go
```diff
@@ -494,6 +497,12 @@ func doScenario(childEnv *env.Env, sc scenario.Scenario) (err error) {
 		}
 	}
 
+	// Populate metrics settings.
+	if viper.IsSet(cfgMetricsAddr) {
+		net.Config().Metrics.Address = viper.GetString(cfgMetricsAddr)
```
kostko, line 502: The statement below implies that `net` could be nil here (although this is probably never the case in practice as all scenarios use fixtures), so this whole section should be moved into that if body below.

abukosek: Oops, fixed :)

go/oasis-test-runner/cmd/cmp/cmp.go
```diff
@@ -674,6 +676,9 @@ func Register(parentCmd *cobra.Command) {
 	)
 	cmpFlags.String(cfgMetricsNetDevice, "lo", "network device traffic to compare")
 
+	cmpFlags.String(cfgMetricsAddr, "127.0.0.1:3000", "metrics pull address")
```
peternose, line 679: This address flag overwrites the address flag in the `root.go` file on line `cmp.Register(rootCmd)` as they both share the same name. Therefore, condition `viper.IsSet(cfgMetricsAddr)` is never true and metrics don't work. If I change the name of the cmp address flag or don't register the cmp command, then metrics work as expected. 

Tested locally with the following command. 
`.buildkite/scripts/test_e2e.sh -s=e2e.runtime.keymanager-restart --metrics.address 127.0.0.1:9091`

abukosek: Ugh, it's this dumb viper bug again...  I'll make a PR with a workaround for this.

go/oasis-test-runner/cmd/root.go
```diff
@@ -498,6 +501,12 @@ func doScenario(childEnv *env.Env, sc scenario.Scenario) (err error) {
 	// datadir for some scenarios exceed the maximum unix socket path length.
 	if net != nil {
 		net.Config().UseShortGrpcSocketPaths = true
+
+		// Populate metrics settings.
+		if viper.IsSet(cfgMetricsAddr) {
```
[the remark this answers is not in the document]
abukosek: I've just copy&pasted the original code here :)
I think the reason why the original author chose to use `IsSet` is that the default value is not always blank and we only want to set the metrics stuff if the user has set it.

go/oasis-test-runner/cmd/cmp/cmp.go
abukosek: Should be fixed in https://github.com/oasisprotocol/oasis-core/pull/5247 :)

PR #5245, opened 2023-04-11
Cache TCB bundles and lower refreshing frequency

jberci:


go/runtime/host/sgx/ecdsa.go
```diff
@@ -18,6 +20,19 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/runtime/host/protocol"
 )
 
+const (
+	tcbCacheKey = "tcb bundle cache"
```
kostko, line 24:
```suggestion
	tcbCacheKey = "tcb_bundle_cache"
```

go/runtime/host/sgx/ecdsa.go
```diff
@@ -105,6 +110,75 @@ func (ec *teeStateECDSA) Update(ctx context.Context, sp *sgxProvisioner, conn pr
...
+	case nil, persistent.ErrNotFound:
+		// Not cached yet. Not an error.
+	default:
+		// Can't get it... an error, but we can still try downloading it.
+		sp.logger.Warn("error checking common store for cached TCB bundle",
+			"err", err,
+		)
+	}
+
+	if cachedBundle == nil {
```
kostko, line 142: It would be good if this bundle-cache logic was in a separate struct so it could be more easily unit-tested with various bundles/timestamps/etc.

go/runtime/registry/config.go
```diff
@@ -188,7 +189,7 @@ func newConfig(dataDir string, consensus consensus.Backend, ias ias.Endpoint) (*
 					Consensus:         consensus,
 					SandboxBinaryPath: sandboxBinary,
 					InsecureNoSandbox: insecureNoSandbox,
-				})
+				}, commonStore)
```
kostko, line 192: I would rather put this into the `hostSgx.Config` structure, similar to the others.

go/runtime/host/sgx/ecdsa.go
```diff
@@ -66,17 +67,7 @@ func (ec *teeStateECDSA) Init(ctx context.Context, sp *sgxProvisioner, runtimeID
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to get quote: %w", err)
-	}
-
-	var quote pcs.Quote
-	if err = quote.UnmarshalBinary(rawQuote); err != nil {
-		return nil, fmt.Errorf("failed to parse quote: %w", err)
-	}
-
+func (ec *teeStateECDSA) downloadBundle(ctx context.Context, sp *sgxProvisioner, quote pcs.Quote) (*pcs.TCBBundle, error) {
```
kostko, line 70: Pass `*pcs.Quote` instead.

go/runtime/host/sgx/sgx.go
```diff
@@ -85,7 +89,7 @@ type teeStateImpl interface {
 	Init(ctx context.Context, sp *sgxProvisioner, runtimeID common.Namespace, version version.Version) ([]byte, error)
 
 	// Update updates the TEE state and returns a new attestation.
-	Update(ctx context.Context, sp *sgxProvisioner, conn protocol.Connection, report []byte, nonce string) ([]byte, error)
+	Update(ctx context.Context, serviceStore *persistent.ServiceStore, sp *sgxProvisioner, conn protocol.Connection, report []byte, nonce string) ([]byte, error)
```
kostko, line 92: Why is this needed given that it is already available in `sp.serviceStore`?

go/runtime/host/sgx/tcb_cache.go
```diff
@@ -0,0 +1,95 @@
...
+	tcbCacheSlowRefreshInterval = 24 * time.Hour
+)
+
+type tcbBundleCache struct {
+	bundle         *pcs.TCBBundle `json:"bundle"`
+	expectedExpiry string         `json:"expected_expiry"`
+	lastUpdate     string         `json:"last_update"`
+}
+
+func checkTCBCache(store *persistent.ServiceStore, sp *sgxProvisioner) (bundle *pcs.TCBBundle, refresh bool) {
```
kostko, line 24: Similar you can grab the store from `sp`.

go/runtime/host/sgx/sgx.go
```diff
@@ -35,6 +36,9 @@ const (
 	sandboxMountRuntime   = "/runtime"
 	sandboxMountSignature = "/runtime.sig"
 
+	// The service name for the common store to use for TCB cache bookkeeping.
+	serviceStoreName = "sgx_runtime_host"
```
kostko, line 40:
```suggestion
	serviceStoreName = "runtime_host_sgx"
```

go/runtime/host/sgx/tcb_cache.go
```diff
@@ -0,0 +1,95 @@
...
+			"err", err,
+		)
+	}
+
+	// Decide if the cache needs to be refreshed.
+	refresh = true
+	if stored != nil {
+		refresh = func() bool {
+			var expectedExpiry time.Time
+			if expectedExpiry, err = time.Parse(pcs.TimestampFormat, stored.expectedExpiry); err != nil {
```
kostko, line 44: Why parse these each time instead of already storing it as `time.Time`?

kostko (commented):
Nice! Some minor comments.

go/runtime/host/sgx/tcb_cache.go
```diff
@@ -0,0 +1,88 @@
...
+	tcbCacheKey = "tcb_bundle_cache"
+
+	tcbCacheRefreshThreshold    = 14 * 24 * time.Hour
+	tcbCacheSlowRefreshInterval = 24 * time.Hour
+)
+
+type tcbBundleCache struct {
+	bundle         *pcs.TCBBundle `json:"bundle"`
+	expectedExpiry time.Time      `json:"expected_expiry"`
+	lastUpdate     time.Time      `json:"last_update"`
```
kostko, line 21: I believe this will not work (e.g. unmarshalling will just zero-initialize these) as private fields cannot be deserialized properly (so you need to capitalize them to export).

Would be good to have some basic tests of the cache that makes sure storage/retrieval works. To make testing easier you could create a structure (e.g. `tcbCache`) that one can create and holds a service store and logger. Then in tests you can more easily (e.g. using a test service store) test it without the need to create an `sgxProvisioner`.

go/runtime/host/sgx/ecdsa.go
```diff
@@ -129,19 +164,33 @@ func (ec *teeStateECDSA) Update(ctx context.Context, sp *sgxProvisioner, conn pr
...
+		tcbBundle = cachedTCBBundle
+	} else {
+		// If downloaded already, don't try again but just return the last error.
+		if freshTCBBundle != nil {
+			return nil, err
+		}
+
+		// If not downloaded yet this time round, try forcing. Any errors are fatal.
+		var downloadedTCBBundle *pcs.TCBBundle
+		if downloadedTCBBundle, err = ec.downloadBundle(ctx, sp, &quote); err != nil {
```
peternose, line 183: Feels like we are repeating ourselves. Maybe we merge this part with the upper one.
```
	tcbBundle, err = func() (*pcs.TCBBundle, error) {
		if cachedTCBBundle, err := tcbCache.check(); err == nil {
			if err = ec.verifyBundle(quote, quotePolicy, freshTCBBundle, sp, "cached"); err == nil {
				return cachedTCBBundle, nil
			}
		}

		downloadedTCBBundle, err := ec.downloadBundle(ctx, sp, &quote);
		if err != nil {
			return nil, err
		}
		if err = ec.verifyBundle(quote, quotePolicy, freshTCBBundle, sp, "cached"); err != nil {
			return nil, err
		}
		tcbCache.cache(downloadedTCBBundle)
		return downloadedTCBBundle, nil
	}()
	if err != nil {
		return nil, err
	}
```

peternose (commented):
Left few comments, ignore those which you find irrelevant.

go/runtime/host/sgx/tcb_cache.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+	}
+
+	// Decide if the cache needs to be refreshed.
+	refresh = true
+	if stored != nil {
+		refresh = func() bool {
+			now := tc.now()
+
+			// Wait for the first two weeks, then check once daily.
```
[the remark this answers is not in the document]
jberci: I think it is: if the bundle is expected to be valid for around a month, then there shouldn't be any fetches for the first two weeks of that, in the last two weeks they should be daily, and after expiration there shouldn't be any throttling.

go/runtime/host/sgx/ecdsa.go
jberci: I'm not sure the snippets are equivalent: the download must not be unconditional, but if it succeeds, its bundle should take precedence over whatever was in the cache.

peternose: You are right. I thought that refresh is mandatory, but is not, as we use the cached bundle if download fails, which fine.

go/runtime/host/sgx/ecdsa.go
```diff
@@ -129,19 +150,48 @@ func (ec *teeStateECDSA) Update(ctx context.Context, sp *sgxProvisioner, conn pr
...
+				sp.logger.Warn("error downloading TCB refresh",
+					"err", err,
+				)
+			}
+		}
+
+		if err = ec.verifyBundle(quote, quotePolicy, fresh, sp, "fresh"); err == nil {
+			ec.tcbCache.cache(fresh)
+			return fresh, nil
+		} else if err = ec.verifyBundle(quote, quotePolicy, cached, sp, "cached"); err == nil {
```
peternose, line 171:
```suggestion
		}
		if err = ec.verifyBundle(quote, quotePolicy, cached, sp, "cached"); err == nil {
```

go/runtime/host/sgx/ecdsa.go
```diff
@@ -129,19 +150,48 @@ func (ec *teeStateECDSA) Update(ctx context.Context, sp *sgxProvisioner, conn pr
...
+			return cached, nil
+		} else {
```
peternose, line 173:
```suggestion
```

peternose (approved):

go/runtime/host/sgx/tcb_cache.go
```diff
@@ -0,0 +1,119 @@
...
+	var err error
+
+	// Check if we have a copy in the local store.
+	var stored *tcbBundleCache
+	switch err = tc.serviceStore.GetCBOR([]byte(tcbCacheKey), &stored); err {
+	case nil, persistent.ErrNotFound:
+		// Not cached yet. Not an error.
+	default:
+		// Can't get it... an error, but we can still try downloading it.
+		if tc.logger != nil {
```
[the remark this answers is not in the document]
ptrus: Or also create the logger in the `mockTbcCache`. I think logging an error here makes sense, since this is an internal to cache error, that's intended for the node operator not for the calling code.

go/runtime/host/sgx/tcb_cache.go
```diff
@@ -0,0 +1,115 @@
...
+			tc.logger.Warn("error checking common store for cached TCB bundle",
+				"err", err,
+			)
+		}
+	}
+
+	// Decide if the cache needs to be refreshed.
+	if err != nil {
```
ptrus, line 49: I think it would be clearer if this conditional is moved into the above switch? e.g.
```golang
switch err {
  case nil:
     // No error, continues below.
  case persistent.ErrNotFound:
    // Not cached yet, not an error, but needs refresh.
    return nil, true
  default:
    // Log the error and needs refresh.
    ...
    return nil, true
}
```

peternose (approved):
Now, we can probably remove this [line](https://github.com/oasisprotocol/oasis-core/blob/master/go/common/sgx/pcs/http.go#L88). 

Tested on an sgx instance. The bundle is downloaded, cached and used on restarts :+1:

kostko (approved):

kostko (approved):

PR #5246, opened 2023-04-11
changelog: Fix outdated links in README

kostko:


pro-wh (approved):
indeed it does redirect there

kostko:
Yeah seems it was a leftover since we transferred the repo a few years ago. Only noticed it because I was looking at our changelog setup to see if we could use something for the SDK.

PR #5255, opened 2023-04-24
.buildkite: Switch CI jobs to using /dev/sgx_enclave devices

gw0:
Because Intel will eventually roll-out SGX2, the new Buildkite agents with `/dev/sgx_enclave` execute jobs with `queue=sgx1`. The old Buildkite agents with `/dev/isgx` execute jobs with `queue=intel-sgx`. This PR switches SGX CI jobs to the new Buildkite agents and renames them.

### DIFF
--- a/.buildkite/code.pipeline.yml
+++ b/.buildkite/code.pipeline.yml
@@ -30,7 +30,7 @@ docker_plugin_default_config: &docker_plugin_default_config
     # Propapage the tmpfs mount.
     - /tmp:/tmp
     # NOTE: When changing the environment variables below, also copy the changes
-  # to the docker_plugin_sgx_config.
+    # to the docker_plugin_sgx1_config.
   environment:
     - "LC_ALL=C.UTF-8"
     - "LANG=C.UTF-8"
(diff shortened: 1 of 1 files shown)


peternose (approved):

kostko:
Note that this has nothing to do with SGX1 vs. SGX2. It just has to do with the legacy out of tree SGX kernel driver, where `/dev/isgx` is the legacy OOT drive while `/dev/sgx_enclave` is the in-tree driver. So even SGX1 with the in-tree driver (e.g. a recent enough Linux kernel) will use `/dev/sgx_enclave`. The legacy driver should no longer be used anywhere in new installs.

kostko:
Given this I would avoid using "sgx1" in configuration names.

gw0:
> Given this I would avoid using "sgx1" in configuration names.

Of course, both are SGX1. But eventually SGX2 will come and then we will need a new name for that (to distinguish which jobs require which version).

We just need a different queue name for a smooth transition from the old legacy driver to the new in-tree driver (without renaming hacks). What queue name would you suggest?

kostko (approved):
Oh, I see, then it makes sense, sorry for the confusion.

kostko:
Also don't forget about the stable branch.

PR #5258, opened 2023-05-04
.buildkite: Improve namespacing of CI caches and build artifacts

gw0:
Experimental attempt to reorganize and improve namespacing of CI caches and build artifacts. By introducing explicit levels at which local caches and artifacts can be shared (`global_cache`, `branch_cache`, `build_cache`), we should be able to support multiple Buildkite agents per node and increase their utilization.

### DIFF
--- a/.buildkite/benchmarks.pipeline.yml
+++ b/.buildkite/benchmarks.pipeline.yml
@@ -6,20 +6,26 @@ docker_plugin_default_config: &docker_plugin_default_config
   volumes:
     - /var/lib/buildkite-agent/.coveralls:/root/.coveralls
     - /var/lib/buildkite-agent/.codecov:/root/.codecov
-    # Shared Rust incremental compile caches.
-    - /var/tmp/cargo_ic/debug:/var/tmp/artifacts/debug/incremental
-    - /var/tmp/cargo_ic/debug_sgx:/var/tmp/artifacts/x86_64-unknown-linux-sgx/debug/incremental
-    # Shared Rust package checkouts directory.
-    - /var/tmp/cargo_pkg/git:/root/.cargo/git
-    - /var/tmp/cargo_pkg/registry:/root/.cargo/registry
-    # Shared Rust SGX standard library artifacts cache.
-    - /var/tmp/xargo_cache:/root/.xargo
-    # Shared Go package checkouts directory.
-    - /var/tmp/go_pkg:/root/go/pkg
     # Intel SGX Application Enclave Services Manager (AESM) daemon running on
     # the Buildkite host.
     - /var/run/aesmd/aesm.socket:/var/run/aesmd/aesm.socket
-    - /var/tmp/benchmarks:/var/tmp/benchmarks
+    # Shared Go package checkouts directory.
+    - /storage/buildkite/global_cache/go_pkg:/root/go/pkg
+    # Shared Rust package checkouts directory.
+    - /storage/buildkite/global_cache/cargo_git:/root/.cargo/git
+    - /storage/buildkite/global_cache/cargo_registry:/root/.cargo/registry
+    # Shared Rust SGX standard library artifacts cache.
+    - /storage/buildkite/global_cache/xargo_cache:/root/.xargo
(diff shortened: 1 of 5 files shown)


kostko (commented):
Why the timeout changes?

peternose:
> Why the timeout changes?

Maybe he is trying to fix (experiment) timeout errors:
https://buildkite.com/oasisprotocol/oasis-core-ci/builds?branch=master
https://buildkite.com/oasisprotocol/oasis-core-ci/builds/11175#_

gw0:
> Why the timeout changes?

Because downloading of the Buildkite artifacts is taking too long...

Our old Buildkite agents were located in Azure `eastus` region and required <1 min to download all the binary artifacts from Buildkite's S3 bucket. Our new Buildkite agents are located on a bare-metal provider in `eu-am1` region and require 8-18 minutes to download them. Then there is to little time left for complete the E2E test... Example difference for downloading just a single artifact:

```
## oasisprotocol-buildkitesgx-medium00001J (in azure/eastus)
+ buildkite-agent artifact download simple-keymanager-upgrade.sgxs .
2023-04-20 16:37:30 INFO   Searching for artifacts: "simple-keymanager-upgrade.sgxs"
2023-04-20 16:37:36 INFO   Found 1 artifacts. Starting to download to: /tmp/artifacts
2023-04-20 16:37:37 INFO   Successfully downloaded "simple-keymanager-upgrade.sgxs" 25464960 bytes

## buildkite0-default-sgx1-9 (in bare-metal/eu-am1)
+ buildkite-agent artifact download simple-keymanager-upgrade.sgxs .
2023-04-20 20:32:04 INFO   Searching for artifacts: "simple-keymanager-upgrade.sgxs"
2023-04-20 20:32:14 INFO   Found 1 artifacts. Starting to download to: /tmp/artifacts
2023-04-20 20:34:52 INFO   Successfully downloaded "simple-keymanager-upgrade.sgxs" 25464960 bytes
```

My hypothesis is that Buildkite's S3 bucket is located in AWS `us-east-?` region and it is slower to download to another region. Another option is that AWS has a DDoS protection on S3 buckets and 10 Buildkite agents attempting to download the same files triggers it.

As a workaround I implemented trivial caching of these downloaded Buildkite agents in PR https://github.com/oasisprotocol/oasis-core/pull/5251 (no clue why this isn't a standard Buildkite feature). But it is not enough, and they still fail the first time.

Another way to improve the situation would also be to compress the artifacts, because individual files are very large (~150MB) and they can be compressed to 1/3rd of their size (no clue why this isn't a standard Buildkite feature to compress them during transfer).

This PR goes in the direction of supporting multiple Buildkite agents per server (with the goal that all E2E jobs start simultaneously). By sharing these binary artifacts between all agents on a server the impact that it takes so long to download them is minimized. I already experimented by running them simultaneously, but there is a race condition between parallel jobs downloading binary artifacts and causing them to still be downloaded multiple times -- this is why it is still a draft PR.

kostko:
Should we also consider using our own storage provider for the artifacts, located closer? See the [Buildkite documentation](https://buildkite.com/docs/pipelines/artifacts#storage-providers-encryption-and-retention).

gw0:
Solved, downloading Buildkite artifacts from our S3 bucket that is closer to our Buildkite agents solved the timeout issues. Downloading all Buildkite artifacts now typically takes around 30 seconds and up to 2.5 minutes. If the artifacts are already in the local cache (introduced in https://github.com/oasisprotocol/oasis-core/pull/5255) it typically takes 8 seconds.

Will now test if everything still works when running multiple Buildkite agents per node...

gw0:
Still need to tweak the timeouts, the rest is ready for review. It seems all E2E tasks can now run in parallel which is a huge speed-up, e.g.:
- https://buildkite.com/oasisprotocol/oasis-core-ci/builds/11308

gw0:
Set timeouts of all E2E tests to 20 minutes (some were 15 minutes), because they all run in parallel and proceeding CI jobs need to wait for all to finish. Most E2E tests finish in 4-11 min, but their duration is non-deterministic and occasionally gets as high as 19 min.

PR #5267, opened 2023-05-22
tests: Support cross-version upgrade tests

peternose:


tests/upgrade/README.md
```diff
@@ -0,0 +1,90 @@
...
+)
+```
+
+## Write tests
+
+1. Write and register the pre-upgrade test scenario by placing the code in the
+`./tests/upgrade/pre/scenario/e2e` folder.
+
+2. Write and register the post-upgrade test scenario using the same scenario
+name, and place the code in the `./tests/upgrade/post/scenario/e2e folder`.
```
kostko, line 90:
```suggestion
name, and place the code in the `./tests/upgrade/post/scenario/e2e` folder.
```

tests/upgrade/post/Makefile
```diff
@@ -0,0 +1,30 @@
...
+	GO_EXTRA_FLAGS += -tags jemalloc
+endif
+
+# Set all target as the default target.
+all: build
+
+# Build.
+build:
+	@$(ECHO) "$(MAGENTA)*** Building oasis-test-runner...$(OFF)"
+	@$(GO) build $(GOFLAGS) $(GO_EXTRA_FLAGS) -o oasis-test-runner
```
kostko, line 19: Should this be named something different to avoid confusion with the "regular" test runner? Maybe something like `oasis-test-upgrade`?

tests/upgrade/pre/Makefile
```diff
@@ -0,0 +1,30 @@
...
+	GO_EXTRA_FLAGS += -tags jemalloc
+endif
+
+# Set all target as the default target.
+all: build
+
+# Build.
+build:
+	@$(ECHO) "$(MAGENTA)*** Building oasis-test-runner...$(OFF)"
+	@$(GO) build $(GOFLAGS) $(GO_EXTRA_FLAGS) -o oasis-test-runner
```
kostko, line 19: Same as above.

.buildkite/scripts/test_upgrade.sh
```diff
@@ -0,0 +1,128 @@
...
+# test_upgrade.sh
+############################################################
+
+# Helpful tips on writing build scripts:
+# https://buildkite.com/docs/pipelines/writing-build-scripts
+set -euxo pipefail
+
+# Branches to test.
+pre_upgrade_git_branch="peternose/stable/22.2.x/backport-5265"
+post_upgrade_git_branch="peternose/internal/upgrade-tests-refactoring"
```
kostko, line 16: Make sure to change this before merging.

tests/upgrade/post/go.mod
```diff
@@ -0,0 +1,202 @@
...
+	github.com/google/flatbuffers => github.com/google/flatbuffers v1.12.1
+
+	// v1.5.0 has broken uint parsing, use my commit with fixes instead until
+	// the maintainers merge my PR: https://github.com/spf13/cast/pull/144
+	github.com/spf13/cast => github.com/oasisprotocol/cast v0.0.0-20220606122631-eba453e69641
+
+	github.com/tendermint/tendermint => github.com/oasisprotocol/tendermint v0.34.23-oasis2
+
+	// Required because of https://github.com/libp2p/go-libp2p-pubsub/issues/467.
+	github.com/whyrusleeping/timecache => github.com/oasisprotocol/timecache v0.0.0-20220102191729-558b1c931038
```
kostko, line 19: Unrelated to this PR, but this can be removed (also from master) as go-libp2p-pubsub 0.9.3 no longer uses this.

tests/upgrade/post/go.mod
```diff
@@ -0,0 +1,202 @@
...
+	// module caches apparently have a messed up copy that causes
+	// build failures.
+	// https://github.com/google/flatbuffers/issues/6466
+	github.com/google/flatbuffers => github.com/google/flatbuffers v1.12.1
+
+	// v1.5.0 has broken uint parsing, use my commit with fixes instead until
+	// the maintainers merge my PR: https://github.com/spf13/cast/pull/144
+	github.com/spf13/cast => github.com/oasisprotocol/cast v0.0.0-20220606122631-eba453e69641
+
+	github.com/tendermint/tendermint => github.com/oasisprotocol/tendermint v0.34.23-oasis2
```
kostko, line 16: This will need to be changed after rebase to master (now uses the new CometBFT repository).

tests/upgrade/post/go.mod
```diff
@@ -0,0 +1,202 @@
+module github.com/oasisprotocol/oasis-core/test-upgrade
+
+go 1.19
+
+replace (
+	// Updates the version used by badgerdb, because some of the Go
+	// module caches apparently have a messed up copy that causes
+	// build failures.
+	// https://github.com/google/flatbuffers/issues/6466
+	github.com/google/flatbuffers => github.com/google/flatbuffers v1.12.1
```
kostko, line 10: Is this still needed?

tests/upgrade/pre/scenario/e2e/runtime/trust-root-km-upgrade.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	return f, nil
+}
+
+func (sc *trustRootKeymanagerUpgradeImpl) Run(childEnv *env.Env) error {
+	ctx := context.Background()
+
+	// Build a simple key/value runtime with embedded trust root, start the network, register
+	// runtimes and run the test client.
```
kostko, line 80: I assume this does not also build the key manager with trust root? We should do that as well to make sure the key manager also survives (should be the same as runtime, but in production both have trust roots configured).

peternose: That is correct. In fact, our key manager tests have three issues that need to be addressed:
1. We have not been testing the key manager runtime with an embedded trust root, as mentioned above.
2. We are testing insecure operations that do not require the key manager at all. For example, the trust root scenario uses the unencrypted `SimpleKeyValueScenario`.
3. We need to ensure that the compute nodes clear the cache whenever changes are made to the key manager. This is optional because currently, a faulty key manager can cause tests to fail with a consensus failure if at least one of the nodes has an empty cache.

These problems will be addressed in another pull request. The focus is on setting up the environment for upgrade tests.

tests/upgrade/post/go.mod
peternose: No, the badgerdb uses the same version.

See https://github.com/dgraph-io/badger/blob/v3.2103.4/go.mod#L14.

tests/upgrade/pre/scenario/e2e/runtime/trust-root-km-upgrade.go
kostko: Agreed, I've created tasks for these items.

kostko (approved):

PR #5268, opened 2023-05-23
go: Switch to CometBFT v0.34.27

abukosek:
This PR switches from our fork of Tendermint to our fork of CometBFT, which is a drop-in replacement.

Here is our fork of CometBFT that contains the exact same patches as our Tendermint fork: https://github.com/oasisprotocol/cometbft/tree/v0.34.27-oasis1


go/consensus/tendermint/db/badger/badger.go
```diff
@@ -9,10 +9,10 @@ import (
 	"strings"
 	"sync"
 
+	dbm "github.com/cometbft/cometbft-db"
```
peternose, line 12:
```suggestion
	cdb "github.com/cometbft/cometbft-db"
```
Have no idea where `dbm` comes from :)

go/go.mod
```diff
@@ -71,11 +71,10 @@ require (
 )
 
 require (
-	github.com/DataDog/zstd v1.4.1 // indirect
 	github.com/Workiva/go-datastructures v1.0.53 // indirect
 	github.com/benbjohnson/clock v1.3.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
-	github.com/btcsuite/btcd v0.22.1 // indirect
+	github.com/btcsuite/btcd/btcec/v2 v2.2.1 // indirect
```
kostko, line 77: :tada:

kostko (approved):

peternose (commented):
Running `go mod tidy` returns an error.

```
➜  go git:(821075e18) ✗ go mod tidy
github.com/oasisprotocol/oasis-core/go/consensus/tendermint/abci imports
        github.com/tendermint/tendermint/abci/types imports
        github.com/tendermint/tendermint/crypto/secp256k1 imports
        github.com/btcsuite/btcd/btcec/v2/ecdsa tested by
        github.com/btcsuite/btcd/btcec/v2/ecdsa.test imports
        github.com/btcsuite/btcd/chaincfg/chainhash: ambiguous import: found package github.com/btcsuite/btcd/chaincfg/chainhash in multiple modules:
        github.com/btcsuite/btcd v0.20.1-beta (/home/peter/go/pkg/mod/github.com/btcsuite/btcd@v0.20.1-beta/chaincfg/chainhash)
        github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1 (/home/peter/go/pkg/mod/github.com/btcsuite/btcd/chaincfg/chainhash@v1.0.1)

**Comment by kostko** [comment]
> Running `go mod tidy` returns an error.

That's strange, we run this as part of CI?

File: go/consensus/tendermint/db/badger/badger.go (hunk 1 above)
Comment (abukosek) [reply go/consensus/tendermint/db/badger/badger.go:12]: We use this elsewhere for the Tendermint DB import, so I standardized on it :)

Comment (peternose) [reply go/consensus/tendermint/db/badger/badger.go:12]: Your changes make sense, but not sure why was named `dbm` in the first place. As we switched to `cometbft`, we could rename to `cdb` or something similar everywhere.

**Comment by abukosek** [comment]
> Running `go mod tidy` returns an error.

I have no idea how to fix this, it seems like a problem in an upstream package or `go mod tidy` itself.

**Comment by abukosek** [comment]
> > Running `go mod tidy` returns an error.
> 
> I have no idea how to fix this, it seems like a problem in an upstream package or `go mod tidy` itself.

This will probably fix it: https://github.com/golang/go/issues/60352
But I'm not waiting for 1.20.5 to be released just for this, going to go ahead with the merge now :)

File: go/consensus/tendermint/db/badger/badger.go (hunk 1 above)
Comment (abukosek) [reply go/consensus/tendermint/db/badger/badger.go:12]: What if they rename it again?

**Comment by peternose** [comment]
> That's strange, we run this as part of CI?

We should probably fix `common.mk` (replace ; with &&):
```
$(GO) mod tidy; \
```

**Comment by abukosek** [comment]
> We should probably fix `common.mk` (replace ; with &&):
> 
> ```
> $(GO) mod tidy; \
> ```

We can do this after Go 1.20.5 is released, which should have a fix for this issue, otherwise all our builds will keep failing...

**Comment by peternose** [comment]
> I have no idea how to fix this, it seems like a problem in an upstream package or `go mod tidy` itself.

I think you need to specify which version to use. I guess that should be 1.0.1, but I'm not sure if this breaks anything.

One solution is to clear the `require` section in the `go.mod` file and then run the following commands:
```
go get github.com/btcsuite/btcd/chaincfg/chainhash@v1.0.1
go mod tidy
```

PR #5261, opened 2023-05-16
go/roothash: Treat failure indications as stragglers

kostko:

PR #5271, opened 2023-05-24
go/roothash: Wait for proposer commit in discrepancy resolution

kostko:
Fixes an issue introduced in #5261.

### DIFF
--- a/go/roothash/api/commitment/pool.go
+++ b/go/roothash/api/commitment/pool.go
@@ -449,6 +449,12 @@ func (p *Pool) ProcessCommitments(didTimeout bool) (OpenCommitment, error) {
 			return nil, ErrMajorityFailure
 		}
 
+		// If we already have the proposer commitment, we can try to resolve early based on majority
+		// vote. Otherwise we need to wait for the proposer commitment.
+		if proposerCommit == nil {
+			break
+		}
+
 		for _, ent := range votes {
 			if ent.tally >= minVotes {
 				// Majority agrees on a commit, result is determined regardless of additional
@@ -456,9 +462,6 @@ func (p *Pool) ProcessCommitments(didTimeout bool) (OpenCommitment, error) {
 				//
 				// Make sure that the majority commitment is the same as the proposer commitment. We
 				// must return the proposer commitment as that one contains additional data.
-				if proposerCommit == nil {
-					return nil, ErrNoProposerCommitment
-				}
 				if !proposerCommit.MostlyEqual(ent.commit) {
 					return nil, ErrBadProposerCommitment
 				}
@@ -503,7 +506,7 @@ func (p *Pool) ProcessCommitments(didTimeout bool) (OpenCommitment, error) {
 		}
 	}
 
-	if commits < required {
+	if commits < required || proposerCommit == nil {
 		return nil, ErrStillWaiting
 	}
 
--- a/go/roothash/api/commitment/pool_test.go
+++ b/go/roothash/api/commitment/pool_test.go
@@ -743,14 +743,14 @@ func TestPoolTwoCommitments(t *testing.T) {
 }
 
 func TestPoolFailureIndicatingCommitment(t *testing.T) {
+	genesisTestHelpers.SetTestChainContext()
+
 	rt, sks, committee, nl := generateMockCommittee(t, nil)
 	sk1 := sks[0]
 	sk2 := sks[1]
 	sk3 := sks[2]
 
 	t.Run("FailureIndicating", func(t *testing.T) {
-		genesisTestHelpers.SetTestChainContext()
-
 		// Create a pool.
 		pool := Pool{
 			Runtime:   rt,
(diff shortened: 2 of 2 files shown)


go/roothash/api/commitment/pool.go
```diff
@@ -449,16 +449,19 @@ func (p *Pool) ProcessCommitments(didTimeout bool) (OpenCommitment, error) {
 			return nil, ErrMajorityFailure
 		}
 
+		// If we already have the proposer commitment, we can try to resolve early based on majority
+		// vote. Otherwise we need to wait for the proposer commitment.
+		if proposerCommit == nil {
```
peternose, line 454: In resolution mode, if we don't have the proposer commitment and timeout happened, we return `ErrInsufficientVotes` instead of `ErrNoProposerCommitment`. 
Could be fixed by moving the first block in `if didTimeout {` to the end.

peternose (commented):
Test looks ok, fails without changes and passes with with fixes.

Unrelated observations:
- I think that switching between resolution and detection makes the code very hard to read.
- `failures >= minVotes` could be changed to `required != 0 && failures > required - minVotes` to gain 1 vote when we have even number of members.
- We could optimise the detection by comparing all commits with each other (like we do with votes) instead of waiting for the proposer commit and then compare it to the others.
- We could optimise the resoulution by not waiting for more commits if no group of votes can reach the majority.

Example code would be:
```
func (p *Pool) ProcessCommitments(didTimeout bool) (OpenCommitment, error) {
	if p.Committee == nil {
		return nil, ErrNoCommittee
	}

	switch p.Committee.Kind {
	case scheduler.KindComputeExecutor:
	default:
		panic("roothash/commitment: unknown committee kind: " + p.Committee.Kind.String())
	}

	type vote struct {
		commit OpenCommitment
		tally  int
	}

	var total, commits, failures int

	// Gather votes.
	votes := make(map[hash.Hash]*vote)
	for _, n := range p.Committee.Members {
		switch {
		case !p.Discrepancy && n.Role != scheduler.RoleWorker:
			continue
		case p.Discrepancy && n.Role != scheduler.RoleBackupWorker:
			continue
		}

		total++
		commit, ok := p.ExecuteCommitments[n.PublicKey]
		if !ok {
			continue
		}
		commits++

		if commit.IsIndicatingFailure() {
			failures++
			continue
		}

		k := commit.ToVote()
		if v, ok := votes[k]; !ok {
			votes[k] = &vote{
				commit: commit,
				tally:  1,
			}
		} else {
			v.tally++
		}

		// In case of discrepancy detection, as soon as there is a discrepancy we can proceed
		// with discrepancy resolution. No need to wait for all commits.
		if !p.Discrepancy && len(votes) > 1 {
			p.Discrepancy = true
			return nil, ErrDiscrepancyDetected
		}
	}

	// Determine whether the proposer has submitted a commitment.
	proposer, err := GetTransactionScheduler(p.Committee, p.Round)
	if err != nil {
		return nil, ErrNoCommittee
	}
	proposerCommit, _ := p.ExecuteCommitments[proposer.PublicKey]

	// Handle detection/resolution.
	switch p.Discrepancy {
	case false:
		// Discrepancy detection.
		allowedStragglers := int(p.Runtime.Executor.AllowedStragglers)

		// If it is already known that the number of valid commitments will not exceed the required
		// threshold, there is no need to wait for the timer to expire. Instead, proceed directly to
		// the discrepancy resolution mode, regardless of any additional commits.
		if failures > allowedStragglers {
			p.Discrepancy = true
			return nil, ErrDiscrepancyDetected
		}

		if proposerCommit == nil {
			// If we timed out but the proposer did not submit a commitment, fail the round.
			// TODO: Consider slashing for this offense.
			if didTimeout {
				return nil, ErrNoProposerCommitment
			}
			return nil, ErrStillWaiting
		}

		required := total

		// While a timer is running, all nodes are required to answer.
		//
		// After the timeout has elapsed, a limited number of stragglers are allowed.
		if didTimeout {
			required -= allowedStragglers
			commits -= failures // Since failures count as stragglers.
		}

		if commits < required {
			return nil, ErrStillWaiting
		}
		return proposerCommit, nil
	case true:
		// Discrepancy resolution.
		required := (total / 2) + 1

		if total != 0 && failures > total-required {
			// Too many failures, the round will fail regardless of additional commits.
			logger.Warn("discrepancy resolution majority failed",
				logging.LogEvent, LogEventDiscrepancyMajorityFailure,
			)
			return nil, ErrMajorityFailure
		}

		max := 0
		remaining := total - failures

		for _, v := range votes {
			remaining -= v.tally
			if v.tally > max {
				max = v.tally
			}

			if v.tally >= required {
				// Majority agrees on a commit, result is determined regardless of additional
				// commits.
				//
				// Make sure that the majority commitment is the same as the proposer commitment. We
				// must return the proposer commitment as that one contains additional data.
				if !proposerCommit.MostlyEqual(v.commit) {
					return nil, ErrBadProposerCommitment
				}

				return proposerCommit, nil
			}
		}

		// No majority commitment so far.
		if max+remaining < required {
			return nil, ErrInsufficientVotes
		}
		return nil, ErrStillWaiting
	}

	panic("dead code")
}
```

kostko:
Let's defer any further refactoring to a separate issue/PR to make this more clear.

peternose (approved):

Issue #467, opened 2018-06-21
Issues with batch processing in consensus frontend


kostko:
Several consensus frontend issues related to batch processing:
* After a batch is finalized in the consensus frontend, the leader should immediately check if there is something left in the incoming queue instead of waiting for the batch timeout.
* Incoming calls should be immediately added to the incoming queue instead of going through the command channel.
* After finalization, notifying clients should be done in a separate spawned task.
* Batches need to be added back to the incoming queue after failures.

PR #5272, opened 2023-05-25
go: Remove timecache replace directive in go.mod file

peternose:
The replace directive for github.com/whyrusleeping/timecache has been removed since the go-libp2p-pubsub library version 0.9.3 no longer utilizes it.

@kostko Was problem [#467](https://github.com/libp2p/go-libp2p-pubsub/issues/467) fixed in https://github.com/libp2p/go-libp2p-pubsub/tree/master/timecache?


kostko (approved):

kostko:
Yes, the issue has been fixed a while back with the introduction of a background goroutine that periodically sweeps. I've closed the upstream issue.

PR #5282, opened 2023-06-08
runtime/consensus/roothash: Implement executor commitment structures 

peternose:
Implement executor commitment related structures in Rust, which should be latter used to verify commitments.

For the time being, we will overlook the cumbersome naming issue that results in `ec.header.header` and do the refactoring of `ComputeResultsHeader` and `ExecutorCommitmentHeader` latter.

### DIFF
--- a/runtime/src/consensus/registry.rs
+++ b/runtime/src/consensus/registry.rs
@@ -275,6 +275,23 @@ impl Node {
         }
         false
     }
+
+    /// Searches for an existing supported runtime descriptor
+    /// in runtimes with the specified version and returns it.
+    pub fn get_runtime(&self, runtime_id: &Namespace, version: &Version) -> Option<NodeRuntime> {
+        if let Some(rts) = &self.runtimes {
+            for rt in rts {
+                if runtime_id != &rt.id {
+                    continue;
+                }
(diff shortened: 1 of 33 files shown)


runtime/src/consensus/registry.rs
```diff
@@ -275,6 +275,21 @@ impl Node {
         }
         false
     }
+
+    pub fn get_runtime(&self, runtime_id: &Namespace, version: &Version) -> Option<NodeRuntime> {
```
kostko, line 279: Add documentation comment.

runtime/src/consensus/roothash/commitment/pool.rs
```diff
@@ -0,0 +1,647 @@
...
+impl<F> MessageValidator for F
+where
+    F: Fn(&[Message]) -> Result<()>,
+{
+    fn validate(&self, msgs: &[Message]) -> Result<()> {
+        (*self)(msgs)
+    }
+}
+
+/// A serializable pool of commitments that can be used to perform
```
kostko, line 35: This one does not seem to be serializable :-)

kostko (approved):

PR #5286, opened 2023-06-16
Enable fixgenesis features

jberci:
Update the `fixgenesis` command to be in line with the v62 consensus migration.


.changelog/5286.fix.md
kostko, line 1: The kind probably needs to be changed to `feature` or `internal`.

kostko (approved):

kostko (approved):

PR #5292, opened 2023-06-22
go/consensus: Add block metadata to allow same-block state validation

kostko:
Based on #5285

### DIFF
--- a/go/consensus/api/meta.go
+++ b/go/consensus/api/meta.go
@@ -1,0 +1,34 @@
+package api
+
+import (
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
+	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"
+)
+
+// MethodMeta is the method name for the special block metadata transaction.
+var MethodMeta = transaction.NewMethodName(ModuleName, "Meta", BlockMetadata{})
+
+// BlockMetadataMaxSize is the maximum size of a fully populated and signed block metadata
+// transaction.
+//
+// This should be less than any reasonably configured MaxTxSize.
+const BlockMetadataMaxSize = 16_384
+
+// BlockMetadata contains additional metadata related to the executing block.
+//
+// The metadata is included in the form of a special transaction where this structure is the
+// transaction body.
+type BlockMetadata struct {
+	// StateRoot is the state root after executing all logic in the block.
+	StateRoot hash.Hash `json:"state_root"`
+}
+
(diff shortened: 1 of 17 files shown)


go/consensus/tendermint/abci/system.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/oasisprotocol/oasis-core/go/consensus/tendermint/api"
+	tmcrypto "github.com/oasisprotocol/oasis-core/go/consensus/tendermint/crypto"
+)
+
+// prepareSystemTxs prepares a list of system transactions to be included in a proposed block in
+// case where the node is currently the block proposer.
+func (mux *abciMux) prepareSystemTxs() [][]byte {
+	var systemTxs [][]byte
+
+	// Append block metadata as a system transaction.
```
peternose, line 21: We could for loop over `SystemMethods`, do a `switch`  and put this code under `case consensus.MethodMeta`.

go/consensus/tendermint/abci/system.go
```diff
@@ -0,0 +1,122 @@
...
+	if len(mux.state.proposal.hash) == 0 {
+		return nil
+	}
+
+	var hasBlockMetadata bool
+	for _, tx := range mux.state.blockCtx.SystemTransactions {
+		switch tx.Method {
+		case consensus.MethodMeta:
+			// Block metadata, verify state root.
+			if hasBlockMetadata {
```
peternose, line 88: Having a counter instead of a bool flag could remove this `if`. Not sure what is better :)

peternose (approved):
Looks great! Left some minor comments, though.

go/consensus/api/meta.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
+	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"
+)
+
+// MethodMeta is the method name for the special block metadata transaction.
+var MethodMeta = transaction.NewMethodName(ModuleName, "Meta", BlockMetadata{})
+
+// BlockMetadataMaxSize is the maximum size of a fully populated block metadata transaction.
+const BlockMetadataMaxSize = 10_000
```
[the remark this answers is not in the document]
kostko: Not really, this is needed to ensure that the block would not overflow after adding the metadata transaction so it is fine if there is some buffer. I've now also added a sanity check during system transaction creation.

PR #5299, opened 2023-06-29
go/consensus/cometbft: Cleanup, data directory rename and migrate seed address book

kostko:


abukosek (approved):

abukosek:
We should also fix the upgrade test to rename the directories before launching the new node or add this migration to the node itself like you've already done for the seed node (this would also be easier for the users, since they could just copy their old data directory without having to rename anything inside).

go/consensus/cometbft/seed/seed.go
```diff
@@ -125,6 +126,21 @@ func (s *Service) GetAddresses() ([]node.ConsensusAddress, error) {
 	return []node.ConsensusAddress{addr}, nil
 }
 
+func initSeedDataDir(dataDir string) (string, error) {
+	seedDataDir := filepath.Join(dataDir, "seed")
+
+	// Check if there is a legacy data directory and migrate it so we can get the old address book.
+	legacyDataDir := filepath.Join(dataDir, "tendermint-seed")
```
peternose, line 133: After the upgrade, this code will become obsolete. To ensure its removal is not forgotten, I prefer preparing a draft pull request in advance for these cases.

peternose (approved):

PR #5300, opened 2023-07-03
runtime/consensus/tendermint/verifier: Support same-block validation

peternose:
WIP: Error: root: failed to run scenario: root: failed to run scenario: log watcher client-0/log: log assertion failed: round failure detected


go/worker/client/committee/node.go
```diff
@@ -140,60 +139,28 @@ func (n *Node) Query(ctx context.Context, round uint64, method string, args []by
...
 	dsc := n.commonNode.CurrentDescriptor
-	latestRound := n.commonNode.CurrentBlock.Header.Round
 	n.commonNode.CrossNode.Unlock()
 
 	if dsc == nil {
 		return nil, api.ErrNoHostedRuntime
 	}
 	maxMessages := dsc.Executor.MaxMessages
 
-	var resolvedRound uint64
```
kostko, line 151: Should this be feature-based? Otherwise things will degrade immediately after consensus/node upgrade until the runtime is also upgraded.

peternose: Yes, as the sent block would be one height lower.

go/worker/client/committee/node.go
```diff
@@ -148,52 +150,80 @@ func (n *Node) Query(ctx context.Context, round uint64, method string, args []by
...
-		// Get consensus light block for state after executing block at given height.
-		lb, err := n.commonNode.Consensus.GetLightBlockForState(ctx, annBlk.Height)
-		if err != nil {
-			return nil, fmt.Errorf("client: failed to get light block at height %d: %w", annBlk.Height, err)
-		}
-		epoch, err := n.commonNode.Consensus.Beacon().GetEpoch(ctx, annBlk.Height)
-		if err != nil {
-			return nil, fmt.Errorf("client: failed to get epoch at height %d: %w", annBlk.Height, err)
-		}
+	// TODO: Remove once same block consensus validation is deployed.
```
kostko, line 158: Created a task for this so we don't forget.

kostko (approved):

PR #5301, opened 2023-07-03
go/control: Add runtime provisioner type to host status output

abukosek:
This PR adds the runtime provisioner type to the status output.

### DIFF
--- a/go/oasis-node/cmd/node/node_control.go
+++ b/go/oasis-node/cmd/node/node_control.go
@@ -331,6 +331,19 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
 			}
 		}
 
+		// Fetch provisioner type.
+		_, provisioner, err := rt.Host(ctx)
+		switch {
+		case err != nil:
+			n.logger.Error("failed to fetch host configuration",
+				"err", err,
+			)
+		case provisioner != nil:
+			status.Provisioner = provisioner.Name()
+		default:
+			status.Provisioner = "none"
+		}
+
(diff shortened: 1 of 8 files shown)


go/oasis-node/cmd/node/node_control.go
```diff
@@ -331,6 +331,20 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
...
+		_, provisioner, err := rt.Host(ctx)
+		if err != nil {
+			n.logger.Error("failed to fetch host configuration",
+				"err", err,
+			)
+		} else {
+			if provisioner != nil {
+				status.Provisioner = provisioner.Name()
+			} else {
+				status.Provisioner = "none"
```
peternose, line 344:
```
		switch {
		case err != nil:
			n.logger.Error("failed to fetch host configuration",
				"err", err,
			)
		case provisioner != nil:
			status.Provisioner = provisioner.Name()
		default:
			status.Provisioner = "none"
		}
```
Could also use an empty string, as the comment says this is the name of the provisioner.

peternose (approved):
```
{
  "runtimes": {
    "8000000000000000000000000000000000000000000000000000000000000000": {
      "provisioner": "sandbox"
    }
  },
}
```
:+1:

go/oasis-node/cmd/node/node_control.go
abukosek: I think "none" is a bit more user-friendly in this case :)

peternose (approved):

PR #5318, opened 2023-07-11
go: Remove TLS certificate rotation

jberci:


go/common/identity/identity.go
```diff
@@ -82,88 +66,10 @@ type Identity struct {
 	// connecting to the sentry node's control connection.  It is never rotated.
 	TLSSentryClientCertificate *tls.Certificate
 
-	// DoNotRotateTLS flag is true if we mustn't rotate the TLS certificates below.
-	DoNotRotateTLS bool
-
 	// tlsSigner is a node TLS certificate signer.
 	tlsSigner signature.Signer
```
kostko, line 70: I think this (and certificate?) no longer needs an accessor function given it is no longer modified and so can be exposed directly similar to the other signers? This also probably means that the lock itself and mutation methods are no longer needed?

go/common/identity/identity.go
```diff
@@ -191,22 +97,6 @@ func (i *Identity) SetTLSCertificate(cert *tls.Certificate) {
...
-// GetNextTLSCertificate returns the next TLS certificate.
-func (i *Identity) GetNextTLSCertificate() *tls.Certificate {
-	i.RLock()
-	defer i.RUnlock()
-
-	return i.nextTLSCertificate
-}
-
 // GetTLSPubKeys returns a list of currently valid TLS public keys.
 func (i *Identity) GetTLSPubKeys() []signature.PublicKey {
```
kostko, line 101: Is this used anywhere now?

runtime/src/consensus/registry.rs
```diff
@@ -64,7 +64,9 @@ pub struct TLSInfo {
     pub pub_key: signature::PublicKey,
 
     /// Public key that will be used for establishing connections after certificate rotation (if enabled).
-    pub next_pub_key: signature::PublicKey,
+    ///
+    /// Unused, certificates are not rotated anymore.
```
kostko, line 68: Drop the documentation comment as below given that the field is deprecated.

runtime/src/consensus/registry.rs
```diff
@@ -64,7 +64,9 @@ pub struct TLSInfo {
...
+    pub _deprecated_next_pub_key: Option<signature::PublicKey>,
```
kostko, line 69: Add the `#[cbor(rename = "next_pub_key", optional)]` annotation.

go/common/identity/identity.go
```diff
@@ -366,21 +173,12 @@ func doLoadOrGenerate(dataDir string, signerFactory signature.SignerFactory, sho
...
+		TLSSigner:                  memory.NewFromRuntime(cert.PrivateKey.(ed25519.PrivateKey)),
+		TLSCertificate:             cert,
 		TLSSentryClientCertificate: sentryClientCert,
-		tlsRotationNotifier:        pubsub.NewBroker(false),
-		dataDir:                    dataDir,
 	}, nil
 }
 
-func ephemeralKeyPath(dataDir, generation string) string {
-	return filepath.Join(dataDir, fmt.Sprintf("%s%s.pem", tlsEphemeralKeyBaseFilename, generation))
```
peternose, line 381: Do we need to remove these files, to not pollute the datadir?

go/oasis-node/cmd/node/node_control.go
```diff
@@ -175,7 +176,7 @@ func (n *Node) getIdentityStatus() control.IdentityStatus {
 	return control.IdentityStatus{
 		Node:      n.Identity.NodeSigner.Public(),
 		Consensus: n.Identity.ConsensusSigner.Public(),
-		TLS:       n.Identity.GetTLSPubKeys(),
+		TLS:       []signature.PublicKey{n.Identity.TLSSigner.Public()},
```
peternose, line 179: Can we remove the array?

runtime/src/consensus/state/registry.rs
```diff
@@ -120,7 +120,7 @@ mod test {
 
     #[test]
     fn test_registry_state_interop() {
-        // Keep in sync with go/consensus/tendermint/apps/registry/state/interop/interop.go.
+        // Keep in sync with go/consensus/cometbft/apps/registry/state/interop/interop.go.
```
peternose, line 123: Can also change in other files.

runtime/src/consensus/registry.rs
peternose: Can we remove these two fields, as runtimes will get upgraded after nodes?

kostko: In theory runtimes could observe v2 descriptors in state and should be able to decode them?

go/oasis-node/cmd/node/node_control.go
kostko: The only downside is that it breaks compatibility. But otherwise making it consistent with the others makes sense.

go/oasis-test-runner/scenario/e2e/runtime/sentry.go
```diff
@@ -189,9 +176,7 @@ func (s *sentryImpl) Run(ctx context.Context, childEnv *env.Env) error { // noli
 		return fmt.Errorf("failed to get status for compute1: %w", err)
 	}
 	compute1ServerPublicKeys := make(map[signature.PublicKey]bool)
```
peternose, line 178: This variable is never used :shrug:

kostko: Even though this is resolved, it still does not seem to be used anywhere? In that case we should just remove it.

go/common/identity/identity.go
kostko: I would suggest removing them as part of `doLoadOrGenerate` in case they exist.

peternose: Agree. I would also add a TODO comment and prepare a draft pull request to remove this cleanup code immediately after the next release.

go/oasis-test-runner/scenario/e2e/runtime/sentry.go
```diff
@@ -168,16 +165,6 @@ func (s *sentryImpl) Run(ctx context.Context, childEnv *env.Env) error { // noli
...
-	if err != nil {
-		return fmt.Errorf("sentry: error loading compute node identity: %w", err)
-	}
-	// Make sure compute0 has disabled certificate rotation.
-	if !compute0Identity.DoNotRotateTLS {
-		return fmt.Errorf("sentry: compute-0 does not have disabled certificate rotation")
-	}
-
 	compute1 := s.Net.ComputeWorkers()[1]
 	// Query for compute 1 TLS public keys.
```
peternose, line 169: I have no idea what was tested here. I see 2 options:
- Remove this code.
- Test if fetching works.
```
	// Query for compute 1 status.
	compute1Ctrl, err := oasis.NewController(compute1.SocketPath())
	if err != nil {
		return err
	}
	if _, err := compute1Ctrl.GetStatus(ctx); err != nil {
		return fmt.Errorf("failed to get status for compute1: %w", err)
	}
```

jberci: I removed the code, since the call is also made in other tests.

kostko (approved):

peternose (approved):

kostko (approved):

PR #5319, opened 2023-07-11
staking: Fix reward distribution when common pool is exhausted

abukosek:
This PR fixes the scenario when the reward schedule hasn't completed yet, but the common pool has been exhausted.

### DIFF
--- a/go/consensus/cometbft/apps/staking/state/state.go
+++ b/go/consensus/cometbft/apps/staking/state/state.go
@@ -1164,11 +1164,8 @@ func (s *MutableState) DiscardGovernanceDeposit(
 }
 
 // AddRewards computes and transfers a staking reward to active escrow accounts.
-// If an error occurs, the pool and affected accounts are left in an invalid state.
-// This may fail due to the common pool running out of stake. In this case, the
-// returned error's cause will be `staking.ErrInsufficientBalance`, and it should
-// be safe for the caller to roll back to an earlier state tree and continue from
-// there.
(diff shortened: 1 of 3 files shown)


go/consensus/cometbft/apps/staking/state/state.go
```diff
@@ -1219,6 +1219,11 @@ func (s *MutableState) AddRewards(
 			continue
 		}
 
+		if q.Cmp(commonPool) == 1 {
```
peternose, line 1222: If this condition is true (`q > commonPool`), should we lower the reward and give as much as we can (`q = commmonPool.Clone()`)?

> // If an error occurs, the pool and affected accounts are left in an invalid state.
> // This may fail due to the common pool running out of stake.

The method's comment is not correct anymore, as the pool will never run out of stake.

This method is not fair. For example:
- If I have unlimited escrow balance, my rewards will always be too big and therefore skipped while others will receive their awards until they drain the pool.
- Assuming equal balances and too small pool to cover all rewards, the first in the list will get more rewards than the last one.

However, I'm not sure if this is a problem. If we want fairness, we should distribute the reminder of the pool according to stakes. But that is probably too much.

abukosek: Yeah, we should discuss what to do about this and implement it in a separate PR.  This PR just fixes the panic that would have happened otherwise :)

peternose (approved):

PR #5326, opened 2023-07-18
go/workerpool: support backoff on failures

ptrus:
Adds support for backoff on failures to the workerpool module. 

Use the backoff in the storage sync protocol to fix the case where currently if storage sync requests start failing (e.g. due to a network error), the storage worker will crazily retry and fill up logs / cpu.

TODO: 
- [x] minor todo's in code
- [x] add some unit tests to the workerpool module
- [x] test that this fixes the mentioned storage sync issue

Before fix (with an artificial instant failure):
```
cat node.log | grep "error calling getdiff" | grep "11:22:03" | wc -l
18202
```

after:
```
cat node.log | grep "error calling getdiff" | grep "11:50:5" | wc -l
23
```

From ~18k calls per second, to ~2 per second.

### DIFF
--- a/go/common/workerpool/workerpool.go
+++ b/go/common/workerpool/workerpool.go
@@ -4,15 +4,55 @@ package workerpool
 import (
 	"fmt"
 	"sync"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/eapache/channels"
 
+	cmnBackoff "github.com/oasisprotocol/oasis-core/go/common/backoff"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
 )
 
+type expBackoff struct {
+	lock sync.Mutex
+
+	backoff *backoff.ExponentialBackOff
+	timeout time.Duration
+}
+
+func (b *expBackoff) Success() {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
+	b.backoff.Reset()
+	b.timeout = 0
+}
+
+func (b *expBackoff) Failure() {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
+	b.timeout = b.backoff.NextBackOff()
+}
+
+func (b *expBackoff) Timeout() time.Duration {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
+	return b.timeout
+}
+
+// BackoffConfig is the configuration for the backoff mechanism for the workers.
+type BackoffConfig struct {
+	// MinTimeout is the minimum timeout to wait in case of failures.
+	MinTimeout time.Duration
+	// MaxTimeout is the maximum timeout to wait in case of repeated failures.
+	MaxTimeout time.Duration
+}
+
 type jobDescriptor struct {
 	terminate  bool
-	job        func()
+	job        func() error
 	completeCh chan struct{}
 }
 
@@ -24,6 +64,7 @@ type jobDescriptor struct {
 type Pool struct { // nolint: maligned
 	lock        sync.Mutex
 	workerGroup sync.WaitGroup
+	backoff     *expBackoff
 
 	name string
 
@@ -92,7 +133,7 @@ func (p *Pool) Quit() <-chan struct{} {
 
 // Submit adds a task to the pool's queue and returns a channel that will be closed
 // once the task is complete.
-func (p *Pool) Submit(job func()) <-chan struct{} {
+func (p *Pool) Submit(job func() error) <-chan struct{} {
 	p.lock.Lock()
 	defer p.lock.Unlock()
 
@@ -119,6 +160,15 @@ func (p *Pool) worker() {
 	defer p.workerGroup.Done()
 
 	for {
+		// Wait for the backoff period if configured.
+		if p.backoff != nil {
+			select {
+			case <-p.stopCh:
+				return
+			case <-time.After(p.backoff.Timeout()):
+			}
+		}
+
 		select {
 		case <-p.stopCh:
 			return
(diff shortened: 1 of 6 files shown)


go/worker/storage/committee/node.go
```diff
@@ -424,12 +418,12 @@ func (n *Node) fetchDiff(round uint64, prevRoot, thisRoot storageApi.Root) {
 			rsp, pf, err := n.storageSync.GetDiff(ctx, &storageSync.GetDiffRequest{StartRoot: prevRoot, EndRoot: thisRoot})
 			if err != nil {
 				result.err = err
-				return
```
kostko, line 427: Should maybe just get rid of the if clause now?

go/common/workerpool/workerpool.go
```diff
@@ -4,15 +4,80 @@ package workerpool
...
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/eapache/channels"
 
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
 )
 
+type backoff interface {
```
kostko, line 14: Could we use [`github.com/cenkalti/backoff/v4`](https://pkg.go.dev/github.com/cenkalti/backoff/v4) (which also adds jitter) instead of rolling another backoff implementation?

ptrus: Yeah i think i can use it nicely within the 'expBackoff' struct to avoid manually computing the backoff. Will update.

But i do need to keep the interface/wrapper struct, because i want to share the same backoff between all workers (and cannot do that with plain upstream backoff as it doesn't expose a method to get the current backoff without also incrementing the failure count).

go/common/workerpool/workerpool.go
```diff
@@ -130,23 +203,37 @@ func (p *Pool) worker() {
...
+			} else {
+				p.backoff.Failure()
+			}
 		}
 	}
 }
 
 // New creates and returns a new worker pool with one worker goroutine.
-func New(name string) *Pool {
+func New(name string, backoffCfg *BackoffConfig) *Pool {
```
peternose, line 220: As backoff is optional, I would remove it from the constructor, add a method `WithBackoff`, remove `noBackoff` implementation and rather check if `backoff != nil`.

I guess it is up to debate if the workerpool should be responsible for backoffs or the jobs. Currently, mixing jobs A and B, where A always fails and needs a backoff and B always succeeds, would force B to wait, and the latter would always reset the timer.

ptrus: I thought about it, but the `WithBackoff` version has a downside that the field now needs to be protected by a lock, because it won't be ensured that it is configured before jobs are submitted. The nicest approach would likely be something akin to an option builder pattern, but seems overkill for this single option.
I think passing an optional option/config struct is quite a common pattern in go.

>remove noBackoff implementation and rather check if backoff != nil. 

Any reasons why preferring the `backoff != nil` variant? I had both implemented, seemed nicer to avoid the `backoff != nil` checks. The only benefit I see is that it avoids the need for the internal `backoff` interface. Don't have a strong opinion about either though.

> I guess it is up to debate if the workerpool should be responsible for backoffs or the jobs. Currently, mixing jobs A and B, where A always fails and needs a backoff and B always succeeds, would force B to wait, and the latter would always reset the timer

For cases where jobs are not mixed, having this in the workerpool avoids duplicating it, so it kinda makes sense i think? 

> As currently the pool is responsible, the lock is actually not needed.

Which lock? The one in `expBackoff`? It's needed.

peternose: > I think passing an optional option/config struct is quite a common pattern in go.

That's true. Another option would also be to create 2 constructors. And we should probably not add the config to the constructor, but an interface instead. The backoff constructor should take the config.

> Any reasons why preferring the backoff != nil variant? I

To avoid creating a new timer with zero duration and giving garbage collector less work.

> For cases where jobs are not mixed, having this in the workerpool avoids duplicating it, so it kinda makes sense i think?

Yeah, its probably fine. Was just saying because the pool accepts arbitrary jobs, so they could be mixed.

> Which lock? The one in expBackoff is needed.

I would say not, as `Success`, `Failure` and `Timeout` are called only in `worker` method.

go/common/workerpool/workerpool.go
ptrus: Updated

go/common/workerpool/workerpool.go
ptrus: > To avoid creating a new timer with zero duration and giving garbage collector less work.

:+1: switched already in any case.

> are called only in worker method.

There can be multiple workers https://github.com/oasisprotocol/oasis-core/blob/457ee23f9a772e78a900d2fd1875132ceb1d978f/go/common/workerpool/workerpool.go#L62-L64

go/common/workerpool/workerpool.go
```diff
@@ -147,6 +213,16 @@ func New(name string) *Pool {
 		logger:       logging.GetLogger(fmt.Sprintf("workerpool/%s", name)),
 	}
 
+	if cfg != nil && cfg.Backoff != nil {
+		backoff := cmnBackoff.NewExponentialBackOff()
```
peternose, line 217: I would move this to `newExpBackoff(cfg *BackoffConfig)` or `newExpBackoff(minTimeout time.Duration, maxTimeout time.Duration)` and as said before, maybe change the method to accept the backoff instead of creating one.

go/common/workerpool/workerpool.go
```diff
@@ -119,6 +160,15 @@ func (p *Pool) worker() {
 	defer p.workerGroup.Done()
 
 	for {
+		// Wait for the backoff period if configured.
+		if p.backoff != nil {
+			select {
+			case <-p.stopCh:
+				return
+			case <-time.After(p.backoff.Timeout()):
```
peternose, line 168: Not sure if this works as intended. For example, if we have unlimited number of workers, it could happen that even if all jobs fail, we never wait before execution. But that is probably not relevant in the long run.

peternose (approved):

go/common/workerpool/workerpool.go
ptrus: Yeah, if a worker already waited for the `backoff.Timeout()` duration before it accepted a job, it won't delay further.

On a related note, I actually just noticed that the storage-sync worker pool is actually shared between all runtimes, so technically it is "mixing jobs A and B" and could potentially end up in a state that for one runtime all jobs are failing, while not for the other one. So I will probably move the backoff into the storage worker as you suggested before, and remove the backoff from the workerpool.

ptrus:
Fixed in: https://github.com/oasisprotocol/oasis-core/pull/5341

PR #5351, opened 2023-08-17
runtime: Bump ed25519-dalek, sha2, hmac, rsa

kostko:
* runtime: Bump ed25519-dalek to 2.0.0
* runtime: Bump sha2 to 0.10.7, hmac to 0.12.1
* runtime: Bump rsa to 0.9.2


runtime/src/common/bytes.rs
```diff
@@ -11,7 +11,7 @@
 macro_rules! impl_bytes {
     ($name:ident, $size:expr, $doc:expr) => {
         #[doc=$doc]
-        #[derive(Clone, Copy)]
+        #[derive(Clone, Copy, zeroize::Zeroize)]
```
peternose, line 14: Do we want to zeroize all, even public keys and hashes?

runtime/src/common/crypto/x25519.rs
```diff
@@ -38,6 +38,9 @@ impl PrivateKey {
 impl From<[u8; PRIVATE_KEY_LENGTH]> for PrivateKey {
     /// Load private key from a byte array.
     fn from(bytes: [u8; PRIVATE_KEY_LENGTH]) -> PrivateKey {
+        // We must clamp to match previous x25519-dalek behavior and the test vectors.
+        let bytes = curve25519_dalek::scalar::clamp_integer(bytes);
```
peternose, line 42: Could also change test vectors, as secrets/scalars will always get clamped (when reading in the old version, when doing an operation in the new one).

runtime/src/common/bytes.rs
kostko: No, this only implements `Zeroize` which provides the `zeroize` method. It does not automatically zeroize (e.g. on drop).

runtime/src/common/crypto/x25519.rs
kostko: I guess that's true. I was concerned about other use cases, e.g. for the RNG. But that is ephemeral so it may be fine regardless?

peternose (approved):

PR #5356, opened 2023-08-24
go: Bump go to 1.21.0

peternose:
In one of the commits I decided to remove some unused parameters (e.g. `ctx`, `dataDir`, `identity`). If needed, they can be added latter. However, I did not remove unused parameters in `MaybeStartSubtree` and `PutNode`, as it is strange that the functions are not using them.

TODOs (probably in another PR):
- [ ] Replace deprecated `elliptic.Unmarshal` function.
- [ ] Fix linting of `tests/upgrade`.
- [ ] Build `tests/upgrade/post` with go 1.21.


.buildkite/go/build.sh
```diff
@@ -29,6 +29,10 @@ pushd go
...
 popd
 
-pushd tests/upgrade
-  make all
+pushd tests/upgrade/pre
+  OASIS_GO=go1.19.10 make all
 popd
+
+pushd tests/upgrade/post
+  OASIS_GO=go1.20.2 make all
```
kostko, line 37: Will this change to 1.21 with this bump or what is the plan?

peternose: I will bump to 1.21 once this PR is merged to master.

kostko (approved):

PR #5357, opened 2023-08-29
Add observer node role

kostko:


abukosek (approved):

go/registry/api/api.go
```diff
@@ -1180,6 +1180,30 @@ func VerifyRuntime( // nolint: gocyclo
 			}
 		}
 	}
+	// Ensure valid per-role policy if present.
+	if perRole := rt.AdmissionPolicy.PerRole; perRole != nil {
```
peternose, line 1184: Feels like these validation checks could be moved to `AdmissionPolicy`, `EntityWhitelist`, `PerRoleAdmissionPolicy`, ...

go/consensus/cometbft/apps/registry/admission.go
```diff
@@ -0,0 +1,237 @@
...
+	newNode *node.Node,
+	rt *registry.Runtime,
+	epoch beacon.EpochTime,
+) error {
+	if len(rt.AdmissionPolicy.PerRole) == 0 {
+		return nil // No per-role policy specified.
+	}
+
+	// Iterate over all valid roles (each entry in the map can only have a single role).
+	for _, role := range node.Roles() {
```
peternose, line 187: I would probably iterate through `rt.AdmissionPolicy.PerRole`, to get rid of the previous `if`.

go/registry/api/runtime.go
```diff
@@ -225,6 +225,24 @@ type EntityWhitelistConfig struct {
 type RuntimeAdmissionPolicy struct {
 	AnyNode         *AnyNodeRuntimeAdmissionPolicy         `json:"any_node,omitempty"`
 	EntityWhitelist *EntityWhitelistRuntimeAdmissionPolicy `json:"entity_whitelist,omitempty"`
+
+	// PerRole is a per-role admission policy that must be satisfied in addition to the global
+	// admission policy for a specific role.
+	PerRole map[node.RolesMask]PerRoleAdmissionPolicy `json:"per_role,omitempty"`
```
peternose, line 231: Not sure what we gain with `role -> entity -> maxNodes` policy as we already have `entity -> role -> maxNodes` policy.

go/consensus/cometbft/apps/registry/admission.go
```diff
@@ -0,0 +1,237 @@
...
+import (
+	beacon "github.com/oasisprotocol/oasis-core/go/beacon/api"
+	"github.com/oasisprotocol/oasis-core/go/common/node"
+	"github.com/oasisprotocol/oasis-core/go/consensus/cometbft/api"
+	registryState "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/apps/registry/state"
+	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
+)
+
+// policyFn is an admission policy verification function.
+type policyFn func(
```
peternose, line 12: Do we need new types here? Maybe it would be better to move structs here and add 2 new methods, `Validate` and `Verify`.

go/worker/registration/worker.go
```diff
@@ -1094,6 +1094,18 @@ func New(
 		rp.SetAvailable(func(*node.Node) error { return nil })
 	}
 
+	// When a node is a client node and it has an entity configured, we register it with the
+	// observer role as this may be needed for confidential runtimes.
```
peternose, line 1098: Can you provide more details about the observer role?

Currently, the only change I see is in `ComputeRuntimeAllowedRoles`, which permits observers to register compute runtimes. If observers are passive compute nodes, they should likely also participate in P2P topics.

Should the observer role and the compute role be mutually exclusive?

kostko: > If observers are passive compute nodes, they should likely also participate in P2P topics.

They are basically "client nodes that are registered on chain so that their eligibility can be enforced by an on-chain policy". They already participate in some P2P topics (e.g. the transaction propagation ones).

> Should the observer role and the compute role be mutually exclusive?

Yes, at least it doesn't make sense for a node to register for both (which should already be the case in the code above). Or do you think that a node should not even be allowed to register for both? I guess it doesn't make much sense to do so.

go/registry/api/api.go
```diff
@@ -1180,6 +1180,30 @@ func VerifyRuntime( // nolint: gocyclo
...
+		for role, cfg := range perRole {
+			if !role.IsSingleRole() {
```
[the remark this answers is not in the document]
kostko: What extra validation would you do?

go/registry/api/runtime.go
kostko: Imagine where you only want to enforce a whitelist for a specific role, but not any others.

go/consensus/cometbft/apps/registry/admission.go
kostko: You probably mean move these functions into `registry/api`. Yeah that actually makes more sense.

go/registry/api/api.go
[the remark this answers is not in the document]
peternose: None, I missed that `RoleReserved` includes all unused roles, not only (2) and 3.

go/worker/registration/worker.go
peternose: I was thinking the node should not be allowed to register for both. But I guess this should not be a problem.

peternose (approved):

PR #5359, opened 2023-09-05
roothash: Store past runtime state and I/O roots in consensus state

abukosek:
This adds the ability to store past N runtime state and I/O roots in consensus state.

### DIFF
--- a/go/consensus/cometbft/apps/roothash/messages.go
+++ b/go/consensus/cometbft/apps/roothash/messages.go
@@ -131,6 +131,12 @@ func (app *rootHashApplication) changeParameters(ctx *api.Context, msg interface
 	if err != nil {
 		return nil, fmt.Errorf("roothash: failed to load consensus parameters: %w", err)
 	}
+	var needToDeletePastRoots bool
+	if changes.MaxPastRootsStored != nil && *changes.MaxPastRootsStored < params.MaxPastRootsStored {
+		// If we've reduced the number of past roots stored, we need to delete
+		// the excess when applying the new parameters.
+		needToDeletePastRoots = true
+	}
 	if err = changes.SanityCheck(); err != nil {
 		return nil, fmt.Errorf("roothash: failed to validate consensus parameter changes: %w", err)
 	}
@@ -143,6 +149,13 @@ func (app *rootHashApplication) changeParameters(ctx *api.Context, msg interface
 
 	// Apply changes.
 	if apply {
+		if needToDeletePastRoots {
+			err = state.ShrinkPastRoots(ctx, params.MaxPastRootsStored)
+			if err != nil {
+				return nil, fmt.Errorf("roothash: failed to shrink past stored roots: %w", err)
+			}
+		}
+
 		if err = state.SetConsensusParameters(ctx, params); err != nil {
 			return nil, fmt.Errorf("roothash: failed to update consensus parameters: %w", err)
 		}
--- a/go/consensus/cometbft/apps/roothash/messages_test.go
+++ b/go/consensus/cometbft/apps/roothash/messages_test.go
@@ -1,6 +1,7 @@
 package roothash
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -9,7 +10,9 @@ import (
 	abciAPI "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/api"
 	roothashState "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/apps/roothash/state"
 	governance "github.com/oasisprotocol/oasis-core/go/governance/api"
+	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
 	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
+	"github.com/oasisprotocol/oasis-core/go/roothash/api/block"
 )
 
 func TestChangeParameters(t *testing.T) {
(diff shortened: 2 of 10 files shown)


go/consensus/cometbft/apps/roothash/state/state.go
```diff
@@ -303,6 +353,56 @@ func (s *MutableState) SetRuntimeState(ctx context.Context, state *roothash.Runt
...
 	}
+
+	// Clean previously stored state and I/O roots if we're over the maximum.
+	params, err := s.ConsensusParameters(ctx)
+	var maxStored uint32
+	if err == nil {
+		maxStored = params.MaxPastRootsStored
+	}
+
+	roots, err := s.PastRoots(ctx, state.Runtime.ID)
```
kostko, line 364: It doesn't seem optimal that you load all of the past roots just to (in the most common case) add one root and remove one root. You could probably just blindly remove the root at round `state.CurrentBlock.Header.Round - MaxPastRoots`.

go/roothash/api/api.go
```diff
@@ -571,6 +571,10 @@ type ConsensusParameters struct {
 
 	// MaxEvidenceAge is the maximum age of submitted evidence in the number of rounds.
 	MaxEvidenceAge uint64 `json:"max_evidence_age"`
+
+	// MaxPastRootsStored is the maximum number of past runtime state and I/O
+	// roots that are stored in the consensus state.
+	MaxPastRootsStored uint32 `json:"max_past_roots_stored,omitempty"`
```
kostko, line 577: Also update fixgenesis (and the test runner's default genesis generation) to configure a sensible value for this (e.g. it should default to something like 1200 rounds).

go/consensus/cometbft/apps/roothash/state/state.go
```diff
@@ -51,6 +51,11 @@ var (
 	//
 	// Value is CBOR-serialized message.IncomingMessage.
 	inMsgQueueKeyFmt = keyformat.New(0x29, keyformat.H(&common.Namespace{}), uint64(0))
+	// Previous state and I/O runtime roots.
+	// Key format is: 0x2a H(<runtime-id>) <round>
+	// Value is CBOR-serialized array of past N state and I/O roots.
```
kostko, line 56: This is not true, the value is just the array containing state and I/O roots for the given round, in order. I would rather define something like the following (in the roothash API) and serialize that:
```go
type PastRoots struct {
	_ struct{} `cbor:",toarray"` //nolint

	StateRoot hash.Hash
	IORoot    hash.Hash
}
```
This gets serialized as `[]hash.Hash{StateRoot, IORoot}`.

abukosek: Yeah, I know, it's an old comment, I was just in the process of updating it :)

go/consensus/cometbft/apps/roothash/state/state.go
abukosek: Done, please take a look :)

go/consensus/cometbft/apps/roothash/state/state.go
```diff
@@ -270,6 +278,69 @@ func (s *ImmutableState) IncomingMessageQueue(ctx context.Context, runtimeID com
...
+
+	// Round -> [state, I/O] roots.
+	ret := make(map[uint64]roothash.PastRoots)
+	for it.Seek(pastRootsFmt.Encode(&runtimeID)); it.Valid(); it.Next() {
+		var rtID keyformat.PreHashed
+		var round uint64
+		if !pastRootsFmt.Decode(it.Key(), &rtID, &round) {
+			break
+		}
+
```
kostko, line 313: I think this should also break in case of `runtimeID` mismatch (also below in `PastRootsCount` and other methods)? See other similar iterators.

go/consensus/cometbft/apps/roothash/state/state.go
```diff
@@ -270,6 +278,69 @@ func (s *ImmutableState) IncomingMessageQueue(ctx context.Context, runtimeID com
...
+		ret[round] = roots
+	}
+
+	return ret, nil
+}
+
+// PastRootsCount returns the number of past state and I/O roots in storage
+// for the given runtime ID.
```
kostko, line 326: Maybe mention that this is more efficient than calling `len(s.PastRoots(id))` as it avoids deserialization.

go/consensus/cometbft/apps/roothash/state/state.go
```diff
@@ -303,6 +374,82 @@ func (s *MutableState) SetRuntimeState(ctx context.Context, state *roothash.Runt
 	if err := s.ms.Insert(ctx, ioRootKeyFmt.Encode(&state.Runtime.ID), ioRoot); err != nil {
 		return api.UnavailableStateError(err)
 	}
+
+	// Clean previously stored state and I/O roots if we're over the maximum.
+	params, err := s.ConsensusParameters(ctx)
+	var maxStored uint32
+	if err == nil {
+		maxStored = params.MaxPastRootsStored
+	}
```
kostko, line 383: The `ConsensusParameters` method failing is a critical error and should not happen in practice.
```suggestion
	params, err := s.ConsensusParameters(ctx)
	if err != nil {
		return api.UnavailableStateError(err)
	}
	maxStored := params.MaxPastRootsStored
```

kostko (commented):
Also add some tests with more runtimes at the same time.

go/consensus/cometbft/apps/roothash/messages.go
```diff
@@ -131,6 +131,12 @@ func (app *rootHashApplication) changeParameters(ctx *api.Context, msg interface
 	if err != nil {
 		return nil, fmt.Errorf("roothash: failed to load consensus parameters: %w", err)
 	}
+	var needToDeletePastRoots bool
```
peternose, line 134:
```suggestion
	oldMaxPastRootsStored := params.MaxPastRootsStored 
```
I think that this way you can get rid of 1 `if` and 1 comparison.

go/consensus/cometbft/apps/roothash/state/state.go
```diff
@@ -51,6 +51,14 @@ var (
 	//
 	// Value is CBOR-serialized message.IncomingMessage.
 	inMsgQueueKeyFmt = keyformat.New(0x29, keyformat.H(&common.Namespace{}), uint64(0))
+	// Previous state and I/O runtime roots.
+	// Key format is: 0x2a H(<runtime-id>) <round>
+	// Value is CBOR-serialized array of state and I/O roots for that round.
```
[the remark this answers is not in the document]
kostko: The serialization format of `roothash.RoundRoots` is an array to make encoding more compact. But yeah that is an internal detail.

go/roothash/api/api.go
```diff
@@ -640,3 +651,12 @@ func VerifyRuntimeParameters(rt *registry.Runtime, params *ConsensusParameters)
 	}
 	return nil
 }
+
+// PastRoots holds the per-round state and I/O roots that are stored in
+// consensus state.
+type PastRoots struct {
+	_ struct{} `cbor:",toarray"` //nolint
```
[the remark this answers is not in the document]
kostko: More compact encoding given that there are many roots stored. The overhead is probably not that large though.

go/consensus/cometbft/apps/roothash/state/state.go
```diff
@@ -303,6 +376,82 @@ func (s *MutableState) SetRuntimeState(ctx context.Context, state *roothash.Runt
...
+
+	// Clean previously stored state and I/O roots if we're over the maximum.
+	params, err := s.ConsensusParameters(ctx)
+	if err != nil {
+		return api.UnavailableStateError(err)
+	}
+	maxStored := params.MaxPastRootsStored
+
+	// Add state and I/O roots for this round if enabled.
+	if maxStored > 0 {
```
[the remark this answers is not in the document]
peternose: I agree, but it could be done in steps. I'm just not sure if it's worth it.

go/consensus/cometbft/apps/roothash/state/state.go
```diff
@@ -51,6 +51,14 @@ var (
 	//
 	// Value is CBOR-serialized message.IncomingMessage.
 	inMsgQueueKeyFmt = keyformat.New(0x29, keyformat.H(&common.Namespace{}), uint64(0))
+	// Previous state and I/O runtime roots.
```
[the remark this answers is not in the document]
abukosek: ?

go/consensus/cometbft/apps/roothash/state/state.go
```diff
@@ -51,6 +51,14 @@ var (
...
+	// Key format is: 0x2a H(<runtime-id>) <round>
```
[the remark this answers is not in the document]
abukosek: All the other definitions do not have blank lines in them, this would break the existing style used in the file.

go/consensus/cometbft/apps/roothash/state/state.go
```diff
@@ -270,6 +278,71 @@ func (s *ImmutableState) IncomingMessageQueue(ctx context.Context, runtimeID com
...
+	var roots roothash.PastRoots
+	if err = cbor.Unmarshal(raw, &roots); err != nil {
+		return nil, api.UnavailableStateError(err)
+	}
+	return &roots, nil
+}
+
+// PastRoots returns the past N state and I/O roots stored for the given
+// runtime ID.
+func (s *ImmutableState) PastRoots(ctx context.Context, runtimeID common.Namespace) (map[uint64]roothash.PastRoots, error) {
```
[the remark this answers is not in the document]
abukosek: Yes, because that method can fail and return `nil` for the roots.  This is consistent with the usage elsewhere in the codebase.

go/consensus/cometbft/apps/roothash/messages.go
abukosek: I think this microoptimization is unnecessary, since the entire change parameters call is going to be slow anyway.

go/consensus/cometbft/apps/roothash/state/state.go
```diff
@@ -303,6 +376,82 @@ func (s *MutableState) SetRuntimeState(ctx context.Context, state *roothash.Runt
...
+			// Nothing to delete.
+			continue
+		}
+
+		numPastRootsToDelete := numStoredRoots - max
+
+		it := s.is.NewIterator(ctx)
+		defer it.Close()
+
+		var keysToRemove [][]byte
```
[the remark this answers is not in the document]
abukosek: I don't feel comfortable skipping the verification steps, especially since they're not going to be the bottleneck here.
We could save a tiny bit of RAM by doing removals on the fly, though, so that's maybe worth doing as an optimization later if necessary.
I've modelled this after `RemoveExpiredEvidence`, which also first collects all the keys and then removes them.

go/consensus/cometbft/apps/roothash/state/state.go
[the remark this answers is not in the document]
peternose:
```suggestion
	// pastRootsFmt is the key format for the previous state and I/O runtime roots.
```
The comment should start with the name of the variable, like others do.

go/roothash/api/api.go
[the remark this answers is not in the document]
peternose: Not sure. I just thought that we don't want to store GB of data in the state. I wanted to suggest 1000, but the default value was set to 1200.

go/consensus/cometbft/apps/roothash/state/state.go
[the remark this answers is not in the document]
peternose: Quite the opposite. See for example:
```
	// evidenceKeyFmt is the key format used for storing valid misbehaviour evidence.
	//
	// Key format is: 0x24 <H(runtime-id) (hash.Hash)> <round (uint64)> <evidence-hash (hash.Hash)>
	evidenceKeyFmt = keyformat.New(0x24, keyformat.H(&common.Namespace{}), uint64(0), &hash.Hash{})
```

go/consensus/cometbft/apps/roothash/messages.go
peternose: OK, marking suggestion as resolved.
I'm aware that optimization is negligible, but less code sometimes improves readability.

go/consensus/cometbft/apps/roothash/state/state.go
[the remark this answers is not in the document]
abukosek:
```

```
and
```
//
```
are not the same thing :P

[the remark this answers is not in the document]
peternose: Sorry, didn't notice that. My mistake :smile:

go/roothash/api/api.go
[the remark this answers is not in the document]
abukosek: If space is ever a problem, `MaxPastRootsStored` can always be reduced and the space will be freed up when the parameter changes.
Changing this to something astronomically large probably wouldn't pass a governance vote anyway :)

abukosek:
Ready for re-review :)

go/roothash/api/api.go
peternose, line 577: Should this be `uint64`, similar as age and other stuff that depend on round which is `uint64`?

go/consensus/cometbft/apps/roothash/state/state_test.go
```diff
@@ -147,4 +202,75 @@ func TestSeparateRuntimeRoots(t *testing.T) {
 	ioRoot, err := st.IORoot(ctx, runtime.ID)
 	require.NoError(err, "IORoot")
 	require.EqualValues(blk.Header.IORoot, ioRoot)
+
+	roots, err := st.PastRoundRoots(ctx, runtime.ID)
+	require.NoError(err, "PastRoundRoots")
+	require.EqualValues(1, len(roots))
+	require.EqualValues(len(roots), st.PastRoundRootsCount(ctx, runtime.ID))
```
peternose, line 209:
```suggestion
	require.EqualValues(1, st.PastRoundRootsCount(ctx, runtime.ID))
```

go/consensus/cometbft/apps/roothash/state/state_test.go
```diff
@@ -147,4 +202,75 @@ func TestSeparateRuntimeRoots(t *testing.T) {
...
+	require.EqualValues(blk1.Header.IORoot, roots[1].IORoot)
+
+	rfr1, err = st.RoundRoots(ctx, runtime.ID, 1)
+	require.NoError(err, "RoundRoots 1")
+	require.EqualValues(rfr1.StateRoot, roots[1].StateRoot)
+	require.EqualValues(rfr1.IORoot, roots[1].IORoot)
+
+	// Round 2.
+	blk2 := block.NewEmptyBlock(blk1, 2, block.Normal)
+	err = blk2.Header.StateRoot.UnmarshalHex("0000000000000000000000000000000000000000000000000000000000000003")
```
peternose, line 254:
```suggestion
	err = blk2.Header.StateRoot.UnmarshalHex("0000000000000000000000000000000000000000000000000000000000000005")
```

go/consensus/cometbft/apps/roothash/state/state_test.go
```diff
@@ -147,4 +202,75 @@ func TestSeparateRuntimeRoots(t *testing.T) {
...
+	require.NoError(err, "UnmarshalHex")
+	err = blk2.Header.IORoot.UnmarshalHex("0000000000000000000000000000000000000000000000000000000000000004")
```
peternose, line 256:
```suggestion
	err = blk2.Header.IORoot.UnmarshalHex("0000000000000000000000000000000000000000000000000000000000000006")
```

go/consensus/cometbft/apps/roothash/state/state_test.go
```diff
@@ -147,4 +202,75 @@ func TestSeparateRuntimeRoots(t *testing.T) {
...
+		CurrentBlockHeight: 3,
+		LastNormalRound:    2,
+		LastNormalHeight:   3,
+	})
+	require.NoError(err, "SetRuntimeState")
+
+	roots, err = st.PastRoundRoots(ctx, runtime.ID)
+	require.NoError(err, "PastRoundRoots")
+	require.EqualValues(2, len(roots))
+	require.EqualValues(len(roots), st.PastRoundRootsCount(ctx, runtime.ID))
```
peternose, line 271:
```suggestion
	require.EqualValues(2, st.PastRoundRootsCount(ctx, runtime.ID))
```

go/consensus/cometbft/apps/roothash/state/state_test.go
```diff
@@ -147,4 +202,75 @@ func TestSeparateRuntimeRoots(t *testing.T) {
...
+		CurrentBlockHeight: 2,
+		LastNormalRound:    1,
+		LastNormalHeight:   2,
+	})
+	require.NoError(err, "SetRuntimeState")
+
+	roots, err = st.PastRoundRoots(ctx, runtime.ID)
+	require.NoError(err, "PastRoundRoots")
+	require.EqualValues(2, len(roots))
+	require.EqualValues(len(roots), st.PastRoundRootsCount(ctx, runtime.ID))
```
peternose, line 241:
```suggestion
	require.EqualValues(2, st.PastRoundRootsCount(ctx, runtime.ID))
```

go/consensus/cometbft/apps/roothash/state/state_test.go
```diff
@@ -116,11 +116,66 @@ func TestSeparateRuntimeRoots(t *testing.T) {
...
+	stateRoot, err := st.StateRoot(ctx, runtime.ID)
+	require.NoError(err, "StateRoot")
+	require.EqualValues(blk.Header.StateRoot, stateRoot)
+
+	ioRoot, err := st.IORoot(ctx, runtime.ID)
+	require.NoError(err, "IORoot")
+	require.EqualValues(blk.Header.IORoot, ioRoot)
+}
+
+func TestPastRuntimeRoots(t *testing.T) {
```
peternose, line 159: Unit test for shrink should be in this file.

go/consensus/cometbft/apps/roothash/messages_test.go
```diff
@@ -88,3 +91,360 @@ func TestChangeParameters(t *testing.T) {
...
+
+	// Prepare & apply proposal for disabling storing the roots.
+	changeMaxPastRootsStored(require, app, ctx, state, 0)
+
+	// Switch context.
+	ctx.Close()
+	ctx = appState.NewContext(abciAPI.ContextBeginBlock)
+
+	// Round 7 (0 root pairs stored).
+	_ = advanceRuntimeState(require, ctx, blk, blk6, runtime, 7)
```
peternose, line 439: Do we need to advance to test shrink?

go/consensus/cometbft/apps/roothash/state/state.go
```diff
@@ -303,6 +399,93 @@ func (s *MutableState) SetRuntimeState(ctx context.Context, state *roothash.Runt
...
+
+		numPastRootsToDelete := numStoredRoots - max
+
+		it := s.is.NewIterator(ctx)
+		defer it.Close()
+
+		// We need to pre-hash the runtime ID, so we can compare it below.
+		hID := keyformat.PreHashed(id.Hash())
+
+		var keysToRemove [][]byte
```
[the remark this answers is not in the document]
abukosek: I used to do this, but Yawning told me it makes no difference in the generated code.
The evidence code below also does it the same way.

go/consensus/cometbft/apps/roothash/state/state.go
```diff
@@ -303,6 +399,93 @@ func (s *MutableState) SetRuntimeState(ctx context.Context, state *roothash.Runt
...
+	// Go through all runtimes, so we can delete extra stored past
+	// roots for each one, where it's needed.
+	runtimes, err := s.Runtimes(ctx)
+	if err != nil {
+		return err
+	}
+
+	for _, r := range runtimes {
+		id := r.Runtime.ID
+		numStoredRoots := s.PastRoundRootsCount(ctx, id)
```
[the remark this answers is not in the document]
abukosek: Oh come on, stop code golfing :D

go/roothash/api/api.go
abukosek: I think `uint64` would bloat the storage way too much -- this parameter sets the max past roots stored per runtime, not globally, `uint32` should be more than enough.

go/consensus/cometbft/apps/roothash/state/state_test.go
abukosek: Your change is semantically different from what I want the test to actually test :)

go/consensus/cometbft/apps/roothash/state/state_test.go
abukosek: It cannot, since the change parameters call can only be called from outside the `state` package.

go/consensus/cometbft/apps/roothash/messages_test.go
abukosek: It's nicer this way and more closely models how the code will be used in real life, rather than hacking something together :)

go/consensus/cometbft/apps/roothash/state/state.go
```diff
@@ -303,6 +399,93 @@ func (s *MutableState) SetRuntimeState(ctx context.Context, state *roothash.Runt
...
+		newRoots := cbor.Marshal(roothash.RoundRoots{
+			StateRoot: state.CurrentBlock.Header.StateRoot,
+			IORoot:    state.CurrentBlock.Header.IORoot,
+		})
+
+		// Delete the oldest root to make room for the new one.
+		// We ignore the error because the root might not be there, so it's OK
+		// if this fails.
+		if newRound >= uint64(maxStored) {
+			_ = s.ms.Remove(ctx, pastRootsFmt.Encode(&state.Runtime.ID, newRound-uint64(maxStored)))
```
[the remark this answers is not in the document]
abukosek: It can if the storage is remote :)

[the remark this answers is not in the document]
kostko: It would not fail if the root is not there, but it could fail if there is an actual storage problem. In which case the error should be propagated.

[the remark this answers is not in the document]
abukosek: OK, I'll add that.

go/roothash/api/api.go
kostko: I don't think it would bloat storage as in deterministic CBOR encoding the value would be stored in the most compact way. It makes sense for this to be of the same type as the rounds.

abukosek: I didn't mean the `uint64` parameter, but the actual storage used for past rounds -- do we really need to store more than 4294967296 rounds of state and I/O roots per runtime??

kostko: No, definitely not, I think the argument was just for it to have the same type as the round to avoid casts.

abukosek: Ah, ok, I guess I can change it to `uint64` then.

go/consensus/cometbft/apps/roothash/state/state.go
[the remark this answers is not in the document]
peternose: Strange, as benchmarks show more memory allocations without suggested changes.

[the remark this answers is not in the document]
abukosek: :man_shrugging:
That was a while ago, so maybe they made changes to the compiler :)  I'll change it to your suggestion.

go/consensus/cometbft/apps/roothash/state/state.go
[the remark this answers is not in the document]
peternose: Sorry, but I had to mention, as this could speed things if we decide to shrink 1200 roots. But I never tested this locally. No need to change the code, though.

go/consensus/cometbft/apps/roothash/state/state_test.go
peternose: I just thought it is easier to read this way.

go/consensus/cometbft/apps/roothash/state/state.go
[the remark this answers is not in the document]
abukosek: Yeah, I think we should do these optimizations only after it's clear that we need them, since they make the code less readable :)
We can always make another PR to optimize this if it becomes a bottleneck.

go/consensus/cometbft/apps/roothash/state/state_test.go
abukosek: The `PastRoundRootsCount` method and `len(roots)` are basically the same, except that the method is more optimized -- putting a hardcoded number in this check is slightly confusing to me.

go/consensus/cometbft/apps/roothash/state/state_test.go
peternose: What you are testing is that `ChangeParameters` correctly shrinks roots by calling `ShrinkPastRoots`. I would unit test the latter method alone, e.g. by inserting few roots and check if some are removed.

go/consensus/cometbft/apps/roothash/messages_test.go
peternose: The problem that I saw here was that some roots should be removed immediately after `changeMaxPastRootsStored` is called, but in your test we also call `advanceRuntimeState` which calls `SetRuntimeState` which removes the oldest root. So I cannot be sure that `changeMaxPastRootsStored` removed the roots correctly.

abukosek: OK, I'll add an extra test.

go/consensus/cometbft/apps/roothash/state/state.go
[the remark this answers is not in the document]
peternose: Maybe, have no idea. I read somewhere that the capacity should be doubled if there is not enough space, which the following output confirms (without any compiler optimizations).
```
len: 0 , cap: 0
len: 1 , cap: 1
len: 2 , cap: 2
len: 3 , cap: 4
len: 4 , cap: 4
len: 5 , cap: 8
len: 6 , cap: 8
len: 7 , cap: 8
len: 8 , cap: 8
len: 9 , cap: 16
```

go/consensus/cometbft/apps/roothash/state/state_test.go
abukosek: Done, please take a look :)

go/consensus/cometbft/apps/roothash/state/state.go
```diff
@@ -303,6 +399,93 @@ func (s *MutableState) SetRuntimeState(ctx context.Context, state *roothash.Runt
 	if err := s.ms.Insert(ctx, ioRootKeyFmt.Encode(&state.Runtime.ID), ioRoot); err != nil {
```
peternose, line 399: @kostko Can we remove this? Don't see anyone using it.

peternose (approved):

go/consensus/cometbft/apps/roothash/state/state.go
kostko: AFAIK nothing is using it and now that there are historic roots we could remove it. I/O roots are useful for proving runtime transactions and events.

peternose: Done in https://github.com/oasisprotocol/oasis-core/pull/5377.

kostko: I was only referring to the IO root. We cannot simply remove the state root as the existing runtimes are using it for light client verification.

PR #5366, opened 2023-09-18
Multiple updates from SDK port to 23.0

kostko:
* runtime: Add conversion methods for common::crypto::x25519
* go/oasis-net-runner: Skip policy on non-TEE builds
* runtime: Always use same-block verification for latest state query

### DIFF
--- a/runtime/src/common/crypto/x25519.rs
+++ b/runtime/src/common/crypto/x25519.rs
@@ -48,6 +48,24 @@ impl Default for PrivateKey {
     }
 }
 
+impl AsRef<[u8]> for PrivateKey {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_ref()
+    }
+}
+
+impl From<PrivateKey> for x25519_dalek::StaticSecret {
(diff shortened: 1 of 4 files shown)


runtime/src/common/crypto/x25519.rs
```diff
@@ -61,9 +79,9 @@ impl cbor::Decode for PrivateKey {
 
     fn try_from_cbor_value(value: cbor::Value) -> Result<Self, cbor::DecodeError> {
         let mut bytes: [u8; PRIVATE_KEY_LENGTH] = cbor::Decode::try_from_cbor_value(value)?;
-        let pk = PrivateKey(x25519_dalek::StaticSecret::from(bytes));
+        let sk = PrivateKey(x25519_dalek::StaticSecret::from(bytes));
```
peternose, line 82: I named this `sk` to distinguish the secret key from the public key `pk`. I blame Diffie and Hellman for this problem, as they should have forbidden naming secret key private key :smiley:

peternose (approved):
:+1:

PR #5370, opened 2023-09-20
ci: Remove daily artifacts on successful test run

gw0:
In case of success daily artifacts are not interesting and can be removed, otherwise the script will error out before reaching the added line.

### DIFF
--- a/.buildkite/scripts/daily_txsource.sh
+++ b/.buildkite/scripts/daily_txsource.sh
@@ -5,13 +5,15 @@ set -euxo pipefail
 # Script invoked from .buildkite/longtests.pipeline.yml
 
 if [[ $BUILDKITE_RETRY_COUNT == 0 ]]; then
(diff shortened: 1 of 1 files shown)


.buildkite/scripts/daily_txsource.sh
```diff
@@ -12,6 +12,9 @@ if [[ $BUILDKITE_RETRY_COUNT == 0 ]]; then
         --scenario_timeout 900m \
         --scenario e2e/runtime/txsource-multi \
         "$@"
+
+    # In case of success daily artifacts are not interesting and can be removed.
+    rm -rf /var/tmp/longtests/*
```
peternose, line 17: Can we remove line 8?

peternose: I guess not, because of failed runs.

peternose (approved):

kostko (approved):

PR #5376, opened 2023-09-28
go/oasis-node/cmd: Deprecate some subcommands

kostko:


peternose (approved):
```
➜  oasis-node git:(6b20dbe44) ✗ ./oasis-node consensus -h
Command "consensus" is deprecated, use the `oasis` CLI instead.
```
:+1:

ptrus (approved):

PR #5377, opened 2023-09-28
go/consensus/cometbft/apps/roothash: Remove last state and I/O roots

peternose:


go/consensus/cometbft/apps/roothash/state/state.go
```diff
@@ -31,14 +31,6 @@ var (
 	//
 	// Key format is: 0x24 <H(runtime-id) (hash.Hash)> <round (uint64)> <evidence-hash (hash.Hash)>
 	evidenceKeyFmt = keyformat.New(0x24, keyformat.H(&common.Namespace{}), uint64(0), &hash.Hash{})
-	// stateRootKeyFmt is the key format used for runtime state roots.
-	//
-	// Value is the runtime's latest state root.
-	stateRootKeyFmt = keyformat.New(0x25, keyformat.H(&common.Namespace{}))
```
kostko, line 37: This will break all runtimes, so it cannot be removed.

.changelog/5377.internal.md
kostko, line 1: This is actually very breaking.

peternose: Agree. Decided to postpone changes until runtimes start reading from the past roots (if they ever will).

PR #5382, opened 2023-10-04
buildkite: Fix uploading artifacts when e2e test fails

peternose:
Problem: Failed tests didn't have artifacts.

PR #5383, opened 2023-10-05
buildkite: Fix uploading artifacts when E2E test fails or timeouts

gw0:
E2E tests are missing artifacts if they fail or timeout. This is a bit cleaner alternative to https://github.com/oasisprotocol/oasis-core/pull/5382 that also handles timeouts. It adds a trap signal that executes the artifact upload command no matter the outcome (mimicking the behavior of `artifact_paths:`).

### DIFF
--- a/.buildkite/code.pipeline.yml
+++ b/.buildkite/code.pipeline.yml
@@ -200,8 +200,8 @@ steps:
     depends_on:
       - "build-rust-runtimes"
     command:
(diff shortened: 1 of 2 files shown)


kostko (approved):

peternose (approved):

peternose (approved):
Looks good. If some tests don't respect ctx, they will run more than 20min and we can fix them. If other commands hang up, we can again set `timeout_in_minutes=25`.

PR #5388, opened 2023-10-09
changelog: Assemble changes for 23.0 release

kostko:


peternose (approved):

pro-wh (approved):
reproduced locally

minor differences in the order when it comes to multiple items that applied to the same issue:

- #5050
- #5085
- #5005
- #5007
- #5351

tjanez (approved):

PR #5389, opened 2023-10-09
go/registry: Do not verify node TEE capabilities during genesis

kostko:


ptrus (approved):

ptrus (approved):

peternose:
If we don't validate, should we reset TEE capabilities and let the node register again?

peternose:
> If we don't validate, should we reset TEE capabilities and let the node register again?

Stupid comment, I see that we don't store this data.

peternose (approved):

PR #5390, opened 2023-10-09
host/sgx/epid: ensure attestation API calls hit the same proxy

ptrus:
Refactors the IAS proxy client to expose separate clients for each configured IAS proxy, instead of load-balancing internally between endpoints on a per-request basis.

This is required because the attestation procedure requires three calls to the IAS endpoint (`GetSPIDInfo`, `GetSigRL`, `VerifyEvidence`) and these should contact the same endpoint.

Retries other proxies on failure.

### DIFF
--- a/go/runtime/host/sgx/epid.go
+++ b/go/runtime/host/sgx/epid.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/binary"
 	"fmt"
+	"time"
 
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
@@ -20,6 +21,11 @@ type teeStateEPID struct {
 	teeStateImplCommon
 
 	epidGID uint32
+
+	// prevIAS is the index of the IAS server that was used for the last successful attestation.
+	// This is used as a heuristic to first query the IAS server that is likely able to
+	// successfully do the attestation.
+	prevIAS int
 }
 
 func (ep *teeStateEPID) Init(ctx context.Context, sp *sgxProvisioner, runtimeID common.Namespace, version version.Version) ([]byte, error) {
@@ -36,11 +42,6 @@ func (ep *teeStateEPID) Init(ctx context.Context, sp *sgxProvisioner, runtimeID 
 }
 
 func (ep *teeStateEPID) Update(ctx context.Context, sp *sgxProvisioner, conn protocol.Connection, report []byte, nonce string) ([]byte, error) {
-	spidInfo, err := sp.ias.GetSPIDInfo(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("error while requesting SPID info: %w", err)
-	}
-
 	// Check if new format of attestations is supported in the consensus layer and use it.
 	regParams, err := sp.consensus.Registry().ConsensusParameters(ctx, consensus.HeightLatest)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


.changelog/5390.feature.md
kostko, line 1: Probably should be classified as a bugfix.

go/runtime/host/sgx/epid.go
```diff
@@ -52,17 +44,55 @@ func (ep *teeStateEPID) Update(ctx context.Context, sp *sgxProvisioner, conn pro
 	}
 	supportsAttestationV1 := (regParams.TEEFeatures != nil && regParams.TEEFeatures.SGX.PCS)
 
+	// XXX: Could randomize endpoint order or order it on success rates and or response times.
```
kostko, line 47: I would just put it as a TODO (and open an issue), let's start with a simple solution for now as it is already much better than the old one.

kostko (approved):

.changelog/5390.feature.md
ptrus: Updated

kostko (approved):

go/runtime/host/sgx/sgx.go
```diff
@@ -60,7 +60,7 @@ type Config struct {
 	LoaderPath string
 
 	// IAS is the Intel Attestation Service endpoint.
```
peternose, line 62:
```suggestion
	// IAS are the Intel Attestation Service endpoints.
```

go/runtime/host/sgx/epid.go
```diff
@@ -52,17 +44,54 @@ func (ep *teeStateEPID) Update(ctx context.Context, sp *sgxProvisioner, conn pro
...
+	// but (re-)attestation are not so frequent that it would matter much.
+	for i, iasClient := range sp.ias {
+		resp, err := ep.update(ctx, sp, conn, report, nonce, supportsAttestationV1, iasClient)
+		switch {
+		case err == nil:
+			return resp, nil
+		case i == len(sp.ias)-1:
+			return nil, err
+		default:
+			sp.logger.Warn("error obtaining attestation, trying next IAS server", "err", err)
```
peternose, line 57: I prefer printing this warn for all errors.
```
		if err != nil {
			sp.logger.Warn("error obtaining attestation, trying next IAS server", "err", err)

			if i == len(sp.ias)-1 {
				return nil, err
			}

			select {
			case <-time.After(50 * time.Millisecond):
				continue
			case <-ctx.Done():
				return nil, ctx.Err()
			}
		}
```

go/runtime/host/sgx/epid.go
```diff
@@ -52,17 +44,54 @@ func (ep *teeStateEPID) Update(ctx context.Context, sp *sgxProvisioner, conn pro
 	}
 	supportsAttestationV1 := (regParams.TEEFeatures != nil && regParams.TEEFeatures.SGX.PCS)
 
+	// TODO: Could consider implementing a strategy for more optimized endpoint selection,
```
peternose, line 47: Agree, we don't need to do anything advanced, but we could at least start where we ended.

go/runtime/host/sgx/epid.go
peternose: We should update metrics here (see `teeAttestationsFailed`).

go/runtime/host/sgx/epid.go
```diff
@@ -52,17 +44,54 @@ func (ep *teeStateEPID) Update(ctx context.Context, sp *sgxProvisioner, conn pro
...
+		resp, err := ep.update(ctx, sp, conn, report, nonce, supportsAttestationV1, iasClient)
+		switch {
+		case err == nil:
+			return resp, nil
+		case i == len(sp.ias)-1:
+			return nil, err
+		default:
+			sp.logger.Warn("error obtaining attestation, trying next IAS server", "err", err)
+			select {
+			case <-time.After(50 * time.Millisecond):
```
peternose, line 59: Why do we need to wait?

ptrus: Seemed reasonable to backoff at least a tiny bit before retrying the next server

go/runtime/host/sgx/epid.go
ptrus: Should we? It depends on what the intended measurement of the metric is. Do we want to record a failed attestation if the problem is on (one of the) proxies?

peternose: Maybe not. Actually, we can even add another metrics latter if we need to. Just that it is clear to operators what metrics mean.

peternose (approved):

go/runtime/host/sgx/epid.go
ptrus: Added, check again.

peternose (approved):

PR #5405, opened 2023-10-17
Database loading statuses

jberci:


go/worker/storage/worker.go
```diff
@@ -95,9 +97,13 @@ func (w *Worker) registerRuntime(commonNode *committeeCommon.Node, checkpointerC
 		}
 	}
 
-	localStorage, err := NewLocalBackend(commonNode.Runtime.DataDir(), id)
-	if err != nil {
-		return fmt.Errorf("can't create local storage backend: %w", err)
+	storageCtor := func(ctx context.Context) (storageApi.LocalBackend, error) {
```
kostko, line 100: Is there a reason why we don't create the local storage backend in `committee/node`? This would avoid passing in this factory function.

go/worker/storage/committee/node.go
```diff
@@ -200,6 +202,31 @@ func NewNode(
 
 	n.ctx, n.ctxCancel = context.WithCancel(context.Background())
 
+	go func() {
```
kostko, line 205: Isn't doing this async dangerous? Where do we ensure that start is not called before this completes? Related goroutines should probably block on a channel that gets closed once this deferred init completes?

go/worker/storage/worker.go
```diff
@@ -95,9 +97,13 @@ func (w *Worker) registerRuntime(commonNode *committeeCommon.Node, checkpointerC
...
+		localStorage, err := NewLocalBackend(commonNode.Runtime.DataDir(), id)
+		if err != nil {
+			return nil, fmt.Errorf("can't create local storage backend: %w", err)
+		}
+		commonNode.Runtime.RegisterStorage(localStorage)
```
kostko, line 105: Are there any issues if this happens async?

PR #5407, opened 2023-10-19
fix: control status blocking before chain init with runtimes

ptrus:
The first commit adds the early-query scenario for runtimes and does some minor needed refactoring so that the base `Scenario` struct implements the `Scenario` interface.

Second one fixes the issue (`runtime.Host()` doesn't wait for the registry descriptor anymore, which was blocking the GetStatus query).

TODO:  also backport to stable/23.0.x once approved


go/oasis-test-runner/scenario/e2e/runtime/early_query.go
```diff
@@ -10,31 +10,39 @@ import (
...
 	}
 
 	// EarlyQueryInitHeight is the same as EarlyQuery scenario but with an initial height set.
 	EarlyQueryInitHeight scenario.Scenario = &earlyQueryImpl{
-		Scenario:      *NewScenario("early-query/init-height"),
+		Scenario:      e2e.NewScenario("early-query/init-height"),
 		initialHeight: 42,
 	}
+
+	EarlyQueryRuntime scenario.Scenario = &earlyQueryImpl{
```
kostko, line 29: Add documentation comment.

go/oasis-test-runner/scenario/e2e/runtime/early_query.go
```diff
@@ -105,5 +113,20 @@ func (sc *earlyQueryImpl) Run(ctx context.Context, _ *env.Env) error {
 		return fmt.Errorf("node does not report itself to be a validator at genesis")
 	}
 
+	if sc.runtime {
+		// Context witha bit longer timeout for the runtime node get status query.
```
kostko, line 117:
```suggestion
		// Context with a bit longer timeout for the runtime node get status query.
```

go/oasis-node/cmd/node/node_control.go
```diff
@@ -332,7 +332,7 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
 		}
 
 		// Fetch provisioner type.
-		_, provisioner, err := rt.Host(ctx)
+		_, provisioner, err := rt.Host()
 		switch {
 		case err != nil:
```
kostko, line 337: Should we also set provisioner to something if there is an error? Maybe `"unknown"` or `"pending"` similar?

ptrus: `"pending"` makes sense to me

go/oasis-test-runner/scenario/e2e/runtime/early_query.go
ptrus: Check again please, I also changed so that in the runtime scenario all queries are done against the compute node (to not re-test querying the validator as it is already done in other tests)

PR #5408, opened 2023-10-19
fix consensus light client status shows empty peer IDs

ptrus:
Alternatively could also set it to something like `<uninitialized>` in the status instead of skipping.

TODO: backport


kostko (approved):

kostko:
> Alternatively could also set it to something like <uninitialized> in the status instead of skipping.

No, I think it makes more sense to just skip these.

PR #5410, opened 2023-10-19
config/migrate: Automatically configure external P2P addresses for validators

ptrus:


go/oasis-node/cmd/config/migrate/migrate.go
```diff
@@ -724,13 +726,45 @@ func doMigrateConfig(cmd *cobra.Command, args []string) {
...
+			if err != nil {
+				logger.Error("failed to parse URI from consensus.external_address", "err", err)
+				os.Exit(1)
+			}
+			// Set P2P port if not set.
+			p := url.Port()
+			if _, ok := m(newCfg["p2p"])["port"]; !ok && p != "" {
+				// Port to integer.
+				if pp, err := strconv.Atoi(p); err == nil {
+					m(newCfg["p2p"])["port"] = pp
```
kostko, line 755: Port cannot be the same as the CometBFT P2P port as this is serving a different protocol and such a configuration will not work. If the port is not set, I would use a default one. Also make sure that the `p2p.port` is used in `p2p.registration.addresses` by default (and not the one from `external_address`).

ptrus: Makes sense, I misunderstood the issue description. I guess 9200 makes sense as a default (similar to known seed)?

go/oasis-node/cmd/config/migrate/migrate.go
```diff
@@ -724,13 +725,41 @@ func doMigrateConfig(cmd *cobra.Command, args []string) {
...
+			url, err := url.Parse(eaStr)
+			if err != nil {
+				logger.Error("failed to parse URI from consensus.external_address", "err", err)
+				os.Exit(1)
+			}
+			// Set P2P port if not set.
+			if _, ok := m(newCfg["p2p"])["port"]; !ok {
+				m(newCfg["p2p"])["port"] = 9200
+			}
+			// Set P2P registration addresses if not set.
```
ptrus, line 753: Would it make sense for this to append the address in the (weird) edge case where "p2p.port" would not be set, but "p2p.registration.addresses" would be set?

ptrus: Yeah maybe, but then we'd need to check again that the port used there is not already used for CometBFT.

kostko: Yeah, maybe just do nothing in that case and just emit a warning?

ptrus: Done :+1:

go/oasis-node/cmd/config/migrate/migrate.go
```diff
@@ -724,13 +725,43 @@ func doMigrateConfig(cmd *cobra.Command, args []string) {
...
+			if _, ok := m(m(newCfg["p2p"])["registration"])["addresses"]; !ok {
+				mkSubMap(m(newCfg["p2p"]), "registration")
+				m(m(newCfg["p2p"])["registration"])["addresses"] = []string{url.Hostname() + ":9200"}
```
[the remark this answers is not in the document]
ptrus: Good idea, also refactored a bit since now the "external_address" is only needed for the "registartion.addresses" part.

kostko (approved):

PR #5421, opened 2023-10-30
Install setuptools to make Towncrier fork work with Python 3.12

buberdds:
Keep CI alive while we do not have Towncrier fork updated 
sample issue from other repo https://github.com/oasisprotocol/oasis-wallet-web/actions/runs/6692075152/job/18180527014?pr=1747

### DIFF
--- a/.changelog/5421.internal.md
+++ b/.changelog/5421.internal.md
@@ -1,0 +1,1 @@
+Install setuptools to make Towncrier fork work with Python 3.12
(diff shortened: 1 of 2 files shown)


tjanez:
> Can we solve this problem with `pip install setuptools`?

Indeed, in Python 3.12 the `setuptools` are no longer pre-installed in virtual environments so one needs to explicitly install them via `pip install setuptools` [[1]][release], [[2]][gh].

[release]: https://docs.python.org/3.12/whatsnew/3.12.html
[gh]: https://github.com/python/cpython/issues/95299

tjanez (approved):

peternose (approved):

PR #5426, opened 2023-11-06
go/worker/compute/executor: Minor fixes

peternose:


go/worker/compute/executor/committee/discrepancy.go
```diff
@@ -15,6 +15,18 @@ type discrepancyEvent struct {
 }
 
 func (n *Node) handleDiscrepancy(ctx context.Context, ev *discrepancyEvent) {
+	// TODO: Add round number to discrepancy events (breaking change)
```
kostko, line 18: Not consensus breaking as non-provable events are not part of the events root and are only processed locally. The change is somewhat breaking for downstream tooling that may not expect the new fields, but with 23.0.x the gRPC clients should handle extra fields gracefully.

go/worker/compute/executor/committee/hooks.go
```diff
@@ -35,12 +33,13 @@ func (n *Node) HandleNewBlockEarlyLocked(*runtime.BlockInfo) {
...
+	// Stop sending blocks if the worker has stopped.
 	select {
-	case <-n.blockInfoCh:
+	case <-n.stopCh:
+		return
 	default:
 	}
 
-	// Non-blocking send.
+	// Don't skip any blocks, as the round worker must finalize the final round.
```
kostko, line 31: Could instead use a ring channel with a size of 1, so that we always only queue the last block in case the worker falls behind? Or would that still cause issues?

kostko (approved):
Also don't forget to backport.

go/worker/compute/executor/committee/hooks.go
peternose: Removed this commit, as we don't need to process all blocks. If we skip some, the only problem will be that processed transactions won't be removed from the pool. However, they will be removed latter as their check will fail.

go/worker/compute/executor/committee/discrepancy.go
peternose: I wasn't aware of that. I have resolved this TODO, so please take a look at this commit, as well as the last two, which have been recently added.

kostko (approved):

PR #5427, opened 2023-11-07
common/config: move internal UNIX socket path option to config file

ptrus:
TODO: backport

### DIFF
--- a/go/oasis-node/cmd/common/grpc/grpc.go
+++ b/go/oasis-node/cmd/common/grpc/grpc.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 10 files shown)


kostko (approved):

go/oasis-node/cmd/common/grpc/grpc.go
```diff
@@ -75,15 +68,10 @@ func NewServerLocal(installWrapper bool) (*cmnGrpc.Server, error) {
 	if dataDir == "" {
```
peternose, line 68: Can this also be removed?

ptrus: Yeah I'll look into removing this in a follow up as this was already merged. I don't see a reason why the datadir was required here specifically - even before the change in this PR.

peternose: > I don't see a reason why the datadir was required here specifically - even before the change in this PR.

To construct the path, I guess.

ptrus: Yeah but this didn't change with the PR. It still uses the same data dir config value to construct it (unless overriden).

PR #5428, opened 2023-11-07
go/worker: fix storageCommitLatency timing

ptrus:
golangci-lint >=1.55.0 found this. (should also update version used in CI)


kostko (approved):
Looks like #5426 also includes a similar fix :)

ptrus:
I see :D  i'll close this then

PR #5433, opened 2023-11-08
go/runtime/host/multi: Propagate consensus sync requests to next version

kostko:
Previously periodic consensus sync requests were not propagated to the next (e.g. upcoming) runtime version. This could result in the runtime's consensus view going stale which would make the attestations too old so they would be rejected during scheduling.


peternose (approved):
Looks good, I would just reorder lines to keep `host := agg.active.host` together with `if` check, and `host.Call` together with its error handling.

We could also call the next host in a go routine, to speed things up. Not sure if needed, though.

Not part of this PR, but it looks like call could be removed from the retry. It would be more obvious that call to next rt will only be made once if the retried multiple times.

```
	var (
		activeHost host.Runtime
		nextHost   host.Runtime
	)

	callFn := func() error {
		agg.l.RLock()
		defer agg.l.RUnlock()

		if agg.active == nil {
			return ErrNoActiveVersion
		}
		activeHost = agg.active.host

		if agg.next != nil {
			nextHost = agg.next.host
		}

		return nil
	}
```

go/runtime/host/multi/multi.go
```diff
@@ -143,32 +143,69 @@ func (agg *Aggregate) GetCapabilityTEE() (*node.CapabilityTEE, error) {
...
+		_, nextErr := nextHost.Call(ctx, body)
+		if nextErr != nil {
+			agg.logger.Warn("failed to propagate runtime request to next version",
+				"id", agg.ID(),
+				"err", nextErr,
+			)
+		}
+	}
+
+	return rsp, err
```
peternose, line 208:
```suggestion
	return activeHost.Call(ctx, body)
```
Could then also rename `nextErr` to `err`.

go/runtime/host/multi/multi.go
```diff
@@ -143,32 +143,69 @@ func (agg *Aggregate) GetCapabilityTEE() (*node.CapabilityTEE, error) {
...
+	// time SetVersion is being called to update the version with the cross node lock acquired.
+
+	rsp, err := activeHost.Call(ctx, body)
+
+	// Check if request should be propagated to the next version.
+	if nextHost != nil && shouldPropagateToNextVersion(body) {
+		_, nextErr := nextHost.Call(ctx, body)
+		if nextErr != nil {
+			agg.logger.Warn("failed to propagate runtime request to next version",
+				"id", agg.ID(),
```
peternose, line 202: In another PR we could also update the logger, e.g. `logging.GetLogger("runtime/host/multi").With("id", id)`. I guess here the version is important, but it can also be read from the logs.

go/runtime/host/multi/multi.go
```diff
@@ -143,32 +143,69 @@ func (agg *Aggregate) GetCapabilityTEE() (*node.CapabilityTEE, error) {
...
+func shouldPropagateToNextVersion(body *protocol.Body) bool {
+	switch {
+	case body.RuntimeConsensusSyncRequest != nil:
+		// Consensus view of the next version should be up to date as otherwise signed attestations
+		// will be stale, resulting in them being rejected by the consensus layer.
+		return true
+	case body.RuntimeKeyManagerPolicyUpdateRequest != nil,
+		body.RuntimeKeyManagerStatusUpdateRequest != nil,
+		body.RuntimeKeyManagerQuotePolicyUpdateRequest != nil:
+		// Key manager updates should be propagated so that the runtime is ready when activated.
```
peternose, line 157: This won't work as the new runtimes will fail to decode `body`. The reason is that old key managers support policy updates (and not status updates), while new ones support status updates (and not policy updates).

See: https://github.com/oasisprotocol/oasis-core/blob/master/go/runtime/registry/host.go#L649-L658

Solutions;
- Do not propagate `RuntimeKeyManagerPolicyUpdateRequest`.
- Propagate `RuntimeKeyManagerPolicyUpdateRequest` only if `nextRT.Features.KeyManagerStatusUpdates == false` (probably the same as previous one). 
- Update `host.go` (a lot of work).

Policy updates will be removed once we deploy 23.0.x.

kostko: So the only problematic case that will be seen in practice (e.g. no downgrades) is where the active version is old (supporting policy updates) and the next version is new (supporting status updates). In this case the new runtime will not recognize the method and return an error (which will be logged but otherwise ignored).

This means that maybe we don't need to handle this explicitly?

peternose: Discussed offline. The new runtime will not explode and will treat `RuntimeKeyManagerPolicyUpdateRequest` as invalid. So this can be marked as resolved.

PR #5434, opened 2023-11-08
[BACKPORT/23.0.x] go/runtime/host/multi: Propagate consensus sync requests to next version

kostko:
Backport of #5433.

### DIFF
--- a/go/runtime/host/multi/multi.go
+++ b/go/runtime/host/multi/multi.go
@@ -143,32 +143,67 @@ func (agg *Aggregate) GetCapabilityTEE() (*node.CapabilityTEE, error) {
 	return agg.active.host.GetCapabilityTEE()
 }
 
+// shouldPropagateToNextVersion checks whether the given runtime request should also be propagated
+// to the next version that is pending activation.
+func shouldPropagateToNextVersion(body *protocol.Body) bool {
+	switch {
+	case body.RuntimeConsensusSyncRequest != nil:
+		// Consensus view of the next version should be up to date as otherwise signed attestations
+		// will be stale, resulting in them being rejected by the consensus layer.
+		return true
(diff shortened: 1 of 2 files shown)


go/runtime/host/multi/multi.go
```diff
@@ -143,32 +143,69 @@ func (agg *Aggregate) GetCapabilityTEE() (*node.CapabilityTEE, error) {
...
+func shouldPropagateToNextVersion(body *protocol.Body) bool {
+	switch {
+	case body.RuntimeConsensusSyncRequest != nil:
+		// Consensus view of the next version should be up to date as otherwise signed attestations
+		// will be stale, resulting in them being rejected by the consensus layer.
+		return true
+	case body.RuntimeKeyManagerPolicyUpdateRequest != nil,
+		body.RuntimeKeyManagerStatusUpdateRequest != nil,
+		body.RuntimeKeyManagerQuotePolicyUpdateRequest != nil:
+		// Key manager updates should be propagated so that the runtime is ready when activated.
```
peternose, line 157: This won't work as the new runtimes will fail to decode `body`. The reason is that old key managers support policy updates (and not status updates), while new ones support status updates (and not policy updates).

See: https://github.com/oasisprotocol/oasis-core/blob/master/go/runtime/registry/host.go#L649-L658

Solutions;
- Do not propagate `RuntimeKeyManagerPolicyUpdateRequest`.
- Propagate `RuntimeKeyManagerPolicyUpdateRequest` only if `nextRT.Features.KeyManagerStatusUpdates == false` (probably the same as previous one). 
- Update `host.go` (a lot of work).

Policy updates will be removed once we deploy 23.0.x.

peternose (approved):

PR #5450, opened 2023-11-16
runtime: Create controller in RpcClient constructor

kostko:
Previously if no RPC calls were initiated by the runtime, the client controller task was never spawned which caused quote policy update requests to pile up in the command queue, eventually blocking the entire runtime from processing requests.

Since the async runtime is now available early on during initialization, we can spawn the controller in the RpcClient constructor, avoiding these problems.

### DIFF
--- a/runtime/src/enclave_rpc/client.rs
+++ b/runtime/src/enclave_rpc/client.rs
@@ -1,12 +1,5 @@
 //! Enclave RPC client.
-use std::{
-    collections::HashSet,
-    mem,
-    sync::{
-        atomic::{AtomicBool, Ordering},
-        Arc, Mutex,
-    },
-};
+use std::{collections::HashSet, mem, sync::Arc};
 
 use anyhow;
 use thiserror::Error;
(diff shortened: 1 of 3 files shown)


ptrus (approved):

runtime/src/enclave_rpc/client.rs
```diff
@@ -365,18 +353,18 @@ impl RpcClient {
...
+        // Create the controller task and start it.
+        let controller = Controller {
+            max_retries: 3,
+            nodes,
+            session: MultiplexedSession::new(builder),
+            transport,
+            cmdq: rx,
+            cmdq_tx: tx.clone(),
+        };
+        tokio::spawn(controller.run());
```
peternose, line 365: Looks good, although I'm not a big fan of creating tasks or goroutines in the constructor, as they are hidden from the caller, not under caller's control and they usually cause troubles in tests (in this case, one must ensure Tokio runtime is available).

kostko: True, but this way it avoids a needless atomic + mutex.

PR #5456, opened 2023-11-20
go/p2p: Improve seed node performance

peternose:
Some nodes cannot bootstrap because the seed node has too many open connections.
```
{"attempted":1,"caller":"scope.go:584","current":77,"direction":"Inbound","edge":"system","error":"system: cannot reserve inbound connection: resource limit exceeded","level":"debug","limit":77,"module":"libp2p:rcmgr","msg":"blocked connection from constraining edge","scope":"conn-9516","stat":{"NumStreamsInbound":0,"NumStreamsOutbound":0,"NumConnsInbound":77,"NumConnsOutbound":0,"NumFD":77,"Memory":0},"ts":"2023-11-16T21:09:27.902582862Z","usefd":true}
{"caller":"listener.go:99","error":"conn-9516: system: cannot reserve inbound connection: resource limit exceeded","level":"debug","module":"libp2p:upgrader","msg":"resource manager blocked accept of new connection","ts":"2023-11-16T21:09:27.902778381Z"}
```
The problem lies in the default settings of the resource manager, which are set too low. Consequently, the connection manager never trims connections down to the low watermark (set at 100 by default) because the high watermark (set at 130 by default) is never attained. Increasing the number of inbound connections for seed nodes to 128+ (extra connections come from autoscaling), should solve the problem.

### DIFF
--- a/go/p2p/discovery/bootstrap/client.go
+++ b/go/p2p/discovery/bootstrap/client.go
@@ -133,6 +133,13 @@ func (c *client) Advertise(ctx context.Context, ns string, _ ...discovery.Option
 
 	pf.RecordSuccess()
 
+	// Close connections after every call because requests to the seed node are infrequent.
+	if err = c.rc.Close(c.seed.ID); err != nil {
+		c.logger.Warn("failed to close connections to seed node",
+			"err", err,
+		)
+	}
+
 	return res.TTL, nil
 }
 
@@ -237,6 +244,13 @@ func (c *client) fetchPeers(ctx context.Context, ns string, limit int) []peer.Ad
 		c.nextDiscovery = now.Add(c.backoff.NextBackOff())
 
 		pf.RecordFailure()
+	}
+
+	// Close connections after every call because requests to the seed node are infrequent.
+	if err := c.rc.Close(c.seed.ID); err != nil {
+		c.logger.Warn("failed to close connections to seed node",
+			"err", err,
+		)
 	}
 
 	return cache.peers
--- a/go/p2p/host.go
+++ b/go/p2p/host.go
@@ -8,7 +8,9 @@ import (
 	"github.com/libp2p/go-libp2p"
 	"github.com/libp2p/go-libp2p/core"
 	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
+	rcmgr "github.com/libp2p/go-libp2p/p2p/host/resource-manager"
 	"github.com/libp2p/go-libp2p/p2p/net/conngater"
 	"github.com/libp2p/go-libp2p/p2p/net/connmgr"
 	"github.com/multiformats/go-multiaddr"
@@ -36,6 +38,12 @@ type HostConfig struct {
 func NewHost(cfg *HostConfig) (host.Host, *conngater.BasicConnectionGater, error) {
 	id := api.SignerToPrivKey(cfg.Signer)
 
+	// Set up a resource manager so that we can reserve more resources.
+	rm, err := NewResourceManager()
+	if err != nil {
+		return nil, nil, err
+	}
+
 	// Set up a connection manager so we can limit the number of connections.
 	cm, err := NewConnManager(&cfg.ConnManagerConfig)
 	if err != nil {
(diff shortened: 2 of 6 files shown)


go/p2p/discovery/bootstrap/client.go
```diff
@@ -133,6 +133,13 @@ func (c *client) Advertise(ctx context.Context, ns string, _ ...discovery.Option
 
 	pf.RecordSuccess()
 
+	// Close connections after every call because requests to the seed node are infrequent.
```
kostko, line 136: We should also do the same on the server side to handle clients that don't close connections.

go/p2p/host.go
```diff
@@ -197,3 +206,28 @@ func (cfg *ConnGaterConfig) Load() error {
...
+func NewResourceManager() (network.ResourceManager, error) {
+	// Use the default resource manager for non-seed nodes.
+	if config.GlobalConfig.Mode != config.ModeSeed {
+		return nil, nil
+	}
+
+	// Tweak limits for seed nodes.
+	defaultLimits := rcmgr.DefaultLimits
+	defaultLimits.SystemBaseLimit.ConnsInbound = rcmgr.DefaultLimits.SystemBaseLimit.ConnsOutbound
+	defaultLimits.SystemLimitIncrease.ConnsInbound = rcmgr.DefaultLimits.SystemLimitIncrease.ConnsOutbound
```
ptrus, line 220: Should `StreamsInbound` also be updated, as it is smaller compared to `StreamsOutbound`?

go/p2p/discovery/bootstrap/client.go
peternose: Agree, but its a bit uglier because currently the server only handles streams. And I'm not sure if the caller will receive the data if the connection is closed immediately after the stream.

go/p2p/host.go
```diff
@@ -197,3 +206,28 @@ func (cfg *ConnGaterConfig) Load() error {
...
+// NewConnManager constructs a new connection manager.
+func NewResourceManager() (network.ResourceManager, error) {
+	// Use the default resource manager for non-seed nodes.
+	if config.GlobalConfig.Mode != config.ModeSeed {
+		return nil, nil
+	}
+
+	// Tweak limits for seed nodes.
+	defaultLimits := rcmgr.DefaultLimits
+	defaultLimits.SystemBaseLimit.ConnsInbound = rcmgr.DefaultLimits.SystemBaseLimit.ConnsOutbound
```
ptrus, line 219: Maybe add a comment about what these values are (e.g. 64 -> 128). Imo, it might actually be nicer to specify values here (with some comments) instead of reusing variables from the default limits.

Btw do we expect these to be enough? On mainnet there's a lot more nodes (vs testnet) that will be trying to connect to the seed node. I guess it should be fine assuming the connections will get closed quickly and clients will be retrying.

peternose: Yes, they should be enough. As stated in the description, the connection manager will trim connections if resource manager's limits are larger than the high watermark.

go/p2p/discovery/bootstrap/client.go
peternose: @kostko  Tested locally. If connection is closed after the stream, the caller may receive an error indicating that the stream was reset. Therefore, if the server intends to close connections, it should wait a few seconds before doing so.

Connection handling is the responsibility of the connection manager. To speed up connection closing, we can set the `PeerGracePeriod` configuration for seed nodes to 2 seconds (default 20 seconds), for instance. Optionally, we might consider introducing default configurations for various roles.

kostko: > Connection handling is the responsibility of the connection manager. To speed up connection closing, we can set the PeerGracePeriod configuration for seed nodes to 2 seconds (default 20 seconds), for instance.

Sounds like the cleanest solution to me.

peternose: No, connection is not closed after 2 secs. It is just out of the grace period and if the number of connections is 130+ (equal or above high watermark) it can be closed.

If we want to keep the number of connections low, we would need to set low watermark to 1 and high to 2. This way, all connections but 1 will be closed after 2 seconds. Currently, we cannot set this in the config as we hard coded difference between low and high (30), so changes to the config are needed.

kostko: Ok, that's good as well (e.g. that connections are only closed under stress, as long as this results in new connections succeeding).

kostko (approved):

PR #5462, opened 2023-11-22
go/consensus/cometbft/light: Don't crash when signed header unavailable

kostko:


ptrus (approved):

go/consensus/cometbft/light/service.go
```diff
@@ -179,13 +182,11 @@ func (c *client) worker() {
 	}
 	if err = trustLocalBlock(c.ctx, lastRetainedHeight); err != nil {
 		c.logger.Error("failed to store last retained block into trust store", "err", err)
-		return
 	}
 
 	// Store latest block into trust store.
 	if err = trustLocalBlock(c.ctx, consensus.HeightLatest); err != nil {
```
peternose, line 188: Can this be removed, as `ch` will receive current epoch immediately?

go/consensus/cometbft/light/service.go
```diff
@@ -161,6 +161,9 @@ func (c *client) worker() {
 		if err != nil {
 			return fmt.Errorf("failed to convert protobuff light block into cometbft light block: %w", err)
 		}
+		if tmLb.SignedHeader == nil {
```
peternose, line 164: When is signed header empty?
Can we use `ValidateBasic` here?

go/consensus/cometbft/light/service.go
kostko: No, as the light client constructor would fail if there are no light blocks available in store. So in case both of these fail, the entire light client construction would fail (which is correct and it would be an unusual case, signifying that something must be wrong with the node).

peternose (approved):

go/consensus/cometbft/light/service.go
peternose: > Using `ValidateBasic` here should avoid possible problems in other components of the light client.

Agree. `SaveLightBlock` is ok implemented, as `Store` should be responsible for storing not for validation.

peternose (approved):

PR #5465, opened 2023-11-22
go/sgx/pcs: GetPCKCertificateChain client and minor fixes

ptrus:
As part of work on supporting PPID PCS attestation the `GetPCKCertificateChain` http client was implemented and some minor fixes were done to the existing code.

Although the PPID PCS attestation remains unsupported (due to aesmd api limitations) it makes sense getting this in for potential future implementation.

### DIFF
--- a/go/common/sgx/pcs/http.go
+++ b/go/common/sgx/pcs/http.go
@@ -75,6 +75,7 @@ func (hc *httpClient) doPCSRequest(ctx context.Context, u *url.URL, method, body
 			"method", method,
 			"url", u,
 		)
+		resp.Body.Close()
 		return nil, fmt.Errorf("pcs: response status error: %s", http.StatusText(resp.StatusCode))
 	}
 
--- a/go/common/sgx/pcs/http.go
+++ b/go/common/sgx/pcs/http.go
@@ -1,8 +1,10 @@
 package pcs
 
 import (
+	"bytes"
 	"context"
 	"crypto/x509"
+	"encoding/binary"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
@@ -27,6 +29,7 @@ const (
 	pcsAPIGetTCBInfoPath        = "/certification/v4/tcb"
 	pcsAPIGetQEIdentityPath     = "/certification/v4/qe/identity"
 	pcsAPICertChainHeader       = "TCB-Info-Issuer-Chain"
+	pcsAPIPCKIIssuerChainHeader = "SGX-PCK-Certificate-Issuer-Chain"
 )
 
 // HTTPClientConfig is the Intel SGX PCS client configuration.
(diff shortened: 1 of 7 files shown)


go/runtime/host/sgx/ecdsa.go
```diff
@@ -114,6 +114,10 @@ func (ec *teeStateECDSA) Update(ctx context.Context, sp *sgxProvisioner, conn pr
 		// We have a PPID, need to retrieve PCK certificate first.
 		// TODO: Fetch PCK certificate based on PPID and include it in the quote, replacing the
 		//       PPID certification data with the PCK certificate chain certification data.
+		//       e.g. sp.pcs.GetPCKCertificateChain(ctx, nil, data.PPID, data.CPUSVN, data.PCESVN, data.PCEID)
+		//
+		//	 Due to aesmd QuoteEx APIs not supporting certification data this currently
+		//       cannot be easily implemented. Instead we rely on a quote provider to be installed.
 		return nil, fmt.Errorf("PPID certification data not yet supported")
```
kostko, line 121: Maybe change the error message to say something about the quote provider not being installed?

go/common/sgx/pcs/quote.go
```diff
@@ -456,6 +457,17 @@ func (qs *QuoteSignatureECDSA_P256) Verify(
 		return err
 	}
 
+	// Ensure PCK certification information matches the quote.
```
kostko, line 460: Could these changes break consensus? Or would this just fail with a different error below?

ptrus: I think that the `qs.QESignature.Verify` below should fail unless something weird was going on. But we can skip this change for now to be extra sure.

go/runtime/host/sgx/ecdsa.go
ptrus: Yeah good idea

go/common/sgx/pcs/quote.go
```diff
@@ -456,6 +457,17 @@ func (qs *QuoteSignatureECDSA_P256) Verify(
...
+	if !bytes.Equal(pckInfo.CPUSVN[:], qs.QEReport.CPUSVN[:]) {
+		return fmt.Errorf("pcs/quote: PCK CPUSVN: '%v', does not match QE report CPUSVN: '%v'",
+			hex.EncodeToString(pckInfo.CPUSVN[:]),
+			hex.EncodeToString(qs.QEReport.CPUSVN[:]),
+		)
+	}
+	if pckInfo.PCESVN != int32(header.PCESVN) {
```
peternose, line 467: We should probably change `pcesvn` from `int32` to `int16` everywhere.

go/common/sgx/pcs/http.go
```diff
@@ -133,6 +136,80 @@ func (hc *httpClient) GetTCBBundle(ctx context.Context, fmspc []byte) (*TCBBundl
...
+			PCESVN           string `json:"pcesvn"`
+			PCEID            string `json:"pceid"`
+		}{
+			PlatformManifest: hex.EncodeToString(platformData),
+			CPUSVN:           hex.EncodeToString(cpusvn[:]),
+			PCESVN:           hex.EncodeToString(pcesvnBytes[:]),
+			PCEID:            hex.EncodeToString(pceidBytes[:]),
+		})
+		if merr != nil {
+			return nil, fmt.Errorf("pcs: failed to marshal PCK certificate request payload: %w", err)
```
peternose, line 176:
```suggestion
			return nil, fmt.Errorf("pcs: failed to marshal PCK certificate request payload: %w", merr)
```

go/common/sgx/pcs/quote.go
ptrus: Sure ok. Likely non breaking because a (malovent) non uint16 `pcesvn` would fail somewhere in the validation chain now already.

go/common/sgx/pcs/quote.go
kostko: Let's skip these checks for now.

peternose: Or you can only print the error.

kostko: I would avoid pulling in the logger here. Also it seems this is actually making the unit tests fail so I would assume this check is not entirely per spec.

go/common/sgx/pcs/http.go
```diff
@@ -133,6 +136,80 @@ func (hc *httpClient) GetTCBBundle(ctx context.Context, fmspc []byte) (*TCBBundl
...
+		}{
+			PlatformManifest: hex.EncodeToString(platformData),
+			CPUSVN:           hex.EncodeToString(cpusvn[:]),
+			PCESVN:           hex.EncodeToString(pcesvnBytes[:]),
+			PCEID:            hex.EncodeToString(pceidBytes[:]),
+		})
+		if merr != nil {
+			return nil, fmt.Errorf("pcs: failed to marshal PCK certificate request payload: %w", merr)
+		}
+		rsp, err = hc.doPCSRequest(ctx, u, http.MethodPost, "application/json", bytes.NewReader(payload), false) // nolint: bodyclose
```
peternose, line 178: Not part of this PR, but I think that `doPCSRequest` should call `rsp.Body.Close()` if response status is not OK (same holds for `doIASRequest`).

kostko (approved):

peternose (approved):
Code looks good, but I did not test GET and POST requests. Did you?

go/common/sgx/pcs/http.go
ptrus: Good notice, updated :+1:

peternose (approved):

ptrus:
> Code looks good, but I did not test GET and POST requests. Did you?

Yes did, as part of trying to get the PPID PCS attestation to work, which in the end didn't work (without installing a quote provider) due to a unrelated issue.

PR #5466, opened 2023-11-23
go/consensus/cometbft/light: Try multiple sources when fetching blocks

kostko:


go/consensus/cometbft/light/service.go
```diff
@@ -213,6 +214,7 @@ func (c *client) worker() {
 		providers[0],  // Primary provider.
 		providers[1:], // Witnesses.
 		c.store,
+		cmtlight.PruningSize(config.GlobalConfig.Consensus.Prune.NumLightBlocksKept),
```
ptrus, line 217: Would it make more sense to use 0 (disabled pruning) if `config.GlobalConfig.Consensus.Prune.Strategy` is "none"? 
And then also mention in the docs that it defaults to 1000 if pruning is enabled.

kostko: Good idea, will change to honor the strategy as well.

go/consensus/cometbft/light/service.go
```diff
@@ -254,14 +256,70 @@ func (c *client) GetLightBlock(ctx context.Context, height int64) (*consensus.Li
...
+		directPeerQuerySource,
+	} {
+		lb, pf, err := src()
+		if err == nil {
+			return lb, pf, nil
+		}
+
+		c.logger.Debug("failed to fetch light block from source",
+			"err", err,
+			"source", srcIdx,
```
ptrus, line 315:
```suggestion
			"source_idx", srcIdx,
```

go/consensus/cometbft/light/internal.go
```diff
@@ -27,19 +27,61 @@ type lightClient struct {
...
+
+		var pf rpc.PeerFeedback
+		result, pf, err = fn(provider)
+		if err == nil {
+			return result, pf, nil
+		}
+	}
+
+	// Trigger primary provider refresh if everything fails.
+	providers[0].RefreshPeer()
```
ptrus, line 59: Any reasons to not trigger the refresh also on the witnesses?

peternose: Not sure why this is useful. If remote nodes don't have blocks, their p2p rating will never decrease. Therefore, the peer manager will always serve nodes in the same order, except the first 5, which are shuffled. Since witnesses use 2 nodes, this refresh actually chooses 1 node out of 3.

go/consensus/cometbft/light/service.go
peternose: I would remove this and try to log this in the sources, using the name instead of the idx. In couple of months, I will forgot the order and will have to look at the source code to find out which source failed.

go/consensus/cometbft/light/service.go
kostko: Done.

go/consensus/cometbft/light/service.go
```diff
@@ -254,14 +256,70 @@ func (c *client) GetLightBlock(ctx context.Context, height int64) (*consensus.Li
...
 	}
 
-	// Try local backend first.
-	lb, err := c.consensus.GetLightBlock(ctx, height)
-	if err == nil {
-		return lb, rpc.NewNopPeerFeedback(), nil
+	// Local backend source.
+	localBackendSource := func() (*consensus.LightBlock, rpc.PeerFeedback, error) {
+		lb, err := c.consensus.GetLightBlock(ctx, height)
+		if err == nil {
```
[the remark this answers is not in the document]
kostko: Done.

ptrus (approved):

peternose (approved):

PR #5472, opened 2023-11-24
governance: Fix PrettyPrint for ChangeParametersProposal

ptrus:
Old:
```
Method: governance.SubmitProposal
Body:
  (invalid)
```

New:
```
Method: governance.SubmitProposal
Body:
  Change Parameters:
    Module: staking
    Changes:
      - Parameter: upgrade_cancel_min_epoch_diff
        Value: 15
```

### DIFF
--- a/go/governance/api/api.go
+++ b/go/governance/api/api.go
@@ -4,6 +4,7 @@ package api
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
 	"fmt"
 	"io"
 	"reflect"
@@ -22,10 +23,6 @@ import (
 
 // ModuleName is a unique module name for the governance backend.
 const ModuleName = "governance"
-
-// ProposalContentInvalidText is the textual representation of an invalid
-// ProposalContent.
-const ProposalContentInvalidText = "(invalid)"
 
 var (
 	// ErrInvalidArgument is the error returned on malformed argument(s).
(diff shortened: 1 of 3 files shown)


go/governance/api/api.go
```diff
@@ -125,12 +126,15 @@ func (p *ProposalContent) Equals(other *ProposalContent) bool {
 // given writer.
 func (p ProposalContent) PrettyPrint(ctx context.Context, prefix string, w io.Writer) {
 	switch {
-	case p.Upgrade != nil && p.CancelUpgrade == nil:
+	case p.Upgrade != nil:
```
ptrus, line 129: I think it's fair to assume that the caller validates the proposal before pretty printing it? (that ensures only a single field is set)

kostko (approved):

go/governance/api/api.go
peternose: Or, you could also print all fields if they are set, to not hide that the proposal is invalid.

peternose (approved):

go/governance/api/api.go
```diff
@@ -233,8 +237,18 @@ func (p *ChangeParametersProposal) Equals(other *ChangeParametersProposal) bool
 // PrettyPrint writes a pretty-printed representation of ChangeParametersProposal to the given
 // writer.
 func (p *ChangeParametersProposal) PrettyPrint(_ context.Context, prefix string, w io.Writer) {
+	var changes map[string]interface{}
+	if err := cbor.Unmarshal(p.Changes, &changes); err != nil {
+		fmt.Fprintf(w, "%s  <error: %s>\n", prefix, err)
```
peternose, line 242: Not sure what is better, this
```
Change Parameters:
    <error: cbor: cannot unmarshal positive integer into Go value of type map[string]interface {}>
    <malformed: AQ==>
```
or this 
```
Change Parameters:
    cbor: cannot unmarshal positive integer into Go value of type map[string]interface {}
```

ptrus: No opinions. The `<error>` `<malformed>` is used for unparsable transactions, that's why i used it.

go/governance/api/api.go
ptrus: Yeah i guess

PR #5478, opened 2023-11-27
buildkite: Retry step if cargo build fails

peternose:
Failed step:

```
+ cargo build --all --locked --exclude simple-keyvalue
    Updating crates.io index
error: Unable to update registry `crates-io`
 
Caused by:
  failed to fetch `https://github.com/rust-lang/crates.io-index`
 
Caused by:
  an unknown git error occurred
+ cleanup
+ '[' 0 == 1 ']'
+ CLEANING_UP=1
+ pkill -P 4578
+ true
+ wait
+ CLEANING_UP=0
🚨 Error: The command exited with status 101
user command error: The plugin docker command hook exited with status 101
```

### DIFF
--- a/.buildkite/code.pipeline.yml
+++ b/.buildkite/code.pipeline.yml
@@ -86,6 +86,9 @@ retry: &retry_agent_failure
     # Buildkite lost contact with the agent or it stopped reporting.
     - exit_status: -1
       limit: 2
+    # Cargo build error (e.g., Unable to update registry `crates-io`).
+    - exit_status: 101
+      limit: 10
     # ERRO[0092] error waiting for container: unexpected EOF.
     - exit_status: 125
       limit: 2


peternose:
@gw0

.buildkite/code.pipeline.yml
```diff
@@ -86,6 +86,9 @@ retry: &retry_agent_failure
     # Buildkite lost contact with the agent or it stopped reporting.
     - exit_status: -1
       limit: 2
+    # Cargo build error (e.g., Unable to update registry `crates-io`).
+    - exit_status: 101
+      limit: 10
```
gw0, line 91:
```suggestion
      limit: 2
```
I would suggest a lower retry limit, because 10 retries multiplied by the number of parallel build jobs (without exponential back-off) will definitely not help an already-failing external service to recover.

gw0 (approved):

.buildkite/code.pipeline.yml
peternose: I'm testing what happens if I set limit to 10. Previous version used 2, and all retries failed so it did not solve anything.

gw0:
> Closing, as there are too many retries needed for a step to complete.

If Cargo (or any package manager) is failing it is an upstream issue and they will probably fix it within an hour. Therefore the solution is not to quickly retry multiple times, but to wait for e.g. 5-15 min between retries.

PR #5496, opened 2023-12-01
docker/aesmd: Built epid and pcs versions of the AESMD docker image

ptrus:


.github/workflows/docker.yml
```diff
@@ -100,15 +100,31 @@ jobs:
             org.opencontainers.image.created=${{ steps.determine_tag.outputs.created }}
             org.opencontainers.image.revision=${{ github.sha }}
 
-      - name: "Rebuild ghcr.io/oasisprotocol/aesmd:${{ steps.determine_tag.outputs.tag }}"
+      - name: "Rebuild ghcr.io/oasisprotocol/aesmd:${{ steps.determine_tag.outputs.tag }}-pcs"
```
kostko, line 103: It would be better if we used different image names, e.g. `aesmd-dcap` vs `aesmd-epid`.

PR #5503, opened 2023-12-11
runtime/sgx/pcs: Make useful attestation verification methods public

ptrus:
Make some of the PCS attestation helper functions public so that these can be reused (e.g. in the [attestation-tool](https://github.com/oasisprotocol/tools/tree/main/attestation-tool)). Most of the structs are already public.

### DIFF
--- a/runtime/src/common/sgx/pcs.rs
+++ b/runtime/src/common/sgx/pcs.rs
@@ -195,11 +195,7 @@ impl QuoteBundle {
             .timestamp();
 
         // Perform quote verification.
-        let mut verifier = Verifier {
-            tcb_info,
-            qe_identity,
-            tcb_level: None,
-        };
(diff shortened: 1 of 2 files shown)


peternose (approved):

runtime/src/common/sgx/pcs.rs
```diff
@@ -244,13 +244,14 @@ impl QuoteBundle {
     }
 }
 
-struct Verifier {
+/// Quoting Enclave ECDSA P-256 verifier.
+pub struct QeEcdsaP256Verifier {
```
kostko, line 248: You probably also want to make the fields public as otherwise you won't be able to construct it? Or make a constructor (e.g. `new(tcb_info, qe_identity)`) and an accessor for retrieving the `tcb_level`.

ptrus: Thats right, updated!

kostko (approved):

PR #5512, opened 2024-01-03
go/storage: Add storage metrics for GetDiff

ptrus:


go/storage/api/metrics.go
```diff
@@ -60,6 +61,19 @@ type metricsWrapper struct {
 	Backend
 }
 
+func (w *metricsWrapper) GetDiff(ctx context.Context, request *GetDiffRequest) (WriteLogIterator, error) {
+	start := time.Now()
+	it, err := w.Backend.GetDiff(ctx, request)
```
peternose, line 66: `GetDiff` function returns an iterator and not a complete result like `SyncGet` function below. Therefore, the metric could return values that are not expected unless we iterate over all write logs (e.g. the next write log is fetched from the database in a background go routine).

go/storage/api/metrics.go
```diff
@@ -60,6 +61,19 @@ type metricsWrapper struct {
...
+	storageLatency.With(labelGetDiff).Observe(time.Since(start).Seconds())
```
peternose, line 67: Errors should have lower latency and could corrupt your data. Therefore, it would be nice to also add a label stating whether the call returned and error or not (and maybe which type). But I guess this doesn't matter, as other functions use the same approach.

go/storage/api/metrics.go
ptrus: But this metric is still useful for someone that is aware of this?

go/storage/api/metrics.go
ptrus: Yeah noticed, decided to keep unified with the rest of the code. Can update in future if needed. I think this is fine, as long as one is aware of this.

peternose (approved):

PR #5523, opened 2024-01-14
runtime: Increase number of processing threads in SGX

kostko:


ptrus (approved):

tests/runtimes/simple-keyvalue/Cargo.toml
```diff
@@ -17,7 +17,7 @@ path = "src/upgraded.rs"
 [package.metadata.fortanix-sgx]
 heap-size = 134217728
 stack-size = 2097152
-threads = 8
+threads = 74
```
ptrus, line 20: These threads are bind to CPU cores right, e.g. only cpu-cores number of enclave threads can run at the same time? So it might be be good to update the "recommended" specs to a bit more than 4 in the documentation once this is live, as it likely will be beneficial?

tests/runtimes/simple-keyvalue/Cargo.toml
```diff
@@ -17,7 +17,7 @@ path = "src/upgraded.rs"
 [package.metadata.fortanix-sgx]
 heap-size = 134217728
 stack-size = 2097152
```
ptrus, line 19: I assume stack and heap size values are large enough to handle the potentially increased memory usage due to the more threads?

kostko: Heap size likely needs to be bumped as query caches are thread-local. Will do some adjustments.

kostko: Stack size is per-thread so this should be fine. Bumped heap size and reduced the number of worker threads for now as we can also do internal load balancing across multiple runtime enclaves if needed.

ptrus (approved):

runtime/THREADS.md
```diff
@@ -1,12 +1,12 @@
 # Threads
 
-The following **8 threads** are used by the runtime:
+The following **26 threads** are used by the runtime:
```
peternose, line 3: What is the reason for this increase (e.g. are there any performance issues in production)?

peternose (approved):

runtime/THREADS.md
kostko: Better performance for confidential client nodes that need to handle a lot of queries.

PR #5531, opened 2024-01-18
mkvs/proofs: Remove implicit internal leaf nodes from proofs

ptrus:
Internal MKVS nodes always implicitly contained the full leaf node in the proof. Instead it should be added as a regular child node in the proof. That way an unneded large value (e.g. value in a leaf node of an internal node on the path to the actual node we're proving) won't inflate the proofs size.

Change is backwards compatible, existing proofs are unmarshaled as V0.

TODO:
- [x] use the (commented out from existing test cases) V0 base64 proof encodings in a new test as V0 proof test vectors 
- [x] add a test case with non-nil internal leaf-node 
- [x] update rust part

### DIFF
--- a/go/storage/mkvs/checkpoint/chunk.go
+++ b/go/storage/mkvs/checkpoint/chunk.go
@@ -28,7 +28,11 @@ func createChunk(
 	nextOffset node.Key,
 	err error,
 ) {
-	it := tree.NewIterator(ctx, mkvs.WithProof(root.Hash))
+	it := tree.NewIterator(
+		ctx,
+		// V1 checkpoints use V0 proofs.
+		mkvs.WithProofBuilder(syncer.NewProofBuilderV0(root.Hash, root.Hash)),
+	)
 	defer it.Close()
 
 	// We build the chunk until the proof becomes too large or we have reached the end.
@@ -82,6 +86,7 @@ func restoreChunk(ctx context.Context, ndb db.NodeDB, chunk *ChunkMetadata, r io
 	// Reconstruct the proof.
 	var decodeErr error
 	var p syncer.Proof
+	p.V = checkpointProofsVersion
 	for {
 		if ctx.Err() != nil {
 			return ctx.Err()
--- a/go/storage/mkvs/lookup.go
+++ b/go/storage/mkvs/lookup.go
@@ -140,9 +140,6 @@ func (t *tree) doGet(
 				}
 			}
 
-			// Omit the proof builder as the leaf node is always included with
-			// the internal node itself.
-			opts.proofBuilder = nil
 			return t.doGet(ctx, n.LeafNode, bitLength, key, opts, false)
 		}
 
--- a/go/storage/mkvs/lookup.go
+++ b/go/storage/mkvs/lookup.go
@@ -8,6 +8,9 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/syncer"
 )
 
+// Use version 0 proofs in sync requests for now.
+const syncProofsVersion uint16 = 0
+
 // Implements Tree.
 func (t *tree) Get(ctx context.Context, key []byte) ([]byte, error) {
 	t.cache.Lock()
@@ -48,12 +51,15 @@ func (t *tree) SyncGet(ctx context.Context, request *syncer.GetRequest) (*syncer
 	// Remember where the path from root to target node ends (will end).
 	t.cache.markPosition()
 
-	pb := syncer.NewProofBuilder(request.Tree.Root.Hash, request.Tree.Position)
+	pb, err := syncer.NewProofBuilderForVersion(request.Tree.Root.Hash, request.Tree.Position, request.ProofVersion)
+	if err != nil {
+		return nil, err
+	}
 	opts := doGetOptions{
 		proofBuilder:    pb,
 		includeSiblings: request.IncludeSiblings,
 	}
-	if _, err := t.doGet(ctx, t.cache.pendingRoot, 0, request.Key, opts, false); err != nil {
+	if _, err = t.doGet(ctx, t.cache.pendingRoot, 0, request.Key, opts, false); err != nil {
 		return nil, err
 	}
 	proof, err := pb.Build(ctx)
@@ -75,6 +81,7 @@ func (t *tree) newFetcherSyncGet(key node.Key, includeSiblings bool) readSyncFet
 			},
 			Key:             key,
 			IncludeSiblings: includeSiblings,
+			ProofVersion:    syncProofsVersion,
 		})
 		if err != nil {
 			return nil, err
@@ -127,7 +134,6 @@ func (t *tree) doGet(
 
 		// Does lookup key end here? Look into LeafNode.
 		if key.BitLength() == bitLength {
-			// Include siblings before disabling the proof builder for the leaf node.
 			if opts.includeSiblings {
 				// Also fetch the left and right siblings.
 				_, err = t.doGet(ctx, n.Left, bitLength, key, opts, true)
@@ -140,6 +146,12 @@ func (t *tree) doGet(
 				}
 			}
 
+			if pb := opts.proofBuilder; pb != nil && pb.Version() == 0 {
+				// Omit the proof builder as the leaf node is always included with
+				// the internal node itself in V0 proofs.
+				opts.proofBuilder = nil
+			}
+
 			return t.doGet(ctx, n.LeafNode, bitLength, key, opts, false)
 		}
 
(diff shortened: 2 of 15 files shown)


go/storage/mkvs/syncer/proof.go
```diff
@@ -113,7 +116,10 @@ func (b *ProofBuilder) Size() uint64 {
 
 // Build tries to build the proof.
 func (b *ProofBuilder) Build(ctx context.Context) (*Proof, error) {
-	var proof Proof
+	proof := Proof{
+		Versioned: cbor.NewVersioned(1),
```
kostko, line 120: Add a `LatestProofVersion` constant.

Also probably we would need the builder to support both V0 and V1, so that old runtimes are not affected by this change.

go/storage/mkvs/node/node.go
```diff
@@ -386,8 +386,25 @@ func (n *InternalNode) ExtractUnchecked() Node {
 }
 
 // CompactMarshalBinary encodes an internal node into binary form without
-// any hash pointers (e.g., for proofs).
+// any hash pointers and also doesn't include the leaf node (e.g., for proofs).
 func (n *InternalNode) CompactMarshalBinary() (data []byte, err error) {
```
kostko, line 390: We need to support serialization of V0 proofs for backwards compatibility.

kostko (commented):
Also the read syncer interface should be changed so that it can request a specific version of the proof to make sure that old runtimes will continue to (implicitly) request V0 until they upgrade.

go/storage/mkvs/lookup.go
```diff
@@ -48,12 +48,15 @@ func (t *tree) SyncGet(ctx context.Context, request *syncer.GetRequest) (*syncer
 	// Remember where the path from root to target node ends (will end).
 	t.cache.markPosition()
 
-	pb := syncer.NewProofBuilder(request.Tree.Root.Hash, request.Tree.Position)
+	pb, err := syncer.NewProofBuilderForVersion(request.Tree.Root.Hash, request.Tree.Position, 1)
```
peternose, line 51:
```suggestion
	pb := syncer.NewProofBuilder(request.Tree.Root.Hash, request.Tree.Position)
```

go/storage/mkvs/syncer/proof.go
```diff
@@ -33,21 +44,37 @@ type proofNode struct {
 
 // ProofBuilder is a Merkle proof builder.
 type ProofBuilder struct {
-	root     hash.Hash
-	subtree  hash.Hash
-	included map[hash.Hash]*proofNode
-	size     uint64
+	proofVersion uint16
```
peternose, line 47:
```suggestion
	proofVersion uint8
```

go/storage/mkvs/syncer/proof.go
```diff
@@ -33,21 +44,37 @@ type proofNode struct {
...
+	root         hash.Hash
+	subtree      hash.Hash
+	included     map[hash.Hash]*proofNode
+	size         uint64
 }
 
 // NewProofBuilder creates a new Merkle proof builder for the given root.
 func NewProofBuilder(root, subtree hash.Hash) *ProofBuilder {
 	return &ProofBuilder{
```
peternose, line 56:
```suggestion
	return NewProofBuilderForVersion(root, subtree, LatestProofVersion)
```

go/storage/mkvs/syncer/proof.go
```diff
@@ -33,21 +44,37 @@ type proofNode struct {
...
+		subtree:      subtree,
+		included:     make(map[hash.Hash]*proofNode),
 	}
 }
 
+// NewProofBuilderForVersion creates a new Merkle proof builder for the given root
+// in a given proof version format.
+func NewProofBuilderForVersion(root, subtree hash.Hash, proofVersion uint16) (*ProofBuilder, error) {
+	if proofVersion < MinimumProofVersion || proofVersion > LatestProofVersion {
+		return nil, fmt.Errorf("proof builder: invalid proof version: %d", proofVersion)
```
peternose, line 68:
```suggestion
		return nil, fmt.Errorf("proof: invalid proof version: %d", proofVersion)
```
I prefer not to return errors in constructors, if possible. As the programmer will select the version, we can maybe add a constructor for each version, e.g. `NewProofBuilder`, `NewProofBuilderV0`, `NewProofBuilderV1`, ... and use this constructor only as a helper (if needed).

go/storage/mkvs/checkpoint/chunk.go
```diff
@@ -103,6 +103,7 @@ func restoreChunk(ctx context.Context, ndb db.NodeDB, chunk *ChunkMetadata, r io
 		p.Entries = append(p.Entries, entry)
 	}
 	p.UntrustedRoot = chunk.Root.Hash
+	p.V = 1
```
peternose, line 106:
```suggestion
	p.V = syncer.LatestProofVersion
```

go/storage/mkvs/node/node.go
```diff
@@ -411,13 +407,47 @@ func (n *InternalNode) CompactMarshalBinary() (data []byte, err error) {
...
+	if n.LeafNode == nil {
+		leafNodeBinary = make([]byte, 1)
+		leafNodeBinary[0] = PrefixNilNode
+	} else {
+		if leafNodeBinary, err = n.LeafNode.Node.MarshalBinary(); err != nil {
+			return nil, fmt.Errorf("mkvs: failed to marshal leaf node: %w", err)
+		}
 	}
 
+	data = make([]byte, 1+DepthSize+len(n.Label)+len(leafNodeBinary))
```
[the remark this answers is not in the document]
ptrus: 👍 updated (also elsewhere)

go/storage/mkvs/lookup.go
```diff
@@ -175,6 +180,12 @@ func (t *tree) doGet(
 		}
 
 		if opts.includeSiblings {
+			// Also fetch the leaf.
```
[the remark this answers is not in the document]
ptrus: Updated, used the first suggested version.

go/storage/mkvs/syncer/proof.go
ptrus: > ..As the programmer will select the version. 

That's not true in this case. The ReadSyncer request will contain the proof version (for backwards compatibility so that existing runtimes will get the current versions of proofs returned).

In places where the programmer selects the version the `NewProofBuilder` constructor is used (which also defaults to latest version).

go/storage/mkvs/syncer/proof.go
ptrus: Using `uint16` so that it is consistent with the version in the `Proof`s, which embeds our common `cbor.Versioned` helper struct that uses `uint16` for versions.

go/storage/mkvs/syncer/proof.go
```diff
@@ -33,19 +56,50 @@ type proofNode struct {
...
+		proofVersion: LatestProofVersion,
+		root:         root,
+		subtree:      subtree,
+		included:     make(map[hash.Hash]*proofNode),
+	}
+}
+
+// NewProofBuilderV0 creates a new version 0 proof builder for the given root.
+func NewProofBuilderV0(root, subtree hash.Hash) *ProofBuilder {
+	return &ProofBuilder{
```
kostko, line 78: Could also use `NewProofBuilderForVersion`.

go/storage/mkvs/syncer/proof_test.go
```diff
@@ -35,12 +34,29 @@ func TestProofExtraNodes(t *testing.T) {
...
+
+	_, err = verifier.VerifyProof(context.Background(), rootHash, &proof)
+	require.Error(err, "proof with extra data should fail to validate")
 }
 
 func FuzzProof(f *testing.F) {
 	// Seed corpus.
-	rawProof, _ := base64.StdEncoding.DecodeString("omdlbnRyaWVzhUoBASQAa2V5IDACRgEBAQAAAlghAsFltYRhD4dAwHOdOmEigY1r02pJH6InhiibKlh9neYlWCECpsJnkjOnIgc4+yfvpsqCcIYHh5eld1hNMWTT7arAfHFYIQLhNTLWRbks1RBf52ulnlOTO+7D5EZNMYFzTx8U46sCnm51bnRydXN0ZWRfcm9vdFggWeZ8L9wIuOEN0Iu2uO/mFPzJZey4liX5fxf4fwcQRhM=")
-	f.Add(rawProof)
+	rawProofV0, _ := base64.StdEncoding.DecodeString("omdlbnRyaWVzhUoBASQAa2V5IDACRgEBAQAAAlghAsFltYRhD4dAwHOdOmEigY1r02pJH6InhiibKlh9neYlWCECpsJnkjOnIgc4+yfvpsqCcIYHh5eld1hNMWTT7arAfHFYIQLhNTLWRbks1RBf52ulnlOTO+7D5EZNMYFzTx8U46sCnm51bnRydXN0ZWRfcm9vdFggWeZ8L9wIuOEN0Iu2uO/mFPzJZey4liX5fxf4fwcQRhM=")
```
kostko, line 56: Also re-run the fuzzer.

go/storage/mkvs/syncer/proof.go
```diff
@@ -68,19 +120,43 @@ func (b *ProofBuilder) Include(n node.Node) {
 	// Node is available, serialize it.
 	var err error
 	var pn proofNode
-	pn.serialized, err = n.CompactMarshalBinary()
+	switch b.proofVersion {
+	case 0:
+		// In version 0, the leaf included is in the internal node.
```
kostko, line 125:
```suggestion
		// In version 0, the leaf is included in the internal node.
```

runtime/src/storage/mkvs/sync/proof.rs
```diff
@@ -450,49 +633,65 @@ uO/mFPzJZey4liX5fxf4fwcQRhM=",
...
-        let proof = tree
-            .get_proof(&keys[9])
-            .expect("get proof keys[9] works")
-            .expect("proof keys[9] exists");
-        assert_eq!(
-            base64::encode(cbor::to_vec(proof)),
-		    "omdlbnRyaWVzhUoBASQAa2V5IDACWCECJueKTLbwFMAiJitvfP3+tOruv3XChOjYSpH3U9/Xo/1GAQEDAIACWCECMMFu3slwotsl8hQsxQ/VPkrMtYMEsIrJAUH5PvSglANUAQAFAGtleSA5BwAAAHZhbHVlIDludW50cnVzdGVkX3Jvb3RYIFnmfC/cCLjhDdCLtrjv5hT8yWXsuJYl+X8X+H8HEEYT",
-        );
+        for tc in vec![
+            // Note: Tree::get_proof doesn't rust support version 0 proofs.
```
kostko, line 650:
```suggestion
            // Note: Tree::get_proof doesn't support version 0 proofs.
```

kostko (approved):

PR #5532, opened 2024-01-20
go/worker/keymanager: Refactor access control

peternose:
This is the initial step to decouple the key manager secrets feature from the rest of the codebase. The ultimate goal is to make the methods `CallEnclave` and `localCallEnclave`, and the access list accessible to all workers and watchers.

Workers and watchers:
- Key manager node watcher
- Key manager client runtime watcher and multiple runtime node watchers
- Secrets worker
- CHURP worker

Observe that access control defined by the key manager policy (`MayQuery`) was tightened and moved from the client runtime watcher to the method `authorize`, as this is specific to secrets feature and not to the key manager in general.


go/worker/keymanager/watcher.go
```diff
@@ -90,23 +103,21 @@ func (knw *kmNodeWatcher) watchNodes() {
 			peers[n.P2P.ID] = struct{}{}
 		}
 
-		knw.w.accessList.UpdateNodes(knw.w.runtimeID, nodes)
-		if pm := knw.w.commonWorker.P2P.PeerManager(); pm != nil {
```
peternose, line 87: Anyone knows why was this check useful and if it can be removed?

kostko: My only thought would be maybe when the no-op p2p is in use for archive nodes? But not sure if key manager stuff can even be started in that case? Maybe this is no longer needed since you cannot configure a key manager node in archive mode as the mode can be either archive or key manager.

go/worker/keymanager/watcher.go
```diff
@@ -108,3 +126,203 @@ func (knw *kmNodeWatcher) rebuildActiveNodeIDs(nodeList []*node.Node) map[signat
 
 	return m
 }
+
+// kmRuntimeWatcher is a key manager runtime watcher tasked with monitoring
+// compute runtimes utilizing the specified key manager and initiating
+// a runtime node watcher for each identified runtime.
+type kmRuntimeWatcher struct {
+	sync.RWMutex
```
kostko, line 134: Same as above.

go/worker/keymanager/worker.go
```diff
@@ -177,49 +174,43 @@ func (w *Worker) Initialized() <-chan struct{} {
...
+		method = req.Method
 	default:
 		// Local queries are not allowed.
 		return nil, fmt.Errorf("unsupported RPC kind")
 	}
 
+	// Handle access control.
+	switch {
+	case method == "" && kind == enclaverpc.KindNoiseSession:
+		// Anyone can connect.
```
kostko, line 202: I assume this will be changed in a separate PR?

go/worker/keymanager/worker.go
```diff
@@ -306,6 +297,88 @@ func (w *Worker) localCallEnclave(method string, args interface{}, rsp interface
...
+		}
+	case api.RPCMethodReplicateMasterSecret, api.RPCMethodReplicateEphemeralSecret:
+		// Replication is restricted to peers within the same key manager runtime.
+		if !w.accessList.Runtimes(peerID).Contains(w.runtimeID) {
+			return fmt.Errorf("not a key manager")
+		}
+		return nil
+	}
+
+	panic("unreachable code")
```
kostko, line 379: Why panic here instead of returning an error? This seems dangerous in case something is modified above.

go/worker/keymanager/worker.go
peternose: Yes. A peer will be allowed to connect if it has the authorization to invoke at least one of the methods accessible through the noise session.

go/worker/keymanager/worker.go
peternose: Agree, it safer. I usually panic in this case as execution of this part of the code is not what the programmer wanted, so something is seriously wrong.

go/worker/keymanager/watcher.go
peternose: You are right. The no-op p2p is used only for archive nodes as the `init` method disables the key manager in this case. So it is currently safe to remove this check.

kostko (approved):

PR #5534, opened 2024-01-22
go/oasis-net-runner: Add ability to set node log level and format

abukosek:
Previously, every node started by the net runner had the default
log level of debug, while now it's possible to set it to other
levels.  The log format can also be changed.

### DIFF
--- a/go/oasis-net-runner/cmd/root.go
+++ b/go/oasis-net-runner/cmd/root.go
@@ -139,6 +139,10 @@ func runRoot(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("root: failed to instantiate fixture: %w", err)
 	}
 
+	// Set logging level and format for all spawned nodes.
+	net.Config().NodeLogLevel = viper.GetString(cfgLogLevel)
+	net.Config().NodeLogFormat = viper.GetString(cfgLogFmt)
+
 	// Start the network and keep it running.
 	if err = net.Start(); err != nil {
 		logger.Error("failed to start network",
--- a/go/oasis-test-runner/oasis/network.go
+++ b/go/oasis-test-runner/oasis/network.go
@@ -162,6 +162,12 @@ type NetworkCfg struct { // nolint: maligned
 	// externally-provided.
 	UseShortGrpcSocketPaths bool `json:"-"`
 
+	// NodeLogLevel is the log level to use for created nodes.
+	NodeLogLevel string `json:"node_log_level,omitempty"`
+
+	// NodeLogFormat is the log format to use for created nodes.
+	NodeLogFormat string `json:"node_log_format,omitempty"`
+
 	// Nodes lists the names of nodes to be created, enabling an N:M mapping between physical node
 	// processes and the features they host. If a feature is specified as attached to a node that
 	// isn't listed here, a new node will be created automatically, so this list can normally be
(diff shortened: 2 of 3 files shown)


go/oasis-net-runner/cmd/root.go
```diff
@@ -139,6 +139,10 @@ func runRoot(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("root: failed to instantiate fixture: %w", err)
 	}
 
+	// Set logging level and format for all spawned nodes.
+	net.Config().NodeLogLevel = viper.GetString(cfgLogLevel)
```
peternose, line 143: Feels like this doesn't belong here. It should be in `GetFixture` or together with `var logFmt logging.Format`, not sure. 

Are we abusing `cfgLogFmt` and `cfgLogLevel` here and using them for two different purposes?

go/oasis-test-runner/oasis/network.go
```diff
@@ -638,8 +644,16 @@ func (net *Network) startOasisNode(
 
 	cfg.Common.DataDir = node.DataDir()
 	cfg.Common.Log.Level = make(map[string]string)
-	cfg.Common.Log.Level["default"] = "debug"
-	cfg.Common.Log.Format = "json"
+	if net.Config().NodeLogLevel != "" {
+		cfg.Common.Log.Level["default"] = net.Config().NodeLogLevel
+	} else {
+		cfg.Common.Log.Level["default"] = "debug"
```
peternose, line 650: Should this be moved to `config.DefaultConfig()`?

go/oasis-test-runner/oasis/network.go
```diff
@@ -638,8 +644,16 @@ func (net *Network) startOasisNode(
 
 	cfg.Common.DataDir = node.DataDir()
 	cfg.Common.Log.Level = make(map[string]string)
```
peternose, line 646: Is this necessary? It looks like `config.DefaultConfig` already initializes it. But maybe `Marshal` sets it back to `nil` :shrug:

go/oasis-net-runner/cmd/root.go
abukosek: No, I'm using the same log level and format here on purpose (that's why this code is also here and not in `fixtures.go`) -- the idea is to set just the log level and format for the net runner and it gets automatically inherited for all the nodes it spawns.  This is way more convenient than having a separate flag just for this or setting the log config for each node individually.

go/oasis-test-runner/oasis/network.go
abukosek: Probably not, since setting the default log level to `debug` is not a sensible default and is specific to the test runner.  If the fixture was loaded from a file, the log level and format fields might not be populated, so we still need these checks here.

go/oasis-test-runner/oasis/network.go
abukosek: Yeah, this is just to make sure it's really initialized (it might not be when loading from file if the fields aren't set).

go/oasis-test-runner/oasis/network.go
```diff
@@ -162,6 +162,12 @@ type NetworkCfg struct { // nolint: maligned
 	// externally-provided.
 	UseShortGrpcSocketPaths bool `json:"-"`
 
+	// NodeLogLevel is the log level to use for created nodes.
+	NodeLogLevel string `json:"node_log_level,omitempty"`
```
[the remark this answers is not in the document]
peternose: Yeah, that would be even better.

peternose (approved):

PR #5537, opened 2024-01-23
Include the mode of the node in control status output

ptrus:


.changelog/5537.bugfix.md
```diff
@@ -0,0 +1 @@
+Include the mode of the node in control status output
```
ptrus, line 1: Since cbor decoding in grpc allows extra fields, it might be ok to not consider this change breaking

kostko (approved):

peternose (approved):

PR #5542, opened 2024-01-25
go/runtime/txpool: Remove rechecked transactions from seen cache

kostko:
In case a transaction is rejected because it fails a re-check pass, it should also be removed from the seen cache as it may be resubmitted later when it could become valid.

### DIFF
--- a/go/runtime/txpool/txpool.go
+++ b/go/runtime/txpool/txpool.go
@@ -83,6 +83,11 @@ type TransactionPool interface {
 	// ClearProposedBatch clears the proposal queue.
 	ClearProposedBatch()
 
+	// RejectTxs indicates that given transaction hashes have been rejected during block processing.
+	// Queues that can remove those transactions will do so. Additionally, these transactions will
+	// be removed from the already seen cache as they can potentially become valid in the future.
+	RejectTxs(txs []hash.Hash)
+
 	// HandleTxsUsed indicates that given transaction hashes are processed in a block. Queues that
 	// can remove those transactions will do so.
 	HandleTxsUsed(txs []hash.Hash)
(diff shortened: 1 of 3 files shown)


go/runtime/txpool/txpool.go
```diff
@@ -333,6 +338,16 @@ func (t *txPool) FinishScheduling() {
...
 }
 
+func (t *txPool) RejectTxs(hashes []hash.Hash) {
+	// Remove rejected transactions from the already seen cache to allow them to be resubmitted as
+	// they may become valid in the future.
+	for _, h := range hashes {
+		t.seenCache.Remove(h)
+	}
+
+	t.HandleTxsUsed(hashes)
```
gw0, line 348: A bit confusing... Are rejected transactions "processed in a block" (description of `HandleTxsUsed`)?

kostko: No, but the logic is the same (e.g. they should be removed from queue). Maybe the naming is a bit awkward.

go/runtime/txpool/txpool.go
```diff
@@ -83,6 +83,11 @@ type TransactionPool interface {
 	// ClearProposedBatch clears the proposal queue.
 	ClearProposedBatch()
 
+	// RejectTxs indicates that given transaction hashes have been rejected during block processing.
+	// Queues that can remove those transactions will do so. Additionally, these transactions will
+	// be removed from the already seen cache as they can potentially become valid in the future.
```
peternose, line 88: The final two sentences might be too detailed for an interface as they delve into implementation specifics. But so are some other comments.

peternose (approved):
Looks good. 
- Why is this better than submitting a new transaction?
- Do we simplify DoS attacks with these changes?
- Is it possible that rejected transactions will travel from node to node and never removed from the network?
- Do we need to limit the number of retries?

kostko:
> Why is this better than submitting a new transaction?

Because the same transaction may become valid in the future. The future may also be due to the async nature of the network where some nodes may be slightly behind others.

> Do we simplify DoS attacks with these changes?

Just in the sense that there may be more load on transaction checks (which are very fast). But someone can already generate slightly different transactions and spam those.

> Is it possible that rejected transactions will travel from node to node and never be removed from the network?

Nodes only forward transactions that pass checks. Also the p2p layer has its own deduplication logic.

> Do we need to limit the number of retries?

You mean like having a separate cache that would count the number of times a tx hash was rejected and after a certain amount just immediately reject without passing any checks? We could consider this if it becomes a problem in practice.

peternose:
> some nodes may be slightly behind others.

This makes sense.

> But someone can already generate slightly different transactions and spam those.

If he takes this action now, the impact will be more significant, as the rejected transactions will accumulate in the network. However, according to your answer, these transactions will not be forwarded so I guess this is not a problem.

> You mean like having a separate cache that would count the number of times a tx hash was rejected and after a certain amount just immediately reject without passing any checks?

Yes.

> We could consider this if it becomes a problem in practice.

Ok.

PR #5544, opened 2024-01-29
go/keymanager: Move master and ephemeral secrets code to extension

peternose:
Having code for master and ephemeral secrets and code for churp in the same package is not transparent and introduces naming problems. This PR tries to refactor the code to achieve the following result:
```go
// Backend is a key manager management implementation.
type Backend interface {
	Secrets() secrets.Backend

	Churp() churp.Backend
}
```
Separating runtime suspension (`suspendRuntimes`) and status generation (`generateStatus`) on epoch change makes this PR consensus breaking.


go/worker/keymanager/config/config.go
```diff
@@ -7,6 +7,14 @@ type Config struct {
 	RuntimeID string `yaml:"runtime_id"`
 	// Base64-encoded public keys of unadvertised peers that may call protected methods.
 	PrivatePeerPubKeys []string `yaml:"private_peer_pub_keys"`
+
+	// Churp is configuration for Churp application.
```
kostko, line 11: Did this leak? It shouldn't be part of this PR.

peternose: Yes, will remove.

kostko (approved):

PR #5545, opened 2024-01-29
go/worker/client: Fix observer node registration

kostko:
Previously a node configured as an observer node would forget to register for all of its configured runtimes, causing the registration to fail.

### DIFF
--- a/go/worker/client/worker.go
+++ b/go/worker/client/worker.go
@@ -2,14 +2,18 @@
 package client
 
 import (
+	"fmt"
+
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/grpc"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
+	"github.com/oasisprotocol/oasis-core/go/common/node"
 	"github.com/oasisprotocol/oasis-core/go/config"
 	"github.com/oasisprotocol/oasis-core/go/runtime/client/api"
 	"github.com/oasisprotocol/oasis-core/go/worker/client/committee"
 	workerCommon "github.com/oasisprotocol/oasis-core/go/worker/common"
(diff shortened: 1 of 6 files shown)


go/worker/client/worker.go
```diff
@@ -124,8 +129,19 @@ func (w *Worker) registerRuntime(commonNode *committeeCommon.Node) error {
 		"runtime_id", id,
 	)
 
+	// When a node is a client node and it has an entity configured, we register it with the
+	// observer role as this may be needed for confidential runtimes.
+	var rp registration.RoleProvider
+	if config.GlobalConfig.Mode == config.ModeClient {
+		var err error
+		rp, err = w.registration.NewRuntimeRoleProvider(node.RoleObserver, id)
```
peternose, line 137: Should `rp.SetAvailable(commonNode.RegisterNodeRuntime)` be moved here?

go/worker/client/worker.go
```diff
@@ -124,8 +129,19 @@ func (w *Worker) registerRuntime(commonNode *committeeCommon.Node) error {
 		"runtime_id", id,
 	)
 
+	// When a node is a client node and it has an entity configured, we register it with the
+	// observer role as this may be needed for confidential runtimes.
+	var rp registration.RoleProvider
+	if config.GlobalConfig.Mode == config.ModeClient {
```
peternose, line 135: Can we merge this `if` with `if config.GlobalConfig.Mode == config.ModeStatelessClient` into a switch?

peternose (approved):
Looks good, should solve the problem.

go/worker/client/worker.go
kostko: Actually, we should probably also register for the observer role for stateless clients.

go/worker/client/worker.go
kostko: Done.

peternose (approved):

PR #5548, opened 2024-01-31
go/common/keyformat: Ensure consensus and runtime prefixes are unique

peternose:


kostko (commented):
Could you maybe change this so that you can basically specify a namespace where it needs to be unique? Consensus looks fine, but runtime as it is currently doesn't make sense (e.g. history db keys and runtime transaction tree don't need to have unique prefixes -- those only need to be unique within that IO tree).

peternose:
> history db keys and runtime transaction tree don't need to have unique prefixes

Wasn't sure about that and wanted to simplify. Will add namespaces.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -32,37 +32,37 @@ var (
 	// nodeKeyFmt is the key format for nodes (node hash).
 	//
 	// Value is serialized node.
-	nodeKeyFmt = keyformat.New(0x00, &hash.Hash{})
+	nodeKeyFmt = keyformat.Consensus(0x00, &hash.Hash{})
```
kostko, line 35: This should not be consensus, this is another database for storing runtime and consensus state (both use the same but these keys here are internal). I don't think we should hardcode "runtime history" and all the database backend namespaces in the key format package though?

Could we have a notion of namespaces that are specified externally? E.g. something like:
```golang
// Somewhere we would define the namespace.
var Consensus = keyformat.NewNamespace()
// Then we can uset the namespace to create new key formats.
var fooFmt = Consensus.New(0x00, ...)
```
This (or similar) would make it more flexible to use in external packages without the need to change the `keyformat` package.

peternose: You are right. I reverted some changes and added namespace constructor (with a namespace name to have clearer panic message). Now there are 4 namespaces: consensus, runtime db, runtime tx, and badger. Key formats defined in tests and in migration don't belong to any namespace (duplicates are allowed).

go/storage/mkvs/db/badger/badger.go
```diff
@@ -29,40 +29,43 @@ const (
 )
 
 var (
+	// keyFormat is the namespace for the runtime transaction key formats.
```
kostko, line 32:
```suggestion
	// keyFormat is the namespace for the badger database key formats.
```

go/consensus/api/api.go
```diff
@@ -65,6 +66,9 @@ var (
 	SystemMethods = map[transaction.MethodName]struct{}{
 		MethodMeta: {},
 	}
+
+	// KeyFormat is the namespace for the consensus key formats.
```
kostko, line 70:
```suggestion
	// KeyFormat is the namespace for the consensus state key formats.
```

kostko (commented):
Looks good now, just two minor comment fixes.

kostko (approved):

PR #5550, opened 2024-02-06
docs: Fix broken link to CometBFT metrics page

matevz:


peternose (approved):

peternose:
Should we fix also `https://docs.cometbft.com/main/cometbft-core/using-cometbft.html#transactions`?

matevz:
> Should we fix also `https://docs.cometbft.com/main/cometbft-core/using-cometbft.html#transactions`?

Nice catch, fixed.

PR #5551, opened 2024-02-07
go/keymanager/churp: Add create and update methods

peternose:


go/consensus/cometbft/apps/registry/query.go
```diff
@@ -81,13 +82,35 @@ func (rq *registryQuerier) NodeStatus(ctx context.Context, id signature.PublicKe
...
 		return nil, fmt.Errorf("failed to get epoch: %w", err)
 	}
 
-	nodes, err := rq.state.Nodes(ctx)
+	nodes, err := rq.state.Nodes(ctx, ids)
+	if err != nil {
+		return nil, err
+	}
+
+	// Filter out expired nodes.
```
kostko, line 96: Could move this into a helper function and reuse.

go/consensus/cometbft/apps/keymanager/churp/txs.go
```diff
@@ -0,0 +1,253 @@
...
+	"fmt"
+
+	beacon "github.com/oasisprotocol/oasis-core/go/beacon/api"
+	tmapi "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/api"
+	churpState "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/apps/keymanager/churp/state"
+	"github.com/oasisprotocol/oasis-core/go/consensus/cometbft/apps/keymanager/common"
+	"github.com/oasisprotocol/oasis-core/go/keymanager/churp"
+)
+
+func (ext *churpExt) create(ctx *tmapi.Context, req *churp.CreateRequest) error {
```
kostko, line 13: Should creating a new churp instance require additional stake? As it seems to increase periodic processing complexity. Or is there a reasonable limit per runtime?

go/registry/api/grpc.go
```diff
@@ -26,7 +26,9 @@ var (
 	// methodGetNodeStatus is the GetNodeStatus method.
 	methodGetNodeStatus = serviceName.NewMethod("GetNodeStatus", IDQuery{})
 	// methodGetNodes is the GetNodes method.
-	methodGetNodes = serviceName.NewMethod("GetNodes", int64(0))
+	methodGetNodes = serviceName.NewMethod("GetNodes", IDsQuery{})
```
kostko, line 29: Uh, this can cause some weird breakage.

peternose: Wasn't sure which components and how many use this endpoint. I thought that renaming was a better solution than using name `NodesByID`.

go/consensus/cometbft/apps/keymanager/churp/txs.go
peternose: Currently no limit (except `uint8`). Larger stake makes sense, but I guess I need to add an option to delete/disable an instance then.

go/consensus/cometbft/apps/keymanager/churp/txs.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	ctx.EmitEvent(tmapi.NewEventBuilder(ext.appName).TypedAttribute(&churp.UpdateEvent{
+		Status: status,
+	}))
+
+	return nil
+}
+
+func (ext *churpExt) onEpochChange(ctx *tmapi.Context, epoch beacon.EpochTime) error {
```
kostko, line 193: This probably shouldn't be in `txs.go`, maybe a separate `epoch.go` or similar.

go/consensus/cometbft/apps/keymanager/churp/txs.go
```diff
@@ -0,0 +1,253 @@
...
+	state := churpState.NewMutableState(ctx.State())
+
+	statuses, err := state.Statuses(ctx)
```
kostko, line 196: This needs to somehow at least skip suspended runtimes, otherwise this list can only grow (even if additional stake requirements are not met and runtime is suspended) and iteration will be slower and slower?

go/consensus/cometbft/apps/keymanager/churp/txs.go
peternose: I will implement this latter, as I need to first include churp statuses in the genesis file and update method for calculating stake from it.

kostko (approved):

go/registry/api/grpc.go
ptrus: The grpc API changes should probably have a breaking changelog fragment since external services use it to communicate with the node (e.g. oasis-cli, oasis-wallet, rosetta-gateway...), there also is a public deployment of it.

kostko: I would try to avoid breakage if possible.

peternose: Removed commit that renamed grpc methods and changed `fetchKeys` (in commit 
_go/keymanager: Move helper functions to a separate package_) to query nodes one by one like we had before.

PR #5554, opened 2024-02-12
keymanager/runtime: Use insecure RPC requests for ephemeral public keys

peternose:
Removing support for outdated key manager clients, which fetched public key via noise sessions.

### DIFF
--- a/.changelog/5554.internal.md
+++ b/.changelog/5554.internal.md
@@ -1,0 +1,1 @@
+keymanager/runtime: Use insecure RPC requests for ephemeral public keys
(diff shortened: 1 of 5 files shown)


kostko (approved):
Looks good, will the second commit be merged as well or is it just temporary?

peternose:
> Looks good, will the second commit be merged as well or is it just temporary?

Temporary. The storage-early-state-sync test is flaky and I couldn't produce the problem locally. So I borrowed this PR. The problem is that on startup compute node 1 tries to fetch light block from compute node 0, fails and stops. I think it doesn't try to fetch it from validators 0-2 as it is still trying to connect to them.

PR #5556, opened 2024-02-14
go: Bump go to 1.22.0

ptrus:
Also bumps:
- golangci-lint to 1.56.1
- gofumpt to 0.6.0
- goimports to 0.18.0

The second commit addresses all unused-parameter lint errors identified by the updated golangci-lint. I considered disabling this lint rule but kept it since it might detect cases where parameter is not used by mistake (there was one such benign case in tests).


kostko (approved):

peternose:
> In Go 1.22, each iteration of the loop creates new variables, to avoid accidental sharing bugs. 

😍

kostko:
Indeed ;-)

PR #5562, opened 2024-02-20
secret-sharing/churp: Implement the dealer

peternose:


secret-sharing/src/vss/polynomial.rs
```diff
@@ -0,0 +1,226 @@
...
+    /// Creates a bivariate polynomial with random coefficients.
+    pub fn random(deg_x: u8, deg_y: u8) -> Self {
+        let deg_x = deg_x as usize;
+        let deg_y = deg_y as usize;
+
+        let mut b = Vec::with_capacity(deg_x + 1);
+        for _ in 0..b.capacity() {
+            let mut bi = Vec::with_capacity(deg_y + 1);
+            for _ in 0..bi.capacity() {
+                let bij = Fp::random(rand_core::OsRng);
```
kostko, line 44: Would be better if the RNG was passed in as a parameter.

secret-sharing/src/vss/matrix.rs
```diff
@@ -0,0 +1,184 @@
...
+            for mij in mi {
+                bytes.extend_from_slice(mij.to_bytes().as_ref());
+            }
+        }
+
+        bytes
+    }
+
+    /// Attempts to create a verification matrix from its byte representation.
+    pub fn from_bytes(bytes: Vec<u8>) -> Option<Self> {
```
kostko, line 79: Would be good to have some fuzzing for the (de)serialization functions. For an example see the runtime package which uses hongfuzz.

peternose: Added. Not sure how sdk runs `cargo hfuzz run ...` :shrug:

kostko: One needs to run it manually currently.

secret-sharing/src/vss/fuzz/main.rs
```diff
@@ -0,0 +1,89 @@
...
+
+fn main() {
+    loop {
+        fuzz!(|data: &[u8]| {
+            fuzz_bivariate_polynomial_random(data);
+            fuzz_bivariate_polynomial_from_seed(data);
+            fuzz_bivariate_polynomial_from_bytes(data);
+
+            fuzz_verification_matrix_random(data);
+            fuzz_verification_matrix_from_seed(data);
```
kostko, line 15: This probably makes it a bit awkward for the fuzzer as the same data is used for different code paths. Ideally you would have one case for each of these or split the data up.

kostko (approved):

PR #5566, opened 2024-02-23
governance: Remove unchanged proposal parameters

amela:
Previously, when creating a governance proposal, the writer would output unchanged parameters and show them as nil value.

### DIFF
--- a/.changelog/5566.bugfix.md
+++ b/.changelog/5566.bugfix.md
@@ -1,0 +1,4 @@
+governance/ChangeParametersProposal: Display only changed parameters
+
+Omit the unchanged parameter values when pretty-printing the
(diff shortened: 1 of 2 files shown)


.changelog/5566.bugfix.md
```diff
@@ -0,0 +1,4 @@
+governance: Remove unchanged proposal parameters
```
ptrus, line 1:
```suggestion
governance/ChangeParametersProposal: Display only changed parameters

Omit the unchanged parameter values when pretty-printing the `ChangeParameterProposal`.
``` 

Maybe something like this, to make it clearer that this change only affects the pretty-printing of the proposals.

ptrus (approved):

PR #5569, opened 2024-02-24
keymanager/src/churp: Handle init RPC requests

peternose:


keymanager/src/churp/handler.rs
```diff
@@ -0,0 +1,263 @@
...
+};
+
+/// A handoff interval that disables handoffs.
+pub const HANDOFFS_DISABLED: EpochTime = 0xffffffffffffffff;
+
+/// Context used for the application request signature.
+const APPLICATION_REQUEST_SIGNATURE_CONTEXT: &[u8] =
+    b"oasis-core/keymanager/churp: application request";
+
+const CHECKSUM_VERIFICATION_MATRIX_CUSTOM: &[u8] =
```
kostko, line 47: Add documentation comment.

keymanager/src/churp/handler.rs
```diff
@@ -0,0 +1,263 @@
...
+                    .store_bivariate_polynomial(polynomial, churp_id, round)?;
+
+                dealer
+
+                // TODO: Delete previous polynomials. How? Should we call
+                //       the host to clean storage?
+            }
+        };
+
+        // Keep the most recent dealers in the memory.
```
kostko, line 241:
```suggestion
        // Keep the most recent dealers in memory.
```

runtime/src/consensus/state/keymanager/churp.rs
```diff
@@ -0,0 +1,330 @@
...
+        let h = Hash::digest_bytes(runtime_id.as_ref());
+        match self.mkvs.get(&StatusKeyFmt((h, churp_id)).encode()) {
+            Ok(Some(b)) => Ok(Some(self.decode_status(&b)?)),
+            Ok(None) => Ok(None),
+            Err(err) => Err(StateError::Unavailable(anyhow!(err))),
+        }
+    }
+
+    fn decode_status(&self, data: &[u8]) -> Result<Status, StateError> {
+        print!("{:?}", data);
```
kostko, line 60: Leftover?

keymanager/src/churp/handler.rs
```diff
@@ -0,0 +1,263 @@
...
+                };
+
+                // Encrypt and store the polynomial in case of a restart.
+                let polynomial = dealer.bivariate_polynomial();
+                self.storage
+                    .store_bivariate_polynomial(polynomial, churp_id, round)?;
+
+                dealer
+
+                // TODO: Delete previous polynomials. How? Should we call
```
kostko, line 236: Currently there is no API for removals. We could add one if needed.

keymanager/src/churp/handler.rs
```diff
@@ -0,0 +1,263 @@
...
+        let beacon_verifier = BeaconState::new(consensus_verifier.clone());
+
+        let dealers = RwLock::new(LruCache::new(
+            NonZeroUsize::new(DEALERS_CACHE_SIZE).unwrap(),
+        ));
+
+        Self {
+            signer,
+            runtime_id,
+            node_id: Default::default(),
```
kostko, line 89: Will this be filled in later? It can be fetched from the host.

keymanager/src/churp/types.rs
```diff
@@ -0,0 +1,49 @@
+//! CHURP types used by the worker-host protocol.
+use oasis_core_runtime::common::{crypto::signature::Signature, namespace::Namespace};
+
+/// Signed public key.
+#[derive(Clone, Debug, Default, PartialEq, Eq, cbor::Encode, cbor::Decode)]
+pub struct InitRequest {
+    /// A unique identifier within the key manager runtime.
+    #[cbor(optional)]
```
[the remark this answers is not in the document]
peternose: I think that everything that has default/zero value could be optional, to save space. I can change that, though.
The only thing that I see as a problem here is that currently rust version of cbor doesn't ignore empty optional structs when encoding, which can cause problems with signatures as the go side will encode the data differently.

keymanager/src/churp/handler.rs
peternose: Didn't know that, actually. Will change this so that we store only 2 values, one for the current round and one for the next one.

keymanager/src/churp/handler.rs
peternose: Yes, that was the plan (but I forgot about it). Will add another commit.

keymanager/src/churp/storage.rs
```diff
@@ -0,0 +1,245 @@
...
+        sgx::seal::new_deoxysii,
+    },
+    storage::KeyValue,
+};
+
+use super::Error;
+
+const BIVARIATE_POLYNOMIAL_SEAL_CONTEXT: &[u8] =
+    b"oasis-core/keymanager/churp: bivariate polynomial";
+const BIVARIATE_POLYNOMIAL_STORAGE_KEY_PREFIX: &[u8] = b"keymanager/churp: bivariate polynomial";
```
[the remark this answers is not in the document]
peternose: Key manager uses `_`, e.g. `keymanager_master_secret`. Will change to that.

keymanager/src/churp/handler.rs
peternose: Actually, one is enough. Will change latter if needed.

keymanager/src/churp/types.rs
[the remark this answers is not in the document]
peternose: I was wrong. We probably gain nothing with this, plus, structures that are signed should have no optional fields.

keymanager/src/churp/handler.rs
kostko: In that case, just remove the comment.

PR #5570, opened 2024-02-26
go/keymanager/churp: Add a flag which enables the extension

peternose:


go/consensus/cometbft/apps/keymanager/churp/ext.go
```diff
@@ -79,6 +88,16 @@ func (*churpExt) EndBlock(*tmapi.Context) error {
...
 }
 
+// EndBlock implements api.Extension.
 func (ext *churpExt) InitChain(*tmapi.Context, types.RequestInitChain, *genesis.Document) error {
 	return nil
 }
+
+func (*churpExt) enabled(ctx *tmapi.Context) (bool, error) {
+	regState := registryState.NewMutableState(ctx.State())
+	regParams, err := regState.ConsensusParameters(ctx)
```
kostko, line 98: Why is this not in the key manager parameters? I guess because now all of the parameters are only for the secrets part and there are no global parameters for the key manager backend?

go/registry/api/api.go
```diff
@@ -1403,6 +1403,9 @@ type ConsensusParameters struct {
 	// disabled outside of the genesis block.
 	DisableKeyManagerRuntimeRegistration bool `json:"disable_km_runtime_registration,omitempty"`
 
+	// EnableKeyManagerCHURP is true iff the CHURP extension for the key manager is enabled.
+	EnableKeyManagerCHURP bool `json:"enable_km_churp,omitempty"`
```
kostko, line 1407: Should this rather be something like a list of enabled KM extension identifiers (`EnableKeyManagerExtensions`)? To not introduce one flag per extension.

peternose: Could do that. But my plan was to remove this after the upgrade.

go/consensus/cometbft/apps/keymanager/churp/ext.go
peternose: Yes, currently there are no global parameters and in the registry we already have `DisableKeyManagerRuntimeRegistration`. Again, this could be removed after the upgrade.

go/consensus/cometbft/apps/keymanager/churp/ext.go
```diff
@@ -79,6 +88,16 @@ func (*churpExt) EndBlock(*tmapi.Context) error {
 	return nil
 }
 
+// EndBlock implements api.Extension.
```
kostko, line 91:
```suggestion
// InitChain implements api.Extension.
```

go/oasis-node/cmd/genesis/genesis.go
```diff
@@ -376,6 +377,10 @@ func AppendRegistryState(doc *genesis.Document, entities, runtimes, nodes []stri
 		regSt.Parameters.EnableRuntimeGovernanceModels[gm] = true
 	}
 
+	if viper.GetBool(CfgRegistryTEEFeaturesSGXSignedAttestations) {
```
kostko, line 380: Wrong cfg?

PR #5571, opened 2024-02-26
go/upgrade/migrations: Prepare handler for version 24.0.0

peternose:


kostko (commented):
Can we add an E2E test for the migration and also remove the `consensus_v62` and `consensus_max_allowances` migrations while we're here?

Looks good otherwise.

peternose:
> Can we add an E2E test for the migration and also remove the `consensus_v62` and `consensus_max_allowances` migrations while we're here?

Tests removed and a new one added. Also, `ConsenusParameters` method was added to gRPC, and by default, the CHURP extension was disabled in tests and has to be enabled manually.

kostko (approved):

PR #5572, opened 2024-02-26
go/worker/keymanager: Add churp worker

peternose:


go/worker/keymanager/churp.go
```diff
@@ -0,0 +1,471 @@
...
+	if status.RuntimeID != w.kmWorker.runtimeID {
+		return
+	}
+	if _, ok := w.churps[status.ID]; !ok {
+		return
+	}
+	w.churps[status.ID] = status
+
+	w.logger.Debug("handle status update",
+		"status", fmt.Sprintf("%+v", status),
```
kostko, line 180: Does simply passing it as `status` not JSON-encode the output when using JSON logs?

go/worker/keymanager/churp.go
```diff
@@ -0,0 +1,471 @@
...
+			churpID: status.ID,
+			round:   status.Round,
+			epoch:   epoch,
+			height:  height,
+			index:   -1,
+			cancel:  nil,
+		}
+
+		s.submissions[status.ID] = info
+		s.queue.Push(info)
```
kostko, line 284: How does this insert into the right place given that queue should be sorted by height? Shouldn't this use `heap.Push`?

go/worker/keymanager/churp.go
```diff
@@ -0,0 +1,471 @@
...
+	}
+
+	// Stop and remove the submission if the application has already
```
kostko, line 287: Why are we adding first and then checking (and maybe immediately removing) instead of not adding at all? I guess this would only rarely happen.

go/worker/keymanager/churp.go
peternose: It does.

go/worker/keymanager/churp.go
peternose: Yes, fixed also for `Pop`.

go/worker/keymanager/churp.go
peternose: The code was shorter and more transparent this way. I can change.

kostko: No it's fine.

go/worker/keymanager/churp.go
```diff
@@ -0,0 +1,471 @@
...
+	q[i], q[j] = q[j], q[i]
+}
+
+// Push appends an element to the heap.
+func (q *SubmissionQueue) Push(x any) {
+	x.(*submissionInfo).index = len(*q)
+	*q = append(*q, x.(*submissionInfo))
+}
+
+// Pop removes the smallest element from the heap.
```
kostko, line 452: I guess this comment should be changed?

kostko (approved):

PR #5578, opened 2024-02-28
sgx: Support early updates for ECDSA TCB infos

kostko:


go/common/sgx/pcs/http.go
```diff
@@ -120,6 +119,9 @@ func (hc *httpClient) GetTCBBundle(ctx context.Context, fmspc []byte) (*TCBBundl
 
 	// Then fetch QE identity.
 	u = hc.getUrl(pcsAPIGetQEIdentityPath)
+	q = u.Query()
+	q.Set("update", string(update))
```
peternose, line 123: I'm still not a fan of introducing a new type for everything, because than you have to cast like here.

go/runtime/host/sgx/ecdsa.go
```diff
@@ -184,7 +187,13 @@ func (ec *teeStateECDSA) Update(ctx context.Context, sp *sgxProvisioner, conn pr
 		}
 		ec.tcbCache.cache(fresh, pckInfo.FMSPC)
 		return fresh, nil
-	}()
+	}
+	var tcbBundle *pcs.TCBBundle
+	for _, update := range []pcs.UpdateType{pcs.UpdateStandard, pcs.UpdateEarly} {
+		if tcbBundle, err = getTcbBundle(update); err == nil {
```
peternose, line 193: I would log error here, to log also errors from line 174. Or maybe warn standard and error early.

go/runtime/host/sgx/ecdsa.go
```diff
@@ -144,14 +144,15 @@ func (ec *teeStateECDSA) Update(ctx context.Context, sp *sgxProvisioner, conn pr
 	// also do their own verification).
 	// Check bundles in order: fresh first, then cached, then try downloading again if there was
 	// no scheduled refresh this time.
-	tcbBundle, err := func() (*pcs.TCBBundle, error) {
+	getTcbBundle := func(update pcs.UpdateType) (*pcs.TCBBundle, error) {
 		var fresh *pcs.TCBBundle
 
 		cached, refresh := ec.tcbCache.check(pckInfo.FMSPC)
```
peternose, line 150: Does the cache need to distinguish between early and standard (e.g. to overwrite early when standard is available)?

kostko: No, it doesn't matter which TCB info is used as long as it passes verification against the quote policy. Over time standard becomes the same as early anyway.

peternose (approved):

PR #5583, opened 2024-03-05
Bump Rust toolchain to 2024-03-04, update dependencies

kostko:


.changelog/5583.internal.md
```diff
@@ -0,0 +1 @@
+Bump Rust toolchain to 2024-03-04
```
peternose, line 1: We could make this one trivial, as now we will have duplicate change logs for this bump.

peternose (approved):

.changelog/5583.internal.md
kostko: No, because I copied the new fragment over the old one there will be one entry which will link to both PRs.

PR #5589, opened 2024-03-07
go/runtime/client: Add the GetUnconfirmedTransactions method

abukosek:
Similarly to `GetUnconfirmedTransactions` in the consensus API, this new method returns the currently pending runtime transactions from the runtime transaction pool.

TODO:
- [ ] tests


go/worker/client/service.go
```diff
@@ -238,6 +239,35 @@ func (s *service) GetTransactionsWithResults(ctx context.Context, request *api.G
...
 
+// Implements api.RuntimeClient.
+func (s *service) GetUnconfirmedTransactions(_ context.Context, runtimeID common.Namespace) ([][]byte, error) {
+	rt := s.w.commonWorker.GetRuntime(runtimeID)
+	if rt == nil {
+		return nil, api.ErrNotFound
+	}
+
+	// Get current proposed transactions from the runtime's transaction pool.
+	proposedTxs := rt.TxPool.GetProposedTxs()
```
kostko, line 250: Hm, this doesn't seem right, so proposed transactions are just those that have been proposed for the current batch (e.g. on executor nodes). You should instead be getting the pending transactions from all of the different queues, similar to what `GetSchedulingSuggestion` is doing (but without holding the lock).

kostko (changes requested):

go/worker/client/service.go
abukosek: OK, I'll fix it. That part of the code is very confusing.

abukosek: Please re-review :)

go/runtime/txpool/txpool.go
```diff
@@ -126,6 +126,9 @@ type TransactionPool interface {
 
 	// PendingCheckSize returns the number of transactions currently pending to be checked.
 	PendingCheckSize() int
+
+	// GetPendingTxs returns the currently pending transactions.
```
kostko, line 130: I think it would be more clear to just say that it returns all of transactions currently queued in the transaction pool. And maybe call the method `GetTxs()`.

go/runtime/txpool/local_queue.go
```diff
@@ -69,6 +69,18 @@ func (lq *localQueue) HandleTxsUsed(hashes []hash.Hash) {
...
 }
 
+func (lq *localQueue) PeekAll() []*TxQueueMeta {
+	lq.l.Lock()
+	defer lq.l.Unlock()
+
+	allTxs := lq.txs
+	var txs []*TxQueueMeta
+	for _, tx := range allTxs {
+		txs = append(txs, &TxQueueMeta{raw: tx.Raw(), hash: tx.Hash(), firstSeen: tx.FirstSeen()})
```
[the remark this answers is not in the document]
abukosek: I thought it would be better to clone them, since they get returned to code outside of the txpool package and I don't know how the existing txpool code manipulates the structures, so returning pointers doesn't seem safe to me.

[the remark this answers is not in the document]
kostko: All other methods return existing pointers and I don't see any places that would mutate them. They also don't contain any fields that could reasonably change (data, hash and first seen time).

go/worker/client/service.go
```diff
@@ -238,6 +239,31 @@ func (s *service) GetTransactionsWithResults(ctx context.Context, request *api.G
...
+
+	// Get currently pending transactions from the runtime's transaction pool.
+	pendingTxs := rt.TxPool.GetTxs()
+
+	// Sort transactions by first seen time (ascending).
+	sort.SliceStable(pendingTxs, func(i, j int) bool {
+		a := pendingTxs[i]
+		b := pendingTxs[j]
+		return a.FirstSeen().Before(b.FirstSeen())
+	})
```
kostko, line 257: I wouldn't change the sort order as these should be sorted by priority already.

abukosek: Oh, OK, I'll remove this.

go/runtime/txpool/local_queue.go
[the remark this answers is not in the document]
abukosek: OK, I've changed it to just copy the pointers into a new slice :)

go/runtime/txpool/local_queue.go
```diff
@@ -69,6 +69,17 @@ func (lq *localQueue) HandleTxsUsed(hashes []hash.Hash) {
 	lq.txs = keptTxs
 }
 
+func (lq *localQueue) PeekAll() []*TxQueueMeta {
+	lq.l.Lock()
+	defer lq.l.Unlock()
+
+	var txs []*TxQueueMeta
```
kostko, line 76: Preallocate the slice.

go/runtime/txpool/main_queue.go
```diff
@@ -114,6 +114,15 @@ func (mq *mainQueue) GetSchedulingExtra(offset *hash.Hash, limit uint32) []*TxQu
 	return txs
 }
 
+func (mq *mainQueue) PeekAll() []*TxQueueMeta {
+	allTxs := mq.inner.getAll()
+	var txs []*TxQueueMeta
```
kostko, line 119: Preallocate the slice if you can get the size (also in `TakeAll`?).

abukosek: Done, I've also modified other `mainQueue` methods to preallocate their output slices.

go/runtime/txpool/queues.go
```diff
@@ -50,6 +50,8 @@ type UsableTransactionSource interface {
 	// HandleTxsUsed is a callback to indicate that the scheduler is done with a set of txs, by hash. For most
 	// implementations, remove it from internal storage.
 	HandleTxsUsed(hashes []hash.Hash)
+	// PeekAll returns all transactions without removing them.
+	PeekAll() []*TxQueueMeta
```
peternose, line 54: Why don't you name this `GetTxs` and move it after/before `GetTxByHash`?

go/runtime/txpool/rim_queue.go
```diff
@@ -36,6 +36,17 @@ func (rq *rimQueue) HandleTxsUsed([]hash.Hash) {
 	// The roothash module manages the incoming message queue on its own, so we don't do anything here.
 }
 
+func (rq *rimQueue) PeekAll() []*TxQueueMeta {
+	rq.l.RLock()
+	defer rq.l.RUnlock()
+
+	var txs []*TxQueueMeta
```
peternose, line 43: Preallocate

go/runtime/txpool/main_queue.go
```diff
@@ -107,17 +107,26 @@ func (mq *mainQueue) HandleTxsUsed(hashes []hash.Hash) {
...
 	txMetas := mq.inner.getPrioritizedBatch(offset, limit)
-	var txs []*TxQueueMeta
+	txs := make([]*TxQueueMeta, 0, len(txMetas))
 	for _, txMeta := range txMetas {
 		txs = append(txs, &txMeta.TxQueueMeta) //nolint:gosec
 	}
 	return txs
 }
 
+func (mq *mainQueue) PeekAll() []*TxQueueMeta {
```
peternose, line 117:
```go
func (mq *mainQueue) GetTxsToPublish() []*TxQueueMeta {
	return mq.PeekAll()
}
```
To avoid code duplication. Maybe we can even get rid of `GetTxsToPublish`, did not check that.

go/runtime/txpool/main_queue.go
```diff
@@ -107,17 +107,26 @@ func (mq *mainQueue) HandleTxsUsed(hashes []hash.Hash) {
...
+	allTxs := mq.inner.getAll()
```
peternose, line 118: Maybe we could add another method to return `TxQueueMeta` or refactor `getAll()`, as it looks like quite a few methods only use `TxQueueMeta` .

go/runtime/txpool/local_queue.go
```diff
@@ -69,6 +69,12 @@ func (lq *localQueue) HandleTxsUsed(hashes []hash.Hash) {
 	lq.txs = keptTxs
 }
 
+func (lq *localQueue) PeekAll() []*TxQueueMeta {
+	lq.l.Lock()
+	defer lq.l.Unlock()
+	return append(make([]*TxQueueMeta, 0, len(lq.txs)), lq.txs...)
```
peternose, line 75: I prefer 3 lines. Should we preallocate also in `GetTxsToPublish`?

go/runtime/client/tests/tester.go
```diff
@@ -229,6 +229,11 @@ func testQuery(
 		Data:      []byte("test checktx request"),
 	})
 	require.NoError(t, err, "CheckTx")
+
+	// Get the number of unconfirmed transactions.
+	utxs, err := c.GetUnconfirmedTransactions(ctx, runtimeID)
+	require.NoError(t, err, "GetUnconfirmedTransactions")
+	require.True(t, len(utxs) == 0)
```
peternose, line 236: Is there a test where this is > 0?

go/runtime/txpool/queues.go
abukosek: Because the operation is basically `TakeAll` without removing anything from the queues :D
`Get` is too vague -- does it remove txs from the queue or not? :)

go/runtime/txpool/main_queue.go
abukosek: That seems like a job for another PR :)

go/runtime/txpool/local_queue.go
abukosek: I put this on a single line to better fit with the rest of the coding style in the txpool package (a few other existing methods also do this).

go/runtime/client/tests/tester.go
abukosek: No. But the consensus `GetUnconfirmedTransactions` test is similar :P

I did initially want to make this nicer, but the mock runtime host that's used in unit tests sucks and doing this in the e2e tests at exactly the right time to make it deterministic is super-hard. So I think like with the consensus method that does the same, the test for this should be in the block explorer or wherever they plan on using this.

go/runtime/txpool/queues.go
peternose: `GetTxsToPublish`, `GetTxByHash` and `GetTxs` (in txpool.go) also do not remove txs, so maybe its not that vague :)

peternose (approved):

kostko (approved):

PR #5592, opened 2024-03-09
secret-sharing/src/vss: Add Lagrange interpolation

peternose:
I guess the multiplier could be optimized if we didn't clone so much.

### DIFF
--- a/.changelog/5592.feature.md
+++ b/.changelog/5592.feature.md
@@ -1,0 +1,1 @@
+secret-sharing/src/vss: Add Lagrange interpolation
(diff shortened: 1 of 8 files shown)


kostko (commented):
Would be good to see some quick benchmarks.

peternose:
> Would be good to see some quick benchmarks.

I think that the speed is actually not that important, and we can deal with it latter, because:
- handoffs (which use interpolation to reconstruct share) are done in the background,
- key reconstruction (which use interpolation to reconstruct key from shared received from key manager nodes) in the key manager client should probably (partially) precompute Lagrange coefficients as the committee `xs` is known in advance.

> The multiplier is doing quite some allocation but this could probably be improved further later.

I know. Agreed.

kostko (approved):

PR #5605, opened 2024-03-20
go/consensus/cometbft/beacon: Fix GetEpochBlock implementation

kostko:


go/consensus/cometbft/beacon/beacon.go
```diff
@@ -101,47 +106,66 @@ func (sc *serviceClient) GetFutureEpoch(ctx context.Context, height int64) (*bea
...
+	if cachedHeight, ok := sc.epochCache.Get(epoch); ok {
+		return cachedHeight.(int64), nil
+	}
+
+	lowHeight, err := sc.backend.GetLastRetainedVersion(ctx)
+	if err != nil {
+		return -1, fmt.Errorf("failed to query last retained version: %w", err)
+	}
+
+	blk, err := sc.backend.GetCometBFTBlock(ctx, consensus.HeightLatest)
```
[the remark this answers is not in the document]
kostko: The `consensus.HeightLatest` is a special marker value, not an actual height. So we need to always first resolve the height to an actual height.

go/consensus/cometbft/beacon/beacon.go
```diff
@@ -101,47 +106,66 @@ func (sc *serviceClient) GetFutureEpoch(ctx context.Context, height int64) (*bea
...
+	height := hiHeight
+
 	// Find historic epoch.
-	//
-	// TODO: This is really really inefficient, and should be optimized,
-	// maybe a cache of the last few epochs, or a binary search.
-	height := consensus.HeightLatest
-	for {
+	const maxIterations = 20 // Should be good enough for most use cases.
+	for range maxIterations {
```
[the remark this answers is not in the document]
kostko: Sure, I thought of that as well, but even doing a few steps is much better than the previous O(n) implementation. If we see this is still a problem we can optimize further.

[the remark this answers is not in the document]
peternose: Since bisection is done on `int64` (max 64 steps), this limit could also be removed. But then the stop condition should be improved (one should check if the low/high epoch hasn't changed). Currently, if we use a very high value for epoch (value greater than the last epoch), we will do 20 steps always, and always querying the same height .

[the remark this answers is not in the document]
peternose: > I thought of that as well, but even doing a few steps is much better than the previous O(n) implementation

Agree, unless these few steps need to query the database or fetch blocks from other nodes. In that case, every removed step is a major improvement. Not sure if this is the case here.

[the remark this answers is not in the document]
kostko: This is querying the local database, so it should be pretty fast for a few queries. Previously the problem was if one was querying an epoch very far in the past (there is a use case for this in the CLI) this did a lot of queries.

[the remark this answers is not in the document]
kostko: Added an early stopping condition.

go/consensus/cometbft/beacon/beacon.go
```diff
@@ -101,47 +106,69 @@ func (sc *serviceClient) GetFutureEpoch(ctx context.Context, height int64) (*bea
...
 		}
 
-		if epoch == pastEpoch {
-			return height, nil
+		switch {
+		case epoch == pastEpoch:
+			_ = sc.epochCache.Put(pastEpoch, epochHeight)
+			return epochHeight, nil
+		case epoch < pastEpoch:
+			hiHeight = epochHeight // epochHeight <= height
```
peternose, line 159: These two comments are confusing, as `epochHeight` is always smaller or equal to `height`.

go/consensus/cometbft/beacon/beacon.go
```diff
@@ -101,47 +106,69 @@ func (sc *serviceClient) GetFutureEpoch(ctx context.Context, height int64) (*bea
...
 
-		height--
-
-		// The initial height can be > 1, but presumably this does not
-		// matter, since we validate that epoch > sc.baseEpoch.
-		if pastEpoch == 0 || height <= 1 {
-			return -1, fmt.Errorf("failed to find historic epoch (minimum: %d requested: %d)", pastEpoch, epoch)
+		// Determine next height as the midpoint between the two.
+		nextHeight := (lowHeight + hiHeight) / 2
+		if nextHeight == height {
```
peternose, line 166: Just adding a note, that the right approach would compare epochs and not heights, as here we could be doing `log(epoch_block_size)` steps in vain.

peternose (approved):

go/consensus/cometbft/beacon/beacon.go
kostko: That is exactly what the comment says :P that if the epoch is lower, the smaller should be chosen to potentially skip more steps and `epochHeight` is always smaller. I'll amend the comment to say this in words.

peternose: Ups, I see they are the same :) I would just remove them.

go/consensus/cometbft/beacon/beacon.go
kostko: True, let me fix that real quick.

peternose (approved):

go/consensus/cometbft/beacon/beacon.go
```diff
@@ -101,47 +106,70 @@ func (sc *serviceClient) GetFutureEpoch(ctx context.Context, height int64) (*bea
...
+	}
+
+	lowHeight, err := sc.backend.GetLastRetainedVersion(ctx)
+	if err != nil {
+		return 0, fmt.Errorf("failed to query last retained version: %w", err)
+	}
+
+	blk, err := sc.backend.GetCometBFTBlock(ctx, consensus.HeightLatest)
+	if err != nil {
+		return -1, err
```
ptrus, line 130: I other places 0 is returned.

kostko: 🤦

PR #5606, opened 2024-03-21
Add observer nodes automatically to ACL

jberci:


go/worker/keymanager/watcher.go
```diff
@@ -284,6 +284,32 @@ func (w *rtNodeWatcher) watch(ctx context.Context) {
 	}
 	defer sub.Close()
 
+	// Maintain a list of allowable nodes in addition to the committee members.
+	allowableNodes := map[signature.PublicKey]struct{}{}
```
kostko, line 288: Maybe rename this to `observerNodes` to be more explicit.

go/worker/keymanager/watcher.go
```diff
@@ -321,6 +352,14 @@ func (w *rtNodeWatcher) watch(ctx context.Context) {
 				// nodes have been set (even if the new set is empty).
 				continue
 			}
+		case ne := <-nodeCh:
+			if ne.IsRegistration {
+				if ne.Node.HasRoles(node.RoleObserver) && ne.Node.HasRuntime(w.runtimeID) {
+					allowableNodes[ne.Node.ID] = struct{}{}
+				}
```
kostko, line 359: Node should probably be deleted in case the roles/runtimes change in a way that the above condition no longer holds.

go/worker/keymanager/watcher.go
```diff
@@ -293,6 +319,11 @@ func (w *rtNodeWatcher) watch(ctx context.Context) {
 				watcher.Reset()
 				defer watcher.Freeze(0)
 
+				for sig := range observerNodes {
+					_, _ = watcher.WatchNode(ctx, sig)
```
peternose, line 323: I understand that we want to allow executor committee members at the start of every epoch, but I guess the observer nodes should be allowed immediately?

Why `sig` and not `pk`, `id` or `nodeID`?

go/worker/keymanager/watcher.go
```diff
@@ -321,6 +352,12 @@ func (w *rtNodeWatcher) watch(ctx context.Context) {
 				// nodes have been set (even if the new set is empty).
 				continue
 			}
+		case ne := <-nodeCh:
+			if ne.IsRegistration && ne.Node.HasRoles(node.RoleObserver) && ne.Node.HasRuntime(w.runtimeID) {
+				observerNodes[ne.Node.ID] = struct{}{}
+			} else {
+				delete(observerNodes, ne.Node.ID)
+			}
```
peternose, line 360: Should we add `continue` to not update the access list every time?

go/worker/keymanager/watcher.go
```diff
@@ -284,6 +284,32 @@ func (w *rtNodeWatcher) watch(ctx context.Context) {
 	}
 	defer sub.Close()
 
+	// Maintain a list of allowable nodes in addition to the committee members.
+	observerNodes := map[signature.PublicKey]struct{}{}
+
+	nodeCh, nodeSub, err := w.consensus.Registry().WatchNodes(ctx)
```
peternose, line 290: I guess we could remove `watcher.WatchNodeUpdates()` as now we have to watch all nodes anyway. But I'm not sure :thinking:

kostko: Yeah, one would just need to introduce a separate `computeNodes` map that is updated during epoch transitions and then refresh it below in `nodeCh` handler.

Because that is basically what the "watcher" is doing anyway and I think that would actually be cleaner.

go/worker/keymanager/watcher.go
```diff
@@ -303,26 +325,39 @@ func (w *rtNodeWatcher) watch(ctx context.Context) {
...
+						computeNodes[nd.ID] = nd
 					}
 				}
 			}()
-		case nu := <-ch:
-			if nu.Reset {
-				// Ignore reset events to avoid clearing the access list before setting a new one.
-				// This is safe because a reset event is always followed by a freeze event after the
-				// nodes have been set (even if the new set is empty).
+		case ne := <-nodeCh:
```
kostko, line 351: This should also now update `computeNodes` on changes.

go/worker/keymanager/watcher.go
```diff
@@ -303,26 +325,48 @@ func (w *rtNodeWatcher) watch(ctx context.Context) {
...
-				// Ignore reset events to avoid clearing the access list before setting a new one.
-				// This is safe because a reset event is always followed by a freeze event after the
-				// nodes have been set (even if the new set is empty).
-				continue
+		case ne := <-nodeCh:
+			if !ne.IsRegistration {
+				if _, known := observerNodes[ne.Node.ID]; known {
+					delete(observerNodes, ne.Node.ID)
+				} else if _, known := computeNodes[ne.Node.ID]; known {
+					delete(computeNodes, ne.Node.ID)
```
kostko, line 356: This probably cannot happen as a node cannot disappear mid-epoch.

go/worker/keymanager/watcher.go
```diff
@@ -303,26 +325,48 @@ func (w *rtNodeWatcher) watch(ctx context.Context) {
...
 				}
 			}()
-		case nu := <-ch:
-			if nu.Reset {
-				// Ignore reset events to avoid clearing the access list before setting a new one.
-				// This is safe because a reset event is always followed by a freeze event after the
-				// nodes have been set (even if the new set is empty).
-				continue
+		case ne := <-nodeCh:
+			if !ne.IsRegistration {
```
kostko, line 352: Maybe use a `switch` for this and add comments for both cases (e.g. registration vs. expiry).

kostko (approved):

go/worker/keymanager/watcher.go
```diff
@@ -303,26 +325,50 @@ func (w *rtNodeWatcher) watch(ctx context.Context) {
...
+				clear(computeNodes)
 
 				for _, cm := range cms {
 					if cm.Kind != scheduler.KindComputeExecutor {
 						continue
 					}
 
 					for _, member := range cm.Members {
-						_, _ = watcher.WatchNode(ctx, member.PublicKey)
+						nd, err := w.consensus.Registry().GetNode(ctx, &registry.IDQuery{
```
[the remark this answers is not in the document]
peternose: Ups, my comment was wrong. You cannot use that method, so please ignore this comment.

PR #5610, opened 2024-03-22
testing: buildkite dcap

ptrus:
Update SGX Buildkite agents to use DCAP.

TODO:
- [x] test if this works
- [x] update all sgx buildkite agents
- [x] change buildkite agents queue back to `queue: sgx1` before merging
- [ ] maybe backoprt this PR, in case we will need a stable release?


.changelog/5610.feature.md
```diff
@@ -0,0 +1 @@
+ci: Update SGX tests to run DCAP
```
kostko, line 1: Maybe change this to `internal`.

go/common/sgx/pcs/tcb.go
```diff
@@ -271,15 +272,21 @@ func (ti *TCBInfo) validateTCBLevel(
...
 		// These are ok.
-	default:
-		return &TCBOutOfDateError{
-			Kind:        TCBKindPlatform,
-			Status:      tcbLevel.Status,
-			AdvisoryIDs: tcbLevel.AdvisoryIDs,
+		return nil
+	case StatusOutOfDate, StatusOutOfDateConfigurationNeeded:
+		// Ok if `OASIS_UNSAFE_LAX_AVR_VERIFY` is set.
+		if os.Getenv("OASIS_UNSAFE_LAX_AVR_VERIFY") != "" {
```
kostko, line 278: Don't do it like that as it seems quite dangerous. I would handle it similarly to how we have `unsafeAllowDebugEnclaves` and then set this in `oasis-node/cmd/common` based on some debug configuration option (gated also on `flags.DebugDontBlameOasis()`).

ptrus: Updated

kostko (approved):

PR #5613, opened 2024-03-28
go/worker/keymanager: Fix node ACL management

jberci:


.changelog/5613.bugfix.md
kostko, line 1: This can be `trivial` as the change was introduced in the same release.

go/worker/keymanager/watcher.go
```diff
@@ -313,41 +314,48 @@ func (w *rtNodeWatcher) watch(ctx context.Context) {
...
-					RuntimeID: w.runtimeID,
-				})
-				if err != nil {
-					w.logger.Error("failed to fetch runtime committee",
-						"err", err,
-					)
-					return
+			// Get executor committee members.
+			// Old ones need to be cleared out of the ACL even in case of errors.
+			clear(computeNodes)
```
kostko, line 319: I would move this above and add a blank line so it stands out on its own.

go/worker/keymanager/watcher.go
```diff
@@ -313,41 +314,48 @@ func (w *rtNodeWatcher) watch(ctx context.Context) {
...
+			})
+			if err != nil {
+				w.logger.Error("failed to fetch runtime committee",
+					"err", err,
+				)
+				break
+			}
+
+			// Get a list of all nodes at once, so we can match committee members with their descriptors.
+			allNodes, err := w.consensus.Registry().GetNodes(ctx, consensus.HeightLatest)
```
kostko, line 332: Why are we fetching many nodes that we don't actually need? The previous version only fetches committee members.

peternose: My comment was wrong, so this is actually my mistake. I thought we have method `GetNodesByID`, but we don't.

kostko (approved):

peternose (approved):

PR #5622, opened 2024-04-02
go/archive: fix runtime queries on archive nodes

ptrus:
TODO: Also backport to 22.2.x

Fix runtime queries on archive nodes. Also adds a runtime query to the archive_api E2E test.

Runtime queries are failing on archive due to the `CurrentBlock` not being set in the common node.

### DIFF
--- a/go/consensus/cometbft/full/archive.go
+++ b/go/consensus/cometbft/full/archive.go
@@ -16,6 +16,7 @@ import (
 	cmtcore "github.com/cometbft/cometbft/rpc/core"
 	"github.com/cometbft/cometbft/state"
 	"github.com/cometbft/cometbft/store"
+	cmttypes "github.com/cometbft/cometbft/types"
 	"github.com/spf13/viper"
 
 	"github.com/oasisprotocol/oasis-core/go/common/identity"
@@ -37,6 +38,7 @@ type archiveService struct {
 	*commonNode
 
 	abciClient abcicli.Client
+	eb         *cmttypes.EventBus
 
 	quitCh chan struct{}
 
@@ -47,6 +49,10 @@ type archiveService struct {
 func (srv *archiveService) Start() error {
 	if srv.started() {
 		return fmt.Errorf("cometbft: service already started")
+	}
+
+	if err := srv.eb.Start(); err != nil {
+		return err
 	}
 
 	if err := srv.commonNode.start(); err != nil {
@@ -69,6 +75,15 @@ func (srv *archiveService) Start() error {
 			}
 		}
 	}()
+
+	// Start command dispatchers for all the service clients.
+	srv.serviceClientsWg.Add(len(srv.serviceClients))
+	for _, svc := range srv.serviceClients {
+		go func() {
+			defer srv.serviceClientsWg.Done()
+			srv.serviceClientWorker(srv.ctx, svc)
+		}()
+	}
 
 	srv.commonNode.finishStart()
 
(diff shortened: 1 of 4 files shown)


go/consensus/cometbft/full/archive.go
```diff
@@ -236,3 +246,34 @@ func NewArchive(
...
+	logger := t.Logger.With("service", sd.Name())
+	logger.Info("starting command dispatcher")
+
+	// Service client event loop.
+	var height int64
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case cmd := <-sd.Commands():
```
kostko, line 269: The `serviceClientWorker` in the non-archive full node first waits for a block before processing commands. Not sure if this is relevant here.

go/consensus/cometbft/full/archive.go
```diff
@@ -70,6 +72,12 @@ func (srv *archiveService) Start() error {
 		}
 	}()
 
+	// Start command dispatchers for all the service clients.
+	srv.serviceClientsWg.Add(len(srv.serviceClients))
+	for _, svc := range srv.serviceClients {
+		go srv.serviceClientWorker(srv.ctx, svc)
```
peternose, line 82:
```suggestion
		go func ()  {
			defer srv.serviceClientsWg.Done()
			srv.serviceClientWorker(srv.ctx, svc)
		}()
```
This way, additions and subtractions to the `WaitGroup` are in the same place (less prone to bugs), and the method `serviceClientWorker` is not dependent on it (can be tested without it).

But if you want to keep it in line with `fullService`, then just leave it as it is.

go/consensus/cometbft/full/archive.go
```diff
@@ -236,3 +249,33 @@ func NewArchive(
...
+	logger.Info("starting command dispatcher")
+
+	// Service client event loop.
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case cmd := <-sd.Commands():
+			// New command.
+			if err := svc.DeliverCommand(ctx, 0, cmd); err != nil {
```
peternose, line 273: A comment explaining why `height=0` is appropriate in this case would be helpful.

go/consensus/cometbft/full/archive.go
```diff
@@ -236,3 +249,33 @@ func NewArchive(
...
+
+	logger := srv.Logger.With("service", sd.Name())
+	logger.Info("starting command dispatcher")
+
+	// Service client event loop.
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case cmd := <-sd.Commands():
```
peternose, line 271: Most apps don't emit commands, so one could stop the routine immediately. 
```go
if sd.Commands() == nil {
    return
```

go/consensus/cometbft/full/archive.go
ptrus: Updated

go/consensus/cometbft/full/archive.go
ptrus: Updated

go/consensus/cometbft/full/archive.go
ptrus: Probably since a non-archive node can be without blocks, and it needs the height of the latest block to be passed in the `DeliverCommand`. I have updated the archive node worker to also query the latest block on start and use the height in the `DeliverCommand`.

kostko (approved):

PR #5627, opened 2024-04-04
go/runtime/bundle: Multiple minor fixes

kostko:


go/runtime/bundle/manifest.go
```diff
@@ -148,6 +151,38 @@ func (c ComponentID) String() string {
...
+
+	if kind == ComponentRONL && len(atoms) > 1 && len(atoms[1]) > 0 {
+		return fmt.Errorf("RONL component must have an empty name, got: %s", atoms[1])
+	}
+
+	c.Kind = kind
+	if len(atoms) > 1 {
+		c.Name = atoms[1]
+	} else {
+		c.Name = ""
```
peternose, line 185:
```suggestion
```

go/runtime/bundle/manifest.go
```diff
@@ -148,6 +151,38 @@ func (c ComponentID) String() string {
 	return string(c.Kind) + " (" + c.Name + ")"
 }
 
+// MarshalText serializes the component identifier into text form.
+func (c ComponentID) MarshalText() ([]byte, error) {
+	return []byte(string(c.Kind) + kindNameSeparator + c.Name), nil
```
peternose, line 160: I would expect that the component identifier is unique, but is not as `"ronl."` and `"ronl"` unmarshal the same way.

go/runtime/bundle/manifest.go
```diff
@@ -148,6 +151,38 @@ func (c ComponentID) String() string {
...
+	case string(ComponentRONL):
+		kind = ComponentRONL
+	case string(ComponentROFL):
+		kind = ComponentROFL
+	default:
+		return fmt.Errorf("malformed component kind: %s", atoms[0])
+	}
+
+	if kind == ComponentRONL && len(atoms) > 1 && len(atoms[1]) > 0 {
+		return fmt.Errorf("RONL component must have an empty name, got: %s", atoms[1])
```
peternose, line 178:
```suggestion
	if c.Kind == ComponentRONL && len(c.Name) > 0 {
		return fmt.Errorf("RONL component must have an empty name, got: %s", c.Name)
	}
```

peternose (approved):

go/runtime/bundle/manifest.go
peternose: Ignore, as you probably don't want to change `c` if you return an error.

go/runtime/bundle/manifest.go
kostko: Hm? You don't know what the value was before `UnmarshalText` was called and this makes sure it is reset.

go/runtime/bundle/manifest.go
kostko: Yeah exactly.

go/runtime/bundle/bundle.go
```diff
@@ -152,6 +143,23 @@ func (bnd *Bundle) EnclaveIdentity(id ComponentID) (*sgx.EnclaveIdentity, error)
 	if err := mrEnclave.FromSgxs(bytes.NewReader(d)); err != nil {
 		return nil, fmt.Errorf("runtime/bundle: failed to derive SGX MRENCLAVE for '%s': %w", id, err)
 	}
+	return &mrEnclave, nil
+}
+
+// EnclaveIdentity returns the SGX enclave identity of the given component.
+func (bnd *Bundle) EnclaveIdentity(id ComponentID) (*sgx.EnclaveIdentity, error) {
```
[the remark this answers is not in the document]
kostko: Good idea, will change.

go/runtime/bundle/manifest.go
kostko: Fixed.

PR #5628, opened 2024-04-05
go/keymanager/churp: Allow nodes to confirm handoff completion

peternose:


go/consensus/cometbft/apps/keymanager/churp/txs.go
```diff
@@ -304,6 +290,128 @@ func (ext *churpExt) apply(ctx *tmapi.Context, req *churp.SignedApplicationReque
...
+	}
+	if err = req.VerifyRAK(rak); err != nil {
+		return fmt.Errorf("keymanager: churp: invalid signature: %w", err)
+	}
+
+	if ctx.IsCheckOnly() {
+		return nil
+	}
+
+	// Charge gas for this operation.
```
kostko, line 364: I think gas accounting should be done before any expensive operations like reading state or verifying signatures (also elsewhere).

go/consensus/cometbft/apps/keymanager/churp/txs.go
```diff
@@ -304,6 +290,128 @@ func (ext *churpExt) apply(ctx *tmapi.Context, req *churp.SignedApplicationReque
...
+	for _, app := range status.Applications {
+		if !app.Reconstructed {
+			handoffCompleted = false
+			break
+		}
+	}
+	if handoffCompleted {
+		status.Handoff = status.NextHandoff
+		status.Checksum = status.NextChecksum
+		status.Committee = maps.Keys(status.Applications)
```
kostko, line 395: This is non-deterministic and will cause consensus failures. You need to sort the entries.

secret-sharing/src/vss/scalar.rs
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    slice.copy_from_slice(bytes);
+
+    let point = match Fp::from_repr(repr).into() {
+        None => return None,
+        Some(bij) => bij,
+    };
+
+    Some(point)
```
kostko, line 30: Shouldn't this just be:
```rust
Fp::from_repr(repr).into()
```

secret-sharing/src/churp/player.rs
```diff
@@ -53,20 +61,62 @@ where
...
+where
+    D: DealerParams,
+{
+    fn from(state: SecretShare<D::Group>) -> Player<D> {
+        Player { share: state }
+    }
+}
+
+/// Secret share of the shared secret.
+#[derive(PartialEq, Eq)]
```
kostko, line 92: Where do you need to do comparisons? These will not be constant-time so can leak secret state.

go/keymanager/churp/status.go
```diff
@@ -70,13 +56,40 @@ type Status struct {
...
+	//
+	// If an insufficient number of applications is received, the next handoff
+	// will be delayed by one epoch.
+	NextHandoff beacon.EpochTime `json:"next_handoff"`
+
+	// NextChecksum is the hash of the verification matrix from the current
+	// handoff.
+	//
+	// The first candidate to confirm share reconstruction is the source
+	// of truth for the checksum. All other candidates need to confirm
```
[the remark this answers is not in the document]
peternose: > Is there a way to prove this in a way that the consensus layer could easily (cheaply) verify? 

No, at least I don't see it. And taking into account that we can replace checksum with some other commitment.
To prevent DoS attacks, we could add slashing. But 

> Also all other candidates or is a threshold enough?

Threshold is enough. But we want to give participants enough time to do the handoff. So maybe all candidates to complete before the end of an epoch, and threshold at the end of an epoch.

secret-sharing/src/churp/player.rs
peternose: In tests only. Agree, need to remove this.

go/keymanager/churp/status.go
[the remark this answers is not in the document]
kostko: Ok, that sounds good. No slashing needed for now.

secret-sharing/src/churp/player.rs
kostko: You can do it only for test builds:
```rust
#[cfg_attr(test, derive(PartialEq, Eq))]
```

peternose: Tnx, didn't know that. This is needed in the next part, so will remove for now.

kostko (approved):

PR #5630, opened 2024-04-08
go/runtime/txpool: Don't abort runtime if node is not synced yet

abukosek:
If the node hasn't finished syncing, `checkTxBatch` previously caused the runtime to be aborted, even though it wasn't the runtime's fault.
Now the checks are retried after the node is finished syncing.

### DIFF
--- a/.changelog/5630.bugfix.md
+++ b/.changelog/5630.bugfix.md
@@ -1,0 +1,6 @@
+go/runtime/txpool: Don't abort runtime if node is not synced yet
+
+If the node hasn't finished syncing, `checkTxBatch` previously
+caused the runtime to be aborted, even though it wasn't the
+runtime's fault.
+Now the checks are retried after the node is finished syncing.
--- a/go/runtime/txpool/txpool.go
+++ b/go/runtime/txpool/txpool.go
@@ -28,6 +28,9 @@ const (
 	checkTxTimeout = 15 * time.Second
 	// checkTxRetryDelay is the time to wait before queuing a check tx retry.
 	checkTxRetryDelay = 1 * time.Second
+	// checkTxWaitRoundSyncedTimeout is the time to wait for block to be
+	// synced with storage at the start of check tx.
+	checkTxWaitRoundSyncedTimeout = 5 * time.Second
 	// abortTimeout is the maximum time the runtime can spend aborting.
 	abortTimeout = 5 * time.Second
 	// maxRepublishTxs is the maximum amount of transactions to republish.
@@ -460,20 +463,31 @@ func (t *txPool) checkTxBatch(ctx context.Context, rr host.RichRuntime) {
 		return
 	}
 
+	// Ensure block round is synced to storage.
+	waitSyncCtx, cancelWaitSyncCtx := context.WithTimeout(ctx, checkTxWaitRoundSyncedTimeout)
+	defer cancelWaitSyncCtx()
+
+	t.logger.Debug("ensuring block round is synced", "round", bi.RuntimeBlock.Header.Round)
+	if _, err = t.history.WaitRoundSynced(waitSyncCtx, bi.RuntimeBlock.Header.Round); err != nil {
+		// Block round isn't synced yet, so make sure the batch check is
+		// retried later to avoid aborting the runtime, as it is not its fault.
+		t.logger.Info("block round is not synced yet, retrying transaction batch check later",
+			"round", bi.RuntimeBlock.Header.Round,
+			"err", err,
+		)
+		t.checkTxCh.In() <- struct{}{}
+		return
+	}
+
+	// Pop the next batch from the queue, check it, and notify submitters.
 	batch := t.checkTxQueue.pop()
 	if len(batch) == 0 {
 		return
 	}
 
 	results, err := func() ([]protocol.CheckTxResult, error) {
-		checkCtx, cancel := context.WithTimeout(ctx, checkTxTimeout)
-		defer cancel()
-
-		// Ensure block round is synced to storage.
-		t.logger.Debug("ensuring block round is synced", "round", bi.RuntimeBlock.Header.Round)
-		if _, err = t.history.WaitRoundSynced(checkCtx, bi.RuntimeBlock.Header.Round); err != nil {
-			return nil, err
-		}
+		checkCtx, cancelCheckCtx := context.WithTimeout(ctx, checkTxTimeout)
+		defer cancelCheckCtx()
 
 		// Check batch.
 		rawTxBatch := make([][]byte, 0, len(batch))
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2736,9 +2736,9 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.22.2"
+version = "0.22.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e87c9956bd9807afa1f77e0f7594af32566e830e088a5576d27c5b6f30f49d41"
+checksum = "bf4ef73721ac7bcd79b2b315da7779d8fc09718c6b3d2d1b2d94850eb8c18432"
 dependencies = [
  "log",
  "rustls-pki-types",
--- a/go/worker/common/committee/node.go
+++ b/go/worker/common/committee/node.go
@@ -229,11 +229,6 @@ func (n *Node) Start() error {
 		return fmt.Errorf("failed to start group services: %w", err)
 	}
 
-	// Start the transaction pool.
-	if err := n.TxPool.Start(); err != nil {
-		return fmt.Errorf("failed to start transaction pool: %w", err)
-	}
-
 	go n.worker()
 	if cmmetrics.Enabled() {
 		go n.metricsWorker()
@@ -615,6 +610,14 @@ func (n *Node) worker() {
 	}
 	n.logger.Info("consensus has finished initial synchronization")
 	atomic.StoreUint32(&n.consensusSynced, 1)
+
+	// Start the transaction pool after consensus is synced.
+	if err := n.TxPool.Start(); err != nil {
+		n.logger.Error("failed to start transaction pool",
+			"err", err,
+		)
+		return
+	}
 
 	// Wait for the runtime.
 	rt, err := n.Runtime.ActiveDescriptor(n.ctx)
--- a/tests/runtimes/simple-rofl/Cargo.toml
+++ b/tests/runtimes/simple-rofl/Cargo.toml
@@ -21,7 +21,7 @@ anyhow = "1.0"
 async-trait = "0.1.77"
 mbedtls = { version = "0.12.3", features = ["x509"] }
 rand = "0.8.5"
-rustls = { version = "0.22.2", default-features = false }
+rustls = { version = "0.22.4", default-features = false }
 rustls-mbedcrypto-provider = { version = "0.0.2" }
 rustls-mbedpki-provider = { version = "0.0.2" }
 tokio = { version = "1.36.0", features = ["rt", "rt-multi-thread", "sync"] }


go/runtime/txpool/txpool.go
```diff
@@ -460,20 +466,49 @@ func (t *txPool) checkTxBatch(ctx context.Context, rr host.RichRuntime) {
...
+		notReady = true
+		t.logger.Debug("consensus is not synced yet")
+	}
+
+	if !notReady {
+		// Node is synced, ensure block round is synced to storage.
+		waitSyncCtx, cancelWaitSyncCtx := context.WithTimeout(ctx, checkTxWaitRoundSyncedTimeout)
+		defer cancelWaitSyncCtx()
+
+		t.logger.Debug("ensuring block round is synced", "round", bi.RuntimeBlock.Header.Round)
```
peternose, line 485:
```suggestion
		t.logger.Debug("ensuring runtime block is synced", "round", bi.RuntimeBlock.Header.Round)
```

go/runtime/txpool/txpool.go
```diff
@@ -894,6 +930,7 @@ func New(
 		host:                 host,
 		history:              history,
 		txPublisher:          txPublisher,
+		consensus:            consensus,
```
peternose, line 933: Feels like the pool shouldn't have this dependency. Maybe the node should start the pool in the `worker` method once consensus is synced, not sure.

go/runtime/txpool/txpool.go
```diff
@@ -460,20 +466,49 @@ func (t *txPool) checkTxBatch(ctx context.Context, rr host.RichRuntime) {
...
+
+		t.logger.Debug("ensuring block round is synced", "round", bi.RuntimeBlock.Header.Round)
+		if _, err = t.history.WaitRoundSynced(waitSyncCtx, bi.RuntimeBlock.Header.Round); err != nil {
+			notReady = true
+			t.logger.Debug("block round is not synced yet", "err", err)
+		}
+	}
+
+	if notReady {
+		// Node or block round aren't synced yet, so make sure the batch
```
peternose, line 493: I would say that the  `checkWorker` should wait for consensus sync, as this is a necessary precondition for all checks. So you don't need an extra timeout and the pool can start processing immediately after sync.

go/runtime/txpool/txpool.go
```diff
@@ -460,20 +466,49 @@ func (t *txPool) checkTxBatch(ctx context.Context, rr host.RichRuntime) {
...
+		t.logger.Debug("consensus is not synced yet")
+	}
+
+	if !notReady {
+		// Node is synced, ensure block round is synced to storage.
+		waitSyncCtx, cancelWaitSyncCtx := context.WithTimeout(ctx, checkTxWaitRoundSyncedTimeout)
+		defer cancelWaitSyncCtx()
+
+		t.logger.Debug("ensuring block round is synced", "round", bi.RuntimeBlock.Header.Round)
+		if _, err = t.history.WaitRoundSynced(waitSyncCtx, bi.RuntimeBlock.Header.Round); err != nil {
```
peternose, line 486: If the block is synced and we call this method, do we need to wait for 6 seconds for the next block to be committed so that the `blocksNotifier` can broadcast the last block? I think that the history doesn't respond immediately as it doesn't track the latest synced round.

Why should txs check wait for round sync?

I would reorganize the code so that this call is not indented in `if`, because it looks like the happy path should  always check this. And I would think of blocking here until the round is synced. I think that currently it is possible for the check to never execute, if syncing takes too long. The reason is that the check call will fail because the block is not synced yet, but the next call will wait for a higher block to be synced, so it will need to wait and fail again.

.changelog/5630.bugfix.md
```diff
@@ -0,0 +1,6 @@
+go/runtime/txpool: Don't abort runtime if node is not synced yet
+
+If the node hasn't finished syncing, `checkTxBatch` previously
+caused the runtime to be aborted, even though it wasn't the
```
peternose, line 4: Why should txs check wait for a round sync? And if the runtime really aborts, do we need to fix the runtime instead?

go/runtime/txpool/txpool.go
peternose: So maybe this function should start like this:
```go
bi, lastSyncedBlockProcessed, err := t.getLastSyncedBlockInfo()
```

.changelog/5630.bugfix.md
```diff
@@ -0,0 +1,6 @@
+go/runtime/txpool: Don't abort runtime if node is not synced yet
```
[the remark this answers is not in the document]
abukosek: The commit message is already in imperative mood.

go/runtime/txpool/txpool.go
```diff
@@ -460,20 +466,49 @@ func (t *txPool) checkTxBatch(ctx context.Context, rr host.RichRuntime) {
...
+	select {
+	case <-t.consensus.Synced():
+		// Node is synced.
+	default:
+		// Node is not synced.
+		notReady = true
+		t.logger.Debug("consensus is not synced yet")
+	}
+
+	if !notReady {
```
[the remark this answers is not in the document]
abukosek: The semantics are different though.  If you look at the bigger picture, `notReady` makes more sense :)

go/runtime/txpool/txpool.go
abukosek: I would prefer to retain the original debug message, since I didn't write it.

go/runtime/txpool/txpool.go
abukosek: This is the only way to enable calling the `consensus.Synced` method.  I'm pretty sure the node can't move txpool init to after the consensus is synced because the txpool methods can be called before that happens and if it's not inited, they would just crash.

go/runtime/txpool/txpool.go
abukosek: It still needs to wait for block sync, though.

go/runtime/txpool/txpool.go
abukosek: Hmm.  I will think about this tomorrow.

.changelog/5630.bugfix.md
abukosek: No, the runtime is aborted erroneously -- it's not the runtime's fault, but the fault of the code in `checkTxBatch`, which this PR fixes!  The previous code initiates the runtime abort, not the runtime itself -- in fact, the runtime doesn't even receive anything, because the block isn't synced yet.

.changelog/5630.bugfix.md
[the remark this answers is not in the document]
peternose: You are right :facepalm: 
However, the commit is usually read like this: `This commit will: Don't abort runtime if node is not synced yet`.

go/runtime/txpool/txpool.go
peternose: Yes, wait here for block sync, and for consensus there.

go/runtime/txpool/txpool.go
peternose: Agree. I was just thinking about how would you unit-test this. And you would need to mock the consensus. On the other hand, starting the worker from the outside once consensus is synced, is cleaner.

go/runtime/txpool/txpool.go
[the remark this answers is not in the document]
peternose: Well, maybe it still could be improved. I would expect the code to look like this:
```go
// Check if we are ready.

if notReady {
    // Return, error or fix so that we will be ready.
}

// Now that we are ready, do something.
```

go/runtime/txpool/txpool.go
kostko: > I think that the history doesn't respond immediately as it doesn't track the latest synced round.

In case the node has local storage available, history also waits for storage sync.

> Why should txs check wait for round sync?

Because otherwise it can't do the checks since the state will not be available and accessing the trie will fail.

.changelog/5630.bugfix.md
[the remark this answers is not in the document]
abukosek: That doesn't seem a rule that this repository follows strictly :)
```
$ git log --format=oneline | grep -i "don't" | wc -l
80
```

I don't know how to write a title that still fits into the length limit and conveys the same amount of information as the existing one :man_shrugging:

go/runtime/txpool/txpool.go
abukosek: But if the node loses consensus sync, we still need to check for it here, so we don't blame the runtime.  I can add an additional wait for consensus sync into `checkWorker`, but we still need to retain it here, I think?

peternose: Loses sync? I think that once consensus is synced, the call to `Consensus.Synced()` will always return a closed channel.

abukosek: Oh, OK, I will make it wait in the worker and just check for block sync here then.

.changelog/5630.bugfix.md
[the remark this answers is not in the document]
peternose: > That doesn't seem a rule that this repository follows strictly :)
:smile: 

Just leave as it is. I just read somewhere the following and I think this is a good practice.
> A properly formed Git commit subject line should always be able to complete the following sentence:
> If applied, this commit will (your subject line here).

go/runtime/txpool/txpool.go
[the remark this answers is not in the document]
abukosek: Moving the consensus sync wait has simplified this part of the code and the `notReady` bool no longer exists :)

go/runtime/txpool/txpool.go
abukosek: Done, please have another look :)

go/runtime/txpool/txpool.go
```diff
@@ -460,20 +468,35 @@ func (t *txPool) checkTxBatch(ctx context.Context, rr host.RichRuntime) {
...
+	if _, err = t.history.WaitRoundSynced(waitSyncCtx, bi.RuntimeBlock.Header.Round); err != nil {
+		// Block round isn't synced yet, so make sure the batch check is
+		// retried later to avoid aborting the runtime, as it is not its fault.
+		t.logger.Info("block round is not synced yet, retrying transaction batch check later",
+			"round", bi.RuntimeBlock.Header.Round,
+			"err", err,
+		)
+
+		go func() {
+			time.Sleep(checkTxWaitForNodeSyncRetryDelay)
```
peternose, line 485: This can be removed if you wait for round synced for additional `checkTxWaitForNodeSyncRetryDelay`. Also, I don't see a point to wait in the background, as if the block is not synced, the next call will also fail. So we can remove go routine and just send to the channel to call this method again immediately. 

```go
	waitSyncCtx, cancelWaitSyncCtx := context.WithTimeout(ctx, checkTxWaitRoundSyncedTimeout)
	defer cancelWaitSyncCtx()

	t.logger.Debug("ensuring block round is synced", "round", bi.RuntimeBlock.Header.Round)
	if _, err = t.history.WaitRoundSynced(waitSyncCtx, bi.RuntimeBlock.Header.Round); err != nil {
		// Block round isn't synced yet, so make sure the batch check is
		// retried later to avoid aborting the runtime, as it is not its fault.
		t.logger.Info("block round is not synced yet, retrying transaction batch check later",
			"round", bi.RuntimeBlock.Header.Round,
			"err", err,
		)
		t.checkTxCh.In() <- struct{}{}
		return
	}
```

go/runtime/txpool/txpool.go
```diff
@@ -659,6 +682,13 @@ func (t *txPool) checkWorker() {
 		return
 	}
 
+	// Wait for consensus to sync.
```
peternose, line 685:
```suggestion
	// Wait for consensus to sync, .... (description why).
```

go/runtime/txpool/txpool.go
peternose: Did you think about the comment above (that we always need to wait up to 6s)?

peternose: So I would update the `runtimeHistory` to store the last synced block into a local variable (probably locked) in `Commit`, and then in `WaitRoundSynced` first check the local variable, and then listen to blocks.

abukosek: As we discussed in the meeting, there is no 6s delay when the node is synced, since the history pubsub broker is instantiated with `publishLastOnSubscribe` set to `true` (https://github.com/oasisprotocol/oasis-core/blob/master/go/runtime/history/history.go#L411), so it will get the last block when subscribing :)

abukosek:
All review comments have been addressed, so please take another look :)

go/runtime/txpool/txpool.go
peternose: The `blockNotifier` pushes the last block immediately, so my comment is way off and can be ignored. Resolving.

go/runtime/txpool/txpool.go
kostko: This is a good point, the `TxPool` is started from the common worker (`worker/common/committee/node.go`) in its `Start` method. You could instead start it after the common worker checks that consensus is synced.

go/runtime/txpool/txpool.go
```diff
@@ -28,6 +29,12 @@ const (
 	checkTxTimeout = 15 * time.Second
 	// checkTxRetryDelay is the time to wait before queuing a check tx retry.
 	checkTxRetryDelay = 1 * time.Second
+	// checkTxWaitForNodeSyncRetryDelay is the time to wait before retrying
+	// check tx while waiting for the node to sync.
+	checkTxWaitForNodeSyncRetryDelay = 30 * time.Second
+	// checkTxWaitRoundSyncedTimeout is the time to wait for block to be
+	// synced with storage at the start of check tx.
+	checkTxWaitRoundSyncedTimeout = 15 * time.Second
```
kostko, line 37: I would reduce this timeout to something like 5 seconds as it will be retried anyway.

abukosek: Done :)  I've removed `checkTxWaitForNodeSyncRetryDelay` as suggested by Peter and reduced `checkTxWaitRoundSyncedTimeout` to 5 seconds.

go/runtime/txpool/txpool.go
```diff
@@ -460,20 +463,31 @@ func (t *txPool) checkTxBatch(ctx context.Context, rr host.RichRuntime) {
...
 
+	// Ensure block round is synced to storage.
+	waitSyncCtx, cancelWaitSyncCtx := context.WithTimeout(ctx, checkTxWaitRoundSyncedTimeout)
+	defer cancelWaitSyncCtx()
+
+	t.logger.Debug("ensuring block round is synced", "round", bi.RuntimeBlock.Header.Round)
+	if _, err = t.history.WaitRoundSynced(waitSyncCtx, bi.RuntimeBlock.Header.Round); err != nil {
+		// Block round isn't synced yet, so make sure the batch check is
+		// retried later to avoid aborting the runtime, as it is not its fault.
+		t.logger.Info("block round is not synced yet, retrying transaction batch check later",
```
peternose, line 474:
```suggestion
		t.logger.Warn("block round is not synced yet, retrying transaction batch check later",
```
Should this be `Warn`? :man_shrugging:

peternose (approved):

go/runtime/txpool/txpool.go
abukosek: Probably not, since the user can't really do anything about it?

kostko (approved):

PR #5632, opened 2024-04-09
runtime: Add EndorsedCapabilityTEE::verify_endorsement

kostko:
This can be useful when only checking the endorsement itself is important as the content will be verified separately.

### DIFF
--- a/runtime/src/consensus/registry.rs
+++ b/runtime/src/consensus/registry.rs
@@ -191,18 +191,26 @@ pub struct EndorsedCapabilityTEE {
 }
 
 impl EndorsedCapabilityTEE {
-    /// Verify endorsed TEE capability is valid.
-    pub fn verify(
-        &self,
-        policy: &sgx::QuotePolicy,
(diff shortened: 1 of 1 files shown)


runtime/src/consensus/registry.rs
```diff
@@ -191,18 +191,26 @@ pub struct EndorsedCapabilityTEE {
...
-    /// Verify endorsed TEE capability is valid.
-    pub fn verify(
-        &self,
-        policy: &sgx::QuotePolicy,
-    ) -> anyhow::Result<VerifiedEndorsedCapabilityTEE> {
-        // Verify node endorsement.
+    /// Verify the endorsement signature is valid.
+    ///
+    /// **This does not verify the TEE capability itself, use `verify` for that.**
+    pub fn verify_endorsement_only(&self) -> anyhow::Result<()> {
```
peternose, line 197:
```suggestion
    pub fn verify_endorsement(&self) -> anyhow::Result<()> {
```

peternose (approved):

PR #5633, opened 2024-04-10
go/keymanager/churp: Minor fixes

peternose:


kostko (approved):

peternose:
Fixed the `multiplier` to return `None` (which equals to 1) if no values are given.

PR #5637, opened 2024-04-11
Multiple minor runtime-related improvements

kostko:


go/runtime/registry/config.go
```diff
@@ -247,13 +248,32 @@ func newConfig( //nolint: gocyclo
...
-			// For compute nodes we also want to run the ROFL component.
-			if config.GlobalConfig.Mode == config.ModeCompute {
-				// TODO: Support configuring named components.
-				wantedComponents = append(wantedComponents, bundle.ComponentID{Kind: bundle.ComponentROFL})
+			for _, comp := range bnd.Manifest.Components {
+				if comp.ID() == component.ID_RONL {
+					continue // Always enabled above.
+				}
+
+				// By default honor the status of the component itself.
```
peternose, line 259: Could also be done top-down:

```go
// Check for any overrides in the node configuration.
compCfg, ok := config.GlobalConfig.Runtime.GetComponent(comp.ID())
if ok && compCfg.Disabled {
	continue
}

// On non-compute nodes, assume all components are disabled by default.
if !ok && config.GlobalConfig.Mode != config.ModeCompute {
	continue
}

// By default honor the status of the component itself.
if !ok && comp.Disabled {
	continue
}
```

peternose (approved):

PR #5641, opened 2024-04-17
Build runtime binaries in release mode

peternose:
Running SGX tests with binaries built in debug mode can be extremely slow, so build everything in release mode. This should also speed up E2E tests.

### DIFF
--- a/go/oasis-test-runner/rust/builder.go
+++ b/go/oasis-test-runner/rust/builder.go
@@ -82,6 +82,7 @@ func (b *Builder) Build(pkg string) error {
 	}
 
 	cmd.Args = append(cmd.Args, "--package", pkg)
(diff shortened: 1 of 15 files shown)


go/oasis-test-runner/rust/builder.go
```diff
@@ -82,6 +82,7 @@ func (b *Builder) Build(pkg string) error {
 	}
 
 	cmd.Args = append(cmd.Args, "--package", pkg)
+	cmd.Args = append(cmd.Args, "--release")
```
kostko, line 85: The `elf2sgxs` command also needs this flag.

peternose: Yes, figured that out because tests are failing.

kostko (approved):

PR #5642, opened 2024-04-17
Add support for mock SGX builds

kostko:


common.mk
```diff
@@ -351,3 +351,10 @@ endif
 # For more details, see:
 # https://goreleaser.com/customization/build/#define-build-tag
 export GORELEASER_CURRENT_TAG := $(RELEASE_TAG)
+
+# If mock SGX is configured, define extra runtime build flags.
+ifdef OASIS_UNSAFE_MOCK_SGX
+	OASIS_RUNTIME_NONSGX_FLAGS := --features debug-mock-sgx
+else
+	OASIS_RUNTIME_NONSGX_FLAGS :=
```
peternose, line 359: Can this be removed?

runtime/src/common/sgx/mod.rs
```diff
@@ -112,3 +116,30 @@ pub struct VerifiedQuote {
...
+/// Generate a report for the given target enclave.
+#[cfg(target_env = "sgx")]
+pub fn report_for(target_info: &sgx_isa::Targetinfo, report_data: &[u8; 64]) -> sgx_isa::Report {
+    sgx_isa::Report::for_target(target_info, report_data)
+}
+
+/// Generate a report for the given target enclave.
+#[cfg(not(target_env = "sgx"))]
+pub fn report_for(_target_info: &sgx_isa::Targetinfo, report_data: &[u8; 64]) -> sgx_isa::Report {
+    let ei = EnclaveIdentity::current().unwrap();
```
peternose, line 129: This should panic on non sgx non debug-mock-sgx builds.

go/runtime/host/sgx/sgx.go
```diff
@@ -82,6 +82,10 @@ type Config struct {
 
 	// InsecureNoSandbox disables the sandbox and runs the loader directly.
 	InsecureNoSandbox bool
+	// InsecureMock runs non-SGX binaries but treats it as if it would be running in an enclave,
+	// using mock quotes and reports. This is useful in tests so most SGX code can be tested even
```
peternose, line 86:
```suggestion
	// using mock quotes and reports.
	//
	// This is useful in tests so most SGX code can be tested even
```

runtime/src/identity.rs
```diff
@@ -5,42 +5,35 @@ use std::{
...
 /// Context used for computing the RAK digest.
-#[cfg_attr(not(target_env = "sgx"), allow(unused))]
 const RAK_HASH_CONTEXT: &[u8] = b"oasis-core/node: TEE RAK binding";
 /// Context used for deriving the nonce used in quotes.
-#[cfg_attr(not(target_env = "sgx"), allow(unused))]
 const QUOTE_NONCE_CONTEXT: &[u8] = b"oasis-core/node: TEE quote nonce";
 
 /// A dummy RAK seed for use in non-SGX tests where integrity is not needed.
-#[cfg(not(target_env = "sgx"))]
+#[cfg(not(any(target_env = "sgx", feature = "debug-mock-sgx")))]
```
peternose, line 33: Do we panic (or prevent) anywhere if `target_env = "sgx"` and `feature = "debug-mock-sgx"`?

runtime/src/enclave_rpc/client.rs
```diff
@@ -490,7 +465,18 @@ impl RpcClient {
             args: cbor::to_value(args),
         };
 
-        let (pfid, inner) = match self.execute_call(request, kind).await {
+        // We can only retry transport errors as for those we are sure that the request was not
```
peternose, line 468: But do we retry only transport errors? I don't see any checks for `RpcClientError`. :thinking:

peternose (commented):
Great stuff!

Mock is disabled by default. Was this planned to be used only locally or also in CI? I guess the former, as otherwise non-sgx non-mock code could be neglected.

kostko:
> Great stuff!
> 
> 
> 
> Mock is disabled by default. Was this planned to be used only locally or also in CI? I guess the former, as otherwise non-sgx non-mock code could be neglected.

Currently only locally. We can add one test for mock SGX (as we also test non-mock SGX anyway). It will also be used more in the SDK CI.

go/common/sgx/pcs/mock.go
```diff
@@ -0,0 +1,125 @@
...
+
+import (
+	"encoding/binary"
+	"fmt"
+)
+
+// mockCertificationDataChain is the certificate chain that is used when generating mock quotes. The
+// chain is valid and rooted in an Intel root CA, but the mock quote is not actually signed by the
+// leaf PCK certificate key (that would require breaking SGX).
+const mockCertificationDataChain = `-----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
kostko: But the package is called `sgx` not `mock`?

.buildkite/code.pipeline.yml
```diff
@@ -245,7 +245,7 @@ steps:
         --scenario e2e/runtime/keymanager-.+
     env:
       # Unsafe flags needed as the trust-root test rebuilds the enclave with embedded trust root data.
-      OASIS_UNSAFE_LAX_AVR_VERIFY: "1"
```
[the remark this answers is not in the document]
kostko: Yes, if you check history it was like this before. We could add a build that tests e2e with checks enabled (but lax so CI agents don't need to be up-to-date).

runtime/src/common/sgx/mod.rs
kostko: It already does as those builds will return `None`.

runtime/src/enclave_rpc/client.rs
kostko: What kind of other errors will return an `Err`, I think all should be transport errors, but let me check.

runtime/src/identity.rs
kostko: Good point, will add a check in `lib.rs`.

runtime/src/common/sgx/mod.rs
peternose: What I meant was that this will panic on those builds. Is that ok or should we handle this differently?

runtime/src/enclave_rpc/client.rs
peternose: I didn't check that. But the comment is confusing, as I would except that we check that there is a transport error.

go/common/sgx/pcs/mock.go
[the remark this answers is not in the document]
peternose: Ups, commented too fast. You are right, its fine.

ptrus (approved):

peternose (approved):

PR #5648, opened 2024-04-19
go/staking: Read token information from state when available

kostko:
This unifies handling of the two remaining attributes so everything is consistently read from state, in a backwards compatible manner.

### DIFF
--- a/go/staking/api/api.go
+++ b/go/staking/api/api.go
@@ -114,11 +114,11 @@ var (
 // Backend is a staking implementation.
 type Backend interface {
 	// TokenSymbol returns the token's ticker symbol.
-	TokenSymbol(ctx context.Context) (string, error)
+	TokenSymbol(ctx context.Context, height int64) (string, error)
 
 	// TokenValueExponent is the token's value base-10 exponent, i.e.
 	// 1 token = 10**TokenValueExponent base units.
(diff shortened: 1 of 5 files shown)


go/staking/api/api.go
```diff
@@ -1183,6 +1183,13 @@ type Genesis struct {
 
 // ConsensusParameters are the staking consensus parameters.
 type ConsensusParameters struct { // nolint: maligned
+	// TokenSymbol is the token's ticker symbol.
+	// Only upper case A-Z characters are allowed.
```
peternose, line 1187: Where is this enforced?

go/staking/api/api.go
```diff
@@ -1183,6 +1183,13 @@ type Genesis struct {
...
+	TokenSymbol string `json:"token_symbol,omitempty"`
```
peternose, line 1188: When/where will these two fields be set?

peternose (approved):

go/staking/api/api.go
kostko: It is not, but it cannot really be changed outside an upgrade handler.

go/staking/api/api.go
kostko: Currently only via an upgrade handler. I would keep it this way for now.

PR #5650, opened 2024-04-21
go/governance: Add support for proposal metadata

kostko:


go/governance/api/api.go
```diff
@@ -60,21 +59,38 @@ var (
...
 // ValidateBasic performs basic proposal content validity checks.
-func (p *ProposalContent) ValidateBasic() error {
-	numProposals := 0
-	values := reflect.ValueOf(*p)
-	for i := 0; i < values.NumField(); i++ {
-		if !values.Field(i).IsNil() {
-			numProposals++
+func (p *ProposalContent) ValidateBasic(params *ConsensusParameters) error {
+	// Validate metadata if present.
+	if p.Metadata != nil {
```
peternose, line 73: Should metadata be mandatory?
``` go
switch params.AllowProposalMetadata {
	case true:
		if p.Metadata == nil {
			return fmt.Errorf("proposal metadata required")
		}
		if err := p.Metadata.ValidateBasic(); err != nil {
			return err
		}
	case false:
		if p.Metadata != nil {
			// Must be consistent with error during unmarshal.
			return ErrInvalidArgument
		}
	}
```

peternose (approved):

PR #5653, opened 2024-04-23
go/keymanager/churp: Require stake for every churp scheme

peternose:


go/common/entity/entity.go
```diff
@@ -146,6 +147,11 @@ func (e *Entity) Save(baseDir string) error {
 	return os.WriteFile(entityPath, b, fileMode)
 }
 
+// StakingAddress returns the staking address for the entity.
+func (e *Entity) StakingAddress() staking.Address {
```
kostko, line 151: The reason why this is not here is that we tried to avoid dependencies to non-common (e.g. staking) from common libraries.

go/oasis-node/cmd/genesis/migrate.go
```diff
@@ -357,6 +358,9 @@ NodeLoop:
 	// Reduce maximum consensus layer block size.
 	newDoc.Consensus.Parameters.MaxBlockSize = 1_048_576 // 1 MiB
 
+	// Add new staking thresholds.
+	newDoc.Staking.Parameters.Thresholds[staking.KindKeyManagerChurp] = *quantity.NewFromUint64(10_000_000_000_000)
```
kostko, line 362: We should make sure that this works with existing genesis files that don't have this threshold set. I believe there is some validation that would currently fail.

go/common/entity/entity.go
peternose: Ok, will remove both.

go/oasis-node/cmd/genesis/migrate.go
peternose: If I don't set this, the `SanityCheck` of staking parameters will fail as the new threshold kind is not defined. But I don't want to remove it from the valid threshold list.

go/consensus/cometbft/apps/keymanager/churp/txs.go
```diff
@@ -44,6 +47,11 @@ func (ext *churpExt) create(ctx *tmapi.Context, req *churp.CreateRequest) error
 		return fmt.Errorf("keymanager: churp: invalid signer")
 	}
 
+	// Add stake claim to the key manager owner.
+	if err = addStakeClaim(ctx, kmRt.EntityID, req.RuntimeID, req.ID); err != nil {
```
kostko, line 51: This should be done after all validation is done (you have a TODO below on line 86). Also since you are now updating multiple things you should start a storage transaction to ensure atomicity, e.g.:
```golang
// Start a new transaction and rollback in case we fail.
ctx = ctx.NewTransaction()
defer ctx.Close()
// ...
ctx.Commit()
```

go/consensus/cometbft/apps/keymanager/churp/txs.go
```diff
@@ -417,6 +425,39 @@ func (ext *churpExt) computeNextHandoff(ctx *tmapi.Context) (beacon.EpochTime, e
...
 }
 
+func addStakeClaim(ctx *tmapi.Context, entityID signature.PublicKey, runtimeID common.Namespace, churpID uint8) error {
+	regState := registryState.NewMutableState(ctx.State())
+
+	regParams, err := regState.ConsensusParameters(ctx)
+	if err != nil {
+		return err
+	}
+	if regParams.DebugBypassStake {
```
kostko, line 435: As you mentioned, this could probably be moved to staking parameters. Since it is a debug parameter it is not breaking as no debug parameters can be set in production networks. But we can refactor later in a separate PR.

go/consensus/cometbft/apps/keymanager/churp/txs_test.go
```diff
@@ -205,6 +215,26 @@ func (s *TxTestSuite) TestCreate() {
 		require.ErrorContains(s.T(), err, "invalid config: unsupported group, ID 100")
 	})
 
+	s.Run("not enough stake", func() {
```
kostko, line 218: Also add a test with sufficient stake and check that stake claims are created correctly.

go/keymanager/churp/api.go
```diff
@@ -62,6 +65,23 @@ var DefaultConsensusParameters = ConsensusParameters{
 	GasCosts: DefaultGasCosts,
 }
 
+const (
+	// StakeClaimCreate is the stake claim template used for creating
+	// new CHURP schemes.
+	StakeClaimCreate = "keymanager.churp.Create.%s.%d"
```
kostko, line 71: Probably should not be "Create" but rather "Instance" or "Scheme" or similar?

go/staking/api/api.go
```diff
@@ -742,6 +743,7 @@ const (
 	KindNodeKeyManagerName    = "node-keymanager"
 	KindRuntimeComputeName    = "runtime-compute"
 	KindRuntimeKeyManagerName = "runtime-keymanager"
+	KindKeyManagerChurpName   = "keymanager-churp"
 )
 
 // ThresholdKinds are the valid threshold kinds.
```
kostko, line 749: Add comment in place of `KindKeyManagerChurpName` explaining why it is not there.

kostko (commented):
Also update `e2e/genesis_file.go` scenario to ensure that the current genesis file is still valid (e.g. update the genesis file URL, set `genesisNeedsUpgrade` to `false` and `genesisDocumentHash` to the latest mainnet genesis document hash).

go/consensus/cometbft/apps/keymanager/churp/txs.go
kostko: Yeah. So in practice in those cases a failure would mean that something catastrophic has happened and the node would not be able to continue anyway. But it makes sense to be safe.

kostko (approved):

PR #5655, opened 2024-04-23
.github/dependabot: fix empty excludes

ptrus:
Wanted to keep empty `exclude-patterns` and `ignore` in the config, so that one can clearly see where these should be added if needed (e.g. for dependencies that will be harder to update/will require more changes and should be updated in separate PRs).

But:

> The property '#/updates/2/groups/rust/exclude-patterns' did not contain a minimum number of items 1

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -52,9 +52,11 @@ updates:
         patterns:
           - "*"
         # Excluded dependencies are updated in separate PRs.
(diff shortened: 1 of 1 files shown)


kostko (approved):
:grin:

peternose:
I would add a comment to the code, similar to what is written above, just shorter.

peternose (approved):

PR #5657, opened 2024-04-23
go/consensus: Add global minimum gas price

kostko:


go/consensus/pricediscovery/pricediscovery.go
```diff
@@ -0,0 +1,147 @@
...
+			pd.mu.Unlock()
+			return
+		}
+
+		pd.mu.Lock()
+		pd.computedGasPrice = maxPrice
+		pd.mu.Unlock()
+	}()
+
+	if len(pd.blockPrices) < int(pd.windowSize) {
```
ptrus, line 107: I know this comes from web3-gw, but we can initialize `blockPrices` to zeroes in `New` and then this conditional is not needed here.

go/consensus/pricediscovery/pricediscovery.go
```diff
@@ -0,0 +1,147 @@
...
+
+	pd.mu.Lock()
+	defer pd.mu.Unlock()
+
+	pd.minGasPrice = mgp
+}
+
+func (pd *priceDiscovery) processBlock(_ context.Context, _ *consensus.Block) {
+	// Currently transactions are not ordered by price, so track price as zero. After the mempool
+	// is refactored, change this to properly compute the median gas price.
```
ptrus, line 73: Once this is done, should we also add some small constant to the computed median price? (to ensure suggested price increases if blocks are full and median is also the minimum price).

Maybe just mention it in the todo comment here.

go/consensus/pricediscovery/pricediscovery.go
```diff
@@ -0,0 +1,161 @@
...
+	//
+	// We should also make sure to add some margin over the median in case of full blocks.
+	pd.trackPrice(quantity.NewQuantity())
+}
+
+func (pd *priceDiscovery) trackPrice(price *quantity.Quantity) {
+	// One item always gets added added to the prices array.
+	// Bump the current index for next iteration.
+	defer func() {
+		pd.blockPricesCurrentIdx = (pd.blockPricesCurrentIdx + 1) % int(pd.windowSize)
```
peternose, line 89:
```suggestion
		pd.blockPricesCurrentIdx = (pd.blockPricesCurrentIdx + 1) % int(len(pd.blockPrices))
```
You can get rid of this field.

go/consensus/pricediscovery/pricediscovery.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	pd.mu.Lock()
+	defer pd.mu.Unlock()
+
+	pd.minGasPrice = mgp
+}
+
+func (pd *priceDiscovery) processBlock(_ context.Context, _ *consensus.Block) {
+	// Currently transactions are not ordered by price, so track price as zero. After the mempool
```
peternose, line 78: Why don't you order them?

go/consensus/pricediscovery/pricediscovery.go
```diff
@@ -0,0 +1,161 @@
...
+
+	logger *logging.Logger
+}
+
+// GasPrice implements consensus.PriceDiscovery.
+func (pd *priceDiscovery) GasPrice(_ context.Context) (*quantity.Quantity, error) {
+	pd.mu.RLock()
+	defer pd.mu.RUnlock()
+
+	// Choose the maximum of (fallback, min, computed) gas prices.
```
peternose, line 51: This could also be computed after every block. Here, we would just serve the computed value.

go/consensus/pricediscovery/pricediscovery.go
```diff
@@ -0,0 +1,161 @@
+// Package pricediscovery implements dynamic price discovery.
+package pricediscovery
```
peternose, line 2: Should move interface, `staticPriceDiscovery` and `noOpPriceDiscovery` here?

go/consensus/pricediscovery/pricediscovery.go
peternose: If everyone used the suggested price, wouldn't the price increase indefinitely?

go/consensus/pricediscovery/pricediscovery.go
kostko: This requires the mempool to be changed. Sorting it only here wouldn't reflect queueing reality.

peternose: True. Maybe you should write that in the comment.

go/consensus/pricediscovery/pricediscovery.go
kostko: Only if blocks continue to be full (e.g. this is only applied in case the block is considered full, otherwise price is tracked as zero). Maybe this is not obvious now that the implementation is missing, but this logic is basically taken from oasis-web3-gateway.

go/consensus/pricediscovery/pricediscovery.go
```diff
@@ -0,0 +1,161 @@
...
+	// is refactored, change this to properly compute the median gas price.
+	//
+	// We should also make sure to add some margin over the median in case of full blocks.
+	pd.trackPrice(quantity.NewQuantity())
+}
+
+func (pd *priceDiscovery) trackPrice(price *quantity.Quantity) {
+	// One item always gets added added to the prices array.
+	// Bump the current index for next iteration.
+	defer func() {
```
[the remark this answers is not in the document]
kostko: Yeah probably not, this was taken from oasis-web3-gateway and after some changes I guess it could be refactored a bit.

go/consensus/pricediscovery/pricediscovery.go
kostko: Interface cannot be moved as it would then result in a circular import which is not supported in Go. I will move `staticPriceDiscovery` and `noOpPriceDiscovery` is only used by the no-op submission manager so it should just stay there.

go/consensus/pricediscovery/pricediscovery.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}
+
+	// No full blocks among last `windowSize` blocks.
+	if maxPrice.IsZero() {
+		maxPrice = nil
+	}
+	pd.computedGasPrice = maxPrice
+
+	pd.blockPrices[pd.blockPricesCurrentIdx] = price
```
peternose, line 91: These two lines should be moved to the top, as now you have a delay of 1 block.

peternose (approved):

go/consensus/pricediscovery/pricediscovery.go
kostko: :facepalm:

PR #5659, opened 2024-04-23
go/upgrade: Support upgrades without stopping the node

kostko:


go/upgrade/upgrade.go
```diff
@@ -279,8 +279,29 @@ func (u *upgradeManager) ConsensusUpgrade(privateCtx interface{}, currentEpoch b
...
+			// Check if we can proceed in place (e.g. without a restart).
+			u.shouldStop = func() bool {
+				if err := pu.Descriptor.EnsureCompatible(); err != nil {
+					// Not compatible, we must stop.
+					return true
+				}
+				// Check if the migration handler has a startup stage.
+				handler, err := migrations.GetHandler(pu.Descriptor.Handler)
+				if err != nil {
+					// Handler not available, we must stop.
```
peternose, line 291: Should we log this, as this is not expected?
I would also log "skipping startup upgrade".

go/upgrade/upgrade.go
```diff
@@ -279,8 +279,29 @@ func (u *upgradeManager) ConsensusUpgrade(privateCtx interface{}, currentEpoch b
 			if err := u.flushDescriptorLocked(); err != nil {
 				return err
 			}
-			u.shouldStop = true // Ensure we really stop before proceeding.
-			return api.ErrStopForUpgrade
+			// Check if we can proceed in place (e.g. without a restart).
+			u.shouldStop = func() bool {
+				if err := pu.Descriptor.EnsureCompatible(); err != nil {
```
peternose, line 284: Not part of this PR, but it looks like we can change the descriptor after it was proposed and accepted. Do we check that the fingerprint of the upgrade methods stayed the same? :thinking:

peternose (approved):

go/upgrade/upgrade.go
kostko: It could be expected as you can have an upgrade that does not change the protocol versions but only adds an upgrade handler. Will add a log entry for skipping startup upgrade though.

go/upgrade/upgrade.go
kostko: What do you mean "change the descriptor"? Upgrade descriptors are local (e.g. can be submitted even without governance as that is the mechanism that was present before there was governance support) and governance can just push them to all nodes.

In theory node operators could cancel these locally, but there is a fallback mechanism for this case where the consensus governance service would trigger a descriptor reinstall and node shutdown in case the upgrade did not happen. Without it there would be risk of state corruption.

PR #5662, opened 2024-04-25
go/vault: Add simple consensus layer vault

kostko:


go/consensus/cometbft/api/context.go
```diff
@@ -207,6 +207,14 @@ func (c *Context) CallerAddress() staking.Address {
 	return c.callerAddress
 }
 
+// CallDepth returns the call depth.
+func (c *Context) CallDepth() int {
+	if c.parent == nil {
+		return 0
+	}
+	return c.parent.CallDepth() + 1
```
ptrus, line 215: Could store parents call depth in the constructor, to avoid recursing through all parents, but probably doesn't really matter since the depth is limited to a low number.

go/vault/api/vault.go
```diff
@@ -0,0 +1,135 @@
...
+	return NewVaultAddress(v.CreatedBy, v.CreatedNonce)
+}
+
+// IsActive retruns true iff the vault is currently active (processing withdrawals).
+func (v *Vault) IsActive() bool {
+	return v.State == StateActive
+}
+
+// Authorities returns the list of all vault authorities.
+func (v *Vault) Authorities() []*Authority {
```
ptrus, line 55: Seems a bit weird putting authorities with different scopes into the same list, without any info about the scope. This should likely return a map or an object which would also contain info with the scope of each authority (e.g. admin, suspend). Or maybe even remove this method if it's rarely needed.

go/vault/api/vault.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// Verify checks whether the passed set of addresses is sufficient to authorize an action.
+func (a *Authority) Verify(addresses []staking.Address) bool {
+	addressMap := make(map[staking.Address]struct{})
+	for _, addr := range a.Addresses {
+		addressMap[addr] = struct{}{}
+	}
+	var count int
+	for _, addr := range addresses {
```
ptrus, line 114: This list needs to be deduplicated (or remove the address from `addressMap` on first match).

go/vault/api/vault.go
```diff
@@ -0,0 +1,135 @@
...
+	for _, addr := range a.Addresses {
+		if _, ok := addressMap[addr]; ok {
+			return fmt.Errorf("duplicate address in authority: %s", addr)
+		}
+		addressMap[addr] = struct{}{}
+	}
+	return nil
+}
+
+// Contains checks whether the set of authorities contains the given address.
```
ptrus, line 97:
```suggestion
// Contains checks whether the authority contains the given address.
```

go/vault/api/vault.go
ptrus: Also add a test for this case.

kostko: Duplicate addresses are rejected on validation, there is already a test case for this, but let me check it may only be in transaction tests.

kostko: But agree that it is better to do this in both places so it works correctly even in case validation was not performed.

kostko: Oh I see, you mean duplicate addresses in `addresses`, yeah let me fix that.

go/vault/api/vault.go
kostko: This is only used in a single place which is early rejection for action cancellations (when the action itself has not yet been fetched from the database so the exact authority is not yet available). We initially reject any callers that are not part of any authority and in the second phase (when action is available) we re-check and only accept the action-specific authority.

ptrus: Yeah maybe consider just explicitly getting both admin and suspend authorities in the list there. Or making this a private method and only used there. Or keep as it is.

Not a big deal, just seems a bit dangerous having a public `Authorities` method which then returns a mix of authorities with different scopes (and without the scope info).

go/consensus/cometbft/apps/vault/query.go
```diff
@@ -0,0 +1,67 @@
+package vault
```
ptrus, line 1: Could update txsource/workload/queries.go with the added queries.

Although these won't really test much unless there is also a vault workload added that populates the state. This should probably also be added, but can be done in a later PR.

go/vault/api/vault.go
kostko: The idea is that if a new authority is added, it would be updated in the same place (e.g. here) and the rest of the code would continue to function as before.

go/consensus/cometbft/apps/vault/transactions.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+	default:
+		return err
+	}
+
+	var removeAction bool
+	switch args.Action {
+	case nil:
+		// Cancel existing action.
+		removeAction = true
```
ptrus, line 146: Maybe canceling an existing action should also emit an event.

go/consensus/cometbft/apps/vault/transactions.go
```diff
@@ -0,0 +1,213 @@
...
+			Action:    *args.Action,
+		}
+	default:
+		return err
+	}
+
+	var removeAction bool
+	switch args.Action {
+	case nil:
+		// Cancel existing action.
```
ptrus, line 145: Does this mean that a single authority address can cancel a pending action, even if the authority threshold is >1? 

Could a single compromised/malicious authority address then prevent updates for a specific vault by constantly submitting cancel action transactions for any pending action?

kostko: Yes. The idea is that this is a safety feature so malicious/errorneous transactions are easier to abort. This is modeled after other existing multi-sig implementations, but it is indeed an availability tradeoff in case of malicious actors.

ptrus: Ok. Could a simple improvement be to at least require an Admin authority for cancelation? So that suspend authority (and any other added authority in future) cannot prevent Admin actions?

go/vault/api/action.go
```diff
@@ -0,0 +1,391 @@
...
+}
+
+// PrettyType returns a representation of ActionUpdateWithdrawPolicy that can be used for pretty
+// printing.
+func (au ActionUpdateWithdrawPolicy) PrettyType() (interface{}, error) {
+	return au, nil
+}
+
+// AuthorityKind is the kind of the authority being updated.
+type AuthorityKind uint8
```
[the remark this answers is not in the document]
kostko: Good idea, I'll change it.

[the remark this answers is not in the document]
kostko: Done.

go/consensus/cometbft/apps/vault/transactions.go
kostko: Added.

go/vault/api/policy.go
```diff
@@ -0,0 +1,88 @@
...
+	"io"
+
+	"github.com/oasisprotocol/oasis-core/go/common/quantity"
+	"github.com/oasisprotocol/oasis-core/go/staking/api/token"
+)
+
+// AddressState is the state stored for the given address.
+type AddressState struct {
+	// WithdrawPolicy is the active withdraw policy.
+	WithdrawPolicy WithdrawPolicy `json:"withdraw_policy"`
```
peternose, line 15: Shouldn't this policy be applied to the vault?

go/consensus/cometbft/apps/vault/action.go
```diff
@@ -0,0 +1,76 @@
...
+		addrState, err := state.AddressState(ctx, vlt.Address(), action.UpdateWithdrawPolicy.Address)
+		switch err {
+		case nil:
+		case vault.ErrNoSuchState:
+			addrState = &vault.AddressState{}
+		default:
+			return err
+		}
+
+		addrState.WithdrawPolicy = action.UpdateWithdrawPolicy.Policy
```
peternose, line 45: Bucket and amount could be set to 0 after update. Or only if `LimitInterval` changes.

go/vault/api/action.go
```diff
@@ -0,0 +1,378 @@
...
+
+// Authorities returns the authorities of the given vault that can authorize this action.
+func (au *ActionUpdateAuthority) Authorities(vault *Vault) []*Authority {
+	return []*Authority{
+		&vault.AdminAuthority,
+	}
+}
+
+// Apply applies the authority update to the given vault.
+func (au *ActionUpdateAuthority) Apply(vault *Vault) error {
```
peternose, line 351:
```suggestion
func (au *ActionUpdateAuthority) Apply(vault *Vault) {
```

go/consensus/cometbft/apps/vault/transactions.go
```diff
@@ -0,0 +1,218 @@
...
+			evExec := &vault.ActionExecutedEvent{
+				Vault: args.Vault,
+				Nonce: args.Nonce,
+			}
+			err = app.executeAction(ctx, vlt, &pendingAction.Action)
+			switch {
+			case api.IsUnavailableStateError(err):
+				// Propagate state unavailability errors.
+				return err
+			default:
```
peternose, line 188:
```suggestion
			case pendingAction.Action.ExecuteMessage == nil:
				return err
			default:
```

go/vault/api/policy.go
kostko: What do you mean? This is a per-address policy for the given vault.

go/consensus/cometbft/apps/vault/state/state.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// ConsensusParameters returns the vault consensus parameters.
+func (s *ImmutableState) ConsensusParameters(ctx context.Context) (*vault.ConsensusParameters, error) {
+	raw, err := s.is.Get(ctx, parametersKeyFmt.Encode())
+	if err != nil {
+		return nil, api.UnavailableStateError(err)
+	}
+	if raw == nil {
+		return &vault.ConsensusParameters{}, nil
```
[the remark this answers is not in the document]
kostko: As long as we want to allow sync from genesis, we cannot just remove/return an error as we need to load the `enabled` flag when the parameters are not set.

go/vault/api/policy.go
peternose: Do we want to have a similar policy for vault? So how much can all addresses withdraw together in some time period.

go/consensus/cometbft/apps/vault/action.go
```diff
@@ -0,0 +1,76 @@
...
+// executeAction executes a given action in the context of a vault. Assumes the action has already
+// been validated before execution.
+func (app *vaultApplication) executeAction(ctx *api.Context, vlt *vault.Vault, action *vault.Action) error {
+	switch {
+	case action.Suspend != nil:
+		// Suspend a vault.
+		oldState := vlt.State
+		vlt.State = vault.StateSuspended
+
+		ctx.EmitEvent(api.NewEventBuilder(app.Name()).TypedAttribute(&vault.StateChangedEvent{
```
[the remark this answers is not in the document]
kostko: Good point, we should only emit if the state actually changes.

go/consensus/cometbft/apps/vault/state/state.go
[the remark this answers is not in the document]
peternose: We could have something above all the apps, which would tell you which apps are enabled. But I guess this approach is also fine. So from now on, all new apps will return empty parameters with `enabled` flag set to `false`.

go/vault/api/vault.go
```diff
@@ -0,0 +1,137 @@
...
+			return fmt.Errorf("duplicate address in authority: %s", addr)
+		}
+		addressMap[addr] = struct{}{}
+	}
+	return nil
+}
+
+// Contains checks whether the authority contains the given address.
+func (a *Authority) Contains(address staking.Address) bool {
+	for _, addr := range a.Addresses {
```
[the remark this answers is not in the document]
kostko: Thought about it, but serialization is not the nicest. There should never be so many addresses that this is a problem (e.g. currently limited to 32 addresses max).

go/vault/api/action.go
```diff
@@ -0,0 +1,378 @@
...
+			return err
+		}
+	}
+	if a.UpdateAuthority != nil {
+		actions++
+		if err := a.UpdateAuthority.Validate(params); err != nil {
+			return err
+		}
+	}
+	if actions != 1 {
```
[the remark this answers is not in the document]
kostko: So you would do the checks twice? The frontend should be made such that you can't even construct actions that have multiple actions set.

go/vault/api/methods.go
```diff
@@ -0,0 +1,124 @@
...
+	return c, nil
+}
+
+// AuthorizeAction is an action authorization call body.
+type AuthorizeAction struct {
+	// Vault is the address of the target vault.
+	Vault staking.Address `json:"vault"`
+	// Nonce is the action nonce.
+	Nonce uint64 `json:"nonce"`
+	// Action is the action that should be authorized. If the action is not set, this means that the
```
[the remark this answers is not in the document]
kostko: I tried to minimize the amount of methods, but sure that would indeed be cleaner.

go/vault/api/vault.go
```diff
@@ -0,0 +1,137 @@
...
+
+const (
+	StateActive    = 0
+	StateSuspended = 1
+)
+
+// Vault contains metadata about a vault.
+type Vault struct {
+	// CreatedBy is the address of the vault creator.
+	CreatedBy staking.Address `json:"created_by"`
```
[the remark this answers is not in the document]
kostko: Owner implies special privileges and this is not the case here (e.g. this field is immutable). Sure "creator" works as well.

go/vault/api/vault.go
```diff
@@ -0,0 +1,137 @@
...
+	// CreatedNonce is the nonce of the vault creator's account at time of creation.
+	CreatedNonce uint64 `json:"created_nonce"`
```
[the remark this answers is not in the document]
kostko: Yes the initial implementation had autoincrementing identifiers, but this is bad because all identifiers are predictable in advance. We could also go with block height and creation transaction index, but that lacks entropy. We could add a salt but that is awkward. I think the creator address and nonce works nicely.

go/consensus/cometbft/apps/vault/action.go
```diff
@@ -0,0 +1,76 @@
...
+			return err
+		}
+
+		ctx.EmitEvent(api.NewEventBuilder(app.Name()).TypedAttribute(&vault.PolicyUpdatedEvent{
+			Vault:   vlt.Address(),
+			Address: action.UpdateWithdrawPolicy.Address,
+		}))
+	case action.UpdateAuthority != nil:
+		// Update a vault authority.
+		err := action.UpdateAuthority.Apply(vlt)
```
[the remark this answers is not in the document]
kostko: Hm I guess in this case the last `AuthorizeAction` transaction would be invalid as well (since it needs to include the action which is validated), so the only way to get rid of the action would be to cancel it. But it makes sense to explicitly validate again.

go/vault/api/policy.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// CurrentBucket specifies the interval we are currently doing accounting for.
+	CurrentBucket uint64 `json:"bucket"`
+	// CurrentAmount specifies the amount already withdrawn in the current interval.
+	CurrentAmount quantity.Quantity `json:"amount"`
+}
+
+// AuthorizeWithdrawal performs withdrawal authorization. In case withdrawal is allowed, the state
+// is also updated to reflect the additional withdrawal.
+func (as *AddressState) AuthorizeWithdrawal(height int64, amount *quantity.Quantity) bool {
```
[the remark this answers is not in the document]
kostko: Hm, but `Withdrawal` makes it sound as if it also does the withdrawal itself? But this does not move any tokens.

go/vault/api/events.go
```diff
@@ -0,0 +1,107 @@
...
+	Vault staking.Address `json:"vault"`
+	// OldState is the old vault state.
+	OldState State `json:"old_state"`
+	// NewState is the new vault state.
+	NewState State `json:"new_state"`
+}
+
+// EventKind returns a string representation of this event's kind.
+func (e *StateChangedEvent) EventKind() string {
+	return "suspended"
```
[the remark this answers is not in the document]
kostko: Ooops, leftover :)

go/vault/api/action.go
```diff
@@ -0,0 +1,378 @@
...
+//
+// It should only be used for pretty printing.
+type PrettyActionExecuteMessage struct {
+	Method transaction.MethodName `json:"method"`
+	Body   interface{}            `json:"body,omitempty"`
+}
+
+// ActionUpdateWithdrawPolicy is the action to update the withdraw policy for a given address.
+type ActionUpdateWithdrawPolicy struct {
+	// Address is the address the policy update is for. A nil address stands for the default
```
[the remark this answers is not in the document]
kostko: Not anymore, initial version allowed a catch-all policy, but I don't think this is a good idea.

go/consensus/cometbft/apps/vault/transactions.go
```diff
@@ -0,0 +1,218 @@
...
+			pendingAction.AuthorizedBy = append(pendingAction.AuthorizedBy, ctx.CallerAddress())
+			if err = state.SetPendingAction(ctx, args.Vault, pendingAction); err != nil {
+				return err
+			}
+		}
+
+		// Check if action has become executable.
+		var canExecute bool
+		for _, auth := range pendingAction.Action.Authorities(vlt) {
+			if auth.Verify(pendingAction.AuthorizedBy) {
```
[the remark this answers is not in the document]
kostko: Currently there can only be a single pending action due to the action nonce check (this is on purpose to keep the initial implementation simple).

go/consensus/cometbft/apps/vault/transactions.go
```diff
@@ -0,0 +1,218 @@
...
+	case vault.ErrNoSuchAction:
+		// In case there is no existing action, this must not be a cancellation.
+		if args.Action == nil {
+			return vault.ErrInvalidArgument
+		}
+
+		// Create a new pending action.
+		pendingAction = &vault.PendingAction{
+			Nonce:     args.Nonce,
+			CreatedAt: ctx.BlockHeight() + 1, // Current height is ctx.BlockHeight() + 1.
```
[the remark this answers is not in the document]
kostko: I would not add timeouts for now, this is mostly for easier introspection. Could also be removed.

go/consensus/cometbft/apps/vault/transactions.go
```diff
@@ -0,0 +1,218 @@
...
+			ctx.EmitEvent(api.NewEventBuilder(app.Name()).TypedAttribute(evExec))
+		}
+	}
+
+	if removeAction {
+		if err = state.RemovePendingAction(ctx, args.Vault, args.Nonce); err != nil {
+			return err
+		}
+
+		vlt.Nonce++
```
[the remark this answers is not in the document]
kostko: It seems safer to always increase the nonce so that any signed messages are invalidated.

go/consensus/cometbft/apps/vault/transactions.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	// Ensure vault exists.
+	vlt, err := state.Vault(ctx, args.Vault)
+	if err != nil {
+		return err
+	}
+
+	// Validate action nonce. Currently queuing multiple future actions is not allowed.
+	if args.Nonce != vlt.Nonce {
```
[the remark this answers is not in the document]
kostko: Not strictly needed, but it seems safer to have it to ensure that authorized messages are unique and cannot be replayed. It is true that these are currently sent as part of transactions which have their own nonces, but having independent nonces seems easier to reason about.

go/vault/api/action.go
```diff
@@ -0,0 +1,378 @@
...
+// Equal returns true iff one action is equal to another.
+func (a *Action) Equal(other *Action) bool {
+	return bytes.Equal(cbor.Marshal(a), cbor.Marshal(other))
+}
+
+// Authorities returns the authorities of the given vault that can authorize this action.
+func (a *Action) Authorities(vault *Vault) []*Authority {
+	switch {
+	case a == nil:
+		// Any authority can cancel an action when the action itself is not known. In the second
```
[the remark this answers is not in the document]
kostko: No longer needed as I added an explicit `CancelAction` method.

go/vault/api/methods.go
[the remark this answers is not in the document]
kostko: Done.

go/vault/api/action.go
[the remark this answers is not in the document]
peternose: I would first count and then validate, for the reason stated above (or return an error as soon as count is greater than 1). Currently, we can validate all actions, just to figure out that there are too many actions.

go/vault/api/vault.go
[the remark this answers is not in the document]
peternose: Oh, now I understand why you used metadata in comments. Well, I would still not expect this metadata in the `Vault` struct, just fields `Address`, `Nonce`, `State`, ...

go/vault/api/policy.go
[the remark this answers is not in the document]
peternose: You are right, tokens are moved latter. But the state is updated, though. Since I have no alternative name, you can leave it as it is.

go/vault/api/api.go
```diff
@@ -0,0 +1,135 @@
...
+	// Cleanup cleans up the backend.
+	Cleanup()
+}
+
+// VaultQuery is a query for data about a given vault.
+type VaultQuery struct {
+	// Height is the query height.
+	Height int64 `json:"height"`
+	// Address is the vault address.
+	Address staking.Address `json:"address"`
```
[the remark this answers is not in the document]
peternose: I just wanted to be consistent with other queries, but you are probably right. You can ignore the comment.

go/consensus/cometbft/apps/vault/transactions.go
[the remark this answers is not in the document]
peternose: At that point I was not sure about that. Resolving.

go/consensus/cometbft/apps/vault/transactions.go
```diff
@@ -0,0 +1,218 @@
...
+	"fmt"
+
+	"github.com/oasisprotocol/oasis-core/go/common/errors"
+	"github.com/oasisprotocol/oasis-core/go/consensus/cometbft/api"
+	stakingState "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/apps/staking/state"
+	vaultState "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/apps/vault/state"
+	vault "github.com/oasisprotocol/oasis-core/go/vault/api"
+)
+
+func (app *vaultApplication) create(ctx *api.Context, args *vault.Create) error {
```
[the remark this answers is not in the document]
peternose: I suggested that because `authorizeAction` is a long method and at the end I had no idea where `args.Nonce` comes from. I checked `staking` app, and there you have names like `burn *staking.Burn`. Name `args` is just so general, that I would try to avoid it. Maybe `req` should do it, not sure, you decide.

go/consensus/cometbft/apps/vault/state/state.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// CreateVault creates a new vault.
+//
+// NOTE: This operation performs multiple actions so it should be wrapped in a transaction.
+func (s *MutableState) CreateVault(ctx context.Context, vlt *vault.Vault) error {
+	addr := vlt.Address()
+
+	// Sanity check to make sure that the vault doesn't already exist. Note that this is not
+	// possible during regular operation due to nonce advancing.
```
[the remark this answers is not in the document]
peternose: True. Unfortunately, this change just made a dependency between staking state and vault state, and added an extra responsibility to vault state, i.e. to update staking state. You can leave it as it is, and we can refactor latter if we encounter cyclic dependency.

go/consensus/cometbft/apps/vault/transactions.go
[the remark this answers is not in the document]
peternose: OK. I just thought that since events already have block height (or at least you know when they were emitted), this is not needed.

go/consensus/cometbft/apps/vault/transactions.go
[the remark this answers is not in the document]
peternose: Good idea (could be added as a comment).

go/consensus/cometbft/apps/vault/transactions.go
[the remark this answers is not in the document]
kostko: Removed.

go/vault/api/action.go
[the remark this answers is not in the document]
kostko: Done.

go/consensus/cometbft/apps/vault/vault.go
```diff
@@ -0,0 +1,122 @@
...
+			return vault.ErrInvalidArgument
+		}
+		return app.authorizeAction(ctx, &args)
+	default:
+		return vault.ErrInvalidArgument
+	}
+}
+
+func (app *vaultApplication) ExecuteMessage(ctx *api.Context, kind, msg interface{}) (interface{}, error) {
+	if enabled, _ := app.Enabled(ctx); !enabled {
```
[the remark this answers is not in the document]
kostko: Good idea, added.

go/vault/api/vault.go
```diff
@@ -0,0 +1,137 @@
...
+	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
+)
+
+var _ prettyprint.PrettyPrinter = (*Authority)(nil)
+
+// State is the vault state.
+type State uint8
+
+const (
+	StateActive    = 0
```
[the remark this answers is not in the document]
kostko: Changed.

go/vault/api/policy.go
kostko: Let's keep it simple for now.

peternose: I would say that having a policy for a vault is simpler than having a policy for every address that can withdraw from it.
But yeah, we can keep it as it is.

kostko: That's true, but we do want to have a list of addresses that are allowed to withdraw anything at all, so per-address state is needed anyway.

go/consensus/cometbft/apps/vault/query.go
kostko: Yeah let's do this in a later PR.

ptrus (approved):

go/consensus/cometbft/apps/vault/transactions.go
```diff
@@ -0,0 +1,218 @@
...
+	defer ctx.Close()
+
+	var pendingAction *vault.PendingAction
+	// Check if there is an existing pending action for this nonce. In this case, we will be
+	// updating the action.
+	pendingAction, err = state.PendingAction(ctx, args.Vault, args.Nonce)
+	switch err {
+	case nil:
+		// Pending action already exists.
+		if !pendingAction.Action.IsAuthorized(vlt, ctx.CallerAddress()) {
```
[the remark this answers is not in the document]
peternose: Why was this not removed?

go/consensus/cometbft/apps/vault/transactions.go
```diff
@@ -0,0 +1,277 @@
...
+	if ctx.IsCheckOnly() {
+		return nil
+	}
+
+	// Charge gas for this transaction.
+	params, err := state.ConsensusParameters(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to fetch consensus parameters: %w", err)
+	}
+	if err = ctx.Gas().UseGas(1, vault.GasOpAuthorizeAction, params.GasCosts); err != nil {
```
peternose, line 236:
```suggestion
	if err = ctx.Gas().UseGas(1, vault.GasOpCancelAction, params.GasCosts); err != nil {
```
Cancel actions could also charge lower gas.

go/consensus/cometbft/apps/vault/vault.go
```diff
@@ -0,0 +1,124 @@
...
+	governanceApi "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/apps/governance/api"
+	stakingapp "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/apps/staking"
+	stakingApi "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/apps/staking/api"
+	vaultState "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/apps/vault/state"
+	vault "github.com/oasisprotocol/oasis-core/go/vault/api"
+)
+
+var (
+	_ api.Application          = (*vaultApplication)(nil)
+	_ api.TogglableApplication = (*vaultApplication)(nil)
```
peternose, line 18:
```suggestion
	_ api.TogglableApplication       = (*vaultApplication)(nil)
	_ api.TogglableMessageSubscriber = (*vaultApplication)(nil)
```

go/consensus/cometbft/apps/vault/vault.go
```diff
@@ -0,0 +1,124 @@
...
+func (app *vaultApplication) ID() uint8 {
+	return AppID
+}
+
+func (app *vaultApplication) Methods() []transaction.MethodName {
+	return vault.Methods
+}
+
+// Enabled implements api.TogglableApplication and api.TogglableMessageSubscriber.
+func (app *vaultApplication) Enabled(ctx *api.Context) (bool, error) {
```
peternose, line 39: I asked already, but will ask again. Is it possible to remove returned `error`, update `ConsensusParameters` to return a special error if parameters are not set, handle this error here and return `false` in that case?

go/vault/api/action.go
```diff
@@ -0,0 +1,360 @@
...
+	// UpdateWithdrawPolicy is the withdraw policy update action.
+	UpdateWithdrawPolicy *ActionUpdateWithdrawPolicy `json:"update_withdraw_policy,omitempty"`
+	// UpdateAuthority is the authority update action.
+	UpdateAuthority *ActionUpdateAuthority `json:"update_authority,omitempty"`
+}
+
+// Validate validates the given action.
+func (a *Action) Validate(params *ConsensusParameters) error {
+	// Ensure there is at most one action set.
+	if !common.ExactlyOneTrue(
```
peternose, line 54: Very elegant, forgot that we have this.

go/vault/api/action.go
```diff
@@ -0,0 +1,360 @@
...
+		err = a.UpdateWithdrawPolicy.Validate()
+	case a.UpdateAuthority != nil:
+		err = a.UpdateAuthority.Validate(params)
+	}
+	return err
+}
+
+// Equal returns true iff one action is equal to another.
+func (a *Action) Equal(other *Action) bool {
+	return bytes.Equal(cbor.Marshal(a), cbor.Marshal(other))
```
peternose, line 79: Off-topic: It would be nice to have a method `cbor.Equal(a, b)` which could return as soon as encodings differ.

go/vault/api/api.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// ConsensusParameters are the vault consensus parameters.
+type ConsensusParameters struct {
+	// Enabled specifies whether the vault service is enabled.
+	Enabled bool `json:"enabled,omitempty"`
+
+	// MaxAuthorityAddresses is the maximum number of addresses that can be configured for each
+	// authority.
+	MaxAuthorityAddresses uint16 `json:"max_authority_addresses,omitempty"`
```
peternose, line 103: Is `uint8` too small?

go/vault/api/action.go
```diff
@@ -0,0 +1,360 @@
...
+	"github.com/oasisprotocol/oasis-core/go/common"
+	"github.com/oasisprotocol/oasis-core/go/common/cbor"
+	"github.com/oasisprotocol/oasis-core/go/common/prettyprint"
+	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"
+	staking "github.com/oasisprotocol/oasis-core/go/staking/api"
+)
+
+var _ prettyprint.PrettyPrinter = (*Action)(nil)
+
+// PendingAction is a pending action.
```
peternose, line 22:
```suggestion
// PendingAction is an action waiting to be ...
```

peternose (approved):

go/consensus/cometbft/apps/vault/vault.go
kostko: But the error could also be a "state unavailable" error which needs to be propagated. If we get rid of the `error` return, then such propagation would not be possible or we would need to panic here which is not ideal as the caller should decide how to handle that.

peternose: Ok. Resolving.

go/vault/api/api.go
kostko: It is not `uint8` but rather `uint16`, that should be enough?

peternose: Wanted to ask why we don't use `uint8` (same as threshold). Is it because it's too small? Maybe it is, but update action would still need to set at 256 addresses, which is 21*256 = ~5kB.

kostko: Ah yes, I think `uint8` should be fine. The current configuration is `32` anyway.

go/consensus/cometbft/apps/vault/transactions.go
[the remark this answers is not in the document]
kostko: Whoops, this was missed, should be removed now.

go/vault/api/api.go
kostko: Changed to `uint8`.

peternose (approved):

PR #5668, opened 2024-04-30
.github/dependabot: group core and tests/upgrade/post

ptrus:
I believe this should make it so that oasis-core and tests/upgrade/post/ projects dependencies will be updated in same PR.

Using the new public beta feature: https://github.blog/changelog/2024-04-29-dependabot-multi-directory-configuration-public-beta-now-available/


kostko (approved):

peternose:
We could also disable or delete `tests/upgrade/post` and `tests/upgrade/post` since this code is specific to Eden upgrade and we are not planing to have dump restore upgrades anymore.

peternose (approved):

ptrus:
>  We could also disable or delete tests/upgrade/pre and tests/upgrade/post since this code is specific to Eden upgrade and we are not planing to have dump restore upgrades anymore.

Lets keep them for now, and try just disabling dependeabot for them. Feel free to remove in a separate PR.

For `test/upgrade/pre` and `test/upgrade/post`, we currently don't have dependabot configured. However dependabot still opens [security updates](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates), these can also be configured via dependabot.yml.  I believe the added configuration should now prevent dependabot to open any updates for these projects. But i'm not 100% sure, lets try it out.

kostko (approved):

PR #5689, opened 2024-05-08
secret-sharing/src/churp: Rename player to shareholder

peternose:


secret-sharing/src/churp/switch.rs
```diff
@@ -595,9 +596,9 @@ where
         Ok(done)
     }
 
-    /// Proactivizes the player with the combined polynomial and verification
-    /// matrix.
-    fn proactivize_player(&mut self) -> Result<Player<S>> {
+    /// Proactivizes the Shareholder with the combined polynomial
```
kostko, line 599:
```suggestion
    /// Proactivizes the shareholder with the combined polynomial
```

kostko (approved):

PR #5703, opened 2024-05-24
keymanager: Increase MAX_FRESH_HEIGHT_AGE

kostko:
This avoids issues where key managers get out of sync during operation and start rejecting otherwise valid requests. It should still be safe as especially since all new runtimes now perform freshness checks.

### DIFF
--- a/keymanager/src/runtime/secrets.rs
+++ b/keymanager/src/runtime/secrets.rs
@@ -73,8 +73,8 @@ const MAX_EPHEMERAL_KEY_AGE: EpochTime = 10;
 /// Maximum age of a fresh height in the number of blocks.
 ///
 /// A height is considered fresh if it is not more than specified amount
-/// of blocks lower than the height of the latest trust root.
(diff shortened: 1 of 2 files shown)


keymanager/src/runtime/secrets.rs
```diff
@@ -73,8 +73,8 @@ const MAX_EPHEMERAL_KEY_AGE: EpochTime = 10;
 /// Maximum age of a fresh height in the number of blocks.
 ///
 /// A height is considered fresh if it is not more than specified amount
-/// of blocks lower than the height of the latest trust root.
-const MAX_FRESH_HEIGHT_AGE: u64 = 50;
+/// of blocks lower than the height of the latest verified height.
+const MAX_FRESH_HEIGHT_AGE: u64 = 1200;
```
peternose, line 77:
```suggestion
const MAX_FRESH_HEIGHT_AGE: u64 = 600; // ~1 hr @ 6s/block
```
I would add a comment. Not sure about the age, but 2h seems a lot.

peternose (approved):

PR #5704, opened 2024-05-24
Reverse TCB fetch order

jberci:


go/runtime/host/sgx/epid.go
```diff
@@ -135,8 +137,8 @@ func (ep *teeStateEPID) update(
 	// Decode the AVR so we can do further checks.
 	avr, decErr := cmnIAS.UnsafeDecodeAVR(avrBundle.Body)
 	if decErr == nil && avr.TCBEvaluationDataNumber < quotePolicy.MinTCBEvaluationDataNumber {
```
kostko, line 139: This condition needs to be changed as early TCB updates will always have higher or equal `TCBEvaluationDataNumber`. Instead we should probably check that `avr.ISVEnclaveQuoteStatus` is not one of `QuoteOK` or `QuoteSwHardeningNeeded` and only in this case fall back to "without early updating".

go/runtime/host/sgx/ecdsa.go
```diff
@@ -193,7 +193,7 @@ func (ec *teeStateECDSA) Update(ctx context.Context, sp *sgxProvisioner, conn pr
 		return fresh, nil
 	}
 	var tcbBundle *pcs.TCBBundle
-	for _, update := range []pcs.UpdateType{pcs.UpdateStandard, pcs.UpdateEarly} {
+	for _, update := range []pcs.UpdateType{pcs.UpdateEarly, pcs.UpdateStandard} {
```
peternose, line 196: Should `tcbCache` distinguish between standard and early TCB info?

go/runtime/host/sgx/epid.go
```diff
@@ -124,22 +124,24 @@ func (ep *teeStateEPID) update(
...
-		return nil, fmt.Errorf("error while verifying attestation evidence: %w", err)
+		return nil, fmt.Errorf("error while verifying attestation evidence with early update: %w", err)
 	}
 
 	// Decode the AVR so we can do further checks.
 	avr, decErr := cmnIAS.UnsafeDecodeAVR(avrBundle.Body)
-	if decErr == nil && avr.TCBEvaluationDataNumber < quotePolicy.MinTCBEvaluationDataNumber {
-		// Retry again with early updating.
-		evidence.EarlyTCBUpdate = true
+	if decErr == nil && avr.ISVEnclaveQuoteStatus != cmnIAS.QuoteOK && avr.ISVEnclaveQuoteStatus != cmnIAS.QuoteSwHardeningNeeded {
```
peternose, line 139: Is the last part really needed? It looks like this status is not returned anymore in v5.

go/runtime/host/sgx/epid.go
```diff
@@ -124,22 +124,24 @@ func (ep *teeStateEPID) update(
...
+		// Retry again without early updating.
+		evidence.EarlyTCBUpdate = false
 		avrBundle, err = iasClient.VerifyEvidence(ctx, &evidence)
 		if err != nil {
-			return nil, fmt.Errorf("error while verifying attestation evidence with early update: %w", err)
+			return nil, fmt.Errorf("error while verifying attestation evidence with normal update: %w", err)
 		}
 		avr, decErr = cmnIAS.UnsafeDecodeAVR(avrBundle.Body)
```
peternose, line 146: Should we also check that `avr.ISVEnclaveQuoteStatus == cmnIAS.QuoteOK`?

go/runtime/host/sgx/ecdsa.go
kostko: I don't think this is needed. Eventually a more advanced cache will be on-chain.

go/runtime/host/sgx/epid.go
kostko: Hm where did you see that? That is not the case AFAIK.

go/runtime/host/sgx/epid.go
kostko: We could, but this is also checked below by the runtime.

go/runtime/host/sgx/epid.go
peternose: p19 in https://api.trustedservices.intel.com/documents/IAS-API-Spec-rev-5.0.pdf

kostko: This is describing an old version (v3) of the API?

peternose: You are right :facepalm:

peternose (approved):

PR #5708, opened 2024-05-30
registry: update proveFreshness tx body

Andrew7234:
The body of the ProveFreshness tx [seems to be outdated](https://github.com/oasisprotocol/oasis-core/blob/master/go/consensus/cometbft/apps/registry/registry.go#L138), this PR switches the type to the [32]byte blob. 

Nexus currently uses a [workaround](https://github.com/oasisprotocol/nexus/blob/main/analyzer/consensus/convert_tx.go#L104) that deserializes the tx body into a custom type.

### DIFF
--- a/.changelog/5708.bugfix.md
+++ b/.changelog/5708.bugfix.md
@@ -1,0 +1,1 @@
+go/registry/api: update ProveFreshness tx body type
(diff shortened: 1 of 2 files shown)


pro-wh (approved):
ya this looks like it was mistakenly copied from above. everything in #4916 expects [32]byte

peternose (approved):

kostko (approved):

PR #5715, opened 2024-06-05
keymanager/src/churp: Fetch key shares and recover key

peternose:
Updates the key manager client to fetch key shares from shareholders and reconstruct the secret key, from which the state key can be derived.


keymanager/src/client/remote.rs
```diff
@@ -584,4 +596,102 @@ impl KeyManagerClient for RemoteClient {
...
+        salt[NAMESPACE_SIZE] = status.id;
+        salt[NAMESPACE_SIZE + 1..].copy_from_slice(&key_id.0);
+
+        let state_key = match status.suite_id {
+            SuiteId::NistP384Sha3_384 => {
+                let mut shares = Vec::with_capacity(min_shares);
+                for encoded_share in encoded_shares {
+                    let share = encoded_share
+                        .try_into()
+                        .map_err(|err: Error| KeyManagerError::Other(err.into()))?;
```
kostko, line 675: This should probably done while fetching shares to make sure enough shares are fetched in case some cannot be decoded?

runtime/src/consensus/state/beacon.rs
```diff
@@ -87,6 +89,26 @@ impl MutableState {
     }
 }
 
+/// Consensus beacon state.
```
kostko, line 92: What is the difference between this and `ImmutableState`? One is for low-level state access and the other one is a more high-level API? Why not just merge the two, to avoid the need to refactor other parts?

peternose: Wanted to keep `ImmutableState` clean as possible (only low lever operations). Will merge.

keymanager/src/client/remote.rs
```diff
@@ -584,4 +596,102 @@ impl KeyManagerClient for RemoteClient {
...
+
+        let mut committee = status.committee;
+        committee.shuffle(&mut OsRng);
+
+        let kind = HandoffKind::CommitteeUnchanged;
+        let player = Player::new(status.threshold, kind);
+        let min_shares = player.min_shares();
+        let mut encoded_shares = Vec::with_capacity(min_shares);
+
+        // TODO: Optimize by fetching key shares concurrently.
```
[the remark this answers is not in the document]
peternose: I think that two things:
- Peer feedback as you already mentioned.
- Support for multiple sessions with different peers+enclaves (currently only the server side handles multiple sessions, but not the client).

An ugly way to solve the last problem would be to create multiple clients. Which is not the best, as it would be nice to have a global client so that sessions can be shared if key manager methods are called from different places. And because of that, the session builder would need refactoring, as connections to the same node could use different policies and enclaves for session establishment.

keymanager/src/client/remote.rs
```diff
@@ -584,4 +596,102 @@ impl KeyManagerClient for RemoteClient {
             .await
             .map_err(|err| KeyManagerError::Other(err.into()))
     }
+
+    async fn state_key(
```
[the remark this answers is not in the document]
peternose: Wanted to exclude the prefix so that method names remain the same if we split this client into two parts, one for secrets and one for churp. Will add prefix for now.

go/consensus/cometbft/apps/keymanager/churp/txs.go
```diff
@@ -545,3 +556,20 @@ func resetHandoff(status *churp.Status, nextHandoff beacon.EpochTime) {
...
 }
+
+func verifyPolicy(ctx *tmapi.Context, policy *churp.SignedPolicySGX) error {
+	// Allow non-empty `MayQuery` field with the 24.2 release.
+	regState := registryState.NewMutableState(ctx.State())
+	regParams, err := regState.ConsensusParameters(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to load registry consensus parameters: %w", err)
+	}
+	if regParams.SoftwareVersion != nil && regParams.SoftwareVersion.ToU64() >= version.MustFromString("24.2").ToU64() {
```
kostko, line 567: Can we have some sort of helper for this? So we can do something like `IsFeatureVersion(ctx, "24.2")` and it would do all the required things? Also rename the field to something like `FeatureVersion` and move to root consensus parameters.

go/worker/keymanager/churp.go
```diff
@@ -130,7 +134,25 @@ func (w *churpWorker) Authorize(_ context.Context, method string, kind enclaverp
...
+		churp.RPCMethodShareDistributionPoint,
+		churp.RPCMethodBivariateShare:
+		return w.authorizeKeyManager(peerID)
+	default:
+		return fmt.Errorf("unsupported method: %s", method)
+	}
+}
+
+func (w *churpWorker) authorizeNode(_ context.Context, peerID core.PeerID) error {
+	// TODO: Which nodes are allowed to query key shares?
```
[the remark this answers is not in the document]
peternose: I refactored domain separation used by CHURP. The context/dst used to encode a `key_id` is now:
- `oasis-core/keymanager/churp: encode SGX policy key ID for runtime <kmID> for churp <ID>` for the first bullet, and
- `oasis-core/keymanager/churp: encode custom policy key ID for runtime <kmID> for churp <ID>` for the second one.

This will encode/hash `key_id` to a different group element, thus creating different shares. Chain context is not taken into account, but maybe it should be? Since churp ID uniquely defines the group and the hash function, we don't need to include that in the domain separation. 

Access policy (executor + observer) is checked on the host node, just like it's implemented for the `secrets`. Key manager runtime only checks that remote enclave is allowed to query key shares for the given runtime ID.

keymanager/src/client/remote.rs
```diff
@@ -218,6 +239,85 @@ impl RemoteClient {
...
+    ) -> Result<StateKey, KeyManagerError> {
+        // Fault detection and blame assignment are not supported,
+        // so the minimal number of key shares will suffice.
+        let kind = HandoffKind::CommitteeUnchanged;
+        let player = Player::new(status.threshold, kind);
+        let min_shares = player.min_shares();
+        let mut shares = Vec::with_capacity(min_shares);
+
+        // Fetch key shares in random order.
+        // TODO: Optimize by fetching key shares concurrently.
```
kostko, line 256: Open an issue to track this.

keymanager/src/client/remote.rs
```diff
@@ -218,6 +239,85 @@ impl RemoteClient {
         key.verify(self.runtime_id, key_pair_id, epoch, now, pk)
             .map_err(KeyManagerError::InvalidSignature)
     }
+
+    async fn churp_recover_state_key<S: Suite>(
+        &self,
+        key_id: KeyPairId,
+        status: churp::Status,
+    ) -> Result<StateKey, KeyManagerError> {
+        // Fault detection and blame assignment are not supported,
```
kostko, line 248: Is this planned? If so, open an issue to track this.

tests/runtimes/simple-keyvalue/src/methods.rs
```diff
@@ -271,23 +271,43 @@ impl Methods {
...
+        key: &[u8],
+        churp_id: u8,
+    ) -> Result<StateKey, String> {
+        // Derive key ID based on key.
+        let key_id = KeyPairId::from(Hash::digest_bytes(key).as_ref());
+
+        // Fetch encryption key.
+        let result = ctx.parent.key_manager.churp_state_key(churp_id, key_id);
+        let state_key = tokio::runtime::Handle::current()
+            .block_on(result)
```
kostko, line 289: Could use `block_on` from `oasis_core_runtime::future`.

go/worker/keymanager/churp.go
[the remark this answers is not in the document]
kostko: > Chain context is not taken into account, but maybe it should be?

What would happen during a possible dump/restore upgrade, would this invalidate all of the keys? We need to be careful.

> Access policy (executor + observer) is checked on the host node, just like it's implemented for the secrets. Key manager runtime only checks that remote enclave is allowed to query key shares for the given runtime ID.

Yeah checks for roles inside the runtime can be added in the future when everyone is using observer nodes.

secret-sharing/src/suites/p384.rs
```diff
@@ -21,7 +17,7 @@ impl GroupDigest for Sha3_384 {
 
     fn hash_to_group(msg: &[u8], dst: &[u8]) -> Result<Self::Output> {
         let msgs = [msg];
-        let dsts = [NIST_P384_SHA3_384_ENC_DST, dst];
+        let dsts = [dst];
```
kostko, line 20: Did you also consider accepting a slice of DSTs or is there a reason against it?

secret-sharing/src/suites/p384.rs
```diff
@@ -7,10 +7,6 @@ use p384::{
 
 use super::{FieldDigest, GroupDigest};
 
-/// Domain separation tag for encoding to NIST P-384 prime field or curve
-/// using the SHA3-384 hash function.
-const NIST_P384_SHA3_384_ENC_DST: &[u8] = b"P384_XMD:SHA3-384_SSWU_RO_";
```
kostko, line 12: Will not using standard DSTs make it harder for third-party implementations?

go/consensus/cometbft/apps/keymanager/churp/txs.go
```diff
@@ -545,3 +556,18 @@ func resetHandoff(status *churp.Status, nextHandoff beacon.EpochTime) {
 	status.NextChecksum = nil
 	status.Applications = nil
 }
+
+func verifyPolicy(ctx *tmapi.Context, policy *churp.SignedPolicySGX) error {
+	// Allow non-empty `MayQuery` field with the 24.2 release.
+	enabled, err := features.IsFeatureVersion(ctx, "24.2")
```
kostko, line 562: We should also prepare an upgrade handler that bumps this (if not part of this PR, then open an issue).

keymanager/src/client/remote.rs
peternose: No. At least not in the near future.

secret-sharing/src/suites/p384.rs
peternose: No, could also use that. The only reason against was to simplify the interface, as I cannot return `dst: &[&[u8]]` from helper method `domain_separation_tag`. I need to create a vector, copy data, and pass the slice, so accepting slices of DSTs is not needed.

go/consensus/cometbft/apps/keymanager/churp/txs.go
peternose: I didn't prepare the handler because the next breaking release could also be 24.3. So if we prepare some bug-fixes/optimizations, we can release them with 24.2 without voting.

secret-sharing/src/suites/p384.rs
kostko: Ok, leave it as-is.

peternose: Plus, the method doesn't behave like `TupleHash`, so dsts `[[1,2],[3]]` and `[[1],[2,3]]` are the same, which could deceive the caller.

go/worker/keymanager/churp.go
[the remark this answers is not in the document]
peternose: > What would happen during a possible dump/restore upgrade, would this invalidate all of the keys? We need to be careful.

Yes, the keys would of course change. Will keep as it is.

> Yeah checks for roles inside the runtime can be added in the future when everyone is using observer nodes.

Ok, we can add this latter.

secret-sharing/src/suites/p384.rs
kostko: I guess if DSTs don't use something like length-prefix encoding for the components (e.g. like TupleHash), then it doesn't make it any easier as soon as there are custom DSTs. So in this case it is fine and this can be ignored.

peternose: By harder for third party implementations you mean that it would be great to have standard dst in the package so that third party apps can easily construct full dst?

IEFT draft _hashing to elliptic curves_ suggestions name `<app>-V<xx>-CS<yy>-<suiteID>-<enc>`  where `<xx>` and `<yy>` are two-digit numbers indicating the version and ciphersuite, respectively, and `<suiteID>` is the Suite ID of the encoding used in ciphersuite `<yy>`. I'm not sure what `<yy>` should be, but I guess in our case the dst should be `oasis-core/keymanager/churp: V00-P384_XMD:SHA3-384_SSWU_RO_-<runtime_id>-<churp_id>-shareholder` or `CHURP-V00-P384_XMD:SHA3-384_SSWU_RO_-<runtime_id>-<churp_id>-shareholder`. However, if we want to be consistent with our signature ctx, the tag should be `oasis-core/keymanager/churp: sharehoder for runtime <id> for churp <id>`. Not sure what is better.

peternose: > length-prefix encoding for the components

Tested locally, and I get the same result. I also checked the code, and length prefix or suffix (not sure what) is added only to the full dst (concatenated components).

go/consensus/cometbft/apps/keymanager/churp/txs.go
peternose: Prepared the handler in #5777, which can be renamed to 24.x latter if needed.

PR #5725, opened 2024-06-24
go/worker/storage: Do not add any particular roles

kostko:
Since the storage node is always coupled with another role, make sure to not add any particular role as otherwise this could cause observer nodes to also register as compute nodes and then misbehave.

### DIFF
--- a/go/worker/registration/worker.go
+++ b/go/worker/registration/worker.go
@@ -769,8 +769,8 @@ func (w *Worker) newRoleProvider(role node.RolesMask, runtimeID *common.Namespac
 		"id", runtimeID,
 		"role", role,
 	)
-	if !role.IsSingleRole() {
-		return nil, fmt.Errorf("RegisterRole: registration role mask does not encode a single role. RoleMask: '%s'", role)
+	if !role.IsEmptyRole() && !role.IsSingleRole() {
+		return nil, fmt.Errorf("registration role mask is non-empty and does not encode a single role: '%s'", role)
 	}
 	if err := w.runtimeRegistry.AddRoles(role, runtimeID); err != nil {
 		w.logger.Info("runtime doesn't exist, roles not registered",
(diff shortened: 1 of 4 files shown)


go/worker/storage/worker.go
```diff
@@ -70,7 +70,9 @@ func (w *Worker) registerRuntime(commonNode *committeeCommon.Node) error {
 		"runtime_id", id,
 	)
 
-	rp, err := w.registration.NewRuntimeRoleProvider(node.RoleComputeWorker, id)
+	// Since the storage node is always coupled with another role, make sure to not add any
+	// particular role here -- instead only serve as a registration gatekeeper.
+	rp, err := w.registration.NewRuntimeRoleProvider(node.RoleEmpty, id)
```
peternose, line 75: Does this return an error as the empty role is not a single role? See [newRoleProvider](https://github.com/oasisprotocol/oasis-core/blob/master/go/worker/registration/worker.go#L772).

go/worker/registration/worker.go
```diff
@@ -769,7 +769,7 @@ func (w *Worker) newRoleProvider(role node.RolesMask, runtimeID *common.Namespac
 		"id", runtimeID,
 		"role", role,
 	)
-	if !role.IsSingleRole() {
+	if role != node.RoleEmpty && !role.IsSingleRole() {
```
peternose, line 772:
```suggestion
	if !role.IsEmptyRole() && !role.IsSingleRole() {
```

go/worker/registration/worker.go
```diff
@@ -769,7 +769,7 @@ func (w *Worker) newRoleProvider(role node.RolesMask, runtimeID *common.Namespac
...
 		return nil, fmt.Errorf("RegisterRole: registration role mask does not encode a single role. RoleMask: '%s'", role)
```
peternose, line 773: The comment needs to be fixed.
What is a string representation of the empty role? Should it be `[no roles]`?

go/worker/storage/worker.go
kostko: It shouldn't since I also updated the method?

go/worker/registration/worker.go
kostko: Why does the error message need to be fixed? It would still be correct as it would never be emitted for the empty role.

peternose (approved):

peternose (approved):

PR #5760, opened 2024-07-04
go/consensus: Add the latest consensus-breaking software version

peternose:
Useful for enabling new consensus-breaking features with upcoming releases. A feature is enabled if the latest consensus-breaking software version is not lower than the version in which the feature is to be enabled.

Example of usage:
```go
if consParams.SoftwareVersion != nil && consParams.SoftwareVersion.ToU64() >= version.MustFromString("24.2").ToU64() {
    // Enable feature.
} else {
    // Disable feature and behave like version 24.1.
}
```

### DIFF
--- a/go/consensus/genesis/genesis.go
+++ b/go/consensus/genesis/genesis.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
+	"github.com/oasisprotocol/oasis-core/go/common/version"
 	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"
 	"github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common/flags"
 )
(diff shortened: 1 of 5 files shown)


go/consensus/genesis/genesis.go
```diff
@@ -42,6 +43,10 @@ type Parameters struct { // nolint: maligned
 
 	// PublicKeyBlacklist is the network-wide public key blacklist.
 	PublicKeyBlacklist []signature.PublicKey `json:"public_key_blacklist,omitempty"`
+
+	// SoftwareVersion represents the latest consensus-breaking software version
+	// that follows calendar versioning (yy.minor[.micro]).
+	SoftwareVersion *version.Version `json:"software_version,omitempty"`
```
kostko, line 49: As mentioned in the other PR, can we rename this to something like `FeatureVersion`?

kostko (commented):
As mentioned in the other PR, can we have some sort of helper for this? So we can do something like `IsFeatureVersion(ctx, "24.2")` and it would do all the required things?

go/consensus/genesis/genesis.go
peternose: Missed the comment. Yes, will change.

kostko (approved):

PR #5764, opened 2024-07-08
go/worker/registration: Support specifying only entity ID

kostko:


go/oasis-test-runner/oasis/byzantine.go
```diff
@@ -74,8 +73,8 @@ func (worker *Byzantine) ModifyConfig() error {
 	worker.AddSeedNodesToConfig()
 
 	if !worker.entity.isDebugTestEntity {
-		dir := worker.entity.dir.String()
-		worker.Config.Registration.Entity = filepath.Join(dir, "entity.json")
+		entityID, _ := worker.entity.ID().MarshalText() // Cannot fail.
```
peternose, line 76: The comment should say why this cannot fail (e.g. Returned err is always nil, MarshalText never fails, ....). Since the method returns an error, we could also return it here and then we don't need to care if the error is possible or not. Or maybe we refactor `MarshalText` to not return an error, just like cbor doesn't return an error when something is encoded.

peternose (approved):

go/oasis-test-runner/oasis/byzantine.go
kostko: We cannot refactor as this is part of the Go standard text marshaller interface.

ptrus (approved):
Nice!

PR #5785, opened 2024-07-23
Fix handling of detached bundles

kostko:
* go/runtime: Automatically enable all configured detached components
* go/runtime/bundle: Use manifest hash at bundle load time

### DIFF
--- a/go/runtime/bundle/bundle.go
+++ b/go/runtime/bundle/bundle.go
@@ -22,6 +22,9 @@ import (
 type Bundle struct {
 	Manifest *Manifest
 	Data     map[string][]byte
+
+	// manifestHash is the original manifest hash of the bundle at time the bundle was loaded.
+	manifestHash hash.Hash
 }
 
 // Validate validates the runtime bundle for well-formedness.
@@ -277,6 +280,9 @@ func (bnd *Bundle) Write(fn string) error {
 		return fmt.Errorf("runtime/bundle: failed to write bundle: %w", err)
 	}
 
+	// Update the manifest hash.
+	bnd.manifestHash = bnd.Manifest.Hash()
+
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


go/runtime/bundle/bundle.go
```diff
@@ -277,6 +280,9 @@ func (bnd *Bundle) Write(fn string) error {
 		return fmt.Errorf("runtime/bundle: failed to write bundle: %w", err)
 	}
 
+	// Update the manifest hash.
+	bnd.manifestHash = bnd.Manifest.Hash()
```
peternose, line 284: Feels like this does not belong here. Should this be moved to `Add`, where we update the manifest?

go/runtime/config/config_test.go
```diff
@@ -0,0 +1,56 @@
...
+			{
+				ID:       component.ID{Kind: component.ROFL, Name: "foo-test"},
+				Disabled: false,
+			},
+		},
+	}
+
+	compCfg, ok := cfg.GetComponent(component.ID{Kind: component.ROFL, Name: "foo-test"})
+	require.True(ok)
+	require.EqualValues(compCfg.ID.Kind, component.ROFL)
```
peternose, line 26: Any special reason why you didn't use just `Equal`? :thinking:

peternose (approved):

go/runtime/bundle/bundle.go
kostko: Hm. My thinking is that `manifestHash` should equal the hash of the manifest that is _written on disk_. Until `Write` is called nothing is written on disk so the hash should not be updated.

go/runtime/config/config_test.go
kostko: Habit :) In this case it would work as well.

go/runtime/bundle/bundle.go
peternose: Doesn't this contradicts the comment?
> manifestHash is the original manifest hash of the bundle at time the bundle was loaded.

I would expect that manifest hash is consistent with manifest in the bundle (in memory).

kostko: No, the entire point is to _not_ be what is in memory as that can be modified (and you can always get the in-memory hash using `bnd.Manifest.Hash()`). The only place where this private `manifestHash` is used is the computation of exploded bundle paths.

peternose (approved):

PR #5790, opened 2024-07-25
go/registry: Deprecate enable key manager CHURP flag

peternose:
Apply changes from:
- #5573

### DIFF
--- a/go/consensus/cometbft/apps/keymanager/churp/ext.go
+++ b/go/consensus/cometbft/apps/keymanager/churp/ext.go
@@ -7,7 +7,6 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/consensus/api"
 	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"
 	tmapi "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/api"
-	registryState "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/apps/registry/state"
 	"github.com/oasisprotocol/oasis-core/go/keymanager/churp"
 )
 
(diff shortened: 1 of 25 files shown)


go/consensus/cometbft/apps/keymanager/churp/ext.go
```diff
@@ -38,10 +37,6 @@ func (ext *churpExt) OnRegister(state tmapi.ApplicationState, _ tmapi.MessageDis
 
 // ExecuteTx implements api.Extension.
 func (ext *churpExt) ExecuteTx(ctx *tmapi.Context, tx *transaction.Transaction) error {
-	if enabled, err := ext.enabled(ctx); err != nil || !enabled {
```
kostko, line 41: I assume the only reason why this doesn't break sync from genesis is because there were no such transaction submitted? Can you make sure sync from genesis works with this change?

peternose: > I assume the only reason why this doesn't break sync from genesis is because there were no such transaction included in blocks on either Testnet or Mainnet (due to failing tx checks)?

Yes, I assumed that there were no CHURP transactions in the consensus before the 24.0 release. But how could there be? Validators should prepare and accept blocks only with valid transactions, and the execution of CHURP transactions failed because CHURP was not enabled.

> Can you make sure sync from genesis works with this change (on both networks)?

Okay, I will test the sync and see how it goes.

peternose: >  Can you make sure sync from genesis works with this change (on both networks)?

Tested Mainnet and Testnet. In both cases client synced successfully.

kostko (approved):

PR #5813, opened 2024-08-13
go/runtime: Add timeout for HostSubmitTx from ROFL apps

kostko:


go/worker/client/committee/node.go
```diff
@@ -107,11 +112,21 @@ func (n *Node) SubmitTx(ctx context.Context, tx []byte) (<-chan *api.SubmitTxRes
...
 	ch := make(chan *api.SubmitTxResult, 1)
-	n.txCh.In() <- &pendingTx{
-		txHash: hash.NewFromBytes(tx),
+	n.txCh.In() <- &wantTx{
+		txHash: txHash,
 		ch:     ch,
 	}
+	// If the context expires, make sure that we are no longer watching the transaction as otherwise
+	// it could get stuck in there forever.
+	context.AfterFunc(ctx, func() {
```
peternose, line 123: I think you have a leak here, if the caller never cancels the `ctx`, as every call creates a new go-routine. And I think you cannot expect the caller to do that.

For example, things will work fine if the caller creates a new `ctx` with timeout, but in the code above, where you use `SubmitTxNoWait`, the `ctx` might be global and canceled when the app finishes.

go/runtime/registry/host_rofl.go
```diff
@@ -181,7 +184,10 @@ func (rh *roflHostHandler) handleHostSubmitTx(
 	switch rq.Wait {
 	case true:
 		// We need to wait for transaction inclusion.
-		rsp, err := rh.client.SubmitTxMeta(ctx, submitRq)
+		submitTxCtx, cancel := context.WithTimeout(ctx, roflSubmitTxTimeout)
```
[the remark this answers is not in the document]
kostko: The caller (e.g. ROFL app itself) should do its own timeout (and it does now, I'll post a PR soon after confirming issues are fixed). This timeout is just so that in case the app doesn't, it still times out.

go/worker/client/committee/node.go
kostko: That is true, but contexts should be cancelled. You are right about each RHP handler not having its own context, this is actually a bug that could cause other resources to leak as well so I'll push a fix here.

kostko: > And I think you cannot expect the caller to do that.

If the caller doesn't signal when it is done, then it is impossible to know when to cancel the wait. Another option would be to return some sort of "subscription" instead of the channel and then ask the caller to call something like `Close` on the subscription when done.

kostko: Added a documentation comment explaining this behavior for callers and pushed some other fixes to ensure contexts get canceled.

peternose: > Of course if the caller doesn't cancel the context then why would it call `Close`? :)

A caller could call this function with `context.Background()` or `context.TODO()`, so you cannot be sure that he will cancel the context. However, if a method returns `stop()`, `close()` or `unsubscribe()`, the caller should call these methods to release resources. 

I prefer that `ctx.cancel()` is used to signal that the work associated with the context should be stopped, and `close()`  when you have a custom cleanup. See for example `AfterFunc`, it returns `stop` instead of relying on `ctx.cancel()`.

peternose: I guess you could solve this problem by separating no wait and wait submission, and use that in the `service`. But I didn't give it much thought.

kostko: The new version mentions this in the function documentation.

peternose (approved):

go/worker/client/committee/node.go
kostko: Pushed a refactored version that returns a subscription with an explicit `Stop`, please take another look.

peternose (approved):

PR #5818, opened 2024-08-19
[BACKPORT/24.1.x] Multiple backports for 24.1.1

kostko:
Backports the following:
* #5785
* #5804 
* #5813


peternose (approved):
Do we need to backport? Why not just releasing unbreaking release 24.2?

kostko:
> Why not just releasing unbreaking release 24.2

Sure, we could do that as well. I would actually like 24.2 more. Let's do it that way.

Issue #5780, opened 2024-07-18
Support bundle components without ELF binary


kostko:
Currently the assumption is that each bundle component always includes an ELF binary. This should not be required (e.g. it doesn't make sense for ROFL apps) and it may also not be always possible in the future (e.g. for Intel TDX). It should also make the runtime provisioner default to the TEE environment to simplify configuration.

PR #5825, opened 2024-08-23
oasis-test-runner: Fix sgx-only fixture

matevz:
Fixes oasis-test-runner and oasis-net-runner crashes in case you want to use SGX binary without elf.

### DIFF
--- a/go/oasis-test-runner/oasis/runtime.go
+++ b/go/oasis-test-runner/oasis/runtime.go
@@ -270,19 +270,21 @@ func (rt *Runtime) toRuntimeBundle(deploymentIndex int) (*bundle.Bundle, error) 
 		elfBin := fmt.Sprintf("component-%d-%s.elf", i, compCfg.Kind)
 		sgxBin := fmt.Sprintf("component-%d-%s.sgx", i, compCfg.Kind)
 
-		binBuf, err := os.ReadFile(compCfg.Binaries[node.TEEHardwareInvalid])
-		if err != nil {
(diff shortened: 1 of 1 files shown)


go/oasis-test-runner/oasis/runtime.go
```diff
@@ -270,19 +270,21 @@ func (rt *Runtime) toRuntimeBundle(deploymentIndex int) (*bundle.Bundle, error)
...
-			return nil, fmt.Errorf("oasis/runtime: failed to read ELF binary: %w", err)
-		}
-		_ = bnd.Add(elfBin, binBuf)
-
 		comp := &bundle.Component{
 			Kind:       compCfg.Kind,
 			Executable: elfBin,
 		}
+		
+		if rt.teeHardware == node.TEEHardwareInvalid {
```
kostko, line 278: Change to a switch statement.

go/oasis-test-runner/oasis/runtime.go
```diff
@@ -270,19 +270,21 @@ func (rt *Runtime) toRuntimeBundle(deploymentIndex int) (*bundle.Bundle, error)
 		elfBin := fmt.Sprintf("component-%d-%s.elf", i, compCfg.Kind)
 		sgxBin := fmt.Sprintf("component-%d-%s.sgx", i, compCfg.Kind)
 
-		binBuf, err := os.ReadFile(compCfg.Binaries[node.TEEHardwareInvalid])
```
kostko, line 273: I don't think bundles without the ELF part are currently supported (see #5780).

matevz:
SGX binaries without ELF are not supported. Closing.

PR #5836, opened 2024-08-27
buildkite: Enable debug assertions in runtimes

peternose:
Tested `debug_assert` in https://buildkite.com/oasisprotocol/oasis-core-ci/builds/14000.

### DIFF
--- a/.buildkite/code.pipeline.yml
+++ b/.buildkite/code.pipeline.yml
@@ -158,6 +158,8 @@ steps:
       # Upload the built artifacts.
       - cd /var/tmp/artifacts/default/release
       - buildkite-agent artifact upload oasis-core-runtime-loader
+    env:
+      RUSTFLAGS: "-C debug-assertions=yes"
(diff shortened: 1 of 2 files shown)


.buildkite/code.pipeline.yml
```diff
@@ -196,6 +198,8 @@ steps:
       - mv simple-keyvalue simple-keyvalue.mocksgx
       - buildkite-agent artifact upload simple-keymanager.mocksgx
       - buildkite-agent artifact upload simple-keyvalue.mocksgx
+    env:
+      RUSTFLAGS: "-C target-feature=+aes,+ssse3 -C debug-assertions=yes"
```
peternose, line 202: Could also modify flags with `export`, not sure what is better.

peternose:
> I assume there is no noticable performance impact?

I didn't notice anything. SGX tests still run from 8-12min, one 18min, which is the same as before.

PR #5838, opened 2024-08-28
go/runtime: Support bundle components without ELF binary

### LINKED ISSUE #5780: Support bundle components without ELF binary

kostko:
Fixes #5780

### DIFF
--- a/go/runtime/registry/config.go
+++ b/go/runtime/registry/config.go
@@ -102,7 +102,62 @@ func newConfig( //nolint: gocyclo
 
 	// Check if any runtimes are configured to be hosted.
 	if haveSetRuntimes || (cmdFlags.DebugDontBlameOasis() && viper.IsSet(CfgDebugMockIDs)) {
+		// By default start with the environment specified in configuration.
 		runtimeEnv := config.GlobalConfig.Runtime.Environment
+
+		// Preprocess runtimes to separate detached from non-detached.
+		type nameKey struct {
+			runtime common.Namespace
+			comp    component.ID
+		}
+
+		var (
+			regularBundles []*bundle.Bundle
+			err            error
+		)
+		detachedBundles := make(map[common.Namespace][]*bundle.Bundle)
+		existingNames := make(map[nameKey]struct{})
+		for _, path := range config.GlobalConfig.Runtime.Paths {
+			var bnd *bundle.Bundle
+			if bnd, err = bundle.Open(path); err != nil {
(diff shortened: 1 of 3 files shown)


go/runtime/registry/config.go
```diff
@@ -102,7 +102,62 @@ func newConfig( //nolint: gocyclo
...
+			switch bnd.Manifest.IsDetached() {
+			case false:
+				// A regular non-detached bundle that has the RONL component.
+				regularBundles = append(regularBundles, bnd)
+			case true:
+				// A detached bundle without the RONL component that needs to be attached.
+				detachedBundles[bnd.Manifest.ID] = append(detachedBundles[bnd.Manifest.ID], bnd)
+
+				// Ensure there are no name conflicts among the components.
+				for compID := range bnd.Manifest.GetAvailableComponents() {
```
peternose, line 140: I forgot why was this not moved outside of the `switch`, to also check that we have only one RONL component for runtime and to also take into account components from non-detached bundles? :thinking:

peternose (approved):

go/runtime/registry/config.go
kostko: Inside the same bundle this is already checked, so this is only needed when you have detached bundles. Unless I'm missing something.

kostko: > Can a non-detached bundle (the one which contains RONL) also contain components

Yes.

> and do we need to compare them against detached for duplicates

Components in the RONL bundle currently override any components from detached bundles, see below:
```rust
// Skip components that already exist in the bundle itself.
if bnd.Manifest.GetComponentByID(detachedComp.ID()) != nil {
	continue
}
```

> Could you have 2 non-detached bundles with the same RONL, do we capture this?

Yes, you can have multiple bundles but they need to be for different versions. If they are for the same version this is an error captured below:
```rust
if _, ok := rh.Runtimes[id][bnd.Manifest.Version]; ok {
	return nil, fmt.Errorf("duplicate runtime '%s' version '%s'", id, bnd.Manifest.Version)
}
```

peternose: OK, than everything works as expected.

PR #5847, opened 2024-09-10
secret-sharing/src/churp: Separate handoff kinds

peternose:
Trying to simply the code by removing `if` conditions for handoff kind in `Handoff` methods.

### DIFF
--- a/keymanager/src/churp/handler.rs
+++ b/keymanager/src/churp/handler.rs
@@ -33,7 +33,10 @@ use oasis_core_runtime::{
 };
 
 use secret_sharing::{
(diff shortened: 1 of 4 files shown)


keymanager/src/churp/handler.rs
```diff
@@ -114,7 +117,7 @@ struct HandoffInfo<G: Group + GroupEncoding> {
     /// The handoff epoch.
     epoch: EpochTime,
     /// The handoff associated with this information.
-    handoff: Arc<Handoff<G>>,
+    handoff: Arc<Box<dyn Handoff<G> + Send + Sync>>,
```
kostko, line 120: You could change the `Handoff` trait to be `: Send + Sync`.

kostko (approved):

PR #5852, opened 2024-09-13
keymanager: Add mock trusted signers for debug mock SGX builds

kostko:
Useful after #5851 to allow configuring the set of mock trusted signers used by net runner.

### DIFF
--- a/keymanager/src/policy/signers.rs
+++ b/keymanager/src/policy/signers.rs
@@ -18,6 +18,38 @@ pub struct TrustedSigners {
     pub threshold: u64,
 }
 
+#[cfg(feature = "debug-mock-sgx")]
+impl TrustedSigners {
+    /// An UNSAFE set of trusted signers using well-known debug keys.
+    pub fn unsafe_mock() -> Self {
+        use oasis_core_runtime::{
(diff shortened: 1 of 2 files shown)


keymanager/src/policy/signers.rs
```diff
@@ -18,6 +18,38 @@ pub struct TrustedSigners {
     pub threshold: u64,
 }
 
+#[cfg(feature = "debug-mock-sgx")]
+impl TrustedSigners {
+    /// An UNSAFE set of trusted signers using well-known debug keys.
+    pub fn mock() -> Self {
```
peternose, line 24:
```suggestion
    pub fn unsafe_mock() -> Self {
```
If you want to emphasize unsafe.

peternose (commented):
Can we remove `trusted_signers()`?

kostko:
> Can we remove trusted_signers()?

In the `simple-keymanager` test runtime? That one doesn't always use the debug-mock-sgx feature though, so the question then is whether exposing this unsafe function is ok.

keymanager/src/policy/signers.rs
kostko: Changed.

peternose (approved):
Looks good then.

PR #5854, opened 2024-09-17
Bump Go to 1.23.1

### LINKED ISSUE #5826: Bump Go to 1.23.2
[Go 1.23 has been released](https://go.dev/doc/go1.23) so we should make sure to update to the latest version.

jberci:
Closes #5826


kostko:
We should probably wait for https://github.com/golang/go/issues/69312 to get fixed in 1.23.x as it impacts libp2p.

peternose:
> as it impacts libp2p.

How did you figure this out? Was it due to failed tests, or are you tracking issues of all our dependencies? :thinking:

kostko:
Yeah, I'm tracking issues and changelogs of some of our key dependencies :)

kostko:
Looks like the issue has been fixed in Go 1.23.2.

go/common/args.go
```diff
@@ -17,8 +17,8 @@ func TrimArgs(osArgs []string) []string {
 
 	for i, w := range osArgs {
 		if w == argSeparator {
-			return append([]string{osArgs[0]}, osArgs[i+1:]...)
+			return append([]string{osArgs[0]}, osArgs[i+1:]...) //nolint:gosec // false G602 out of range
 		}
 	}
-	return osArgs[0:1]
+	return osArgs[0:1] //nolint:gosec // false G602 out of range
```
peternose, line 23: Could also change the code.

```go
func TrimArgs(osArgs []string) []string {
	if len(osArgs) == 0 {
		panic("TrimArgs called with an empty argument list")
	}

	args := []string{osArgs[0]}

	for i, w := range osArgs {
		if w == argSeparator {
			return append(args, osArgs[i+1:]...)
		}
	}

	return args
}
```

jberci: Yup, seems to work... O_o

kostko (approved):

Issue #5766, opened 2024-07-09
Support concurrent key manager client requests


peternose:
Currently, the key manager client cannot make concurrent RPC requests because peer feedback is tied to the last request.

PR #5863, opened 2024-09-24
keymanager/src/client: Fetch churp key shares concurrently

peternose:
Peer feedback, which should not work anymore when fetching key shares, will be fixed in another PR.

### DIFF
--- a/keymanager/src/client/remote.rs
+++ b/keymanager/src/client/remote.rs
@@ -9,6 +9,7 @@ use std::{
 
 use anyhow::anyhow;
 use async_trait::async_trait;
+use futures::stream::{FuturesUnordered, StreamExt};
 use group::GroupEncoding;
 use lru::LruCache;
 use rand::{prelude::SliceRandom, rngs::OsRng};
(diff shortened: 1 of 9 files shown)


keymanager/src/client/remote.rs
```diff
@@ -244,33 +245,46 @@ impl RemoteClient {
...
+                        KeyShareRequest {
+                            id: status.id,
+                            runtime_id: status.runtime_id,
+                            epoch: status.handoff,
+                            key_runtime_id: self.runtime_id,
+                            key_id,
+                        },
+                        vec![node_id],
+                    )
+                    .await
```
kostko, line 277: This seems like it is awaiting for the call the complete before proceeding? The `into_result_with_feedback` should be done further below after the call future resolves.

kostko (commented):
> Peer feedback, which should not work anymore when fetching key shares, will be fixed in another PR.

Does it make sense to open a new issue?

peternose:
> Does it make sense to open a new issue?

No, as this should be fixed in https://github.com/oasisprotocol/oasis-core/issues/5766.

kostko (approved):

PR #5868, opened 2024-09-25
oasis-test-runner: Enable sgx_loader in client node

matevz:
Needed by https://github.com/oasisprotocol/oasis-web3-gateway/pull/625

### DIFF
--- a/go/oasis-test-runner/oasis/client.go
+++ b/go/oasis-test-runner/oasis/client.go
@@ -61,6 +61,8 @@ func (client *Client) ModifyConfig() error {
 	if len(client.runtimes) > 0 {
 		client.Config.Mode = config.ModeClient
 		client.Config.Runtime.Provisioner = client.runtimeProvisioner
+		client.Config.Runtime.SGXLoader = client.net.cfg.RuntimeSGXLoaderBinary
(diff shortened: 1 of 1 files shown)


go/oasis-test-runner/oasis/client.go
```diff
@@ -61,6 +61,8 @@ func (client *Client) ModifyConfig() error {
 	if len(client.runtimes) > 0 {
 		client.Config.Mode = config.ModeClient
 		client.Config.Runtime.Provisioner = client.runtimeProvisioner
+		client.Config.Runtime.SGXLoader = client.net.cfg.RuntimeSGXLoaderBinary
```
peternose, line 64: Is it possible that `net` or `cfg` is `nil`?

kostko (approved):

go/oasis-test-runner/oasis/client.go
kostko: Should never be.

matevz:
Since we won't be running ROFLs on the client nodes on sapphire-localnet yet I think there is no need for this PR anymore.

kostko:
Still makes sense to do it?

PR #5872, opened 2024-09-30
runtime/src/enclave_rpc: Support peer feedback for concurrent requests

peternose:
Example logs:
```json
// First client.
{"caller":"keymanager.go:242","level":"debug","module":"worker/keymanager","msg":"received peer feedback from runtime","peer_feedback":"failure","request_id":11876174929579737116,"ts":"2024-09-30T11:13:55.971535462Z","valid":true}
{"caller":"keymanager.go:242","level":"debug","module":"worker/keymanager","msg":"received peer feedback from runtime","peer_feedback":"success","request_id":11876174929579737117,"ts":"2024-09-30T11:13:55.972354408Z","valid":true}
// No peers found error, so peer feedback is never stored.
{"caller":"keymanager.go:242","level":"debug","module":"worker/keymanager","msg":"received peer feedback from runtime","peer_feedback":"failure","request_id":11876174929579737118,"ts":"2024-09-30T11:13:55.972777804Z","valid":false}
// Second client.
{"caller":"keymanager.go:242","level":"debug","module":"worker/keymanager","msg":"received peer feedback from runtime","peer_feedback":"success","request_id":11876174933874704385,"ts":"2024-09-30T11:14:10.333719181Z","valid":true}
```

### DIFF
--- a/go/worker/common/committee/keymanager.go
+++ b/go/worker/common/committee/keymanager.go
@@ -4,11 +4,13 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/libp2p/go-libp2p/core"
 	"golang.org/x/exp/maps"
 
 	"github.com/oasisprotocol/oasis-core/go/common"
+	"github.com/oasisprotocol/oasis-core/go/common/cache/lru"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
 	cmSync "github.com/oasisprotocol/oasis-core/go/common/sync"
@@ -18,7 +20,18 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/p2p/rpc"
 	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
 	enclaverpc "github.com/oasisprotocol/oasis-core/go/runtime/enclaverpc/api"
+	runtimeKeymanager "github.com/oasisprotocol/oasis-core/go/runtime/keymanager/api"
 	keymanagerP2P "github.com/oasisprotocol/oasis-core/go/worker/keymanager/p2p"
+)
+
+const (
+	// peerFeedbackCacheSize is the maximum number of peer feedbacks
+	// in the cache.
+	peerFeedbackCacheSize = 100
+
+	// maxPeerFeedbackAge is the maximum age of peer feedback in the cache
+	// before it is discarded.
+	maxPeerFeedbackAge = time.Minute
 )
 
 // KeyManagerClientWrapper is a wrapper for the key manager P2P client that handles deferred
@@ -37,6 +50,7 @@ type KeyManagerClientWrapper struct {
 	logger       *logging.Logger
 
 	lastPeerFeedback rpc.PeerFeedback
+	peerFeedbacks    *lru.Cache
 }
 
 // Initialized returns a channel that gets closed when the client is initialized.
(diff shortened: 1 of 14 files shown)


runtime/src/enclave_rpc/client.rs
```diff
@@ -408,11 +471,15 @@ impl RpcClient {
         // Create the command channel.
         let (tx, rx) = mpsc::channel(CMDQ_BACKLOG);
 
+        // Ensure every client generates unique request IDs.
+        let last_request_id = Self::generate_random_request_id();
```
kostko, line 475: Hm, this could in theory overlap? We could use a random ID for each request or a global atomic for these, but then the individual methods (`connect`, `call*`, etc.) would need to return the request ID instead of updating/reading from `self.last_request_id`. Or maybe the odds are small?

runtime/src/enclave_rpc/client.rs
```diff
@@ -550,6 +620,19 @@ impl RpcClient {
             .unwrap();
     }
 
+    /// Generate a random request ID.
+    #[cfg(not(test))]
+    fn generate_random_request_id() -> u64 {
+        let mut rng = OsRng {};
+        rng.next_u64() & (!(1 << 63))
```
kostko, line 627:
```suggestion
        OsRng.next_u64() & (!(1 << 63))
```

go/worker/common/committee/keymanager.go
```diff
@@ -317,3 +410,152 @@ func newKeyManagerNodeTracker(p2p p2p.Service, consensus consensus.Backend, keym
...
+	// timestamp is the time when the feedback was added to the cache.
+	timestamp time.Time
+	// prev is a pointer to the previous node in the cache.
+	prev *peerFeedbackNode
+	// next is a pointer to the next node in the cache.
+	next *peerFeedbackNode
+}
+
+// peerFeedbackCache is a cache that holds the most recent unused peer feedback.
+type peerFeedbackCache struct {
```
kostko, line 429: Consider `container/list` and/or `go/common/cache/lru`.

peternose: Used `lru`, stupid me :facepalm:

runtime/src/enclave_rpc/client.rs
peternose: > Hm, this could in theory overlap? 

In theory could overlap, but probability should be negligible for small number of clients. 

> We could use a random ID for each request

This should be the same. What you loose here is that requests from the same client don't have consecutive numbers.

>  global atomic

Not the best, as runtime restart starts with the same ids.

>  individual methods (`connect`, `call*`, etc.) would need to return the request ID instead of updating/reading from `self.last_request_id`

Don't want to return request from `transport`, as it complicates things and probably doesn't even belong there.

Change to `request_id = client_id << 32 + counter % (1 << 32)`, i.e. I use first 32 bits for client id (a random value on startup which is increased for every new client) and the last 32 bits for counter/call id (cycling counter when reaching max).

runtime/src/enclave_rpc/client.rs
```diff
@@ -311,11 +363,30 @@ impl Controller {
...
-            .map(|(data, _)| data)
+            .map(|rsp| rsp.data);
+
+        let pf = if result.is_err() {
+            types::PeerFeedback::Failure
+        } else {
+            types::PeerFeedback::Success
+        };
+
+        let _ = self.transport.submit_peer_feedback(request_id, pf).await; // Ignore error.
```
peternose, line 384: Maybe peer feedback should never be submitted for close messages, not sure :thinking:
For example, if call fails during connect, we will give a bad feedback, and another one when we gracefully close  the session.

kostko: Yeah probably should not submit it for close as the error is ignored anyway. It serves mostly to notify the other side.

kostko (approved):

PR #5876, opened 2024-09-30
go/storage: Add automatic storage backend detection

### LINKED ISSUE #5875: Default storage backend to pathbadger for new nodes
We should default the storage backend to pathbadger for new nodes, e.g. nodes that don't yet have an existing data directory.

kostko:
Fixes #5875 

The new default storage backend is "auto" which attempts to detect the storage backend that should be used based on existing data directories. When none exist, "pathbadger" is used. When multiple exist, the most recently modified one is used.

This should make newly deployed nodes default to pathbadger.

### DIFF
--- a/go/storage/database/database.go
+++ b/go/storage/database/database.go
@@ -5,7 +5,10 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
 	"path/filepath"
+	"slices"
+	"time"
 
 	"github.com/oasisprotocol/oasis-core/go/storage/api"
 	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/checkpoint"
@@ -14,10 +17,16 @@ import (
 )
 
 const (
+	// BackendNameAuto is the name of the automatic backend detection "backend".
+	BackendNameAuto = "auto"
 	// BackendNameBadgerDB is the name of the BadgerDB backed database backend.
 	BackendNameBadgerDB = "badger"
 	// BackendNamePathBadger is the name of the PathBadger database backend.
 	BackendNamePathBadger = "pathbadger"
+
+	// defaultBackendName is the default backend in case automatic backend detection is enabled and
+	// no previous backend exists.
+	defaultBackendName = BackendNamePathBadger
 
 	checkpointDir = "checkpoints"
 )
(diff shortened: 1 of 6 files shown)


ptrus (approved):

go/worker/storage/config/config.go
```diff
@@ -35,14 +35,17 @@ type CheckpointerConfig struct {
...
+		return err
+	}
+	return nil
 }
 
 // DefaultConfig returns the default configuration settings.
 func DefaultConfig() Config {
 	return Config{
-		Backend:                "badger",
+		Backend:                "auto",
```
peternose, line 48: Can we use here `BackendNameAuto`?

go/storage/database/database.go
```diff
@@ -14,11 +17,17 @@ import (
...
+	// BackendNameAuto is the name of the automatic backend detection "backend".
+	BackendNameAuto = "auto"
 	// BackendNameBadgerDB is the name of the BadgerDB backed database backend.
 	BackendNameBadgerDB = "badger"
 	// BackendNamePathBadger is the name of the PathBadger database backend.
 	BackendNamePathBadger = "pathbadger"
 
+	// autoDefaultBackend is the default backend in case automatic backend detection is enabled and
+	// no previous backend exists.
+	autoDefaultBackend = BackendNamePathBadger
```
peternose, line 29:
```suggestion
	defaultBackendName = BackendNamePathBadger
```

go/storage/database/database.go
```diff
@@ -164,3 +177,59 @@ func (ba *databaseBackend) Checkpointer() checkpoint.CreateRestorer {
...
+			path:      maybeDb,
+			timestamp: fi.ModTime(),
+			name:      b.Name(),
+		})
+	}
+	slices.SortFunc(backends, func(a, b foundBackend) int {
+		return a.timestamp.Compare(b.timestamp)
+	})
+
+	// If no existing backends are available, default to "pathbadger".
```
peternose, line 222:
```suggestion
	// If no existing backends are available, use default.
```
We can change default in the future, which will invalidate this comment.

go/storage/database/database_test.go
```diff
@@ -51,3 +51,92 @@ func doTestImpl(t *testing.T, backend string) {
...
+		// When there is no existing database directory, the default should be used.
+		cfg := api.Config{
+			Backend: "auto",
+			DB:      filepath.Join(tmpDir, DefaultFileName("auto")),
+		}
+		impl, err := New(&cfg)
+		require.NoError(err)
+		impl.Cleanup()
+
+		require.Equal("pathbadger", cfg.Backend)
```
peternose, line 72:
```suggestion
		require.Equal(defaultStorageBackend, cfg.Backend)
```
Same below.

go/worker/storage/config/config.go
kostko: Unfortunately this would currently lead to an import cycle.

peternose (approved):

PR #5878, opened 2024-10-02
go/common/grpc: allow non-tls connections to loopback addresses

ptrus:
This usecase came up in https://github.com/oasisprotocol/oasis-web3-gateway/pull/626 where we want to expose the node from docker via a port (since this works better cross-platform than exposing the socket via a volume).

It was requested that both `oasis-node` and `oasis-cli` work with the exposed endpoint, for better developer experience.

--- 

The added `CfgInsecureLoopback` might be a bit over the top, I'm not sure if we should keep it or not. We could always drop TLS for loopback address. But this could break any existing setup if someone did setup proper TLS with trusted certs on his local setup, so that's why I added it.

The `IsLocalAddress` is made public so that it can be reused in `oasis-sdk` and we won't need to duplicate the not most intuitive gRPC address parsing logic (https://github.com/oasisprotocol/oasis-sdk/pull/2013).

### DIFF
--- a/go/common/grpc/grpc_test.go
+++ b/go/common/grpc/grpc_test.go
@@ -1,0 +1,45 @@
+package grpc
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestIsLocalRPC(t *testing.T) {
+	for _, tc := range []struct {
+		name     string
+		addr     string
+		expected bool
+	}{
+		// Invalid inputs.
+		{"Empty", "", false},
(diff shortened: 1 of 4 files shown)


go/common/grpc/grpc_test.go
```diff
@@ -0,0 +1,47 @@
+package grpc_test
```
kostko, line 1:
```suggestion
package grpc
```

kostko (approved):

PR #5879, opened 2024-10-03
runtime/src/enclave_rpc/client: Support concurrent sessions

peternose:
Measured locally execution time on `compute-0` of `loop` in method `churp_recover_state_key` in test `e2e.runtime.keymanager-churp-txs`. 

Speed-up looks lower than expected, but I didn't investigate the numbers.

```
Before:
Time elapsed: 31.48498ms
Time elapsed: 33.393901ms
Time elapsed: 32.392303ms
Time elapsed: 32.062076ms

After:
Time elapsed: 22.728806ms
Time elapsed: 27.323837ms
Time elapsed: 22.350006ms
Time elapsed: 23.219776ms
```


runtime/src/enclave_rpc/client.rs
```diff
@@ -456,48 +108,112 @@ impl<T> Response<T> {
...
+
+    /// Update allowed remote enclave identities.
+    ///
+    /// Useful if the key manager's policy has changed.
+    pub async fn update_enclaves(&self, enclaves: Option<HashSet<EnclaveIdentity>>) {
+        let sessions = {
+            let mut sessions = self.sessions.lock().await;
+            match sessions.update_enclaves(enclaves) {
+                true => sessions.clear(),
+                false => vec![],
```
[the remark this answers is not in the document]
peternose: Currently, no. On one hand, it's strange that the `update_enclaves` method removes and returns sessions, but on the other hand, I agree with you that this should be done together. Decided to move to `sessions`.

runtime/src/enclave_rpc/sessions.rs
```diff
@@ -0,0 +1,945 @@
...
+        sessions.remove(&session.session_id);
+        self.by_idle_time.remove(&key);
+
+        // If peer doesn't have any more sessions, remove the peer.
+        if sessions.is_empty() {
+            self.by_peer.remove(&session.peer_id);
+        }
+    }
+
+    /// Clear all sessions.
```
kostko, line 508: And return them? Also the method sounds more like `drain` :)

runtime/src/enclave_rpc/client.rs
```diff
@@ -456,48 +108,103 @@ impl<T> Response<T> {
...
+        let sessions = {
+            let mut sessions = self.sessions.lock().await;
+            sessions.update_enclaves(enclaves)
         };
-        tokio::spawn(controller.run());
+        self.close_all(sessions).await;
+    }
 
-        Self { cmdq: tx }
+    /// Update key manager's quote policy.
```
kostko, line 192: Isn't this the "remote end's quote policy" and not necessarily the key manager's? As EnclaveRPC can also be used for something else.

runtime/src/enclave_rpc/client.rs
```diff
@@ -456,48 +108,103 @@ impl<T> Response<T> {
...
+            builder,
+            max_sessions,
+            max_sessions_per_peer,
+            stale_session_timeout,
+        )
+    }
+
+    /// Update allowed remote enclave identities.
+    ///
+    /// Useful if the key manager's policy has changed.
```
kostko, line 183: Not sure if we should mention the key manager in the generic EnclaveRPC client?

kostko (approved):

runtime/src/enclave_rpc/client.rs
peternose: Definitely not. Removing as this was copy pasted from the key manager client.

kostko (approved):

PR #5897, opened 2024-10-11
secret-sharing/src/churp/switch: Verify combined bivariate polynomial

peternose:


secret-sharing/src/churp/dealer.rs
```diff
@@ -89,15 +119,41 @@ where
...
         let deg_x = threshold;
         let deg_y = threshold.checked_mul(2).ok_or(Error::ThresholdTooLarge)?;
-        let bp = BivariatePolynomial::random(deg_x, deg_y, rng);
-        Ok(bp)
+
+        // When using a random RNG and a large prime field, this loop
+        // should execute once with an extremely high probability,
+        // so there is no need to optimize it by randomly selecting
+        // only the problematic coefficients.
+        loop {
```
kostko, line 141: Should we still bound the loop to some number of iterations and fail? Just in the unlikely case the RNG is broken.

kostko (approved):

secret-sharing/src/churp/dealer.rs
peternose: Yes, we should.
- #5914

PR #5928, opened 2024-11-05
secret-sharing: Zeroize sensitive data

peternose:


keymanager/src/churp/storage.rs
```diff
@@ -138,17 +158,31 @@ impl Storage {
...
     ) -> Vec<u8> {
+        // Prepare data for encryption.
         let nonce = Nonce::generate();
-        let plaintext = polynomial.to_bytes();
+        let mut plaintext = polynomial.to_bytes();
         let additional_data = Self::pack_churp_id_epoch(churp_id, epoch);
+
+        // Encrypt data using seal_into so that we can zeroize the plaintext.
+        let mut ciphertext = Vec::with_capacity(plaintext.len() + TAG_SIZE + NONCE_SIZE);
+        unsafe { ciphertext.set_len(plaintext.len() + TAG_SIZE) }
```
kostko, line 174: Why do you need unsafe here? Couldn't you just use `ciphertext.resize` instead?

keymanager/src/churp/storage.rs
```diff
@@ -159,53 +193,98 @@ impl Storage {
...
+        // Prepare data for encryption.
         let share: EncodedVerifiableSecretShare = verifiable_share.into();
         let nonce: Nonce = Nonce::generate();
-        let plaintext = cbor::to_vec(share);
+        let mut plaintext = cbor::to_vec(share);
         let additional_data = Self::pack_churp_id_epoch(churp_id, epoch);
+
+        // Encrypt data using seal_into so that we can zeroize the plaintext.
+        let mut ciphertext = Vec::with_capacity(plaintext.len() + TAG_SIZE + NONCE_SIZE);
+        unsafe { ciphertext.set_len(plaintext.len() + TAG_SIZE) }
```
kostko, line 235: Same as above.

keymanager/src/churp/storage.rs
peternose: This was copied from `seal` function (maybe should add this as a comment). I think that the idea is to speed up encryption by using vector as a buffer and in that case you don't care about the content and you don't need to initialize data by resizing. 

The first example in `set_len` says: _This method can be useful for situations in which the vector is serving as a buffer ..._

kostko: Ok, I would suggest adding a comment.

kostko (approved):

PR #5931, opened 2024-11-06
go/runtime/bundle: Add expected enclave identity to manifest

### LINKED ISSUE #5929: Include expected enclave identity in manifest
Currently the enclave identity is computed on-demand for SGX components. In case of other TEEs like TDX, this is not always possible or may involve a complex process. To this end, we should make expected enclave identity computation a build-time feature and include it in the manifest.

kostko:
Fixes #5929

### DIFF
--- a/go/runtime/bundle/bundle.go
+++ b/go/runtime/bundle/bundle.go
@@ -219,8 +219,33 @@ func (bnd *Bundle) MrSigner(id component.ID) (*sgx.MrSigner, error) {
 	return &mrSigner, nil
 }
 
-// EnclaveIdentity returns the SGX enclave identity of the given component.
+// EnclaveIdentity returns the enclave identity of the given component.
+//
+// Deprecated: Use EnclaveIdentities instead.
 func (bnd *Bundle) EnclaveIdentity(id component.ID) (*sgx.EnclaveIdentity, error) {
+	ids, err := bnd.EnclaveIdentities(id)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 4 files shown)


go/runtime/bundle/bundle.go
```diff
@@ -219,8 +219,33 @@ func (bnd *Bundle) MrSigner(id component.ID) (*sgx.MrSigner, error) {
...
+
+// EnclaveIdentities returns the enclave identities of the given component.
+func (bnd *Bundle) EnclaveIdentities(id component.ID) ([]sgx.EnclaveIdentity, error) {
+	// If the component has a build-time known expected identity, use it.
+	comp := bnd.Manifest.GetComponentByID(id)
+	if comp == nil {
+		return nil, fmt.Errorf("runtime/bundle: component '%s' not available", id)
+	}
+	if len(comp.Identities) > 0 {
+		ids := make([]sgx.EnclaveIdentity, 0, len(comp.Identities))
```
peternose, line 241:
```suggestion
		ids := slices.Clone(comp.Identities)
```

kostko: Oh, nice!

kostko: Oh, but this only works for the same type, these are different now and require a map function as well. Didn't see anything like it in `slices`?

peternose: Overlooked `id.Enclave` :facepalm: I think there is no function that behaves like map, so please ignore this comment.

Issue #5755, opened 2024-07-03
Hot-loading and distribution system for runtime bundles


kostko:
Either our own P2P protocol or we could just store them in IPFS or similar. We could simply chunk the bundle into equally-sized chunks and then provide those to peers. In order to reduce avenues for a DoS attack, only nodes that are registered for a given runtime would be eligible to request bundles.

There should also be support for hot-loading (e.g. without node restart) runtimes (and individual detached components) after they are fetched.

kostko:
An initial step that we can implement immediately could be to just have a backend that can fetch binaries via HTTPS. There would be two steps:
* Looking up the bundle URL from a known manifest hash. E.g. these could be implemented by having a repository of files where the filename is the Base16-encoded manifest hash. The content of the file would be just the URL to use.
* Fetching the actual bundle from the resolved URL.

As long as the manifest hash is available for verification on-chain, it should be fine. We can then gradually transition to a more decentralized version.

Issue #5737, opened 2024-07-03
Periodically clean up cached bundles directory


kostko:
Currently the cached exploded bundles are never removed which can result in a lot of disk use after a few upgrades. We should make it so that cached bundles are periodically removed if they are either:
* For versions lower than the active runtime version (regular bundles).
* From bundles that are no longer present in configuration (regular and detached bundles).

martintomazic:
There might be exception for the paratime clients (?):
> For versions lower than the active runtime version (regular bundles).

Haven't checked this in the code yet, but (see excerpt from the docs; [updating paratimes](https://docs.oasis.io/node/run-your-node/maintenance/handling-network-upgrades#updating-paratimes)):
> If you are running a client node, ParaTime state syncing will be performed regardless of the version installed.

kostko:
That is related to syncing runtime transactions and state. Since we are using state diffs, there is no need to have the correct runtime version as the runtime logic itself is not involved during storage sync.

martintomazic:
>cached bundles are **periodically**  removed 

I was thinking about keeping it simple i.e.:
* Single clean-up during/after [`newRuntimeConfig`](https://github.com/oasisprotocol/oasis-core/blob/907d604502bc0dde829d78a58498f2b34812cb78/go/runtime/registry/config.go#L74) to handle stale exploded bundles no longer in the config.
* For versions lower then the current active one, doing a clean-up where the actual hosted runtime version changes ([see](https://github.com/oasisprotocol/oasis-core/blob/907d604502bc0dde829d78a58498f2b34812cb78/go/runtime/registry/host.go#L161)) and also remove it from the [`hostConfig`](https://github.com/oasisprotocol/oasis-core/b.ob/907d604502bc0dde829d78a58498f2b34812cb78/go/runtime/registry/registry.go#L136) of the corresponding runtime registry?
    *  Am I complicating and should simply watch [`WatchActiveDescriptor`](https://github.com/oasisprotocol/oasis-core/blob/907d604502bc0dde829d78a58498f2b34812cb78/go/runtime/registry/registry.go#L177) instead?
    * I am worried this might be bad idea if the runtime lags slighlty behind the consensus, or is this actually not possible by design?
    
Will hopefully have draft PR ready tmr :)

kostko:
> Single clean-up during/after

I would just make sure that this doesn't block initialization so while it should be fired on startup, it should be done in the background.

[...]

PR #5961, opened 2024-12-06
go/runtime: Support hot-loading of runtime bundles

peternose:
Issue: https://github.com/oasisprotocol/oasis-core/issues/5755

Example status output:

```json
"bundles": [
  {
    "name": "test-runtime",
    "id": "8000000000000000000000000000000000000000000000000000000000000000",
    "components": [
      {
        "kind": "ronl",
        "version": {}
      },
      {
        "kind": "rofl",
        "version": {}
      }
    ]
  }
],
```


go/runtime/bundle/discovery.go
```diff
@@ -0,0 +1,452 @@
...
+	}
+	defer os.Remove(src)
+
+	d.logger.Info("bundle downloaded",
+		"runtime_id", runtimeID,
+		"base_url", baseURL,
+		"filename", filename,
+	)
+
+	if err := d.registry.AddBundle(src, manifestHash); err != nil {
```
ptrus, line 321: It is a bit scary that `AddBundle` needs to "Open" the (untrusted) bundle before verifying the manifest hash.  I am not sure, but the current implementation of `Open` might assume that bundle came from a (somewhat) trusted source, which might not be the case in future. We should ensure that `Open` is safe to use with malicious inputs. 

---

Somehwat related, I remember we talked about https://github.com/oasisprotocol/oasis-core/commit/38edded4cad5cabbe9a78953880448ba540d0d02 . Now that I have more context, shouldn't that actually be the hash of the entire bundle (not just the manifest)? This would prevent this problem and allow us to verify the downloaded file before opening it.

ptrus (commented):
(self-note): only reviewed the discovery part now

go/runtime/bundle/discovery.go
peternose: Validation is actually hidden in the `Open` function, so the bundle doesn't need to be trusted.

We can refactor this to use bundle checksum instead (maybe in another PR), as that might be more intuitive. However, the current approach allows us to remove the bundles entirely as one could just publish the manifest (e.g. on chain) and the users should then fetch binaries from untrusted sources and verify their checksums.

@kostko What are your thoughts?

go/runtime/bundle/registry.go
```diff
@@ -0,0 +1,346 @@
...
+			manifestHash.Hex(),
+		)
+	}
+
+	// Skip already processed bundles.
+	if _, ok := r.bundles[manifestHash]; ok {
+		return nil
+	}
+
+	// Verify that components are unique.
```
ptrus, line 121:
```suggestion
	// Ensure the manifest doesn't include a component version already in the registry.
```

Something like that sounds clearer to me. Otherwise, I thought we were checking that the manifest doesn't contain duplicate components (which is already checked in `Open`).

go/runtime/bundle/discovery.go
kostko: The idea for using the manifest hash was that one would not need to always download all of the bundle artifacts as some may be shared between bundles.

We should probably add a separate `Open` variant (or have the existing one take options), one that takes a manifest hash. Obviously the zip archive parser should be well-fuzzed (which the current implementation is) and since the manifest _must_ be the first file in the archive, we can just verify the manifest hash before continuing to load or parse anything else. I would opt for this variant.

go/runtime/bundle/discovery.go
```diff
@@ -0,0 +1,452 @@
...
+		}
+	}
+
+	if err := common.CopyFile(src, dst); err != nil {
+		d.logger.Error("failed to copy bundle",
+			"err", err,
+			"src", src,
+			"dst", dst,
+		)
+		return fmt.Errorf("failed to open bundle: %w", err)
```
martintomazic, line 452: NIT:
```suggestion
		return fmt.Errorf("failed to copy bundle: %w", err)
```

go/runtime/bundle/registry.go
martintomazic: Agree with the suggested comment.

Question: Why is it problematic to receive a new bundle (defined by manifest), that has a component whose version is already in the registry? Isn't this too prohibitive as it does not allow to (re)-use existing components?

go/runtime/bundle/discovery.go
```diff
@@ -0,0 +1,475 @@
...
+func (d *Discovery) Init() error {
+	// Consolidate all bundles in one place, which could be useful
+	// if we implement P2P sharing in the future.
+	if err := d.copyBundles(); err != nil {
+		return err
+	}
+
+	// Add copied and cached bundles to the registry.
+	if err := d.Discover(); err != nil {
+		return err
```
martintomazic, line 85: `Init` only adds configured bundles to the registry. In addition, some of them may be already previously exploded (cached)? If this is the case somewhere nested `WriteExploded` will only verify...

However, it does not guarantee to add all the "cached" bundles, as looking at `Discover` method you only add the `*.orc` bundles to the registry (i.e. the ones that were configured). This is desired as some of the exploded bundles may be stale, i.e. no longer part of the config.  (Automatic removal will be handled as part of #5737).

Do I understand your logic correctly? If so I would suggest removing `cached bundles` reference from the `Init` toplevel docs and before `d.Discover()` to avoid confusion?

Thanks for your reply! :)

go/runtime/bundle/discovery.go
```diff
@@ -0,0 +1,452 @@
...
+		case <-ticker.C:
+		case <-d.discoverCh:
+		case <-ctx.Done():
+			d.logger.Info("stopping discovery")
+			return
+		}
+
+		_ = d.Discover()
+		d.Download()
+	}
```
martintomazic, line 147: In theory we could sent downloaded bundles (paths) over the channel directly. This way there would be no need to `os.ReadDir(d.bundleDir)`, with further check for every bundle if already part of registry,  for every tick? 

If we assume it is mandatory to call `init` before starting the `*Discovery` service, we could remove `d.Discover()` altogether from this function.

But then... the tick is every 10 minutes anyways so this is not really a performance bottleneck.

go/runtime/bundle/discovery.go
```diff
@@ -0,0 +1,452 @@
...
+	d.mu.RUnlock()
+
+	downloaded := make(map[hash.Hash]struct{})
+	for _, hash := range hashes {
+		if err := d.downloadBundle(runtimeID, hash); err != nil {
+			d.logger.Error("failed to download bundle",
+				"err", err,
+				"runtime_id", runtimeID,
+				"manifest_hash", hash.Hex(),
+			)
```
martintomazic, line 259: Looking at `d.downloadBundle` we join the errors from all the downloads, meaning unless the first `baseUrl` is success, we would receive error here even if we successfully download the bundle using e.g. third baseUrl.

I know we plan to start with single url for the start so this should not be problematic for now, but might become misleading in the future?

martintomazic (commented):
I have checked new discovery and registry part. Overall looks super solid. Will try to review the whole PR/integration tmr :)

go/runtime/bundle/registry.go
peternose: What do you do when two components have the same version but different values (e.g. `Disabled`)? 
For now, we can probably be more strict, and we can loosen this latter.

go/runtime/bundle/discovery.go
peternose: All bundles are copied to the bundle folder.:
- Bundles from the config are copied in `Init` on line 79.
- Bundles fetched from the web are first verified and then copied to the bundle folder.

So the `Discovery` loads all bundles in the config, all bundles that used to be in the config, and all bundles that were downloaded. The idea is that you always have bundle and the extracted folder in the bundle directory. This way, it will be easy to implement p2p2 sharing as one can just check which orc files are in the bundle directory.

go/runtime/bundle/registry.go
```diff
@@ -0,0 +1,346 @@
...
+
+		// Detached components are explicit and they should be enabled by default.
+		if comp.Detached {
+			return true
+		}
+
+		// On non-compute nodes, assume all components are disabled by default.
+		if config.GlobalConfig.Mode != config.ModeCompute {
+			return false
+		}
```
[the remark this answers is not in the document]
peternose: Probably. This was copied, only the order was reversed. 

@kostko Should we return `false` for detached components on compute nodes?

go/runtime/bundle/registry.go
```diff
@@ -0,0 +1,346 @@
...
+	// are added to the registry.
+	WatchVersions(runtimeID common.Namespace) (<-chan version.Version, *pubsub.Subscription)
+
+	// GetManifests returns all known manifests that contain RONL component.
+	GetManifests() []*Manifest
+
+	// GetName returns optional human readable runtime name.
+	GetName(runtimeID common.Namespace, version version.Version) (string, error)
+
+	// GetComponents returns components for the given runtime and version.
```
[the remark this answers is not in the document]
peternose: No, this methods returns all components, e.g. one RONL (version matches) and many ROFLs (the latest version).

go/runtime/bundle/discovery.go
peternose: Could probably do that, but not sure if it is worth. Will think about it.

What I want is to be able to call `Discover` and `Download` without running the discovery, which s useful for tests.

go/runtime/bundle/discovery.go
peternose: No. Observe that we only return error if all downloads fail. So if the first two fail and the third one doesn't, the `downloadBundle` will return `nil`.

martintomazic: Ah my bad. sorry!

go/runtime/bundle/registry.go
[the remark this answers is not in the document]
martintomazic: See the line 51 in the suggestion? I think it might be worth to emphasize that remaining components are latest and not for the version that we pass.

go/runtime/bundle/discovery.go
martintomazic: I see and thanks.

So what I suggest to change (super minor but to avoid confusion): 
```
// Init sets up bundle discovery using node configuration and adds configured
// and cached bundles to the registry.
```
and 
```go
// Add copied and cached bundles to the registry.
```

to
```
// Init sets up bundle discovery using node configuration and adds configured
// bundles (that are guaranteed to be exploded) to the registry.
```
and 
```go
// Add copied bundles (that are guaranteed to be exploded) to the registry.
```

or smtg along the lines, just to emphasize that we are not also moving stale cached bundles (see issue #5737)

Hope it makes sense(?), and feel free to reword as you see fit...

go/runtime/bundle/registry.go
martintomazic: Good point. Feel free to close from my side.

go/runtime/bundle/discovery.go
martintomazic: Agree. Unless trivial to change feel free to keep as is :)

go/runtime/bundle/registry.go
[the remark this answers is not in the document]
peternose: Ups, missed. Looks good, will change.

go/runtime/config/config.go
```diff
@@ -71,21 +72,29 @@ const (
...
+
 	// The runtime environment (sgx, elf, auto).
-	Environment RuntimeEnvironment `yaml:"environment"`
+	Environment RuntimeEnvironment `yaml:"environment,omitempty"`
 
 	// History pruner configuration.
 	Prune PruneConfig `yaml:"prune,omitempty"`
 
 	// RuntimeConfig maps runtime IDs to their respective local configurations.
+	// NOTE: This may go away in the future, use `RuntimeConfig.Config` instead.
```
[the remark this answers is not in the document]
peternose: That was OK. The users should use `Config` field in `type RuntimeConfig struct` .

go/runtime/bundle/discovery.go
martintomazic: > all bundles that used to be in the config

Correct since at init time the `.orc` file (if configured) was moved to `/data/runtimes/bundles` and exploded. 
Note, the nodes that were running previous software (prior to this PR/current version), may have some bundles exploded, later removed them from config (and thus also bundle), so only the exploded part will stay. I guess with #5737 the idea was to remove those stale "exploded" bundles + the ones whose version is now outdated.

Now if we push for the p2p sharing, I guess we go in the other direction, i.e. we further increase the storage load on the nodes as they store essentially all the historical bundles and their exploded paths (technically we could be removing those once outdated and only store bundles). On the positive side we get a more resilient data availability for our software components.

Thinking aloud, if we want to have some kind of p2p sharing we may want only archive nodes responsible for that?  For the normal nodes I don't think we want to store outdated bundles and exploded components. In any case we may want to adapt #5737 cleanup requirements?

Happy to tackle this myself in the follow-up once we align on the strategy.

cc @kostko

go/runtime/host/sandbox/sandbox.go
```diff
@@ -369,18 +369,19 @@ func (r *sandboxedRuntime) startProcess() (err error) {
...
+
+	rtVersion, err := pc.InitHost(initCtx, conn, hi)
+	if err != nil {
 		return fmt.Errorf("failed to initialize connection: %w", err)
 	}
 
-	if r.rtCfg.Components[0].IsRONL() {
+	if id := r.rtCfg.Components[0]; id.IsRONL() {
 		// Make sure the version matches what is configured in the bundle. This check is skipped for
 		// non-RONL components to support detached bundles.
```
kostko, line 380: Since this version is now per-component, it would probably be fine to check for all components. But probably we need to leave it like this for backwards compatibility until all bundles are using per-component versions.

go/runtime/registry/config.go
```diff
@@ -128,7 +128,8 @@ func newRuntimeConfig(dataDir string) (map[common.Namespace]map[version.Version]
 			if runtimes[id] == nil {
 				runtimes[id] = make(map[version.Version]*runtimeHost.Config)
 			}
-			if _, ok := runtimes[id][bnd.Manifest.Version]; ok {
+			version := bnd.Manifest.GetComponentByID(component.ID_RONL).Version
```
kostko: Maybe add a comment that the component cannot be nil because regular bundles are the ones that have the RONL component which was checked above.

go/runtime/bundle/manifest.go
```diff
@@ -226,6 +227,24 @@ type Identity struct {
 	Enclave sgx.EnclaveIdentity `json:"enclave"`
 }
 
+// ExplodedComponent is an exploded runtime component ready for execution.
```
kostko, line 248: Should this be in `bundle.go` instead of `manifest.go`? To me it seems like that would make more sense.

go/runtime/bundle/manifest.go
```diff
@@ -226,6 +227,24 @@ type Identity struct {
...
+type ExplodedComponent struct {
+	*Component
+
+	// Detached returns true iff the bundle containing the component does not
```
kostko, line 252: Should probably be `// Detached is true iff ...`.

go/runtime/bundle/discovery.go
kostko: There is no sense in storing or sharing historic bundles, so any pruning should get rid of them IMO.

go/runtime/bundle/discovery.go
```diff
@@ -0,0 +1,452 @@
...
+	if err != nil {
+		d.logger.Error("failed to construct URL",
+			"err", err,
+			"base_url", baseURL,
+			"filename", filename,
+		)
+		return fmt.Errorf("failed to construct URL: %w", err)
+	}
+
+	src, err := d.fetchBundle(url)
```
kostko, line 319: So this scheme assumes that all bundles are directly downloadable in the registry by their hash. I would like to avoid the need to copy bundles and instead do it like it was suggested in https://github.com/oasisprotocol/oasis-core/issues/5755#issuecomment-2363506316, namely that there would be an indirection step:
* You construct the URL based on the registry base URL and the manifest hash.
* But instead of containing the entire bundle, this URL only contains metadata on where to find the bundle.

This way we can use existing GitHub releases and just point to those releases in these metadata files instead of needing to copy over the actual bundles.

go/runtime/bundle/registry.go
```diff
@@ -0,0 +1,346 @@
...
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	r.logger.Info("adding bundle",
+		"path", path,
+		"manifest_hash", manifestHash,
+	)
+
+	// Open the bundle and release resources when done.
+	bnd, err := Open(path)
```
kostko: This would ideally use the "open with known hash verification" step where the manifest hash is checked before doing anything else (e.g. even parsing the manifest).

go/runtime/bundle/registry.go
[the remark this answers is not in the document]
kostko: No, it is fine if detached components are always enabled when they are explicitly specified in the config.

go/runtime/bundle/discovery.go
```diff
@@ -28,6 +28,9 @@ const (
 
 	// requestTimeout is the time limit for http client requests.
 	requestTimeout = 10 * time.Second
+
+	// maxBundleSizeBytes is the maximum allowed bundle size in bytes.
+	maxBundleSizeBytes = 20 * 1024 * 1024 // 20 MB
```
kostko, line 33: Can this be made configurable in the configuration file? As long as this is only used for RONLs, it may be fine, but ROFL bundles will be much larger and once this mechanism is also used for ROFLs it will become a problem.

go/runtime/bundle/registry.go
kostko: I see this is already implemented in a later commit, resolving.

go/runtime/bundle/bundle.go
```diff
@@ -518,7 +518,9 @@ func (bnd *Bundle) Close() error {
 }
 
 // Open opens and validates a runtime bundle instance.
-func Open(fn string) (*Bundle, error) {
+func Open(fn string, opts ...OpenOption) (*Bundle, error) {
+	options := NewOpenOptions(opts...)
+
 	r, err := zip.OpenReader(fn)
```
kostko, line 524: Unrelated to this PR, but just noticed that if this fails below, the reader is never closed.

go/runtime/bundle/discovery.go
kostko: BTW, we should do some bundle fuzzing, just to be sure these things are somewhat robust. Will open an issue.

kostko: #5969

martintomazic: @kostko I think in this context it would be worth adding some additional doc of the current and intendend status of bundles.

Here is what I have in mind:
* right now we have legacy and non-legacy bundles (components)
* furthermore we will have two ways of adding bundles (via the `paths` or configuring them to download).
* finally, there is this things with versions that is not 100% clear yet (what is allowed to be reused, what if you get same version but different content, exact bundle validation).

I think it may be worth aggregating requirements/invariant somewhere together with legacy context, as in addition to make it more accessible it would also ease us in finding problematic corner cases/interleaving. Maybe a `README.md` under `go/runtime/bundle`? We can later move this to official docs once those features are released?

Hope I am not complicating, we could also offline :).

peternose: I think that runtime bundles don't need to be compatible with all previous versions, so I expect we will remove all these legacy fields once we are done refactoring and all runtimes upgrade to new bundles. So I'm not sure if we need a readme file.

go/runtime/bundle/discovery.go
peternose: Yes, all exploded folders without orc file and all orc files that have historic components should be removed.

Is a bundle historic if it contains RONL 1.0 and ROFL 1.0 and there is a new version of the bundle containing RONL1.1 (which can fetch ROFL 1.0 from the previous bundle)?

kostko: > Is a bundle historic if it contains RONL 1.0 and ROFL 1.0 and there is a new version of the bundle containing RONL1.1 (which can fetch ROFL 1.0 from the previous bundle)?

Only once RONL 1.1 is the active version.

Also note that ROFL components can only be fetched from detached bundles, not from bundles that contain RONLs. If there is a "system ROFL component" (e.g. one that is included in the non-detached bundle) then the expectation is that the new version will include it as well. If it does not, that means that the component has been removed, so in your case the 1.0 version can be pruned.

go/runtime/bundle/discovery.go
peternose: I was thinking about that. I changed this to be configurable with default of 20MB.

go/runtime/bundle/manifest.go
peternose: Agree. But maybe it is even clearer to move it to a component file. So I did that.

go/runtime/registry/config.go
peternose: This was latter changed (moved to registry) so no need to add a comment.

go/runtime/host/sandbox/sandbox.go
peternose: Do ROFLs send you the right version in the `RuntimeInfoResponse`? Maybe we open a new issue for that.

go/control/api/api.go
```diff
@@ -188,6 +193,37 @@ type RuntimeStatus struct {
...
+
+	// Version is the component version.
+	Version version.Version `json:"version,omitempty"`
+
+	// Detached specifies whether the component was in a detached bundled.
+	Detached bool `json:"detached,omitempty"`
+
+	// Disabled specifies whether the component is disabled by default
+	// and needs to be explicitly enabled via node configuration to be used.
+	Disabled bool `json:"disabled,omitempty"`
```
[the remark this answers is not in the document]
peternose: This field is under section `bundle`, so it should tell you if the component is disabled in the bundle. If this setting is overridden in the runtime configuration, this should belong to `runtime` section.

The idea is to display available components, so that you can prepare configuration.

> If this is true, is the component currently disabled or was it just disabled by default and overridden in configuration?

I thought it is obvious from `needs to be explicitly enabled via node configuration to be used` that it is disabled by default.

go/runtime/bundle/discovery.go
peternose: OK, this make sense. RONL + system components in one bundle, and ROFL components in detached bundles.

> Also note that ROFL components can only be fetched from detached bundles, not from bundles that contain RONLs.

Our code doesn't distinguish between ROFL and system components. But it would be nice to have a check that ROFLs are always in a detached bundle, and system components in the runtime bundle.

peternose: Should we add `Kind = "rofl-system"` or  `Kind = "system"` and verify this in the code?

peternose: Ups, maybe we don't need this and we just mark the component as system ROFL once is read from an undetached bundle.

kostko: So "system ROFL component" = non-detached ROFL. There is already the detached flag for this, but maybe it is not obvious? Or there could be an additional component attribute (e.g. `system`), which would only be allowed to be set if the bundle is non-detached?

go/control/api/api.go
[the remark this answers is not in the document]
kostko: Ok.

go/runtime/host/sandbox/sandbox.go
kostko: They do. Just that previously there was only one version in the manifest which corresponded to the RONL component version.

.changelog/5962.cfg.md
```diff
@@ -23,4 +23,6 @@ The following configuration options have been added:
    used to fetch runtime bundles,
 
 - `runtime.repositories` is the list of global URLs used to fetch
-   runtime bundles.
+   runtime bundles,
+
+- `runtime.max_bundle_size_bytes` is the maximum allowed bundle size in bytes.
```
kostko, line 28: We also have the `ParseSizeInBytes` helper to support standard suffixes and then this can be called `max_bundle_size` (without the `_bytes`). See existing config options.

go/runtime/bundle/discovery.go
```diff
@@ -29,8 +29,9 @@ const (
 	// requestTimeout is the time limit for http client requests.
 	requestTimeout = 10 * time.Second
 
-	// maxBundleSizeBytes is the maximum allowed bundle size in bytes.
-	maxBundleSizeBytes = 20 * 1024 * 1024 // 20 MB
+	// maxDefaultBundleSizeBytes is the maximum allowed default bundle size
+	// in bytes.
+	maxDefaultBundleSizeBytes = 20 * 1024 * 1024 // 20 MB
 )
```
martintomazic: Minor: Will `requestTimeout` of 10s be sufficient for big bundles or do we also want to make this configurable/function of size?

martintomazic: I would say for the request to complete since we use it [to set](https://github.com/oasisprotocol/oasis-core/blob/979c1c210b824dbb9c1bb4abac7d46dabd6cc5e5/go/runtime/bundle/discovery.go#L62) `http.Client.Timeout`
```
       // Timeout specifies a time limit for requests made by this
	// Client. The timeout includes connection time, any
	// redirects, and reading the response body. The timer remains
	// running after Get, Head, Post, or Do return and will
	// interrupt reading of the Response.Body.
```

martintomazic: What is the max size we anticipate? For the reference I see that our documentation recommends "Recommended: 1 Gbps internet connection with low latency".

Might be a good idea to future proof us by making this parametric...

martintomazic: The more I think about it there is also the following risk:

Compute nodes may also run runtimes developed by other entities. I they are intentionally or non-intentionally compromised, a bad actor could cheaply post new deployments versions to registry (L1 transaction) that would reference big bundles. In theory you could then fill-up the compute nodes (that are subscribed/run compromised runtime) disk fast.

Worst case if against suggestion they also run validator instance as part of their compute node you would halt down validator instance + runtimes, otherwise only runtimes.

Now we could limit/monitor the total storage used inside `/bundles` dir to avoid this case (can be also config parameter). Unless we do it per-runtime, there is still a risk of this attack that halt all the runtimes that compute node is running? Finally, in this context it might be useful to have a flag that allows/dis-allows automatic downloads of runtime bundles. 

Unlikely scenario but might be worth protecting us against in the follow-up? @peternose

peternose: I increased timeout to 1 minute.

Since only runtime bundles that include RONL component and maybe few system components are downloaded through the client, the bundles should not be large, at least not in the near future. The size of the latest sapphire bundle is 7.45 MB, so limits 20 MB and 1 minute should suffice. 

I'm not worried about your threat model because we have very little runtimes in production. If someone can post new deployments to the registry this is a much bigger problem than big bundles. Actually, it would be great if big bundles would stop your runtime until this is resolved. And lastly, bundle cleanup and bundle limitations will ensure that a node will have only the latest and sometimes second-latest runtime bundle stored locally, so this should be approx. 100 MB (including exploded files).

go/runtime/bundle/discovery.go
peternose: Added indirection step. Now, metadata file containing the URL is downloaded first, and then the bundle.

kostko (approved):

PR #5971, opened 2024-12-12
go/worker/keymanager: Generate secrets immediately with mock beacon

peternose:
Forcing `url` upgrade to 2.5.4 and waiting for new releases of `tendermint-rpc` and `tendermint-config`.

```
Crate:     idna
ID:        RUSTSEC-2024-0421
idna 0.5.0
└── url 2.5.2
    ├── tendermint-rpc 0.39.1
    └── tendermint-config 0.39.1
```

Forcing also `hashbrown` upgrade to 0.15.1 and waiting for new version of `lru`.

```
cargo update -p url
cargo update -p hashbrown@0.15.2
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1275,9 +1275,9 @@ checksum = "e5274423e17b7c9fc20b6e7e208532f9b19825d82dfd615708b70edd83df41f1"
 
 [[package]]
 name = "hashbrown"
-version = "0.15.0"
+version = "0.15.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e087f84d4f86bf4b218b927129862374b72199ae7d8657835f1e89000eea4fb"
+checksum = "bf151400ff0baff5465007dd2f3e717f3fe502074ca563069ce3a6629d07b289"
 dependencies = [
  "allocator-api2",
  "equivalent",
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -3270,28 +3411,23 @@ dependencies = [
...
-[[package]]
-name = "tinyvec_macros"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
-
 [[package]]
 name = "tokio"
-version = "1.40.0"
+version = "1.42.0"
```
ptrus, line 3425: It looks like all packages were updated? (since I don't think tokio is a dependency of `url` or `hashbrown`?).

Btw i think it's fine to update, it's just not mentioned.

peternose: Ups, no that's not fine. Fill fix.

peternose: Or maybe is. I hardcoded versions of url and hashbrown, and ran cargo update, so that this will be upgraded automatically once new versions come out.

ptrus: Can't you do:
```
cargo update -p url
cargo update -p hashbrown
```
if you want to update just these two? (and their dependencies).

peternose: Yes, saw that and making changes

ptrus (approved):

PR #5975, opened 2024-12-17
go/runtime/registry: Simplify creation of provisioners

peternose:
Trying to remove the bundle registry dependency from the `createProvisioner` function, as nodes should generally not have any bundles stored locally upon their initial startup.

These changes also decouple the creation of provisioners from TEE selection. Previously, only client nodes could choose the TEE in which a component would run. Now, node operators can specify the TEE for each component, provided the component supports the chosen TEE. If no TEE is specified, it is automatically selected, with TDX and SGX taking precedence over ELF.

How to configure nonSGX client:
```yaml
runtime:
  paths: # Can be replaced with bundle repository url.
    - {{ runtime_orc_path }} 
  runtimes:
  - id: {{ runtime_id }}
    components:
    - id: ronl
      tee: none # Do not run in SGX or TDX!
```

How to configure SGX client:
```yaml
runtime:
  paths: # Can be replaced with bundle repository url, if you also specify runtime ids.
    - {{ runtime_orc_path }}
  sgx_loader: /node/bin/oasis-core-runtime-loader
```

How to mock SGX:
```yaml
runtime:
  paths: # Can be replaced with bundle repository url, if you also specify runtime ids.
    - {{ runtime_orc_path }}
  debug_mock_tee: true
```

### DIFF
--- a/.buildkite/code.pipeline.yml
+++ b/.buildkite/code.pipeline.yml
@@ -354,7 +354,7 @@ steps:
       - .buildkite/scripts/test_e2e.sh --timeout 20m --scenario e2e/runtime/runtime-encryption
     env:
       OASIS_TEE_HARDWARE: intel-sgx
-      OASIS_UNSAFE_MOCK_SGX: "1"
+      OASIS_UNSAFE_MOCK_TEE: "1"
       OASIS_UNSAFE_SKIP_AVR_VERIFY: "1"
       OASIS_E2E_COVERAGE: enable
       TEST_BASE_DIR: /tmp
--- a/.changelog/5975.cfg.md
+++ b/.changelog/5975.cfg.md
@@ -1,0 +1,47 @@
+go/runtime/config: Support selection of TEE kind
+
+The node operator can now specify the kind of Trusted Execution Environment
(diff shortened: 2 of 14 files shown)


ptrus:
> Do we run client with RuntimeEnvironmentELF and without DebugDontBlameOasis flag set?

Yes I believe so: [For example, running a ParaTime client node for an SGX-enabled ParaTime like Cipher doesn't require having SGX on the ParaTime client node.](https://docs.oasis.io/node/run-your-node/paratime-client-node).

.buildkite/code.pipeline.yml
```diff
@@ -354,7 +354,7 @@ steps:
       - .buildkite/scripts/test_e2e.sh --timeout 20m --scenario e2e/runtime/runtime-encryption
     env:
       OASIS_TEE_HARDWARE: intel-sgx
-      OASIS_UNSAFE_MOCK_SGX: "1"
+      OASIS_UNSAFE_MOCK_TEE: "1"
```
kostko, line 357: Should probably mention this in the internal changelog as this will require changes in other repositories (e.g. SDK).

kostko (approved):

kostko (approved):

PR #5977, opened 2024-12-17
go/runtime/bundle/component: Add ELF metadata

peternose:
Refactored elf executable so that ELF, SGX and TDX have the same structure. Changes should be backwards compatible. `ELF.Executable` overwrites `Executable` if both are set.


go/runtime/bundle/component.go
```diff
@@ -40,8 +41,12 @@ type Component struct {
 	Version version.Version
 
 	// Executable is the name of the runtime ELF executable file if any.
+	// NOTE: This may go away in the future, use `ELFMetadata` instead.
```
kostko, line 44:
```suggestion
	// NOTE: This may go away in the future, use `ELF` instead.
```

kostko (approved):

PR #5981, opened 2024-12-20
go/runtime/host/tdx: Add support for RW drives and formats

kostko:


go/runtime/host/tdx/qemu.go
```diff
@@ -155,10 +155,20 @@ func (q *qemuProvisioner) getSandboxConfig(rtCfg host.Config, _ sandbox.Connecto
...
+			case "raw", "qcow2":
+				// These formats are supported as-is.
+			default:
+				return process.Config{}, fmt.Errorf("format '%s' is not supported", stage2Format)
+			}
 
 			cfg.Args = append(cfg.Args,
 				// Stage 2 drive.
-				"-drive", fmt.Sprintf("format=raw,file=%s,if=none,id=drive0,read-only=on", stage2Image),
+				"-drive", fmt.Sprintf("format=%s,file=%s,if=none,id=drive0,snapshot=on", stage2Format, stage2Image),
```
ptrus, line 171: Should there be a `Stage2ReadOnly` tdx configuration parameter so that one could still have read-only if needed, or is it not useful in practice?

kostko: Not that useful as `snapshot=on` serves a similar role, eg. all modifications will go to a temporary file that will get removed (so the original image is never modified).

If there is no disk-backed ephemeral storage, then everything in stage2 will be either read-only (dm-verity/squashfs) or in memory (tmpfs).

ptrus (approved):

PR #5994, opened 2025-01-10
go/runtime/config: Add default bundle registry

peternose:
Renaming also repository to registry, as the URL doesn't necessary point to a repository.

### DIFF
--- a/go/runtime/config/config.go
+++ b/go/runtime/config/config.go
@@ -131,8 +131,8 @@ type Config struct {
 	// LoadBalancer is the load balancer configuration.
 	LoadBalancer LoadBalancerConfig `yaml:"load_balancer,omitempty"`
 
-	// Repositories is the list of URLs used to fetch runtime bundle metadata.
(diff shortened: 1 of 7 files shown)


go/runtime/config/config.go
```diff
@@ -184,8 +188,8 @@ type RuntimeConfig struct {
 	// Config contains runtime local configuration.
 	Config map[string]interface{} `yaml:"config,omitempty"`
 
-	// Repositories is the list of URLs used to fetch runtime bundle metadata.
-	Repositories []string `yaml:"repositories,omitempty"`
+	// Registries is the list of URLs used to fetch runtime bundle metadata.
```
ptrus, line 191: Should probably add a comment on how the bundle URLs are constructed:
https://github.com/oasisprotocol/oasis-core/blob/92f18910b90544ce9707c454c6a23eb77069beac/go/runtime/bundle/discovery.go#L315

and that the provided base URL doesn't need to be a valid endpoint, just the constructed ones.

kostko (approved):

PR #5995, opened 2025-01-10
go/runtime/bundle: Add some more metadata to manifest

kostko:
Currently not used anywhere yet, but will be used later on. Adding it in early so that the runtime build side (e.g. Oasis CLI) can use it.

### DIFF
--- a/go/runtime/bundle/component.go
+++ b/go/runtime/bundle/component.go
@@ -248,6 +248,8 @@ type TDXResources struct {
 	Memory uint64 `json:"memory"`
 	// CPUCount is the requested number of vCPUs.
 	CPUCount uint8 `json:"cpus"`
+	// GPU is the optional GPU resource configuration.
+	GPU *GPUResource `json:"gpu,omitempty"`
 }
 
 // Validate validates the VM resources.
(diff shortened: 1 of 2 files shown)


go/runtime/bundle/component.go
```diff
@@ -255,6 +260,35 @@ func (r *TDXResources) Validate() error {
 	if r.CPUCount < 1 {
 		return fmt.Errorf("vCPU count must be at least 1")
 	}
+	if r.GPU != nil {
+		err := r.GPU.Validate()
+		if err != nil {
```
peternose, line 265:
```suggestion
		if err := r.GPU.Validate(); err != nil {
```

go/runtime/bundle/component.go
```diff
@@ -255,6 +260,35 @@ func (r *TDXResources) Validate() error {
...
+			return err
+		}
+	}
+	return nil
+}
+
+const (
+	// GPUKindNvidiaH100 is the NVIDIA H100 GPU kind.
+	GPUKindNvidiaH100 = "nvidia-h100"
```
peternose, line 274: Can we remove `Kind`?

go/runtime/bundle/component.go
```diff
@@ -255,6 +260,35 @@ func (r *TDXResources) Validate() error {
...
+	// GPUKindNvidiaH200 is the NVIDIA H200 GPU kind.
+	GPUKindNvidiaH200 = "nvidia-h200"
+)
+
+// GPUResource is the GPU resource descriptor.
+type GPUResource struct {
+	// Kind is the GPU kind. It may be omitted to specify that the kind is not important.
+	Kind string `json:"kind,omitempty"`
```
peternose, line 282: I would not use `Kind`, as is too general. Maybe `Model`, `GPU`, ...

peternose (approved):

go/runtime/bundle/component.go
kostko: I believe this will cause a lint issue due to nested ifs.

go/runtime/bundle/component.go
kostko: Yes, we could just remove it for now.

peternose: I should have been more specific. What I meant is whether we can remove `Kind` from the name and change it to `GPUNvidiaH100` or `gpu.NvidiaH100` instead.

peternose (approved):

PR #6005, opened 2025-01-17
go/runtime/host/tdx: Add support for persistent image overlay

### LINKED ISSUE #5999: Support persistent storage for TDX
There should be support for persistent storage for TDX. It would simply create a separate mutable stage 2 image using `qemu-img` and use the immutable stage 2 image as the backing image.

kostko:
Closes #5999

### DIFF
--- a/go/runtime/bundle/component.go
+++ b/go/runtime/bundle/component.go
@@ -3,11 +3,22 @@ package bundle
 import (
 	"fmt"
 	"path/filepath"
+	"regexp"
 
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/sgx"
 	"github.com/oasisprotocol/oasis-core/go/common/version"
 	"github.com/oasisprotocol/oasis-core/go/runtime/bundle/component"
+)
+
+// componentNameRegexp is the regular expression for valid component names.
+var componentNameRegexp = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
+
+const (
+	// minComponentNameLen is the minimum length of a valid component name.
+	minComponentNameLen = 3
+	// maxComponentNameLen is the maximum length of a valid component name.
+	maxComponentNameLen = 128
 )
 
 // ExplodedComponent is an exploded runtime component ready for execution.
@@ -118,6 +129,15 @@ func (c *Component) Validate() error {
 			return fmt.Errorf("RONL component cannot be disabled")
 		}
 	case component.ROFL:
+		if len(c.Name) < minComponentNameLen {
+			return fmt.Errorf("ROFL component name must be at least %d characters long", minComponentNameLen)
+		}
+		if len(c.Name) > maxComponentNameLen {
+			return fmt.Errorf("ROFL component name must be at most %d characters long", maxComponentNameLen)
+		}
+		if !componentNameRegexp.MatchString(c.Name) {
+			return fmt.Errorf("ROFL component name is invalid (must satisfy: %s)", componentNameRegexp)
+		}
 	default:
 		return fmt.Errorf("unknown component kind: '%s'", c.Kind)
 	}
--- a/go/runtime/bundle/component_test.go
+++ b/go/runtime/bundle/component_test.go
@@ -1,0 +1,40 @@
+package bundle
+
+import (
+	"strings"
+	"testing"
+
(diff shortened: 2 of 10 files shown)


go/runtime/host/tdx/qemu.go
```diff
@@ -211,6 +232,47 @@ func (q *qemuProvisioner) getSandboxConfig(rtCfg host.Config, _ sandbox.Connecto
...
+func (q *qemuProvisioner) createPersistentOverlayImage(
+	rtCfg host.Config,
+	comp *bundle.ExplodedComponent,
+	image string,
+	format string,
+) (string, error) {
+	compID, _ := comp.ID().MarshalText()
+	imageDir := filepath.Join(q.cfg.DataDir, persistentImageDir, rtCfg.ID.String(), string(compID))
+	imageFn := filepath.Join(imageDir, fmt.Sprintf("%s.overlay", filepath.Base(image)))
+	if _, err := os.Stat(imageFn); err == nil {
```
peternose, line 248:
```suggestion
	if _, err := os.Stat(imageFn); err == nil { // err is nil
```
Add some comment to notify the reader that this is non-idiomatic check, as one expects `err != nil`.

You should thrown an error if `Stat` failed and create a file when doesn't exist.

go/runtime/host/tdx/qemu.go
```diff
@@ -211,6 +232,47 @@ func (q *qemuProvisioner) getSandboxConfig(rtCfg host.Config, _ sandbox.Connecto
...
+		"-f", "qcow2",
+		"-b", image,
+		"-F", format,
+		imageFn,
+	)
+	var out strings.Builder
+	cmd.Stderr = &out
+	cmd.Stdout = &out
+	if err := cmd.Run(); err != nil {
+		return "", fmt.Errorf("failed to create persistent overlay image: %w\n%s", err, out.String())
```
peternose, line 270: Error should be at the end.

go/runtime/bundle/component_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/oasisprotocol/oasis-core/go/runtime/bundle/component"
+)
+
+func TestComponentValidation(t *testing.T) {
+	require := require.New(t)
+
+	comp := Component{}
```
peternose, line 15:
```suggestion
	var comp Component
```
Zero values should be defined like this, its more readable.

Could also use table-test. I would add comment at the top that this section only tests names.

go/runtime/host/tdx/qemu.go
kostko: Fixed.

go/runtime/host/tdx/qemu.go
kostko: Fixed.

go/runtime/bundle/component_test.go
kostko: Fixed.

go/runtime/bundle/component.go
```diff
@@ -3,13 +3,24 @@ package bundle
...
+	"regexp"
 
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/sgx"
 	"github.com/oasisprotocol/oasis-core/go/common/version"
 	"github.com/oasisprotocol/oasis-core/go/runtime/bundle/component"
 )
 
+// validComponentNameRegexp is the regular expression for valid component names.
+var validComponentNameRegexp = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
```
peternose, line 15:
```suggestion
var componentNameRegexp = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
```

go/runtime/bundle/component_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/oasisprotocol/oasis-core/go/runtime/bundle/component"
+)
+
+func TestComponentValidation(t *testing.T) {
+	require := require.New(t)
+
+	var comp Component
+	err := comp.Validate()
+	require.ErrorContains(err, "unknown component kind")
+	comp.Kind = component.ROFL
```
peternose, line 18: Could move under `// Test component names.`

peternose (approved):
Nice

PR #6007, opened 2025-01-20
go/runtime/bundle/manifest: Fix legacy manifest version

peternose:
The current solution for handling legacy manifest version updated the RONL component version after the bundle was opened. This was not the best solution as that changed the manifest hash. Furthermore, if the manager read the manifest directly from the exploded data directory, the RONL component version did't not change, causing problems if the node was restarted with cached bundles.

### DIFF
--- a/go/oasis-node/cmd/node/node_control.go
+++ b/go/oasis-node/cmd/node/node_control.go
@@ -13,6 +13,7 @@ import (
 	cmdFlags "github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common/flags"
 	p2p "github.com/oasisprotocol/oasis-core/go/p2p/api"
 	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
+	"github.com/oasisprotocol/oasis-core/go/runtime/bundle/component"
 	storage "github.com/oasisprotocol/oasis-core/go/storage/api"
 	upgrade "github.com/oasisprotocol/oasis-core/go/upgrade/api"
 	keymanagerWorker "github.com/oasisprotocol/oasis-core/go/worker/keymanager/api"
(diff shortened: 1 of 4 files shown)


go/oasis-node/cmd/node/node_control.go
```diff
@@ -359,7 +360,12 @@ func (n *Node) getBundleStatus() ([]control.BundleStatus, error) {
 	bundles := make([]control.BundleStatus, 0, len(manifests))
 
 	for _, manifest := range manifests {
-		explodedComponents, err := bundleRegistry.GetComponents(manifest.ID, manifest.GetVersion())
+		ronl := manifest.GetComponentByID(component.ID_RONL)
+		if ronl == nil {
+			return nil, fmt.Errorf("manifest has no RONL component")
```
kostko, line 365: Ideally these errors should not cause the entire node status to become unavailable.

go/runtime/bundle/manifest.go
```diff
@@ -124,20 +138,24 @@ func (m *Manifest) GetComponentByID(id component.ID) *Component {
...
-func (m *Manifest) GetVersion() version.Version {
-	// We also support legacy manifests which define version at the top-level.
-	for _, comp := range m.Components {
-		if !comp.ID().IsRONL() {
-			continue
-		}
+// UnmarshalJSON customizes the unmarshalling of the manifest.
+func (m *Manifest) UnmarshalJSON(b []byte) (err error) {
+	// Unmarshal into the auxiliary struct to avoid recursion.
+	type Alias Manifest
```
kostko, line 144: There should be no need for this type to be capitalized, private types should work too.

kostko (approved):

go/oasis-node/cmd/node/node_control.go
peternose: This error is not possible as the registry only returns RONL manifests. But I can skip non-ronl manifests, removing the error.

PR #6009, opened 2025-01-21
go/oasis-node/cmd: Remove config migrate command

### LINKED ISSUE #5991: Remove config migrate command
Since all users have already migrated their YAML configuration files to the new format, this command can be deleted. This will enable us to remove legacy configuration fields without affecting the migration tests.

abukosek:
All users have migrated from the old config format, so this command is no longer necessary.

Closes #5991.


peternose (approved):
Those who haven't migrated can use an older version of the Oasis node to do the migration.

Issue #6000, opened 2025-01-14
Fix runtime upgrade with attached and detached ROFLs


peternose:
Make a clear distinction between attached/detached ROFL components within a runtime and upgrade only attached during the upgrade of RONL:
- **Attached**: Same as before. Upon upgrade, all components run in duplicate, for both the old and new versions, then the old one shuts down.
- **Detached**: During the upgrade of RONL, nothing happens; components continue running without stopping (only the routing changes so that requests for RONL are directed to the new RONL).

Issue #5963, opened 2024-12-07
Support ROFL versions via CLI


peternose:
It would be highly beneficial if the ROFL could be upgraded to a different version without requiring a node restart. Additionally, it could be valuable to allow the selection of a specific ROFL version, whether it is the latest version or an older one.

Issue #5889, opened 2024-10-10
Expose component hot-loading via internal gRPC


kostko:
After #5755 brings us support for hot-loading runtime components, we should expose this functionality via internal gRPC, probably as part of the existing node control API, to upgrade detached ROFLs.

PR #6012, opened 2025-01-27
go/control/api: Support ROFL upgrades

peternose:
Until now, all components available for a runtime were upgraded simultaneously.  The upgrade first warmed up the upcoming versions, and once they became active, the old versions were replaced with the new ones. This was not the best as components were upgraded, even if the version hasn't changed.

Now, each component can be upgraded independently. RONL components are upgraded at the specified epoch, while ROFL components are upgraded immediately.

Closes:
- #5963
- #5889 
- #6000 

We will address the attached (system) ROFL components later in:
- #6018


go/runtime/host/composite/host.go
```diff
@@ -98,42 +113,108 @@ func (h *compositeHost) Call(ctx context.Context, body *protocol.Body) (*protoco
...
+	comp, ok := h.comps[id]
+	if !ok {
+		comp = multi.New(h.id)
+		if h.running {
+			comp.Start()
+		}
+		h.comps[id] = comp
+	}
+
+	if err := comp.AddVersion(version, rt); err != nil {
```
kostko, line 210: Are old versions ever removed? I guess they are set to `nil` once stopped (in `multi.Aggregate`)?

peternose: They are only removed if they are started. Will add cleanup in another PR.

go/worker/keymanager/worker.go
```diff
@@ -393,55 +394,64 @@ func (w *Worker) worker() {
...
 				return false
 			}
 		case 1:
-			version = versions[0]
+			comp = comps[0]
 		default:
-			w.logger.Error("expected a single runtime version (got %d)", numVers)
+			w.logger.Error("expected a single runtime component (got %d)", numComps)
 			return false
 		}
```
martintomazic, line 422: This is logic from before (and relates to keymanager) but I am wondering: why is it wrong If I already have multiple ronl components here? I.e. what if I already donwloaded/configured bundles with component for version that has not yet been activated?

go/runtime/registry/host.go
```diff
@@ -2,68 +2,93 @@ package registry
...
-func (n *RuntimeHostNode) ProvisionHostedRuntimeVersion(version version.Version) error {
-	if n.agg.HasVersion(version) {
+// ProvisionHostedRuntimeComponent provisions the given runtime component.
+func (n *RuntimeHostNode) ProvisionHostedRuntimeComponent(comp *bundle.ExplodedComponent) error {
+	n.mu.Lock()
+	defer n.mu.Unlock()
+
+	if !n.isComponentWanted(comp) {
+		return nil
+	}
```
martintomazic, line 60: I know this is logic form before but is this normal/do we want a warning here? Can I have it in the bundle registry, but is not wanted? I guess if I set so in the config but still keep the bundles...?

martintomazic (commented):
I have reviewed ~2/3 and looks super solid! Left some minor questions...

go/runtime/bundle/manager.go
```diff
@@ -212,6 +212,27 @@ func (m *Manager) run(ctx context.Context) {
...
+		)
+		return err
+	}
+
+	if err := m.registerManifest(manifest); err != nil {
+		m.logger.Error("failed to register manifest",
+			"err", err,
+		)
+		return fmt.Errorf("failed to register manifest: %w", err)
+	}
```
martintomazic, line 231: Nit: There is a potential for race condition. I don't see it in practice though.

Technically, a user may add regular bundle (even though this feature is mostly for rofl), right at the time clean-up is happening concurrently for the same version.

In the unlucky interleaving, clean-up removes manifest from the bundle registry, `registerManifest` adds it back, and finally clean-up cleans the exploded bundle dir. Now registry has a wrong view. Furthermore, this would trigger `ProvisionHostedRuntimeComponent` via subscription, for the component whose `ExplodedDataDir` does not exist anymore.

Again I don't see happening this in practice... May there be more obvious problems I don't see?

Finally this make me think: like bundle discovery must be enabled, do we want to do the same for the cleanup?

.changelog/6012.feature.md
```diff
@@ -0,0 +1,6 @@
+go/oasis-node/cmd: Support ROFL upgrades
+
```
martintomazic, line 2: NIT: we are also introducing new fields in the status. Should this be part of the changelog?

martintomazic (approved):
LGTM - happy to approve!

go/runtime/bundle/manager.go
martintomazic: A safer design might be to enqueu this pending explode operation just like we do for download and clean. This way we guarantee single thread operating on the bundles. `AddManifest` may even become private (to prevent other devs doing same mistake later on) making bundle registry a direct dependency of manager. Not sure might be an overkill?

go/runtime/host/composite/host.go
```diff
@@ -98,42 +113,108 @@ func (h *compositeHost) Call(ctx context.Context, body *protocol.Body) (*protoco
...
-func (h *compositeHost) Abort(ctx context.Context, force bool) error {
+// Abort implements host.Runtime.
+func (h *Host) Abort(ctx context.Context, force bool) error {
+	h.mu.RLock()
+	defer h.mu.RUnlock()
+
 	// Only RONL supports aborts.
-	return h.comps[component.ID_RONL].Abort(ctx, force)
+	return h.ronl.Abort(ctx, force)
 }
```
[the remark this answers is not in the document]
peternose: Yes. 

Field `running` means that the runtime component hosts (and their inner hosts) are running. This doesn't mean that the runtime components (binaries) are running as well. For example, if active version of a component is not available (e.g. bundle was not downloaded), the aggregate host will be running but waiting for the active version. Another example would be a component that crashes every 5 seconds. In that case, the aggregate host and the sandbox host for the active version will be running, but the component might be stopped, running, or restarting.

Calling `Abort` sends a request to the runtime component, but currently this method is not supported. Therefore, the sandbox host will kill the runtime component and restart it. So the component will be stopped for a moment, but the sandbox host for the active version, the aggregate host for that component, and the composite host for the runtime will still be running.

go/worker/keymanager/worker.go
peternose: Great question. 

The key manager runtime is a special runtime that allows you to run any version you choose. This is because it cannot be upgraded like "normal" runtimes. An upgraded version would have a different enclave identity and, as a result, would be unable to decrypt the master secret that was previously encrypted and stored locally by the previous version.

To overcome this issue, you must always start a new node (or runtime) and replicate the master secret from an old version. The policy defines who can replicate the secrets and from whom they can be replicated. So when a new deployment becomes active, the old versions must still be allowed/able to run.

Now, back to the question. If key manager node has multiple RONLs, which one should you run? Keep in mind that the old key manager node must continue running the old RONL, even if the consensus indicates that a new version is available and ready for download. 

Until now, we forced that the node must have only one static bundle, that's why are those limitations in the code. The current solution uses the same principle and will work for static bundles, if hot-loading is not enabled. If it is, things will not work for old nodes, because:
- the cleanup will remove the old version,
- there might be two RONL versions if the cleanup hasn't started.

I suggest we fix this problem in another PR (e.g. we forbid hot-loading for key manager nodes). Until then, the key manager operators must not use hot-loading.

go/runtime/registry/host.go
peternose: > I know this is logic form before but is this normal/do we want a warning here?

I would not post warnings here. If needed, the node operator can check the logs which components were provisioned/started.

> Can I have it in the bundle registry, but is not wanted?

Yes. In general, I would like to have control over which ROFL components are running, e.g. via local config. That could be useful for testing, and maybe even in production if running ROFL all the time is expensive. We could change the code to not add it to the registry if it is not wanted, but that would not allow us to enable unwanted components dynamically (not implemented yet, not sure if needed). I guess both solutions are fine. I moved the code here because this node is responsible for orchestration of the components.

go/runtime/bundle/manager.go
peternose: You are right. Yes, doing everything in one routine solves the problem, but on the other hand causes problems because the caller for `Add` would like to have answers immediately and not wait for downloads to finish. Will think about this.

> Like bundle discovery must be enabled, do we want to do the same for the cleanup?

Don't understand this question. Can you elaborate?

go/worker/keymanager/worker.go
martintomazic: Agree with the fix-up PR, feel free to open issue and close :)

go/runtime/bundle/manager.go
martintomazic: >Like bundle discovery must be enabled, do we want to do the same for the cleanup?

I am just thinking, we didn't want to force operators to have bundle discovery enabled by default (they have to configure it). Should we give them same liberty when it comes to clean-up? Are there corner cases when automatic clean-up is not desired, e.g. testing? Also gives us opportunity to shut it off if it comes across as problematic.

> Yes, doing everything in one routine solves the problem, but on the other hand causes problems because the caller for Add would like to have answers immediately and not wait for downloads to finish. 

We could technically parallelize the downloads and set reasonably low limit for download + always trigger `Add` first (albeit not much help if during downloads). But yes still wait time might be too high... I guess we will always have compromise between thread safety and blocking here.

I mean we could also store data of ongoing clean-up and downloads and then only block if overlapping, but this really would over-engineer it.

.changelog/6012.feature.md
peternose: Yes, missed that. Changes to `oasis-node` command, configuration and to `oasis-node control status` output should be in the changelog so that users know what is new and what has changed.

go/runtime/host/composite/host.go
peternose: Currently, the ROFL components cannot be stopped and if we have a ROFL component, we always run the latest version. Therefore, it doesn't make sense to remove aggregated ROFL host from the composite host. We should add this latter when needed.

We do however need to remove obsolete RONL and ROFL versions from the aggregate. This is fixed in #6028, and can be latter changed to keep some old ROFL versions in case the scheduler wants that (or maybe the scheduler just creates a new host with an older version).

PR #6013, opened 2025-01-27
go/control: Show runtime block history indexer status

### LINKED ISSUE #5998: control status: Show current history reindex round number
##### SUMMARY

Show the progress of the history reindex in `oasis control status`.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

go/control,

https://github.com/oasisprotocol/oasis-core/blob/c9acbf23b21d48f5f3de381e84deea5bebb6cd06/go/consensus/cometbft/roothash/roothash.go#L390-L527

##### ADDITIONAL INFORMATION

Related to https://github.com/oasisprotocol/oasis-core/issues/5738 it would be useful for oasis-node to report the current reindex round so that the node operator could estimate when the reindex will finish.

martintomazic:
Show block history indexer status, also closes #5998.

Rendexing:
```bash
      "indexer": {
        "status": "reindexing",
        "last_round": 3951368,
        "reindex_status": {
          "batch_size": 1000,
          "last_height": 18764680,
          "start_height": 18705681,
          "end_height": 25804727,
          "eta": "2025-03-28T19:30:00+01:00"
        }
      },
```
Indexing"

``` bash
      "indexer": {
        "status": "indexing",
        "last_round": xxxxxxx,
      },
```

### DIFF
--- a/.changelog/5998.feature.md
+++ b/.changelog/5998.feature.md
@@ -1,0 +1,18 @@
+go/oasis-node: Display runtime block history indexer status
+
+A new field `indexer` has been added to the `oasis-node control status`
+output under the runtime status section. Unless keymanager runtime,
+this field displays:
+
+ 1. The status of runtime block history indexer.
+ 2. The last indexed round.
+
+Additionally, if history reindex is in progress, it also includes:
+
+- Batch size that is used during reindex.
+- Last consensus height that was indexed.
+- Start and end heights of reindex range.
+- ETA field, which specifies expected time of reindex completion.
+
+This is useful for the node operators, so that they can estimate when their
+node will be ready to accept runtime work.
--- a/go/control/api/api.go
+++ b/go/control/api/api.go
@@ -18,6 +18,7 @@ import (
 	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
 	block "github.com/oasisprotocol/oasis-core/go/roothash/api/block"
 	"github.com/oasisprotocol/oasis-core/go/runtime/bundle/component"
+	"github.com/oasisprotocol/oasis-core/go/runtime/history"
 	storage "github.com/oasisprotocol/oasis-core/go/storage/api"
 	upgrade "github.com/oasisprotocol/oasis-core/go/upgrade/api"
 	commonWorker "github.com/oasisprotocol/oasis-core/go/worker/common/api"
@@ -184,13 +185,14 @@ type RuntimeStatus struct {
 	// LastRetainedHash is the hash of the oldest retained block.
 	LastRetainedHash hash.Hash `json:"last_retained_hash"`
 
-	// Committee contains the runtime worker status in case this node is a (candidate) member of a
-	// runtime committee.
+	// Committee contains the runtime common committee worker status.
 	Committee *commonWorker.Status `json:"committee"`
 	// Executor contains the executor worker status in case this node is an executor node.
 	Executor *executorWorker.Status `json:"executor,omitempty"`
 	// Storage contains the storage worker status in case this node is a storage node.
 	Storage *storageWorker.Status `json:"storage,omitempty"`
+	// Indexer contains the runtime history indexer status in case this runtime has a block indexer.
+	Indexer *history.IndexerStatus `json:"indexer,omitempty"`
 
 	// Provisioner is the name of the runtime provisioner.
 	Provisioner string `json:"provisioner,omitempty"`
--- a/go/oasis-node/cmd/node/node_control.go
+++ b/go/oasis-node/cmd/node/node_control.go
@@ -319,6 +319,11 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
 			}
 		}
 
+		// Fetch history indexer status.
+		if indexer, ok := n.RuntimeRegistry.Indexer(rt.ID()); ok {
+			status.Indexer = indexer.Status()
+		}
+
 		// Fetch provisioner type.
 		status.Provisioner = n.Provisioner.Name()
 
--- a/go/oasis-node/cmd/node/node_control.go
+++ b/go/oasis-node/cmd/node/node_control.go
@@ -208,6 +208,8 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
 			continue
 		}
 
+		logger := n.logger.With("runtime_id", rt.ID())
+
 		var status control.RuntimeStatus
 
 		// Fetch runtime registry descriptor. Do not wait too long for the descriptor to become
@@ -221,10 +223,7 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
 		case context.DeadlineExceeded:
 			// The descriptor may not yet be available. It is fine if we use nil in this case.
 		default:
-			n.logger.Error("failed to fetch registry descriptor",
-				"err", err,
-				"runtime_id", rt.ID(),
-			)
+			logger.Error("failed to fetch registry descriptor", "err", err)
 		}
 
 		// Fetch latest block as seen by this node.
@@ -244,10 +243,7 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
 				Hash:      blk.Header.StateRoot,
 			}
 		default:
-			n.logger.Error("failed to fetch latest runtime block",
-				"err", err,
-				"runtime_id", rt.ID(),
-			)
+			logger.Error("failed to fetch latest runtime block", "err", err)
 		}
 
 		// Fetch latest genesis block as seen by this node.
@@ -260,10 +256,7 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
 			status.GenesisRound = blk.Header.Round
 			status.GenesisHash = blk.Header.EncodedHash()
 		default:
-			n.logger.Error("failed to fetch genesis runtime block",
-				"err", err,
-				"runtime_id", rt.ID(),
-			)
+			logger.Error("failed to fetch genesis runtime block", "err", err)
 		}
 
 		// Fetch the oldest retained block.
@@ -273,10 +266,7 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
 			status.LastRetainedRound = blk.Header.Round
 			status.LastRetainedHash = blk.Header.EncodedHash()
 		default:
-			n.logger.Error("failed to fetch last retained runtime block",
-				"err", err,
-				"runtime_id", rt.ID(),
-			)
+			logger.Error("failed to fetch last retained runtime block", "err", err)
 		}
 
 		// Take storage into account for last retained round.
@@ -284,9 +274,7 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
 			lsb, ok := rt.Storage().(storage.LocalBackend)
 			switch ok {
 			case false:
-				n.logger.Error("local storage backend expected",
-					"runtime_id", rt.ID(),
-				)
+				logger.Error("local storage backend expected")
 			default:
 				// Update last retained round if storage earliest round is higher.
 				if earliest := lsb.NodeDB().GetEarliestVersion(); earliest > status.LastRetainedRound {
@@ -296,10 +284,9 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
 						status.LastRetainedRound = blk.Header.Round
 						status.LastRetainedHash = blk.Header.EncodedHash()
 					default:
-						n.logger.Error("failed to fetch runtime block",
+						logger.Error("failed to fetch runtime block",
 							"err", err,
 							"round", earliest,
-							"runtime_id", rt.ID(),
 						)
 					}
 
@@ -312,10 +299,7 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
 		if rtNode := n.CommonWorker.GetRuntime(rt.ID()); rtNode != nil {
 			status.Committee, err = rtNode.GetStatus()
 			if err != nil {
-				n.logger.Error("failed to fetch common committee worker status",
-					"err", err,
-					"runtime_id", rt.ID(),
-				)
+				logger.Error("failed to fetch common committee worker status", "err", err)
 			}
 		}
 
@@ -323,10 +307,7 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
 		if execNode := n.ExecutorWorker.GetRuntime(rt.ID()); execNode != nil {
 			status.Executor, err = execNode.GetStatus()
 			if err != nil {
-				n.logger.Error("failed to fetch executor worker status",
-					"err", err,
-					"runtime_id", rt.ID(),
-				)
+				logger.Error("failed to fetch executor worker status", "err", err)
 			}
 		}
 
@@ -334,10 +315,7 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
 		if storageNode := n.StorageWorker.GetRuntime(rt.ID()); storageNode != nil {
 			status.Storage, err = storageNode.GetStatus(ctx)
 			if err != nil {
-				n.logger.Error("failed to fetch storage worker status",
-					"err", err,
-					"runtime_id", rt.ID(),
-				)
+				logger.Error("failed to fetch storage worker status", "err", err)
 			}
 		}
 
--- a/go/runtime/history/indexer.go
+++ b/go/runtime/history/indexer.go
@@ -16,17 +16,59 @@ import (
 )
 
 const (
+	statusStarted    = "started"
+	statusReindexing = "reindexing"
+	statusIndexing   = "indexing"
+	statusStopped    = "stopped"
+)
+
+// IndexerStatus is the status of runtime history indexer.
+type IndexerStatus struct {
+	// Status is the concise status of runtime history indexer state.
+	Status string `json:"status"`
+
+	// LastRound is the last runtime round that was indexed.
+	LastRound uint64 `json:"last_round"`
+
+	// ReindexStatus is history reindex status.
+	//
+	// It is nil unless during history reindex.
+	ReindexStatus *ReindexStatus `json:"reindex_status,omitempty"`
+}
+
+type ReindexStatus struct {
+	// BatchSize is the number of blocks to reindex in a single batch.
+	BatchSize uint16 `json:"batch_size"`
+	// LastHeight is the last consensus height that was indexed.
+	LastHeight int64 `json:"last_height"`
+	// StartHeight is the first height of history reindex interval.
+	StartHeight int64 `json:"start_height"`
+	// EndHeight is the last height of history reindex interval.
+	EndHeight int64 `json:"end_height"`
+	// ETA is expected time of history reindex completition.
+	ETA time.Time `json:"eta"`
+}
+
+const (
 	maxPendingBlocks = 10
 )
 
 // BlockIndexer is responsible for indexing and committing finalized
 // runtime blocks from the consensus into the runtime history.
 type BlockIndexer struct {
+	mu       sync.RWMutex
 	startOne cmSync.One
 
 	consensus consensus.Backend
 	history   History
 	batchSize uint16
+
+	status      string
+	lastHeight  int64
+	startHeight int64
+	endHeight   int64
+	lastRound   uint64
+	started     time.Time
 
 	logger *logging.Logger
 }
@@ -54,8 +96,40 @@ func (bi *BlockIndexer) Stop() {
 	bi.startOne.TryStop()
 }
 
+// Status returns runtime block history indexer status.
+func (bi *BlockIndexer) Status() *IndexerStatus {
+	bi.mu.RLock()
+	defer bi.mu.RUnlock()
+
+	status := &IndexerStatus{
+		Status:    bi.status,
+		LastRound: bi.lastRound,
+	}
+
+	if bi.status != statusReindexing {
+		return status
+	}
+
+	elapsed := time.Since(bi.started).Milliseconds()
+	remaining := elapsed * (bi.endHeight - bi.lastHeight) / max((bi.lastHeight-bi.startHeight+1), 1)
+	eta := time.Now().Add(time.Duration(remaining) * time.Millisecond)
+	status.ReindexStatus = &ReindexStatus{
+		BatchSize:   bi.batchSize,
+		LastHeight:  bi.lastHeight,
+		StartHeight: bi.startHeight,
+		EndHeight:   bi.endHeight,
+		ETA:         eta,
+	}
+
+	return status
+}
+
 func (bi *BlockIndexer) run(ctx context.Context) {
 	bi.logger.Info("starting")
+
+	bi.mu.Lock()
+	bi.status = statusStarted
+	bi.mu.Unlock()
 
 	// Subscribe to new runtime blocks.
 	blkCh, blkSub, err := bi.consensus.RootHash().WatchBlocks(ctx, bi.history.RuntimeID())
@@ -86,10 +160,18 @@ func (bi *BlockIndexer) run(ctx context.Context) {
 	// Index new blocks.
 	bi.index(ctx, blkCh)
 	bi.logger.Info("stopping")
+
+	bi.mu.Lock()
+	bi.status = statusStopped
+	bi.mu.Unlock()
 }
 
 func (bi *BlockIndexer) index(ctx context.Context, blkCh <-chan *roothash.AnnotatedBlock) {
 	bi.logger.Debug("indexing")
+
+	bi.mu.Lock()
+	bi.status = statusIndexing
+	bi.mu.Unlock()
 
 	retry := time.Duration(math.MaxInt64)
 	boff := cmnBackoff.NewExponentialBackOff()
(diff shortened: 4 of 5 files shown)


go/oasis-node/cmd/node/node_control.go
```diff
@@ -343,6 +343,17 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
 			}
 		}
 
+		// Fetch last consensus height seen by block history.
+		height, err := rt.History().LastConsensusHeight()
```
martintomazic, line 347: > it would be useful for oasis-node to report the current reindex round 

Notice I report consensus height instead of runtime round. I would argue this is also more appropriate, since reindex iterates over the block height...

go/control/api/api.go
```diff
@@ -189,6 +189,9 @@ type RuntimeStatus struct {
 	// Storage contains the storage worker status in case this node is a storage node.
 	Storage *storageWorker.Status `json:"storage,omitempty"`
 
+	// BlockHistoryLastConsensusHeight is last consensus height seen by block history.
+	BlockHistoryLastConsensusHeight int64 `json:"block_history_last_consensus_height"`
+
```
martintomazic, line 194: >so that the node operator could estimate when the reindex will finish.

I am not sure current node control status is helpful unless node operator is very familiar with oasis?

Hackish alternative may be to only show it if during "history reindex" status right below it, giving more context to the operator:

```bash
diff --git a/go/worker/common/api/api.go b/go/worker/common/api/api.go
index 41e0a46ff..0e050b421 100644
--- a/go/worker/common/api/api.go
+++ b/go/worker/common/api/api.go
@@ -107,6 +107,9 @@ type Status struct {
        // Status is a concise status of the committee node.
        Status StatusState `json:"status"`
 
+       // HistoryReindexHeight is last consensus height seen by block history.
+       HistoryReindexHeight *int64 `json:"history_reindex_height,omitempty"`
+
        // ActiveVersion is the currently active version.
        ActiveVersion *version.Version `json:"active_version"`
 
diff --git a/go/worker/common/committee/node.go b/go/worker/common/committee/node.go
index 9f06bcd15..668e43365 100644
--- a/go/worker/common/committee/node.go
+++ b/go/worker/common/committee/node.go
@@ -275,6 +275,20 @@ func (n *Node) GetStatus() (*api.Status, error) {
        var status api.Status
        status.Status = n.getStatusStateLocked()
 
+       // Include last consensus height if during history reindex.
+       if status.Status == api.StatusStateWaitingHistoryReindex {
+               height, err := n.Runtime.History().LastConsensusHeight()
+               if err != nil {
+                       n.logger.Warn("failed to retrieve last consensus height seen by block history",
+                               "err", err,
+                               "id", n.Runtime.ID(),
+                       )
+               } else {
+                       status.HistoryReindexHeight = &height
+
+               }
+       }

```

go/oasis-node/cmd/node/node_control.go
```diff
@@ -343,6 +343,17 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
...
+		if err == nil {
```
peternose, line 348:
```suggestion
		if err != nil {
```

go/control/api/api.go
```diff
@@ -189,6 +189,9 @@ type RuntimeStatus struct {
 	// Storage contains the storage worker status in case this node is a storage node.
 	Storage *storageWorker.Status `json:"storage,omitempty"`
 
+	// BlockHistoryLastConsensusHeight is last consensus height seen by block history.
+	BlockHistoryLastConsensusHeight int64 `json:"block_history_last_consensus_height"`
```
peternose, line 193: Quite long field name. I would try to find a shorter name or move it under history.
Should this field be grouped with `LastRetainedRound`?

.changelog/5998.feature.md
```diff
@@ -0,0 +1 @@
+Show the progress of the history reindex in oasis control status
```
peternose, line 1: We usually start [changelog](https://github.com/oasisprotocol/oasis-core/blob/v24.3.1/CHANGELOG.md) and git messages with a prefix, in this case `go/oasis_node/cmd` or `go/oasis_node` would be fine. 
After the title we could explain which field was added and what it shows. See previous logs for examples.

go/oasis-node/cmd/node/node_control.go
peternose: Can we show the last height and the last round?
Would percent of indexed blocks help the node operators?

go/oasis-node/cmd/node/node_control.go
```diff
@@ -343,6 +343,17 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
...
+			n.logger.Error("failed to retrieve last consensus height seen by block history",
+				"err", err,
+				"id", rt.ID(),
+			)
+		} else {
```
peternose, line 353: [nit] I'm not a big fan of `if/else`, as it makes code less readable and the reader normally expects only `if` part which handles the error. To be in-line with the code above, you could assign directly to the status field (even if there is an error, assuming that the returned value is zero value) or use `switch` like it is done above. I would use the former, just to be in-line with code above.

Maybe this code could be moved up, where we handle history.

go/oasis-node/cmd/node/node_control.go
kostko: I agree, all of these would be useful.

go/control/api/api.go
kostko: > I am not sure current node control status is helpful unless node operator is very familiar with oasis?

The CLI can provide a nicer computed overlay from this information. But it would be good if the node could estimate the average reindex speed and then an ETA as the range of blocks to reindex is known.

go/control/api/api.go
kostko: Yes, we should shorten it.

martintomazic: Now that I think the cleanest way would probably be to show last round and height as seen by the block history right below `storage` field. I am thinking about exposing this status from the `roothash/api/BlockHistory` interface ([see](https://github.com/oasisprotocol/oasis-core/blob/master/go/roothash/api/history.go)):

```bash
      "latest_round": 10160924,
      "latest_hash": "85867c19bd6ad43521432e187962e3bb19a9731c79cd64f195821bfd90bd7b24",
      "latest_time": "2025-01-31T11:08:05+01:00",
      "latest_state_root": {
        "ns": "000000000000000000000000000000000000000000000000a6d1e3ebf60dff6c",
        "version": 10160924,
        "root_type": 1,
        "hash": "a914f8c8baf05a47757c1d7f7caecebdf614c8b3d1e20aa33d6855ec96a50ba1"
      },
      "genesis_round": 2995927,
      "genesis_hash": "c9f3ca654531b775d944c85e1f00e76944aaf5de1902acfa082ff76e852dba5e",
      "last_retained_round": 10145927,
      "last_retained_hash": "fad686c7eb077ec1a1a5a1ba1c27027a3788b003c94130e78c7dfb25a1d0e8cd",
      // other fields
      "storage": {
        "status": "syncing rounds",
        "last_finalized_round": 10160923
      },
      "history": {
        "last_round": 10160923,
        "last_height": 25058099
      },
```
> Should this field be grouped with LastRetainedRound?
1. `LastRetainedRound` is actually earliest (not latest) runtime block that is also synced to the storage.
    i. You may still get an error and show the latest retained round from the block history even if you don't have it synced to the storage... [See](https://github.com/oasisprotocol/oasis-core/blob/6fc521b26ad32243a5c4722f4d91b1ff926b23e2/go/oasis-node/cmd/node/node_control.go#L291). Is this a bug?
2. History reindex is actually happening and finishes before storage round syncing so ideally this should be separate grouping.

I am thinking could we move `last_retained_round` and `last_retained_hash` to `storage`? This is breaking obviously. More importantly the `storage` status is received from the [storage committee worker](https://github.com/oasisprotocol/oasis-core/blob/5773ab9f72fbf125788088d6c51aceaeedd366d4/go/worker/storage/committee/node.go#L328), whilst `LastRetainedRound` and hash are technically overwritten, directly from the runtime storage ([see](https://github.com/oasisprotocol/oasis-core/blob/6fc521b26ad32243a5c4722f4d91b1ff926b23e2/go/oasis-node/cmd/node/node_control.go#L284), described in **1.i**). As from the [docstring](https://github.com/oasisprotocol/oasis-core/blob/5773ab9f72fbf125788088d6c51aceaeedd366d4/go/worker/storage/api/api.go#L67) the former is synced + finalized, whilst if reading from the storage directly it is only guaranteed to be synced? Not sure what is the difference (if any)?

go/control/api/api.go
```diff
@@ -230,6 +233,14 @@ type SeedStatus struct {
...
 }
 
+// HistoryStatus is the status as observed by the block history.
+type HistoryStatus struct {
+	// LastRound is the round of the latest block.
+	LastRound uint64 `json:"last_round"`
+	// LastHeight is the consensus height of the latest block.
+	LastHeight int64 `json:"last_height"`
+}
+
```
martintomazic, line 243: Do we also want a status like we have for `storage`? e.g. initializing, reindexing blocks, syncing blocks?

If so this logic would have to be moved to [consensus roothash](https://github.com/oasisprotocol/oasis-core/blob/5773ab9f72fbf125788088d6c51aceaeedd366d4/go/consensus/cometbft/roothash/roothash.go#L51), or we  add a method to query it.  Albeit, feels weird adding this functionality to `roothash/api/Backend`...?

go/oasis-node/cmd/node/node_control.go
martintomazic: >Can we show the last height and the last round?

I have added those. Have a bit of a mixed feelings about percentage...

1. Do we want to add it to consensus sync and storage syncing as well (to be consistent)?
2. Should we only show it if during reindex/syncing?
3. The value is not deterministic so we cannot expose it as status, `[reindex/syncing]_progress: 90.1%` looks good?

go/control/api/api.go
martintomazic: > But it would be good if the node could estimate the average reindex speed and then an ETA as the range of blocks to reindex is known.

I would prefer to tackle ETA in another PR here is why:
1. consensus reindex >= storage sync >> history reindex (time needed for my testnet node to sync from start)
2. ETA also depends if you are doing reindex during consensus sync or if you start with already synced node. Need to do some more experiments just to be sure...

go/oasis-node/cmd/node/node_control.go
peternose: > 1. Do we want to add it to consensus sync and storage syncing as well (to be consistent)?

I said it would be nice to have percentages, so that you know how far are you.

> 2. Should we only show it if during reindex/syncing?

Probably no. If we could show 100%, that would mean that the syncing is not in progress.

> 3. The value is not deterministic so we cannot expose it as status

Not sure what you are trying to say here.

go/control/api/api.go
peternose: > History reindex is actually happening and finishes before storage round syncing so ideally this should be separate grouping.

Maybe we show status of both. One part should probably be under `storage`, other under `history`. Although I'm not sure if users will understand the difference. 

> Should this field be grouped with `LastRetainedRound`?

The code which fetches these fields should also be together.

> This is breaking

Breaking for who? I think it is not a problem if we change the status. If node operators depend on it, they can check the changelog and modify their code.

go/control/api/api.go
peternose: > Do we also want a status like we have for `storage`? e.g. initializing, reindexing blocks, syncing blocks?

Yes, I would like to have percentages (and maybe optional status). 

From you example, I would like to have at least 3 fields, two that say which range is `reindexBlocks` working on (800 - 1000), and percentages of reindexed blocks (30/200 = 15%). When this is done, the node will see that too many blocks are still to be indexed, so it will update status by updating range (1000 - 1100) and percentage (0/100 = 0%).

Having field `LastRound` might be a problem when you start doing things in parallel, as some lower rounds might not be processed yet. Range might be better.

martintomazic:
With #6089 merged, I have extended the functionality of this PR to show indexer status. [See](https://github.com/oasisprotocol/oasis-core/pull/6013#issue-2814281446) updated context/description.

Should be ready for a second round of reviews.

go/runtime/history/indexer.go
```diff
@@ -15,6 +15,37 @@ import (
...
+	Status IndexerStatus `json:"status"`
+	// LastIndexedHeight is the last consensus height that was indexed.
+	LastIndexedHeight int64 `json:"last_indexed_height"`
+	// LastIndexedRound is the last runtime round that was indexed.
+	LastIndexedRound uint64 `json:"last_indexed_round"`
+	// StartHeight is the first height of history reindex interval.
+	// It is nil unless reindexing.
+	StartHeight *int64 `json:"start_height,omitempty"`
+	// EndHeight is the last height of history reindex interval.
+	// It is nil unless reindexing.
```
peternose, line 41: I usually add a new line before this because this is not that important information.

Should we always display this so that user knows how long the reindex took? In this case, we would have fields `LastHeight`, `LastRound`, and for reindex `StartHeight`, `EndHeight`, `StartTime`, `EndTime` and `ETA`  which would be zero when finished.

go/runtime/history/indexer.go
```diff
@@ -15,6 +15,37 @@ import (
...
+	StatusIndexing     IndexerStatus = "indexing"
+	StatusReindexing   IndexerStatus = "reindexing"
+)
+
+// BlockIndexerStatus is the status of runtime history indexer.
+type BlockIndexerStatus struct {
+	// Status is the runtime history indexer status.
+	Status IndexerStatus `json:"status"`
+	// LastIndexedHeight is the last consensus height that was indexed.
+	LastIndexedHeight int64 `json:"last_indexed_height"`
```
peternose, line 34:
```suggestion
	LastHeight int64 `json:"last_height"`
```

go/runtime/history/indexer.go
```diff
@@ -45,17 +84,59 @@ func NewBlockIndexer(consensus consensus.Backend, history History) *BlockIndexer
...
+
+	// Set reindex range with ETA iff reindexing.
+	if bi.status != StatusReindexing {
+		return status
+	}
+
+	status.StartHeight = &bi.startHeight
+	status.EndHeight = &bi.endHeight
+
+	// ETA is misleading if running only for a short period.
```
peternose, line 119: I don't care about that, it only complicates things.

```go
	elapsed := time.Since(bi.started).Milliseconds()
	remaining := elapsed * (bi.endHeight - bi.lastHeight) / (bi.lastHeight - bi.startHeight + 1)
	eta := time.Now().Add(time.Duration(remaining) * time.Millisecond)

	status.StartHeight = &bi.startHeight
	status.EndHeight = &bi.endHeight
	status.ETA = &eta
```

go/runtime/history/indexer.go
```diff
@@ -45,17 +84,59 @@ func NewBlockIndexer(consensus consensus.Backend, history History) *BlockIndexer
 
 // Start starts the indexer.
 func (bi *BlockIndexer) Start() {
+	bi.mu.Lock()
+	bi.status = StatusStarting
```
peternose, line 88: I would move this to `run`. But since we already have one status there, I would remove `StatusInitializing`. Observe that this can be called multiple times, which will override the status.

go/runtime/history/indexer.go
```diff
@@ -15,6 +15,37 @@ import (
 	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
 )
 
+// IndexerStatus is the concise status of runtime history indexer.
+type IndexerStatus string
```
peternose, line 19: I hate adding new types unless they have at least one method. 
I would choose another name, e.g. `BlockIndexerState`. You can also rename `BlockIndexer` to `Indexer`, to shorten things.

go/runtime/history/indexer.go
```diff
@@ -45,17 +84,59 @@ func NewBlockIndexer(consensus consensus.Backend, history History) *BlockIndexer
...
+	bi.mu.Lock()
+	bi.status = StatusStarting
+	bi.mu.Unlock()
 	bi.startOne.TryStart(bi.run)
 }
 
 // Stop halts the indexer.
 func (bi *BlockIndexer) Stop() {
+	bi.mu.Lock()
+	bi.status = StatusStopping
```
peternose, line 96: Move to end of `run`.

go/runtime/history/indexer.go
```diff
@@ -15,6 +15,37 @@ import (
 	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
 )
 
+// IndexerStatus is the concise status of runtime history indexer.
+type IndexerStatus string
+
+const (
+	StatusInitializing IndexerStatus = "initializing"
+	StatusStarting     IndexerStatus = "starting"
```
peternose, line 23:
```suggestion
	StatusStarted     IndexerStatus = "started"
```

go/runtime/history/indexer.go
```diff
@@ -15,6 +15,37 @@ import (
...
+	StatusStopping     IndexerStatus = "stopping"
```
peternose, line 24:
```suggestion
	StatusStopped     IndexerStatus = "stopped"
```

go/runtime/history/indexer.go
```diff
@@ -15,6 +15,37 @@ import (
 	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
 )
 
+// IndexerStatus is the concise status of runtime history indexer.
+type IndexerStatus string
+
+const (
```
peternose, line 21:
```suggestion
const (
	StatusStarted     = "started"
	StatusReindexing   = "reindexing"
	StatusIndexing     = "indexing"
	StatusStopped     = "stopped"
)
```
So that you know what is the right order.

go/runtime/history/indexer.go
```diff
@@ -289,6 +383,8 @@ func (bi *BlockIndexer) reindexRange(ctx context.Context, start int64, end int64
 		return err
 	}
 
+	bi.updateLastIndexedHeight(blocks)
```
peternose, line 386: I would move this to `commitBlocks`, and then maybe just pass the last block to remove `if` check, or even better, remove this function and do that in the `commitBlocks`.

go/runtime/history/indexer.go
```diff
@@ -230,6 +317,13 @@ func (bi *BlockIndexer) reindexTo(ctx context.Context, height int64) error {
 		lastHeight = lastRetainedHeight
 	}
 
+	bi.mu.Lock()
+	bi.status = StatusReindexing
+	bi.startHeight = lastHeight
+	bi.endHeight = height
+	bi.started = time.Now()
```
peternose, line 324: Should we change this by extending the end height only? This way, we can estimate next reindex more accurately, plus in the end the user can see the whole range.

go/runtime/history/indexer.go
```diff
@@ -15,6 +15,37 @@ import (
...
+	// StartHeight is the first height of history reindex interval.
+	// It is nil unless reindexing.
+	StartHeight *int64 `json:"start_height,omitempty"`
+	// EndHeight is the last height of history reindex interval.
+	// It is nil unless reindexing.
+	EndHeight *int64 `json:"end_height,omitempty"`
+	// ETA is expected time of history reindex completition.
+	// It is nil unless reindexing and enough progress has been made
+	// to make a meaningful prediction.
+	ETA *time.Time `json:"eta,omitempty"`
```
peternose, line 46: I would expect this to be duration, so how long I still have to wait.

martintomazic: I have changed to duration but I think I would prefer timestamp?

```bash 
        "eta": "2025-03-17T16:38:07+01:00"
```

vs

```bash  
          "eta": 1153740000000
```

I could print it in seconds, but then api type `time.Duration` is misleading since I am not printing nanoseconds. On the other hand timestamp is both human and machine readable.

go/runtime/history/indexer.go
martintomazic: >Should we always display this so that user knows how long the reindex took?

I would keep it simple and avoid unneeded information in the status output. The user is only interested in the status. Finally if it is reindexing it is interested in ETA, others he doesn't care?

How long the reindex took is only interesting to developer, and could be found in the logs...

Happy to change! :)

go/runtime/history/indexer.go
martintomazic: Added `max(..., 1) to avoid integer division.

go/runtime/history/indexer.go
martintomazic: I have kept it as `Status string` for now, happy to change to `IndexerState string`?

Overall we have a bit of inconsistencies in our codebase as I took existing status for the reference.

>I hate adding new types unless they have at least one method.

Generally agree, albeit I like things to be typed when doing if in a given state comparisons...

go/runtime/history/indexer.go
peternose: `LastHeight` should be removed or only showed during reindex, because if the runtime doesn't produce any blocks, the height will stay still and the user might think that indexer is not working. 

Maybe we change to:
- `LastRound` (when indexing)
- `StartHeight`, `LastHeight`, `EndHeight`, `ETA` (when reindexing).

What do you think?

go/runtime/history/indexer.go
peternose: > I have kept it as `Status string` for now, happy to change to `IndexerState string`?

If you named it `history.Status`, then this is status of the history, not of the block indexer.

> Generally agree, albeit I like things to be typed when doing if in a given state comparisons...

Ok, you can keep it as it is. But the trick is that:
- you cannot create `enum` with a new type,
- you never get full safety out of the type system (for example, anyone can change `status` to `status = "some status"` because of implicit conversion)
- you are hiding the underlying type, which means that users have no idea weather to pass `&status` or `status` and how much passing by value costs,
- the costs for casting to and from the underlying type everywhere can be substantial (been there done that, and is horrible).

For me, the only reason for using this is to be consistent with the rest of the code.

go/runtime/history/indexer.go
martintomazic: > the height will stay still and the user might think that indexer is not working.

so will `LastRound`? Not sure this is big deal as other parts will also be halted: common committee status, storage status...?

peternose: Yes, but if the user checks the last round in the consensus, he will see that new runtime blocks are not produced, so it is obvious that this number doesn't change. But if he checks the last height in the consensus, he will see that new consensus blocks arrive, but the indexer is still stuck on the last height. So the indexer is not working?

Since reindex processes heights and index rounds (blocks), maybe we should display as I wrote.

go/runtime/history/indexer.go
martintomazic: I have left it as you suggested initially. 

Closing...

go/runtime/history/indexer.go
```diff
@@ -15,6 +15,35 @@ import (
 	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
 )
 
+const (
+	StatusStarted    = "started"
```
peternose, line 19: I would make this private, as otherwise someone can use `history.StatusStarted` and has no idea to which struct this status belongs (to history?). And public constants should also have comments, if we keep as it is.

go/control/api/api.go
```diff
@@ -184,13 +185,14 @@ type RuntimeStatus struct {
...
-	// Committee contains the runtime worker status in case this node is a (candidate) member of a
-	// runtime committee.
+	// Committee contains the runtime common committee worker status.
 	Committee *commonWorker.Status `json:"committee"`
 	// Executor contains the executor worker status in case this node is an executor node.
 	Executor *executorWorker.Status `json:"executor,omitempty"`
 	// Storage contains the storage worker status in case this node is a storage node.
 	Storage *storageWorker.Status `json:"storage,omitempty"`
+	// History contains the runtime history indexer status.
+	History *history.BlockIndexerStatus `json:"history,omitempty"`
```
peternose, line 195:
```suggestion
	History *history.Status `json:"history,omitempty"`
```

If you want to make this more general, and then define:

```go
// Status is the status of runtime history.
type Status struct {
	// LastHeight is the consensus height of the last committed block.
	LastHeight int64 `json:"last_height"`
	// LastRound is the round of the last committed block.
	LastRound uint64 `json:"last_round"`
	// Reindex is the status of runtime history reindexer.
	Reindex *ReindexStatus `json:"reindex,omitempty"`
}

// ReindexStatus is the status of runtime history reindex.
type ReindexStatus struct {
	// BatchSize is the number of blocks to reindex in a single batch.
	BatchSize int `json:"batch_size,omitempty"`
	// StartHeight is the first height of history reindex interval.
	StartHeight int64 `json:"start_height,omitempty"`
	// EndHeight is the last height of history reindex interval.
	EndHeight int64 `json:"end_height,omitempty"`
	// ETA is expected time of history reindex completition.
	ETA time.Time `json:"eta,omitempty"`
}
```

where `LastHeight` and `LastRound` should be read from the history and not from the indexer.

This way you can get rid of `nil` values and `status` field, shorten comments, and read last block height/round from the right place.

martintomazic: I am not 100% convinced albeit probably looks nicer...

1. I think status is useful to display (also for debugging):
    *  We have issue where worker will silently stop -> you will see `stopped` status.
    * Finally you need to track status since `*ReindexStatus` should only be shown if reindexing else confusing.
3. `LastHeight` and `LastRound` don't have to be read from history since they can simply output the view of the reindexer worker. Just like is done for the storage worker. Block history could probably have a separate status?
4. >LastHeight should be removed or only showed during reindex, because if the runtime doesn't produce any blocks, the height will stay still and the user might think that indexer is not working. 
    * Now you have this problem again, as you cannot expect the user to read doc api (why would node operator search repo...)
    
That said I really like nil pointers have been removed. Personally I would prefer to leave it as is until we finalize abstractions (registry, workers, history, who contains who....).

peternose (approved):

go/control/api/api.go
```diff
@@ -184,13 +185,14 @@ type RuntimeStatus struct {
...
 
-	// Committee contains the runtime worker status in case this node is a (candidate) member of a
-	// runtime committee.
+	// Committee contains the runtime common committee worker status.
 	Committee *commonWorker.Status `json:"committee"`
 	// Executor contains the executor worker status in case this node is an executor node.
 	Executor *executorWorker.Status `json:"executor,omitempty"`
 	// Storage contains the storage worker status in case this node is a storage node.
 	Storage *storageWorker.Status `json:"storage,omitempty"`
+	// Indexer contains the runtime history indexer status in case this is not keymanager runtime.
```
peternose, line 194:
```suggestion
	// Indexer contains the runtime history indexer status in case this runtime has a block indexer.
```

Issue #5738, opened 2024-07-03
Optimize runtime history reindex


kostko:
Possibly doing the reindex in parallel. Should do some experiments.

peternose:
@kostko Did you mean to reindex every runtime in parallel, or to reindex blocks or blocks of blocks in parallel?

kostko:
Both if that would improve performance. We can split it into two issues though.

ptrus:
:+1: for both. Some more context: Reindexing a single runtime already takes a lot of time, so we want/need to parallelize reindexing blocks in that scenario. This will likely have the biggest impact on speeding up the full node initialization. It probably also makes sense to parallelize it across runtimes as well.

One additional related thing that I remember from the past is that we currently block starting any runtimes until all have completed the history reindex. This could likely be improved as well. But probably belongs in a separate issue.

martintomazic:
For the reference:

1. #6050 solved write bottleneck of history reindex (3x increase + enables read optimiziations).
2. #6095 stopped reindexing redundant  round results (further 30% increase).
3. #6089 moved history reindex from roothash service to runtime registry.
    * Runtimes are now automatically reindexed in parallel, so the case where reindex of one runtime blocks already reindexed runtimes is gone.


This gave us following performance gain:
1. `4x` speed-up for single runtime.
2. `4 x n` speed-up for `n` runtimes. (Likely if `n` grows this factor decreases).
 
E.g. having `cipher`, `emerald` and `sapphire` configured resulted in `12-14x` speed-up (benched on testnet starting with synced consensus and empty block history dir). We are happy with that for now, so closing this issue.

TODO:
We could easily get another 2-3x, by optimizing the reading part. This was already [benched](https://docs.google.com/document/d/1Lq-Qe6jrrLUSKIkLLsIGOrfb31Ivi-hZKP0wAC50V1E/edit?tab=t.0#heading=h.97gqwtgjc6xf), where we fetched reads in parallel within a batch (using worker pool - (#6055)).

PR #6031, opened 2025-02-03
go/roothash: Parallelize history reindexes

martintomazic:
Part of #5738 

# What
1. Run multiple history reindexes in parallel:
     * Bottleneck will always be iteration over all consensus blocks...
     * However in case of `n` runtimes we already gain almost `n` speed-up, since reindex was sequentiall for each runtime until now.
2. >One additional related thing that I remember from the past is that we currently block starting any runtimes until all have completed the history reindex. This could likely be improved as well. But probably belongs in a separate issue.
    * This was the case if you added new runtime, so you needed to reindex it. ~If the first received reindex command was for this new runtime, you were blocking ready status of all other runtimes, that might be only few heights behind... This is also fixed here~. This was the case the moment you started reindexing new runtime, since you were also blocking receiving events for other runtimes, even if their history has been reindexed prior to the new runtime history reindex started -> will fix the changelog.

# How to test/benchmark

I am running local testnet with `sapphire`, `emerald` and `cipher` set. I start with synced consensus. I deleted the `data/runtimes` datadir, to trigger history reindex from the start. Before this change it took around 6h (2h per runtime) for it to finish. Now it happens in ~2h20min (in parallel).

## TODO

Benchmark history reindex whilst doing consensus sync.

### DIFF
--- a/go/consensus/cometbft/roothash/roothash.go
+++ b/go/consensus/cometbft/roothash/roothash.go
@@ -437,13 +437,19 @@ func (sc *serviceClient) reindexBlocks(currentHeight int64, bh api.BlockHistory)
 	)
 
 	for height := lastHeight; height <= currentHeight; height++ {
-		// XXX: could soft-fail first few heights in case more heights were
-		// pruned right after the GetLastRetainedVersion query.
-		round, err := sc.reindexBlock(runtimeID, height)
-		if err != nil {
-			return 0, err
+		select {
+		case <-sc.ctx.Done():
+			logger.Info("context cancelled")
+			return lastRound, nil
+		default:
+			// XXX: could soft-fail first few heights in case more heights were
+			// pruned right after the GetLastRetainedVersion query.
+			round, err := sc.reindexBlock(runtimeID, height)
+			if err != nil {
+				return 0, err
+			}
+			lastRound = round
 		}
-		lastRound = round
 	}
 
 	if lastRound == api.RoundInvalid {
@@ -545,7 +551,7 @@ func (sc *serviceClient) ServiceDescriptor() tmapi.ServiceDescriptor {
 func (sc *serviceClient) DeliverCommand(ctx context.Context, height int64, cmd interface{}) error {
 	switch c := cmd.(type) {
 	case *cmdTrackRuntime:
-		sc.handleTrackRuntime(ctx, height, c.runtimeID, c.blockHistory)
+		go sc.handleTrackRuntime(ctx, height, c.runtimeID, c.blockHistory)
 	default:
 		return fmt.Errorf("roothash: unknown command: %T", cmd)
 	}
@@ -574,7 +580,12 @@ func (sc *serviceClient) handleTrackRuntime(ctx context.Context, height int64, r
 	}
 	sc.trackedRuntime[runtimeID] = tr
 	// Request subscription to events for this runtime.
-	sc.queryCh <- app.QueryForRuntime(tr.runtimeID)
+	// This has to be done only after history reindex is done, so that we don't
+	// receive events we are currently reindexing.
+	// Finally, this has to be done always, so that tracking request is not lost.
+	defer func() {
+		sc.queryCh <- app.QueryForRuntime(tr.runtimeID)
+	}()
 
 	// Resolve the correct block finalization height to use for the latest block at the current
 	// height as the current height may not correspond to the latest block finalization height.
--- a/go/consensus/cometbft/roothash/roothash.go
+++ b/go/consensus/cometbft/roothash/roothash.go
@@ -545,58 +545,62 @@ func (sc *serviceClient) ServiceDescriptor() tmapi.ServiceDescriptor {
 func (sc *serviceClient) DeliverCommand(ctx context.Context, height int64, cmd interface{}) error {
 	switch c := cmd.(type) {
 	case *cmdTrackRuntime:
-		// Request to track a new runtime.
-		etr := sc.trackedRuntime[c.runtimeID]
-		if etr != nil {
-			// Ignore duplicate runtime tracking requests unless this updates the block history.
-			if etr.blockHistory != nil || c.blockHistory == nil {
-				break
-			}
-		} else {
-			sc.logger.Debug("tracking new runtime",
-				"runtime_id", c.runtimeID,
-				"height", height,
-			)
-		}
-
-		// We need to start watching a new block history.
-		tr := &trackedRuntime{
-			runtimeID:    c.runtimeID,
-			blockHistory: c.blockHistory,
-		}
-		sc.trackedRuntime[c.runtimeID] = tr
-		// Request subscription to events for this runtime.
-		sc.queryCh <- app.QueryForRuntime(tr.runtimeID)
-
-		// Resolve the correct block finalization height to use for the latest block at the current
-		// height as the current height may not correspond to the latest block finalization height.
-		rs, err := sc.GetRuntimeState(ctx, &api.RuntimeRequest{
-			RuntimeID: tr.runtimeID,
-			Height:    height,
-		})
-		if err != nil {
-			sc.logger.Warn("failed to get runtime state for latest block",
-				"err",
-				"runtime_id", tr.runtimeID,
-				"height", height,
-			)
-			return nil
-		}
-
-		// Emit latest block.
-		if err := sc.processFinalizedEvent(ctx, rs.LastBlockHeight, tr.runtimeID, nil, false); err != nil {
-			sc.logger.Warn("failed to emit latest block",
-				"err", err,
-				"runtime_id", tr.runtimeID,
-				"height", height,
-			)
-		}
-		// Make sure we reindex again when receiving the first event.
-		tr.reindexDone = false
+		sc.handleTrackRuntime(ctx, height, c.runtimeID, c.blockHistory)
 	default:
 		return fmt.Errorf("roothash: unknown command: %T", cmd)
 	}
 	return nil
+}
+
+func (sc *serviceClient) handleTrackRuntime(ctx context.Context, height int64, runtimeID common.Namespace, history api.BlockHistory) {
+	// Request to track a new runtime.
+	etr := sc.trackedRuntime[runtimeID]
+	if etr != nil {
+		// Ignore duplicate runtime tracking requests unless this updates the block history.
+		if etr.blockHistory != nil || history == nil {
+			return
+		}
+	} else {
+		sc.logger.Debug("tracking new runtime",
+			"runtime_id", runtimeID,
+			"height", height,
+		)
+	}
+
+	// We need to start watching a new block history.
+	tr := &trackedRuntime{
+		runtimeID:    runtimeID,
+		blockHistory: history,
+	}
+	sc.trackedRuntime[runtimeID] = tr
+	// Request subscription to events for this runtime.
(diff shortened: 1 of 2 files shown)


go/consensus/cometbft/roothash/roothash.go
```diff
@@ -525,6 +470,78 @@ func (sc *serviceClient) reindexBlocks(currentHeight int64, bh api.BlockHistory)
...
+		// Only process events for the given runtime.
+		if !evRtID.Equal(&runtimeID) {
+			continue
+		}
+		if err = sc.processFinalizedEvent(sc.ctx, height, *evRtID, &ev.Round, true); err != nil {
+			return 0, fmt.Errorf("failed to process finalized event: %w", err)
+		}
+		lastRound = ev.Round
+	}
+	return lastRound, nil
```
martintomazic, line 542: Added this just bo safe/100% consistent with previous logic. Can we have multiple finalized events for one runtime at one height? We may simplify if this is not the case, but also don't want to make assumptions here if we change later on...

go/consensus/cometbft/roothash/roothash.go
```diff
@@ -437,74 +437,13 @@ func (sc *serviceClient) reindexBlocks(currentHeight int64, bh api.BlockHistory)
...
-			if !evRtID.Equal(&runtimeID) {
-				continue
-			}
-			if err = sc.processFinalizedEvent(sc.ctx, height, *evRtID, &ev.Round, true); err != nil {
-				return 0, fmt.Errorf("failed to process finalized event: %w", err)
-			}
-			lastRound = ev.Round
+			return 0, err
 		}
+		lastRound = round
```
peternose: This is not the same as before, because if there are no events, this is set to 0.

go/consensus/cometbft/roothash/roothash.go
```diff
@@ -545,60 +545,64 @@ func (sc *serviceClient) ServiceDescriptor() tmapi.ServiceDescriptor {
...
 	default:
 		return fmt.Errorf("roothash: unknown command: %T", cmd)
 	}
 	return nil
 }
 
+func (sc *serviceClient) handleTrackRuntime(ctx context.Context, height int64, runtimeID common.Namespace, history api.BlockHistory) {
+	// Request to track a new runtime.
+	etr := sc.trackedRuntime[runtimeID]
+	if etr != nil {
```
peternose, line 564:
```suggestion
	if tr, ok := sc.trackedRuntime[runtimeID]; ok {
```

go/consensus/cometbft/roothash/roothash.go
```diff
@@ -437,13 +437,19 @@ func (sc *serviceClient) reindexBlocks(currentHeight int64, bh api.BlockHistory)
...
 	for height := lastHeight; height <= currentHeight; height++ {
-		// XXX: could soft-fail first few heights in case more heights were
-		// pruned right after the GetLastRetainedVersion query.
-		round, err := sc.reindexBlock(runtimeID, height)
-		if err != nil {
-			return 0, err
+		select {
+		case <-sc.ctx.Done():
+			logger.Info("context cancelled")
+			return lastRound, nil
```
peternose, line 443: Why don't we return an error here?

go/consensus/cometbft/roothash/roothash.go
```diff
@@ -437,13 +437,19 @@ func (sc *serviceClient) reindexBlocks(currentHeight int64, bh api.BlockHistory)
...
+		default:
+			// XXX: could soft-fail first few heights in case more heights were
```
peternose, line 445: Happy path should not be indented.

go/consensus/cometbft/roothash/roothash.go
```diff
@@ -437,13 +437,19 @@ func (sc *serviceClient) reindexBlocks(currentHeight int64, bh api.BlockHistory)
 	)
 
 	for height := lastHeight; height <= currentHeight; height++ {
-		// XXX: could soft-fail first few heights in case more heights were
-		// pruned right after the GetLastRetainedVersion query.
-		round, err := sc.reindexBlock(runtimeID, height)
-		if err != nil {
-			return 0, err
+		select {
+		case <-sc.ctx.Done():
```
peternose, line 441: This doesn't make any sense. The long-running functions in `reindexBlock` already use `sc.ctx` and will abort if canceled.

go/consensus/cometbft/roothash/roothash.go
```diff
@@ -545,7 +551,7 @@ func (sc *serviceClient) ServiceDescriptor() tmapi.ServiceDescriptor {
 func (sc *serviceClient) DeliverCommand(ctx context.Context, height int64, cmd interface{}) error {
 	switch c := cmd.(type) {
 	case *cmdTrackRuntime:
-		sc.handleTrackRuntime(ctx, height, c.runtimeID, c.blockHistory)
+		go sc.handleTrackRuntime(ctx, height, c.runtimeID, c.blockHistory)
```
peternose, line 554: Already said once, that if you start a go routine, you need to have control over it. Spawning routines all over the place is not the best practice.

Once you allow calling this method concurrently, synchronization becomes a problem. Your solution:
- won't work if we send two commands requesting to track the same runtime,
- will sometimes panic when accessing `sc.trackedRuntime`,
- ...

go/consensus/cometbft/roothash/roothash.go
```diff
@@ -574,7 +580,12 @@ func (sc *serviceClient) handleTrackRuntime(ctx context.Context, height int64, r
 	}
 	sc.trackedRuntime[runtimeID] = tr
 	// Request subscription to events for this runtime.
-	sc.queryCh <- app.QueryForRuntime(tr.runtimeID)
+	// This has to be done only after history reindex is done, so that we don't
```
peternose, line 583: This comment is not clear. This code needs to execute after reindex is done, but still executes if `GetRuntimeState` fails :thinking:

peternose (commented):
Is this code the final solution for this PR or intermediate hack to show how fast this works?

I would try to parallelize blocks and runtimes together, because I'm not sure if this solution can be upgraded to also parallelize blocks.

go/consensus/cometbft/roothash/roothash.go
peternose: This is not consistent, as already pointed out.

go/consensus/cometbft/roothash/roothash.go
```diff
@@ -533,59 +558,76 @@ func (sc *serviceClient) ServiceDescriptor() tmapi.ServiceDescriptor {
...
+		blockHistory: history,
+	}
+	sc.trackedRuntime[runtimeID] = tr
+	// Request subscription to events for this runtime.
+	// This has to be done only after history reindex is done, so that we don't
+	// receive events we are currently reindexing.
+	// Finally, this has to be done always, so that tracking request is not lost.
+	defer func() {
 		sc.queryCh <- app.QueryForRuntime(tr.runtimeID)
+	}()
```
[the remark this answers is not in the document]
peternose: > This has to be done only after history reindex is done, so that we don't receive events we are currently reindexing.

Not sure if this is fine. If the go scheduler pauses before the query is registered, we could miss few events. If we miss blocks also, this is not fine.

> A good solution was proposed to recursively call reindex until predefined number of heights away from the consensus height?

Doesn't seem like a question.

> If we fail to reindex due to errors for the second time, we are effectively skipping the reindex?

Maybe, didn't check the code.

go/consensus/cometbft/roothash/roothash.go
martintomazic: Sorry completely forgot that I should protect the vars now :/ Will fix :)

go/consensus/cometbft/roothash/roothash.go
martintomazic: You are right since we if  `reindexBlock` returns error we propagate it and exit. For some reason I had in mind we are only logging here (not returning) so I wanted to stop iteration here. Will fix.

go/consensus/cometbft/roothash/roothash.go
peternose: Is this solution even acceptable as now runtime tracking is done aysnc?

martintomazic: This was the case before as calling `TrackRuntime` was not blocking...?

go/consensus/cometbft/roothash/roothash.go
martintomazic: You are right, I think is best to factor this function back, especially since I don't have to do select over context... Thanks for spotting!

go/consensus/cometbft/roothash/roothash.go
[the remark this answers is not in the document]
martintomazic: 2. Is done [here](https://github.com/oasisprotocol/oasis-core/blob/5773ab9f72fbf125788088d6c51aceaeedd366d4/go/consensus/cometbft/roothash/roothash.go#L639) in `ProcessFinalizedEvent` defer statement, modifying tracked runtime pointer.

As you noticed I forget to lock `trackedRuntimes` variable. Handling this is actually quite cumbersome since we have `proccessFinalizedEvent` (requires locking), triggering `reindexBlocks`, which calls back `proccessFinalizedEvent` (requires locking) for every height. Finally we return to original `proccessFinalizedEvent` that triggered history reindex,  which in the defer statement modifies pointer to tracked runtime it started with (its bool flag) (requires locking) :/. Wouldn't be surprised if we could deadlock actually.

Two lessons here:
1. It is almost certain we need to refactor existing history reindex first.
2. I started with this PR as it looked like a very simple way of getting solid performance gains (if configured with multiple runtimes) + fixing the **2.** issue in the context description. I don't think this is the case anymore.

Now that we know there is no such trivial fix (unless complicating our code further), that would give us significant performance gains (even though far from optimal), we should probably focus on the harder task directly, i.e. reindexing blocks in parallel.

In fact we could go back to original idea, where we change `TrackRuntime` to `TrackRuntimes` so that we can trigger block history reindex for all runtimes at the same time, but in parallel for different consensus height intervals? If we go backwards with reindex this could even solve **2.**. Alternatively, doing history reindex for every runtime in parallel, for all consensus blocks in parallel seems wasteful...?

cc @kostko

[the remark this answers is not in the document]
peternose: > Handling this is actually quite cumbersome

Yes, because you are doing the opposite thing what the previous authors did. The service is parsing commands sequentially, so that you don't need to worry about this stuff. 

> looked like a very simple way of getting solid performance gains

It didn’t look that way to me. I already mentioned that runtime optimization isn’t that relevant. What matters is optimizing the processing of consensus blocks. To achieve this, you need to introduce concurrency, which makes the task more complex rather than just a few lines of code.

> focus on the harder task directly, i.e. reindexing blocks in parallel

Didn't I already said, that your solution probably cannot be upgraded to reindexing blocks in parallel.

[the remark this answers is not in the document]
martintomazic: >Not sure if this is fine. If the go scheduler pauses before the query is registered, we could miss few events. If we miss blocks also, this is not fine.

You are right since we call ` sc.GetRuntimeState(ctx, &api.RuntimeRequest)` right after this is registered, meaning what I did in this PR was probably off for another reason.

PR #6037, opened 2025-02-04
Fix observer role registration

### LINKED ISSUE #6019: Observer role registration either fails or is incomplete when multiple runtimes are enabled
##### SUMMARY

A race condition happens when registering an observer node with multiple runtimes enabled on Testnet. It sometimes fails with `node update not allowed` (removal of runtimes is not allowed), sometimes registers with an empty list (should not happen), and sometimes only registers for Emerald Testnet runtime (the only non-TEE). Restarting the node still leaves it in one of the failed states. Unless something else is missing, after the next epoch transition (when the node registration is retried) it successfully re-registers for all runtimes.

It seems the registration process is separate from runtime readiness checks. It registers as only an observer node for Emerald Testnet, because TEE runtimes are not yet ready, but once they become ready it does not re-register. This might also explain why it sometimes complains with `node update not allowed` (because other runtimes are not yet running and it attempts to register only for Emerald Testnet what represents a downgrade).

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

`worker/registration`

##### OASIS NODE VERSION

```paste below
24.3.1
```

##### OS / ENVIRONMENT

Ubuntu 22.04

##### STEPS TO REPRODUCE

1. Set up a ParaTime Client node to register as an observer role for all runtimes on Testnet (needs whitelisting)
```
mode: client
registration:
  entity_id: ...
runtime:
  paths:
    - /node/runtimes/cipher-testnet.orc
    - /node/runtimes/emerald-testnet.orc
    - /node/runtimes/sapphire-testnet.orc
    - /node/runtimes/pontusx-testnet.orc
    - /node/runtimes/pontusx-devnet.orc
```
2. Check registration status in `control status` output, you get one of:
```
  "registration": {
    "last_attempt_successful": false,
    "last_attempt_error_message": "registry: node update not allowed",
## or:
  "registration": {
    "last_attempt_successful": true,
    "descriptor": {
      "runtimes": null,
      "roles": "observer",
## or:
  "registration": {
    "last_attempt_successful": true,
    "descriptor": {
      "runtimes": [
        { "id": "00000000000000000000000000000000000000000000000072c8215e60d5bca7",... }
      ],
      "roles": "observer",
```
Right after it successfully registers for Emerald Testnet, the status of other runtimes is:
```
    "status": "ready",
        "status": "waiting for hosted runtime provision",
        "status": "syncing rounds",
        "status": "waiting for hosted runtime provision",
        "status": "syncing rounds",
        "status": "waiting for hosted runtime provision",
        "status": "syncing rounds",
        "status": "ready",
        "status": "syncing rounds",
        "status": "waiting for hosted runtime provision",
        "status": "syncing rounds",
```

##### EXPECTED RESULTS

Successful registration as observer node for all runtimes.

abukosek:
Closes #6019.


go/worker/client/worker.go
```diff
@@ -58,8 +56,26 @@ func (w *Worker) Start() error {
...
+		for id, rt := range w.runtimes {
 			<-rt.Initialized()
+
+			// After the runtime is initialized, register observer role if applicable.
+			switch config.GlobalConfig.Mode {
+			case config.ModeClient, config.ModeStatelessClient:
+				// When a node is a client node and it has an entity
+				// configured, we register it with the observer role
+				// as this may be needed for confidential runtimes.
+				rp, err := w.registration.NewRuntimeRoleProvider(node.RoleObserver, id)
```
kostko, line 68: I think the role provider needs to be created before initialization starts, so it blocks registration until declared available.

go/worker/client/worker.go
```diff
@@ -131,14 +137,19 @@ func (w *Worker) registerRuntime(commonNode *committeeCommon.Node) error {
...
+		// for confidential runtimes.
 		rp, err := w.registration.NewRuntimeRoleProvider(node.RoleObserver, id)
 		if err != nil {
 			return fmt.Errorf("failed to create role provider: %w", err)
 		}
 
-		rp.SetAvailable(commonNode.RegisterNodeRuntime)
+		// Set as unavailable before initialization starts, so the registration
+		// is blocked until the runtime finishes initializing.
+		rp.SetUnavailable()
```
kostko, line 150: I believe this is the default so no need to set it explicitly.

abukosek:
OK, this should work now, please re-review :)

go/worker/client/worker.go
```diff
@@ -58,8 +59,13 @@ func (w *Worker) Start() error {
 
 	// Wait for all runtimes to be initialized.
 	go func() {
-		for _, rt := range w.runtimes {
+		for id, rt := range w.runtimes {
 			<-rt.Initialized()
+
+			// After the runtime is finished initializing, enable registration if applicable.
```
peternose, line 65:
```suggestion
```

go/worker/client/worker.go
```diff
@@ -185,13 +196,14 @@ func New(
...
-		enabled:      enabled,
-		commonWorker: commonWorker,
-		registration: registration,
-		runtimes:     make(map[common.Namespace]*committee.Node),
-		quitCh:       make(chan struct{}),
-		initCh:       make(chan struct{}),
-		logger:       logging.GetLogger("worker/client"),
+		enabled:             enabled,
+		commonWorker:        commonWorker,
+		registration:        regWorker,
```
peternose, line 201:
```suggestion
		registrationWorker:        registrationWorker,
```
Not the best name, but just to be consistent with other code.

go/worker/client/worker.go
```diff
@@ -131,14 +137,19 @@ func (w *Worker) registerRuntime(commonNode *committeeCommon.Node) error {
...
 		rp, err := w.registration.NewRuntimeRoleProvider(node.RoleObserver, id)
 		if err != nil {
 			return fmt.Errorf("failed to create role provider: %w", err)
 		}
 
-		rp.SetAvailable(commonNode.RegisterNodeRuntime)
+		// The role provider is set as unavailable by default and we want to
+		// leave it like that before initialization starts, so the registration
+		// is blocked until the runtime finishes initializing.
+
```
peternose, line 151:
```suggestion
```

go/worker/client/worker.go
```diff
@@ -58,8 +59,13 @@ func (w *Worker) Start() error {
 
 	// Wait for all runtimes to be initialized.
 	go func() {
-		for _, rt := range w.runtimes {
+		for id, rt := range w.runtimes {
 			<-rt.Initialized()
+
+			// After the runtime is finished initializing, enable registration if applicable.
+			if rp, ok := w.runtimeRoleProvider[id]; ok {
+				rp.SetAvailable(w.commonWorker.GetRuntime(id).RegisterNodeRuntime)
```
peternose, line 67: It would be still nice to check if runtime exits.

go/worker/client/worker.go
abukosek: Why do you want to remove this comment?

go/worker/client/worker.go
abukosek: If I remove this blank line, it will look like the assignment below is somehow related to the text in the comment, which it is not.

go/worker/client/worker.go
abukosek: Is it even possible in this situation for `GetRuntime` to fail?  It iterates over basically the same list of runtimes, which is used to create the local `w.runtimes` map in the beginning.

go/worker/client/worker.go
peternose: The code says it all, the comment is not needed.

go/worker/client/worker.go
peternose: It actually is, because of that reason you are storing this provider locally. You can leave it as it is.

go/worker/client/worker.go
peternose: Probably not, unless we in the future change something (e.g. support tmp runtimes).

go/worker/client/worker.go
```diff
@@ -58,8 +59,13 @@ func (w *Worker) Start() error {
 
 	// Wait for all runtimes to be initialized.
 	go func() {
-		for _, rt := range w.runtimes {
+		for id, rt := range w.runtimes {
 			<-rt.Initialized()
```
kostko, line 63: Are you sure this is waiting long enough? An observer should only be deemed initialized once the runtimes are fully provisioned.

See the executor (`go/worker/compute/executor/committee/node.go`) as an example.
https://github.com/oasisprotocol/oasis-core/blob/299a63f08e5a511a0dc7f9987cbef51cb28cf543/go/worker/compute/executor/committee/node.go#L954-L984

kostko: I guess what the executor is doing is overkill for the client, but it should at least wait for the runtime to be started, e.g. after receiving this event:
https://github.com/oasisprotocol/oasis-core/blob/299a63f08e5a511a0dc7f9987cbef51cb28cf543/go/worker/common/committee/node.go#L588-L602

kostko: You could handle this in the client event hook:
https://github.com/oasisprotocol/oasis-core/blob/299a63f08e5a511a0dc7f9987cbef51cb28cf543/go/worker/client/committee/node.go#L91-L94

peternose: You are right, `initCh` is closed almost immediately, not waiting the runtime to start. Maybe we need to change this, because the common worker has a different approach.

https://github.com/oasisprotocol/oasis-core/blob/master/go/worker/common/committee/node.go#L728-L729

kostko: The client worker first waits for the common worker to initialize, so this is already taken into account. But this still does not wait for the runtime to initialize.

go/worker/client/committee/node.go
```diff
@@ -89,8 +91,15 @@ func (n *Node) HandleNewBlockLocked(*runtime.BlockInfo) {
 }
 
 // HandleRuntimeHostEventLocked is guarded by CrossNode.
-func (n *Node) HandleRuntimeHostEventLocked(*host.Event) {
-	// Nothing to do here.
+func (n *Node) HandleRuntimeHostEventLocked(ev *host.Event) {
+	if n.roleProvider != nil {
```
kostko, line 95:
```suggestion
	if n.roleProvider == nil {
		return
	}
```
And outdent the below.

go/worker/client/committee/node.go
```diff
@@ -89,8 +91,15 @@ func (n *Node) HandleNewBlockLocked(*runtime.BlockInfo) {
...
+		switch {
+		case ev.Started != nil:
+			n.roleProvider.SetAvailable(n.commonNode.RegisterNodeRuntime)
+		case ev.FailedToStart != nil, ev.Stopped != nil:
+			n.roleProvider.SetUnavailable()
+		}
```
kostko, line 101: Also add an explicit `default` that does nothing.

go/worker/client/worker.go
```diff
@@ -167,7 +177,7 @@ func (w *Worker) registerRuntime(commonNode *committeeCommon.Node) error {
 func New(
 	grpcInternal *grpc.Server,
 	commonWorker *workerCommon.Worker,
-	registration *registration.Worker,
+	registrationWorker *registration.Worker,
```
kostko, line 180: This change is no longer needed.

kostko (commented):
Some nits, otherwise I think this looks good now.

go/worker/client/committee/node.go
```diff
@@ -366,14 +378,15 @@ func (n *Node) worker() {
 }
 
 // NewNode creates a new client node.
-func NewNode(commonNode *committee.Node) (*Node, error) {
+func NewNode(commonNode *committee.Node, roleProvider registration.RoleProvider) (*Node, error) {
```
peternose, line 381:
```suggestion
func NewNode(commonNode *committee.Node, opts ...NodeOption) (*Node, error) {
```
Optional parameters should be specified differently, e.g. via variadic options. It is "ugly" to accept `nil` parameters.

go/worker/client/worker.go
```diff
@@ -129,21 +129,31 @@ func (w *Worker) registerRuntime(commonNode *committeeCommon.Node) error {
...
 		if err != nil {
 			return fmt.Errorf("failed to create role provider: %w", err)
 		}
 
-		rp.SetAvailable(commonNode.RegisterNodeRuntime)
+		// The role provider is set as unavailable by default and we want to
+		// leave it like that before initialization starts, so the registration
+		// is blocked until the runtime finishes initializing.
+		// The availability of the role provider is changed in
+		// HandleRuntimeHostEventLocked in worker/client/committee/node.go.
```
peternose, line 151: This is not the best because no one will fix this comment if the code in `node.go` changes. I would generalize this and keep it simple, if I cannot remove it.

Also it is not the best that unavailability is set here, and availability in the other file, as it is hard to track who has control over this flag. Can we maybe move this to `node.go`?

peternose: We could move whole `switch` or add option `WithRoleProvider(roles []Role)`.

go/worker/client/committee/node.go
abukosek: I disagree :P  This is just a single option, it makes no sense to add another layer of abstraction just to pass that.  If we have more optional flags for `NewNode` in the future, we can add the option struct then.

go/worker/client/worker.go
abukosek: Unavailability is the default state and the role provider is created in this file, so...  The executor node also has the role provider passed from outside, so this code is consistent with the existing way this is done.

I still think a comment pointing to the other part of the code is useful, because it saves time the next time someone has to look at this code.  I can't imagine why the code in `node.go` would move from that event handler, the executor node does it the same way.

go/worker/client/committee/node.go
peternose: I disagree again :D Once you allow accepting `nil` parameters, even if that is only one, programmers should start questioning themselves which fields can be `nil` and which not. In general, that should be forbidden, but if you allow it, it should be explicit (e.g. with options).

go/worker/client/worker.go
peternose: > Unavailability is the default state and the role provider is created in this file, so...

I would remove the comment as this is a know fact. The executor and storage worker don't comment that neither. 

> The executor node also has the role provider passed from outside, so this code is consistent with the existing way this is done.

It is not consistent, as executor assumes it is not `nil`, while this one does. We have a similar problem in storage workers, where `roleProvider` should not be `nil`, while `rpcRoleProvider` can be.

> I still think a comment pointing to the other part of the code is useful, because it saves time the next time someone has to look at this code.

It is useful if it is valid. Once this comment becomes invalid, things become ugly.

> I can't imagine why the code in `node.go` would move from that event handler, the executor node does it the same way.

For example, if I change the name `HandleRuntimeHostEventLocked` to `HandleEventLocked` or file `node.go` to `committee_node.go`, this comment becomes invalid. And these are just trivial cases. 

You can leave it as it is, since other workers do the same. Just saying that this is not the best.

peternose (approved):

go/worker/client/committee/node.go
abukosek: :D
I mean, this part of the code is not public-facing and the executor node does it the same way :P Adding the role provider as an option would just overcomplicate the rest of the code with needless boilerplate -- `NewNode` is only called from within this file.  In this particular case, I think it's OK and not confusing.

go/worker/client/worker.go
abukosek: The comment would still point the developer in the right direction, even if the name of the handler or file changes -- this is still useful information, especially, since the codebase is so spread out.  I've added it because it's something I miss a lot in other parts of the code that are in different locations, but interconnected.

If someone renames the handler/file, it's their job to grep the source code for any mentions of the old name and change it :)

go/worker/client/committee/node.go
peternose: This kind of questions arise if you allow nil values and have no idea if some fields can be `nil`:
https://github.com/oasisprotocol/oasis-core/pull/6050/files#r1950633603

PR #6038, opened 2025-02-05
go/runtime/host/sgx: Retry re-attestation faster on failure

### LINKED ISSUE #6036: Retry attestation refresh more often in case of failure
Currently attestation is refreshed on a fixed interval, even in case it fails. This should be changed so in case it fails, it gets retried more often, possibly using exponential backoff.

kostko:
Fixes #6036

### DIFF
--- a/go/runtime/host/sgx/common/common.go
+++ b/go/runtime/host/sgx/common/common.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
+
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
 	"github.com/oasisprotocol/oasis-core/go/common/identity"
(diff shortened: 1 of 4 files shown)


.changelog/6038.feature.md
```diff
@@ -0,0 +1 @@
+go/runtime/host/sgx: Retry re-attestation faster on failure
```
peternose, line 1: This could also be trivial, instead of a feature.

go/runtime/host/sgx/common/common.go
```diff
@@ -155,18 +157,27 @@ func AttestationWorker(
...
+	var retryTicker *backoff.Ticker
+
 	logger = logger.With("runtime_id", hp.Runtime.ID())
 
 	// Get the event emitter.
 	eventEmitter, _ := hp.Runtime.(host.RuntimeEventEmitter)
 
 	for {
+		var retryCh <-chan time.Time
+		if retryTicker != nil {
```
peternose, line 169: I would keep this code together, e.g.
```go
	var (
		retryTicker *backoff.Ticker
		retryCh     <-chan time.Time
	)

	for {
		capabilityTEE, err := updateCapabilityFunc(context.Background(), hp)
		if err != nil {
			// Setup a retry ticker so we retry attestation faster than the configured interval.
			if retryTicker == nil {
				expBackoff := backoff.NewExponentialBackOff(
					backoff.WithMaxElapsedTime(0), // Never stop.
				)
				retryTicker = backoff.NewTicker(expBackoff)
				retryCh = retryTicker.C
			}

			continue
		}

		// Clear retry ticker after successful attestation.
		if retryTicker != nil {
			retryTicker.Stop()
			retryTicker = nil
			retryCh = nil
		}
	}
```

peternose (approved):

go/runtime/host/sgx/common/common.go
```diff
@@ -181,9 +192,24 @@ func AttestationWorker(
...
+				)
+				retryTicker = backoff.NewTicker(expBackoff)
+			}
+
 			continue
 		}
 
+		// Clear retry ticker after successful attestation.
+		if retryTicker != nil {
+			retryTicker.Stop()
```
peternose, line 209: We should probably also stop the ticker if we return and is not nil.

go/runtime/host/sgx/common/common.go
kostko: It is not taken from committee node, at least not on purpose. Where did the select go?

kostko: Also if we do it like this we need to always reset one more thing.

go/runtime/host/sgx/common/common.go
kostko: Fixed.

.changelog/6038.feature.md
kostko: Changed.

go/runtime/host/sgx/common/common.go
peternose: >  Where did the select go?

Removed it for brevity. 

> Also if we do it like this we need to always reset one more thing.

Yes, but we have one `if` less and less scattered code for retry.

peternose (approved):

Issue #6039, opened 2025-02-05
Inconsistent state response when (explicitly) querying the latest block


ptrus:
`beacon.GetEpoch` query via the gRPC for the epoch transition block is inconsistent right after the epoch transition happens. Occasionally, for the epoch transition block, the `GetEpoch` call (with explicitly set height) for the latest block returns the previous epoch. Later if the query is repeated the correct (transitioned) epoch is returned.

Pontentially due to:
https://github.com/oasisprotocol/oasis-core/blob/9e138d8bb78cc579e15bfb34f095cee5a2f178e4/go/consensus/cometbft/api/state.go#L297-L299

since we are explicitly querying the to-be processed block, we actually might get state of the previous block due to the above.

PR #6040, opened 2025-02-05
go/consensus/cometbft: Fail ImmutableState creation if version is missing

ptrus:
Fixes: #6039

### DIFF
--- a/go/consensus/cometbft/abci/state.go
+++ b/go/consensus/cometbft/abci/state.go
@@ -283,17 +283,25 @@ func (s *applicationState) GetCurrentEpoch(ctx context.Context) (beacon.EpochTim
 	if blockHeight == 0 {
 		return beacon.EpochInvalid, nil
 	}
+
+	latestHeight := blockHeight
+	if abciCtx := api.FromCtx(ctx); abciCtx != nil {
+		// If request was made from an ABCI application context, then use blockHeight + 1, to fetch
+		// the epoch at current (future) height. See cometbft/api.NewImmutableState for details.
+		latestHeight++
+	}
+
 	// Check if there is an epoch transition scheduled for the current height. This should be taken
 	// into account when GetCurrentEpoch is called before the time keeping app does the transition.
-	future, err := s.timeSource.GetFutureEpoch(ctx, blockHeight+1)
+	future, err := s.timeSource.GetFutureEpoch(ctx, latestHeight)
 	if err != nil {
-		return beacon.EpochInvalid, fmt.Errorf("failed to get future epoch for height %d: %w", blockHeight+1, err)
+		return beacon.EpochInvalid, fmt.Errorf("failed to get future epoch for height %d: %w", latestHeight, err)
 	}
 	if future != nil && future.Height == blockHeight+1 {
 		return future.Epoch, nil
 	}
 
-	currentEpoch, err := s.timeSource.GetEpoch(ctx, blockHeight+1)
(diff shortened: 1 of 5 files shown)


go/consensus/cometbft/abci/state.go
```diff
@@ -278,28 +278,39 @@ func (s *applicationState) GetEpoch(ctx context.Context, blockHeight int64) (bea
 	return s.timeSource.GetEpoch(ctx, blockHeight)
 }
 
-func (s *applicationState) GetCurrentEpoch(ctx context.Context) (beacon.EpochTime, error) {
+func (s *applicationState) getCurrentEpoch(ctx context.Context, inCommitPhase bool) (beacon.EpochTime, error) {
```
ptrus, line 281: This is not exposed because it's a bit awkward for the caller to have to know whether the method is being called during the commit phase or not. At the moment, only a single internal method (`maybeHaltForUpgrade`) calls this during the commit phase, so there's no need to expose it. Also, if `GetCurrentEpoch` is used in such a case in the future, it will be easily noticeable since the call will fail, and we can refactor then.

I believe a better solution would be for `applicationState` to be aware of the latest available state height and simply use that. I'm not sure if there's an easy way to achieve that.

One alternative I implemented before was for `GetCurrentEpoch` to first try `s.BlockHeight() + 1` and, if it fails with "version not available," internally fall back to querying the state at `s.BlockHeight()`. That approach works as well, but in my opinion it is worse since we "blindly" assume that it's acceptable for the state at +1 to be unavailable.

go/consensus/cometbft/api/state.go
```diff
@@ -294,7 +294,10 @@ func NewImmutableState(ctx context.Context, state ApplicationQueryState, version
 	if state.BlockHeight() == 0 {
 		return nil, consensus.ErrNoCommittedBlocks
 	}
-	if version <= 0 || version > state.BlockHeight() {
+	if version > state.BlockHeight() {
+		return nil, consensus.ErrVersionNotFound
```
kostko, line 298: Would be great to have a test that covers this case.

go/consensus/cometbft/apps/staking/slashing.go
```diff
@@ -69,7 +69,7 @@ func onEvidenceByzantineConsensus(
 	// validator from being scheduled in the next epoch.
 	if penalty.FreezeInterval > 0 {
 		var epoch beacon.EpochTime
-		epoch, err = ctx.AppState().GetEpoch(context.Background(), ctx.BlockHeight()+1)
+		epoch, err = ctx.AppState().GetEpoch(context.Background(), ctx.BlockHeight())
```
kostko, line 72: Why this change given that semantics of `GetEpoch` has not changed? It seems like this is consensus-breaking.

ptrus: I have also updated the `GetCurrentEpoch` code, so that it now similarly checks the provided context, and uses the appropriate height based on that, which should ensure consistent code with existing.

kostko (approved):

go/consensus/cometbft/apps/staking/slashing.go
ptrus: Thanks for pointing this out, so I looked a bit more into it. I have added a comment for now. I also updated the `EpochChanged` fn in `applicationState` to similarly use the correct height, although that one is currently never used outside of abciCtx.

PR #6044, opened 2025-02-07
go/runtime/bundle: Don't abort when some manifests cannot be loaded

kostko:
When upgrading from earlier versions the exploded bundle directory may contain non-bundle subdirectories which should not cause the loader to fail.


peternose (approved):
This solves the problem, but the folder structure will remain invalid indefinitely. If any abnormalities occur, the cached folder should be completely wiped. Maybe we do that in another task.

kostko:
Removing unknown stuff without letting the user know is always somewhat dangerous as the data could be useful. Maybe it is ok in this case since the whole directory is a cache, but let's defer this to another PR.

PR #6055, opened 2025-02-11
Hacky benchmark of history reindex

martintomazic:
Hacky Bench-marking only:

# Warning
Ignore code quality etc, this is benchmark only. Corner case logic is also probably off sometimes, logs are also often just copied.

```bash
 prune:
        strategy: none
```
batching has minor bug if you are doing prunning.

# How to benchmark
Change: 
* `writeWorkers` if you want to write in parallel
* `batchSize` if you want to batch`
* `readWorkers` if withing a batch processing you want to fetch consensus blocks and data in parallel.

** `writeWorkers=1` + `batchSize=1` should give you same benchmark as is on the `main`.** 

grepping for `history reindex0` should output relevant logs.

### DIFF
--- a/.changelog/5738.feature.md
+++ b/.changelog/5738.feature.md
@@ -1,0 +1,4 @@
+go/roothahs: Optimize runtime history reindex
+
+During runtime history reindex, we batch 1000 writes at the same time,
+resulting in 2x speed-up of history reindex.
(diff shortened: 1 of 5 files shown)


martintomazic:
Closing with #5738.

Issue #6014, opened 2025-01-28
Decouple ROFL event notifier and sub-handler


peternose:
The ROFL event notifier should be decoupled from the ROFL sub-handler to ensure that these structures are independent of each other. Additionally, the ROFL event notifier should terminate when the ROFL component is stopped (e.g., during an upgrade) to prevent resource leaks.

Ideally, every RONL component should have a handler and a notifier, and every ROFL component should have a sub-handler and a sub-notifier.

kostko:
This is also causing problems with notifications not being delivered to components in case there are multiple components and some are not working. The reason is sequential call propagation with a shared context. When we have sub-notifiers, this somewhat hacky call propagation can be removed.

PR #6073, opened 2025-02-20
go/runtime/registry: Remove runtime host handler factory

peternose:
Refactored committee node:
- Removed `RuntimeHostHandlerFactory` as it is more cleaner to pass parameters to the constructor instead of passing a factory.
- Removed handler and notifier from the committee node, so that we can easier stop/start the notifier.
- Removed host provisioner (which is now generalized to provision any runtime) from the runtime registry, as it doesn't belong there.
  - It is now globally avaliable to common worker which provisions runtimes and to the key manager worker which provisions key manager runtime (maybe we need to move up also handler and notifier).

Related to:
- #6014

### DIFF
--- a/go/runtime/host/provisioner/provisioner.go
+++ b/go/runtime/host/provisioner/provisioner.go
@@ -1,0 +1,203 @@
+package provisioner
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+
+	"github.com/oasisprotocol/oasis-core/go/common/identity"
+	"github.com/oasisprotocol/oasis-core/go/common/persistent"
+	"github.com/oasisprotocol/oasis-core/go/common/sgx/pcs"
+	"github.com/oasisprotocol/oasis-core/go/config"
+	consensus "github.com/oasisprotocol/oasis-core/go/consensus/api"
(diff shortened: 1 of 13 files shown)


go/runtime/host/provisioner/provisioner.go
```diff
@@ -0,0 +1,198 @@
...
+	hostLoadBalance "github.com/oasisprotocol/oasis-core/go/runtime/host/loadbalance"
+	hostMock "github.com/oasisprotocol/oasis-core/go/runtime/host/mock"
+	hostProtocol "github.com/oasisprotocol/oasis-core/go/runtime/host/protocol"
+	hostSandbox "github.com/oasisprotocol/oasis-core/go/runtime/host/sandbox"
+	hostSgx "github.com/oasisprotocol/oasis-core/go/runtime/host/sgx"
+	hostTdx "github.com/oasisprotocol/oasis-core/go/runtime/host/tdx"
+	"github.com/oasisprotocol/oasis-core/go/runtime/registry"
+)
+
+func New(
```
kostko, line 29: Add documentation comment.

kostko (approved):

PR #6078, opened 2025-02-22
go/keymanager/api: Fix key manager client

peternose:
Moving key manager backend from `ServicesBackend` to `ClientBackend`.

I also refactored clients, trying to follow these guidelines: 
- First define struct, then implement constructor, and finally implement all methods.
- Accept interfaces, return structs.

### DIFF
--- a/go/keymanager/secrets/grpc.go
+++ b/go/keymanager/secrets/grpc.go
@@ -282,12 +282,12 @@ func RegisterService(server *grpc.Server, service Backend) {
 	server.RegisterService(&serviceDesc, service)
 }
 
-// Client is a gRPC keymanager secrets client.
+// Client is a gRPC key manager secrets client.
 type Client struct {
 	conn *grpc.ClientConn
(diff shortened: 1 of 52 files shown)


go/keymanager/secrets/grpc.go
```diff
@@ -12,8 +12,10 @@ import (
 
 var (
 	// serviceName is the gRPC service name.
-	serviceName = cmnGrpc.NewServiceName("KeyManager")
+	serviceName = cmnGrpc.NewServiceName("KeyManager.Secrets")
```
peternose, line 15: Is this too breaking (not sure who even calls this endpoint, since all nodes run consensus)?

kostko: Should be fine.

kostko: Should be fine. Maybe expose both for a few versions and use the new one in clients?

PR #6083, opened 2025-02-25
go/consensus/cometbft/roothash: Publish the last block on subscribe 

peternose:
Replaying the latest block with a hook was not ideal because if the broker still had blocks in the queue while adding a new subscription, it could result in an unexpected block order and potentially cause the last block to be replayed twice. 

Solution with monotonically increasing order solved this problem, but there is a simpler one where we use broker's option to publish the last block on subscription.


kostko (approved):
:rofl:

Issue #6085, opened 2025-02-27
Fix `runtime.History.History.Watchblocks` to emit sequential blocks (no gaps)


martintomazic:
Subscription to `runtime.client.api.WatchBlocks` ([see](https://github.com/oasisprotocol/oasis-core/blob/f94b700b8edcd3e4fa228e748ca45d7c91a3bf06/go/worker/client/service.go#L116)) may return block sequence with missing rounds. 

This may happen because  `runtime.history.History` is not notifying blocks if they are `Committ`-ed via `notify=false` (during reindex)**.

 Note this is only problematic if node `!hasLocalStorage`, as otherwise storage syncing (`worker.storage.committee.Node`) is responsible  for notifying blocks via `StorageSyncCheckpoint`. This worker subscribes to `consensus.cometbft.roothash.ServiceClient.WatchBlocks`, which may have gaps, however it fetches missing rounds. 

As discussed in private, `roothash.ServiceClient` should probably not be responsible for reindexing, nor should `roothahs.BlockHistory` know anything about storage, node types etc... Likely this code needs a thorough refactor.

** We are guaranteed to have at least two reindexes, possibly more if there are errors during subsequent event processing.

martintomazic:
> As discussed in private, roothash.ServiceClient should probably not be responsible for reindexing, nor should roothash.BlockHistory know anything about storage, node types etc... Likely this code needs a thorough refactor.

Idealy, we move [reindexing part](https://github.com/oasisprotocol/oasis-core/blob/master/go/consensus/cometbft/roothash/roothash.go#L680-L718) from `roothash.ServiceClient` to something like `worker.runtime.xxx`. 
The worker should be initialized, just like others, after we have [initialized `RuntimeRegistry`](https://github.com/oasisprotocol/oasis-core/blob/4d22d4d3a0e7ce8b7d767ddbff29c7da3ac73d8a/go/oasis-node/cmd/node/node.go#L228). Worker should  first fetch `runtime.history.History.LastConsensusHeight` and reindex until `n` blocks away from consensus sync. At this point, it should subscribe to `roothash.WatchBlocks`, do another reindex to fill possible gap and then proceed with regular processing. After that, blocks should be always notified without any gaps.

What about problematic [`runtime.history.History`](https://github.com/oasisprotocol/oasis-core/blob/master/go/runtime/history/history.go)?
* `runtime.history.History` does two things: 
    1. it implements `roothash.BlockHistory`, with pruning policy 
[...]

PR #6089, opened 2025-02-28
 go/consensus/cometbft/roothash: Remove history from roothash service

peternose:
The runtime history should be built on top of the roothash service, using observed blocks for block reindexing and round syncing. The roothash service should broadcast blocks as soon as they are finalized in consensus, whereas the block history should broadcast blocks only after they have been reindexed or synced.

Now, we have 3 ways to observe runtime blocks:

1. Method `consensus.RootHash().WatchBlocks` returns blocks immediately once they are finalized in the consensus.
2. Method `rt.History().WatchCommittedBlocks` returns blocks immediately once they are committed to local storage.
3. Method `rt.History().WatchBlocks` returns blocks immediately once they are committed and synced to local storage.

Comparing methods to previous solution:
- Previous 1 == Current 2

### DIFF
--- a/go/consensus/cometbft/roothash/roothash.go
+++ b/go/consensus/cometbft/roothash/roothash.go
@@ -27,10 +27,7 @@ import (
 	runtimeRegistry "github.com/oasisprotocol/oasis-core/go/runtime/registry"
 )
 
-const (
-	crashPointBlockBeforeIndex = "roothash.before_index"
-	reindexWriteBatchSize      = 1000
-)
+const crashPointBlockBeforeIndex = "roothash.before_index"
 
 // ServiceClient is the roothash service client interface.
 type ServiceClient interface {
--- a/go/consensus/cometbft/roothash/roothash.go
+++ b/go/consensus/cometbft/roothash/roothash.go
@@ -585,7 +585,7 @@ func (sc *serviceClient) DeliverCommand(ctx context.Context, height int64, cmd i
 		})
 		if err != nil {
 			sc.logger.Warn("failed to get runtime state for latest block",
-				"err",
+				"err", err,
 				"runtime_id", tr.runtimeID,
 				"height", height,
 			)
@@ -809,7 +809,7 @@ EventLoop:
 				// An executor commit has been processed.
 				var e api.ExecutorCommittedEvent
 				if err := eventsAPI.DecodeValue(val, &e); err != nil {
-					errs = errors.Join(errs, fmt.Errorf("roothash: corrupt ExecutorComitted event: %w", err))
+					errs = errors.Join(errs, fmt.Errorf("roothash: corrupt ExecutorCommitted event: %w", err))
 					continue EventLoop
 				}
 
--- a/go/consensus/cometbft/roothash/roothash.go
+++ b/go/consensus/cometbft/roothash/roothash.go
@@ -46,15 +46,12 @@ type runtimeBrokers struct {
 
 type trackedRuntime struct {
 	runtimeID common.Namespace
-
-	height       int64
-	blockHistory api.BlockHistory
-	reindexDone  bool
+	height    int64
+	round     uint64
 }
 
 type cmdTrackRuntime struct {
-	runtimeID    common.Namespace
-	blockHistory api.BlockHistory
+	runtimeID common.Namespace
 }
 
 type serviceClient struct {
@@ -187,7 +184,7 @@ func (sc *serviceClient) WatchBlocks(_ context.Context, id common.Namespace) (<-
 	sub.Unwrap(ch)
 
 	// Start tracking this runtime if we are not tracking it yet.
-	if err := sc.trackRuntime(sc.ctx, id, nil); err != nil {
+	if err := sc.trackRuntime(sc.ctx, id); err != nil {
 		sub.Close()
 		return nil, nil, err
 	}
@@ -211,7 +208,7 @@ func (sc *serviceClient) WatchEvents(_ context.Context, id common.Namespace) (<-
 	sub.Unwrap(ch)
 
 	// Start tracking this runtime if we are not tracking it yet.
-	if err := sc.trackRuntime(sc.ctx, id, nil); err != nil {
+	if err := sc.trackRuntime(sc.ctx, id); err != nil {
 		sub.Close()
 		return nil, nil, err
 	}
@@ -227,7 +224,7 @@ func (sc *serviceClient) WatchExecutorCommitments(_ context.Context, id common.N
 	sub.Unwrap(ch)
 
 	// Start tracking this runtime if we are not tracking it yet.
-	if err := sc.trackRuntime(sc.ctx, id, nil); err != nil {
+	if err := sc.trackRuntime(sc.ctx, id); err != nil {
 		sub.Close()
 		return nil, nil, err
 	}
@@ -236,15 +233,14 @@ func (sc *serviceClient) WatchExecutorCommitments(_ context.Context, id common.N
 }
 
 // Implements api.Backend.
-func (sc *serviceClient) TrackRuntime(ctx context.Context, history api.BlockHistory) error {
+func (sc *serviceClient) TrackRuntime(_ context.Context, history api.BlockHistory) error {
 	sc.pruneHandler.trackRuntime(history)
-	return sc.trackRuntime(ctx, history.RuntimeID(), history)
+	return nil
 }
 
-func (sc *serviceClient) trackRuntime(ctx context.Context, id common.Namespace, history api.BlockHistory) error {
+func (sc *serviceClient) trackRuntime(ctx context.Context, id common.Namespace) error {
 	cmd := &cmdTrackRuntime{
-		runtimeID:    id,
-		blockHistory: history,
+		runtimeID: id,
 	}
 
 	select {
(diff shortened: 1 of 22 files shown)


go/runtime/history/indexer.go
```diff
@@ -0,0 +1,256 @@
...
+		case <-ctx.Done():
+			return
+		}
+	}
+
+	// Handle new blocks after reindex is complete.
+	for {
+		if err = bi.commitBlock(ctx, blk, true); err != nil {
+			return
+		}
```
martintomazic, line 117: My understanding is previous logic would start reindexing again if error here. This was not best either since it could led to infinite reindexes. 

How about retrying few times here (exponential backoff?) and finally `panic` if not working?

We could do something similar in `emitLatestBlock` (where we call `sc.getLatestBlockAt`), just in case consensus state db is temporarily unavailable/there was unexpected disk error?

go/runtime/history/indexer.go
```diff
@@ -0,0 +1,256 @@
...
+func (bi *BlockIndexer) commitBlocks(ctx context.Context, blocks []*roothash.AnnotatedBlock, notify bool) error {
+	if len(blocks) == 0 {
+		return nil
+	}
+
+	var results []*roothash.RoundResults
+	for _, blk := range blocks {
+		result, err := bi.consensus.RootHash().GetLastRoundResults(ctx, &roothash.RuntimeRequest{
+			RuntimeID: bi.history.RuntimeID(),
+			Height:    blk.Height,
```
martintomazic, line 230: NIT: I would move this back to `reindexBlocks` since `commitBlocks` is now `fetchResultsAndCommitBlocks`. 

Might be also useful if we use worker pool for fetching reads. No strong opinion, since we could trivially do that at that point if needed.

go/runtime/registry/registry.go
```diff
@@ -524,16 +524,19 @@ func (r *runtimeRegistry) NewRuntime(ctx context.Context, runtimeID common.Names
...
 
 		// Start tracking this runtime.
-		if err = r.consensus.RootHash().TrackRuntime(ctx, history); err != nil {
+		if err = r.consensus.RootHash().TrackRuntime(ctx, rt.history); err != nil {
 			return nil, fmt.Errorf("runtime/registry: cannot track runtime %s: %w", runtimeID, err)
 		}
+
+		// Start indexing blocks.
+		indexer := history.NewBlockIndexer(r.consensus, rt.history)
+		indexer.Start()
```
martintomazic, line 539: Do we control this go routine?

martintomazic (commented):
Looks super solid, also unblocks a lot of our work.

Left some minor questions, and happy to approve after that.

// Hope I did not review too early as I just saw is still draft.

go/runtime/history/indexer.go
```diff
@@ -0,0 +1,256 @@
...
+	}()
+
+	// Reindex blocks up to the latest block.
+	var blk *roothash.AnnotatedBlock
+reindex:
+	for {
+		select {
+		case blk = <-blkCh:
+			// Send a new request with the latest height.
+			sendToChannel(reindexCh, blk.Height-1)
```
martintomazic, line 101: Reindex would never start  if runtime is suspended/there are no new blocks? 

`run` could be simplified by first fetching and re-indexing until last consensus height . Then we subscribe to `Roothash.WatchBlocks`. When (if) we receive first block, check if there is gap of consensus heights and reindex the gap as well.

go/worker/client/service.go
```diff
@@ -130,7 +130,11 @@ func (s *service) GetBlock(ctx context.Context, request *api.GetBlockRequest) (*
 	if err != nil {
 		return nil, err
 	}
-	return rt.History().GetBlock(ctx, request.Round)
+	blk, err := rt.History().GetSyncedBlock(ctx, request.Round)
+	if err != nil {
+		return nil, err
+	}
+	return blk.Block, nil
```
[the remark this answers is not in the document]
peternose: I guess it was possible, so my solution was not fine. I reverted these changes, so resolving.

go/runtime/history/indexer.go
peternose: > How about retrying few times here (exponential backoff?) and finally `panic` if not working?

Yes, we can retry. We should not panic, we should exit the indexer instead (ideally returning an error, so that other services can be gracefully stopped).
 
> We could do something similar in `emitLatestBlock` (where we call `sc.getLatestBlockAt`), just in case consensus state db is temporarily unavailable/there was unexpected disk error?

We could, but maybe in another PR if needed.

go/runtime/registry/registry.go
peternose: Fixed, together with startup of the runtime.

go/runtime/history/indexer.go
peternose: > Reindex would never start if runtime is suspended/there are no new blocks?

Reindex will always start because `WatchBlocks` always returns the latest block. So no-blocks scenario is not possible.
 
> `run` could be simplified by first fetching and re-indexing until last consensus height . Then we subscribe to `Roothash.WatchBlocks`. When (if) we receive first block, check if there is gap of consensus heights and reindex the gap as well.

Therefore, we don't need to implement your simplification.

go/runtime/history/indexer.go
```diff
@@ -0,0 +1,313 @@
...
+		}
+
+		if len(blks) > batchSize {
+			bi.logger.Error("terminating")
+			return
+		}
+
+		if err := bi.commitBlocks(blks); err != nil {
+			retry = boff.NextBackOff()
+			continue
```
martintomazic, line 112: So in practice if you fail to commitblock within `6s` it will automatically terminate the service? do you see a scenario where db is down for `20s` and ideally this should not terminate?

martintomazic (commented):
Do we want to add changelog given that this fixes:
1. #6085
2. In theory it also parallelizes reindex (should be much faster if many runtimes), since no reindexer is now run in a separate go routine?

go/runtime/history/indexer.go
martintomazic: >Reindex will always start because WatchBlocks always returns the latest block. So no-blocks scenario is not possible.

In case of first ever runtime block you are always guaranteed to have genesis block? 


>Therefore, we don't need to implement your simplification.

I am starting to think it could simplify `reindex` function a lot if we have recursive function (so no need for extra channells, go routine etc)? No strong opinion on that and definitely not blocking. + You never have many pending blocks inside `WatchBlocks` buffer (albeit it should be in order of `1k` so not problematic at all?)

go/runtime/history/indexer.go
martintomazic: >We could do something similar in emitLatestBlock (where we call sc.getLatestBlockAt), just in case consensus state db is temporarily unavailable/there was unexpected disk error?

My understanding is previously in case of error during `DeliverEvent` because of this case, we marked runtime for reindex.

I see a similar problem as in https://github.com/oasisprotocol/oasis-core/pull/6089#discussion_r1991220208, where due to ephemeral db error, you will now have gap in `roothash.WatchBlocks`? This kind of errors are extremely unlikely but they can happen.

go/runtime/history/indexer.go
```diff
@@ -0,0 +1,313 @@
...
+		case blk := <-blkCh:
+			blks = append(blks, blk)
+		case <-time.After(retry):
+		case <-ctx.Done():
+			bi.logger.Info("stopping")
+			return
+		}
+
+		if len(blks) > batchSize {
+			bi.logger.Error("terminating")
```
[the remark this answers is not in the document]
peternose: Yes, will add a messages. I didn't want to include it as in the logs should be seen that commit failed many times.

go/runtime/history/indexer.go
```diff
@@ -0,0 +1,313 @@
...
+
+	return nil
+}
+
+func (bi *BlockIndexer) reindexRange(ctx context.Context, start int64, end int64) error {
+	bi.logger.Debug("reindexing blocks",
+		"start_height", start,
+		"end_height", end,
+	)
+
```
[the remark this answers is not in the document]
peternose: I think no. Validation is omitted because this is an internal function. Method also allows you to commit batch of any size. However, the `reindexTo` should take care that input parameters are valid and limit blocks.

go/consensus/cometbft/roothash/roothash.go
```diff
@@ -236,15 +230,14 @@ func (sc *serviceClient) WatchExecutorCommitments(_ context.Context, id common.N
 }
 
 // Implements api.Backend.
-func (sc *serviceClient) TrackRuntime(ctx context.Context, history api.BlockHistory) error {
+func (sc *serviceClient) TrackRuntime(_ context.Context, history api.BlockHistory) error {
 	sc.pruneHandler.trackRuntime(history)
```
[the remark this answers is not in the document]
peternose: Yes, also `TrackRuntime` method doesn't belong to this backend. However, this is not that simple, as we need to move pruner from cometbft to consensus API. Opening an issue and can do it in another PR.

go/runtime/history/history.go
```diff
@@ -83,11 +84,14 @@ func (h *runtimeHistory) CommitBatch(blks []*roothash.AnnotatedBlock, notify boo
...
-	if h.hasLocalStorage || !notify {
+	if !notify {
 		return nil
 	}
 	for _, blk := range blks {
-		h.blocksNotifier.Broadcast(blk)
+		h.committedBlocksNotifier.Broadcast(blk)
+		if !h.hasLocalStorage {
+			h.syncedBlocksNotifier.Broadcast(blk)
+		}
```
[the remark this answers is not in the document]
peternose: Maybe. I intentionally used this name as this should be moved to storage worker. But if you assume that sync on nodes without local storage works differently, this name might be fine.

go/runtime/history/indexer.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+}
+
+func (bi *BlockIndexer) reindex(ctx context.Context, blkCh <-chan *roothash.AnnotatedBlock) error {
+	bi.logger.Debug("reindexing",
+		logging.LogEvent, roothash.LogEventHistoryReindexing,
+	)
+
+	var wg sync.WaitGroup
+	defer wg.Wait()
```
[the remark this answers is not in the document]
peternose: Are you sure?

go/runtime/history/indexer.go
```diff
@@ -0,0 +1,313 @@
...
+			Height:    height,
+		})
+		switch {
+		case err == nil:
+		case errors.Is(err, consensus.ErrVersionNotFound):
+			bi.logger.Debug("failed to get runtime state, probably pruned",
+				"err", err,
+				"height", height,
+			)
+			continue
```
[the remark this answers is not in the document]
peternose: This error cannot be caused by an ephemeral database issue. If this error is thrown, the db works as expected, just the runtime state is not available on that height. If there is a database error, the default case should restart reindex when the next block arrives.

I added retry mechanism to reindex also.

go/runtime/history/indexer.go
peternose: No. The service only terminates if we were unable to commit 1000 blocks in a row, but this also means that something is really wrong and probably not recoverable.

go/runtime/history/indexer.go
```diff
@@ -0,0 +1,313 @@
...
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+		break
+	}
+
+	// Commit the last block manually, if the height at which the block
+	// was published has already been pruned. This ensures that there
+	// is always at least one block in the history.
+	height, err := bi.history.LastConsensusHeight()
```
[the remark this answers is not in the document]
peternose: One test was failing. Will check why hook doesn't work.

go/runtime/history/indexer.go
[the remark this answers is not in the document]
martintomazic: I was wrong. (defer is lifo not fifo ...)

go/runtime/history/indexer.go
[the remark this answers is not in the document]
martintomazic: I see. 

Could you add a comment, that you are doing retry during reindex for a very unlikely scenario where error would happen on the last block (e.g. suspended runtime from here on)? Feel free to ignore/not blocking.

Finally, could you add retry [here](https://github.com/oasisprotocol/oasis-core/pull/6089#discussion_r1991284973) as well. After that happy to approve. :)

go/runtime/history/indexer.go
martintomazic: What if someone changes `batch_size=1`, or if we make it parametric in the future?

peternose: Will introduce a new variable and stop abusing the exiting one.

kostko (approved):

go/runtime/history/indexer.go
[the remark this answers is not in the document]
peternose: We need this not because of the pruning, but because we can sync from a newer height. In this case, the consensus block at which the runtime block was published will not be available, even though, the roothash will still know the last block because it is in the runtime state of the last known block.

So my comment was incorrect and was fixed.

go/runtime/history/indexer.go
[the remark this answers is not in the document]
peternose: > Could you add a comment, that you are doing retry during reindex for a very unlikely scenario where error would happen on the last block (e.g. suspended runtime from here on)?

Not sure what are you saying here. If runtime is suspended, the last block should reindex normally, I don't see a problem here.

> Finally, could you add retry [here](https://github.com/oasisprotocol/oasis-core/pull/6089#discussion_r1991284973) as well.

This is still in progress, will reply in that thread.

go/runtime/history/indexer.go
peternose: I have changed the roothash to be more strict and to not allow any gaps. I think that if this error happens, the application will also panic. For example, I we cannot read from the state, how will the application write to it? Therefore, I'm postponing this until needed.

go/runtime/history/indexer.go
peternose: > When (if) we receive first block, check if there is gap of consensus heights and reindex the gap as well.

This might not be more simple because we don't know how many blocks were produced between the last reindex and subscription to the `WatchBlocks`. It could be none or few. So we are back at the begging, which means, that we can never know if the reindex has be completed. We only know that we came very close with heights. 

To solve this you need to start listening first, and then do the reindex. If new blocks arrive before, throw them away and update the height. If no blocks arrive, the index was completed and any blocks still in the channel will be processed normally without any gaps.

> I am starting to think it could simplify reindex function a lot if we have recursive function (so no need for extra channells, go routine etc)?

This causes the same problem, I think. You cannot know when reindex is completed, you only know that you came close.

> You never have many pending blocks inside WatchBlocks buffer

We don't have any pending blocks. If blocks arrive during reindex, we throw them away, cleaning the buffer. Once reindex is completed, index will start processing these blocks, so there should be no pending at all.

peternose:
> Do we want to add changelog

I didn't add this because nothing has really changed, we just cleaned the code.

> In theory it also parallelizes reindex (should be much faster if many runtimes), since no reindexer is now run in a separate go routine?

You are right, I didn't even think about this. I could mention that runtimes are now reindex in parallel, but won't. It is nice tough that we don't block `DeliverCommand` and `DeliverEvent` anymore.

go/runtime/history/indexer.go
martintomazic: > // For now, enforce strict error handling and address if necessary later.
> I have changed the roothash to be more strict and to not allow any gaps. 

You will  only log the error [see](https://github.com/oasisprotocol/oasis-core/blob/dff158f084531943f3c24d8e79db9f9e60ccc590/go/consensus/cometbft/full/services.go#L201), meaning for all subsequent runtime blocks you will simply continue logging warnings. 

>  For example, I we cannot read from the state, how will the application write to it?

I thought epehemeral db errors are reality when working with databases. E.g. reading once may fail for whatever reason;  ephemeral disk problem, to many transactions, locks, rate limiting etc. However, subsequent write or reads may still pass.

Previously, if this scenario happens you would recover due to reindex (but only in case of ephemeral error). Now you will log warnings, whilst in reality we should gracefully stop or try to recover?

martintomazic (approved):
LGTM!

PR #6093, opened 2025-02-28
go/oasis-node: Fix node termination on service failure 

peternose:


peternose:
Put on hold, as this problem is more trickier than I expected and not high priority.

PR #6095, opened 2025-03-04
 go/roothash: Remove RoundResults from history reindex

martintomazic:
`RoundResults` are now on-chain so there is no need to reindex them anymore. See [thread](https://oasisprotocol.slack.com/archives/C075WCGDNJ0/p1741098222045169) for more context.

Related to #5738, I have confirmed this increases our history reindex ~30%, which is in line with what [existing benchmarks](https://docs.google.com/document/d/1Lq-Qe6jrrLUSKIkLLsIGOrfb31Ivi-hZKP0wAC50V1E/edit?tab=t.0#heading=h.2fkeolp01q4u) predict.

### DIFF
--- a/go/runtime/history/db.go
+++ b/go/runtime/history/db.go
@@ -28,10 +28,14 @@ var (
 	//
 	// Value is CBOR-serialized roothash.AnnotatedBlock.
 	blockKeyFmt = keyFormat.New(0x02, uint64(0))
-	// roundResultsKeyFmt is the round result index key format.
+
+	// deprecatedRoundResultsKeyFmt is the deprecated round result index key format.
 	//
 	// Value is CBOR-serialized roothash.RoundResults.
-	roundResultsKeyFmt = keyFormat.New(0x03, uint64(0))
+	//
+	// Deprecated: This key format is deprecated and will be removed in future versions
+	// after we clean up the remaining round results from the block history (#6098).
+	deprecatedRoundResultsKeyFmt = keyFormat.New(0x03, uint64(0)) //nolint:unused
 )
 
 type dbMetadata struct {
@@ -149,13 +153,7 @@ func (d *DB) metadata() (*dbMetadata, error) {
 	return meta, nil
 }
 
-func (d *DB) commit(blks []*roothash.AnnotatedBlock, results []*roothash.RoundResults) error {
-	if len(blks) != len(results) {
-		return fmt.Errorf("length mismatch (blocks: %d round results: %d)",
-			len(blks),
-			len(results),
-		)
-	}
+func (d *DB) commit(blks []*roothash.AnnotatedBlock) error {
 	if len(blks) == 0 {
 		return nil
 	}
@@ -166,7 +164,7 @@ func (d *DB) commit(blks []*roothash.AnnotatedBlock, results []*roothash.RoundRe
 			return err
 		}
 
-		for i, blk := range blks {
+		for _, blk := range blks {
 			rtID := blk.Block.Header.Namespace
 			if !rtID.Equal(&meta.RuntimeID) {
 				return fmt.Errorf("runtime mismatch (expected: %s got: %s)",
@@ -190,10 +188,6 @@ func (d *DB) commit(blks []*roothash.AnnotatedBlock, results []*roothash.RoundRe
 			}
 
 			if err := tx.Set(blockKeyFmt.Encode(blk.Block.Header.Round), cbor.Marshal(blk)); err != nil {
-				return err
-			}
-
-			if err := tx.Set(roundResultsKeyFmt.Encode(blk.Block.Header.Round), cbor.Marshal(results[i])); err != nil {
 				return err
 			}
 
@@ -247,28 +241,6 @@ func (d *DB) getEarliestBlock() (*roothash.AnnotatedBlock, error) {
 		return nil, txErr
 	}
 	return &blk, nil
-}
-
-func (d *DB) getRoundResults(round uint64) (*roothash.RoundResults, error) {
-	var roundResults *roothash.RoundResults
-	txErr := d.db.View(func(tx *badger.Txn) error {
-		item, err := tx.Get(roundResultsKeyFmt.Encode(round))
-		switch err {
-		case nil:
-		case badger.ErrKeyNotFound:
-			return roothash.ErrNotFound
-		default:
-			return err
-		}
-
-		return item.Value(func(val []byte) error {
-			return cbor.UnmarshalTrusted(val, &roundResults)
-		})
-	})
-	if txErr != nil {
-		return nil, txErr
-	}
-	return roundResults, nil
 }
 
 func (d *DB) close() {
--- a/go/runtime/history/prune.go
+++ b/go/runtime/history/prune.go
@@ -120,15 +120,11 @@ func (p *keepLastPruner) Prune(latestRound uint64) error {
 				break
 			}
 
-			if err := tx.Delete(roundResultsKeyFmt.Encode(round)); err != nil {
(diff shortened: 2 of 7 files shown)


go/runtime/history/prune.go
```diff
@@ -120,18 +120,14 @@ func (p *keepLastPruner) Prune(latestRound uint64) error {
...
 					// We can't prune any more rounds in this transaction.
 					break
 				}
 				return err
 			}
 
-			if err := tx.Delete(item.KeyCopy(nil)); err != nil {
-				return err
-			}
-
```
martintomazic, line 134: Observe that previously we were only checking `badger.ErrTxnTooBig` during round results removal. I would argue this had to be checked during annotated block removal as well?

go/runtime/history/db.go
```diff
@@ -28,10 +28,6 @@ var (
 	//
 	// Value is CBOR-serialized roothash.AnnotatedBlock.
 	blockKeyFmt = keyFormat.New(0x02, uint64(0))
-	// roundResultsKeyFmt is the round result index key format.
-	//
-	// Value is CBOR-serialized roothash.RoundResults.
-	roundResultsKeyFmt = keyFormat.New(0x03, uint64(0))
```
kostko, line 34: Make sure to add a comment saying that `0x03` is reserved to avoid breaking with old history indices.

go/runtime/history/prune.go
peternose: Yes, I think this should be fine. However, I did not check what happens if this error occurs. Is only the last `Delete` ignored or is the whole transaction corrupted, reverted, or something else. Maybe you can check this locally.

go/runtime/history/db.go
```diff
@@ -28,10 +28,9 @@ var (
...
+
+	// Do not use 0x03 prefix. Previously it was used for round results key.
+	// Old reindexes may still have legacy data stored.
```
peternose, line 33: I would rather rename to `deprecatedRoundResultsKeyFmt = keyFormat.New(0x03, uint64(0))`, `reservedKeyFmt = keyFormat.New(0x03, uint64(0))`, or `deprecatedKeyFmt = keyFormat.New(0x03, uint64(0))` and update the comment, because the key format will panic if prefix is used twice, while your comment can easily be overlooked. 

See how some other vars are deprecated.

go/runtime/history/prune.go
```diff
@@ -120,18 +120,14 @@ func (p *keepLastPruner) Prune(latestRound uint64) error {
 				break
 			}
 
-			if err := tx.Delete(roundResultsKeyFmt.Encode(round)); err != nil {
```
peternose, line 123: I would clear any stored round results on startup once, since old round results will never be removed from db and will be kept forever. Furthermore, I would prepare a PR which will remove this cleanup and will be merged in some future version of Oasis Core.

go/runtime/history/db.go
peternose: But if we clear the db we can reuse this, right? So if we add cleanup, we can deprecate now, and remove latter.

peternose:
What's up with this?

https://github.com/oasisprotocol/oasis-core/blob/9fc09c76fb30a7ae20e7b25db54bb67a912154fe/go/worker/compute/executor/committee/prune.go#L21-L22

peternose:
> What's up with this?

Oh, I see. This should be fine. Ignore.

martintomazic:
>What's up with this?
This fetches state directly, but does not rely on reindexed history?

peternose:
> This fetches state directly, but does not rely on reindexed history?

So we can remove this?

peternose:
@kostko Since round results are never used, and we only care about the last round results that are stored in the consensus state, we can probably remove this [pruner](https://github.com/oasisprotocol/oasis-core/blob/9fc09c76fb30a7ae20e7b25db54bb67a912154fe/go/worker/compute/executor/committee/prune.go#L12) and revert this [commit](https://github.com/oasisprotocol/oasis-core/pull/3736/commits/3d998ad839bdea9d5c23b054599c45f166afd20b).

go/runtime/history/db.go
martintomazic: >But if we clear the db we can reuse this, right? So if we add cleanup, we can deprecate now, and remove latter.

Do you mind if I create an issue for this and tackle it after #6089, once abstractions are moved as well?

Probably clean-up should not be blocking init/reindex, nor do we want to do it in the midst of reindex as it would slow it down. So likely best would be to clean after reindex has finished? This will also have to be done in batches to avoid `badger.ErrTxnTooBig`.

On the other hand removing should be fast, way faster then reindex so maybe this is not really a problem.

kostko: Yes, this is not a big deal, we can do cleanup separately.

go/runtime/history/db.go
kostko: Ah yes, this is even better.

kostko:
I think so, unless there are cases where we would also need (other) consensus state from any previous heights.

kostko:
Actually, I just remembered that this prune handler must be kept because some runtimes will do historic queries for things that is not available in last consensus state (e.g. delegation/undelegation events that happened while the runtime was suspended or not producing blocks or similar).

go/runtime/history/db.go
martintomazic: #6098 created to track this.

go/runtime/history/prune.go
martintomazic: If you get `badger.ErrTxnTooBig`  you will only ignore current key and subsequent keys, but commit existing ones. In case of a different error you will revert whole transaction.

go/worker/compute/executor/committee/prune.go
```diff
@@ -18,8 +18,8 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
 	height := p.commonNode.CurrentBlockHeight
 	p.commonNode.CrossNode.Unlock()
 
-	// Make sure we never prune past the last successful round as we need that round in history so
-	// we can fetch any needed round results.
+	// Make sure we never prune past the last normal round, as some runtimes will do historic queries
+	// for things that are not available in last consensus state.
```
peternose, line 22:
```suggestion
	// for things that are not available in the last consensus state (e.g. delegation/undelegation
	// events that happened while the runtime was suspended or not producing blocks).
```
I would keep the example, is informative.

go/runtime/history/db.go
```diff
@@ -28,10 +28,10 @@ var (
 	//
 	// Value is CBOR-serialized roothash.AnnotatedBlock.
 	blockKeyFmt = keyFormat.New(0x02, uint64(0))
-	// roundResultsKeyFmt is the round result index key format.
-	//
-	// Value is CBOR-serialized roothash.RoundResults.
-	roundResultsKeyFmt = keyFormat.New(0x03, uint64(0))
+
+	// deprecatedRoundResultsKeyFmt is deprecated. It may be removed once
+	// clean-up is done, since history may still have legacy data stored.
```
peternose, line 33:
```suggestion
	// deprecatedRoundResultsKeyFmt is the deprecated round result index key format.
	//
	// Value is CBOR-serialized roothash.RoundResults.
	//
	// Deprecated: This key format is deprecated and will be removed in future versions
	// after we clean up the remaining round results from the block history (#6098). 
```
Whoever will be reading this will have no idea what the cleanup is neither about which history you are talking about. Be more general or more exact.

```suggestion
	// deprecatedRoundResultsKeyFmt is the deprecated round result index key format.
	//
	// Value is CBOR-serialized roothash.RoundResults.
	//
	// Deprecated: This key format is deprecated and will be removed in future versions.
```

peternose (approved):

PR #6114, opened 2025-03-18
Improve e2e history reindex test

### LINKED ISSUE #6070: Improve e2e history reindex test
>I think the simplest way would be running a [TestClientScenario](https://github.com/oasisprotocol/oasis-core/blob/f94b700b8edcd3e4fa228e748ca45d7c91a3bf06/go/oasis-test-runner/scenario/e2e/runtime/test_client_scenario.go#L121) which queries (runs GetKeyValueTx) queries all the old blocks, at the end of the existing history reindex e2e test.

For context [see](https://github.com/oasisprotocol/oasis-core/pull/6050#discussion_r1961610126).

martintomazic:
Closes #6070

### DIFF
--- a/go/oasis-test-runner/oasis/client.go
+++ b/go/oasis-test-runner/oasis/client.go
@@ -22,6 +22,8 @@ type Client struct {
 	runtimeProvisioner runtimeConfig.RuntimeProvisioner
 	runtimeConfig      map[int]map[string]interface{}
 
+	batchSize uint16
+
 	consensusPort uint16
 	p2pPort       uint16
 }
@@ -33,6 +35,7 @@ type ClientCfg struct {
 	Runtimes           []int
 	RuntimeProvisioner runtimeConfig.RuntimeProvisioner
 	RuntimeConfig      map[int]map[string]interface{}
+	BatchSize          uint16
 }
 
 // UpdateRuntimes updates the client node runtimes.
@@ -66,6 +69,7 @@ func (client *Client) ModifyConfig() error {
 	if len(client.runtimes) > 0 {
 		client.Config.Mode = config.ModeClient
 		client.Config.Runtime.Provisioner = client.runtimeProvisioner
+		client.Config.Runtime.Indexer.BatchSize = client.batchSize
 	}
 
 	client.AddSeedNodesToConfig()
@@ -96,6 +100,7 @@ func (net *Network) NewClient(cfg *ClientCfg) (*Client, error) {
 		runtimes:           cfg.Runtimes,
 		runtimeProvisioner: cfg.RuntimeProvisioner,
 		runtimeConfig:      cfg.RuntimeConfig,
+		batchSize:          cfg.BatchSize,
 		consensusPort:      host.getProvisionedPort(nodePortConsensus),
 		p2pPort:            host.getProvisionedPort(nodePortP2P),
 	}
--- a/go/oasis-test-runner/oasis/client.go
+++ b/go/oasis-test-runner/oasis/client.go
@@ -35,6 +35,11 @@ type ClientCfg struct {
 	RuntimeConfig      map[int]map[string]interface{}
 }
 
+// UpdateRuntimes updates the client node runtimes.
+func (client *Client) UpdateRuntimes(runtimes []int) {
+	client.runtimes = runtimes
+}
+
 func (client *Client) AddArgs(args *argBuilder) error {
 	args.appendNetwork(client.net)
 
--- a/go/oasis-test-runner/oasis/fixture.go
+++ b/go/oasis-test-runner/oasis/fixture.go
@@ -518,6 +518,9 @@ type ClientFixture struct {
 
 	// RuntimeConfig contains the per-runtime node-local configuration.
 	RuntimeConfig map[int]map[string]interface{} `json:"runtime_config,omitempty"`
+
+	// BatchSize is max number of block that block indexer writes at the same time.
+	BatchSize uint16 `json:"batch_size"`
 }
 
 // Create instantiates the client node described by the fixture.
@@ -536,6 +539,7 @@ func (f *ClientFixture) Create(net *Network) (*Client, error) {
 		Runtimes:           f.Runtimes,
 		RuntimeProvisioner: f.RuntimeProvisioner,
 		RuntimeConfig:      f.RuntimeConfig,
+		BatchSize:          f.BatchSize,
 	})
 }
 
--- a/go/oasis-test-runner/scenario/e2e/runtime/helpers_runtime.go
+++ b/go/oasis-test-runner/scenario/e2e/runtime/helpers_runtime.go
@@ -613,30 +613,11 @@ func (sc *Scenario) ensureReplicationWorked(ctx context.Context, km *oasis.Keyma
 	return fmt.Errorf("consensus state missing km status")
 }
 
-func (sc *Scenario) WaitNextRuntimeBlock(ch <-chan *roothash.AnnotatedBlock) (*roothash.AnnotatedBlock, error) {
-	sc.Logger.Info("waiting for next runtime block")
-
-	timer := time.NewTimer(10 * time.Second)
-
-	for {
-		select {
-		case blk, ok := <-ch:
-			if !ok {
-				return nil, fmt.Errorf("runtime block channel closed")
-			}
-			return blk, nil
-		case <-timer.C:
-			return nil, fmt.Errorf("failed to receive runtime block")
-		}
-	}
-}
-
-func (sc *Scenario) WaitRuntimeBlock(ch <-chan *roothash.AnnotatedBlock, round uint64) (*roothash.AnnotatedBlock, error) {
+// WaitRuntimeBlock waits until the runtime block for the specified round is generated.
+func (sc *Scenario) WaitRuntimeBlock(ctx context.Context, ch <-chan *roothash.AnnotatedBlock, round uint64) (*roothash.AnnotatedBlock, error) {
 	sc.Logger.Info("waiting for runtime block",
 		"round", round,
 	)
-
-	timer := time.NewTimer(10 * time.Second)
 
 	for {
 		select {
@@ -655,8 +636,10 @@ func (sc *Scenario) WaitRuntimeBlock(ch <-chan *roothash.AnnotatedBlock, round u
 				continue
 			}
 			return blk, nil
-		case <-timer.C:
-			return nil, fmt.Errorf("failed to receive runtime block")
+		case <-time.After(time.Minute):
+			return nil, fmt.Errorf("runtime block timeout")
+		case <-ctx.Done():
+			return nil, ctx.Err()
 		}
 	}
 }
--- a/go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
+++ b/go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
@@ -55,6 +55,7 @@ func (sc *historyReindexImpl) Fixture() (*oasis.NetworkFixture, error) {
 
 	// Start client without any runtime.
 	f.Clients[0].Runtimes = nil
+	f.Clients[0].BatchSize = 3
 
 	return f, nil
 }
--- a/go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
+++ b/go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
@@ -3,22 +3,23 @@ package runtime
 import (
 	"context"
 	"fmt"
+	"slices"
 
-	consensus "github.com/oasisprotocol/oasis-core/go/consensus/api"
 	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/env"
-	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/log"
 	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis"
-	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis/cli"
 	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario"
 	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
+	"github.com/oasisprotocol/oasis-core/go/roothash/api/block"
+	"github.com/oasisprotocol/oasis-core/go/runtime/client/api"
 	scheduler "github.com/oasisprotocol/oasis-core/go/scheduler/api"
 )
 
-// HistoryReindex is the scenario that triggers roothash history reindexing.
+// HistoryReindex is the scenario that triggers block history reindexing.
 var HistoryReindex scenario.Scenario = newHistoryReindexImpl()
 
 type historyReindexImpl struct {
 	Scenario
+	rtIdx int
 }
 
 func newHistoryReindexImpl() scenario.Scenario {
@@ -36,32 +37,24 @@ func (sc *historyReindexImpl) Fixture() (*oasis.NetworkFixture, error) {
 		return nil, err
 	}
 
-	f.ComputeWorkers = []oasis.ComputeWorkerFixture{
-		{
-			Entity:   1,
-			Runtimes: []int{},
-			LogWatcherHandlerFactories: []log.WatcherHandlerFactory{
-				// Ensure re-indexing happens on the node.
-				oasis.LogAssertRoothashRoothashReindexing(),
-			},
-		},
+	// Find the first compute runtime.
+	rtIdx := slices.IndexFunc(f.Runtimes, func(rt oasis.RuntimeFixture) bool {
+		return rt.Kind == registry.KindCompute
+	})
+	if rtIdx == -1 {
+		return nil, fmt.Errorf("no compute runtime configured")
 	}
+	sc.rtIdx = rtIdx
 
-	// Assumes a single compute runtime.
-	var rtIdx int
-	for idx, rt := range f.Runtimes {
-		if rt.Kind == registry.KindCompute {
-			rtIdx = idx
-			break
-		}
-	}
-	// Compute runtime will be registered later.
-	f.Runtimes[rtIdx].ExcludeFromGenesis = true
-	// Use a single compute node.
+	// Run selected runtime on a single compute node.
+	f.ComputeWorkers = f.ComputeWorkers[:1]
 	f.Runtimes[rtIdx].Executor.GroupSize = 1
 	f.Runtimes[rtIdx].Executor.GroupBackupSize = 0
 	f.Runtimes[rtIdx].Constraints[scheduler.KindComputeExecutor][scheduler.RoleWorker].MinPoolSize.Limit = 1
 	f.Runtimes[rtIdx].Constraints[scheduler.KindComputeExecutor][scheduler.RoleBackupWorker].MinPoolSize.Limit = 0
+
+	// Start client without any runtime.
+	f.Clients[0].Runtimes = nil
 
 	return f, nil
 }
@@ -73,63 +66,89 @@ func (sc *historyReindexImpl) Clone() scenario.Scenario {
 }
 
 func (sc *historyReindexImpl) Run(ctx context.Context, childEnv *env.Env) error {
-	cli := cli.New(childEnv, sc.Net, sc.Logger)
-
 	// Start the network.
 	if err := sc.Net.Start(); err != nil {
 		return err
 	}
 
-	// Wait for one block so that initial epoch can be fetched below.
-	if _, err := sc.WaitBlocks(ctx, 1); err != nil {
-		return fmt.Errorf("failed to wait for one block: %w", err)
-	}
-
-	// Restart compute worker with configured runtime.
+	// Prepare few runtime blocks for reindex.
 	compute := sc.Net.ComputeWorkers()[0]
-	sc.Logger.Info("stopping the compute worker")
-	if err := compute.Stop(); err != nil {
-		return err
-	}
-	var rtIdx int
-	for idx, rt := range sc.Net.Runtimes() {
-		if rt.Kind() == registry.KindCompute {
-			rtIdx = idx
-			break
-		}
-	}
-	// Update worker runtime configuration.
-	compute.UpdateRuntimes([]int{rtIdx})
-	sc.Logger.Info("starting the compute worker")
-	if err := compute.Start(); err != nil {
-		return err
-	}
-
-	// Fetch current epoch.
-	epoch, err := sc.Net.Controller().Beacon.GetEpoch(ctx, consensus.HeightLatest)
-	if err != nil {
-		return fmt.Errorf("failed to get current epoch: %w", err)
-	}
-
-	// Register runtime.
-	compRt := sc.Net.Runtimes()[rtIdx]
-	rtDsc := compRt.ToRuntimeDescriptor()
-	rtDsc.Deployments[0].ValidFrom = epoch + 1
-	if err = sc.RegisterRuntime(childEnv, cli, rtDsc, 0); err != nil {
-		return err
-	}
-
-	// Wait for the compute worker to be ready.
-	sc.Logger.Info("waiting for the compute worker to become ready")
 	computeCtrl, err := oasis.NewController(compute.SocketPath())
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create controller: %w", err)
 	}
-	if err = computeCtrl.WaitReady(ctx); err != nil {
-		return err
+	err = sc.waitForClientRuntimeBlock(ctx, computeCtrl.RuntimeClient, 10)
+	if err != nil {
+		return fmt.Errorf("failed to wait for runtime block: %w", err)
 	}
 
-	// Run client to ensure runtime works.
-	sc.Logger.Info("Starting the basic client")
-	return sc.RunTestClientAndCheckLogs(ctx, childEnv)
+	// Reindex existing runtime blocks and start indexing new ones.
+	client := sc.Net.Clients()[0]
+	client.UpdateRuntimes([]int{sc.rtIdx})
+	if err = client.Restart(ctx); err != nil {
+		return fmt.Errorf("failed to restart %s: %w", client.Name, err)
+	}
+
+	// Verify that indexing works.
+	if err := sc.waitForClientRuntimeBlock(ctx, sc.Net.ClientController().RuntimeClient, 20); err != nil {
+		return fmt.Errorf("failed to wait for runtime block: %w", err)
+	}
+
+	// Verify (re)indexed runtime blocks.
+	for round := uint64(0); round <= 20; round++ {
+		if err := sc.ensureEqualBlock(ctx, computeCtrl.RuntimeClient, sc.Net.ClientController().RuntimeClient, round); err != nil {
+			return fmt.Errorf("failed to ensure %s and %s equal blocks (round: %d): %w", compute.Name, client.Name, round, err)
+		}
+	}
+
+	// Run test client to ensure runtime works.
+	sc.Logger.Info("starting the basic test client")
+	if err = sc.RunTestClientAndCheckLogs(ctx, childEnv); err != nil {
+		return fmt.Errorf("failed to run test client and check logs: %w", err)
+	}
+
+	return nil
+}
+
+func (sc *historyReindexImpl) waitForClientRuntimeBlock(ctx context.Context, client api.RuntimeClient, round uint64) error {
+	rtID := sc.Net.Runtimes()[sc.rtIdx].ID()
+	ch, sub, err := client.WatchBlocks(ctx, rtID)
+	if err != nil {
+		return fmt.Errorf("failed to watch runtime blocks: %w)", err)
+	}
+	defer sub.Close()
+	if _, err := sc.WaitRuntimeBlock(ctx, ch, round); err != nil {
+		return fmt.Errorf("failed to wait for runtime round %d: %w", round, err)
+	}
+	return nil
+}
+
+func (sc *historyReindexImpl) ensureEqualBlock(ctx context.Context, client1, client2 api.RuntimeClient, round uint64) error {
+	blk1, err := sc.fetchRuntimeBlock(ctx, client1, round)
+	if err != nil {
+		return fmt.Errorf("failed to fetch client1's runtime block: %w", err)
+	}
+	blk2, err := sc.fetchRuntimeBlock(ctx, client2, round)
+	if err != nil {
+		return fmt.Errorf("failed to fetch client2's runtime block: %w", err)
+	}
+
+	hash1 := blk1.Header.EncodedHash()
+	hash2 := blk2.Header.EncodedHash()
+	if !hash1.Equal(&hash2) {
+		return fmt.Errorf("block header hash not equal: want %s, got %s", hash1, hash2)
+	}
+
+	return nil
+}
+
+func (sc *historyReindexImpl) fetchRuntimeBlock(ctx context.Context, client api.RuntimeClient, round uint64) (*block.Block, error) {
+	blk, err := client.GetBlock(ctx, &api.GetBlockRequest{
+		RuntimeID: KeyValueRuntimeID,
+		Round:     round,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get block: %w", err)
+	}
+	return blk, nil
 }
--- a/go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
+++ b/go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
@@ -14,10 +14,6 @@ import (
 	scheduler "github.com/oasisprotocol/oasis-core/go/scheduler/api"
 )
 
-const (
-	reindexComputePruneNumKept = 50
-)
-
 // HistoryReindex is the scenario that triggers roothash history reindexing.
 var HistoryReindex scenario.Scenario = newHistoryReindexImpl()
 
@@ -44,12 +40,7 @@ func (sc *historyReindexImpl) Fixture() (*oasis.NetworkFixture, error) {
 		{
 			Entity:   1,
 			Runtimes: []int{},
-			Consensus: oasis.ConsensusFixture{
-				PruneNumKept: reindexComputePruneNumKept,
-			},
 			LogWatcherHandlerFactories: []log.WatcherHandlerFactory{
-				// Ensure ABCI pruning happens on the node.
-				oasis.LogEventABCIPruneDelete(),
 				// Ensure re-indexing happens on the node.
 				oasis.LogAssertRoothashRoothashReindexing(),
 			},
@@ -89,27 +80,9 @@ func (sc *historyReindexImpl) Run(ctx context.Context, childEnv *env.Env) error 
 		return err
 	}
 
-	// Wait for enough block to ensure pruning on the compute node.
-	waitForHeight := int64(reindexComputePruneNumKept + 20)
-	sc.Logger.Info("waiting enough blocks to ensure pruning",
-		"compute_prune_num_kept", reindexComputePruneNumKept,
-		"wait_for_height", waitForHeight,
-	)
-
-	blockCh, blockSub, bErr := sc.Net.Controller().Consensus.WatchBlocks(ctx)
-	if bErr != nil {
-		return fmt.Errorf("failed waiting for block height: %w", bErr)
-	}
-	defer blockSub.Close()
-
-	for newBlk := range blockCh {
-		if newBlk.Height > waitForHeight {
-			break
-		}
-		sc.Logger.Debug("waiting enough blocks to ensure pruning",
-			"current_height", newBlk.Height,
-			"wait_for_height", waitForHeight,
-		)
+	// Wait for one block so that initial epoch can be fetched below.
+	if _, err := sc.WaitBlocks(ctx, 1); err != nil {
+		return fmt.Errorf("failed to wait for one block: %w", err)
 	}
 
 	// Restart compute worker with configured runtime.
--- a/go/oasis-test-runner/scenario/e2e/scenario.go
+++ b/go/oasis-test-runner/scenario/e2e/scenario.go
@@ -143,6 +143,8 @@ func RegisterScenarios() error {
 		Debond,
 		// Consensus state sync.
 		ConsensusStateSync,
+		// Consensus pruning test.
+		ConsensusPrune,
 		// Multiple seeds test.
 		MultipleSeeds,
 		// Seed API test.
--- a/go/runtime/config/config.go
+++ b/go/runtime/config/config.go
@@ -110,6 +110,9 @@ type Config struct {
 	// History pruner configuration.
 	Prune PruneConfig `yaml:"prune,omitempty"`
 
+	// Indexer is history indexer configuration.
+	Indexer IndexerConfig `yaml:"indexer,omitempty"`
+
 	// RuntimeConfig maps runtime IDs to their respective local configurations.
 	// NOTE: This may go away in the future, use `RuntimeConfig.Config` instead.
 	RuntimeConfig map[string]map[string]interface{} `yaml:"config,omitempty"`
@@ -301,10 +304,19 @@ type PruneConfig struct {
 	NumKept uint64 `yaml:"num_kept"`
 }
 
+// IndexerConfig is history indexer configuration.
+type IndexerConfig struct {
+	// BatchSize is max number of blocks committed in a batch during history reindex.
+	//
+	// Setting it to zero uses the default batch size.
+	BatchSize uint16 `yaml:"batch_size,omitempty"`
+}
+
 // LoadBalancerConfig is the load balancer configuration.
 type LoadBalancerConfig struct {
-	// NumInstances is the number of runtime instances to provision for load-balancing. Setting it
-	// to zero (default) or one disables load balancing.
+	// NumInstances is the number of runtime instances to provision for load-balancing.
+	//
+	// Setting it to zero or one disables load balancing.
 	NumInstances uint64 `yaml:"num_instances,omitempty"`
 }
 
@@ -427,6 +439,9 @@ func DefaultConfig() Config {
 			Strategy: "none",
 			Interval: 2 * time.Minute,
 			NumKept:  600,
+		},
+		Indexer: IndexerConfig{
+			BatchSize: 1000,
 		},
 		SentryAddresses: []string{},
 		TxPool: tpConfig.Config{
--- a/go/runtime/history/indexer.go
+++ b/go/runtime/history/indexer.go
@@ -16,7 +16,6 @@ import (
 )
 
 const (
-	batchSize        = 1000
 	maxPendingBlocks = 10
 )
 
@@ -27,18 +26,20 @@ type BlockIndexer struct {
 
 	consensus consensus.Backend
 	history   History
+	batchSize uint16
 
 	logger *logging.Logger
 }
 
 // NewBlockIndexer creates a new block indexer.
-func NewBlockIndexer(consensus consensus.Backend, history History) *BlockIndexer {
+func NewBlockIndexer(consensus consensus.Backend, history History, batchSize uint16) *BlockIndexer {
 	logger := logging.GetLogger("runtime/history/indexer").With("runtime_id", history.RuntimeID())
 
 	return &BlockIndexer{
 		startOne:  cmSync.NewOne(),
 		consensus: consensus,
 		history:   history,
+		batchSize: batchSize,
 		logger:    logger,
 	}
 }
@@ -230,6 +231,7 @@ func (bi *BlockIndexer) reindexTo(ctx context.Context, height int64) error {
 		lastHeight = lastRetainedHeight
 	}
 
+	batchSize := int64(bi.batchSize)
 	for start := lastHeight; start <= height; start += batchSize {
 		end := min(start+batchSize-1, height)
 		if err = bi.reindexRange(ctx, start, end); err != nil {
--- a/go/runtime/registry/registry.go
+++ b/go/runtime/registry/registry.go
@@ -538,7 +538,7 @@ func (r *runtimeRegistry) createRuntime(runtimeID common.Namespace, managed bool
 		r.consensus.Pruner().RegisterHandler(rt.history)
 
 		// Start indexing blocks.
-		indexer := history.NewBlockIndexer(r.consensus, rt.history)
+		indexer := history.NewBlockIndexer(r.consensus, rt.history, config.GlobalConfig.Runtime.Indexer.BatchSize)
 		r.indexers = append(r.indexers, indexer)
 	}
 
--- a/.changelog/6070.cfg.md
+++ b/.changelog/6070.cfg.md
@@ -1,0 +1,3 @@
+go/runtime/history: Make history indexer batch size configurable
+
+Batch size can now be specified via `runtime.indexer.batch_size`.
--- a/go/oasis-test-runner/scenario/e2e/consensus_prune.go
+++ b/go/oasis-test-runner/scenario/e2e/consensus_prune.go
@@ -1,0 +1,89 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/env"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/log"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario"
+)
+
+const (
+	pruneNumKept = 20
+)
+
+// ConsensusPrune is scenario that tests consensus block pruning.
+var ConsensusPrune scenario.Scenario = &consensusPruneImpl{
+	Scenario: *NewScenario("consensus-prune"),
+}
+
+type consensusPruneImpl struct {
+	Scenario
+}
+
+func (sc *consensusPruneImpl) Fixture() (*oasis.NetworkFixture, error) {
+	f, err := sc.Scenario.Fixture()
+	if err != nil {
+		return nil, err
+	}
+
+	f.Validators[0].Consensus.PruneNumKept = pruneNumKept
+	f.Validators[0].LogWatcherHandlerFactories = []log.WatcherHandlerFactory{
+		// Ensure ABCI pruning happens on the node.
+		oasis.LogEventABCIPruneDelete(),
+	}
+
+	return f, nil
+}
+
+func (sc *consensusPruneImpl) Clone() scenario.Scenario {
+	return &consensusPruneImpl{
+		Scenario: *sc.Scenario.Clone().(*Scenario),
+	}
+}
+
+func (sc *consensusPruneImpl) Run(ctx context.Context, _ *env.Env) error {
+	if err := sc.Net.Start(); err != nil {
+		return fmt.Errorf("failed to start network: %w", err)
+	}
+
+	// Wait first few blocks and ensure no pruning.
+	halfPrune := pruneNumKept / 2
+	if _, err := sc.WaitBlocks(ctx, halfPrune); err != nil {
+		return fmt.Errorf("failed waiting %d blocks: %w", halfPrune, err)
+	}
+
+	sc.Logger.Info("ensure no pruning",
+		"start_height", 1,
+		"end_height", halfPrune,
+	)
+	ctrl := sc.Net.Controller()
+	for h := 1; h <= halfPrune; h++ {
+		if _, err := ctrl.Consensus.GetBlock(ctx, int64(h)); err != nil {
+			return fmt.Errorf("pruning too early: %w", err)
+		}
+	}
+
+	// Wait for additional blocks and verify pruning.
+	if _, err := sc.WaitBlocks(ctx, pruneNumKept); err != nil {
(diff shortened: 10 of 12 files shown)


go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -82,14 +85,116 @@ func (sc *historyReindexImpl) Clone() scenario.Scenario {
...
+func (sc *historyReindexImpl) ensureEqualBlock(
+	ctx context.Context,
+	client api.RuntimeClient,
+	round uint64,
+	want *block.Block,
+) error {
+	got, err := client.GetBlock(ctx, &api.GetBlockRequest{
+		RuntimeID: KeyValueRuntimeID,
+		Round:     round,
+	})
```
martintomazic, line 185: Not best, since history reindex is also about annotated block, which `api.RuntimeClient` does not expose...

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -82,14 +85,116 @@ func (sc *historyReindexImpl) Clone() scenario.Scenario {
...
+	}
+
+	// Start new clients, all with different history indexer batch size configuration.
+	// For every client, ensure reindex works by comparing all blocks of default client.
+	for _, batchSize := range []uint16{1, 2, 3, 1000} {
+		if err := sc.startNewClientAndTestReindex(ctx, lastRound, batchSize); err != nil {
+			return fmt.Errorf("wrong history reindex (batchSize: %d): %w", batchSize, err)
+		}
+	}
+
```
martintomazic, line 121: `sc.RunTestClientAndCheckLogs`  produces 4 runtime blocks (`InsertRemoveEncWithSecretsScenario`), where every runtime block is produced every second consensus block.

In theory this is enough to find some corner cases with given batch sizes, but leaves much to be desired. I would prefer to have PBT based style test (generator that randomizes block gaps, random batch sizes etc...). Unfortunately to have this functionality we would have to change quite a lot...

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -82,14 +85,116 @@ func (sc *historyReindexImpl) Clone() scenario.Scenario {
...
 func (sc *historyReindexImpl) Run(ctx context.Context, childEnv *env.Env) error {
-	cli := cli.New(childEnv, sc.Net, sc.Logger)
-
 	// Start the network.
 	if err := sc.Net.Start(); err != nil {
 		return err
 	}
 
 	// Wait for enough block to ensure pruning on the compute node.
+	if err := sc.ensureComputePruning(ctx); err != nil {
```
peternose, line 94: We have method `sc.WaitBlocks(ctx, n)` which would be replace this method.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -4,13 +4,16 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/google/go-cmp/cmp"
```
peternose, line 7: Linter doesn't complain that there is no new line afterwards?

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -82,14 +85,116 @@ func (sc *historyReindexImpl) Clone() scenario.Scenario {
...
+	for _, batchSize := range []uint16{1, 2, 3, 1000} {
+		if err := sc.startNewClientAndTestReindex(ctx, lastRound, batchSize); err != nil {
+			return fmt.Errorf("wrong history reindex (batchSize: %d): %w", batchSize, err)
+		}
+	}
+
+	return nil
+}
+
+func (sc *historyReindexImpl) startNewClientAndTestReindex(
```
peternose, line 125: I would expect this method to be lower, as it is used at the end of the `run`. So if `run` calls methods `A`, `B`, `C` in that order, I would also expect that method `B` is defined below `A` and `C` below `B`. Unless these are publlic methods, which should usually be always on the top.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -156,7 +266,75 @@ func (sc *historyReindexImpl) Run(ctx context.Context, childEnv *env.Env) error
...
+func (sc *historyReindexImpl) fetchLatestRound(ctx context.Context) (uint64, error) {
+	latestBlock, err := sc.Net.ClientController().RuntimeClient.GetBlock(ctx, &api.GetBlockRequest{
+		RuntimeID: KeyValueRuntimeID,
+		Round:     api.RoundLatest,
+	})
+	if err != nil {
+		return 0, fmt.Errorf("failed to get latest block: %w", err)
+	}
+	round := latestBlock.Header.Round
+	sc.Logger.Info("latest block", "round", round)
```
peternose, line 338:
```suggestion
```

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -82,14 +85,116 @@ func (sc *historyReindexImpl) Clone() scenario.Scenario {
...
+
+	// Run client to ensure runtime works.
+	sc.Logger.Info("Starting the basic client")
+	if err := sc.RunTestClientAndCheckLogs(ctx, childEnv); err != nil {
+		return fmt.Errorf("failed to run test client and check logs: %w", err)
+	}
+
+	lastRound, err := sc.fetchLatestRound(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to fetch default client latest round: %w", err)
```
peternose, line 111:
```suggestion
		return fmt.Errorf("failed to fetch latest round: %w", err)
```

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -156,7 +266,75 @@ func (sc *historyReindexImpl) Run(ctx context.Context, childEnv *env.Env) error
 		return err
 	}
 
-	// Run client to ensure runtime works.
-	sc.Logger.Info("Starting the basic client")
-	return sc.RunTestClientAndCheckLogs(ctx, childEnv)
+	return nil
+}
+
+func (sc *historyReindexImpl) starNewClient(batchSize uint16) (*oasis.Client, error) {
```
peternose, line 272: Client node or test client?

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -82,14 +85,116 @@ func (sc *historyReindexImpl) Clone() scenario.Scenario {
...
+		return fmt.Errorf("failed to ensure equal block history: %w", err)
+	}
+
+	return nil
+}
+
+func (sc *historyReindexImpl) ensureEqualHistory(
+	ctx context.Context,
+	client api.RuntimeClient,
+	start, end uint64,
```
peternose, line 153: One can easily overlook `end`.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -82,14 +85,116 @@ func (sc *historyReindexImpl) Clone() scenario.Scenario {
...
+	}
+
+	if err = sc.ensureEqualHistory(ctx, clientCtrl.RuntimeClient, 0, lastRound); err != nil {
+		return fmt.Errorf("failed to ensure equal block history: %w", err)
+	}
+
+	return nil
+}
+
+func (sc *historyReindexImpl) ensureEqualHistory(
```
peternose, line 150: What do we gain with this test? So both clients are running the same code, why wouldn't they have equal blocks?

go/oasis-test-runner/oasis/client.go
```diff
@@ -63,6 +66,10 @@ func (client *Client) ModifyConfig() error {
 		client.Config.Runtime.Provisioner = client.runtimeProvisioner
 	}
 
+	if client.batchSize > 0 {
```
peternose, line 69: To avoid this you can always allow 0 and use a default value if set to 0. Or be strict as you are and not allow 0 at all. Both options are fine.

go/oasis-test-runner/oasis/client.go
```diff
@@ -63,6 +66,10 @@ func (client *Client) ModifyConfig() error {
...
+		client.SetIndexerBatchSize(client.batchSize)
```
peternose, line 70: Why don't you change the config here directly?

go/runtime/history/indexer.go
```diff
@@ -230,6 +230,7 @@ func (bi *BlockIndexer) reindexTo(ctx context.Context, height int64) error {
 		lastHeight = lastRetainedHeight
 	}
 
+	batchSize := int64(config.GlobalConfig.Runtime.Indexer.BatchSize)
```
peternose, line 233: We should avoid reading from the global state if possible. Therefore, you should accept this value in the constructor, either as required or optional parameter.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
peternose: I think you can create your own test scenario with timeouts. This would give you wanted block gaps.

go/go.mod
```diff
@@ -29,6 +29,7 @@ require (
 	github.com/golang/protobuf v1.5.4
 	github.com/golang/snappy v0.0.4
 	github.com/google/btree v1.1.3
+	github.com/google/go-cmp v0.6.0
```
martintomazic, line 32: `oasis-test-runner/scenario/e2e/runtime/history_reindex.go:7:2: import 'github.com/google/go-cmp/cmp' is not allowed from list 'main' (depguard)`

Can I whitelist this dependency, at least for testing? I think is quite standard to use in go codebase, but I also understand if don't want to increase dependency surface... :)

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
martintomazic: >// Property: for every batch size, reindexing blocks 1...N, should
	// produce equal block history as indexing.

By varying batch sizes we simply increase number of test cases/coverage. It is a simplified PBT style test.

 I think is is valuable as until now we didn't have any test at all checking how reindexer works.
 E.g. if you change interval in the for loop or increment wrongly, you will finally have smtg that is failing... (coverage is of course limited)

.changelog/6070.trivial.md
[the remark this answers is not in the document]
martintomazic: Added a changelog. I understand your perspective, but it also gives us ability to fine tune this later, without upgrading oasis version... ?

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -82,14 +85,116 @@ func (sc *historyReindexImpl) Clone() scenario.Scenario {
...
+	clientCtrl, err := oasis.NewController(client.SocketPath())
+	if err != nil {
+		return fmt.Errorf("failed to create client controller: %w", err)
+	}
+
+	if err := sc.waitRound(ctx, lastRound, clientCtrl.RuntimeClient); err != nil {
+		return fmt.Errorf("failed to wait for client to catch round %d: %w", lastRound, err)
+	}
+
+	if err = sc.ensureEqualHistory(ctx, clientCtrl.RuntimeClient, 0, lastRound); err != nil {
```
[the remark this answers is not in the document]
martintomazic: I have kept somehow similar style, as I think is much easier to read what the test is doing (and simply mentally ignore irrelevant details). Not a fan of 3 line functions, but neither of 50+. Happy to change!

go/go.mod
peternose: Less dependency is always better. Can't you compare blocks otherwise, e.g. using `block.Header.EncodedHash()`?

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
peternose: > By varying batch sizes we simply increase number of test cases/coverage. It is a simplified PBT style test.

It doesn't feel like this should be tested in E2E test. Maybe this would be fine for a unit test, but E2E tests are expensive and testing batch sizes might be just too much.

> I think is is valuable as until now we didn't have any test at all checking how reindexer works.

Maybe you only need to test that the state is the same, i.e. that blocks from the client and one compute worker are the same.

go/runtime/config/config.go
```diff
@@ -279,10 +282,19 @@ type PruneConfig struct {
 	NumKept uint64 `yaml:"num_kept"`
 }
 
+// IndexerConfig is history indexer configuration.
+type IndexerConfig struct {
+	// BatchSize is max number of blocks committed in a batch during history reindex.
+	//
+	// Setting it to zero (default), sets default batch size.
```
peternose, line 289:
```suggestion
	// Setting it to zero uses the default batch size.
```
Confusing, as one might think that zero is default.
Good APIs would probably also expose the default value, so that users know what the default value is.

go/oasis-test-runner/scenario/e2e/runtime/scenario.go
```diff
@@ -143,6 +143,9 @@ func (sc *Scenario) Fixture() (*oasis.NetworkFixture, error) {
 	if err = runtimeProvisioner.UnmarshalText([]byte(runtimeProvisionerRaw)); err != nil {
 		return nil, fmt.Errorf("failed to parse runtime provisioner: %w", err)
 	}
+	if err != nil {
```
peternose, line 146: Isn't this checked above?

go/runtime/registry/registry.go
```diff
@@ -538,7 +538,7 @@ func (r *runtimeRegistry) createRuntime(runtimeID common.Namespace, managed bool
 		r.consensus.Pruner().RegisterHandler(rt.history)
 
 		// Start indexing blocks.
-		indexer := history.NewBlockIndexer(r.consensus, rt.history)
+		indexer := history.NewBlockIndexer(r.consensus, rt.history, config.GlobalConfig.Runtime.Indexer.BatchSize)
```
peternose, line 541: Could also move it higher and read it in the constructor of the registry.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -17,6 +17,8 @@ import (
 
 const reindexComputePruneNumKept = 50
 
+var batchSizes = []uint16{1, 2, 3, 100}
```
peternose, line 20: I still think this makes no sense. I would reindex with one client and compare to a compute.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -129,10 +134,13 @@ func (sc *historyReindexImpl) Run(ctx context.Context, childEnv *env.Env) error
...
-	// Configure second client node and test the following property:
-	// Reindexing blocks 1...N, should produce equal block history as indexing.
-	if err := sc.testReindex(ctx, sc.Net.Clients()[1], rtIdx, last.Header.Round); err != nil {
-		return fmt.Errorf("reindex test failed: %w", err)
+	// Configure remaining client nodes and test their reindex.
+	// Property: for every batch size, reindexing blocks 1...N, should
+	// produce equal block history as indexing.
+	for _, client := range sc.Net.Clients()[1:] {
+		if err := sc.testReindex(ctx, client, rtIdx, last.Header.Round); err != nil {
+			return fmt.Errorf("reindex test failed (batch_size: %d): %w", client.BatchSize(), err)
```
peternose, line 142: Is `BatchSize` needed or can we read it from the fixture or local var?

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
peternose: We can leave config. But I would still only test one client and compare its state to compute node.

go/oasis-test-runner/scenario/e2e/runtime/helpers_runtime.go
```diff
@@ -613,31 +613,11 @@ func (sc *Scenario) ensureReplicationWorked(ctx context.Context, km *oasis.Keyma
...
-	}
-}
-
-func (sc *Scenario) WaitRuntimeBlock(ch <-chan *roothash.AnnotatedBlock, round uint64) (*roothash.AnnotatedBlock, error) {
+func (sc *Scenario) WaitRuntimeBlock(ctx context.Context, ch <-chan *roothash.AnnotatedBlock, round uint64) (*roothash.AnnotatedBlock, error) {
 	sc.Logger.Info("waiting for runtime block",
 		"round", round,
 	)
 
-	timer := time.NewTimer(10 * time.Second)
```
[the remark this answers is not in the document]
martintomazic: Then we also have a problem in `sc.WaitBlocks`. Personally I think it is the caller responsibility to define the timeout.... Currently this is done when you set scenario.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -139,3 +168,74 @@ func (sc *historyReindexImpl) registerRuntime(ctx context.Context, childEnv *env
...
+func (sc *historyReindexImpl) testReindex(ctx context.Context, client *oasis.Client, rtIdx int, endRound uint64) error {
+	sc.Logger.Info("test history reindex",
+		"node", client.Name,
+	)
+
+	sc.Logger.Info("configure and restart client node",
+		"rt_idx", rtIdx,
+		"node", client.Name,
+	)
+	client.UpdateRuntimes([]int{rtIdx})
```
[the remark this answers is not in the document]
martintomazic: If I want to trigger reindex for the same range, then the client has to first have synced consensus until latest height, else reindex would be triggered until lower round. In fact I tested locally, if you start node later reindex only goes up to 0 round...

The easiest way to achive this is to have client node running. Then update config + restart to trigger tracking of new runtime, and subsequent reindex. This is very realistic e2e scenario for the reindex.

go/oasis-test-runner/scenario/e2e/runtime/helpers_runtime.go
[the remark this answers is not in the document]
peternose: Yes, in production code the caller should take care of this. But here, this is a helper method and I want to limit the context always. By default, the timeout for test is set to 20 minutes, which is too much waiting time for this function to fail.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -82,55 +84,48 @@ func (sc *historyReindexImpl) Clone() scenario.Scenario {
...
-			"current_height", newBlk.Height,
-			"wait_for_height", waitForHeight,
-		)
+	// Get runtime index.
+	rtIdx, err := sc.rtIdx()
+	if err != nil {
+		return fmt.Errorf("failed to get runtime index: %w", err)
 	}
 
-	// Restart compute worker with configured runtime.
```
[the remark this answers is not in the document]
peternose: Yes, but why was this not done before? Maybe you should ask the author why was this so, as the comments don't explain it.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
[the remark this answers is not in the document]
peternose: Oh, I get it now. You want that consensus in synced before starting reindex. That should be in the comments.

Still, changing the config which is reflected on restart is a bit hacky. I see that compute worker has the same method, which is not used anywhere (can be removed). If you don't find a better way to achieve this, than I guess this solution is also fine. An alternative would be to reindex, clear runtime history from disk, and reindex from startup.

go/oasis-test-runner/scenario/e2e/runtime/helpers_runtime.go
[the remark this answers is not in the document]
martintomazic: Then this number should not be hardcoded but instead calculated dynamically, depending on the `N` of blocks to wait, and the config that defines how frequent blocks are produced. In practice this method as is useless unless few rounds.

Finally there is consistency with existing code:
1. `sc.WaitBlocks` and
2. Other places almost always rely on the scenario timeout. If you believe this is the problem (I do) we should probably wrap context of all client calls and other that use context...

Whilst I agree with you (2.), this is out of scope of this PR. If you don't see this commit as incremental improvement of our code, I would prefer to remove it and not tackle this here... (?)

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -82,55 +84,48 @@ func (sc *historyReindexImpl) Clone() scenario.Scenario {
...
-	sc.Logger.Info("starting the compute worker")
-	if err := compute.Start(); err != nil {
-		return err
+	if rtIdx == -1 {
+		return 0, fmt.Errorf("no compute runtime configured")
 	}
+	return rtIdx, nil
+}
+
+func (sc *historyReindexImpl) registerRuntime(ctx context.Context, childEnv *env.Env, rtIdx int) error {
```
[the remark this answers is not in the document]
martintomazic: This way I save on passing one additional parameter...

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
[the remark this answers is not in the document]
martintomazic: >which is not used anywhere (can be removed).

This was result of my simplification,  that is [pending](https://github.com/oasisprotocol/oasis-core/pull/6114#discussion_r2009888258) approval.

>Still, changing the config which is reflected on restart is a bit hacky. 

Added helper method to `Client` directly for that. I think  this is better then: 1. stop the node, 2. remove hardcoded subdir of block history, 3. start again.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
martintomazic: Removed. 

I have two clients now, one is default (also used by test client transaction submission and generally `sc.Net` relies on it). The other one is configured with synced consensus (reindex). Note compute node is being prunned, so comparing runtime history with it might become flaky. Alternative is to compare second client with default one?

Last commit I added is optional, i.e. we randomize batch size of the second client every run. This way we are using CI resources that we are spending anyways + finally we at least run some batches sometimes (since default is 1000)...

go/oasis-test-runner/scenario/e2e/runtime/helpers_runtime.go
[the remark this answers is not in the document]
peternose: > Then this number should not be hardcoded but instead calculated dynamically, depending on the `N` of blocks to wait, and the config that defines how frequent blocks are produced. 

I would say that this is over-engineering. Maybe just update the code to `case <-time.After(10*time.Second)`, if you expect that the blocks are already being produced. Otherwise, you would need to wait for the first block a bit longer.

> Other places almost always rely on the scenario timeout.

I would not do that here, as I don't want to wait on CI 20 min for this to fail, if I know, that the worst case scenario is around 6s for the next block to be produced.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -101,8 +111,37 @@ func (sc *historyReindexImpl) Run(ctx context.Context, childEnv *env.Env) error
...
+	}
+
+	// Fetch latest round after test client finishes.
+	last, err := sc.fetchRuntimeBlock(ctx, sc.Net.ClientController().RuntimeClient, api.RoundLatest)
+	if err != nil {
+		return fmt.Errorf("failed to get latest block: %w", err)
+	}
+	round := last.Header.Round
+
+	// Second client has synced consensus, which ensures that upon configuring new runtime,
```
peternose, line 132: How do you know that consensus was synced? Maybe you need to fetch the last block also from the second client.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -101,8 +111,37 @@ func (sc *historyReindexImpl) Run(ctx context.Context, childEnv *env.Env) error
...
+	// reindex will be happening from 0...last round. If we started a new client here,
+	// reindex would only happen until first few blocks, as consensus sync is not blocking indexer.
```
peternose, line 134: Should we update the indexer to start after the consensus has synced?

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -120,3 +159,77 @@ func (sc *historyReindexImpl) registerRuntime(ctx context.Context, childEnv *env
...
+	compute := sc.Net.ComputeWorkers()[0]
+	ctrl1, err := oasis.NewController(compute.SocketPath())
+	if err != nil {
+		return fmt.Errorf("failed to create controller (node: %s): %w", compute.Name, err)
+	}
+	ctrl2, err := oasis.NewController(client.SocketPath())
+	if err != nil {
+		return fmt.Errorf("failed to create controller (node: %s): %w", client.Name, err)
+	}
+	indexClient, reindexClient := ctrl1.RuntimeClient, ctrl2.RuntimeClient
```
peternose, line 190: These names are confusing as before you used suffixes 1/2, and now you are using prefixes index/reindex. Options: `client1`, `client2`, or `computeCtrl`, `clientCtrl`, `comptuteRt`, `clientRt` , ...

go/runtime/history/indexer.go
```diff
@@ -230,11 +231,13 @@ func (bi *BlockIndexer) reindexTo(ctx context.Context, height int64) error {
 		lastHeight = lastRetainedHeight
 	}
 
+	batchSize := int64(bi.batchSize)
 	for start := lastHeight; start <= height; start += batchSize {
 		end := min(start+batchSize-1, height)
 		if err = bi.reindexRange(ctx, start, end); err != nil {
 			return fmt.Errorf("failed to reindex batch: %w", err)
 		}
+
```
peternose, line 240:
```suggestion
```

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -69,12 +70,17 @@ func (sc *historyReindexImpl) Fixture() (*oasis.NetworkFixture, error) {
 		},
 	}
 
+	// Add additional client node, with random batch size.
 	if n := len(f.Clients); n != 1 {
 		return nil, fmt.Errorf("expected single client node, got %d", n)
 	}
+	batchSize := rand.IntN(100)
+	sc.Logger.Info("configure second client with random batch size",
+		"batch_size", batchSize)
```
peternose, line 79: Wrong line format.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -69,12 +70,17 @@ func (sc *historyReindexImpl) Fixture() (*oasis.NetworkFixture, error) {
 		},
 	}
 
+	// Add additional client node, with random batch size.
 	if n := len(f.Clients); n != 1 {
 		return nil, fmt.Errorf("expected single client node, got %d", n)
 	}
+	batchSize := rand.IntN(100)
```
peternose, line 77:
```suggestion
	batchSize := rand.UIntN(100)
```
In general, we should avoid random numbers in tests. 

Suppose reindexing doesn't work for N=10, but works for all other numbers. In this case:
- We will merge these changes only to figure out few days or weeks latter that this doesn't work for N=10.
- Once we see that test failed, we cannot run it locally because the test will probably not fail again. So in order to debug it, we first need to change the test and hardcode batch size, fix the bug, and finally undo changes to test.

Now suppose we allow random variables and you have 10 of them in one test. Will that be easy to debug and when will the test fail?

And as I said before, for me, it doesn't make sense to test if reindex works for different batch sizes. If it works for 10, it should also work for 20. If you do test this, you should test edge cases, e.g. 0, 1, 10, and infinity. But I would not do that, not in E2E test. And if you do that, `client.ConfigureRuntime` function doesn't help you anymore, it would be better to have an option to clear the runtime state.

go/runtime/registry/registry.go
peternose: > Not sure this is better, also `config.GlobalConfig` is used in many places in registry add-hoc 

Yes, and that is not the best as you are relying on some external state. How will you unit test this? Ideally, config should be read in main function, then passed to the node constructor, and then the node should pass parameters to its workers.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
peternose: Compute node doesn't need to be pruned, or does it?

Ideally, you would have one client, do reindex, compare state to compute, reset state, reindex again, compare state to compute, ...

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
martintomazic: What would be the benefit?

Cons: Currently in such case most of the blocks goes through `index` function of instead of intended `reindex`?
Pro: you are faster as you don't block reindex with consensus sync.

go/oasis-test-runner/oasis/client.go
```diff
@@ -35,6 +36,17 @@ type ClientCfg struct {
 	RuntimeConfig      map[int]map[string]interface{}
 }
 
+// ConfigureNewRuntime configures client node with a new runtime.
+//
+// Warning: In order to apply configuration changes node has to be restarted.
+func (client *Client) ConfigureNewRuntime(ctx context.Context, rtIdx int) error {
+	client.runtimes = append(client.runtimes, rtIdx)
+	if err := client.Restart(ctx); err != nil {
```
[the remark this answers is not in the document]
martintomazic: Could also be renamed `TrackNewRuntime`... Indeed this is only used/called when new client node is started, so that it overwrites default config. Concretely by registering client node as `Feature` to the normal node. And then calling `ModifyConfig`.

I am not sure I understand why this is actually needed, as ideally  I would want to modify default config, before starting the node. Currently we have this extra abstraction, where every time you start a node, you replace existing config with default config, and finally overwrite it via the Feature...

Ideally default config should be set at init, then we could modify it and finally start it?

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
martintomazic: So this pruning was added because previously we had a bug in reindex: https://github.com/oasisprotocol/oasis-core/pull/3127.

I am wondering should I keep the old test, and make a separate scenario for this one?

>Ideally, you would have one client,

I think you need at least two clients, since default client has a bit of a special status, and a lot of helper methods/fixture rely on having it from the start, e.g. test client scenario submitting runtime transactions. I we do any `start`/`stop` of the client node, or reindex manipulation I prefer to do it on the second client for the sake of consistency/`Net` assumptions.

go/oasis-test-runner/oasis/client.go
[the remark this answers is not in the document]
peternose: Yes, just move restart to test.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
martintomazic: Client is at most one block behind (I haven't observed this in practice though). Since I am comparing much bigger range cca 100 consensus blocks this is not really problematic as I still "stress" reindex, possibly a mix of both (even better). 

If you are worried about reproducibility I am happy to fix that, at the cost of few lines of code more :).

peternose: > Client is at most one block behind 

How do you know? For me, the comment is misleading as it states something that was never tested. Therefore, I suggested, that we check this. An alternative would be to write a better comment.

go/oasis-test-runner/oasis/client.go
```diff
@@ -35,6 +35,13 @@ type ClientCfg struct {
 	RuntimeConfig      map[int]map[string]interface{}
 }
 
+// ConfigureNewRuntime configures client node with a new runtime.
+//
+// Warning: In order to apply configuration changes node has to be restarted.
+func (client *Client) ConfigureNewRuntime(rtIdx int) {
```
peternose, line 41: Warning looks too scary. This should be a note, but I would prefer to remove this part.
The comment is not totaly clear, since you are accepting index and not runtime. Will this also work if we configure the same index twice?
Why don't we keep the same name as in compute worker?

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -120,3 +161,80 @@ func (sc *historyReindexImpl) registerRuntime(ctx context.Context, childEnv *env
...
+	computeCtrl, err := oasis.NewController(compute.SocketPath())
+	if err != nil {
+		return fmt.Errorf("failed to create controller (node: %s): %w", compute.Name, err)
+	}
+	clientCtrl, err := oasis.NewController(client.SocketPath())
+	if err != nil {
+		return fmt.Errorf("failed to create controller (node: %s): %w", client.Name, err)
+	}
+
+	if err := clientCtrl.WaitReady(ctx); err != nil {
```
peternose, line 187: I would group these lines with `Restart`.

peternose (commented):
As you mentioned already, prune test should be done in another test as it makes no sense to test pruning here on compute node. If we try to test it on a client together with reindex, it will only complicate things as the storage worker won't initialize without checkpoint.

I think that the following solution (with minor fixes) is simple and covers reindex + index. My logs show that 11 blocks are reindexed and 16 indexed before the client node stops.
```go

func (sc *historyReindexImpl) Fixture() (*oasis.NetworkFixture, error) {
	f, err := sc.Scenario.Fixture()
	if err != nil {
		return nil, err
	}

	// Find the first compute runtime.
	rtIdx := slices.IndexFunc(f.Runtimes, func(rt oasis.RuntimeFixture) bool {
		return rt.Kind == registry.KindCompute
	})
	if rtIdx == -1 {
		return nil, fmt.Errorf("no compute runtime configured")
	}
	sc.rtIdx = rtIdx

	// Run selected runtime on a single compute node.
	f.ComputeWorkers = f.ComputeWorkers[:1]
	f.Runtimes[rtIdx].Executor.GroupSize = 1
	f.Runtimes[rtIdx].Executor.GroupBackupSize = 0
	f.Runtimes[rtIdx].Constraints[scheduler.KindComputeExecutor][scheduler.RoleWorker].MinPoolSize.Limit = 1
	f.Runtimes[rtIdx].Constraints[scheduler.KindComputeExecutor][scheduler.RoleBackupWorker].MinPoolSize.Limit = 0

	// Start client without any runtime.
	f.Clients[0].Runtimes = nil

	return f, nil
}

func (sc *historyReindexImpl) Run(ctx context.Context, childEnv *env.Env) error {
	// Start the network.
	if err := sc.Net.Start(); err != nil {
		return err
	}

	// Prepare few runtime blocks for reindex.
	compute := sc.Net.ComputeWorkers()[0]
	computeCtrl, err := oasis.NewController(compute.SocketPath())
	if err != nil {
		return fmt.Errorf("failed to create controller: %w", err)
	}
	blk, err := sc.waitRuntimeBlock(ctx, computeCtrl.RuntimeClient, 10)
	if err != nil {
		return fmt.Errorf("failed to wait for runtime block: %w", err)
	}

	// Ensure client has all consensus blocks needed to reindex runtime blocks.
	if err = sc.WaitBlock(ctx, blk.Height, sc.Net.ClientController()); err != nil {
		return fmt.Errorf("failed to wait for block: %w", err)
	}

	// Reindex existing runtime blocks and start indexing new ones.
	client := sc.Net.Clients()[0]
	client.UpdateRuntimes([]int{sc.rtIdx})
	if err = client.Restart(ctx); err != nil {
		return fmt.Errorf("failed to restart %s: %w", client.Name, err)
	}

	// Verify that indexing works.
	if _, err = sc.waitRuntimeBlock(ctx, sc.Net.ClientController().RuntimeClient, 20); err != nil {
		return fmt.Errorf("failed to wait for runtime block: %w", err)
	}

	// Verify (re)indexed runtime blocks.
	for round := uint64(0); round <= 20; round++ {
		if err := sc.ensureEqualBlock(ctx, computeCtrl.RuntimeClient, sc.Net.ClientController().RuntimeClient, round); err != nil {
			return fmt.Errorf("blocks not equal (round: %d): %w", round, err)
		}
	}

	// Run test client to ensure runtime works.
	sc.Logger.Info("starting the basic test client")
	if err = sc.RunTestClientAndCheckLogs(ctx, childEnv); err != nil {
		return fmt.Errorf("failed to run test client and check logs: %w", err)
	}

	return nil
}
```

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -40,123 +37,112 @@ func (sc *historyReindexImpl) Fixture() (*oasis.NetworkFixture, error) {
...
-		if newBlk.Height > waitForHeight {
-			break
-		}
-		sc.Logger.Debug("waiting enough blocks to ensure pruning",
-			"current_height", newBlk.Height,
-			"wait_for_height", waitForHeight,
-		)
+	// Reindex existing runtime blocks and start indexing new ones.
+	client := sc.Net.Clients()[0]
+	client.UpdateRuntimes([]int{sc.rtIdx})
```
martintomazic, line 81: No need to wait for consensus block explicitly since we are not checking the reindex range explicitly but is a mix of both. This simplifies the code so would prefer to keep ti this way.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -3,26 +3,23 @@ package runtime
...
-	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis/cli"
 	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario"
 	registry "github.com/oasisprotocol/oasis-core/go/registry/api"
+	"github.com/oasisprotocol/oasis-core/go/roothash/api/block"
+	"github.com/oasisprotocol/oasis-core/go/runtime/client/api"
 	scheduler "github.com/oasisprotocol/oasis-core/go/scheduler/api"
 )
 
-const (
-	reindexComputePruneNumKept = 50
```
martintomazic, line 18: We have now lost legacy prunning test, that was not best either.

1. There is existing `runtime_prune.go` but not sure we could move it there.
2. If we want to add back the old test my suggestion would be that as part of this scenario we start second client, and configure it with runtime at the end of the test. Then we check blocks were pruned, and finally we wait for latest round (meaning) reindexed succeeded. The old test was checking just that, as range was not properly calculated (taking prunner into account).

martintomazic: As agreed in private we have moved this part to separate test (no runtimes).

ptrus: So we don't test anymore that reindex works on a node that doesn't have complete history (e.g. had some blocks pruned)?

peternose: I think that this was already not tested before. If you have a node with blocks pruned, reindex works, but the problem is that the storage worker will never be ready, as it needs a checkpoint to trust blocks. Consequently, the node will not be ready and if this node is a client node, the test client will wait indefinitely.

peternose: In the original version, consensus blocks are not pruned until runtime is reindexed because the prune handler for runtime is blocking pruning. So pruning is done at the end, when everything is already indexed, I think.

go/oasis-test-runner/scenario/e2e/prune.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario"
+)
+
+const (
+	pruneNumKept = 20
+)
+
+// Prune is scenario that tests consensus block pruning.
+var Prune scenario.Scenario = &prune{
```
peternose, line 19: Structure of this scenario is not consistent with others (see how are scenarios, constructors, ... defined elsewhere).
The name of this test should be `ConsensusPrune`.
Already said that this test doesn't use runtimes so it should be in the `e2e` folder.

go/oasis-test-runner/scenario/e2e/prune.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (sc *prune) Fixture() (*oasis.NetworkFixture, error) {
+	f, err := sc.Scenario.Fixture()
+	if err != nil {
+		return nil, err
+	}
+
+	f.Validators[0].Consensus.PruneNumKept = pruneNumKept
+	f.Validators[0].Consensus.PruneInterval = time.Second
```
peternose, line 34: This is not needed as this is minimum.

go/oasis-test-runner/scenario/e2e/prune.go
```diff
@@ -0,0 +1,87 @@
...
+	)
+	ctrl := sc.Net.Controller()
+	for h := 1; h <= halfPrune; h++ {
+		if _, err := ctrl.Consensus.GetBlock(ctx, int64(h)); err != nil {
+			return fmt.Errorf("pruning too early: %w", err)
+		}
+	}
+
+	// Wait for additional blocks and verify pruning.
+	if _, err := sc.WaitBlocks(ctx, 2*pruneNumKept); err != nil {
```
peternose, line 72: Do we need to multiply with 2?

go/oasis-test-runner/scenario/e2e/prune.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	sc.Logger.Info("ensure no pruning",
+		"start_height", 1,
+		"end_height", halfPrune,
+	)
+	ctrl := sc.Net.Controller()
+	for h := 1; h <= halfPrune; h++ {
+		if _, err := ctrl.Consensus.GetBlock(ctx, int64(h)); err != nil {
+			return fmt.Errorf("pruning too early: %w", err)
```
peternose, line 67: This error probably won't be clear, e.g. `pruning too early: version not found`.

go/oasis-test-runner/scenario/e2e/prune.go
```diff
@@ -0,0 +1,87 @@
...
+		return fmt.Errorf("failed waiting for %d blocks: %w", 2*pruneNumKept, err)
+	}
+
+	sc.Logger.Info("ensure pruned blocks",
+		"start_height", 1,
+		"end_height", pruneNumKept,
+	)
+	for h := 1; h <= pruneNumKept; h++ {
+		if _, err := ctrl.Consensus.GetBlock(ctx, int64(h)); err == nil { // if NO error
+			return fmt.Errorf("pruning failed: unexpected block (height: %d)", h)
```
peternose, line 82: This error is not true, as you don't now if pruning failed. Maybe it just wasn't executed.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -36,100 +37,112 @@ func (sc *historyReindexImpl) Fixture() (*oasis.NetworkFixture, error) {
...
-	epoch, err := sc.Net.Controller().Beacon.GetEpoch(ctx, consensus.HeightLatest)
+	return nil
+}
+
+func (sc *historyReindexImpl) waitRuntimeBlock(ctx context.Context, client api.RuntimeClient, round uint64) error {
+	rtID := sc.Net.Runtimes()[sc.rtIdx].ID()
+	ch, sub, err := client.WatchBlocks(ctx, rtID)
 	if err != nil {
-		return fmt.Errorf("failed to get current epoch: %w", err)
+		return fmt.Errorf("client failed to watch runtime blocks (ID: %s)", rtID)
```
peternose, line 111:
```suggestion
		return fmt.Errorf("failed to watch runtime blocks")
```
Reader will have no idea which client are you talking about.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -36,100 +37,112 @@ func (sc *historyReindexImpl) Fixture() (*oasis.NetworkFixture, error) {
...
+	if err = sc.RunTestClientAndCheckLogs(ctx, childEnv); err != nil {
+		return fmt.Errorf("failed to run test client and check logs: %w", err)
 	}
 
-	// Fetch current epoch.
-	epoch, err := sc.Net.Controller().Beacon.GetEpoch(ctx, consensus.HeightLatest)
+	return nil
+}
+
+func (sc *historyReindexImpl) waitRuntimeBlock(ctx context.Context, client api.RuntimeClient, round uint64) error {
```
peternose, line 107: Forgot to mention that this function has the same name as the inner one, which is a bit confusing. Maybe change to `waitForRuntimeBlockFromClient` or something better.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
```diff
@@ -36,100 +37,112 @@ func (sc *historyReindexImpl) Fixture() (*oasis.NetworkFixture, error) {
...
+	blk1, err := sc.fetchRuntimeBlock(ctx, client1, round)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to fetch client1's runtime block: %w", err)
 	}
-	if err = computeCtrl.WaitReady(ctx); err != nil {
-		return err
+	blk2, err := sc.fetchRuntimeBlock(ctx, client2, round)
+	if err != nil {
+		return fmt.Errorf("failed to fetch client2's runtime block: %w", err)
```
peternose, line 127: I would say that this `client2` is too detailed, and nobody know what this is. If you accepted controllers instead of clients, then you could add the name of the node.

peternose (commented):
I don't know why were changes to scenario to set batch size removed. I said that it doesn't make sense to test sizes 1,2,5, ... but it does make sense to change the default value from 1000 to something that will commit more batches. Since we are reindexing 10 blocks, I would set batch size to something between 2-5.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
martintomazic: too long, I changed to `waitForClientRuntimeBlock`.

Also too subjective.

go/oasis-test-runner/scenario/e2e/prune.go
martintomazic: `pruning too early: rpc error: code = Unknown desc = cometbft: block query failed: height 40 must be less than or equal to the current blockchain height 10`

It will be just fine.

go/oasis-test-runner/scenario/e2e/prune.go
martintomazic: Yes, since we only prune until block `9`, if we increase to `30` to block `19`.

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
martintomazic: >If you accepted controllers instead of clients
Not sure it is possible to get the name out of controller.

>I would say that this client2 is too detailed,
If caller wraps it properly it makes sense.

go/oasis-test-runner/scenario/e2e/prune.go
martintomazic: Copied from `multiple_seeds.go`, also there are more with same style. 

>The name of this test should be ConsensusPrune.

Renamed, but previous was also fine.

>Already said that this test doesn't use runtimes so it should be in the e2e folder.

This was the case already.

go/oasis-test-runner/scenario/e2e/prune.go
peternose: We first wait for 10 blocks, then here for 2x20=40 blocks, which is together 50 blocks. The pruning happens every 1 second, which is ~1 block. Calculation shows, that if we remove 2, the pruning will still happen and we will be left with blocks 10-30 or 11-31. So we can test if blocks 1-9 were removed. 
Multiplication with 2 makes me think that pruning doesn't work as described above. Can you again explain in details why we have this here?

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
peternose: > If caller wraps it properly it makes sense.

How? You do it this way: `blocks not equal (round: 10): failed to fetch client2's runtime block`. So if you don't do it correctly, how will the caller? And from this error you actually see, that this method should return `(bool, error)`, because you are stating that blocks are not equal, which is not true, since you couldn't fetch them.

go/oasis-test-runner/scenario/e2e/prune.go
peternose: > Copied from `multiple_seeds.go`, also there are more with same style.

Oh. This is probably one of few scenarios, as I don't recall seeing a scenario without `Impl` in the name. 
 
> This was the case already.

Ops, I missed that :thinking: Sorry, ignore.

go/oasis-test-runner/scenario/e2e/scenario.go
```diff
@@ -159,6 +159,8 @@ func RegisterScenarios() error {
 		ChangeParametersMinCommissionRate,
 		// Consensus governance change reward schedule test.
 		ChangeParametersRewardSchedule,
+		// Consensus pruning test.
+		ConsensusPrune,
```
peternose, line 163: Could be moved higher under `ConsensusStateSync`.

go/oasis-test-runner/scenario/e2e/prune.go
```diff
@@ -0,0 +1,85 @@
...
+const (
+	pruneNumKept = 20
+)
+
+// ConsensusPrune is scenario that tests consensus block pruning.
+var ConsensusPrune scenario.Scenario = &prune{
+	Scenario: *NewScenario("prune"),
+}
+
+type prune struct {
```
peternose, line 22:
```suggestion
type consensusPruneImpl struct {
```

go/oasis-test-runner/scenario/e2e/prune.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario"
+)
+
+const (
+	pruneNumKept = 20
+)
+
+// ConsensusPrune is scenario that tests consensus block pruning.
+var ConsensusPrune scenario.Scenario = &prune{
+	Scenario: *NewScenario("prune"),
```
peternose, line 19:
```suggestion
	Scenario: *NewScenario("consensus-prune"),
```

go/oasis-test-runner/oasis/fixture.go
```diff
@@ -518,6 +518,9 @@ type ClientFixture struct {
 
 	// RuntimeConfig contains the per-runtime node-local configuration.
 	RuntimeConfig map[int]map[string]interface{} `json:"runtime_config,omitempty"`
+
+	// BatchSize is max number of block that block indexer writes at the same time.
+	BatchSize uint16 `json:"batch_size"`
```
peternose, line 523: This is only for clients, but maybe we want it also for compute nodes. Should we move this to `NodeFixture` or create a new one, e.g. `RuntimeFixture`, where we store runtime specific fixtures for nodes that support runtime (not sure if this already exists or is solved some other way).

go/oasis-test-runner/scenario/e2e/prune.go
peternose: A better solution and more obvious is to add time to the pruner to trigger and prune. So I would remove `2*`, add the following line, and verify up to `halfPrune`. The test should also be faster.
```go
// Give pruner, which triggers every second, enough time to prune.
time.Sleep(2 * time.Second)
```

martintomazic: I like your suggestion as it speeds-up. Tested locally and `2s` is not enough works from `3` on. Added `5` just to be safe. Ideally this would be calculated from `pruneNumKept` and `PruneInterval`. But let's keep it simple :)

go/oasis-test-runner/scenario/e2e/runtime/history_reindex.go
martintomazic: You are right and I am fixing that with.
`fmt.Errorf("failed to ensure %s and %s equal blocks (round: %d): %w", compute.Name, client.Name, round, err)`

>And from this error you actually see, that this method should return (bool, error)

I would say current version is also just fine, but happy to change if you still see this as blocking. error happens if blocks are not equal or if I failed fetching them. Yours is better but then I need two checks and error logging so prefer to keep it simple for such short scope.

Personally (and also inline with the style), best way here imo would be: 
`fmt.Errorf("ensureEqualBlock(%s, %s): %w", compute.Name, client.Name, err)`. You are definitely not going to grep it, and errors are for programmers to debug. But we discussed this already and I get your take.

Finally idealy we would remove redundant `failed to`... since already we have `Error: [lowercase sentence]`, but am keeping it for the sake of consistency. This would in general give us more space and make our errors less opaque :).

go/oasis-test-runner/oasis/fixture.go
martintomazic: I suggest doing that _when/if_ this becomes needed.

peternose (approved):

PR #6116, opened 2025-03-21
go/consensus/cometbft/roothash: Move prune handler

peternose:


go/runtime/history/history.go
```diff
@@ -29,6 +30,7 @@ type Factory func(runtimeID common.Namespace, dataDir string) (History, error)
 // History is the runtime history interface.
 type History interface {
 	roothash.BlockHistory
+	consensus.StatePruneHandler
```
peternose, line 33: Currently, this is not the ideal solution. However, it might become one in the future if the `History` interface is removed, allowing `runtimeHistory` to be renamed to `History` and behave like `StatePruneHandler`.

kostko (approved):

PR #6120, opened 2025-03-22
go/consensus: Clean light client interfaces

peternose:
I think our light client service implementation is not very useful and that the whole service can be removed:
- The CometBFT light client is never used.
- The trust store is never pruned.
- The trust store contains only a few valid light blocks (one per epoch, not older than trust period).
  - This causes issues for the CometBFT light client, as it will give up if the primary provider cannot find light block at the requested height.
- Light blocks provided by the service from remote peers are useful only if they cannot be found locally. 
  - This means, that the local consensus state must be pruned very frequently, i.e., faster than the trust period. I'm not sure if we want to have these kind of settings in production.

Maybe I overlooked something. What am I missing?


kostko (approved):

ptrus:
> The CometBFT light client is never used.
> Light blocks provided by the service from remote peers are useful only if they cannot be found locally.

The main idea was that once we have standalone consensus light client nodes, they would be implemented on top of this and used it (https://github.com/oasisprotocol/oasis-core/issues/3116). However, this was likely premature, as the situation has evolved since then.

ptrus (approved):

PR #6123, opened 2025-03-28
go/runtime: Add support for simple per-component networking

### LINKED ISSUE #5903: Easy network provisioning for ROFL apps
It should be easy for TDX-based ROFL apps to accept incoming network connections. This would be configured in the node configuration in such a way that requests routed to the given incoming TCP/UDP ports would be forwarded to the given VM.

The networking configuration part can be added to the per-component configuration under a new `networking` key:
https://github.com/oasisprotocol/oasis-core/blob/87cf5ac9c835b5bfb8119d2373aaf87f2e232144/go/runtime/config/config.go#L212-L226

And then the QEMU-based provisioner can use it to configure forwarding to the VM:
https://github.com/oasisprotocol/oasis-core/blob/87cf5ac9c835b5bfb8119d2373aaf87f2e232144/go/runtime/host/tdx/qemu.go#L242-L244

kostko:
Fixes #5903

### DIFF
--- a/go/runtime/config/config.go
+++ b/go/runtime/config/config.go
@@ -3,6 +3,7 @@ package config
 
 import (
 	"fmt"
+	"net"
 	"time"
 
 	"gopkg.in/yaml.v3"
@@ -223,6 +224,9 @@ type ComponentConfig struct {
 	// Disabled specifies whether the component is disabled. If a component is specified and not
 	// disabled, it is enabled.
 	Disabled bool `yaml:"disabled,omitempty"`
+
+	// Networking contains the networking configuration for a component.
+	Networking NetworkingConfig `yaml:"networking,omitempty"`
 }
 
 // Validate validates the component configuration.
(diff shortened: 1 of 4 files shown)


go/runtime/config/config.go
```diff
@@ -269,6 +272,24 @@ func (c *ComponentConfig) UnmarshalYAML(value *yaml.Node) error {
...
+}
+
+// IncomingNetworkingConfig describes an IP/port to expose to the component from the host.
+type IncomingNetworkingConfig struct {
+	// IP is the host IP address to expose to the component.
+	IP string `yaml:"ip,omitempty"`
+	// Protocol is the optional protocol to expose to the component.
+	Protocol string `yaml:"protocol,omitempty"`
+	// SrcPort is the source port (on the host).
+	SrcPort uint16 `yaml:"src_port"`
```
matevz, line 288: If you want to forward multiple ports to the same IP, then you make multiple `IncomingNetworkConfig` sections right?

matevz (approved):
I can confirm this works!

go/runtime/host/tdx/qemu.go
```diff
@@ -266,6 +272,45 @@ func (p *qemuProvisioner) getSandboxConfig(cfg host.Config, _ sandbox.Connector,
 	return pcfg, nil
 }
 
+// createNetworkingConfig generates QEMU networking configuration for a component.
+func (p *qemuProvisioner) createNetworkingConfig(cfg host.Config) ([]string, error) {
+	compCfg, _ := config.GlobalConfig.Runtime.GetComponent(cfg.ID, cfg.Component.ID())
+	netdevOpts := []string{"user", "id=net0"}
+
+	// Inbound forwarding.
+	for _, netCfg := range compCfg.Networking.Incoming {
```
ptrus, line 281: Should we validate that same ip,protocol,source port combination doesn't repeat among the rules? 

Otherwise we will probably get (a not very nice) error on quemu start.

go/runtime/config/config.go
kostko: Yeah.

go/runtime/host/tdx/qemu.go
kostko: Added some config validation, please take a look.

ptrus (approved):

PR #6125, opened 2025-03-29
Fix type inference issues caused by deranged crate update

kostko:
The deranged `deranged` crate update has caused havoc in a lot of random crates due to overlapping `PartialOrd` trait implementations.

See https://github.com/jhpratt/deranged/issues/18 and https://github.com/jhpratt/deranged/issues/19.

Currently uses a fix from the https://github.com/fortanix/rust-sgx/pull/726 branch which should be changed to a released version once it becomes available.


runtime-loader/Cargo.toml
```diff
@@ -11,9 +11,12 @@ futures = { version = "0.3.31", features = ["compat", "io-compat"] }
...
 [target.'cfg(target_os = "linux")'.dependencies]
-aesm-client = { version = "0.6.0", features = ["sgxs"] }
-enclave-runner = { version = "0.6.0" }
-sgxs-loaders = { version = "0.4.0" }
+# TODO: Change back to released version once upstream fix released.
+aesm-client = { git = "https://github.com/fortanix/rust-sgx", branch = "raoul/rte-418-fix_enclave_runner", features = ["sgxs"] }
+# TODO: Change back to released version once upstream fix released.
+enclave-runner = { git = "https://github.com/fortanix/rust-sgx", branch = "raoul/rte-418-fix_enclave_runner" }
+# TODO: Change back to released version once upstream fix released.
+sgxs-loaders = { git = "https://github.com/fortanix/rust-sgx", branch = "raoul/rte-418-fix_enclave_runner" }
```
martintomazic, line 19: Oasis docs/development set-up/prerequisites:  `cargo install sgxs-tools`  is  also not build correctly without this patch.

Or is this temp and we can avoid updating documentation...?

kostko: This is temp as mentioned in the comment. See linked upstream PR.

kostko: I think Cargo.lock should contain the exact revision, are you sure it doesn't?

kostko: Oh I see, yeah rev would be better but I hope they will do a new release soon.

peternose: They will probably have to, so we can leave it for now. We just need to be careful that we don't update this branch.

martintomazic: Hm I guess our docker on master is still failing. Which is strange as I would expect it uses `Cargo.toml`?

kostko: That could be failing in the sgxs tools install which would need to be changed to also use this branch (it uses the released version by default). I'm afk so can't check right now.

PR #6127, opened 2025-03-31
changelog: Assemble changes for 25.1 release

kostko:


peternose (approved):

ptrus:
Can we also release 20.12 and 20.10 (with archive nodes updates) before 25.1 (so that the latest release will be 25.1).

ptrus:
Not sure if the automatic release process will work, since it's old branches, hopefully it will 🤞  Otherwise we can maybe manually do it, since it's a one-off thing.

kostko:
Oh, the release is already in progress.

kostko:
Ok, the release process has been stopped.

PR #6132, opened 2025-04-01
go/storage/mkvs/node: Fix SetBit method of the Key

martintomazic:
`SetBit` method was clearing all remaining bits instead of zeroing the chosen one. We could also remove since currently not used.

### DIFF
--- a/go/storage/mkvs/node/key.go
+++ b/go/storage/mkvs/node/key.go
@@ -94,7 +94,7 @@ func (k Key) SetBit(bit Depth, val bool) Key {
 	if val {
 		kb[bit/8] |= mask
 	} else {
-		kb[bit/8] &= mask
+		kb[bit/8] &^= mask
 	}
 	return kb
 }
--- a/go/storage/mkvs/node/key_test.go
+++ b/go/storage/mkvs/node/key_test.go
@@ -6,6 +6,37 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestSetBit(t *testing.T) {
+	testCases := []struct {
+		key  Key
+		bit  Depth
+		val  bool
+		want Key
+		name string
+	}{
+		{Key{0b00000000}, 6, true, Key{0b00000010}, "Set bit"},
+		{Key{0b00000010}, 6, true, Key{0b00000010}, "Set bit (no change)"},
+		{Key{0b00000000, 0b00000000}, 6, true, Key{0b00000010, 0b00000000}, "Set bit in first byte"},
+		{Key{0b00000010, 0b00000000}, 6, true, Key{0b00000010, 0b00000000}, "Set bit in first byte (no change)"},
+		{Key{0b00000000, 0b00000000}, 14, true, Key{0b00000000, 0b00000010}, "Set bit in last byte"},
+		{Key{0b00000000, 0b00000010}, 14, true, Key{0b00000000, 0b00000010}, "Set bit in last byte (no change)"},
+		{Key{0b11111111}, 6, false, Key{0b11111101}, "Unset bit"},
+		{Key{0b11111101}, 6, false, Key{0b11111101}, "Unset bit (no change)"},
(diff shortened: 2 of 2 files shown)


go/storage/mkvs/node/key.go
```diff
@@ -26,12 +26,6 @@ func (k Key) MarshalBinary() (data []byte, err error) {
 	return
 }
 
-// UnmarshalBinary decodes a binary marshaled key including the length in bytes.
-func (k *Key) UnmarshalBinary(data []byte) error {
```
peternose, line 30: I would not remove this function as it is public and may be used in other packages, and I think it is nice that `MarshalBinary` is accompanied by `UnmarshalBinary`. It doesn't hurt anyone to have it there.

go/storage/mkvs/node/key_test.go
```diff
@@ -82,6 +82,13 @@ func TestKeyAppendSplitMerge(t *testing.T) {
 	require.Equal(t, Key{0x41, 0x6b, 0x37}, newKey)
 }
 
+func TestSetBit(t *testing.T) {
+	// Round trip should produce equal key.
```
peternose, line 86: This is a poor test. You should prepare a test vector and test bits manually.

martintomazic: The test catches previous bug. The whole test suite is not best/could cover more. Will add this later if you still see it as blocking.

peternose: > The test catches previous bug.

It does, but it doesn't catch many other bugs.

> The whole test suite is not best/could cover more.

So what? If you're adding a unit test for this function, that shouldn't be an issue.

> Will add this later if you still see it as blocking.

OK, improve it later.

martintomazic: I have followed style from `TestKeyAppendSplitMerge`. Also used vectors from that test.

Ideally this would be a table test, or we could set bit by bit and remove it. Please let me know what do you prefer. :)

peternose: As I said in the previous comment, it should be a table test. And you should have written values in binaries so that it is easier to verify.

```go
func TestSetBit(t *testing.T) {
	testCases := []struct {
		key      Key
		bit      Depth
		val      bool
		expected Key
		name     string
	}{
		{Key{0b00000000}, 6, true, Key{0b00000010}, "Set bit"},
		{Key{0b00000010}, 6, true, Key{0b00000010}, "Set bit (no change)"},
		{Key{0b00000000, 0b00000000}, 6, true, Key{0b00000010, 0b00000000}, "Set bit in first byte"},
		{Key{0b00000010, 0b00000000}, 6, true, Key{0b00000010, 0b00000000}, "Set bit in first byte (no change)"},
		{Key{0b00000000, 0b00000000}, 14, true, Key{0b00000000, 0b00000010}, "Set bit in last byte"},
		{Key{0b00000000, 0b00000010}, 14, true, Key{0b00000000, 0b00000010}, "Set bit in last byte (no change)"},
		{Key{0b11111111}, 6, false, Key{0b11111101}, "Unset bit"},
		{Key{0b11111101}, 6, false, Key{0b11111101}, "Unset bit (no change)"},
		{Key{0b11111111, 0b11111111}, 6, false, Key{0b11111101, 0b11111111}, "Unset bit in first byte"},
		{Key{0b11111101, 0b11111111}, 6, false, Key{0b11111101, 0b11111111}, "Unset bit in first byte (no change)"},
		{Key{0b11111111, 0b11111111}, 14, false, Key{0b11111111, 0b11111101}, "Unset bit in last byte"},
		{Key{0b11111111, 0b11111101}, 14, false, Key{0b11111111, 0b11111101}, "Unset bit in last byte (no change)"},
	}

	for _, tc := range testCases {
		t.Run(tc.name, func(t *testing.T) {
			newKey := tc.key.SetBit(tc.bit, tc.val)
			require.Equal(t, tc.expected, newKey)
		})
	}
}
```

peternose (approved):

PR #6137, opened 2025-04-07
go/consensus/cometbft/api: Simplify consensus service client

peternose:
This will remove `DeliverCommand` from consensus `ServiceClient`, making the API more clean and limited only to two methods `DeliverBlock` and `DeliverEvents`. As a result, anyone with access to blocks and events can now run the client.

### DIFF
--- a/go/consensus/cometbft/roothash/roothash.go
+++ b/go/consensus/cometbft/roothash/roothash.go
@@ -5,6 +5,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
+	"slices"
 	"sync"
 
 	cmtabcitypes "github.com/cometbft/cometbft/abci/types"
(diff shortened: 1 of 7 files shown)


kostko (approved):

go/consensus/cometbft/roothash/roothash.go
```diff
@@ -435,67 +412,86 @@ func (sc *serviceClient) DeliverEvent(ctx context.Context, height int64, tx cmtt
...
+						sc.logger.Error("failed to fetch block",
+							"err", err,
+							"height", h,
+							"runtime_id", tr.runtimeID,
+						)
+						return fmt.Errorf("roothash: failed to fetch block: %w", err)
+					}
+					if err := sc.emitBlock(tr, oldBlk); err != nil {
+						return fmt.Errorf("roothash: failed to emit block: %w", err)
+					}
```
martintomazic, line 447: If runtime is suspended/no blocks are produced during "missing range" then `fetchBlock` would return same runtime block header multiple times, that would then error in `emitBlock` since its the round would not increase?

Do I understand correctly that `sc.emitBlock` here should only be called if round is increased?

martintomazic (commented):
Hats of for this PR(!), as it simplifies our code significantly.

I left one minor question... :)

go/consensus/cometbft/roothash/roothash.go
peternose: Yes, let me fix this. I destroyed this as I moved block ignore out of the `emitBlock`.

martintomazic (approved):

PR #6144, opened 2025-04-09
go/consensus/api: Add method GetLatestHeight

peternose:


go/consensus/api/grpc.go
```diff
@@ -184,11 +190,11 @@ var (
 )
 
 func handlerSubmitTx(
-	srv interface{},
+	srv any,
```
ptrus, line 193: Should we do a commit that updates these across all grpc backends?

ptrus (approved):

kostko (approved):

go/consensus/api/grpc.go
peternose: Actually, I can replace it everywhere, except in `aesm_proto.pb.go`.

PR #6159, opened 2025-04-18
go/runtime: Add configurable CID pool

### LINKED ISSUE #6122: tdx-qemu: Make the CID range configurable
Currently a node will always use the same CID range for vsock addresses. We should make it configurable.

kostko:
Fixes #6122

### DIFF
--- a/go/runtime/host/sandbox/host.go
+++ b/go/runtime/host/sandbox/host.go
@@ -445,6 +445,8 @@ func (h *sandboxHost) manager(ctx context.Context) {
 
 		// Notify subscribers that the runtime has stopped.
 		h.notifier.Broadcast(&host.Event{Stopped: &host.StoppedEvent{}})
+
+		h.cfg.Cleanup(h.rtCfg)
 	}()
 
 	// Subscribe to own events to make sure the cached CapabilityTEE remains up-to-date.
@@ -457,6 +459,13 @@ func (h *sandboxHost) manager(ctx context.Context) {
 		watchdogCh   <-chan time.Time
 	)
 	for {
+		// Terminate immediately when requested.
+		select {
+		case <-ctx.Done():
+			return
+		default:
+		}
+
 		// Make sure to restart the process if terminated.
 		if h.process == nil {
 			firstTickCh := make(chan struct{}, 1)
--- a/go/runtime/host/tdx/cid_pool.go
+++ b/go/runtime/host/tdx/cid_pool.go
@@ -1,0 +1,86 @@
+package tdx
+
+import (
+	"fmt"
+	"math"
+	"sync"
+)
+
+// CidPool is a pool of context IDs (CIDs) that can be assigned to virtual machines (VMs) to
+// identify them in VSOCK connections.
+type CidPool struct {
+	l sync.Mutex
+
+	start uint32
+	end   uint32
+
+	free map[uint32]struct{}
+	used map[uint32]struct{}
+}
+
+// NewCidPool creates a new CID pool containing the given range of CIDs.
+func NewCidPool(start, count uint32) (*CidPool, error) {
+	if start < 10 {
+		return nil, fmt.Errorf("CID identifiers between 0 and 10 are reserved")
+	}
+	if count > 4096 {
+		return nil, fmt.Errorf("maximum CID pool size is 4096")
+	}
+	if start > math.MaxUint32-count {
+		return nil, fmt.Errorf("CID pool would overflow")
+	}
+
+	c := CidPool{
+		start: start,
+		end:   start + count,
+		free:  make(map[uint32]struct{}),
+		used:  make(map[uint32]struct{}),
+	}
+	for cid := start; cid < start+count; cid++ {
+		c.free[cid] = struct{}{}
+	}
+	return &c, nil
+}
+
+// Allocate allocates a CID from the pool.
+func (c *CidPool) Allocate() (uint32, error) {
+	c.l.Lock()
+	defer c.l.Unlock()
+
+	for cid := range c.free {
+		c.used[cid] = struct{}{}
+		delete(c.free, cid)
+		return cid, nil
+	}
+	return 0, fmt.Errorf("no free CIDs available")
+}
+
+// AllocateExact allocates a specific CID from the pool.
+func (c *CidPool) AllocateExact(cid uint32) error {
+	c.l.Lock()
+	defer c.l.Unlock()
+
+	if cid < c.start || cid >= c.end {
+		return fmt.Errorf("CID %d is outside the allocated range (%d - %d)", cid, c.start, c.end)
+	}
+	if _, ok := c.free[cid]; !ok {
+		return fmt.Errorf("CID %d is already in use", cid)
+	}
+
+	c.used[cid] = struct{}{}
+	delete(c.free, cid)
+	return nil
+}
+
+// Release releases the given previously allocated CID back to the pool.
+func (c *CidPool) Release(cid uint32) bool {
+	c.l.Lock()
+	defer c.l.Unlock()
+
+	if _, ok := c.used[cid]; !ok {
+		return false
+	}
+	delete(c.used, cid)
+	c.free[cid] = struct{}{}
+	return true
+}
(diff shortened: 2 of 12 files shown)


go/runtime/host/tdx/cid_pool.go
```diff
@@ -0,0 +1,55 @@
...
+	return 0, fmt.Errorf("no free CIDs available")
+}
+
+// Release releases the given previously allocated CID back to the pool.
+func (c *CidPool) Release(cid uint32) {
+	c.l.Lock()
+	defer c.l.Unlock()
+
+	if _, ok := c.used[cid]; !ok {
+		return
```
peternose, line 51: This could return an `error` or `bool` so that the caller can know if there is a bug in his code that doesn't allocate/release CIDs correctly. In the end, the caller can ignore the returned value.

go/runtime/host/tdx/cid_pool.go
```diff
@@ -0,0 +1,55 @@
+package tdx
+
+import (
+	"fmt"
+	"sync"
+)
+
+// CidPool is a pool of CIDs that can be assigned to VMs.
```
peternose, line 8:
```suggestion
// CidPool is a pool of Content IDs (CIDs) that can be assigned to Virtual Machines (VMs).
```

go/runtime/host/tdx/qemu.go
```diff
@@ -131,12 +129,14 @@ func NewQemuProvisioner(cfg QemuConfig) (host.Provisioner, error) {
 func (p *qemuProvisioner) NewRuntime(cfg host.Config) (host.Runtime, error) {
 	// Assign CID if not explicitly configured.
 	if cfg.Extra == nil {
-		p.l.Lock()
+		cid, err := p.cidPool.Allocate()
```
peternose, line 132: When is this released?

go/runtime/host/tdx/cid_pool.go
```diff
@@ -0,0 +1,55 @@
...
+	if start < 10 {
+		return nil, fmt.Errorf("CID identifiers between 0 and 10 are reserved")
+	}
+
+	c := CidPool{
+		free: make(map[uint32]struct{}),
+		used: make(map[uint32]struct{}),
+	}
+	for cid := start; cid < start+count; cid++ {
+		c.free[cid] = struct{}{}
```
peternose, line 27: If `count` is big, you are wasting memory here. I believe that there will be little used CIDs, so I would rather have only one map `used` and change `Allocate` method to iterate until it finds one that is not used.
If you decide to stay with this approach, you can also remove `used` and solve this with `free`, `start` and `count`.

go/runtime/host/tdx/cid_pool.go
```diff
@@ -0,0 +1,55 @@
...
+type CidPool struct {
+	l sync.Mutex
+
+	free map[uint32]struct{}
+	used map[uint32]struct{}
+}
+
+// NewCidPool creates a new CID pool containing the given range of CIDs.
+func NewCidPool(start, count uint32) (*CidPool, error) {
+	if start < 10 {
```
peternose, line 18: Can also verify overflow `start <= MaxUint32 - count`.

go/runtime/host/tdx/qemu.go
kostko: Ooops, fixed ;)

go/runtime/host/tdx/cid_pool.go
kostko: These are actually Context IDs, let me update the comment.

peternose: Ups. That is why it is good to define abbreviation before you start using it. And for me the best place is in the struct's comment.

go/runtime/host/tdx/cid_pool.go
kostko: I know, initially I had the more complicated variant implemented, but then just simplified it like this. I don't think this will contain a huge amount of entries in practice as running even 1024 virtual machines on a single host would probably not work very well.

And for 1024 CIDs this takes around 4KiB + overhead.

go/runtime/host/tdx/qemu.go
```diff
@@ -131,12 +129,17 @@ func NewQemuProvisioner(cfg QemuConfig) (host.Provisioner, error) {
...
+			CID: cid,
+		}
+		cfg.Cleanup = func() {
+			p.cidPool.Release(cid)
 		}
-		p.nextCid++
-		p.l.Unlock()
 	}
 
 	// Ensure any extra configuration is of the correct type.
```
peternose, line 145: Should we verify that CID is free and not used, and reserve it if in range?

go/runtime/host/host.go
```diff
@@ -32,6 +32,9 @@ type Config struct {
 
 	// LocalConfig is the node-local runtime configuration.
 	LocalConfig map[string]any
+
+	// Cleanup is an optional function that is called when the runtime is stopped.
+	Cleanup func()
```
peternose, line 37: Would this be less hackish if it were moved to the sandbox config and behaved like `GetSandboxConfig`?

go/runtime/host/tdx/cid_pool.go
peternose: I don't see any change.

go/runtime/host/host.go
kostko: Good idea, will change.

go/runtime/host/tdx/cid_pool.go
kostko: What do you mean, it has not yet been pushed?

peternose: Oh, you pushed without this change. I thought it was already fixed in [143d30b](https://github.com/oasisprotocol/oasis-core/pull/6159/commits/143d30b29e4e0b055d11e06915099d0d97fc0dd6).

kostko: Pushed now.

go/runtime/host/tdx/cid_pool.go
kostko: Changed.

go/runtime/host/host.go
kostko: Changed.

go/runtime/host/tdx/cid_pool.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	return 0, fmt.Errorf("no free CIDs available")
+}
+
+// AllocateExact allocates a specific CID from the pool.
+func (c *CidPool) AllocateExact(cid uint32) error {
+	c.l.Lock()
+	defer c.l.Unlock()
+
+	if cid < c.start || cid > c.end {
```
peternose, line 63:
```suggestion
	if cid < c.start || cid >= c.end {
```
Current implementation doesn't include `end`.

go/runtime/host/sandbox/host.go
```diff
@@ -445,6 +445,10 @@ func (h *sandboxHost) manager(ctx context.Context) {
 
 		// Notify subscribers that the runtime has stopped.
 		h.notifier.Broadcast(&host.Event{Stopped: &host.StoppedEvent{}})
+
+		if h.cfg.Cleanup != nil {
```
peternose, line 449: I would disallow `nil` cleanup functions and update the other constructors to set a no-op cleanup instead.

peternose: > Why? Isn't it a nice thing that the default type still works with reasonable defaults?

Yes, this awesome. But unfortunately, you now start wondering which parameters can be `nil` and which not. So now, you need to change line 447 (2 lines above) to
```go
if notifier != nil {
    h.notifier.Broadcast(&host.Event{Stopped: &host.StoppedEvent{}})
}
```
because you are not sure. And more fields you have in the config or struct, the harder is to follow.

To avoid this, one solution is to add comment to `Cleanup` that it can be `nil`, but who reads comments? So the best solution is to program in a way that everything is not `nil`. And if that is not possible, there should be warning comments everywhere.

kostko: Changed to make it consistent with other config options.

peternose (approved):

PR #6165, opened 2025-04-23
go/storage/mkvs: Paralellize checkpoint creation

martintomazic:
**WIP**

Checkpoint creation should be done in parallel, by fetching first few levels (e.g. 8 subtrees) and then creating independent subtrees with independent cache and iterator.

Initially, I was generating proofs for subtrees + one initial chunk that covered everything in between root and subroots. As a result I also had to modify the checkpoint metada && some other hacks.

I think there is a much simpler solution that was prototyped today (still have to fix some stuff). Basically instead of making a proof for subtree, we include the path from the root to subtree and then automatically continue appending nodes to existing proof as we iterate over subtree. -> much simplified code + avoids making `pathbadger` work with subtrees.


go/storage/mkvs/subtree.go
```diff
@@ -0,0 +1,96 @@
...
+	st.tree.Close()
+}
+
+type sbtree struct {
+	root *node.Pointer
+	path []*node.Pointer
+}
+
+func (t *tree) Subtrees(ctx context.Context, pb func() *syncer.ProofBuilder, depth int) ([]Subtree, error) {
+	root := sbtree{
```
martintomazic, line 30: TODO: **write a test suite** and fix the issue:

Currently it works only with `depth=0` (whole tree, with iterator from root). If `depth > 0` then returned subtrees, will produce `Iterator` that only traverses left children... Should be a trivial bug to fix likely in my existing code.

Also handle the case if depth is too big/or very shallow tree on the side...

go/storage/mkvs/subtree.go
```diff
@@ -0,0 +1,96 @@
...
+		return
+	}
+	return
+}
+
+func newSubtreeWithRoot(ndb db.NodeDB, root node.Root, subtree *node.Pointer) *tree {
+	t := New(nil, ndb, root.Type).(*tree)
+	t.cache.setPendingRoot(subtree)
+	t.cache.setSyncRoot(root)
+	return t
```
martintomazic, line 95: `t.cache.setPendingRoot(subtree)`:

This is a hack. Consider making https://github.com/oasisprotocol/oasis-core/blob/c5b9896618d9a02da465846fadebe6b7560ca722/go/storage/mkvs/iterator.go#L202 `Seek` parametric with subtree pointer instead

go/storage/mkvs/subtree.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+type subtree struct {
+	tree         *tree
+	proofBuilder *syncer.ProofBuilder
+}
+
+func (st *subtree) Iterator(ctx context.Context) Iterator {
+	return st.tree.NewIterator(ctx, WithProofBuilder(st.proofBuilder))
+}
```
martintomazic, line 18: Fix this: this should return iterator with independent proof generator...

Concretely move
```
		for _, n := range st.path {
			pb.Include(n.Node)
		}
``` 

here +  store `st.path` as part of `subtree` struct instead of `proofbuilder`...

go/storage/mkvs/mkvs.go
```diff
@@ -107,4 +107,11 @@ type Tree interface {
 
 	// RootType returns the storage root type.
 	RootType() node.RootType
+
+	Subtrees(ctx context.Context, pb func() *syncer.ProofBuilder, depth int) ([]Subtree, error)
+}
+
+type Subtree interface {
+	Iterator(ctx context.Context) Iterator
+	Close()
```
martintomazic, line 116: "accept interfaces return structs" ... should return struct.

go/storage/mkvs/subtree.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return
+	}
+	switch n := nd.(type) {
+	case *node.InternalNode:
+		return n.Left, n.Right, nil
+	case *node.LeafNode:
+		return
+	}
+	return
```
martintomazic, line 88: never use blank returns, missing default.

go/storage/mkvs/subtree.go
martintomazic: Working on a fix for this: Basically the problem is https://github.com/oasisprotocol/oasis-core/blob/2c59bdc646705374b24fb7f075aabc00dca138a8/go/storage/mkvs/iterator.go#L227

since here we pass the whole key from the leaf, but instead we should pass only suffix of the leaf key, since this is subtree...

Also wondering at the same time if iterator could not be implemented simpler, i.e. pre-order traversal (interested only in non-nil leaf nodes that hold keys)....

go/storage/mkvs/subtree.go
martintomazic: Alternative to `Subtree` abstraction is to define it as path that goes from root to subtree of choice (slice of internal nodes). 

Then we simply add `Iterator.SetSubtree(subtree []*node.InternalNode)`, that takes prefix path into consideration when doing iteration (also if generating proof).

go/storage/mkvs/iterator.go
```diff
@@ -226,7 +251,8 @@ func (it *treeIterator) Next() {
 		// next node.
 		key := it.key
 		it.reset()
-		err := it.doNext(atom.ptr, atom.bitDepth, atom.path, key, atom.state)
+		_, suf := key.Split(it.subtreeDepth, key.BitLength()) // TODO
+		err := it.doNext(atom.ptr, atom.bitDepth, atom.path, suf, atom.state)
```
martintomazic, line 255: I am no longer convinced this will work. Original idea was to account for subtree access path from the root, as otherwise we don't respect subtree boundaries.

Unfortunately, this has proven cumbersome, since `doNext` "state machine" is broken, due to broken bitlenth/key equality comparisons. 


**I think I have found a much simpler solution:**
* Store `offset` and `offsetBitLength` as part of the `treeIterator` struct.
* If this is not set this will work like normal iterator just like before.
* If set:
     * `Rewind`  should start with `offset`. (This will automatically add access path into proof - existing logic...) :) 
     * After every `it.doNext` we should check if the new key (if found) has still `offset` as a prefix. If not invalidate the iterator since we have just ended subtree iteration.
 * Finally like before every subtree should be independent mkvs `Tree` instance (i.e. independent cache and iterator to facilitate parallelization).

martintomazic:
**Preliminary benchmarks**

Creating checkpoint for height `5462451`, with corresponding `250GB` mkvs state (no-pruning) **runs `2.5-3x` faster**. This was benchmarked using subtrees  at the `depth` `7` and `10`.

Our **state trie** however is **quite unbalanced**, which results in many subtrees being iterated quickly, which decreases parallelization as we progress towards the checkpoint finish. Moreover, as we increase amount of subtrees, the amount of chunks also increases (many partially filled ones).

I think we should consider alternative algorithm, where there is a worker thread processing subtree chunking.
As soon as subtree has been chunked, we should split one remaining into two, so that we keep the amount of parallelization stable.

martintomazic:
The logic seems to be sound now. I still need to do a bit of a clean-up/sanity tests.

Related to state trie being unbalanced, and the need to dynamically generate subtrees:
* I am becoming more and more convinced existing checkpoint design (`mkvs/checkpoint`) that uses `mkvs.Iterator` as a client is not optimal for efficient checkpoint creation: see [CreateCheckpoint](https://github.com/oasisprotocol/oasis-core/blob/e2db97e334ef39dd2b19be4f589c0cacdfeb1db7/go/storage/mkvs/checkpoint/file.go#L34), [createChunk](https://github.com/oasisprotocol/oasis-core/blob/e2db97e334ef39dd2b19be4f589c0cacdfeb1db7/go/storage/mkvs/checkpoint/chunk.go#L19).
* Ideally `mkvs` package should expose procedure for creating checkpoint, and not rely on `mkvs.Iterator` and `Subtree` abstraction at all? `checkpoint.CreateCheckpoint` should then simply delegate the work to it. E.g:


```go
package mkvs

type CheckpointMetada struct { // This now becomes part of mkvs API.
	Version uint16      `json:"version"`
	Root    node.Root   `json:"root"`
	Chunks  []hash.Hash `json:"chunks"`
}

func CreateCheckpoint(chunksDir string, ndb db.NodeDB, root node.Root, chunkSize uint64) CheckpointMetada {
	// Traverse in parallel (independent caches), and generate proofs (respecting chunkSize) used for chunk binary dump. 
	// Operate directly on pointers, to ease ad-hoc parallel traversal of subtrees.
}
```

kostko:
Moving the logic makes sense to me, but maybe expose the output in the form of a chunk iterator instead of writing to files. Or a database-specific way and then also having the chunk fetch implemented by the database.

This is to make sure that different options can still be implemented, eg. one which does not write to separate files.

PR #6171, opened 2025-04-30
go/beacon/api: Remove SetableBackend interface

peternose:
Removing `SetableBackend` interface, as transaction submission does not belong in the beacon backend implementation. This change also removes the dependency between the beacon backend and the consensus backend/transaction submission manager.

### DIFF
--- a/go/beacon/api/api.go
+++ b/go/beacon/api/api.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/oasisprotocol/oasis-core/go/common/errors"
 	"github.com/oasisprotocol/oasis-core/go/common/pubsub"
+	"github.com/oasisprotocol/oasis-core/go/consensus/api/transaction"
 )
 
 const (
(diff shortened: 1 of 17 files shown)


go/beacon/api/api.go
```diff
@@ -35,6 +36,15 @@ var (
 	// ErrBeaconNotAvailable is the error returned when a beacon is not
 	// available for the requested height for any reason.
 	ErrBeaconNotAvailable = errors.New(ModuleName, 2, "beacon: random beacon not available")
+
+	// MethodSetEpoch is the method name for setting epochs.
+	MethodSetEpoch = transaction.NewMethodName(ModuleName, "SetEpoch", EpochTime(0))
```
peternose, line 41: This change renames module name from `AppName = 000_beacon` to `ModuleName = beacon`. I think that in production this should not be a breaking change, even though we now return a different error when `SetEpoch` transaction is executed.

kostko: `SetEpoch` is a debug-only option which cannot be enabled in production, so it should be fine if there are no other side effects.

kostko: Oh I see, I think error messages are excluded as long as module and error code are the same. And in this case the module would be unknown and error code the default of 1.

kostko (approved):

PR #6176, opened 2025-05-03
go/storagek/mkvs: Simplify Iterator's doNext function

martintomazic:
Whilst working on #6165 (still WIP) I found iterator code quite hard to follow/debug. I tried to simplify it a bit but am still not 100% convinced by the changes ([see comment](https://github.com/oasisprotocol/oasis-core/pull/6176#discussion_r2072365166)).

### DIFF
--- a/go/storage/mkvs/iterator.go
+++ b/go/storage/mkvs/iterator.go
@@ -267,6 +267,17 @@ func (it *treeIterator) doNext(ptr *node.Pointer, bitDepth node.Depth, path, key
 		bitLength := bitDepth + n.LabelBitLength
 		newPath := path.Merge(bitDepth, n.Label, n.LabelBitLength)
 
+		tryNext := func(next *node.Pointer, nextKey node.Key, parentState visitState) (bool, error) {
+			if err := it.doNext(next, bitLength, newPath, nextKey, visitBefore); err != nil {
+				return false, err
+			}
+			if it.key != nil { // Key has been found
+				it.pos = append(it.pos, pathAtom{state: parentState, ptr: ptr, bitDepth: bitDepth, path: path})
+				return true, nil
+			}
+			return false, nil
+		}
+
 		// Check if the key is longer than the current path but lexicographically smaller. In this
 		// case everything in this subtree will be larger so we need to take the first value.
 		takeFirst := bitLength > 0 && key.BitLength() >= bitLength && key.Compare(newPath) < 0
@@ -275,14 +286,8 @@ func (it *treeIterator) doNext(ptr *node.Pointer, bitDepth node.Depth, path, key
 		// Does lookup key end here? Look into LeafNode.
 		if (state == visitBefore && (keyNotLonger || takeFirst)) || state == visitAt {
 			if state == visitBefore {
-				err := it.doNext(n.LeafNode, bitLength, newPath, key, visitBefore)
-				if err != nil {
+				if ok, err := tryNext(n.LeafNode, key, visitAt); ok || err != nil {
 					return err
-				}
-				if it.key != nil {
-					// Key has been found.
-					it.pos = append(it.pos, pathAtom{state: visitAt, ptr: ptr, bitDepth: bitDepth, path: path})
-					return nil
 				}
 			}
 			// Key has not been found, continue with search for next key.
@@ -298,14 +303,8 @@ func (it *treeIterator) doNext(ptr *node.Pointer, bitDepth node.Depth, path, key
 		// Continue recursively based on a bit value.
 		if (state == visitAt && (!key.GetBit(bitLength) || takeFirst)) || state == visitAtLeft {
 			if state == visitAt {
-				err := it.doNext(n.Left, bitLength, newPath, key, visitBefore)
-				if err != nil {
+				if ok, err := tryNext(n.Left, key, visitAtLeft); ok || err != nil {
 					return err
-				}
-				if it.key != nil {
-					// Key has been found.
-					it.pos = append(it.pos, pathAtom{state: visitAtLeft, ptr: ptr, bitDepth: bitDepth, path: path})
-					return nil
 				}
 			}
 			// Key has not been found, continue with search for next key.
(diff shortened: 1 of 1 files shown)


go/storage/mkvs/iterator.go
```diff
@@ -260,73 +260,63 @@ func (it *treeIterator) doNext(ptr *node.Pointer, bitDepth node.Depth, path, key
...
+		takeFirst := bitLength > 0 && key.BitLength() >= bitLength && key.Compare(newPath) < 0
+		keyNotLonger := key.BitLength() <= bitLength
+
+		switch state {
+		case visitBefore:
+			if keyNotLonger || takeFirst {
+				// Try leaf first, since it is lexicographically smallest.
+				if ok, err := tryNext(n.LeafNode, key, visitAt); ok || err != nil {
 					return err
 				}
```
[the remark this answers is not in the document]
kostko: I like the switch approach, let me look at it some more though.

go/storage/mkvs/iterator.go
```diff
@@ -260,73 +260,63 @@ func (it *treeIterator) doNext(ptr *node.Pointer, bitDepth node.Depth, path, key
...
-				if err != nil {
+		advanceKeyToRight := func(k node.Key) node.Key {
+			k, _ = k.Split(bitLength, k.BitLength())
+			return k.AppendBit(bitLength, true)
+		}
+
+		// Check if the key is longer than the current path but lexicographically smaller. In this
+		// case everything in this subtree will be larger so we need to take the first value.
+		takeFirst := bitLength > 0 && key.BitLength() >= bitLength && key.Compare(newPath) < 0
+		keyNotLonger := key.BitLength() <= bitLength
```
peternose, line 287: I would also unify names of variables and rename few of them, to make code easier to read:
- `keyNotLonger` to `keyLonger`,
- `bitLength` to `newBitLength` or `newDepth`,
- use either `bitLength` or `nodeDepth`, not both, maybe even something shorter like `depth`.

martintomazic: Fixed.

I have kept `keyNotLonger` since we have `key <= other`. `keyLonger` would work for ` key > other`?

martintomazic:
Do we want to also rename `Depth` type [definition](https://github.com/oasisprotocol/oasis-core/blob/06115b95d51123d457c8d5d7b2a5c7fdd11670f6/go/storage/mkvs/node/depth.go#L8) to `BitDepth` possibly `BitLen`? We use `*depth`, `*length` or `keyLen` interchangeably in lots of places so not really sure how to solve it best.

~Finally, is there a reason for `Key` not to be struct type and also store  `bitLength` inside struct e.g.:~

```go
type Key struct {
	val: []byte,
	len: node.Depth,
}
```
~Starting to explore this, as this would allow us to  stop passing bitLen variables...~



^^ Lots of changes for questionable gain.

go/storage/mkvs/iterator.go
```diff
@@ -260,73 +260,63 @@ func (it *treeIterator) doNext(ptr *node.Pointer, bitDepth node.Depth, path, key
...
+			if keyNotLonger {
+				key = key.AppendBit(newBitDepth, false)
+			}
+
+			if !key.GetBit(newBitDepth) || takeFirst {
+				if ok, err := tryNext(n.Left, key, visitAtLeft); ok || err != nil {
+					return err
 				}
+				key = advanceKeyToRight(key)
 			}
```
martintomazic, line 308: Observation: I think this only works because we only `Seek` or `Insert` keys that are at least byte size. This function on the other hand promises bit level precision. E.g. problem could arise if you are searching for a key  with 2 bits only, but your `newPath=3`.  Then `key.AppendBit(newBitDepth, false)` may produce a new 8 bit key.  This is problematic since you may be skipping first bigger key that may be at 4 bits depth, since `ketNotLonger` will not trigger in such case (you are skipping leave nodes).

go/storage/mkvs/iterator.go
```diff
@@ -260,73 +260,63 @@ func (it *treeIterator) doNext(ptr *node.Pointer, bitDepth node.Depth, path, key
...
+		// Check if the key is longer than the current path but lexicographically smaller. In this
+		// case everything in this subtree will be larger so we need to take the first value.
+		takeFirst := newBitDepth > 0 && key.BitLength() >= newBitDepth && key.Compare(newPath) < 0
+		keyNotLonger := key.BitLength() <= newBitDepth
+
+		switch state {
+		case visitBefore:
+			if keyNotLonger || takeFirst {
+				// Try leaf first, since it is lexicographically smallest.
+				if ok, err := tryNext(n.LeafNode, key, visitAt); ok || err != nil {
```
ptrus, line 293: Not sure if this is relevant or not, but i think that in the old code, for this case, we called `it.doNext` with `path`, and now we use the `newPath`.

martintomazic: Correct. As pointed in 57f5d86ee2abc6444a374f7214bc40c4ffd29803, previously we were passing `bitLength` (now  called `newBitDepth`) together with `path` (old one), i.e. vars mismatched.  I guess this inconsistency  had no effect, since the recursive call for leaf node then ignored them anyways (still does).

ptrus: Oh, sorry i mised that. Thx.

go/storage/mkvs/iterator.go
```diff
@@ -260,73 +260,63 @@ func (it *treeIterator) doNext(ptr *node.Pointer, bitDepth node.Depth, path, key
...
-					return nil
-				}
+		tryNext := func(next *node.Pointer, nextKey node.Key, parentState visitState) (bool, error) {
+			if err := it.doNext(next, newBitDepth, newPath, nextKey, visitBefore); err != nil {
+				return false, err
 			}
-			// Key has not been found, continue with search for next key.
-			if key.BitLength() <= bitLength {
-				key = key.AppendBit(bitLength, false)
+			if it.key != nil { // Key has been found
```
kostko:
```suggestion
			if it.key != nil { // Key has been found.
```

kostko (approved):

PR #6179, opened 2025-05-05
ci: Release pre-release versions by default

kostko:


.goreleaser.yml
```diff
@@ -89,3 +89,5 @@ snapshot:
 
 release:
   name_template: "{{.ProjectName}} {{.Version}}"
+  prerelease: true
```
peternose, line 92: Why is this not done with a RC tag?

>  If set to auto, will mark the release as not ready for production
>  in case there is an indicator for this in the tag e.g. v1.0.0-rc1
>  If set to true, will mark the release as not ready for production.
>  Default: false.
>  prerelease: auto

kostko: We want to keep the exact same version and not change the binary in any way.

peternose: Not the standard way, but it's fine.

peternose (approved):

PR #6180, opened 2025-05-06
go/consensus/api: Add method GetBlockResults

peternose:
Method `GetBlockResults` is needed by the light client node to serve events.

If performance degrades due to unmarshalling block results metadata in every `GetEvents` call, we can retain the backend interface and use `GetCometBFTBlockResults` if implemented, falling back to `GetBlockResults` otherwise.


kostko (approved):
Would be good to open an issue to benchmark the difference (if any) before release.

kostko (approved):

PR #6194, opened 2025-05-19
go/common/sgx/pcs: Add support for specific TCB data evaluation numbers

### LINKED ISSUE #6121: pcs: Support new tcbEvaluationDataNumber parameter
Intel PCS API recently introduced a new `tcbEvaluationDataNumber` parameter when retrieving TCB info structures. This is in addition to the already supported `update=early` parameter and allows retrieval of TCB infos with a specific TCB evaluation data number value (e.g. as required by the policy).

We should support this in the PCS client and the SGX/TDX attestation flows such that in case `update=early` is not suitable (e.g. fails for the local platform), it should attempt to fetch specific TCB evaluation data number values (in descending order).

For details see the [updated PCS API documentation](https://api.portal.trustedservices.intel.com/content/documentation.html).

kostko:
Fixes #6121

### DIFF
--- a/go/common/sgx/pcs/tcb.go
+++ b/go/common/sgx/pcs/tcb.go
@@ -742,3 +742,25 @@ type EnclaveTCBLevel struct {
 	Status      TCBStatus `json:"tcbStatus"`
 	AdvisoryIDs []string  `json:"advisoryIDs"`
 }
+
+// SignedTCBEvaluationDataNumbers is the signed TCB evaluation data numbers response body.
+type SignedTCBEvaluationDataNumbers struct {
+	Numbers   TCBEvaluationDataNumbers `json:"tcbEvaluationDataNumbers"`
+	Signature string                   `json:"signature"`
+}
+
+// TCBEvaluationDataNumbers is the TCB evaluation data numbers body.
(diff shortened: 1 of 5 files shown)


go/common/sgx/pcs/tcb.go
```diff
@@ -742,3 +742,25 @@ type EnclaveTCBLevel struct {
 	Status      TCBStatus `json:"tcbStatus"`
 	AdvisoryIDs []string  `json:"advisoryIDs"`
 }
+
+// SignedTCBEvaluationDataNumbers is the signed TCB evaluation data numbers response body.
+type SignedTCBEvaluationDataNumbers struct {
+	Numbers   TCBEvaluationDataNumbers `json:"tcbEvaluationDataNumbers"`
+	Signature string                   `json:"signature"`
```
peternose, line 749: Should this signature be verified?

kostko: I don't see a need for it in this case, this is being done in an untrusted environment (host node) and we just need a list of numbers to request from the same service. Verifying the signature requires a bunch of extra logic for no clear benefit.

peternose (approved):
Looks good.

PR #6197, opened 2025-05-20
go/runtime: Add log manager

### LINKED ISSUE #6064: Add endpoints to stream logs from a given runtime component
There should be an easy way to stream logs from a given runtime component via internal gRPC. This should also be exposed to apps via the bundle manager RHP API.

kostko:
Fixes #6064


go/runtime/log/directories.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"path/filepath"
+
+	"github.com/oasisprotocol/oasis-core/go/runtime/config"
+)
+
+// LogsDir is the name of the directory located inside the node's runtimes directory which contains
+// the component logs.
+const LogsDir = "logs"
```
ptrus, line 11: Does this constant need to be exported?

go/runtime/log/log.go
```diff
@@ -0,0 +1,126 @@
...
+	return l.file.Close()
+}
+
+// Watch starts watching the log file for changes, returning a channel that receives updates.
+//
+// The passed context is used to stop the watcher.
+func (l *Log) Watch(ctx context.Context, follow bool) (<-chan string, error) {
+	watcher, err := tail.TailFile(l.file.Name(), tail.Config{
+		ReOpen:        true,
+		Follow:        follow,
```
ptrus, line 93: Doesn't `TailFile` require that `Follow` is true if `ReOpen` is true?

.changelog/6197.feature.md
```diff
@@ -0,0 +1,4 @@
+Add log manager
```
ptrus, line 1:
```suggestion
runtimes: Add log manager
```

or something alike, to add context that this is for runtimes.

go/common/logging/multi.go
```diff
@@ -0,0 +1,39 @@
...
+
+func newMultiLogger(loggers []log.Logger) *multiLogger {
+	return &multiLogger{
+		loggers: loggers,
+	}
+}
+
+func (l *multiLogger) Log(keyvals ...any) error {
+	for _, logger := range l.loggers {
+		if logger == nil {
```
ptrus, line 19: Is there a use case for this being nil, or would it be reasonable to expect that this should not be nil? (unless a developer error), and should we emit this check?

go/runtime/host/logger.go
```diff
@@ -22,7 +23,9 @@ const maxLogBufferSize = 10_000
 // It hardcodes some assumptions about the format of the runtime logs.
 type RuntimeLogWrapper struct {
 	// Logger for wrapper-internal info/errors.
-	logger *logging.Logger
+	internalLogger *logging.Logger
+	// Runtime-specific logger that should be used as an additional log sink.
+	runtimeLogger log.Logger
```
ptrus, line 28: Feels a bit weird mixing two different loggers (go-kit and the oasis-core package), but I guess you used it due to the simpler interface?

go/runtime/host/logger.go
```diff
@@ -129,7 +137,38 @@ func (w RuntimeLogWrapper) processLogLine(line []byte) {
 	case "ERRO":
 		rtLogger.Error(msg, kv...)
 	default:
-		w.logger.Warn("log line from runtime has no known error level set, using INFO", "log_line", string(line))
+		w.internalLogger.Warn("log line from runtime has no known error level set, using INFO", "log_line", string(line))
 		rtLogger.Info(msg, kv...)
 	}
 }
+
+// filterLogger is a logger wrapper that filters out specific keys.
```
ptrus, line 145: Maybe mention the reasoning or the types of keys that are filtered out. Could also rename it to something a bit less general. Because without looking at the code, based on the current name, I expected it to accept the field names to be filtered out.

go/runtime/log/log.go
kostko: You are right. I need to add some tests ;)

go/runtime/host/logger.go
kostko: Yeah I agree. I'll check if this could be refactored so that we would not need the external interface.

go/runtime/host/logger.go
kostko: Ah yeah, this is a private filter only used for this specific use case. Will update.

go/runtime/host/logger.go
kostko: Changed, please take another look.

go/runtime/host/logger.go
kostko: Made it more reusable and moved to the logging package.

go/common/logging/multi.go
kostko: Changed.

go/common/logging/filter.go
```diff
@@ -0,0 +1,48 @@
...
+			return fmt.Errorf("missing value for key '%s'", keyvals[i])
+		}
+		filteredKvs = append(filteredKvs, keyvals[i], keyvals[i+1])
+	}
+
+	return l.logger.Log(filteredKvs...)
+}
+
+// NewFilterLogger creates a logger wrapper that filters out specific keys.
+func NewFilterLogger(base *Logger, excludeFields map[string]struct{}) *Logger {
```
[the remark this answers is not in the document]
kostko: But the returned value is already a struct?

go/common/logging/filter.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"fmt"
+
+	"github.com/go-kit/log"
+)
+
+// filterLogger is a logger wrapper that filters out specific keys.
+type filterLogger struct {
+	logger        log.Logger
+	excludeFields map[string]struct{}
```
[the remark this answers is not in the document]
kostko: Good idea.

go/common/logging/filter.go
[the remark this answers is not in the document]
peternose: Ups, missed that. It still feels strange that we are creating `FilterLogger` but returning `Logger`.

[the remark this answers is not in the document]
kostko: But there are no interfaces for loggers so this would not work. I do not want to refactor the entire logging system. Previously this returned the low-level logger interface but based on comment from @ptrus this was changed to not expose internals.

go/runtime/log/manager.go
```diff
@@ -0,0 +1,104 @@
...
+	delete(m.logs[runtimeID], componentID)
+	if len(m.logs[runtimeID]) == 0 {
+		delete(m.logs, runtimeID)
+	}
+}
+
+func (m *Manager) getLogDirectory(runtimeID common.Namespace, componentID component.ID) string {
+	rawRuntimeID, _ := runtimeID.MarshalText()
+	rawComponentID, _ := componentID.MarshalText()
+	logDir := filepath.Join(m.logsDir, string(rawRuntimeID), string(rawComponentID))
```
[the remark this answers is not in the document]
kostko: Dots and slashes are not valid in component names (only: `[a-zA-Z0-9_-]+`), but yeah we should do some defense in depth.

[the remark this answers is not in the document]
kostko: Using `os.Root` is somewhat awkward currently as it doesn't have the `RemoveAll` and/or `MkdirAll` functions though.

go/runtime/log/log.go
```diff
@@ -0,0 +1,141 @@
...
+func (l *Log) Close() error {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	defer close(l.closeCh)
+	return l.file.Close()
+}
+
+// Watch starts watching the log file for changes, returning a channel that receives updates.
+//
+// The passed context is used to stop the watcher.
```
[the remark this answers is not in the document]
kostko: Makes sense, just that the two EOF cases (`closeCh` and `!ok`) should not result in an error, but normal return.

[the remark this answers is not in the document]
kostko: But this does make it somewhat more awkward for the caller...

go/runtime/host/logger.go
```diff
@@ -32,11 +34,12 @@ type RuntimeLogWrapper struct {
 }
 
 // NewRuntimeLogWrapper creates a new RuntimeLogWrapper.
-func NewRuntimeLogWrapper(logger *logging.Logger, suffixes ...any) *RuntimeLogWrapper {
+func NewRuntimeLogWrapper(internalLogger *logging.Logger, runtimeLogger *logging.Logger, suffixes ...any) *RuntimeLogWrapper {
```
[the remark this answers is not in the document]
kostko: There is also a nop logger that you can pass if you don't want to pass `nil`. Currently the only place we pass `nil` is in tests.

go/runtime/host/logger.go
ptrus: Thanks, looks good to me now 👍

go/runtime/host/logger.go
ptrus: 👍

go/runtime/log/manager.go
```diff
@@ -0,0 +1,104 @@
...
+	"sync"
+
+	"github.com/oasisprotocol/oasis-core/go/common"
+	"github.com/oasisprotocol/oasis-core/go/common/logging"
+	"github.com/oasisprotocol/oasis-core/go/config"
+	"github.com/oasisprotocol/oasis-core/go/runtime/bundle/component"
+)
+
+// logFn is the filename of the log file.
+const logFn = "log"
```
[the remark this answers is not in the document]
peternose: I think we are using `fn` in our codebase for filenames (only inside functions/methods), and suffix `Fn` for functions (e.g. `filterFn, ctorFn, counterIncFn`).

go/runtime/log/manager.go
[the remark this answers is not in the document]
peternose: True, it's awkward. We can leave it as it is.

go/runtime/log/log_test.go
```diff
@@ -0,0 +1,45 @@
...
+	ch := make(chan string)
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for line := range ch {
+			lines = append(lines, line)
+		}
+	}()
+
+	err = log.Watch(context.Background(), ch, false)
```
peternose, line 37: Use timeout of a second to avoid hanging.

go/runtime/registry/host_rofl_mgmt.go
```diff
@@ -87,6 +90,21 @@ func (rh *roflHostHandler) handleVolumeManagement(rq *enclaverpc.Request) (any,
...
+// handleLogManagement handles log management local RPCs.
+func (rh *roflHostHandler) handleLogManagement(ctx context.Context, rq *enclaverpc.Request) (any, error) {
+	switch rq.Method {
+	case rofl.MethodLogGet:
+		// Get logs.
+		var args rofl.LogGetRequest
+		if err := cbor.Unmarshal(rq.Args, &args); err != nil {
+			return nil, err
+		}
+		return rh.handleLogGet(ctx, &args)
```
peternose, line 102: How will runtime call this endpoint, as currently this method returns all lines, and not only new lines from the last call?

go/runtime/log/log_test.go
```diff
@@ -0,0 +1,45 @@
...
+	require := require.New(t)
+
+	tmpDir := t.TempDir()
+	logFn := filepath.Join(tmpDir, "log")
+	log, err := newLog(logFn, 1024)
+	require.NoError(err, "newLog")
+	defer log.Close()
+
+	logger := log.Logger()
+	logger.Info("hello log world")
```
peternose, line 22:
```suggestion
	logger.Info("info message")
	logger.Warn("warn message")
	logger.Error("error message")

	expectedOutput := []string{
		`{"level":"info","msg":"info message"}`,
		`{"level":"warn","msg":"warn message"}`,
		`{"level":"error","msg":"error message"}`,
	}	
```

go/runtime/registry/host_rofl_mgmt.go
kostko: Added a way to seek by timestamp.

go/runtime/registry/host_rofl_mgmt.go
```diff
@@ -246,6 +264,66 @@ func (rh *roflHostHandler) handleVolumeList(rq *rofl.VolumeListRequest) (*rofl.V
...
+		return nil, fmt.Errorf("component not found")
+	}
+
+	log, err := rh.getLogManager().Get(rh.parent.runtime.ID(), requestedComponentID)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: Support streaming logs.
+	var (
```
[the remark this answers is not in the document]
peternose: I'm missing something, as I think this should work. Once `Watch` finishes with or without an error, the channel is closed (`defer close(ch)`), the go routine stops writing because of the closed channel, `Read` function waits for go routine to exit (`defer wg.Wait()`), and in the end, all lines are returned.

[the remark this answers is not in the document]
kostko: AFAIK defer would run after the return expression.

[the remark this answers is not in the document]
peternose: Oh, crap. You are right. It would work with named returned values, but that is uglier.

peternose (approved):

Issue #5752, opened 2024-07-03
Optimize checkpoint creation


kostko:
Currently checkpoint creation can be rather slow and we should optimize it, especially as the state grows.

martintomazic:
# Optimize Checkpoint Creation Design

## Context

_Checkpoint_ is state dump at a given _height_, that facilitates [state sync](https://docs.oasis.io/node/run-your-node/advanced/sync-node-using-state-sync/). It consists of many _chunks_ to enable parallel and partial downloads from the p2p network. Chunks are serialized and compressed merkle proofs, which prevent malicious nodes spamming with invalid data.

Creation is currently optional (altruistic). To achieve deterministic output, yet some level of configuration (e.g. chunks size), some parameters are stored as part of runtime/consensus policy.

## Motivation
Creating a checkpoint for latest height of _Sapphire Mainnet_, with state pruning enabled is done in a matter of minutes.  If no pruning is set, checkpoint creation becomes a matter of hours (bigger overall state -> slower reads).

_Creation needs to be fast and optimal, so that we can enable it on bulkier nodes, with possible smaller creation interval._


## Requirements

Checkpoint creation should be:
* R1: `Ndb` implementation agnostic.
* R2: Deterministic.
* R3: Partial and provable, to facilitate fetching from untrusted p2p network.
* R4: Interoperable, i.e. network may consists of peers broadcasting different versions at the same time.
    - Mixing chunks from different versions is of course not allowed.
* R5: Introducing new version should not introduce availability issues/overload:
    - If we have only few peers per checkpoint already, further splitting it may be problematic.
  
## Existing Implementation
Currently, chunks are created by [iterating over keyset](https://github.com/oasisprotocol/oasis-core/blob/5aa6d1e41093cfbfbebd69669ee49da0b315472e/go/storage/mkvs/checkpoint/chunk.go#L19) for the given height. As the iterator's `proofbuilder` exceeds the `chunk_size` limit, a chunk is written into file. Next, we create new iterator, starting with the previous `offset` and repeat procedure until no keys are left.

Internally, this iteration triggers in-order traversal of the state trie, resulting in repeated calls to Badger's [`txn.Get`](https://pkg.go.dev/github.com/dgraph-io/badger#Txn.Get).


## Benchmarks and Profiling

### Reproducibility
CPU: 12th Gen Intel i7-12800H (20) @ 4.700GHz
GPU: Intel Alder Lake-P
Memory:  31760MiB
[...]

PR #6164, opened 2025-04-23
go/storage/mkvs: Prefetch nodes during Tree iteration

martintomazic:
**DRAFT**

**Context:**
Part of #5752.

**Possible Optimization:**
By prefetching nodes during checkpoint creation (order is predictable...), we could save on transaction overhead.

I expect this optimization to be more pronounced with pathbadger, since due to higher locality of data, I would expect more cache hits, making transaction overhead optimization more visible.

**Result**
https://github.com/oasisprotocol/oasis-core/pull/6164#issuecomment-2929863970

PR #6204, opened 2025-05-27
go/storage/mkvs: Parallelize checkpoint creation

### LINKED ISSUE #6218: Introduce v2 checkpointing algorithm (parallel)

martintomazic:
Closes #6218 

For context please read [design doc](https://github.com/oasisprotocol/oasis-core/issues/5752#issuecomment-2941540845). To see where this PR fits in, see _Implementation and Validation Plan/2_.

For the discussion about need to release new version read the https://github.com/oasisprotocol/oasis-core/pull/6204#issuecomment-2964592470 and replies.

### DIFF
--- a/go/oasis-test-runner/scenario/e2e/runtime/scenario.go
+++ b/go/oasis-test-runner/scenario/e2e/runtime/scenario.go
@@ -264,7 +264,13 @@ func (sc *Scenario) Fixture() (*oasis.NetworkFixture, error) {
 					},
 				},
 			},
-			{RuntimeProvisioner: runtimeProvisioner, Entity: 1, Runtimes: []int{1}},
+			{
+				RuntimeProvisioner: runtimeProvisioner,
+				Entity:             1,
+				Runtimes:           []int{1},
+				// Use a parallel chunking algorithm for the second compute node to show interoperability.
+				CheckpointParallelChunker: true,
+			},
 		},
 		Sentries: []oasis.SentryFixture{},
 		Seeds:    []oasis.SeedFixture{{}},
--- a/go/oasis-test-runner/oasis/compute.go
+++ b/go/oasis-test-runner/oasis/compute.go
@@ -37,10 +37,11 @@ type Compute struct { // nolint: maligned
 
 	sentryIndices []int
 
-	storageBackend          string
-	disablePublicRPC        bool
-	checkpointSyncDisabled  bool
-	checkpointCheckInterval time.Duration
+	storageBackend            string
+	disablePublicRPC          bool
+	checkpointSyncDisabled    bool
+	checkpointCheckInterval   time.Duration
+	checkpointParallelChunker bool
 
 	sentryPubKey  signature.PublicKey
 	consensusPort uint16
@@ -62,10 +63,11 @@ type ComputeCfg struct {
 
 	SentryIndices []int
 
-	StorageBackend          string
-	DisablePublicRPC        bool
-	CheckpointSyncDisabled  bool
-	CheckpointCheckInterval time.Duration
+	StorageBackend            string
+	DisablePublicRPC          bool
+	CheckpointSyncDisabled    bool
+	CheckpointCheckInterval   time.Duration
+	CheckpointParallelChunker bool
 }
 
 // UpdateRuntimes updates the worker node runtimes.
@@ -172,6 +174,7 @@ func (worker *Compute) ModifyConfig() error {
 	worker.Config.Storage.CheckpointSyncDisabled = worker.checkpointSyncDisabled
 	worker.Config.Storage.Checkpointer.Enabled = true
 	worker.Config.Storage.Checkpointer.CheckInterval = worker.checkpointCheckInterval
+	worker.Config.Storage.Checkpointer.ParallelChunker = worker.checkpointParallelChunker
 
 	// Sentry configuration.
 	sentries, err := resolveSentries(worker.net, worker.sentryIndices)
@@ -230,18 +233,19 @@ func (net *Network) NewCompute(cfg *ComputeCfg) (*Compute, error) {
 	}
 
 	worker := &Compute{
-		Node:                    host,
-		storageBackend:          cfg.StorageBackend,
-		sentryIndices:           cfg.SentryIndices,
-		disablePublicRPC:        cfg.DisablePublicRPC,
-		checkpointSyncDisabled:  cfg.CheckpointSyncDisabled,
-		checkpointCheckInterval: cfg.CheckpointCheckInterval,
-		sentryPubKey:            sentryPubKey,
-		runtimeProvisioner:      cfg.RuntimeProvisioner,
-		consensusPort:           host.getProvisionedPort(nodePortConsensus),
-		p2pPort:                 host.getProvisionedPort(nodePortP2P),
-		runtimes:                cfg.Runtimes,
-		runtimeConfig:           cfg.RuntimeConfig,
+		Node:                      host,
+		storageBackend:            cfg.StorageBackend,
+		sentryIndices:             cfg.SentryIndices,
+		disablePublicRPC:          cfg.DisablePublicRPC,
+		checkpointSyncDisabled:    cfg.CheckpointSyncDisabled,
+		checkpointCheckInterval:   cfg.CheckpointCheckInterval,
+		checkpointParallelChunker: cfg.CheckpointParallelChunker,
+		sentryPubKey:              sentryPubKey,
+		runtimeProvisioner:        cfg.RuntimeProvisioner,
+		consensusPort:             host.getProvisionedPort(nodePortConsensus),
+		p2pPort:                   host.getProvisionedPort(nodePortP2P),
+		runtimes:                  cfg.Runtimes,
+		runtimeConfig:             cfg.RuntimeConfig,
 	}
 
 	// Remove any exploded bundles on cleanup.
--- a/go/oasis-test-runner/oasis/fixture.go
+++ b/go/oasis-test-runner/oasis/fixture.go
@@ -401,8 +401,9 @@ type ComputeWorkerFixture struct {
 
 	Sentries []int `json:"sentries,omitempty"`
 
-	CheckpointCheckInterval time.Duration `json:"checkpoint_check_interval,omitempty"`
-	CheckpointSyncEnabled   bool          `json:"checkpoint_sync_enabled,omitempty"`
+	CheckpointCheckInterval   time.Duration `json:"checkpoint_check_interval,omitempty"`
+	CheckpointSyncEnabled     bool          `json:"checkpoint_sync_enabled,omitempty"`
+	CheckpointParallelChunker bool          `json:"checkpoint_parallel_chunker,omitempty"`
 
 	// Runtimes contains the indexes of the runtimes to enable.
 	Runtimes []int `json:"runtimes,omitempty"`
@@ -434,10 +435,11 @@ func (f *ComputeWorkerFixture) Create(net *Network) (*Compute, error) {
 			Entity:                      entity,
 			ExtraArgs:                   f.ExtraArgs,
 		},
-		RuntimeProvisioner:      f.RuntimeProvisioner,
-		StorageBackend:          f.StorageBackend,
-		SentryIndices:           f.Sentries,
-		CheckpointCheckInterval: f.CheckpointCheckInterval,
+		RuntimeProvisioner:        f.RuntimeProvisioner,
+		StorageBackend:            f.StorageBackend,
+		SentryIndices:             f.Sentries,
+		CheckpointCheckInterval:   f.CheckpointCheckInterval,
+		CheckpointParallelChunker: f.CheckpointParallelChunker,
 		// The checkpoint syncing flag is intentionally flipped here.
 		// Syncing should normally be enabled, but normally disabled in tests.
 		CheckpointSyncDisabled: !f.CheckpointSyncEnabled,
--- a/go/oasis-test-runner/scenario/e2e/runtime/txsource.go
+++ b/go/oasis-test-runner/scenario/e2e/runtime/txsource.go
@@ -477,8 +477,9 @@ func (sc *txSourceImpl) Fixture() (*oasis.NetworkFixture, error) {
 	var computeWorkers []oasis.ComputeWorkerFixture
 	for i := 0; i < sc.numComputeNodes; i++ {
 		computeWorkers = append(computeWorkers, oasis.ComputeWorkerFixture{
-			Entity:   1,
-			Runtimes: []int{1},
+			Entity:                    1,
+			Runtimes:                  []int{1},
+			CheckpointParallelChunker: i%2 == 0, // use a mix of new and old checkpointing algorithm
 		})
 	}
 	f.ComputeWorkers = computeWorkers


martintomazic:
Same benchmark as run by https://github.com/oasisprotocol/oasis-core/pull/6165#issuecomment-2889259677 was around around as fast, i.e. `5x` faster then master:
**Pros**:
* Less partially filled chunks, because previous algorithm has lot's of single sized subtrees at the beginning.  
* Previous algorithm works only with high `depth`, in order to go deep enough into EVM part (most state). This one works just fine with smaller parallelization. 
* Likely we could improve this algorithm further, making it a desired solution. 

**Cons**:
* Maybe slightly more complex.


**Possible improvements:**
* >Wait for all subtrees to create one chunk,  so that final output is stable.
    * **Is there alternative?** We have to limit parallelization, and have deterministic output at the end...
*  Maintaining enough threads is quite naive:
     *   Whenever less subtree tasks then expected, try to split task (at most two children).
     *  Tasks for `threads=30` (capped by 60 then): `1, 2, 1, 1, 2, 4, 8, 12, 22, 26, 30, 27, 24, 21, 22, 26, 37, 56, 44, 58, 55, 49, 44, 44, 48, 40, 25, 17, 4`. Ideally it would be more like `30, 30, 30, ...`

**Update**: this is now exact,  potentially we see 10% speed-up but hard to say since this seems like standard deviation.

martintomazic:
> **Possible improvements:**
> Wait for all subtrees to create one chunk, so that final output is stable.

Benched a "non-deterministic" version (thread limited, but no explicit coordination/waiting). The results are the same, i.e. we are capped at ~`5x` speed-up.

We observed this in two different algorithms, meaning we are likely hitting nodedb and or  current badgerdb configuration read limitation.

> Maintaining enough threads is quite naive

Doesn't seem to make much difference, if the code is simpler will keep it.

**Next steps:**

- [x]  Explore reading nodes in batch, to avoid transaction overhead (~I expect additional `2x` here~):
    * See [results](https://github.com/oasisprotocol/oasis-core/pull/6164#issuecomment-2929863970) (no performance gain).
- [x] Explore using `v1` proofs instead of `v0` (we may be able to reduce checkpoint size, no speed-up expected):
    * I replaced `v1` proofs in the [current sequential code](https://github.com/oasisprotocol/oasis-core/blob/0857a45f126d96cbe1f9558cd91d8988f99b33f5/go/storage/mkvs/checkpoint/chunk.go#L34). The difference seems to be `3.8KB` less for the checkpoint of size `4.1GB`.
    * Would prefer continue using `v0` proofs since this is negligible and code stays simpler. Parallel version should behave the same?
- [x] Final clean-up of this PR once above is checked (1-2 day max for all).
- [x] ~Reorganize `checkpoint` package, test suite for both versions~, add regression tests.

As a follow-up, we may want to explore if it is possible to optimize above mentioned db read bottleneck:

martintomazic:
>Explore reading nodes in batch, to avoid transaction overhead 

As from the https://github.com/oasisprotocol/oasis-core/pull/6164#issuecomment-2929863970, there is no performance gain.

go/storage/mkvs/checkpoint/file.go
```diff
@@ -7,37 +7,39 @@ import (
...
+type chunker interface {
+	createChunks(ctx context.Context, root node.Root, chunksDir string, chunkSize uint64) ([]hash.Hash, error)
+	Version() uint16
 }
 
-func (fc *fileCreator) CreateCheckpoint(ctx context.Context, root node.Root, chunkSize uint64) (meta *Metadata, err error) {
-	tree := mkvs.NewWithRoot(nil, fc.ndb, root)
-	defer tree.Close()
-
+func createCheckpoint(ctx context.Context, ndb db.NodeDB, root node.Root, dataDir string, chunkSize uint64, chunker chunker) (meta *Metadata, err error) {
```
martintomazic: This is a bit of an anti-pattern, that will have to be fixed as part of `checkpoint` package reorg.

As one can see both `Creator` implementations differ in only this logic. Moreover, as the name suggests they should probably not be responsible for serving checkpoints/chunks. As such storage, and bookkeeping should be propagated upwards to `Checkpointer`?

martintomazic:
After exploring #6216 and #6219 I believe we don't want (need) to release new version.

Two chunking algorithms indeed produce different checkpoint hash for the given root, however chunk format as well as restoration logic is still the same. Moreover, consensus checkpoint sync (see [snapshots.go](https://github.com/oasisprotocol/oasis-core/blob/master/go/consensus/cometbft/abci/snapshots.go), cometbft [syncer.go](https://github.com/cometbft/cometbft/blob/f69352842a56f3d3e2c3df27d7586f4da9d57c56/statesync/syncer.go)), and runtime's [checkpoint sync](https://github.com/oasisprotocol/oasis-core/blob/master/go/worker/storage/committee/checkpoint_sync.go), work perfectly fine with multiple checkpoints per root.

Added some minor tweaks to above and made checkpointing algorithm configurable. This way we can upgrade incrementally, possibly starting with some test nodes, then default to new algorithm for the runtimes during next release. Consensus we may or may not upgrade at the end, given that checkpointing is already fast.. 

Contrary to original idea I am starting to think #5752 could be closed in a single PR.

Next steps:
- [x] Explain above in more details, possibly short doc on the issue and update context of this PR.
- [x] Final clean-up: Improve/add missing comments, simplify some code if possible.
- [x] Regression, e2e, extend unit tests for new corner cases?

_Definitely this would benefit for preliminary review/high level feedback._

kostko:
> however chunk format as well as restoration logic is still the same

Does that mean that even old clients can restore new checkpoints? Or will they encounter any issues?

martintomazic:
>Does that mean that even old clients can restore new checkpoints? Or will they encounter any issues?

They can, without issues.

Let's examine a hypothetical situation for the Sapphire state sync:
1. Old client [requests](https://github.com/oasisprotocol/oasis-core/blob/7619fdd1d5e2ea6af4d2fd89f8a2e72624f7728e/go/worker/storage/p2p/sync/client.go#L69) checkpoint list from it's peerset.
2. It may receive a mix of new and old checkpoints, sorting them by height at the  ([callsite](https://github.com/oasisprotocol/oasis-core/blob/7619fdd1d5e2ea6af4d2fd89f8a2e72624f7728e/go/worker/storage/committee/checkpoint_sync.go#L278C16-L278C33)).
3. We have two roots per height so we expect at least two checkpoints per height.
     * Existing peerset may also return byzantine checkpoints. 
4. We start restoring from latest height [here](https://github.com/oasisprotocol/oasis-core/blob/7619fdd1d5e2ea6af4d2fd89f8a2e72624f7728e/go/worker/storage/committee/checkpoint_sync.go#L381), keeping track of [remaining](https://github.com/oasisprotocol/oasis-core/blob/7619fdd1d5e2ea6af4d2fd89f8a2e72624f7728e/go/worker/storage/committee/checkpoint_sync.go#L399) roots. 
5. If checkpoint for height is invalid we proceed to next checkpoint, until all roots are filled.
    * If we fail, we start with lower height. Finally, if we run out of heights state sync fails.
6.  **Observe**: _It is completely valid to e.g. restore state root from the new checkpointing algorithm and IO root from the old one._ 
   
Consensus state sync: [see](https://github.com/cometbft/cometbft/blob/f69352842a56f3d3e2c3df27d7586f4da9d57c56/statesync/syncer.go#L145).

martintomazic:
A bit of a nitpicking/corner case exploration not directly related to above:
1. We are not punishing  byzantine peers (e0a5dfbf12590ef15455517129e0e35e8f609318) during consensus state sync.
     * Cometbft [handling](https://github.com/cometbft/cometbft/blob/f69352842a56f3d3e2c3df27d7586f4da9d57c56/statesync/snapshots.go#L212) should probably also remove all checkpoints originating only from byzantine author.
          * Rn consensus state sync is fast, also checkpoints are not ordered by hash, so not critical. + out of our control.
     *  For runtimes this is more problematic,  [see handling](https://github.com/oasisprotocol/oasis-core/blob/7619fdd1d5e2ea6af4d2fd89f8a2e72624f7728e/go/worker/storage/committee/checkpoint_sync.go#L140), since checkpoint restoration takes 10+ min and order is [predictable](https://github.com/oasisprotocol/oasis-core/blob/9b8f8fda13822f31ea78d901615fd3d3aca4df5e/go/worker/storage/committee/checkpoint_sync.go#L305) (why?) so we could mine them. 
2. ~My understanding both our as well as cometbft p2p network is point to point and not gossip-sub like.~
    * Had we released new version for the new checkpoints, cometbft [orders](https://github.com/cometbft/cometbft/blob/f69352842a56f3d3e2c3df27d7586f4da9d57c56/statesync/snapshots.go#L176) checkpoints by format, before availability.
         *  This is not ideal for the early upgraders, since initially all clients are first redirected to them.
    * For our runtime state sync, I have sorted checkpoints by availability also 9b8f8fda13822f31ea78d901615fd3d3aca4df5e.
        * Cons is if we want to first test it in production with single node, likely nobody will be fetching our checkpoints (although old client would not have this optimization so probably doesn't hold.)

go/storage/mkvs/checkpoint/file.go
```diff
@@ -62,13 +65,22 @@ func (fc *fileCreator) CreateCheckpoint(ctx context.Context, root node.Root, chu
...
+		// Deprecated.
+		ch = &seqChunker{ndb: fc.ndb, root: root, chunkSize: chunkSize}
+	}
+	chunks, err := ch.chunk(ctx, fp)
 	if err != nil {
-		return nil, fmt.Errorf("checkpoint: failed to create chunks: %w", err)
+		return nil, fmt.Errorf("checkpoint: failed to create chunks (chunker threads: %d): %w", chunkerThreads, err)
 	}
 
 	meta = &Metadata{
```
martintomazic: NIT:

`Creator` (`file.go`) currently does two things:
1. The part of the code marked in this comment
2. Registry/manager functionality

I would say `Creator` should have single method that complements `Restorer`:
```go
CreateCheckpoint(ctx context.Context, root node.Root, chunkSize uint64, chunkerThreads uint16) (*Metadata, error)
``` 

Remaining part of this file should probably be extracted to smtg  like `registry.go`, possibly this could be even moved outside of the `mkvs` package.

If there will be no need to release new version, I think we can leave it as is for now...

go/storage/mkvs/checkpoint/subtree.go
```diff
@@ -0,0 +1,239 @@
...
+
+	return writeChunk(proof, w)
+}
+
+func (ct *subtree) trim() {
+	// Remove all trailing nodes whose right subtree state is pending/being explored.
+	for len(ct.pending) > 0 && ct.pending[len(ct.pending)-1].visitState == visitAtRight {
+		ct.pending = ct.pending[:len(ct.pending)-1]
+	}
+}
```
martintomazic: Ideally, running old sequential algorithm (`chunkerThreads=0`) and new one with `chunkerThreads=1` (also sequential), should produce the same result.

Currently, this is not the case, since trimming is quite naive here. In the previous version, we got this check via iterator's `next` method. [see](https://github.com/oasisprotocol/oasis-core/blob/a6123ed3337e5a4c03d43eca6be8b0b174c7aebd/go/storage/mkvs/checkpoint/file.go#L92).

go/storage/mkvs/checkpoint/subtree.go
```diff
@@ -0,0 +1,239 @@
...
+	visitAtRight
+)
+
+type pathAtom struct {
+	nd         node.Node
+	visitState visitState
+}
+
+// subtree is a subtree that is being chunked.
+type subtree struct {
```
martintomazic, line 34: NIT:

Should we create nested `task` package and only expose methods that the client of this abstraction should be using:
1. `NewSubtree` 
2. `NextChunk`
3. `Split`
4. `IsFinished`

go/worker/storage/config/config.go
```diff
@@ -31,6 +31,11 @@ type CheckpointerConfig struct {
 	Enabled bool `yaml:"enabled"`
 	// Storage checkpointer check interval.
 	CheckInterval time.Duration `yaml:"check_interval"`
+	// ChunkerSubtreesCount specifies the target number of parallel subtree chunkers
+	// used during checkpoint creation.
+	//
+	// Setting it to 0, will use old sequential chunker.
+	ChunkerSubtreesCount uint16 `yaml:"chunker_subtrees_count"`
 }
```
martintomazic: Probably we should add a warning this should not be arbitrary configured...

go/consensus/cometbft/abci/snapshots.go
```diff
@@ -212,7 +212,11 @@ func (mux *abciMux) ApplySnapshotChunk(req types.RequestApplySnapshotChunk) type
 			"err", err,
 		)
 
-		return types.ResponseApplySnapshotChunk{Result: types.ResponseApplySnapshotChunk_REJECT_SNAPSHOT}
+		// Given that snapshot was invalid, the sender must have been malicious.
+		return types.ResponseApplySnapshotChunk{
+			Result:        types.ResponseApplySnapshotChunk_REJECT_SNAPSHOT,
+			RejectSenders: []string{req.Sender},
+		}
```
martintomazic, line 219: Tentative. Opinions?

martintomazic (commented):
The PR should be ready for first round of reviews.

go/consensus/cometbft/abci/snapshots.go
kostko: Makes sense to me :+1:

go/consensus/cometbft/config/config.go
```diff
@@ -115,6 +115,11 @@ type CheckpointerConfig struct {
 	Disabled bool `yaml:"disabled"`
 	// ABCI state checkpointer check interval.
 	CheckInterval time.Duration `yaml:"check_interval"`
+	// ChunkerSubtreesCount specifies the target number of parallel subtree chunkers
+	// used during checkpoint creation.
+	//
+	// Setting it to 0, will use the old sequential chunker.
+	ChunkerSubtreesCount uint16 `yaml:"chunker_subtrees_count"`
```
kostko: Should this be named something more high-level like "chunker parallelization" or "chunker threads" or similar?

go/storage/mkvs/checkpoint/checkpoint_test.go
```diff
@@ -491,3 +492,269 @@ func testPruneGapAfterCheckpointRestore(t *testing.T, factory dbApi.Factory) {
 	err = ndb2.Prune(checkpointRootVersion)
 	require.NoError(err, "Prune(%d)", checkpointRootVersion)
 }
+
+// FuzzCreateRestore tests checkpoint creation and restoration.
```
kostko, line 553: Nice! How long did you run the fuzzers for?

go/storage/mkvs/checkpoint/chunk.go
```diff
@@ -16,65 +18,213 @@ import (
...
+	if it.Valid() {
+		it.Next()
+		nextOffset = it.Key()
+	}
+
+	chunkHash, err = writeChunk(proof, w)
+	return
+}
+
+type parallChunker struct {
```
kostko:
```suggestion
type parallelChunker struct {
```

go/storage/mkvs/checkpoint/chunk.go
```diff
@@ -16,65 +18,213 @@ import (
...
-	root node.Root,
-	offset node.Key,
-	chunkSize uint64,
-	w io.Writer,
-) (
+type writerFactory interface {
+	next() (int, io.WriteCloser, error)
+}
+
+type seqChunker struct {
```
kostko, line 47: Add documentation comments briefly describing what `seqChunker` and `parallelChunker` do.

go/storage/mkvs/checkpoint/chunk.go
```diff
@@ -16,65 +18,213 @@ import (
...
+}
+
+func (pc *parallChunker) chunk(ctx context.Context, wf writerFactory) ([]hash.Hash, error) {
+	root, err := newSubtree(pc.ndb, pc.root)
+	if err != nil {
+		return nil, err
+	}
+	pending := []*subtree{root}
+
+	// chunking must be deterministic!
```
kostko: Can you expand on this comment, what does it apply to?

go/storage/mkvs/checkpoint/chunk.go
```diff
@@ -16,65 +18,213 @@ import (
...
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
+
+		pending, err = pc.splitTasks(pending, 0)
+		if err != nil {
+			return nil, fmt.Errorf("chunk: splitting chunking tasks: %w", err)
+		}
+		fmt.Println(len(pending))
```
kostko: Drop debug code.

go/storage/mkvs/checkpoint/chunk.go
```diff
@@ -16,65 +18,213 @@ import (
...
+		children, err := task.split()
+		if err != nil {
+			return nil, fmt.Errorf("splitting chunking task: %w", err)
+		}
+		tasksN = append(tasksN, children...)
+	}
+
+	if len(tasksN) < int(pc.threads) {
+		return pc.splitTasks(tasksN, tried+1)
+	}
```
kostko: Why recursion? Given that it is only at the end it should be easy to refactor to an iterative version.

go/storage/mkvs/checkpoint/chunk.go
```diff
@@ -16,65 +18,213 @@ import (
...
+
+		chunks = append(chunks, hashes...)
 
+		pending = pc.filterFinished(pending)
+	}
+
+	return chunks, nil
+}
+
+func (pc *parallChunker) splitTasks(tasks []*subtree, tried int) ([]*subtree, error) {
```
kostko: Add documentation comment on what this does.

go/storage/mkvs/checkpoint/chunk.go
```diff
@@ -16,65 +18,213 @@ import (
...
+	}
+
+	return tasksN, nil
+}
+
+func (pc *parallChunker) createChunks(ctx context.Context, wf writerFactory, tasks []*subtree) ([]hash.Hash, error) {
+	group, ctx := errgroup.WithContext(ctx)
+
+	chunks := make([]hash.Hash, len(tasks))
+	var mu sync.RWMutex
```
kostko: Why not a regular mutex?

go/storage/mkvs/checkpoint/chunk.go
```diff
@@ -16,65 +18,213 @@ import (
...
+
+	return chunks, nil
+}
+
+func (pc *parallChunker) filterFinished(tasks []*subtree) []*subtree {
+	var pending []*subtree
+	for _, task := range tasks {
+		if !task.isFinished() {
+			pending = append(pending, task)
+		}
```
kostko:
```suggestion
		if task.isFinished() {
			continue
		}
		pending = append(pending, task)
```

go/storage/mkvs/checkpoint/file.go
```diff
@@ -143,17 +131,17 @@ func (fc *fileCreator) GetCheckpoints(_ context.Context, request *GetCheckpoints
 
 func (fc *fileCreator) GetCheckpoint(_ context.Context, version uint16, root node.Root) (*Metadata, error) {
 	// Currently we only support a single version.
-	if version != checkpointVersion {
+	if version != v1 {
 		return nil, ErrCheckpointNotFound
 	}
 
-	checkpointFilename := filepath.Join(
+	metaFile := filepath.Join(
```
kostko:
```suggestion
	metaPath := filepath.Join(
```

go/storage/mkvs/checkpoint/file.go
```diff
@@ -206,19 +194,19 @@ func (fc *fileCreator) DeleteCheckpoint(_ context.Context, version uint16, root
 
 func (fc *fileCreator) GetCheckpointChunk(_ context.Context, chunk *ChunkMetadata, w io.Writer) error {
 	// Currently we only support a single version.
-	if chunk.Version != checkpointVersion {
+	if chunk.Version != v1 {
 		return ErrChunkNotFound
 	}
 
-	chunkFilename := filepath.Join(
+	chunkFile := filepath.Join(
```
kostko: Why the change?

go/storage/mkvs/checkpoint/file.go
```diff
@@ -237,3 +225,19 @@ func NewFileCreator(dataDir string, ndb db.NodeDB) (Creator, error) {
 		ndb:     ndb,
 	}, nil
 }
+
+type fileProvider struct {
```
kostko, line 230: Add documentation comment.

go/storage/mkvs/checkpoint/chunk.go
```diff
@@ -16,65 +18,213 @@ import (
...
 
-func createChunk(
-	ctx context.Context,
-	tree mkvs.Tree,
-	root node.Root,
-	offset node.Key,
-	chunkSize uint64,
-	w io.Writer,
-) (
+type writerFactory interface {
```
kostko, line 35: Add documentation comment.

go/worker/storage/config/config.go
```diff
@@ -31,6 +31,11 @@ type CheckpointerConfig struct {
 	Enabled bool `yaml:"enabled"`
 	// Storage checkpointer check interval.
 	CheckInterval time.Duration `yaml:"check_interval"`
+	// ChunkerSubtreesCount specifies the target number of parallel subtree chunkers
+	// used during checkpoint creation.
+	//
+	// Setting it to 0, will use old sequential chunker.
+	ChunkerSubtreesCount uint16 `yaml:"chunker_subtrees_count"`
```
kostko: Same comment as above about the naming.

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -289,13 +289,22 @@ func (n *Node) getCheckpointList() ([]*storageSync.Checkpoint, error) {
 		return nil, err
 	}
 
-	// Sort checkpoints by version, descending.
-	sort.Slice(list, func(i, j int) bool {
-		// Descending!
-		if list[j].Root.Version == list[i].Root.Version {
-			return bytes.Compare(list[j].Root.Hash[:], list[i].Root.Hash[:]) < 0
+	// Sort checkpoints by version, then by number of peers, descending.
+	sort.Slice(list, func(i, j int) bool { // TODO add a unit test for that.
```
kostko: What's the plan with the TODO?

go/storage/mkvs/checkpoint/subtree.go
```diff
@@ -0,0 +1,239 @@
...
+	rootPtr := node.Pointer{
+		Clean: true,
+		Hash:  root.Hash,
+	}
+
+	rootT := subtree{
+		ndb:  ndb,
+		root: root,
+	}
+	// if root is empty, we should stil return non-empty state,
```
kostko:
```suggestion
	// If root is empty we should still return non-empty state
```

go/storage/mkvs/checkpoint/subtree.go
```diff
@@ -0,0 +1,239 @@
...
+	// that will produce empty proof.
```
kostko:
```suggestion
	// that will produce an empty proof.
```

go/storage/mkvs/checkpoint/subtree.go
```diff
@@ -0,0 +1,239 @@
...
+	pending []pathAtom
+}
+
+func newSubtree(ndb db.NodeDB, root node.Root) (*subtree, error) {
+	rootPtr := node.Pointer{
+		Clean: true,
+		Hash:  root.Hash,
+	}
+
+	rootT := subtree{
```
kostko:
```suggestion
	rootSubtree := subtree{
```

go/storage/mkvs/checkpoint/subtree.go
```diff
@@ -0,0 +1,239 @@
...
+
+	proof, err := pb.Build(ctx)
+	if err != nil {
+		return hash.Hash{}, err
+	}
+
+	return writeChunk(proof, w)
+}
+
+func (ct *subtree) trim() {
```
kostko: Add documentation comment.

go/storage/mkvs/checkpoint/subtree.go
```diff
@@ -0,0 +1,239 @@
...
+	// Remove all trailing nodes whose right subtree state is pending/being explored.
+	for len(ct.pending) > 0 && ct.pending[len(ct.pending)-1].visitState == visitAtRight {
+		ct.pending = ct.pending[:len(ct.pending)-1]
+	}
+}
+
+func (ct *subtree) isFinished() bool {
```
kostko: Add documentation comment.

go/storage/mkvs/checkpoint/subtree.go
```diff
@@ -0,0 +1,239 @@
...
+	return len(ct.pending) == 0
+}
+
+func (ct *subtree) split() ([]*subtree, error) {
```
kostko: Add documentation comment.

go/storage/mkvs/checkpoint/subtree.go
```diff
@@ -0,0 +1,239 @@
...
+		if child == nil {
+			return nil
+		}
+
+		cNd, err := ct.ndb.GetNode(ct.root, child)
+		if err != nil {
+			return fmt.Errorf("getting node from nodedb (ptr hash: %.8s): %w", child.Hash, err)
+		}
+
+		copied := append([]node.Node(nil), ct.path...)
```
kostko:
```suggestion
		copied := append([]node.Node{}, ct.path...)
```

go/storage/mkvs/checkpoint/subtree.go
```diff
@@ -0,0 +1,239 @@
...
+		return []*subtree{ct}, nil
+	}
+
+	var tasks []*subtree
+	addTask := func(parent node.Node, child *node.Pointer) error {
+		if child == nil {
+			return nil
+		}
+
+		cNd, err := ct.ndb.GetNode(ct.root, child)
```
kostko:
```suggestion
		childNode, err := ct.ndb.GetNode(ct.root, child)
```

kostko (commented):
Can you change the E2E tests to use a mix of old and new chunkers among different nodes?

go/worker/storage/config/config.go
kostko: You mean to avoid fragmenting available checkpoints? I don't think we need a warning, we just need sane defaults.

kostko: You could also change the configuration option to be less flexible:
* `none` (old sequential algorithm)
* `parallel` (new parallel algorithm, with a sane number of default threads)

go/storage/mkvs/checkpoint/subtree.go
kostko: What is the difference with the iterator state? The "next offset" is used there instead of the pending list (which is generated each time the iterator seeks), so it should point to the same node given equal constraints (chunk size, etc.)?

go/storage/mkvs/checkpoint/subtree.go
martintomazic: https://github.com/oasisprotocol/oasis-core/pull/6204#discussion_r2148855695

Related to documentation would this be overkill?

Create chunk in such case would have to return `proof` if we want to avoid package cycles (`writeChunk` is problematic otherwise).

kostko: I don't think we need to split it off to a separate package for now.

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -289,13 +289,22 @@ func (n *Node) getCheckpointList() ([]*storageSync.Checkpoint, error) {
...
+		a, b := list[i], list[j]
+		switch { // Descending!
```
ptrus: Could shorten this a bit, and (imo) also make it clearer:

```golang
	switch {
	case a.Root.Version != b.Root.Version:
		return a.Root.Version > b.Root.Version
	case len(a.Peers) != len(b.Peers):
		return len(a.Peers) > len(b.Peers)
	default:
		return bytes.Compare(a.Root.Hash[:], b.Root.Hash[:]) > 0
	}
```

Also consider using the newer `slices.SortFunc`.

go/storage/mkvs/checkpoint/file.go
```diff
@@ -11,48 +11,47 @@ import (
...
-	checkpointProofsVersion = 0
+	chunksDirname = "chunks"
+	metaFilename  = "meta"
+	v1            = 1
+
+	// Versions 1 of checkpoint chunks use proofs version 0.
+	//
+	// Using proof version 1 (latest), does not reduce chunk size, and makes code more complex.
+	//
+	// Consider reevaluating when introducing new checkpoint version, if newer proof version.
```
ptrus:
```suggestion
	// Consider reevaluating when introducing a new checkpoint version.
```

go/storage/mkvs/checkpoint/subtree.go
```diff
@@ -0,0 +1,261 @@
...
+					return
+				}
+			case visitAtRight:
+			}
+		}
+		ct.pending = ct.pending[:len(ct.pending)-1]
+	}
+}
+
+func (ct *subtree) isFinished() bool {
```
ptrus: Maybe clearer name would be `hasPending` or just inline this check in the two places this is called from?

go/worker/storage/config/config.go
ptrus: Agree with the suggestion only to allow configuring which algorithm to use.

go/storage/mkvs/checkpoint/checkpoint_test.go
```diff
@@ -491,3 +492,337 @@ func testPruneGapAfterCheckpointRestore(t *testing.T, factory dbApi.Factory) {
...
+	defer tree1.Close()
+	it1 := tree1.NewIterator(ctx)
+	defer it1.Close()
+	tree2 := mkvs.NewWithRoot(nil, ndb2, root)
+	defer tree2.Close()
+	it2 := tree2.NewIterator(ctx)
+	defer it2.Close()
+	it1.Rewind()
+	it2.Rewind()
+	for count := 0; it1.Valid(); it1.Next() {
```
ptrus, line 868: Note that this doesn't ensure that `ndb2` doesn't have any additional entries. Should ensure that `it2.Next()` reached the end as well after the loop ends.

go/storage/mkvs/checkpoint/checkpoint_test.go
```diff
@@ -491,3 +492,337 @@ func testPruneGapAfterCheckpointRestore(t *testing.T, factory dbApi.Factory) {
...
+		if !it2.Valid() {
+			return fmt.Errorf("key missing in the second database (index %d)", count)
+		}
+		key1, key2 := it1.Key(), it2.Key()
+		if !bytes.Equal(key1, key2) {
+			return fmt.Errorf("keys not equal (index %d): want %s, got %s", count, key1, key2)
+		}
+		val1, val2 := it1.Key(), it2.Key()
```
ptrus: ❗  This should take the `Value` not the key?

go/storage/mkvs/checkpoint/checkpoint_test.go
```diff
@@ -491,3 +492,337 @@ func testPruneGapAfterCheckpointRestore(t *testing.T, factory dbApi.Factory) {
...
+	return nil
+}
+
+// populateDB populates database with n random entries.
+func populateDB(ctx context.Context, ndb dbApi.NodeDB, ns common.Namespace, n uint16, rnd *rand.Rand) (node.Root, error) {
+	tree := mkvs.New(nil, ndb, node.RootTypeState)
+	defer tree.Close()
+
+	for i := 0; i < int(n); i++ {
+		key := make([]byte, rnd.Intn(10)) // TODO Set limit to node.Key max size.?
```
ptrus: It might make sense generating also larger keys and values. Perhaps we could even sample key-pairs from a realistic database? (e.g. a mainnet storage dump?).  This could help create more representative test cases, which could (at least in theory) reveal edge cases that smaller synthetic inputs wouldn't. 

Not sure how much it would impact and if it's worth exploring.

go/storage/mkvs/checkpoint/checkpoint_test.go
martintomazic: So far about `1-2`h max. I plan to run it overnight before merging... Also they need a bit of improvement as @ptrus spotted.

The positive part is that if I change arbitrary operation/condition it seems to catch errors immediately.

Finally, I am planning to add a simple regression test: e.g. add 1000 entries and assert expected checkpoint hashes for `0`, `1`, `10`, and `15` threads. Leaving a comment that if this is intentional change, fuzzers should be re-run.

go/worker/storage/committee/checkpoint_sync.go
martintomazic: I am still planning to add one as part of this PR (should be trivial). 

_Do we see any reason against this ordering or could it bite us in the future_? 

I.e. is there a situation where we would want to release new version and everyone (including old clients) to prioritize it, just like is done for consensus state sync rn. Also is there a reason why we sort it by hash at last?

Partially relates to old  [comment](https://github.com/oasisprotocol/oasis-core/pull/6204#issuecomment-2967266418). Maybe I am overthinking and we should leave it as is...

go/storage/mkvs/checkpoint/checkpoint_test.go
martintomazic: Currently, my ultimate benchmark/correctness test is creating checkpoint from local Sapphire Mainnet database, and then restoring it to new db and ensuring equality, using helpers above.

Previously, this was part of this PR, but has been removed when opening it for a review. As you spotted, equality helper needs to be improved, but hopefully was working already with the keys. Moreover, this TODO should be removed and we should use bigger keys and values... Will think about if we can make it even more realistic.

**Related to realistic test cases/rolling this up in prodution:**
Before defaulting to parallel algorithm in the next release, it would probably be a good idea to only release this with configuration option, still defaulting to old algorithm?

We could first spin a few nodes and observe if there is any issue with creation/peers. Only then it is probably a good idea to default to it. Or I am being paranoid and we have testnet for that. :)

go/storage/mkvs/checkpoint/subtree.go
martintomazic: Fixed with 6cd2661b1407eb2a34e5155f8292d04f89d1e078. In short `trim` was not thorough, so sometimes we produced (redundant) empty chunks.

To make it work, I had to work around a harmless (?) proof builder bug 4d1c7d268a5871f44808bd9303caaf2fd0f10cf7.  I am thinking we should probably document/warn on top of `ProofBuilder.Size`?  Happy to add as part of this PR...

Overall new creation logic (minor tweak) feels more robust and should also work with `v1` proofs, so I like it more.

Now that we have equality, `seqChunker` is already redundant and could be replaced with `parallelChunker`, `threads=1`. I am still inclined to:
1. First release this for runtimes only.
2. Later release for the consensus as well, and entirely remove old sequential chunker?

martintomazic:
Will try to address remaining threads tmr.

Obviously before merging I will need to rerun final sanity/benchmarks. Will post them here.

Possibly did a mistake by not offloading last few commits (integration) to #6219.

go/storage/mkvs/checkpoint/subtree.go
kostko: Yeah that's ok, after merging do open an issue to remove the old chunker so we have it planned.

go/storage/mkvs/checkpoint/checkpoint_test.go
ptrus: I agree with your points. My suggestion was more of a minor note, that it might be worth considering using more realistic key-value pairs in Fuzzing tests as well. But feel free to disregard if it’s not worth the added complexity.

go/worker/storage/committee/node.go
```diff
@@ -226,11 +226,17 @@ func NewNode(
 				return nil, fmt.Errorf("failed to retrieve genesis block: %w", rerr)
 			}
 
+			var threads uint16
+			if config.GlobalConfig.Storage.Checkpointer.ParallelChunker {
+				threads = 12
+			}
```
martintomazic: Chunk size is currently set to `16MB`. Worst case for single runtime all profbuilder are full, meaning around `200MB` memory spike during creation. I guess this should be fine.

We could calculate/cap this dynamically in case `chunkSize` is increased so that memory does not grow unbounded...

go/worker/storage/config/config.go
```diff
@@ -51,8 +53,9 @@ func DefaultConfig() Config {
 		PublicRPCEnabled:       false,
 		CheckpointSyncDisabled: false,
 		Checkpointer: CheckpointerConfig{
-			Enabled:       false,
-			CheckInterval: 1 * time.Minute,
+			Enabled:         false,
+			CheckInterval:   1 * time.Minute,
+			ParallelChunker: false,
```
martintomazic, line 58: To release parallel chunking algorithm for the runtimes only this flag should be set to `true`.

Didn't include configuration changelog since config is mostly for us to test in this case, possibly we will remove it later?

go/storage/mkvs/checkpoint/checkpoint_test.go
martintomazic: Totally makes sense. I noticed there are `n+` places that require initialized and pre-populated db  for testing. Might be beneficial to unify; smtg general purpose and performant + realistic. I can create issue for it but will be low priority for now?

Updated key/val size to `100 bytes`. E.g. Sapphire is prefix + 32 bytes EVM keys so should be better now?

go/storage/mkvs/checkpoint/checkpoint_test.go
martintomazic: Run overnight, no issues!

Strangely initially I had to limit number of items to `2000` or limit parallelization to `5` (default is `20`) else fuzzer would panic during initial exploration (resources, single case taking to long?). Swapping the order of `threads` and `chunkSize` in corpus made this redundant :/.

go/storage/mkvs/checkpoint/file.go
martintomazic: I tried to be consistent with `const` definitions on top where we now use `*filename`/`*dirname` for part of the path. Then `*dir` or `*path` (as you suggested above, previously `file`) is now used for the absolute path. 

Happy to remove/change!

martintomazic:
Should be ready for a final review. 

For me probably the last question would be if we want [this](https://github.com/oasisprotocol/oasis-core/pull/6204#discussion_r2155700032) at all.

Did a sanity bench by spinning Sapphire client node (state sync, mainnet):
* Using old sequential algorithm, it took `13min` to create checkpoint.
* Parallel one finished in `3.6min` -> `3.6` speed-up. 
* Given the previous benchmarks, as the auxiliary state increases (no pruning), improvement increases (`5x` measured).
    *  Obviously checkpoint creation becomes order of magnitude slower for both.

martintomazic (commented):
Should be ready for the final review. 

For me probably the last question would be if we want [this](https://github.com/oasisprotocol/oasis-core/pull/6204#discussion_r2155700032) at all.

Did a sanity bench by spinning Sapphire client node (state sync, mainnet):
* Using old sequential algorithm, it took `13min` to create checkpoint.
* Parallel one finished in `3.6min` -> `3.6` speed-up. 
* Given the previous benchmarks, as the auxiliary state increases, improvement increases (`5x` measured).
    *  Obviously checkpoint creation becomes order of magnitude slower for both.

go/storage/mkvs/checkpoint/checkpoint_test.go
```diff
@@ -491,3 +493,395 @@ func testPruneGapAfterCheckpointRestore(t *testing.T, factory dbApi.Factory) {
 	err = ndb2.Prune(checkpointRootVersion)
 	require.NoError(err, "Prune(%d)", checkpointRootVersion)
 }
+
+// TestCreateRegression is regression test for checkpoint creation.
+//
+// If change is intentional re-run the fuzzers below.
+func TestCreateRegression(t *testing.T) {
```
martintomazic, line 500: Could be moved to `e2e` also provided there is determinism. This way we would also capture expected configuration.

go/consensus/cometbft/full/full.go
```diff
@@ -544,6 +544,11 @@ func (t *fullService) lazyInit() error { // nolint: gocyclo
 	pruneCfg.NumKept = config.GlobalConfig.Consensus.Prune.NumKept
 	pruneCfg.PruneInterval = max(config.GlobalConfig.Consensus.Prune.Interval, time.Second)
 
+	var chunkerThreads uint16
+	if config.GlobalConfig.Storage.Checkpointer.ParallelChunker {
+		chunkerThreads = 12
```
ptrus: Move this into a constant, with some short reasoning/explanation for the picked value.

go/oasis-test-runner/scenario/e2e/runtime/scenario.go
```diff
@@ -264,7 +264,13 @@ func (sc *Scenario) Fixture() (*oasis.NetworkFixture, error) {
 					},
 				},
 			},
-			{RuntimeProvisioner: runtimeProvisioner, Entity: 1, Runtimes: []int{1}},
+			{
+				RuntimeProvisioner: runtimeProvisioner,
+				Entity:             1,
+				Runtimes:           []int{1},
+				// Use a parallel chunking algorithm for the second compute node to show interoperability.
```
ptrus, line 271: Let's also do this in the [long-running txsource tests](https://github.com/oasisprotocol/oasis-core/blob/113878af787d6c6f8da22d6b8a33f6a249180c8b/go/oasis-test-runner/scenario/e2e/runtime/txsource.go#L478-L484)

go/storage/mkvs/checkpoint/checkpoint_test.go
```diff
@@ -491,3 +493,395 @@ func testPruneGapAfterCheckpointRestore(t *testing.T, factory dbApi.Factory) {
...
+		if err != nil {
+			t.Fatalf("Create new file creator: %v", err)
+		}
+		var chunkSize uint64 = 1000
+		cp, err := fc.CreateCheckpoint(ctx, root, chunkSize, tc.threads)
+		if err != nil {
+			t.Fatalf("Create checkpoint (rootHash: %.8s, chunkSize: %d): %v", root.Hash, chunkSize, err)
+		}
+
+		// Ensure regression.
```
ptrus:
```suggestion
		// Ensure no regression.
```

go/storage/mkvs/checkpoint/chunk.go
```diff
@@ -16,65 +18,247 @@ import (
...
+	chunk(ctx context.Context, wf writerFactory) ([]hash.Hash, error)
+}
+
+// writerFactor is a factory that provides writers a chunker can write to.
+type writerFactory interface {
+	// next returns next writer, together with its index.
+	next() (int, io.WriteCloser, error)
+}
+
+// SeqChunker implements chunker interface.
```
ptrus:
```suggestion
// seqChunker implements chunker interface.
```

go/storage/mkvs/checkpoint/subtree.go
```diff
@@ -0,0 +1,276 @@
...
+// hasNext is true when the whole subtree has been chunked.
+func (s *subtree) hasNext() bool {
+	return len(s.pending) == 0
+}
+
+// split splits the subtree chunking task into 0-2 subtrees.
+//
+// In order to respect invariant 1 and 2 it can return:
+//   - 0 subtrees if the subtree has been fully chunked already.
+//   - 1 subtree if splitting is not possible but subtree has still pending work.
```
ptrus:
```suggestion
//   - 1 subtree if splitting is not possible and subtree has pending work.
```

go/worker/storage/committee/node.go
```diff
@@ -226,11 +226,17 @@ func NewNode(
 				return nil, fmt.Errorf("failed to retrieve genesis block: %w", rerr)
 			}
 
+			var threads uint16
+			if config.GlobalConfig.Storage.Checkpointer.ParallelChunker {
+				threads = 12
```
ptrus: Move this into a constant, with some short reasoning/explanation for the picked value.

go/worker/storage/committee/checkpoint_sync.go
kostko: The hash sorting is just to have some deterministic order. I would not change the order beyond what is proposed here.

martintomazic: Added a simple test. :)

go/worker/storage/config/config.go
kostko: This is ok.

go/storage/mkvs/checkpoint/chunk.go
```diff
@@ -16,65 +18,247 @@ import (
...
-	w io.Writer,
-) (
+// Chunker enables splitting database state into chunks, where chunks are merkle proofs
+// representing part of the state.
+type chunker interface {
+	// Chunk creates chunks and writes each into writer provided by the factory.
+	//
+	// A list of hashes corresponding to created chunks is returned.
+	//
+	// A chunker implementation is expected to produce deterministic result.
```
kostko:
```suggestion
	// A chunker implementation is expected to produce a deterministic result.
```

go/storage/mkvs/checkpoint/chunk.go
```diff
@@ -16,65 +18,247 @@ import (
...
+	// This allows using chunks from multiple chunker instances for state
+	// restoration.
+	chunk(ctx context.Context, wf writerFactory) ([]hash.Hash, error)
+}
+
+// writerFactor is a factory that provides writers a chunker can write to.
+type writerFactory interface {
+	// next returns next writer, together with its index.
```
kostko:
```suggestion
	// next returns the next writer together with its index.
```

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -289,15 +290,23 @@ func (n *Node) getCheckpointList() ([]*storageSync.Checkpoint, error) {
...
+	// Sort checkpoints by version, then by number of peers, descending.
+	sortCheckpoints(list)
+
+	return list, nil
+}
+
+// sortCheckpoints sorts the slice in-place (descending by version, peers, hash).
+func sortCheckpoints(s []*storageSync.Checkpoint) {
+	slices.SortFunc(s, func(a, b *storageSync.Checkpoint) int {
+		if cmp := cmp.Compare(b.Root.Version, a.Root.Version); cmp != 0 {
```
kostko: You could use [`cmp.Or`](https://pkg.go.dev/cmp#Or) to simplify the comparator function.

kostko (approved):

PR #6205, opened 2025-05-27
go/consensus/cometbft/api: Truncate block time to seconds

peternose:
Just checking if this works.


kostko (commented):
This also has some light client provider-related changes. Is this on purpose?

kostko (approved):

PR #6207, opened 2025-05-29
go/consensus/api: Deprecate GetSignerNonce

peternose:
The `GetSignerNonce` method should be deprecated and eventually removed, as it does not belong in the core consensus backend. Callers should retrieve the nonce via the account query in the staking backend instead. If this approach proves too complex, a helper method can be added to the staking backend to simplify access.

### DIFF
--- a/go/consensus/api/api.go
+++ b/go/consensus/api/api.go
@@ -159,6 +159,8 @@ type Backend interface {
 
 	// GetSignerNonce returns the nonce that should be used by the given
 	// signer for transmitting the next transaction.
+	//
+	// Deprecated: Use staking backend instead.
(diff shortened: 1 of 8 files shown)


go/worker/common/committee/keymanager.go
```diff
@@ -337,6 +337,12 @@ func (nt *nodeTracker) Nodes(nodes []signature.PublicKey) map[core.PeerID]signat
 }
 
 func (nt *nodeTracker) trackKeymanagerNodes(ctx context.Context) {
+	select {
+	case <-nt.consensus.Synced():
```
peternose, line 341: The consensus needs to be synced so that `WatchStatuses` method doesn't fail when the status notifier hook is executed on subscription.

kostko (approved):
Ideally the "staking" service would be named "accounts".

PR #6214, opened 2025-06-04
runtime: Fix log manager request structure

kostko:


runtime/src/host/log_manager.rs
```diff
@@ -28,15 +28,18 @@ impl LogManager for Protocol {
...
+/// The `PermissionLogView` permission is required to call this method.
 #[derive(Clone, Debug, Default, cbor::Encode, cbor::Decode)]
 pub struct LogGetRequest {
     /// Labels to filter the bundles by. All labels must match and only the first bundle is used.
     pub labels: BTreeMap<String, String>,
     /// Identifier of the component in the bundle.
     pub component_id: String,
+    /// An optional UNIX timestamp to filter log entries by. Only entries with higher timestamps
+    /// will be returned.
+    pub since: u64,
```
ptrus, line 43: Although usually rust will only serialize this (and go deserialize) should this still have  the`#[cbor(optional)]` tag to match the `omitempty` on the go size?

kostko: Fixed.

ptrus (approved):

PR #6224, opened 2025-06-13
Add support for component label attestation

### LINKED ISSUE #6158: Allow nodes to attest to per-component metadata
Nodes should be able to attest that specific per-component metadata (e.g. set from labels by the ROFL scheduler) is set to a given value. One use case for this is to allow a ROFL app instance to prove that it is running with a specific instance ID on the given provider.

These attestations should be retrievable by the apps via a RPC call.

kostko:
Fixes #6158 

This allows apps to request that the node attests to labels attached to those apps have specific values which can be used to prove specific statements about individual components hosted by a specific node.


go/runtime/bundle/component.go
```diff
@@ -38,6 +38,9 @@ type ExplodedComponent struct {
 	// containing the component has been extracted.
 	ExplodedDataDir string
 
+	// Labels are labels attached to the this exploded component.
```
peternose:
```suggestion
	// Labels are labels attached to this exploded component.
```

or 

```suggestion
	// Labels are labels attached to the exploded component.
```

go/runtime/rofl/api/attestation.go
```diff
@@ -0,0 +1,51 @@
...
+
+	// Key is the label key.
+	Key string `json:"key"`
+	// Value is the label value.
+	Value string `json:"value"`
+}
+
+// AttestLabels signs the given list of labels and returns the signature.
+func AttestLabels(signer signature.Signer, labels []*AttestedLabel) (*signature.RawSignature, error) {
+	return signature.SignRaw(signer, AttestLabelsSignatureContext, cbor.Marshal(labels))
```
peternose: If two components have the same labels, the same attest response can be used. Should we include component ID in the signature?

kostko: Ah yes, we need to include the component RAK.

kostko: Fixed.

go/runtime/registry/host_rofl.go
```diff
@@ -267,6 +273,67 @@ func (rh *roflHostHandler) handleHostRegisterNotify(
...
+	}
+
+	identity, err := rh.parent.env.GetNodeIdentity()
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve label attestation key")
+	}
+
+	ri, ok := rh.comps[rh.id]
+	if !ok {
+		return nil, fmt.Errorf("failed to retrieve component RAK")
```
peternose:
```suggestion
		return nil, fmt.Errorf("failed to retrieve component")
```

PR #6228, opened 2025-06-18
Change label attestation to use a map of labels

kostko:


go/runtime/registry/host_rofl.go
```diff
@@ -296,11 +296,10 @@ func (rh *roflHostHandler) handleAttestLabels(args *rofl.AttestLabelsRequest) (*
 		return nil, fmt.Errorf("too many labels specified (max: %d)", rofl.MaxAttestLabels)
 	}
 
-	labels := make([]*rofl.AttestedLabel, 0, len(args.Labels))
+	labels := make(map[string]string, len(args.Labels))
 	for _, key := range args.Labels {
 		// NOTE: We do not discriminate between an empty value and label not being set.
```
peternose:
```suggestion
		// NOTE: We do not discriminate between a zero value and label not being set.
```

go/runtime/registry/host_rofl.go
```diff
@@ -322,13 +321,13 @@ func (rh *roflHostHandler) handleAttestLabels(args *rofl.AttestLabelsRequest) (*
 		RAK:    teeCap.RAK,
 	}
 
-	signature, err := rofl.AttestLabels(identity.NodeSigner, la)
+	encLa, signature, err := rofl.AttestLabels(identity.NodeSigner, la)
```
peternose, line 324: This could be split into 2 methods, `AttestLabels` and `EncodeLabels`.

kostko: But encode labels is just `cbor.Marshal(labels)` which is the standard way we encode most things. We could of course just get rid of `AttestLabels` and use the context manually.

PR #6233, opened 2025-06-19
go/common/sgx: Also cache TCB evaluation data numbers

### LINKED ISSUE #6230: Cache list of TCB evaluation data numbers
We should cache the list of valid TCB evaluation data numbers so it can be leveraged when the PCS service is offline (other TCB artifacts are already cached).

kostko:
Fixes #6230

### DIFF
--- a/go/common/sgx/pcs/cache.go
+++ b/go/common/sgx/pcs/cache.go
@@ -11,11 +11,20 @@ import (
 )
 
 const (
-	tcbCacheKey = "tcb_bundle_cache"
+	tcbBundleCacheKeyPrefix                = "tcb_bundle_cache"
+	tcbEvaluationDataNumbersCacheKeyPrefix = "tcb_evaluation_data_numbers_cache"
 
 	tcbCacheRefreshThreshold    = 14 * 24 * time.Hour
 	tcbCacheSlowRefreshInterval = 24 * time.Hour
 )
+
+func tcbBundleCacheKey(teeType TeeType) []byte {
+	return []byte(fmt.Sprintf("%s.%d", tcbBundleCacheKeyPrefix, teeType))
+}
+
+func tcbEvaluationDataNumbersCacheKey(teeType TeeType) []byte {
+	return []byte(fmt.Sprintf("%s.%d", tcbEvaluationDataNumbersCacheKeyPrefix, teeType))
+}
 
 func readBundleMinTimestamp(bundle *TCBBundle) (time.Time, error) {
 	var err error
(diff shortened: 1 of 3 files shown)


go/common/sgx/pcs/cache.go
```diff
@@ -50,18 +59,57 @@ type tcbBundleCache struct {
...
+		tc.logger.Warn("error checking common store for cached TCB evaluation data numbers",
+			"err", err,
+		)
+		return nil, true
+	}
+
+	now := tc.now()
+	delta := now.Sub(stored.LastUpdate)
+	refresh := delta > tcbCacheSlowRefreshInterval
+	return stored.Numbers, refresh
```
peternose, line 92: Would be better to return `nil` when refresh is needed?

go/common/sgx/pcs/cache.go
```diff
@@ -50,18 +59,57 @@ type tcbBundleCache struct {
...
+}
+
+func (tc *tcbCache) cacheEvaluationDataNumbers(teeType TeeType, numbers []uint32) {
+	cached := tcbEvaluationDataNumbersCache{
+		Numbers:    numbers,
+		LastUpdate: tc.now(),
+	}
+	if err := tc.serviceStore.PutCBOR(tcbEvaluationDataNumbersCacheKey(teeType), cached); err != nil {
+		tc.logger.Error("could not store new TCB evaluation data numbers to cache, ignoring",
```
peternose, line 101: This should be logged by the caller.

go/common/sgx/pcs/cache_test.go
```diff
@@ -169,7 +201,8 @@ func TestTCBCache(t *testing.T) {
 	} {
 		t.Run(name, func(t *testing.T) {
 			fun(t, store, &tcbBundle)
-			_ = store.Delete([]byte(tcbCacheKey))
```
peternose, line 172: Will this remain in the cache forever if not deleted in the new version?

go/common/sgx/pcs/cache.go
kostko: No, because refresh can fail and cached data may be better than failing attestation.

go/common/sgx/pcs/cache.go
kostko: This is consistent with other code in the file. I don't think the caller can do much here.

go/common/sgx/pcs/cache_test.go
kostko: The old keys? Yeah, we can clear or migrate.

peternose (approved):

go/common/sgx/pcs/cache_test.go
kostko: Added migration.

peternose (approved):

PR #6235, opened 2025-06-20
go/consensus/cometbft: Add stateless client node

peternose:
Things that should be done in another PR:

- [ ] Support multiple providers (implement composite provider).
- [ ] Restart watching blocks if connection is lost (switch to another provider and start fetching where the node left off).
- [ ] Support TLS.


go/consensus/cometbft/stateless/core.go
```diff
@@ -0,0 +1,838 @@
...
+		return fmt.Errorf("malformed block meta last commit: %w", err)
+	}
+	if !bytes.Equal(lastCommit.Hash(), lb.LastCommitHash) {
+		return fmt.Errorf("mismatched block meta last commit")
+	}
+
+	return nil
+}
+
+func (c *Core) verifyBlockResults(ctx context.Context, results *consensusAPI.BlockResults, lb *cmttypes.LightBlock) (x *api.BlockResultsMeta, err error) {
```
peternose: Note that the block result verification differs from that of CometBFT's [light RPC node](https://github.com/oasisprotocol/cometbft/blob/v0.37.15-oasis1/light/rpc/client.go#L382), which appears to be incorrect.

go/consensus/cometbft/api/api.go
```diff
@@ -160,18 +160,25 @@ func QueryForApp(eventApp string) cmtpubsub.Query {
 // BlockMeta is the CometBFT-specific per-block metadata.
 type BlockMeta struct {
 	// Header is the CometBFT block header.
-	Header *cmttypes.Header `json:"header"`
+	Header []byte `json:"header"`
```
kostko, line 163: Is this breaking for clients, e.g. Nexus? If so, this must be highlighted in the changelog entry. cc @ptrus

.changelog/6235.feature.md
```diff
@@ -0,0 +1,13 @@
+go/consensus/cometbft: Add stateless client node
+
+One can now start a stateless client as follows:
+
+```yaml
+mode: client-stateless
+# ... sections not relevant are omitted ...
+consensus:
+    providers:
```
kostko, line 11: Do providers need to run the new version of the node as well? This should probably be explicitly mentioned somewhere.

go/consensus/cometbft/cometbft.go
```diff
@@ -30,7 +37,163 @@ func New(
...
+	return stateless.NewServices(provider, lightClient, cfg)
+}
+
+func createProvider() (*consensusAPI.Client, error) {
+	addresses := config.GlobalConfig.Consensus.Providers
+	if len(addresses) == 0 {
+		return nil, fmt.Errorf("no providers configured")
+	}
+
+	return stateless.NewProvider(addresses[0])
```
kostko, line 154: Should it fail or warn if multiple providers are configured? Just discarding the rest silently doesn't seem ideal.

go/consensus/cometbft/stateless/provider.go
```diff
@@ -0,0 +1,23 @@
...
+
+	cmnGrpc "github.com/oasisprotocol/oasis-core/go/common/grpc"
+	consensusAPI "github.com/oasisprotocol/oasis-core/go/consensus/api"
+)
+
+// NewProvider creates a new consensus provider for the stateless client.
+func NewProvider(address string) (*consensusAPI.Client, error) {
+	conn, err := cmnGrpc.Dial(
+		address,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
```
kostko, line 15: It should be possible to also use regular TLS verification (via configuration).

go/runtime/host/provisioner/provisioner.go
```diff
@@ -34,6 +35,7 @@ func New(
 	commonStore *persistent.CommonStore,
 	identity *identity.Identity,
 	consensus consensus.Service,
+	genesisDoc *genesisAPI.Document,
```
kostko, line 38: Could it instead get the genesis document from `consensus`?

go/consensus/cometbft/api/api.go
peternose: Yes, only for Nexus.

.changelog/6235.feature.md
peternose: Yes, because I have changed `BlockMeta`. Will mention in changelog.

go/consensus/cometbft/cometbft.go
peternose: No, because this will be implemented very soon. This is just temporarily.

go/consensus/cometbft/stateless/provider.go
peternose: OK, will open an issue and we add in another PR.

go/runtime/host/provisioner/provisioner.go
peternose: Yes, but it is not as efficient and we would be calling `consensus` during initialization of structs, which is not the best, as consensus might not be up (p2p peers are definitely not available at this stage). A better solution would call `consensus` once on startup when things are already running.

peternose: An alternative is to create custom functions for these values, but again, we would be calling endpoints from constructor.

kostko: But doesn't `consensus.GetGenesisDocument` just return the document it has available locally (looking at `Core`)? No initialization or remote calls needed or am I missing something?

kostko: Anyway, an extra parameter is not a big deal, so it is ok to leave it as is.

peternose: It reads the file again from the disk, as we optimized so that the file is not kept in the memory. Not a remote call.

kostko:
> Things that should be done in another PR:

File issues for the remaining parts.

go/consensus/cometbft/stateless/core.go
```diff
@@ -0,0 +1,838 @@
...
+	// stateRootCacheCapacity is the capacity of the state root LRU cache.
+	stateRootCacheCapacity = 128
+
+	// resultsHashCacheCapacity is the capacity of the results hash LRU cache.
+	resultsHashCacheCapacity = 128
+)
+
+// Core is a stateless implementation of consensus backend.
+type Core struct {
+	mu sync.Mutex
```
ptrus: Maybe mention that this is only used to protect access to the `latestBlock`. Alternatively move it down to `latestBlock` field and/or name it `lbLock`?

go/consensus/cometbft/full/common.go
```diff
@@ -969,3 +911,59 @@ func newCommonNode(ctx context.Context, cfg CommonConfig) *commonNode {
 		startedCh:             make(chan struct{}),
 	}
 }
+
+func TransactionResultsFromCometBFT(height int64, txs [][]byte, responses []*cmtabcitypes.ResponseDeliverTx) ([]*results.Result, error) {
```
ptrus, line 917: Missing documentation comment.  Is it useful to have this exported?

go/consensus/cometbft/crypto/merkle/merkle.go
```diff
@@ -0,0 +1,94 @@
...
+		hashes = append(hashes, hashTransaction(tx))
+	}
+	return hashes
+}
+
+func encodeProof(proof *cmtmerkle.Proof) []byte {
+	return cbor.Marshal(proof)
+}
+
+func decodeProof(proof []byte) (*cmtmerkle.Proof, error) {
```
ptrus, line 82: I see the exported-before-unexported layout pattern, but wouldn’t it be clearer to move short internal helpers that are only used once directly below the method that uses them? Might improve readability in this case.

go/consensus/cometbft/api/api.go
ptrus: Yes, please add a changelog entry. We should ensure that Nexus is updated to support this before it is deployed live on testnet/mainnet.

go/consensus/cometbft/crypto/merkle/merkle.go
```diff
@@ -0,0 +1,94 @@
+package merkle
```
martintomazic, line 3: NIT: Add package docstring.

How about adding simple test suite for exported methods?

go/consensus/cometbft/stateless/service.go
```diff
@@ -0,0 +1,118 @@
+package stateless
```
martintomazic, line 2: NIT: missing package description.

go/consensus/cometbft/stateless/core.go
```diff
@@ -0,0 +1,838 @@
...
+
+	c.mu.Lock()
+	c.latestBlock = blk
+	c.mu.Unlock()
+
+	return nil
+}
+
+func (c *Core) verifyBlock(blk *consensusAPI.Block, lb *cmttypes.LightBlock) error {
+	if blk.Height != lb.Height {
```
martintomazic: For the follow-up:

Should we add some tests that thoroughly checks verification methods (applies to other below as well)? 

Ideally if the structs evolve and we forget to add validation this should capture it as well. Not sure go offers out of the box generators/mutators to easily fuzz this...

go/consensus/cometbft/api/api.go
martintomazic: Is this still needed given that you are truncating miliseconds in verification?

martintomazic (commented):
Nice!

go/consensus/cometbft/api/api.go
peternose: Is needed, because to compute hash and to verify the last commit, we need milliseconds. We could also get them somewhere else, but the right approach would be to use protobuf for encoding all CometBFT structures.

go/consensus/cometbft/stateless/core.go
peternose: I added a test to check if verification works on mainnet using one of the blocks.

If structs evolve, verification hashes should also change. If that’s not the case and the struct’s fields are verified by some other means, I doubt we can create a test to detect this. However, we could test whether the signature of a struct has changed, and if it has, someone should update the verification and the struct’s signature accordingly.

go/consensus/cometbft/crypto/merkle/merkle.go
peternose: I would also like `encodeProof`, `decodeProof` and `encodeProofs` to be together, e.g. for testing or because they can be moved to `Encoder` struct, `encoder.go` file,  or something, which in then not possible. Since I'm not sure what is better, I'm keeping the current organization.

go/consensus/cometbft/full/common.go
peternose: It is useful because both full node and stateless node require it. Ideally, it should be moved to a more appropriate location, but I’m not sure which one would be best. For now, I’ve decided to leave it here.

go/consensus/cometbft/stateless/core.go
peternose: Just moved down, without renaming.

go/consensus/cometbft/stateless/core.go
```diff
@@ -0,0 +1,845 @@
...
+		Transactions: txs,
+		Results:      txResults,
+	}, nil
+}
+
+// GetUnconfirmedTransactions implements api.Backend.
+func (c *Core) GetUnconfirmedTransactions(ctx context.Context) ([][]byte, error) {
+	// Unconfirmed transactions cannot be verified.
+	return c.provider.GetUnconfirmedTransactions(ctx)
+}
```
martintomazic, line 362: NIT: How about setting this to zero value given that it cannot be verified -> no sense to use it? Possibly same for lastRetainedHeight?

Alternative is:
1. An error saying stateless client does not support it.
2. New stateless API subset interface.
3.  Ignore this for now, and tackle as part of multiple providers PR where this methods are even more subtle. 

Probably 3. is best and we think about that later?

go/consensus/cometbft/stateless/core.go
```diff
@@ -0,0 +1,845 @@
...
+
+	select {
+	case <-c.startWatchingBlocksCh:
+		height, err := c.lightClient.LastTrustedHeight()
+		if err != nil {
+			break
+		}
+		return height, nil
+	default:
+	}
```
martintomazic, line 694: Why use light client if no client started watching blocks otherwise provider?  Maybe leave a comment.

go/consensus/cometbft/stateless/services.go
```diff
@@ -0,0 +1,332 @@
...
+	}()
+
+	var errs error
+	for range 2 {
+		err := <-errCh
+		errs = errors.Join(errs, err)
+		cancel()
+	}
+
+	return errs
```
martintomazic: NIT: How about using `errgroup` instead, assuming you only care about first error:

```go
g, ctx := errgroup.WithContext(ctx)
g.Go (...)
g.Go(...)
return g.wait()

```

go/consensus/cometbft/stateless/core.go
```diff
@@ -0,0 +1,845 @@
...
+	// TODO: Verify the latest block results once we add LastResultsHash
+	// to the system block metadata transaction (#6210).
+	lastHeight, err := c.lightClient.LastTrustedHeight()
+	if err != nil || lastHeight <= lb.Height {
+		c.logger.Warn("skipping verification of block results", "height", lb.Height)
+
+		if results.Height != lb.Height {
+			return nil, fmt.Errorf("mismatched block height")
+		}
+		return api.NewBlockResultsMeta(results)
```
martintomazic: Shouldn't you always return error if light client fails to return last trusted height?

go/consensus/cometbft/stateless/core.go
martintomazic: Ok this is actually a great regression test.

How about adding some unhappy paths as well?

With thoroughness check I was mostly worried about "malicious" provider that would pass some faulty data if validation had a bug.

peternose: Added few unhappy paths.

go/consensus/cometbft/stateless/core.go
peternose: I would not set it to zero value neither return an error, as I prefer showing these transactions to the caller and he can decide how to proceed. For transactions, I think current solution is the best. For last retained height, probably point 2, meaning that we should clean the API so that stateless client doesn't need to support this call. We can clean this latter.

go/consensus/cometbft/stateless/core.go
peternose: Because the light client is already up to date, as we are subscribed to new blocks. Added a comment.

go/consensus/cometbft/stateless/services.go
peternose: Nice, will use this of course!

go/consensus/cometbft/stateless/core.go
peternose: Ideally yes, but no because otherwise we cannot verify the latest block and the client will always be 1 block behind. For now, we never verify block results for the last block and this is not secure. But we will change that once we update block metadata transaction with the latest results hash. But that is breaking and takes time.

peternose: Sorry, I misunderstood. Yes, I should return an error here if there is no height.

.changelog/6235.breaking.md
```diff
@@ -0,0 +1,6 @@
+go/consensus/cometbft: Switch to Protobuf encoding for block metadata
```
ptrus, line 1: Staring with https://github.com/oasisprotocol/nexus/pull/1082 Nexus will support the new format, so feel free to proceed with merging/deploying this whenever.

ptrus (approved):

.changelog/6235.breaking.md
peternose: Deployed in 25.5.

PR #6239, opened 2025-06-24
go/worker/storage/committee: Fix Stuck Storage Finalization

### LINKED ISSUE #6238: go/worker/storage: Full apply queue can prevent storage finalization
On one of the nodes I noticed that it wasn't syncing runtime storage and the log was full of the following entries:
```
{"caller":"node.go:1140","err":"storage/database: failed to Apply: mkvs/pathbadger: too many non-finalized versions","level":"error","module":"worker/storage/committee","msg":"can't apply write log","new_root":"<Root ns=000000000000000000000000000000000000000000000000a6d1e3ebf60dff6c version=12194581 type=state-root hash=df7e17cf172934f86a81175fa0b7fd4a3f547d4294320bd9b652de0a8a88d178>","old_root":"<Root ns=000000000000000000000000000000000000000000000000a6d1e3ebf60dff6c version=12194580 type=state-root hash=d5a9d0eabd50ae918574341b4ee68d694b6be47478af887b9003678f5f13c200>","runtime_id":"000000000000000000000000000000000000000000000000a6d1e3ebf60dff6c","ts":"2025-06-21T02:31:43.094401697Z"}
```

It looks like somehow there were many non-finalized applies. Looking at the code it seems like it can happen that the finalization step never happens if there are many diffs in the queue, which can lead to the above condition if too many non-finalized roots versions exist in the database.

Just changing the order should work, e.g. putting finalizations before draining the apply queue.

martintomazic:
Closes #6238

### DIFF
--- a/.changelog/6239.bugfix.md
+++ b/.changelog/6239.bugfix.md
@@ -1,0 +1,1 @@
+go/worker/storage/committee: Fix stuck storage finalization
--- a/go/worker/storage/committee/node.go
+++ b/go/worker/storage/committee/node.go
@@ -26,6 +26,7 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/runtime/host"
 	storageApi "github.com/oasisprotocol/oasis-core/go/storage/api"
 	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/checkpoint"
+	dbApi "github.com/oasisprotocol/oasis-core/go/storage/mkvs/db/api"
 	mkvsDB "github.com/oasisprotocol/oasis-core/go/storage/mkvs/db/api"
 	workerCommon "github.com/oasisprotocol/oasis-core/go/worker/common"
 	"github.com/oasisprotocol/oasis-core/go/worker/common/committee"
@@ -1101,22 +1102,21 @@ func (n *Node) worker() { // nolint: gocyclo
 	pendingApply := &minRoundQueue{}
 	pendingFinalize := &minRoundQueue{}
 
-	// Main processing loop. When a new block comes in, its state and io roots are inspected and their
-	// writelogs fetched from peers in case we don't have them locally yet. Fetches are
-	// asynchronous and, once complete, trigger local Apply operations. These are serialized
-	// per round (all applies for a given round have to be complete before applying anything for following
-	// rounds) using the fetched diffs and pending finalization priority queue. Once a round has all its write
-	// logs applied, a Finalize for it is triggered, again serialized by round but otherwise asynchronous
-	// (pendingFinalization and cachedLastRound).
+	// Main processing loop. When a new block arrives, its state and I/O roots are inspected.
+	// If missing locally, diffs are fetched from peers, possibly for many rounds in parallel,
+	// including all missing rounds since the last fully applied one. Fetched diffs are then applied
+	// in round order, ensuring no gaps. Once a round has all its roots applied, background finalization
+	// for that round is triggered asynchronously, not blocking concurrent fetching and diff application.
 mainLoop:
 	for {
-		// Drain the Apply and Finalize queues first, before waiting for new events in the select
-		// below. Applies are drained first, followed by finalizations (which are asynchronous
-		// but serialized, i.e. only one Finalize can be in progress at a time).
+		// Drain the Apply and Finalize queues first, before waiting for new events in the select below.
 
-		// Apply any writelogs that came in through fetchDiff, but only if they are for the round
-		// after the last fully applied one (lastFullyAppliedRound).
-		if len(*pendingApply) > 0 && lastFullyAppliedRound+1 == (*pendingApply)[0].GetRound() {
+		// Apply fetched writelogs, but only if they are for the round after the last fully applied one
+		// and current number of pending roots to be finalized is smaller than max allowed.
+		applyNext := pendingApply.Len() > 0 &&
+			lastFullyAppliedRound+1 == (*pendingApply)[0].GetRound() &&
+			pendingFinalize.Len() < dbApi.MaxPendingVersions-1 // -1 since one may be already finalizing.
+		if applyNext {
 			lastDiff := heap.Pop(pendingApply).(*fetchedDiff)
 			// Apply the write log if one exists.
 			err = nil
@@ -1172,15 +1172,13 @@ mainLoop:
 			continue
 		}
 
-		// Check if any new rounds were fully applied and need to be finalized. Only finalize
-		// if it's the round after the one that was finalized last (cachedLastRound).
-		// The finalization happens asynchronously with respect to this worker loop and any
-		// applies that happen for subsequent rounds (which can proceed while earlier rounds are
-		// still finalizing).
+		// Check if any new rounds were fully applied and need to be finalized.
+		// Only finalize if it's the round after the one that was finalized last.
+		// As a consequence at most one finalization can be happening at the time.
 		if len(*pendingFinalize) > 0 && cachedLastRound+1 == (*pendingFinalize)[0].GetRound() {
 			lastSummary := heap.Pop(pendingFinalize).(*blockSummary)
 			wg.Add(1)
-			go func() {
+			go func() { // Don't block fetching and applying remaining rounds.
 				defer wg.Done()
 				n.finalize(lastSummary)
 			}()
--- a/go/worker/storage/committee/node.go
+++ b/go/worker/storage/committee/node.go
@@ -1068,7 +1068,7 @@ func (n *Node) worker() { // nolint: gocyclo
 				"awaiting_retry", syncing.awaitingRetry,
 			)
 
-			prev := hashCache[i-1] // Closures take refs, so they need new variables here.
+			prev := hashCache[i-1]
 			this := hashCache[i]
 			prevRoots := make([]storageApi.Root, len(prev.Roots))
 			copy(prevRoots, prev.Roots)
@@ -1090,12 +1090,10 @@ func (n *Node) worker() { // nolint: gocyclo
 				if !syncing.outstanding.contains(rootType) && syncing.awaitingRetry.contains(rootType) {
 					syncing.scheduleDiff(rootType)
 					fetcherGroup.Add(1)
-					n.fetchPool.Submit(func(round uint64, prevRoot, thisRoot storageApi.Root) func() {
-						return func() {
-							defer fetcherGroup.Done()
-							n.fetchDiff(round, prevRoot, thisRoot)
-						}
-					}(this.Round, prevRoots[i], this.Roots[i]))
+					n.fetchPool.Submit(func() {
+						defer fetcherGroup.Done()
+						n.fetchDiff(this.Round, prevRoots[i], this.Roots[i])
+					})
 				}
 			}
 		}
@@ -1184,10 +1182,10 @@ mainLoop:
 		if len(*outOfOrderFinalizable) > 0 && cachedLastRound+1 == (*outOfOrderFinalizable)[0].GetRound() {
 			lastSummary := heap.Pop(outOfOrderFinalizable).(*blockSummary)
 			fetcherGroup.Add(1)
-			go func(lastSummary *blockSummary) {
+			go func() {
 				defer fetcherGroup.Done()
 				n.finalize(lastSummary)
-			}(lastSummary)
+			}()
 			continue
 		}
 
--- a/go/worker/storage/committee/node.go
+++ b/go/worker/storage/committee/node.go
@@ -392,35 +392,38 @@ func (n *Node) fetchDiff(round uint64, prevRoot, thisRoot storageApi.Root) {
 		case <-n.ctx.Done():
 		}
 	}()
+
 	// Check if the new root doesn't already exist.
-	if !n.localStorage.NodeDB().HasRoot(thisRoot) {
-		result.fetched = true
-		if thisRoot.Hash.Equal(&prevRoot.Hash) {
-			// Even if HasRoot returns false the root can still exist if it is equal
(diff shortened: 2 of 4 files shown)


go/worker/storage/committee/node.go
```diff
@@ -1102,18 +1094,37 @@ func (n *Node) worker() { // nolint: gocyclo
...
+		// still finalizing).
+		if len(*outOfOrderFinalizable) > 0 && cachedLastRound+1 == (*outOfOrderFinalizable)[0].GetRound() {
+			lastSummary := heap.Pop(outOfOrderFinalizable).(*blockSummary)
+			wg.Add(1)
+			go func(lastSummary *blockSummary) {
+				defer wg.Done()
+				n.finalize(lastSummary)
+			}(lastSummary)
+			continue
+		}
```
martintomazic: Could it happen that fetchers/diff application is faster then finalization? In such case would it be better to add a var and count/limit pending non-finalized versions instead?

In theory this case is handled below via retry...

kostko: Yeah we have `maxInFlightRounds` for diffs, but no such thing for finalizations. It may make sense to have it for finalizations as well and pause fetching new diffs if finalizations are slow (even though they should be fast).

But as you said, currently the database will reject new applies after reaching a limit, and with the changed order this should now be handled correctly (e.g. finalizations will be spawned before any new applies will be queued).

go/worker/storage/committee/node.go
```diff
@@ -1065,7 +1059,7 @@ func (n *Node) worker() { // nolint: gocyclo
 				"awaiting_retry", syncing.awaitingRetry,
 			)
 
-			prev := hashCache[i-1] // Closures take refs, so they need new variables here.
+			prev := hashCache[i-1]
```
kostko: Why remove this comment?

go/worker/storage/committee/node.go
```diff
@@ -1086,13 +1080,11 @@ func (n *Node) worker() { // nolint: gocyclo
...
-					n.fetchPool.Submit(func(round uint64, prevRoot, thisRoot storageApi.Root) func() {
-						return func() {
-							defer fetcherGroup.Done()
-							n.fetchDiff(round, prevRoot, thisRoot)
-						}
-					}(this.Round, prevRoots[i], this.Roots[i]))
+					wg.Add(1)
+					n.fetchPool.Submit(func() {
+						defer wg.Done()
+						n.fetchDiff(this.Round, prevRoots[i], this.Roots[i])
```
kostko, line 1091: I assume this is now fine because of new loop variable binding semantics since Go 1.22?

go/worker/storage/committee/node.go
```diff
@@ -1102,18 +1094,37 @@ func (n *Node) worker() { // nolint: gocyclo
...
 	// logs applied, a Finalize for it is triggered, again serialized by round but otherwise asynchronous
 	// (outOfOrderFinalizable and cachedLastRound).
 mainLoop:
 	for {
-		// Drain the Apply and Finalize queues first, before waiting for new events in the select
-		// below. Applies are drained first, followed by finalizations (which are asynchronous
-		// but serialized, i.e. only one Finalize can be in progress at a time).
+		// Drain the Apply and Finalize queues before waiting for new events in the select below.
+		//
+		// Finalizations (asynchronous, single-threaded), must always run before applies, too
```
kostko:
```suggestion
		// Finalizations (asynchronous, single-threaded), must always run before applies to
```

go/worker/storage/committee/node.go
```diff
@@ -1102,18 +1094,37 @@ func (n *Node) worker() { // nolint: gocyclo
...
+		// The finalization happens asynchronously with respect to this worker loop and any
+		// applies that happen for subsequent rounds (which can proceed while earlier rounds are
+		// still finalizing).
+		if len(*outOfOrderFinalizable) > 0 && cachedLastRound+1 == (*outOfOrderFinalizable)[0].GetRound() {
+			lastSummary := heap.Pop(outOfOrderFinalizable).(*blockSummary)
+			wg.Add(1)
+			go func(lastSummary *blockSummary) {
+				defer wg.Done()
+				n.finalize(lastSummary)
+			}(lastSummary)
```
kostko: Why is this wrapping needed here, won't the variable in the closure reference the correct `lastSummary` anyway (which is always created from the `heap.Pop` call?

go/worker/storage/committee/node.go
martintomazic: >In theory this case is handled below via retry...

Actually this is still not good  `cachedLastRound` is not updated until the pending diffs are applied. So the same problem would happen.

Will rewrite and use limit instead, (incrementing for any applied root and decrementing for any finalized).

kostko: Ah you are right.

go/worker/storage/committee/node.go
```diff
@@ -61,27 +61,31 @@ const (
 	// maxInFlightRounds is the maximum number of rounds that should be fetched before waiting
 	// for them to be applied.
 	maxInFlightRounds = 100
+
+	// maxNonFinalizedRoots is the maximum number of allowed non-finalized roots.
+	// It should be set lower than the limit configured by the underlying node database.
+	maxNonFinalizedRoots = 1000
```
martintomazic: Do we want to export this as part of the `nodedb` API? I set limit to `1000`, for `pathbadger` is is `4000`, badger I think has none.

go/worker/storage/committee/node.go
martintomazic: I think it was here because of legacy `i` var capture in the for loop?

.changelog/6239.bugfix.md
```diff
@@ -0,0 +1 @@
+go/worker/storage/committee: Fix Stuck Storage Finalization
```
kostko:
```suggestion
go/worker/storage/committee: Fix stuck storage finalization
```

go/worker/storage/committee/node.go
```diff
@@ -61,27 +61,31 @@ const (
...
 )
 
 type roundItem interface {
 	GetRound() uint64
 }
 
-// outOfOrderRoundQueue is a Round()-based min priority queue.
-type outOfOrderRoundQueue []roundItem
+// minRoundPQ is a Round()-based min priority queue.
+type minRoundPQ []roundItem
```
kostko:
```suggestion
type minRoundQueue []roundItem
```

go/worker/storage/committee/node.go
```diff
@@ -392,35 +396,38 @@ func (n *Node) fetchDiff(round uint64, prevRoot, thisRoot storageApi.Root) {
...
-		}
+	// Even if HasRoot returns false the root can still exist if it is equal
+	// to the previous root and the root was emitted by the consensus committee
+	// directly (e.g., during an epoch transition).
+	if thisRoot.Hash.Equal(&prevRoot.Hash) {
+		result.writeLog = storageApi.WriteLog{}
+		return
 	}
+
+	// New root does not yet exist in storage and we need to fetch it from peer.
```
kostko:
```suggestion
	// New root does not yet exist in storage and we need to fetch it from a peer.
```

go/worker/storage/committee/node.go
kostko: Yes, I think it makes sense to expose it.

martintomazic: Moved, also "fixed"  `>=`  check with `>` inside pathbadger. 

I think should be safe since less restrictive also metadata is cleaned upon restart...

Using `finalizing` I also made it work if in the future  `maxNonFinalizedRoots = 1`. Maybe an overkill, but previously this would not work.

go/worker/storage/committee/node.go
```diff
@@ -1149,41 +1169,41 @@ mainLoop:
...
-			fetcherGroup.Add(1)
-			go func(lastSummary *blockSummary) {
-				defer fetcherGroup.Done()
+		// Check if any new rounds were fully applied and need to be finalized.
+		// Only finalize if it's the round after the one that was finalized last.
+		// As a consequence at most one finalization can be happening at the time.
+		if len(*pendingFinalize) > 0 && cachedLastRound+1 == (*pendingFinalize)[0].GetRound() {
+			lastSummary := heap.Pop(pendingFinalize).(*blockSummary)
+			wg.Add(1)
+			go func() { // don't block fetching and applying remaining rounds.
```
kostko:
```suggestion
			go func() { // Don't block fetching and applying remaining rounds.
```

go/worker/storage/committee/node.go
```diff
@@ -1266,49 +1286,48 @@ mainLoop:
...
-				if err != nil {
-					n.logger.Error("failed to flush synced state",
-						"err", err,
-					)
-				}
-				storageWorkerLastFullRound.With(n.getMetricLabels()).Set(float64(finalized.summary.Round))
+			if finalized.err != nil {
+				// Request a node shutdown given that syncing is effectively blocked.
+				_ = n.commonNode.HostNode.RequestShutdown(n.ctx, false)
+			}
```
kostko: This is missing a `break`.

go/worker/storage/committee/node.go
```diff
@@ -1102,23 +1116,29 @@ func (n *Node) worker() { // nolint: gocyclo
...
-		if len(*outOfOrderDoneDiffs) > 0 && lastFullyAppliedRound+1 == (*outOfOrderDoneDiffs)[0].GetRound() {
-			lastDiff := heap.Pop(outOfOrderDoneDiffs).(*fetchedDiff)
+		// Drain the Apply and Finalize queues first, before waiting for new events in the select below.
+
+		// Apply fetched writelogs, but only if they are for the round after the last fully applied one
+		// and current number of pending roots to be finalized is smaller than max allowed.
+		maxPending := dbApi.MaxPendingVersions
+		if n.isFinalizing() {
+			maxPending--
+		}
```
kostko: Why is this needed? Looks like it adds quite some code just to subtract one here? You could just always subtract one.

go/worker/storage/committee/node.go
```diff
@@ -1149,41 +1169,41 @@ mainLoop:
...
-					// with respect to Apply operations for subsequent rounds.
-					lastFullyAppliedRound = lastDiff.round
-					heap.Push(outOfOrderFinalizable, summary)
-				}
+				continue
+			}
+			syncing.outstanding.remove(lastDiff.thisRoot.Type)
+
+			if !syncing.outstanding.isEmpty() || !syncing.awaitingRetry.isEmpty() {
+				continue
```
kostko:
```suggestion

			syncing.outstanding.remove(lastDiff.thisRoot.Type)
			if !syncing.outstanding.isEmpty() || !syncing.awaitingRetry.isEmpty() {
				continue
```

go/worker/storage/committee/node.go
```diff
@@ -1149,41 +1169,41 @@ mainLoop:
...
+		if len(*pendingFinalize) > 0 && cachedLastRound+1 == (*pendingFinalize)[0].GetRound() {
+			lastSummary := heap.Pop(pendingFinalize).(*blockSummary)
+			wg.Add(1)
+			go func() { // don't block fetching and applying remaining rounds.
+				defer wg.Done()
 				n.finalize(lastSummary)
-			}(lastSummary)
+			}()
 			continue
 		}
```
kostko, line 1187: Should we process finalization notifications (`finalizeCh`) _before_ other cases? E.g. have an additional `select` with an empty `default` branch before the below select and move handling finalize notifications here. This way finalize would always be checked even if other events (e.g. blocks or sync completions) are coming in (which can possibly trigger more round fetches and skip finalizations).

go/worker/storage/committee/node.go
martintomazic: Added a `return` instead given that is sensible to exit.

go/worker/storage/committee/node.go
martintomazic: I also dislike it. 

Placed it here if in the future someone places `MaxPendingVersions=1`. subtracting -1 here would prevent any further applications. Not checking this however may cause adding 1, whilst 1 is already in the finalization, resulting in 2 and breaking max pending contract.

Suggestions? Maybe an error? Overthinking ?

go/worker/storage/committee/node.go
martintomazic: Oh you probably meant to move this part up as well right below `finalizeCh` to logically group finalization.

Possible we could remove `continue` here as well.

martintomazic: Slightly cons to the recent 2 commits (combined): Now application and finalization may give more blocking to diff fetching leaving some extra idle time. Not sure how critical this is, but in ideal scenario I think this should be 3 independent processes each one respecting its limit: `maxFetcher`, `maxNonApplied` `maxFinalization=1`....

go/worker/storage/committee/node.go
kostko: I don't think ever having it as 1 is realistic. I would instead add a comment to where the constant is defined that it must be greater than 1 or similar.

go/worker/storage/committee/node.go
kostko: Ok, let's not do this in this PR and we can refactor the pipeline separately.

go/worker/storage/committee/node.go
martintomazic: Omitted the comment altogether since e2e would fail anyway and don't want to write in the API assumptions clients have about it.

go/worker/storage/committee/node.go
martintomazic: Actually `break main` is more appropriate so that we wait for `wg.Wait` to prevent goroutine leaks :).

PR #6242, opened 2025-06-29
go/storage/committee: Optimize state sync

### LINKED ISSUE #6241: Optimize runtime iterative state sync
Currently, it takes days (Sapphire, mainnet) for iterative state sync to finish.

State sync should be benchmarked and optimized.

Possible bottlenecks:
1. Default fetcher pool has `4` workers ([see](https://github.com/oasisprotocol/oasis-core/blob/69f0d2999efaccf81bdec4796cbaead11bd54851/go/worker/storage/committee/node.go#L166)).
    * Possibly increase, also bottleneck could be number of peers for diff fetching.
2. Fetching, applying diffs and finalizing versions ([see](https://github.com/oasisprotocol/oasis-core/blob/master/go/worker/storage/committee/node.go#L1114:L1308)) are three completely independent processes.
    * Currently this is not implemented optimally and processes may block each other.
    * See original discussion [here](https://github.com/oasisprotocol/oasis-core/pull/6239#discussion_r2171232847).

martintomazic:
Closes #6241 

WIP

TODO:
- [ ]  bench the new performance.
- [ ] Make ready for review: Minor refactors/simplications + fix docs.

**Motivation**:
TODO

**Benchmarks**:
Preliminary benchmarks of  `master` :
1. finalization is around 2-5 times faster than the application. 
2. Fetching is also around 2-5 times faster than the syncing on my local machine with very average wifi.
   * I will repeat benchmarks in the cloud. 
3. Applying and finalizing is order of magnitude slower when syncing many blocks, compared to when syncing with synced worker.
    *  This happens when checkpoint sync is disabled or just after it. Makes sense as badger has more to clean load.
   
I think the behavior, we observed in #6238 was due to very fast fetchers or likely badger doing compaction, which caused the rounds pending finalization to go above 5k. (fixed by capping max non-finalized).

**Next Steps**
1. NIT: Consider refactoring further and get rid of  `inFlight` structs or possibly simplify the fetcher worker doing to much stuff.

Fix existing issues you found in the code:
- [ ] Unlikely issue of potentially registering availability to early. (`master`) 
    * After syncing checkpoint there could be many block headers in the infinite channels pending diff sync, but `nudgeAvailability` may act on the first block already.
- [ ] Existing issue in the master where genesis checkpoint may not be checkpointed if syncing from it (it will be however checkpointed on the first restart of the node).

### DIFF
--- a/go/worker/storage/statesync/diff_sync.go
+++ b/go/worker/storage/statesync/diff_sync.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sync"
 	"time"
 
 	"github.com/oasisprotocol/oasis-core/go/common"
@@ -18,6 +17,7 @@ import (
 	storageApi "github.com/oasisprotocol/oasis-core/go/storage/api"
 	dbApi "github.com/oasisprotocol/oasis-core/go/storage/mkvs/db/api"
 	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/sync/errgroup"
 )
 
 const (
@@ -65,11 +65,6 @@ type fetchedDiff struct {
 
 func (d *fetchedDiff) GetRound() uint64 {
 	return d.round
-}
-
-type finalizedResult struct {
-	summary *blockSummary
-	err     error
 }
 
 // Fetcher fetches storage diffs from the peers.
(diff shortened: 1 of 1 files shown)


go/worker/storage/committee/node.go
```diff
@@ -1268,12 +1208,54 @@ mainLoop:
...
-		case finalized := <-n.finalizeCh:
+			// Trigger next round diff application (if fetched already), provided
+			// current number of pending roots to be finalized is be smaller than max allowed.
+			applyNext := pendingApply.Len() > 0 &&
+				lastFullyAppliedRound+1 == (*pendingApply)[0].GetRound() &&
+				pendingFinalize.Len() < dbApi.MaxPendingVersions-1 // -1 since one may be already finalizing.
+			if !applyNext {
+				break
+			}
+			applyCh <- heap.Pop(pendingApply).(*fetchedDiff)
```
martintomazic: This is wrong, as nothing will empty the pending apply queue, potentially lagging behind too much.

~Wrapping with for loop could do~. Will fix tmr or find better solution.

PR #6244, opened 2025-07-01
keymanager/src/crypto: Deprecate checksum in key pair

peternose:
I don't think this checksum is ever used, or is it? If not, it might be best to remove it, since it's unrelated to the keys derived from ephemeral secrets.

### DIFF
--- a/.changelog/6244.internal.md
+++ b/.changelog/6244.internal.md
@@ -1,0 +1,1 @@
+keymanager/src/crypto: Deprecate checksum in key pair
--- a/keymanager/src/crypto/kdf.rs
+++ b/keymanager/src/crypto/kdf.rs
@@ -177,9 +177,7 @@ impl Inner {
     }
 
     // Derive ephemeral or long-term keys from the given secret.
-    fn derive_keys(&self, secret: Secret, xof_custom: &[u8]) -> Result<KeyPair> {
-        let checksum = self.get_checksum()?;
-
+    fn derive_keys(&self, secret: Secret, xof_custom: &[u8]) -> KeyPair {
         // Note: The `name` parameter for cSHAKE is reserved for use by NIST.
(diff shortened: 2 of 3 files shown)


peternose:
> So we need to make sure clients making this optional are deployed first and only then we can drop it? The current PR seems to already drop it in responses, won't this break existing clients?

Won't dropped values decode to default?

```rust
#[test]
fn test_roundtrip() {
    let sk = x25519::PrivateKey::generate();
    let pk = x25519::PublicKey::from(&sk);
    let keypair = KeyPair {
        input_keypair: super::InputKeyPair { pk, sk },
        state_key: StateKey([1; STATE_KEY_SIZE]),
        _deprecated_checksum: None,
    };
    let encoded = cbor::to_vec(keypair);
    let _: KeyPairOld = cbor::from_slice(&encoded).unwrap();

    let sk = x25519::PrivateKey::generate();
    let pk = x25519::PublicKey::from(&sk);
    let keypair = KeyPair {
        input_keypair: super::InputKeyPair { pk, sk },
        state_key: StateKey([1; STATE_KEY_SIZE]),
        _deprecated_checksum: Some(vec![1, 2, 3]),
    };
    let encoded = cbor::to_vec(keypair);
    let _: KeyPairOld = cbor::from_slice(&encoded).unwrap();

    let sk = x25519::PrivateKey::generate();
    let pk = x25519::PublicKey::from(&sk);
    let keypair = KeyPairOld {
        input_keypair: super::InputKeyPair { pk, sk },
        state_key: StateKey([1; STATE_KEY_SIZE]),
        checksum: vec![1, 2, 3],
    };
    let encoded = cbor::to_vec(keypair);
    let _: KeyPair = cbor::from_slice(&encoded).unwrap();
}
```

Issue #5750, opened 2024-07-03
Make storage request more peers if no checkpoints can be found


kostko:
In case the storage worker is unable to initialize because there are no usable checkpoints available, make it able to request additional peers from the P2P subsystem (they could be requested through discovery etc.).

Similar should probably be done for other protocols, e.g. in case key manager client is unable to do queries it should try to discover more peers etc.

martintomazic:
**Context:**
Target number of protocol peers is maintained by peer manager [service](https://github.com/oasisprotocol/oasis-core/blob/6ac1216230f21fbdcda720e27a12e95178f6e62f/go/p2p/peermgmt/peermgr.go#L289:L297) that ensures both the minimum number of consensus registered peers and total number of peers (registered or non-registered). The former are being discovered via consensus service, while the latter are discovered via seed nodes discovery ([see](https://github.com/oasisprotocol/oasis-core/blob/6ac1216230f21fbdcda720e27a12e95178f6e62f/go/p2p/peermgmt/peermgr.go#L367:L382)).

**Problem:**
If host node has sufficient number of peers connected, but none of them has usable checkpoints, host might never trigger new discovery** even though with current peerset it may never have checkpoints available (or at least in reasonable time).

** Assuming peers that never record neither success nor failure for a given protocol are not pruned and replaced fast.

**Naive solution**:
1. Add new method to [`p2p.Service`](https://github.com/oasisprotocol/oasis-core/blob/6ac1216230f21fbdcda720e27a12e95178f6e62f/go/p2p/api/api.go#L55) or  [`rpc.P2P`](https://github.com/oasisprotocol/oasis-core/blob/6ac1216230f21fbdcda720e27a12e95178f6e62f/go/p2p/rpc/rpc.go#L14) interface:
```go
// ForcePeerDiscovery discovers up to max new peers for the specified protocol.
ForcePeerDiscovery(p core.ProtocolID, max int)
```
2. If [no checkpoints received ](https://github.com/oasisprotocol/oasis-core/blob/6ac1216230f21fbdcda720e27a12e95178f6e62f/go/worker/storage/p2p/sync/client.go#L98) (or only genesis), call method above.
3. Repeat similar to **2.** for other protocols.

**Alternative/Complementary solution**:
[...]

PR #6262, opened 2025-07-10
Split storage sync P2P protocol

### LINKED ISSUE #5751: Separate storage checkpoint sync p2p protocol
Currently both runtime checkpoint sync and block sync use the same p2p protocol. We should separate the two to support different parameters and better discovery as there may be less nodes available that expose checkpoints.

This should be done in a backwards-compatible manner where both protocols are advertised and used, with the new protocol being prioritized. Then the old protocol can slowly be removed.

martintomazic:
Closes #5751, blocked by #6270.

Likely solves root cause of #5750.

**Open for discussion:**
1. E2e test to be more confident in integration part, even though existing integration test was helpful at catching bugs.
    * Cons: Additional config parameter, e2e test, and wrapping on the node side that delegates to appropriate client.
    * Alternatives: 
        * Adapt oasis tesnet runner to receive path to "legacy" binary (would also require to change CI).
        * Bootstrap a few nodes on the testnet and manually monitor p2p dashboard before release.
3. Extending rpc client (458f76f6274bbc22f7daab56a4b7882cc71dc216) with fallback protocol is likely the way to go.
    * Cons: Since existing storage sync has few methods, it might be simpler for the new clients to take the legacy client and fallback to it manually, avoiding the need for this change and additional [fallback manager](https://github.com/oasisprotocol/oasis-core/blob/f619b56b601b037ec7f0d7a60b499f8956714ae8/go/worker/storage/p2p/checkpoints/client.go#L116:L143).
4. Should new protocols alias legacy protocol definitions/methods given they are semantically equivalent?
    * Pros: Less duplication && less type fighting. 
    * Cons1: When removing the legacy protocol, things will have to be copied again (double work).
    * Cons2: In a way not aliasing protects against accidental changes in one manifesting in other.

### DIFF
--- a/go/p2p/protocol/protocol.go
+++ b/go/p2p/protocol/protocol.go
@@ -12,33 +12,32 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/p2p/api"
 )
 
-type protocolRegistry struct {
+// Registry is responsible for ensuring unique protocol ids.
+type Registry struct {
 	mu        sync.Mutex
 	protocols map[core.ProtocolID]struct{}
 }
 
-func newProtocolRegistry() *protocolRegistry {
-	return &protocolRegistry{
+func NewRegistry() *Registry {
+	return &Registry{
 		protocols: make(map[core.ProtocolID]struct{}),
 	}
 }
 
-var registry = newProtocolRegistry()
-
 // ValidateProtocolID panics if the protocol id is not unique.
-func ValidateProtocolID(p core.ProtocolID) {
-	registry.mu.Lock()
-	defer registry.mu.Unlock()
+func (r *Registry) ValidateProtocolID(p core.ProtocolID) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
 
-	if _, ok := registry.protocols[p]; ok {
+	if _, ok := r.protocols[p]; ok {
 		panic(fmt.Sprintf("p2p/protocol: protocol or topic with name '%s' already exists", p))
 	}
-	registry.protocols[p] = struct{}{}
+	r.protocols[p] = struct{}{}
 }
 
 // ValidateTopicID panics if the topic id is not unique.
-func ValidateTopicID(topic string) {
-	ValidateProtocolID(core.ProtocolID(topic))
+func (r *Registry) ValidateTopicID(topic string) {
+	r.ValidateProtocolID(core.ProtocolID(topic))
 }
 
 // NewProtocolID generates a protocol identifier for a consensus P2P protocol.
--- a/go/p2p/rpc/client.go
+++ b/go/p2p/rpc/client.go
@@ -269,8 +269,8 @@ type Client interface {
 }
 
 type client struct {
-	host       core.Host
-	protocolID protocol.ID
+	host        core.Host
+	protocolIDs []protocol.ID
 
 	listeners struct {
 		sync.RWMutex
@@ -486,7 +486,7 @@ func (c *client) call(
 	stream, err := c.host.NewStream(
 		ctx,
 		peerID,
-		c.protocolID,
+		c.protocolIDs...,
 	)
 	if err != nil {
 		return fmt.Errorf("failed to open stream: %w", err)
@@ -612,15 +612,15 @@ func retryFn(ctx context.Context, fn func() error, maxRetries uint64, retryInter
 }
 
 // NewClient creates a new RPC client for the given protocol.
-func NewClient(h host.Host, p protocol.ID) Client {
+func NewClient(h host.Host, p protocol.ID, fallback ...protocol.ID) Client {
 	if h == nil {
 		// No P2P service, use the no-op client.
 		return &nopClient{}
 	}
 
 	return &client{
-		host:       h,
-		protocolID: p,
+		host:        h,
+		protocolIDs: append([]protocol.ID{p}, fallback...),
 		listeners: struct {
 			sync.RWMutex
 			m map[ClientListener]struct{}
--- a/go/worker/storage/p2p/checkpointsync/client.go
+++ b/go/worker/storage/p2p/checkpointsync/client.go
@@ -33,6 +33,9 @@ type Client interface {
 		request *GetCheckpointChunkRequest,
 		cp *Checkpoint,
 	) (*GetCheckpointChunkResponse, rpc.PeerFeedback, error)
+
+	// IsReady is true when protocol client is aware of at least one remote peer.
+	IsReady() bool
 }
 
 // Checkpoint contains checkpoint metadata together with peer information.
@@ -112,6 +115,10 @@ func (c *client) GetCheckpointChunk(
 
 func (c *client) getBestPeers(opts ...rpc.BestPeersOption) []core.PeerID {
 	return append(c.mgr.GetBestPeers(opts...), c.fallbackMgr.GetBestPeers(opts...)...)
+}
+
+func (c *client) IsReady() bool {
+	return len(c.getBestPeers()) > 0
 }
 
 // NewClient creates a new checkpoint sync protocol client.
--- a/go/worker/storage/p2p/checkpointsync/client.go
+++ b/go/worker/storage/p2p/checkpointsync/client.go
@@ -1,4 +1,4 @@
-package sync
+package checkpointsync
 
 import (
 	"context"
@@ -10,23 +10,20 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/p2p/protocol"
 	"github.com/oasisprotocol/oasis-core/go/p2p/rpc"
 	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/checkpoint"
+	"github.com/oasisprotocol/oasis-core/go/worker/storage/p2p/sync"
 )
 
 const (
 	// minProtocolPeers is the minimum number of peers from the registry we want to have connected
-	// for StorageSync protocol.
+	// for checkpoint sync protocol.
 	minProtocolPeers = 5
 
-	// totalProtocolPeers is the number of peers we want to have connected for StorageSync protocol.
+	// totalProtocolPeers is the number of peers we want to have connected for checkpoint sync protocol.
 	totalProtocolPeers = 10
 )
 
-// Client is a storage sync protocol client.
+// Client is a checkpoint sync protocol client.
 type Client interface {
-	// GetDiff requests a write log of entries that must be applied to get from the first given root
-	// to the second one.
-	GetDiff(ctx context.Context, request *GetDiffRequest) (*GetDiffResponse, rpc.PeerFeedback, error)
-
 	// GetCheckpoints returns a list of checkpoint metadata for all known checkpoints.
 	GetCheckpoints(ctx context.Context, request *GetCheckpointsRequest) ([]*Checkpoint, error)
 
@@ -47,26 +44,14 @@ type Checkpoint struct {
 }
 
 type client struct {
-	rcC  rpc.Client
-	rcD  rpc.Client
-	mgrC rpc.PeerManager
-	mgrD rpc.PeerManager
-}
-
-func (c *client) GetDiff(ctx context.Context, request *GetDiffRequest) (*GetDiffResponse, rpc.PeerFeedback, error) {
-	var rsp GetDiffResponse
-	pf, err := c.rcD.CallOne(ctx, c.mgrD.GetBestPeers(), MethodGetDiff, request, &rsp,
-		rpc.WithMaxPeerResponseTime(MaxGetDiffResponseTime),
-	)
-	if err != nil {
-		return nil, nil, err
-	}
-	return &rsp, pf, nil
+	rc          rpc.Client
+	mgr         rpc.PeerManager
+	fallbackMgr rpc.PeerManager
 }
 
 func (c *client) GetCheckpoints(ctx context.Context, request *GetCheckpointsRequest) ([]*Checkpoint, error) {
 	var rsp GetCheckpointsResponse
-	rsps, pfs, err := c.rcC.CallMulti(ctx, c.mgrC.GetBestPeers(), MethodGetCheckpoints, request, rsp)
+	rsps, pfs, err := c.rc.CallMulti(ctx, c.getBestPeers(), MethodGetCheckpoints, request, rsp)
 	if err != nil {
 		return nil, err
 	}
@@ -115,7 +100,8 @@ func (c *client) GetCheckpointChunk(
 	}
 
 	var rsp GetCheckpointChunkResponse
-	pf, err := c.rcC.CallOne(ctx, c.mgrC.GetBestPeers(opts...), MethodGetCheckpointChunk, request, &rsp,
+	peers := c.getBestPeers(opts...)
+	pf, err := c.rc.CallOne(ctx, peers, MethodGetCheckpointChunk, request, &rsp,
 		rpc.WithMaxPeerResponseTime(MaxGetCheckpointChunkResponseTime),
 	)
 	if err != nil {
@@ -124,32 +110,35 @@ func (c *client) GetCheckpointChunk(
 	return &rsp, pf, nil
 }
 
-// GetStorageSyncProtocolID returns unique storage sync protocol id for the specified chain context and runtime id.
-func GetStorageSyncProtocolID(chainContext string, runtimeID common.Namespace) core.ProtocolID {
-	return protocol.NewRuntimeProtocolID(chainContext, runtimeID, StorageSyncProtocolID, StorageSyncProtocolVersion)
+func (c *client) getBestPeers(opts ...rpc.BestPeersOption) []core.PeerID {
+	return append(c.mgr.GetBestPeers(opts...), c.fallbackMgr.GetBestPeers(opts...)...)
 }
 
-// NewClient creates a new storage sync protocol client.
+// NewClient creates a new checkpoint sync protocol client.
+//
+// Previously, it was part of the storage sync protocol. To enable seamless rolling
+// upgrades of the network, this client has a fallback to the old legacy protocol.
+// The new protocol is prioritized.
+//
+// Warning: This client only registers the checkpoint sync protocol with the P2P
+// service. To enable advertisement of the legacy protocol, it must be registered
+// separately.
 func NewClient(p2p rpc.P2P, chainContext string, runtimeID common.Namespace) Client {
-	// Use two separate clients and managers for the same protocol. This is to make sure that peers
-	// are scored differently between the two use cases (syncing diffs vs. syncing checkpoints). We
-	// could consider separating this into two protocols in the future.
-	pid := protocol.NewRuntimeProtocolID(chainContext, runtimeID, StorageSyncProtocolID, StorageSyncProtocolVersion)
+	pid := protocol.NewRuntimeProtocolID(chainContext, runtimeID, CheckpointSyncProtocolID, CheckpointSyncProtocolVersion)
+	fallbackPid := sync.GetStorageSyncProtocolID(chainContext, runtimeID)
+	rc := rpc.NewClient(p2p.Host(), pid, fallbackPid)
+	mgr := rpc.NewPeerManager(p2p, pid)
+	rc.RegisterListener(mgr)
 
-	rcC := rpc.NewClient(p2p.Host(), pid)
-	mgrC := rpc.NewPeerManager(p2p, pid)
-	rcC.RegisterListener(mgrC)
-
-	rcD := rpc.NewClient(p2p.Host(), pid)
-	mgrD := rpc.NewPeerManager(p2p, pid)
-	rcD.RegisterListener(mgrD)
+	// Fallback protocol requires a separate manager to manage peers that also support legacy protocol.
+	fallbackMgr := rpc.NewPeerManager(p2p, fallbackPid)
+	rc.RegisterListener(fallbackMgr)
 
 	p2p.RegisterProtocol(pid, minProtocolPeers, totalProtocolPeers)
 
 	return &client{
-		rcC:  rcC,
-		rcD:  rcD,
-		mgrC: mgrC,
-		mgrD: mgrD,
+		rc:          rc,
+		mgr:         mgr,
+		fallbackMgr: fallbackMgr,
 	}
 }
--- a/go/worker/storage/p2p/checkpointsync/client.go
+++ b/go/worker/storage/p2p/checkpointsync/client.go
@@ -120,9 +120,8 @@ func (c *client) getBestPeers(opts ...rpc.BestPeersOption) []core.PeerID {
 // upgrades of the network, this client has a fallback to the old legacy protocol.
 // The new protocol is prioritized.
 //
-// Warning: This client only registers the checkpoint sync protocol with the P2P
-// service. To enable advertisement of the legacy protocol, it must be registered
-// separately.
+// Finally, it ensures underlying p2p service starts tracking protocol peers
+// for both new and legacy protocol.
 func NewClient(p2p rpc.P2P, chainContext string, runtimeID common.Namespace) Client {
 	pid := protocol.NewRuntimeProtocolID(chainContext, runtimeID, CheckpointSyncProtocolID, CheckpointSyncProtocolVersion)
 	fallbackPid := synclegacy.GetStorageSyncProtocolID(chainContext, runtimeID)
@@ -135,6 +134,7 @@ func NewClient(p2p rpc.P2P, chainContext string, runtimeID common.Namespace) Cli
 	rc.RegisterListener(fallbackMgr)
 
 	p2p.RegisterProtocol(pid, minProtocolPeers, totalProtocolPeers)
+	p2p.RegisterProtocol(fallbackPid, minProtocolPeers, totalProtocolPeers)
 
 	return &client{
 		rc:          rc,
--- a/go/worker/storage/p2p/checkpointsync/client.go
+++ b/go/worker/storage/p2p/checkpointsync/client.go
@@ -10,7 +10,7 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/p2p/protocol"
 	"github.com/oasisprotocol/oasis-core/go/p2p/rpc"
 	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/checkpoint"
-	"github.com/oasisprotocol/oasis-core/go/worker/storage/p2p/sync"
+	"github.com/oasisprotocol/oasis-core/go/worker/storage/p2p/synclegacy"
 )
 
 const (
@@ -125,7 +125,7 @@ func (c *client) getBestPeers(opts ...rpc.BestPeersOption) []core.PeerID {
 // separately.
 func NewClient(p2p rpc.P2P, chainContext string, runtimeID common.Namespace) Client {
 	pid := protocol.NewRuntimeProtocolID(chainContext, runtimeID, CheckpointSyncProtocolID, CheckpointSyncProtocolVersion)
-	fallbackPid := sync.GetStorageSyncProtocolID(chainContext, runtimeID)
+	fallbackPid := synclegacy.GetStorageSyncProtocolID(chainContext, runtimeID)
 	rc := rpc.NewClient(p2p.Host(), pid, fallbackPid)
 	mgr := rpc.NewPeerManager(p2p, pid)
 	rc.RegisterListener(mgr)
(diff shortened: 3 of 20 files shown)


go/worker/storage/p2p/diff/protocol.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/libp2p/go-libp2p/core"
+	"github.com/oasisprotocol/oasis-core/go/common/node"
+	"github.com/oasisprotocol/oasis-core/go/common/version"
+	"github.com/oasisprotocol/oasis-core/go/p2p/peermgmt"
+	"github.com/oasisprotocol/oasis-core/go/p2p/protocol"
+	"github.com/oasisprotocol/oasis-core/go/storage/api"
+)
+
+// StorageDiffProtocolID is a unique protocol identifier for the storage diff protocol.
+const StorageDiffProtocolID = "storagediff"
```
kostko: One is `checkpoints` the other is `storagediff`? Maybe `checkpointsync` and `diffsync`? Or `storagecheckpoints` and `storagediffs`?

go/worker/storage/p2p/sync_test/interoperable_test.go
```diff
@@ -0,0 +1,395 @@
...
+	err = assertEqualCheckpoints(cps, want, getMeta)
+	require.NoError(err, "Assert expected checkpoints response")
+}
+
+func testNewClients(ctx context.Context, t *testing.T, host p2pApi.Service, backend storageApi.Backend) {
+	require := require.New(t)
+
+	// Test diff sync protocol.
+	diffClient := diffsync.NewClient(host, chainContext, runtimeID)
+	time.Sleep(1 * time.Second)
```
martintomazic: This is already flaky in the CI. Ideally there should be explicit synchronization or increase the timeout.

Let's see if we opt for e2e only with config.

go/worker/storage/p2p/sync_test/interoperable_test.go
```diff
@@ -0,0 +1,395 @@
...
+	defer clean1()
+
+	peer2, clean2 := mustStartNewPeer(t, dataDir, 2, backend, peerKind)
+	defer clean2()
+
+	err := peer1.Host().Connect(context.Background(), peer.AddrInfo{
+		ID:    peer2.Host().ID(),
+		Addrs: peer2.Host().Addrs(),
+	})
+	require.NoError(t, err, "Connecting host to peer")
```
martintomazic, line 103: This is bypassing correct discovery. E.g. this test won't fail if you forget to advertise yourself later in the integration part of the code.

Either move this test to e2e or have a discovery mock. The latter would require factoring out global configuration out of p2p2 service and making discovery parametric. 

A lot of our services use global config (in consistent manner) instead of having them parametric. Since this would be the first such change and would likely affect the style of other code ideally this should be separate PR.

go/worker/storage/p2p/checkpointsync/client.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
+	"github.com/oasisprotocol/oasis-core/go/p2p/protocol"
+	"github.com/oasisprotocol/oasis-core/go/p2p/rpc"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/checkpoint"
+	"github.com/oasisprotocol/oasis-core/go/worker/storage/p2p/synclegacy"
+)
+
+const (
+	// minProtocolPeers is the minimum number of peers from the registry we want to have connected
+	// for StorageSync protocol.
```
kostko:
```suggestion
	// for CheckpointSync protocol.
```

go/worker/storage/p2p/checkpointsync/client.go
```diff
@@ -0,0 +1,144 @@
...
+// separately.
+func NewClient(p2p rpc.P2P, chainContext string, runtimeID common.Namespace) Client {
+	pid := protocol.NewRuntimeProtocolID(chainContext, runtimeID, CheckpointSyncProtocolID, CheckpointSyncProtocolVersion)
+	fPid := synclegacy.GetStorageSyncProtocolID(chainContext, runtimeID)
+	rc := rpc.NewClient(p2p.Host(), pid, fPid)
+	mgr := rpc.NewPeerManager(p2p, pid)
+	rc.RegisterListener(mgr)
+
+	// Fallback protocol requires a separate manager to manage peers that also support legacy protocol.
+	fMgr := rpc.NewPeerManager(p2p, fPid)
```
kostko:
```suggestion
	fallbackMgr := rpc.NewPeerManager(p2p, fPid)
```

go/worker/storage/p2p/checkpointsync/client.go
```diff
@@ -0,0 +1,144 @@
...
+// Previously, it was part of the storage sync protocol. To enable seamless rolling
+// upgrades of the network, this client has a fallback to the old legacy protocol.
+// The new protocol is prioritized.
+//
+// Warning: This client only registers the checkpoint sync protocol with the P2P
+// service. To enable advertisement of the legacy protocol, it must be registered
+// separately.
+func NewClient(p2p rpc.P2P, chainContext string, runtimeID common.Namespace) Client {
+	pid := protocol.NewRuntimeProtocolID(chainContext, runtimeID, CheckpointSyncProtocolID, CheckpointSyncProtocolVersion)
+	fPid := synclegacy.GetStorageSyncProtocolID(chainContext, runtimeID)
```
kostko:
```suggestion
	fallbackPid := synclegacy.GetStorageSyncProtocolID(chainContext, runtimeID)
```

go/worker/storage/p2p/diffsync/client.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/oasisprotocol/oasis-core/go/common"
+	"github.com/oasisprotocol/oasis-core/go/p2p/protocol"
+	"github.com/oasisprotocol/oasis-core/go/p2p/rpc"
+	"github.com/oasisprotocol/oasis-core/go/worker/storage/p2p/synclegacy"
+)
+
+const (
+	// minProtocolPeers is the minimum number of peers from the registry we want to have connected
+	// for storage diff protocol.
```
kostko:
```suggestion
	// for diff sync protocol.
```

go/worker/storage/p2p/diffsync/client.go
```diff
@@ -0,0 +1,73 @@
...
+// Previously, it was part of the storage sync protocol. To enable seamless rolling
+// upgrades of the network, this client has a fallback to the old legacy protocol.
+// The new protocol is prioritized.
+//
+// Warning: This client only registers the diff sync protocol with the P2P
+// service. To enable advertisement of the legacy protocol, it must be registered
+// separately.
+func NewClient(p2p rpc.P2P, chainContext string, runtimeID common.Namespace) Client {
+	pid := protocol.NewRuntimeProtocolID(chainContext, runtimeID, DiffSyncProtocolID, DiffSyncProtocolVersion)
+	fPid := synclegacy.GetStorageSyncProtocolID(chainContext, runtimeID)
```
kostko:
```suggestion
	fallbackPid := synclegacy.GetStorageSyncProtocolID(chainContext, runtimeID)
```

go/worker/storage/p2p/diffsync/client.go
```diff
@@ -0,0 +1,73 @@
...
+	rc := rpc.NewClient(p2p.Host(), pid, fPid)
+	mgr := rpc.NewPeerManager(p2p, pid)
+	rc.RegisterListener(mgr)
+
+	// Fallback protocol requires a separate manager to manage peers that also support legacy protocol.
+	fMgr := rpc.NewPeerManager(p2p, fPid)
```
kostko:
```suggestion
	fallbackMgr := rpc.NewPeerManager(p2p, fPid)
```

go/worker/storage/p2p/sync_test/interoperable_test.go
kostko: Yeah ideally config would be passed explicitly, but everything would need to be refactored to be consistent. I would open a separate issue for this.

go/worker/storage/p2p/checkpointsync/protocol.go
```diff
@@ -0,0 +1,74 @@
...
+
+// GetCheckpointChunkResponse is a response to a GetCheckpointChunk request.
+type GetCheckpointChunkResponse struct {
+	Chunk []byte `json:"chunk,omitempty"`
+}
+
+func init() {
+	peermgmt.RegisterNodeHandler(&peermgmt.NodeHandlerBundle{
+		ProtocolsFn: func(n *node.Node, chainContext string) []core.ProtocolID {
+			if !n.HasRoles(node.RoleComputeWorker | node.RoleStorageRPC) {
```
kostko, line 68: Probably we should add observer nodes as well (also elsewhere).

go/worker/storage/p2p/sync_test/interoperable_test.go
```diff
@@ -0,0 +1,395 @@
...
+	cpVersion uint16
+	root      node.Root
+	hash1     hash.Hash = hashFromBytes([]byte("hash1"))
+	hash2     hash.Hash = hashFromBytes([]byte("hash2"))
+)
+
+func (bm *backendMock) GetCheckpoints(_ context.Context, request *checkpoint.GetCheckpointsRequest) ([]*checkpoint.Metadata, error) {
+	cpVersion = request.Version
+	root = node.Root{
+		// Existing bug: storagesync p2p server does not set namespace, nor does checkpoint.ChunkProvider validates it.
```
kostko: Not sure if this should be considered a bug, because the protocol is always per runtime so the namespace is implicit.

kostko:
> Likely solves root cause of https://github.com/oasisprotocol/oasis-core/issues/5750.

Note that as I see it currently, you always register for the checkpoint sync protocol, even if the checkpointer is disabled? In this case, this alone will not solve the above issue.

kostko:
> E2e test to be more confident in integration part, even though existing integration test was helpful at catching bugs.

Yeah that could become complicated fast and the e2e suite likely needs a major refresh so I would do it separately.

> Should new protocols alias legacy protocol definitions/methods given they are semantically equivalent?

I think copying is the safer option.

go/p2p/protocol/protocol.go
```diff
@@ -12,33 +12,32 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/p2p/api"
 )
 
-type protocolRegistry struct {
+// Registry is responsible for ensuring unique protocol ids.
```
peternose, line 15:
```suggestion
// Registry is responsible for ensuring unique protocol IDs.
```

go/worker/storage/p2p/sync_test/interoperable_test.go
martintomazic: You are right, I will simply ignore this like other part of the code do.

Possibly I could add a comment to the api request definition that this filed is redundant/not used.

martintomazic:
>Note that as I see it currently, you always register for the checkpoint sync protocol, even if the checkpointer is disabled? In this case, this alone will not solve the above issue.

Nice catch!

I guess correct solution to it is the _complementary solution_ from the https://github.com/oasisprotocol/oasis-core/issues/5750#issuecomment-3042912231
>[RegisterProtocol](https://github.com/oasisprotocol/oasis-core/blob/6ac1216230f21fbdcda720e27a12e95178f6e62f/go/p2p/peermgmt/peermgr.go#L173) in addition to triggering tracking and peer management for the given protocol (fetch part), is also responsible for triggering background [advertisement](https://github.com/oasisprotocol/oasis-core/blob/6ac1216230f21fbdcda720e27a12e95178f6e62f/go/p2p/peermgmt/peermgr.go#L191) (serve part). Ideally, advertisement would be [RegisterProtocolServer](https://github.com/oasisprotocol/oasis-core/blob/6ac1216230f21fbdcda720e27a12e95178f6e62f/go/p2p/p2p.go#L338) responsibility.
We could keep it simple and start advertising only after sufficient number of checkpoints. Likely we could generalize this and propagate generic readiness closure up to advertisement's [retry](https://github.com/oasisprotocol/oasis-core/blob/6ac1216230f21fbdcda720e27a12e95178f6e62f/go/p2p/peermgmt/discovery.go#L216).

^^ or independent and then only register if checkpointer is enabled.

**update**: This has to be independent method. E.g. registering light client protocol server should not start advertisement as peers are found here by other means.

go/worker/storage/p2p/sync_test/interoperable_test.go
kostko: That's a good idea.

go/p2p/rpc/client.go
```diff
@@ -612,15 +612,15 @@ func retryFn(ctx context.Context, fn func() error, maxRetries uint64, retryInter
 }
 
 // NewClient creates a new RPC client for the given protocol.
-func NewClient(h host.Host, p protocol.ID) Client {
+func NewClient(h host.Host, p protocol.ID, fallback ...protocol.ID) Client {
```
peternose, line 615: I think you need to accept `pids []protocol.ID` because go-libp2p implementation doesn't guarantee that the first ID will be used if possible.

go/worker/storage/p2p/sync/client.go
```diff
@@ -124,6 +124,11 @@ func (c *client) GetCheckpointChunk(
 	return &rsp, pf, nil
 }
 
+// GetStorageSyncProtocolID returns unique storage sync protocol id for the specified chain context and runtime id.
+func GetStorageSyncProtocolID(chainContext string, runtimeID common.Namespace) core.ProtocolID {
```
peternose, line 131: Maybe rename to `NewStorageSyncProtocolID` or `StorageSyncProtocolID`.

go/worker/storage/p2p/diffsync/server.go
```diff
@@ -0,0 +1,62 @@
...
+			return nil, rpc.ErrBadRequest
+		}
+
+		return s.handleGetDiff(ctx, &rq)
+	default:
+		return nil, rpc.ErrMethodNotSupported
+	}
+}
+
+func (s *service) handleGetDiff(ctx context.Context, request *GetDiffRequest) (*GetDiffResponse, error) {
```
peternose, line 30:
```suggestion
func (s *service) handleGetDiff(ctx context.Context, req *GetDiffRequest) (*GetDiffResponse, error) {
```

go/worker/storage/p2p/diffsync/server.go
```diff
@@ -0,0 +1,62 @@
...
+		if err != nil {
+			return nil, err
+		}
+		rsp.WriteLog = append(rsp.WriteLog, chunk)
+	}
+	return &rsp, nil
+}
+
+// NewServer creates a new storage diff protocol server.
+func NewServer(chainContext string, runtimeID common.Namespace, backend api.Backend) rpc.Server {
```
peternose, line 59: To avoid returning interface, I would create function `NewService` and let the caller create `rpc.Server`. Or create 2 functions.

go/worker/storage/p2p/diffsync/protocol.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/oasisprotocol/oasis-core/go/storage/api"
+)
+
+// DiffSyncProtocolID is a unique protocol identifier for the diff sync protocol.
+const DiffSyncProtocolID = "diffsync"
+
+// DiffSyncProtocolVersion is the supported version of the diff sync protocol.
+var DiffSyncProtocolVersion = version.Version{Major: 1, Minor: 0, Patch: 0}
+
+// Constants related to the GetDiff method.
```
peternose, line 29: I prefer individual comments, as these are public constants.

go/worker/storage/p2p/diffsync/client.go
```diff
@@ -0,0 +1,73 @@
...
+	return &rsp, pf, nil
+}
+
+// NewClient creates a new diff sync protocol client.
+//
+// Previously, it was part of the storage sync protocol. To enable seamless rolling
+// upgrades of the network, this client has a fallback to the old legacy protocol.
+// The new protocol is prioritized.
+//
+// Warning: This client only registers the diff sync protocol with the P2P
```
peternose: Warning in a constructor is a bit scary.

go/worker/storage/p2p/checkpointsync/server.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil, err
+	}
+
+	return &GetCheckpointsResponse{
+		Checkpoints: cps,
+	}, nil
+}
+
+func (s *service) handleGetCheckpointChunk(ctx context.Context, request *GetCheckpointChunkRequest) (*GetCheckpointChunkResponse, error) {
+	// TODO: Use stream resource manager to track buffer use.
```
peternose, line 53: Who will do this, do we have an issue for it? If not, this should be changed to a suggestion.

go/worker/storage/p2p/checkpointsync/server.go
```diff
@@ -0,0 +1,74 @@
+package checkpointsync
```
peternose, line 1: Should we move these two packages to a subfolder `sync`, e.g. `sync/checkpoint` and `sync/diff`?

go/worker/storage/p2p/checkpointsync/protocol.go
```diff
@@ -0,0 +1,74 @@
...
+
+// GetCheckpointsResponse is a response to a GetCheckpoints request.
+type GetCheckpointsResponse struct {
+	Checkpoints []*checkpoint.Metadata `json:"checkpoints,omitempty"`
+}
+
+// Constants related to the GetCheckpointChunk method.
+const (
+	MethodGetCheckpointChunk          = "GetCheckpointChunk"
+	MaxGetCheckpointChunkResponseTime = 60 * time.Second
```
peternose:
```suggestion
	MaxGetCheckpointChunkResponseTime = time.Minute
```

go/worker/storage/p2p/checkpointsync/protocol.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/oasisprotocol/oasis-core/go/common/node"
+	"github.com/oasisprotocol/oasis-core/go/common/version"
+	"github.com/oasisprotocol/oasis-core/go/p2p/peermgmt"
+	"github.com/oasisprotocol/oasis-core/go/p2p/protocol"
+	"github.com/oasisprotocol/oasis-core/go/storage/api"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/checkpoint"
+)
+
+// CheckpointSyncProtocolID is a unique protocol identifier for the checkpoint sync protocol.
+const CheckpointSyncProtocolID = "checkpointsync"
```
peternose, line 21: Can this, version, and methods be private?

go/worker/storage/p2p/checkpointsync/client.go
```diff
@@ -0,0 +1,144 @@
...
+	// minProtocolPeers is the minimum number of peers from the registry we want to have connected
+	// for StorageSync protocol.
+	minProtocolPeers = 5
+
+	// totalProtocolPeers is the number of peers we want to have connected for StorageSync protocol.
+	totalProtocolPeers = 10
+)
+
+// Client is a checkpoints sync protocol client.
+type Client interface {
```
peternose, line 26: Do we need this interface?

martintomazic:
Let's get #6270 before this one to also tackle https://github.com/oasisprotocol/oasis-core/pull/6262#issuecomment-3083721728 here.

Will handle remaining threads later on.

go/worker/storage/p2p/sync_test/interoperable_test.go
martintomazic: If we use cbor codec for grpc, omitting unused field should be both forward and backward compatible? So possibly we could remove it altogether (separate PR of course). Else I will just leave a comment.

kostko: What do you mean remove it? The same root structure is used in places where it needs the namespace.

kostko: I would just leave a comment for now.

martintomazic: >What do you mean remove it? The same root structure is used in places where it needs the namespace.

I meant removing the namespace from request (backend) not from root: https://github.com/oasisprotocol/oasis-core/blob/2bb1b3b9860b5dec20633cb32ddde6b61f7bb11e/go/storage/mkvs/checkpoint/checkpoint.go#L52
**Update**: or update all callsites to pass it, add new field to storage/checkpoint sync request and validate in the backend  (omitting validation in case of zero value not to be breaking).

Will go with comment for this PR as this would be out of scope. :)

go/worker/storage/p2p/checkpointsync/client.go
```diff
@@ -0,0 +1,144 @@
...
+// NewClient creates a new checkpoint sync protocol client.
+//
+// Previously, it was part of the storage sync protocol. To enable seamless rolling
+// upgrades of the network, this client has a fallback to the old legacy protocol.
+// The new protocol is prioritized.
+//
+// Warning: This client only registers the checkpoint sync protocol with the P2P
+// service. To enable advertisement of the legacy protocol, it must be registered
+// separately.
+func NewClient(p2p rpc.P2P, chainContext string, runtimeID common.Namespace) Client {
```
peternose, line 132: Move constructor after struct type definition.

go/worker/storage/p2p/checkpointsync/client.go
```diff
@@ -0,0 +1,144 @@
...
+	pid := protocol.NewRuntimeProtocolID(chainContext, runtimeID, CheckpointSyncProtocolID, CheckpointSyncProtocolVersion)
+	fallbackPid := sync.GetStorageSyncProtocolID(chainContext, runtimeID)
```
peternose: I would also prepare PR which will clean these things.

go/worker/storage/p2p/checkpointsync/client.go
```diff
@@ -0,0 +1,144 @@
...
+		rpc.WithMaxPeerResponseTime(MaxGetCheckpointChunkResponseTime),
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+	return &rsp, pf, nil
+}
+
+func (c *client) getBestPeers(opts ...rpc.BestPeersOption) []core.PeerID {
+	return append(c.mgr.GetBestPeers(opts...), c.fallbackMgr.GetBestPeers(opts...)...)
```
peternose, line 117: I'm not sure if this will work as expected. Maybe we should have just 2 clients, everyone with its own protocol ID and then we can switch to the legacy one in methods like `GetCheckpointChunk` if we don't get an answer from the first one.

go/worker/storage/p2p/checkpointsync/protocol.go
```diff
@@ -0,0 +1,74 @@
...
+type GetCheckpointsRequest struct {
+	Version uint16 `json:"version"`
+}
+
+// GetCheckpointsResponse is a response to a GetCheckpoints request.
+type GetCheckpointsResponse struct {
+	Checkpoints []*checkpoint.Metadata `json:"checkpoints,omitempty"`
+}
+
+// Constants related to the GetCheckpointChunk method.
```
peternose, line 46: Every constant should have its own comment.

go/worker/storage/p2p/checkpointsync/server.go
martintomazic: Existing comment, also most of code was copied. Will take comments into account and try to improve overall quality :)

martintomazic: Let's discuss all style comments that also affect other 4 p2p packages in one place to keep them consistent.
If minor happy to make them here, else let's sneak in one trivial code quality PR before this one.

>To avoid returning interface, I would create function NewService and let the caller create rpc.Server. Or create 2 functions. https://github.com/oasisprotocol/oasis-core/pull/6262#discussion_r2213186113

So we could create `rpc.Server` from the `rpc.Service` inside `RegisterProtocolServer`?  Probably better as adheres to "_return struct accept interfaces_" . `server.go` then becomes `service.go`, likely some comments need to be updated.


> Do we need this interface? https://github.com/oasisprotocol/oasis-core/pull/6262#discussion_r2213280559

Probably no. We can remove it and move comments to concrete structs. Interface would be only useful if structs would be used in multiple places and you would need interface in all this places due to e.g. mocks? In case of single usage it would still not make sense as interface should be then defined at the callsite.

> Move constructor after struct type definition. https://github.com/oasisprotocol/oasis-core/pull/6262#discussion_r2218467497

I prefer `New` at the bottom, majority of core code I have seen follows this pattern, even if the struct is on top.


> Can this, version, and methods be private? https://github.com/oasisprotocol/oasis-core/pull/6262#discussion_r2213276856

Probably yes, especially since we exposed `ProtocolID` in the other PR. No strong opinion.


> Every constant should have its own comment. https://github.com/oasisprotocol/oasis-core/pull/6262#discussion_r2218483024

As you said this should probably private. Normally we would leave the following comment:
```go
// MethodGetCheckpoints is the name of GetCheckpoint method.
```

No strong preference as this is opaque anyways.

> https://github.com/oasisprotocol/oasis-core/pull/6262#discussion_r2213178962

I would keep `request` to be consistent with this and other p2p packages.

go/p2p/protocol/protocol.go
martintomazic: All comments below use `id`. I think we use it 50:50 in our code but happy to change everything to uppercase which I also prefer.

go/p2p/rpc/client.go
martintomazic: Isn't this just syntactic sugar for slice? The main problem as you point out is that interface does not promise order.

In practice, [implementation](https://github.com/libp2p/go-libp2p/blob/7b7c3ed4ce048c99a865d4349b5f7484a6a93c04/p2p/host/basic/basic_host.go#L649:L655) favors protocols from left to right.

Ideally, we should not rely on this. If we had many methods as part of our protocol likely this would be the only sensible approach. But we don't, so maybe we can make it simpler.

go/worker/storage/p2p/checkpointsync/server.go
martintomazic: I would prefer not to. We should avoid using common names (checkpoint, diff) for package names since this causes clashes at the import site and need to rename.

go/worker/storage/p2p/checkpointsync/client.go
martintomazic: Why? I think the test confirms it works, but indeed this is probably simpler and thus better, given that we only have 3 methods to fallback manually. Will think a bit after rebasing this PR on top of #6270.

~In general this would also enable proper e2e testing (we could remove existing integration test). E2e test can be registered to not execute by default and used later as sanity again when removing support for both protocols.~

~Probably to make this work we should refactor existing storage and checkpoint sync  to avoid returning custom `Checkpoint` type that wraps metadata and slice of peer feedbacks, but instead return slice of metadata with slice of slices of peer feedback. [See](https://github.com/oasisprotocol/oasis-core/blob/462a7ebaa920f3c35c2253568df490d33f995520/go/worker/storage/p2p/checkpointsync/client.go#L39:L84). Or handle this type discrepancy on the call site. Uglier I would want but doable~

martintomazic:
Will consider simplifying with 2 clients again and then I tag you all for a final review :).

go/worker/storage/p2p/checkpointsync/server.go
peternose: > I prefer New at the bottom, majority of core code I have seen follows this pattern, even if the struct is on top.

I think this is a bad style, and I see that every day when I'm searching for constructors. Since the number of methods varies, the constructor should be at the top.  They do the same in Java.

> So we could create rpc.Server from the rpc.Service inside RegisterProtocolServer

This might not be the best idea, since someone might want to create a server without a service. So I would let the caller handle this. First, he needs to create a service, then server, and then register.

> I would keep request to be consistent with this and other p2p packages.

Is this true? Because I know that I always use `req` and `rsp` (both have 3 letters), and Jernej mostly uses `rq` and `rsp`. Strange to see `request` as whole word.

martintomazic: >Is this true?

Yes, all p2p packages are exact copies of each other in terms of style and code organization.

> I think this is a bad style

Probably. 

Checked stdlib and indeed ~70% in favour of the style you mention after a quick sample. Generally, for any style/standardization that is more restrictive that official go style I would _suggest updating contributing/style_ explaining _why_. Especially, if most of code uses various patterns and changing this.

Personally, I would prefer to keep the style and shortcomings of existing 5 packages (_all 7 comments above_). The style should be corrected as a separate PR depending on priority.

go/worker/storage/p2p/checkpointsync/client.go
martintomazic: Tried this in new PR https://github.com/oasisprotocol/oasis-core/pull/6277

PR #6263, opened 2025-07-11
go/runtime/registry/notifier: Refactor notifier

peternose:
These changes should run only one ROFL notifier for all ROFL components. Once all ROFL apps are updated and support `get_config` method, we can remove support for `RegisterNotify` and remove/decouple ROFL notifier from ROFL handler.

### DIFF
--- a/go/runtime/registry/handler.go
+++ b/go/runtime/registry/handler.go
@@ -40,6 +40,9 @@ type RuntimeHostHandlerEnvironment interface {
 
 	// GetRuntimeRegistry returns the runtime registry.
 	GetRuntimeRegistry() Registry
+
+	// GetROFLNotifier returns the ROFL notifier.
+	GetROFLNotifier() (*ROFLNotifier, error)
 }
 
 // RuntimeHostHandler is a runtime host handler suitable for compute runtimes. It provides the
@@ -124,7 +127,11 @@ func (h *runtimeHostHandler) Handle(ctx context.Context, rq *protocol.Body) (*pr
 func (h *runtimeHostHandler) NewSubHandler(comp *bundle.ExplodedComponent) (host.RuntimeHandler, error) {
 	switch comp.Kind {
 	case component.ROFL:
-		return newSubHandlerROFL(comp, h)
+		roflNotifier, err := h.env.GetROFLNotifier()
+		if err != nil {
+			return nil, err
+		}
+		return newSubHandlerROFL(comp, h, roflNotifier)
 	default:
 		return nil, fmt.Errorf("cannot create sub-handler for component '%s'", comp.Kind)
 	}
--- a/go/runtime/registry/handler_rofl.go
+++ b/go/runtime/registry/handler_rofl.go
@@ -36,13 +36,13 @@ type roflHostHandler struct {
 	comp  *bundle.ExplodedComponent
 	comps map[component.ID]host.Runtime
 
-	client        runtimeClient.RuntimeClient
-	eventNotifier *roflEventNotifier
+	client       runtimeClient.RuntimeClient
+	roflNotifier *ROFLNotifier
 
 	logger *logging.Logger
 }
 
-func newSubHandlerROFL(comp *bundle.ExplodedComponent, parent *runtimeHostHandler) (*roflHostHandler, error) {
+func newSubHandlerROFL(comp *bundle.ExplodedComponent, parent *runtimeHostHandler, roflNotifier *ROFLNotifier) (*roflHostHandler, error) {
 	client, err := parent.env.GetRuntimeRegistry().Client()
 	if err != nil {
 		return nil, err
@@ -53,13 +53,13 @@ func newSubHandlerROFL(comp *bundle.ExplodedComponent, parent *runtimeHostHandle
 		With("component_id", comp.ID())
 
 	return &roflHostHandler{
-		parent:        parent,
-		id:            comp.ID(),
-		comp:          comp,
-		comps:         make(map[component.ID]host.Runtime),
-		client:        client,
-		eventNotifier: newROFLEventNotifier(parent.runtime, client, logger),
-		logger:        logger,
+		parent:       parent,
+		id:           comp.ID(),
+		comp:         comp,
+		comps:        make(map[component.ID]host.Runtime),
+		client:       client,
+		roflNotifier: roflNotifier,
+		logger:       logger,
 	}, nil
 }
 
@@ -74,7 +74,7 @@ func (rh *roflHostHandler) AttachRuntime(id component.ID, rt host.Runtime) error
 	if id != rh.id {
 		return nil
 	}
-	return rh.eventNotifier.AttachRuntime(rt)
+	return nil
 }
 
 // getLocalStorageKey returns a properly namespaced version of the local storage key.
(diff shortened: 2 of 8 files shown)


go/runtime/registry/handler.go
```diff
@@ -40,6 +40,9 @@ type RuntimeHostHandlerEnvironment interface {
 
 	// GetRuntimeRegistry returns the runtime registry.
 	GetRuntimeRegistry() Registry
+
+	// GetROFLNotifier returns the ROFL notifier.
+	GetROFLNotifier() (*RoflNotifier, error)
```
kostko: Should the type be `ROFLNotifier` or should the method be `GetRoflNotifier`?

go/runtime/registry/notifier.go
```diff
@@ -61,526 +42,169 @@ const (
...
-
-	return &runtimeHostNotifier{
-		startOne:  cmSync.NewOne(),
-		runtime:   runtime,
-		host:      host,
-		consensus: consensus,
-		notifyCh:  make(chan *componentNotifyFuncWithQueue, notifyMainQueueSize),
-		logger:    logger,
-	}
+	notifyCh chan Notification
```
kostko: Why pass notifications by value?

go/runtime/registry/handler_rofl.go
kostko, line 1: We should also rename `host_rofl_mgmt.go` to `handler_rofl_mgmt.go`.

go/runtime/registry/notifier.go
```diff
@@ -61,526 +42,169 @@ const (
...
-				select {
-				case <-ctx.Done():
-				case <-time.After(keyManagerUpdateRetryInterval):
-					continue
-				}
-			}
-			break
-		}
-	})
+	notifyCh chan notifyFunc
```
kostko: Previously this was a ring channel, prioritizing the most recent notifications in case of slow consumers. This now prioritizes old notifications when new ones arrive which I think makes less sense?

go/runtime/registry/notifier_policy.go
```diff
@@ -0,0 +1,331 @@
...
+	nf := Notification{
+		comp:   compID,
+		queue:  queueKeyManagerStatus,
+		notify: notify,
+	}
+
+	if err := n.notifier.Queue(nf); err != nil {
+		n.logger.Error("failed to queue notification",
+			"component_id", nf.comp,
+			"queue0", nf.queue,
```
kostko:
```suggestion
			"queue", nf.queue,
```

go/runtime/registry/notifier_policy.go
```diff
@@ -0,0 +1,331 @@
...
+	nf := Notification{
+		comp:   compID,
+		queue:  queueKeyManagerQuotePolicy,
+		notify: notify,
+	}
+
+	if err := n.notifier.Queue(nf); err != nil {
+		n.logger.Error("failed to queue notification",
+			"component_id", nf.comp,
+			"queue0", nf.queue,
```
kostko:
```suggestion
			"queue", nf.queue,
```

go/runtime/registry/notifier_rofl.go
```diff
@@ -0,0 +1,337 @@
...
+		ctx, cancel := context.WithTimeout(ctx, roflNotifyTimeout)
+		defer cancel()
+
+		rsp, err := rr.Call(ctx, &protocol.Body{
+			RuntimeQueryRequest: &protocol.RuntimeQueryRequest{
+				Method: methodGetConfig,
+			},
+		})
+		if err != nil {
+			n.logger.Warn("failed to query config",
```
kostko, line 264: Should this really be a warning? AFAIK it will constantly trigger for any legacy runtimes that don't provide the implementation.

go/runtime/registry/notifier_rofl.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+// Notifications are notification settings.
+type Notifications struct {
+	/// Blocks subscribe to runtime block notifications.
+	Blocks bool `json:"blocks,omitempty"`
+	/// Events subscribe to runtime event notifications associated
+	/// with the specified tags.
+	Events [][]byte `json:"events,omitempty"`
+}
```
kostko: This seems like it belongs to `go/runtime/rofl/api`.

runtime/src/app/application.rs
```diff
@@ -62,8 +69,10 @@ pub trait App: Send + Sync {
 
     /// Called for runtime queries.
     async fn query(&self, method: &str, args: Vec<u8>) -> Result<Vec<u8>> {
-        // Default implementation rejects all requests.
-        bail!("method not supported");
+        match method {
+            METHOD_GET_CONFIG => Ok(cbor::to_vec(self.get_config())),
```
kostko: This is not ideal as the trait implementor must take care to provide this implementation in case one overrides the `query` method. I would rather that this is handled automatically by the caller of `query` and such "special" methods cannot be overriden.

go/runtime/registry/notifier_rofl.go
```diff
@@ -0,0 +1,337 @@
...
+	"github.com/oasisprotocol/oasis-core/go/runtime/host/protocol"
+	"github.com/oasisprotocol/oasis-core/go/runtime/transaction"
+)
+
+const (
+	// roflNotifyTimeout is the maximum amount of time runtime notification handling can take.
+	roflNotifyTimeout = 2 * time.Second
+
+	/// methodGetConfig is the name of the `get_config` method.
+	methodGetConfig = "get_config"
```
kostko: Should this be namespaced? Eg. `rofl.GetConfig` or similar.

Also it belongs to `go/runtime/rofl/api`.

go/runtime/registry/notifier_rofl.go
peternose: Yes, this will trigger for legacy ROFL components only few times and not constantly, until the component registers. After that, you will see no more warnings.

go/runtime/registry/notifier.go
peternose: To optimize things (copy is better than allocation) and in general it is better to pass values than references between routines. However, sending a pointer feels more natural, but I think it is not the recommended path.

go/runtime/registry/notifier_rofl.go
kostko: Ok.

runtime/src/app/application.rs
peternose: Agree. Moved to dispatcher. Could also create internal queries, custom RHP functions, .. but this might be just fine.

go/runtime/registry/notifier.go
peternose: Yes, this is not fine. Changed it back. However, the notifier will reject notifications if there are too many of them. Before, it blocked the caller, which can be a problem if the notifier is not running. Will leave it like this for now, and we can change it latter if needed.

go/runtime/registry/notifier.go
peternose: Changed to pointer.

go/runtime/registry/handler.go
peternose: The type should be `ROFLNotifier`, changed.

runtime/src/app/application.rs
```diff
@@ -62,8 +69,10 @@ pub trait App: Send + Sync {
 
     /// Called for runtime queries.
     async fn query(&self, method: &str, args: Vec<u8>) -> Result<Vec<u8>> {
-        // Default implementation rejects all requests.
-        bail!("method not supported");
+        match method {
+            METHOD_ROFL_GET_CONFIG => Ok(cbor::to_vec(self.get_config())),
+            _ => bail!("method not supported"),
+        }
```
kostko: This can now be removed?

go/runtime/registry/notifier.go
```diff
@@ -50,6 +30,9 @@ const (
 	queueKeyManagerStatus      = "key-manager/status"
 	queueKeyManagerQuotePolicy = "key-manager/quote-policy"
 	queueConsensusSync         = "consensus-sync"
+	queueROFLBlock             = "rofl/block"
+	queueROFLTags              = "rofl/tags"
+	queueROFLSettings          = "rofl/settings"
```
kostko: Should this be "config" instead of "settings"?

kostko (approved):

PR #6265, opened 2025-07-15
go/control: Show local checkpoint heights

### LINKED ISSUE #5823: control status: Show checkpoint height(s)
##### SUMMARY

Currently, there is no nice way to obtain the (latest?) checkpoint height that can be used for fast sync unless you already run a node and you check your `datadir/consensus/state/mkvs_storage.badger.db/checkpoints` folder. Add a list of available checkpoints to the `control status` output.

martintomazic:
Closes #5823.

Overlap with https://github.com/oasisprotocol/docs/pull/1363#discussion_r2238048832.

### DIFF
--- a/go/consensus/api/api.go
+++ b/go/consensus/api/api.go
@@ -375,6 +375,9 @@ type Status struct {
 
 	// P2P is the P2P status of the node.
 	P2P *P2PStatus `json:"p2p,omitempty"`
+
+	// Checkpoint is the checkpoints status of the node.
+	Checkpoint *CheckpointStatus `json:"checkpoint,omitempty"`
 }
 
 // P2PStatus is the P2P status of a node.
@@ -390,6 +393,14 @@ type P2PStatus struct {
 
 	// Peers is a list of node's peers.
 	Peers []string `json:"peers"`
+}
+
+// CheckpointStatus is the checkpoints status of the node.
+type CheckpointStatus struct {
+	// Heights is a list of consensus heights for which the node has checkpoints locally.
+	//
+	// Heights are sorted in descending order.
+	Heights []uint64 `json:"heights"`
 }
 
 // Service is an interface that a consensus backend service must provide.
--- a/go/consensus/cometbft/full/common.go
+++ b/go/consensus/cometbft/full/common.go
@@ -3,6 +3,7 @@ package full
 import (
 	"context"
 	"fmt"
+	"slices"
 	"sync"
 	"sync/atomic"
 
@@ -840,9 +841,36 @@ func (n *commonNode) GetStatus(ctx context.Context) (*consensusAPI.Status, error
 			// Failed to load validator set.
 			status.IsValidator = false
 		}
+
+		if n.Checkpointer() != nil {
+			status.Checkpoint = n.fetchCheckpointStatus(ctx)
+		}
 	}
 
 	return status, nil
+}
+
+// fetchCheckpointStatus fetches checkpoint status.
+//
+// In case of zero checkpoints or error a nil status is returned.
+func (n *commonNode) fetchCheckpointStatus(ctx context.Context) *consensusAPI.CheckpointStatus {
+	cps, err := n.mux.State().Storage().GetCheckpoints(ctx, &checkpoint.GetCheckpointsRequest{
+		Version: 1,
+	})
+	if err != nil {
+		n.Logger.Error("failed to fetch checkpoints status", "err", err)
+		return nil
+	}
+	var heights []uint64
+	for _, cp := range cps {
+		heights = append(heights, cp.Root.Version)
+	}
+	if len(heights) <= 0 {
+		return nil
+	}
+	slices.Sort(heights)
+	slices.Reverse(heights)
+	return &consensusAPI.CheckpointStatus{Heights: heights}
 }
 
 // Unimplemented methods.
--- a/go/consensus/cometbft/full/common.go
+++ b/go/consensus/cometbft/full/common.go
@@ -831,13 +831,14 @@ func (n *commonNode) GetStatus(ctx context.Context) (*consensusAPI.Status, error
 			valSetHeight = status.GenesisHeight
 		}
 		vals, err := n.stateStore.LoadValidators(valSetHeight)
-		if err != nil {
-			// Failed to load validator set.
-			status.IsValidator = false
-		} else {
+		switch err {
+		case nil:
 			consensusPk := n.identity.ConsensusSigner.Public()
 			consensusAddr := []byte(crypto.PublicKeyToCometBFT(&consensusPk).Address())
 			status.IsValidator = vals.HasAddress(consensusAddr)
+		default:
+			// Failed to load validator set.
+			status.IsValidator = false
 		}
 	}
 
--- a/.changelog/5265.feature.md
+++ b/.changelog/5265.feature.md
@@ -1,0 +1,6 @@
+go/oasis-node: Display consensus checkpoint heights
+
+A new field `heights` has been added to the `oasis-node control status` output
+under the `consensus.checkpoint` status section. If checkpointer is not
+disabled and there is at least one checkpoint, this field displays a list
+of local checkpoint heights.


go/consensus/cometbft/full/common.go
```diff
@@ -831,13 +831,29 @@ func (n *commonNode) GetStatus(ctx context.Context) (*consensusAPI.Status, error
...
+		case nil:
 			consensusPk := n.identity.ConsensusSigner.Public()
 			consensusAddr := []byte(crypto.PublicKeyToCometBFT(&consensusPk).Address())
 			status.IsValidator = vals.HasAddress(consensusAddr)
+		default:
+			// Failed to load validator set.
+			status.IsValidator = false
+		}
+
+		cps, err := n.mux.State().Storage().GetCheckpoints(ctx, &checkpoint.GetCheckpointsRequest{
```
kostko: So this currently only displays local checkpoints? It would be cool if some component could actually see what is available on the p2p network. Because if the node is not generating checkpoints, it will not show anything.

martintomazic: Hm but can you trust those checkpoints until validating them? I would say for the trusted height we should only trust local checkpoints or validated ones? But then this is long running process and should be separate from the control status?

_I already have some concerns with the current approach since to get local checkpoints we parse local dir. Alternatively checkpoints could be cached. The status is still fast and ideally we avoid that unless needed._

We could have trust height/hash status endpoint with sensible default if we want to get this data from the nodes that don't create checkpoints? So CI could query both?

kostko: The list of checkpoints from p2p is just to get a sense of what heights are available.

To get the trusted hash you would then query verified consensus block headers (locally).

kostko: Also note that the trusted height/hash should be BEFORE a checkpoint not at or after.

martintomazic: Ok I see. So a two step on the cli side. Will remove the hash part also and only return heights.

> It would be cool if some component could actually see what is available on the p2p network.

~After a quick look I think the main challenge is cometbft's state sync is internal, so to be able to query available snapshots from the p2p, we would need to set-up custom p2p client on our side.~

I guess we already have a `storagesync` (soon to be `checkpointsync`) `GetCheckpoints` method:
1. Consider making independent method (e.g. `GetUntrustedCheckpointHeights`) that should be exposed somewhere in the [client sdk](https://github.com/oasisprotocol/oasis-sdk/blob/3f1f6f4aa4a7800b1c176fea3cbb2faaba915ddb/client-sdk/go/connection/connection.go#L43-L52), likely part of consensus service.
    * If local checkpoints just return their local heights else delegate a call to p2p client.
2. Standalone service that tracks checkpoints.
     * Display them via status (no lag since cached) and no need for new method.
     * If you have checkpoints locally probably we don't even need this service.

martintomazic: > So this currently only displays local checkpoints? It would be cool if some component could actually see what is available on the p2p network. Because if the node is not generating checkpoints, it will not show anything.

_Do we really need this as it seems to complicate things: oasis CLI has all consensus parameters (debonding period, checkpoint frequency) available. Why not add a command that calculates/fetches recommended period, height and hash using parameters and consensus client that it has available locally?_

If this is not an option I would go in this direction:
1. Local and p2p checkpoints are two different things and as such should be reported differently.
    * CLI should first check local checkpoints and only then request p2p checkpoints.
2. Displaying local checkpoints as part of the node status seems reasonable, especially as this can be used for "debugging", e.g. "Am I producing checkpoints?"
3. P2p checkpoints: having additional component that caches available checkpoints, to display them under status (maybe different field) would be mis-using status a bit? Alternatively is to add additional rpc method which cli should use (yes latency)?

>Also note that the trusted height/hash should be BEFORE a checkpoint not at or after.

Is this true? I think I just synced using same height. Maybe you were thinking about runtime state sync that happens one height later? But for this one we are using history reindex as a form of "light client" so this is really not a problem.

go/consensus/api/api.go
```diff
@@ -375,6 +375,9 @@ type Status struct { // nolint: maligned
 
 	// P2P is the P2P status of the node.
 	P2P *P2PStatus `json:"p2p,omitempty"`
+
+	// CheckpointHeights is the list of node's local checkpoint heights.
+	CheckpointHeights []uint64 `json:"checkpoint_heights,omitempty"`
```
peternose: We could put this under `CheckpointStatus` or `SyncStatus`, as we might add additional fields latter. For example, we could display a list of local heights and a list of P2P heights.

go/consensus/cometbft/full/common.go
```diff
@@ -831,13 +831,26 @@ func (n *commonNode) GetStatus(ctx context.Context) (*consensusAPI.Status, error
...
+			Version: 1,
+		})
+		switch err {
```
peternose:
```suggestion
		if err != nil {
```

go/consensus/cometbft/full/common.go
peternose: Yes, this is true, as trusted height should be before the checkpoint so that you can do forward verification with bisection, which is fast.

I prefer adding this feature to the CLI. This way, the status only shows the status of the node, while the CLI can query the P2P network for available checkpoints.

martintomazic: >Also note that the trusted height/hash should be BEFORE a checkpoint not at or after.

Tried again and it worked. So AT works, after of course not.

As discussed today let's only display local checkpoints for now.

 We may add a method later for p2p checkpoints, although what is the value? We can only serve checkpoints current node sees via p2p client, but then the one using this heights will have different peers! So Probably just local checkpoints or even simpler doing some simple math on the cli side from consensus parameters gives you same guarantees :).

kostko: Yeah I agree, simpler is better.

go/consensus/cometbft/full/common.go
```diff
@@ -831,19 +832,46 @@ func (n *commonNode) GetStatus(ctx context.Context) (*consensusAPI.Status, error
...
 			consensusPk := n.identity.ConsensusSigner.Public()
 			consensusAddr := []byte(crypto.PublicKeyToCometBFT(&consensusPk).Address())
 			status.IsValidator = vals.HasAddress(consensusAddr)
+		default:
+			// Failed to load validator set.
+			status.IsValidator = false
+		}
+
+		if status.Checkpoint, err = n.fetchCheckpointStatus(ctx); err != nil {
+			return nil, fmt.Errorf("failed to fetch checkpoints: %w", err)
```
kostko: We should avoid failing the entire status result on errors as this can make it harder to diagnose issues.

go/consensus/cometbft/full/common.go
```diff
@@ -831,19 +832,45 @@ func (n *commonNode) GetStatus(ctx context.Context) (*consensusAPI.Status, error
...
+		case nil:
 			consensusPk := n.identity.ConsensusSigner.Public()
 			consensusAddr := []byte(crypto.PublicKeyToCometBFT(&consensusPk).Address())
 			status.IsValidator = vals.HasAddress(consensusAddr)
+		default:
+			// Failed to load validator set.
+			status.IsValidator = false
 		}
+
+		status.Checkpoint = n.fetchCheckpointStatus(ctx)
```
martintomazic: If you previously had checkpointer enabled (and you have some checkpoints), then you restart a node and disable checkpoints, you would still show old checkpoints. Could this confuse node operators?

go/consensus/cometbft/full/common.go
martintomazic: You are right. 

Do you want me to fix this for whole `GetStatus` method? Happy to do it here, but would like to agree on the general pattern so that remaining `GetStatus` functions ~section~ could be eventually refactored as well:
1. Probably status should take in pointer to a struct like is the case for json unmarshal?
2. I propose the struct is then populated, any errors encountered are joined (`errors.Join`) and returned.
    * It is the caller responsibility to decide what to do with an error (only log, but still display status).
    * Alternative is that status methods log, but then they don't have to return an error (not sure I like).
4. Currently  we log as `Error` (e.g. runtime status). I think this should be `Warning` or `Info` instead, since error should be reserved critical erros?

go/consensus/cometbft/full/common.go
kostko: Don't forget to also open a corresponding issue in the CLI.

go/consensus/cometbft/full/common.go
kostko: No, let's handle any further refactor separately.

go/consensus/cometbft/full/common.go
kostko: You could also show information on whether the checkpointer is currently enabled or not.

martintomazic: >You could also show information on whether the checkpointer is currently enabled or not.

You mean to only show it if the checkpointer is enabled? I think this would be good.

Alternative is to show the status of the checkpointer directly:
`consensus.checkpointer.enabled/heights` or possibly `consensus.state.checkpointer.enabled/heights`?

I notice for the consensus we tend to display "domain view" (current solution),  whereas for the runtimes we tend to show status of independent components (alternative solution).

martintomazic: I have kept is simple and only show it if checkpointer is enabled.

PR #6270, opened 2025-07-18
go/p2p: Ensure only server providers advertise themselves

martintomazic:
Tackles https://github.com/oasisprotocol/oasis-core/pull/6262#issuecomment-3083790722, see existing slack [discussion](https://oasisprotocol.slack.com/archives/C075WCGDNJ0/p1752760962203459).

_Only hosts that serve p2p protocol should advertise themselves via seed nodes for peer discovery._

### DIFF
--- a/go/p2p/api/api.go
+++ b/go/p2p/api/api.go
@@ -68,6 +68,11 @@ type Service interface {
 	Publish(ctx context.Context, topic string, msg any)
 
 	// RegisterHandler registers a message handler for the specified runtime and topic kind.
+	//
+	// In addition, it triggers advertisement of the host's readiness to serve the specified topic,
+	// allowing remote peers without existing connection to discover it.
+	//
+	// Ensure your server is indeed ready, to avoid advertising prematurely.
 	RegisterHandler(topic string, handler Handler)
 
 	// BlockPeer blocks a specific peer from being used by the local node.
@@ -83,6 +88,11 @@ type Service interface {
 	RegisterProtocol(p core.ProtocolID, minPeers int, totalPeers int)
 
 	// RegisterProtocolServer registers a protocol server for the given protocol.
+	//
+	// In addition, it triggers advertisement of the host's readiness to serve the specified protocol,
+	// allowing remote peers without existing connection to discover it.
+	//
+	// Ensure your server is indeed ready, to avoid advertising prematurely.
 	RegisterProtocolServer(srv rpc.Server)
 
 	// GetMinRepublishInterval returns the minimum republish interval that needs to be respected by
--- a/go/p2p/peermgmt/discovery.go
+++ b/go/p2p/peermgmt/discovery.go
@@ -100,6 +100,10 @@ func (d *peerDiscovery) startAdvertising(ns string) {
 	case d.advCh <- struct{}{}:
 	default:
 	}
+
+	d.logger.Debug("triggered protocol advertisement",
+		"protocol", ns,
+	)
 }
 
 // stopAdvertising stops advertising the given namespace.
--- a/go/p2p/peermgmt/peermgr.go
+++ b/go/p2p/peermgmt/peermgr.go
@@ -168,9 +168,9 @@ func (m *PeerManager) NumTopicPeers(topic string) int {
 	return len(m.pubsub.ListPeers(topic))
 }
 
-// RegisterProtocol starts tracking and managing peers that support the given protocol.
+// TrackProtocolPeers starts tracking and managing peers that support the given protocol.
 // If the protocol is already registered, its values are updated.
-func (m *PeerManager) RegisterProtocol(p core.ProtocolID, minPeers int, totalPeers int) {
+func (m *PeerManager) TrackProtocolPeers(p core.ProtocolID, minPeers int, totalPeers int) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
@@ -188,8 +188,6 @@ func (m *PeerManager) RegisterProtocol(p core.ProtocolID, minPeers int, totalPee
 	}
 
 	m.protocols[p] = &watermark{minPeers, totalPeers}
-	m.discovery.startAdvertising(string(p))
-
 	m.logger.Debug("protocol registered",
 		"protocol", p,
 		"min_peers", minPeers,
@@ -197,9 +195,9 @@ func (m *PeerManager) RegisterProtocol(p core.ProtocolID, minPeers int, totalPee
 	)
 }
 
-// RegisterTopic starts tracking and managing peers that support the given topic.
+// TrackTopicPeers starts tracking and managing peers that support the given topic.
 // If the topic is already registered, its values are updated.
-func (m *PeerManager) RegisterTopic(topic string, minPeers int, totalPeers int) {
+func (m *PeerManager) TrackTopicPeers(topic string, minPeers int, totalPeers int) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
@@ -217,7 +215,6 @@ func (m *PeerManager) RegisterTopic(topic string, minPeers int, totalPeers int) 
 	}
 
 	m.topics[topic] = &watermark{minPeers, totalPeers}
-	m.discovery.startAdvertising(topic)
 
 	m.logger.Debug("topic registered",
 		"topic", topic,
@@ -226,9 +223,23 @@ func (m *PeerManager) RegisterTopic(topic string, minPeers int, totalPeers int) 
 	)
 }
 
-// UnregisterProtocol stops managing peers that support the given protocol.
+// AdvertiseProtocol starts advertising readiness to serve the specified protocol.
+//
+// This enables remote peers without existing connection to find the host node.
+func (m *PeerManager) AdvertiseProtocol(p core.ProtocolID) {
+	m.discovery.startAdvertising(string(p))
+}
+
+// AdvertiseTopic starts advertising readiness to serve the specified topic.
+//
+// This enables remote peers without existing connection to find the host node.
+func (m *PeerManager) AdvertiseTopic(topic string) {
+	m.discovery.startAdvertising(topic)
+}
+
+// StopTrackingProtocolPeers stops managing peers that support the given protocol.
 // If the protocol is not registered, this is a noop operation.
-func (m *PeerManager) UnregisterProtocol(p core.ProtocolID) {
+func (m *PeerManager) StopTrackingProtocolPeers(p core.ProtocolID) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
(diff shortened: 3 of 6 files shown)


go/worker/common/committee/node.go
```diff
@@ -956,6 +956,7 @@ func NewNode(
 
 	// Register transaction sync service.
 	p2pHost.RegisterProtocolServer(txsync.NewServer(chainContext, runtime.ID(), n.TxPool))
+	p2pHost.AdvertiseProtocol(txsync.RuntimeProtocolID(chainContext, runtime.ID()))
```
martintomazic: This is a bit weird, previously only executor node would advertise itself via `txsync.NewClient`, but everyone else would still run a server. 

Now I am registering everyone. And then peer manager will ensure `min` number registered on consensus and total number of peers that are either on consensus or not (can be found via advertisement I am doing here).

**update:** observation is still the same.

go/p2p/api/api.go
```diff
@@ -83,8 +83,16 @@ type Service interface {
...
+	//
+	// Normally, registered protocol server should be also advertised (AdvertiseProtocol),
+	// so that remote peers find the host node.
 	RegisterProtocolServer(srv rpc.Server)
 
+	// AdvertiseProtocol starts advertising the host as available for serving the specified protocol.
+	//
+	// It should be called only after the protocol server is running and ready to serve.
+	AdvertiseProtocol(p core.ProtocolID)
+
```
martintomazic: Alternative is to not expose `AdvertiseProtocol` and automate this call as part of the `RegisterProtocolServer`.

Pros: Inline with `RegisterHandler` and generally less duplication/risk of forgetting to register advertisement.
Cons: Less control in the future, breaking SRP.

go/worker/storage/p2p/sync/protocol.go
```diff
@@ -20,6 +21,11 @@ const StorageSyncProtocolID = "storagesync"
 // StorageSyncProtocolVersion is the supported version of the storage sync protocol.
 var StorageSyncProtocolVersion = version.Version{Major: 2, Minor: 0, Patch: 0}
 
+// ProtocolID returns the runtime storage sync protocol ID.
+func ProtocolID(chainContext string, runtimeID common.Namespace) core.ProtocolID {
+	return protocol.NewRuntimeProtocolID(chainContext, runtimeID, StorageSyncProtocolID, StorageSyncProtocolVersion)
```
peternose, line 26: Can we make ID and version now private?

go/p2p/peermgmt/peermgr.go
```diff
@@ -188,15 +188,21 @@ func (m *PeerManager) RegisterProtocol(p core.ProtocolID, minPeers int, totalPee
...
 		"protocol", p,
 		"min_peers", minPeers,
 		"total_peers", totalPeers,
 	)
 }
 
+// AdvertiseProtocol starts advertising readiness to serve the specified protocol.
+func (m *PeerManager) AdvertiseProtocol(p core.ProtocolID) {
+	m.discovery.startAdvertising(string(p))
+	m.logger.Debug("triggered protocol advertisement",
```
peternose: This comment (a bit changed) should be in `startAdvertising`.

go/worker/common/committee/node.go
```diff
@@ -956,6 +956,7 @@ func NewNode(
 
 	// Register transaction sync service.
 	p2pHost.RegisterProtocolServer(txsync.NewServer(chainContext, runtime.ID(), n.TxPool))
```
peternose: Should we change the server to accept protocol ID or to return it via function call, so that we don't create it twice in the next step?

go/p2p/api/api.go
```diff
@@ -83,8 +83,16 @@ type Service interface {
...
 
 	// RegisterProtocolServer registers a protocol server for the given protocol.
+	//
+	// Normally, registered protocol server should be also advertised (AdvertiseProtocol),
+	// so that remote peers find the host node.
 	RegisterProtocolServer(srv rpc.Server)
 
+	// AdvertiseProtocol starts advertising the host as available for serving the specified protocol.
+	//
+	// It should be called only after the protocol server is running and ready to serve.
```
peternose: Is this true? I think you can call this method anytime, as it will only add protocol to the map of advertised namespaces, and when the discovery starts, the advertising begins.

go/p2p/api/api.go
peternose: Since every server will advertise, maybe it would be the best to not have separate function for this.

peternose (commented):
> This is not optimal: before that we should check all connected peers via m.host.Peerstore().GetProtocols(remoteID). This way we avoid round-trip to seed nodes. More importantly, we can reuse existing connections and only open a new stream for a new protocol.

Once we connect to a node, we always take into account all its protocols before connecting new ones. See `NumProtocolPeers`. So when we are out of nodes, we should discover them via seed nodes.

go/p2p/peermgmt/peermgr.go
martintomazic: Not sure I understand. I need to expose this method as part of the peermgr and thus comment. Or maybe you mean to improve doc of existing `startAdvertising`?

go/worker/storage/p2p/sync/protocol.go
martintomazic: After fixup this change is no longer needed.

It would be probably still desired as it makes code consistent with light client protocol file. Logically also feels part of the `protocol.go`.

I assume there is no client of this constants outside of core rn...? Any pref or we keep as is?

go/p2p/api/api.go
martintomazic: The idea was to warn future user to think twice if indeed is time to register it already. E.g. for the checkpoints protocol probably we should only register once you have at least one checkpoint ready to serve...

Let me know if the new comment is clearer or we omit.

go/p2p/peermgmt/peermgr.go
peternose: Sorry, I meant that logging doesn't belong here, as someone could call `startAdvertising` from somewhere else. So this method should be in `startAdvertising` or lower.

go/p2p/peermgmt/peermgr.go
```diff
@@ -188,15 +188,23 @@ func (m *PeerManager) RegisterProtocol(p core.ProtocolID, minPeers int, totalPee
...
 		"protocol", p,
 		"min_peers", minPeers,
 		"total_peers", totalPeers,
 	)
 }
 
+// AdvertiseProtocol starts advertising readiness to serve the specified protocol.
+//
+// This enables remote peers without existing connection to find the host node.
+func (m *PeerManager) AdvertiseProtocol(p core.ProtocolID) {
```
peternose: You should do the same for topics, i.e. you should also remove `startAdvertising` from `RegisterTopic`.

go/worker/keymanager/p2p/server.go
```diff
@@ -48,9 +49,13 @@ func (s *service) handleCallEnclave(ctx context.Context, request *CallEnclaveReq
 	}, nil
 }
 
+// RuntimeProtocolID returns the runtime protocol ID for the specified chain context and runtime ID.
+func RuntimeProtocolID(chainContext string, runtimeID common.Namespace) core.ProtocolID {
```
peternose: Why is this not named `ProtocolID`?

peternose:
> ^^ not an issue basic host runs identity protocol automatically.

What is identity protocol?

go/p2p/peermgmt/peermgr.go
```diff
@@ -198,8 +198,8 @@ func (m *PeerManager) RegisterProtocol(p core.ProtocolID, minPeers int, totalPee
 // AdvertiseProtocol starts advertising readiness to serve the specified protocol.
 //
 // This enables remote peers without existing connection to find the host node.
-func (m *PeerManager) AdvertiseProtocol(p core.ProtocolID) {
-	m.discovery.startAdvertising(string(p))
+func (m *PeerManager) AdvertiseProtocol(p string) {
```
peternose: Could also just add method `AdvertiseTopic`.

peternose (commented):
Looks good, will approve once commits are merged.

martintomazic:
>What is identity protocol?

https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/identify/id.go - responsible for keeping protobook up to date.

This is complementary to your answer https://github.com/oasisprotocol/oasis-core/pull/6270#pullrequestreview-3035597064 as to how seed nodes are last resource.

go/p2p/peermgmt/peermgr.go
martintomazic: Changed. There is also `UnregisterProtocol` and `UnregisterTopic`.

Question:  It is a bit annoying to split two additional methods given they were only used for testing in the first place. Maybe we only update the comment for the unregister part and point out it still does two tings? Possibly make them private.

Happy to  split the two methods is just ideally public methods are tested, and since we are not using them feels like redundant extra work. 

Currently, advertisement cancellation works even without this methods as your node stops sending heartbeats, thus seed nodes automatically remove you.

peternose: The discovery and peer manager also support protocol removal, although that’s not needed in our case (not sure why I included it earlier, maybe just for completeness). Previously, registering protocols and topics would both start advertising and initiate node discovery. Now, these actions are split into two separate methods: one for discovery and one for advertising. Therefore, if you modify the registration logic, you’ll also need to update the unregistration accordingly. It might also be helpful to consider renaming the `Register*` methods to reflect this change. Yes, it is annoying, but you should be consistent. You could also remove support for protocol removal (not sure if this is good, though).

martintomazic: Did all this for the peer manager.

Maybe we should now improve p2p service definition as well (`api.go`).

`RegisterHandler` (gossip) both tracks peers, registers a handler, and advertizes. I am fine with breaking SRP  here as was done for `RegisterProtocolServer` (registers server and advertizes) to avoid duplication and only update comment.

`RegisterProtocol` could be renamed to `TrackProtocolPeers` like was done in the manager, or we could remove this method and expand `PeerManager` interface adding  `TrackProtocolPeers` to it instead.

_I would maybe opt for only updating `RegisterHandler` comment for the start...?_

go/p2p/peermgmt/discovery.go
```diff
@@ -100,6 +100,10 @@ func (d *peerDiscovery) startAdvertising(ns string) {
 	case d.advCh <- struct{}{}:
 	default:
 	}
+
+	d.logger.Debug("triggered protocol advertisement",
```
peternose, line 104: This could also be optional as I just saw that we log `started advertising` when the running thread picks up a new namespace.

PR #6277, opened 2025-07-23
Split storage sync P2P protocol (manual fallback)

### LINKED ISSUE #5751: Separate storage checkpoint sync p2p protocol

martintomazic:
As discussed in #6262 it might be simpler to manually handle fallback to the legacy protocol.

Closes #5751.

Regarding the style, I would prefer to keep it consistent with other p2p packages even if not optimal: https://github.com/oasisprotocol/oasis-core/pull/6262#discussion_r2220496029

### DIFF
--- a/go/p2p/rpc/client.go
+++ b/go/p2p/rpc/client.go
@@ -364,6 +364,10 @@ func (c *client) CallMulti(
 ) ([]any, []PeerFeedback, error) {
 	c.logger.Debug("call multiple", "method", method)
 
+	if len(peers) == 0 {
+		return nil, nil, fmt.Errorf("no peers given to service the request")
+	}
+
 	co := NewCallMultiOptions(opts...)
 
 	// Prepare the request.
--- a/go/worker/common/p2p/txsync/client.go
+++ b/go/worker/common/p2p/txsync/client.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
-	"github.com/oasisprotocol/oasis-core/go/p2p/protocol"
 	"github.com/oasisprotocol/oasis-core/go/p2p/rpc"
 )
 
@@ -82,7 +81,7 @@ func (c *client) GetTxs(ctx context.Context, request *GetTxsRequest) (*GetTxsRes
 
 // NewClient creates a new transaction sync protocol client.
 func NewClient(p2p rpc.P2P, chainContext string, runtimeID common.Namespace) Client {
-	pid := protocol.NewRuntimeProtocolID(chainContext, runtimeID, TxSyncProtocolID, TxSyncProtocolVersion)
+	pid := ProtocolID(chainContext, runtimeID)
 	mgr := rpc.NewPeerManager(p2p, pid)
 	rc := rpc.NewClient(p2p.Host(), pid)
 	rc.RegisterListener(mgr)
--- a/go/worker/storage/committee/checkpoint_sync.go
+++ b/go/worker/storage/committee/checkpoint_sync.go
@@ -11,9 +11,11 @@ import (
 	"sync"
 	"time"
 
+	"github.com/oasisprotocol/oasis-core/go/p2p/rpc"
 	storageApi "github.com/oasisprotocol/oasis-core/go/storage/api"
 	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/checkpoint"
-	storageSync "github.com/oasisprotocol/oasis-core/go/worker/storage/p2p/sync"
+	"github.com/oasisprotocol/oasis-core/go/worker/storage/p2p/checkpointsync"
+	"github.com/oasisprotocol/oasis-core/go/worker/storage/p2p/synclegacy"
 )
 
 const (
@@ -55,7 +57,7 @@ type chunk struct {
 	*checkpoint.ChunkMetadata
 
 	// checkpoint points to the checkpoint this chunk originated from.
-	checkpoint *storageSync.Checkpoint
+	checkpoint *checkpointsync.Checkpoint
 }
 
 type chunkHeap struct {
@@ -101,12 +103,7 @@ func (n *Node) checkpointChunkFetcher(
 		defer cancel()
 
 		// Fetch chunk from peers.
-		rsp, pf, err := n.storageSync.GetCheckpointChunk(chunkCtx, &storageSync.GetCheckpointChunkRequest{
-			Version: chunk.Version,
-			Root:    chunk.Root,
-			Index:   chunk.Index,
-			Digest:  chunk.Digest,
-		}, chunk.checkpoint)
+		rsp, pf, err := n.fetchChunk(chunkCtx, chunk)
 		if err != nil {
 			n.logger.Error("failed to fetch chunk from peers",
 				"err", err,
@@ -117,7 +114,7 @@ func (n *Node) checkpointChunkFetcher(
 		}
 
 		// Restore fetched chunk.
-		done, err := n.localStorage.Checkpointer().RestoreChunk(chunkCtx, chunk.Index, bytes.NewBuffer(rsp.Chunk))
+		done, err := n.localStorage.Checkpointer().RestoreChunk(chunkCtx, chunk.Index, bytes.NewBuffer(rsp))
 		cancel()
 
 		switch {
@@ -157,7 +154,47 @@ func (n *Node) checkpointChunkFetcher(
 	}
 }
 
-func (n *Node) handleCheckpoint(check *storageSync.Checkpoint, maxParallelRequests uint) (cpStatus int, rerr error) {
+// fetchChunk fetches chunk using checkpoint sync p2p protocol client.
+//
+// In case of no peers or error, it fallbacks to the legacy storage sync protocol.
+func (n *Node) fetchChunk(ctx context.Context, chunk *chunk) ([]byte, rpc.PeerFeedback, error) {
+	rsp1, pf, err := n.checkpointSync.GetCheckpointChunk(
+		ctx,
+		&checkpointsync.GetCheckpointChunkRequest{
+			Version: chunk.Version,
+			Root:    chunk.Root,
+			Index:   chunk.Index,
+			Digest:  chunk.Digest,
+		},
+		&checkpointsync.Checkpoint{
+			Metadata: chunk.checkpoint.Metadata,
+			Peers:    chunk.checkpoint.Peers,
+		},
+	)
+	if err == nil { // if NO error
+		return rsp1.Chunk, pf, nil
+	}
+
+	rsp2, pf, err := n.legacyStorageSync.GetCheckpointChunk(
+		ctx,
+		&synclegacy.GetCheckpointChunkRequest{
+			Version: chunk.Version,
+			Root:    chunk.Root,
+			Index:   chunk.Index,
+			Digest:  chunk.Digest,
+		},
+		&synclegacy.Checkpoint{
+			Metadata: chunk.checkpoint.Metadata,
+			Peers:    chunk.checkpoint.Peers,
+		},
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+	return rsp2.Chunk, pf, nil
+}
+
+func (n *Node) handleCheckpoint(check *checkpointsync.Checkpoint, maxParallelRequests uint) (cpStatus int, rerr error) {
 	if err := n.localStorage.Checkpointer().StartRestore(n.ctx, check.Metadata); err != nil {
 		// Any previous restores were already aborted by the driver up the call stack, so
 		// things should have been going smoothly here; bail.
@@ -276,13 +313,11 @@ func (n *Node) handleCheckpoint(check *storageSync.Checkpoint, maxParallelReques
 	}
 }
 
-func (n *Node) getCheckpointList() ([]*storageSync.Checkpoint, error) {
+func (n *Node) getCheckpointList() ([]*checkpointsync.Checkpoint, error) {
 	ctx, cancel := context.WithTimeout(n.ctx, cpListsTimeout)
 	defer cancel()
 
-	list, err := n.storageSync.GetCheckpoints(ctx, &storageSync.GetCheckpointsRequest{
-		Version: 1,
-	})
+	list, err := n.fetchCheckpoints(ctx)
 	if err != nil {
 		n.logger.Error("failed to retrieve any checkpoints",
 			"err", err,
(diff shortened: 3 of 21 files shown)


go/p2p/rpc/client.go
```diff
@@ -364,6 +364,10 @@ func (c *client) CallMulti(
 ) ([]any, []PeerFeedback, error) {
 	c.logger.Debug("call multiple", "method", method)
 
+	if len(peers) == 0 {
+		return nil, nil, fmt.Errorf("no peers given to service the request")
+	}
+
```
martintomazic, line 370: Maybe this change is not even needed given that I check for empty slice. Possibly I could break some stuff that "wrongly" relied on empty slices/didn't handle errors properly.

Update: Not a problem `txsync` is retrying correctly.

go/oasis-test-runner/scenario/e2e/runtime/checkpoint_sync.go
```diff
@@ -0,0 +1,117 @@
...
+// CheckpointSync tests interoperability of the new checkpoint and diff sync
+// p2p protocols with the legacy storage sync p2p protocol.
+//
+// The test checks that hosts that serve both protocols are compatible
+// with clients that fallback to both.
+//
+// To simulate legacy host comment out fallback to the new protocols
+// inside storage committee worker and disable registration of new checkpoint
+// and diff sync protocols. This is not tested automatically as it would
+// further polute existing code and require additional config flags.
```
martintomazic: I tried all 6 permutations locally (also sanity failures).

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -51,11 +52,16 @@ func (cfg *CheckpointSyncConfig) Validate() error {
 	return nil
 }
 
+type checkpoint struct {
```
peternose: I don't see why is this change better.

go/worker/storage/p2p/checkpointsync/client.go
```diff
@@ -0,0 +1,126 @@
...
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+	return &rsp, pf, nil
+}
+
+// NewClient creates a new checkpoint sync protocol client.
+//
+// Moreover, it ensures underlying p2p service starts tracking protocol peers
```
peternose: Is this true, as I see only one `pid` in the constructor?

go/worker/storage/committee/node.go
```diff
@@ -128,7 +130,9 @@ type Node struct { // nolint: maligned
 
 	localStorage storageApi.LocalBackend
 
-	storageSync storageSync.Client
+	diffSync       diffsync.Client
+	checkpointSync checkpointsync.Client
+	legacySync     synclegacy.Client
```
peternose: I would still name it `storageSync`.

go/worker/storage/committee/node.go
```diff
@@ -272,9 +276,20 @@ func NewNode(
...
 
-	// Register storage sync service.
-	commonNode.P2P.RegisterProtocolServer(storageSync.NewServer(commonNode.ChainContext, commonNode.Runtime.ID(), localStorage))
-	n.storageSync = storageSync.NewClient(commonNode.P2P, commonNode.ChainContext, commonNode.Runtime.ID())
+	// Advertise and serve legacy storage sync protocol.
+	commonNode.P2P.RegisterProtocolServer(synclegacy.NewServer(commonNode.ChainContext, commonNode.Runtime.ID(), localStorage))
+	// Advertise and serve diff sync protocol server.
+	commonNode.P2P.RegisterProtocolServer(diffsync.NewServer(commonNode.ChainContext, commonNode.Runtime.ID(), localStorage))
+	// Advertise and serve checkpoint sync protocol server if checkpoints are enabled.
+	if checkInterval < mkvsCp.CheckIntervalDisabled {
```
peternose:
```suggestion
	if checkInterval != mkvsCp.CheckIntervalDisabled {
```

go/worker/storage/committee/node.go
```diff
@@ -272,9 +276,20 @@ func NewNode(
 		node:   n,
 	})
 
-	// Register storage sync service.
-	commonNode.P2P.RegisterProtocolServer(storageSync.NewServer(commonNode.ChainContext, commonNode.Runtime.ID(), localStorage))
-	n.storageSync = storageSync.NewClient(commonNode.P2P, commonNode.ChainContext, commonNode.Runtime.ID())
+	// Advertise and serve legacy storage sync protocol.
```
peternose: From here on there are too many `what` comments, I would remove them and optionally add a comment explaining why do we need to serve the legacy protocol.

go/worker/storage/committee/node.go
```diff
@@ -430,13 +445,29 @@ func (n *Node) fetchDiff(round uint64, prevRoot, thisRoot storageApi.Root) {
...
 		return
 	}
 	result.pf = pf
-	result.writeLog = rsp.WriteLog
+	result.writeLog = wl
+}
+
+// getDiff fetches writelog using diff sync p2p protocol client.
+//
+// In case of no peers, it fallbacks to the legacy storage sync protocol.
```
peternose: The first part is not true, as there could be another error.

go/worker/storage/committee/node.go
```diff
@@ -430,13 +445,29 @@ func (n *Node) fetchDiff(round uint64, prevRoot, thisRoot storageApi.Root) {
 	ctx, cancel := context.WithCancel(n.ctx)
 	defer cancel()
 
-	rsp, pf, err := n.storageSync.GetDiff(ctx, &storageSync.GetDiffRequest{StartRoot: prevRoot, EndRoot: thisRoot})
+	wl, pf, err := n.getDiff(ctx, prevRoot, thisRoot)
```
peternose, line 447:
```suggestion
	rsp, pf, err := n.diffSync.GetDiff(ctx, &diffsync.GetDiffRequest{StartRoot: prevRoot, EndRoot: thisRoot})
	if err != nil {
	        // Support legacy protocol.
		rsp, pf, err := n.legacySync.GetDiff(ctx, &synclegacy.GetDiffRequest{StartRoot: prevRoot, EndRoot: thisRoot})
		if err == nil {
			result.pf = pf
			result.writeLog = rsp.WriteLog
			return
		}
		result.err = err
		return
	}
	result.pf = pf
	result.writeLog = rsp.WriteLog
```
I prefer to write it this way, as it is easily to remove the legacy stuff afterwards.

go/worker/storage/config/config.go
```diff
@@ -20,6 +20,8 @@ type Config struct {
 	PublicRPCEnabled bool `yaml:"public_rpc_enabled,omitempty"`
 	// Disable initial storage sync from checkpoints.
 	CheckpointSyncDisabled bool `yaml:"checkpoint_sync_disabled,omitempty"`
+	// Disable serving legacy storage sync p2p protocol.
+	LegacySyncServerDisabled bool `yaml:"legacy_sync_server_disabled,omitempty"`
```
peternose: Adding a new config just for a test is not the best. Can we do the test without it?

go/oasis-test-runner/scenario/e2e/runtime/checkpoint_sync.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Make the first compute worker check for checkpoints more often.
+	f.ComputeWorkers[0].CheckpointCheckInterval = 1 * time.Second
+	// Configure runtime for storage checkpointing.
+	f.Runtimes[1].Storage.CheckpointInterval = 10
+	f.Runtimes[1].Storage.CheckpointNumKept = 10
+	f.Runtimes[1].Storage.CheckpointChunkSize = 1 * 1024
```
peternose:
```suggestion
	f.Runtimes[1].Storage.CheckpointChunkSize = 1024
```

go/oasis-test-runner/scenario/e2e/runtime/checkpoint_sync.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (sc *checkpointSync) Fixture() (*oasis.NetworkFixture, error) {
+	f, err := sc.Scenario.Fixture()
+	if err != nil {
+		return nil, err
+	}
+
+	// Make the first compute worker check for checkpoints more often.
+	f.ComputeWorkers[0].CheckpointCheckInterval = 1 * time.Second
```
peternose:
```suggestion
	f.ComputeWorkers[0].CheckpointCheckInterval = time.Second
```

go/oasis-test-runner/scenario/e2e/runtime/checkpoint_sync.go
```diff
@@ -0,0 +1,118 @@
...
+	// Configure runtime for storage checkpointing.
+	f.Runtimes[1].Storage.CheckpointInterval = 10
+	f.Runtimes[1].Storage.CheckpointNumKept = 10
+	f.Runtimes[1].Storage.CheckpointChunkSize = 1 * 1024
+	// Serve both legacy and new protocols.
+	for i := range f.ComputeWorkers {
+		f.ComputeWorkers[i].LegacySyncServerDisabled = false
```
peternose: I think that this doesn't do anything. And it looks like this settings is obsolete, as it is never set to `true` :thinking:

go/oasis-test-runner/scenario/e2e/runtime/checkpoint_sync.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/checkpoint"
+)
+
+// CheckpointSync tests interoperability of the new checkpoint and diff sync
+// p2p protocols with the legacy storage sync p2p protocol.
+//
+// The test checks that hosts that serve both protocols are compatible
+// with clients that fallback to both.
+//
+// To simulate legacy host comment out fallback to the new protocols
```
peternose: So this test does nothing and should be removed when we clean the legacy stuff?

go/worker/common/p2p/txsync/client.go
```diff
@@ -82,7 +81,7 @@ func (c *client) GetTxs(ctx context.Context, request *GetTxsRequest) (*GetTxsRes
 
 // NewClient creates a new transaction sync protocol client.
 func NewClient(p2p rpc.P2P, chainContext string, runtimeID common.Namespace) Client {
-	pid := protocol.NewRuntimeProtocolID(chainContext, runtimeID, TxSyncProtocolID, TxSyncProtocolVersion)
+	pid := ProtocolID(chainContext, runtimeID)
```
peternose, line 84: The last commit has unfinished message: `go/worker: Use ProtocolID helper to simplify (what?)` (probably code).

go/oasis-test-runner/scenario/e2e/runtime/checkpoint_sync.go
peternose: Nice, but should this test be merged to master?

go/worker/storage/committee/checkpoint_sync.go
martintomazic: This is needed because `checkpointsync` and `storagesync` both return custom checkpoint type, so ideally the worker should define its own type to handle/unify type mismatch.

After trying a bit this seemed like the cleanest solution but open to other suggestions.

go/worker/storage/committee/node.go
martintomazic: If possible I would keep `legacy` word especially if we omit comments as you suggest (agree). `legacyStorageSync` is a bit long thus this name.

go/worker/storage/committee/node.go
martintomazic: Probably, but at the same time it makes existing functions quite big, especially two relating to checkpoints. This makes code harder to read/follow (e.g. `fetchChunk`). 

I would prefer readability here, mostly because copying back in this case would be trivial and non-error prone?

go/worker/storage/config/config.go
martintomazic: I see your perspective and spent quite some time thinking about alternatives but this is actually the only way to enable testing. I would prefer to keep it but maybe we don't register e2e to the test runner by default?

Pros: 
1. It already enables us to do a canary deployment of nodes that only serve new protocol (useful feature actually).
2. Indeed I would be the one removing it, but generally a lot of time could pass in between, code can change a lot etc. If possible I would keep this and be nice to my future self.

go/oasis-test-runner/scenario/e2e/runtime/checkpoint_sync.go
martintomazic: Yes I would maybe suggest to set it to `true`? This way I already proof to future developer (me in this case) that when we remove support for legacy client, mixed clients (legacy at that time) will be interoperable.

go/worker/storage/committee/checkpoint_sync.go
peternose: But once we remove the legacy `storagesync`, the problem disappears. Meaning that you should have used `checkpointsync` checkpoint type and convert `storagesync` type to that one; or refactor `storagesync` to use `checkpointsync` checkpoint type instead.

go/worker/storage/committee/node.go
peternose: Name `legacySync` doesn't tell you much. So in this case, I would prefer `legacyStorageSync`. We also have similar long names like `deprecated...` to emphasize that this field/struct is deprecated and should not be used.

go/worker/storage/committee/node.go
peternose: Copying back is not totally trivial as you are removing an internal function and the diff will be huge, that's why I usually do it this way. But if you immediately prepare a PR that will remove this legacy stuff, this is also fine. And yes, `fetchChunk` may be harder to refactor to my proposal, but it might be possible. 

You can leave it like this, resolving.

go/worker/storage/config/config.go
peternose: > 1. It already enables us to do a canary deployment of nodes that only serve new protocol (useful feature actually).

Maybe you could deploy a custom branch to those nodes. 

> Indeed I would be the one removing it, but generally a lot of time could pass in between, code can change a lot etc.

Yes, you would be the one removing it, but the problem is that you cannot remove it cleanly if people start using this feature.

> If possible I would keep this and be nice to my future self.

I would probably use another branch for this test, and never merge it. If you merge this, you still need to add a description of this config field to the changelog.

peternose: If you really want it, you can rename it to `DebugLegacySyncServerDisabled` and use it together with `cmdFlags.DebugDontBlameOasis()`. See how other similar fields are used.

go/worker/storage/committee/node.go
martintomazic: I replaced this with `config.GlobalConfig.Storage.Checkpointer.Enabled ` which should be easier to read and more correct until I get rid of global config.

I realised this when factoring out checkpointer creation, so added additional commit for it.

I know you prefer exported methods on top and unexported on the bottom. I have intentionally keep it below the `New` to ease readability and avoid jumping since is logically part of the init/new. Probably this is the main reason why style does not enforce strict ordering of exported > private. If you disagree I am happy to change.

go/worker/storage/committee/node.go
martintomazic: Grouped service registration and client creation both under two short comments, to keep it consistent with the flavor of existing and other workers where we use comments (even if redundant) to break structure.

go/worker/storage/config/config.go
martintomazic: >If you really want it, you can rename it to DebugLegacySyncServerDisabled and use it together with cmdFlags.DebugDontBlameOasis(). See how other similar fields are used.

Nice! :) 

I created #6279 that does removal and am keeping this test there for now.

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -296,9 +331,36 @@ func (n *Node) getCheckpointList() ([]*storageSync.Checkpoint, error) {
...
+		return list1, nil
+	}
+
+	list2, err := n.legacyStorageSync.GetCheckpoints(ctx, &synclegacy.GetCheckpointsRequest{
+		Version: 1,
+	})
+	if err != nil {
+		return nil, err
+	}
+	var cps []*checkpointsync.Checkpoint
```
peternose, line 351:
```suggestion
	cps := make([]*checkpointsync.Checkpoint, 0, len(list2))
```
I usually always create a slice with the right capacity, if the number of elements is known.

go/worker/storage/committee/node.go
```diff
@@ -208,8 +212,38 @@ func NewNode(
...
+
+	// Create p2p protocol clients.
+	n.legacyStorageSync = synclegacy.NewClient(commonNode.P2P, commonNode.ChainContext, commonNode.Runtime.ID())
+	n.diffSync = diffsync.NewClient(commonNode.P2P, commonNode.ChainContext, commonNode.Runtime.ID())
+	n.checkpointSync = checkpointsync.NewClient(commonNode.P2P, commonNode.ChainContext, commonNode.Runtime.ID())
+
+	return n, nil
+}
+
+func (n *Node) newCheckpointer(commonNode *committee.Node, localStorage storageApi.LocalBackend) (checkpoint.Checkpointer, error) {
```
peternose:
```suggestion
func newCheckpointer(ctx context.Context, commonNode *committee.Node, localStorage storageApi.LocalBackend) (checkpoint.Checkpointer, error) {
```

peternose (approved):
Looks good to me :+1:

go/worker/storage/committee/checkpoint_sync.go
martintomazic: I try to avoid indexes unless many elements (performance), as I find the code harder to read with indexes, so would prefer to keep as is but also happy to change!

PR #6278, opened 2025-07-29
go/worker/storage/config: Default to parallel chunking algorithm

### LINKED ISSUE #6219: Integrate and release v2 checkpointing algorithm
Checkpoint version would be an optional configuration parameter. The default would then change across releases. From the perspective of client it doesn't really matter which checkpoint it is restoring from.

The following will have to be tackled:
* E2E test that shows interoperability of two versions.
* Likely we need fallback/extend peerset in case of missing checkpoints/chunks.
* Both old and new checkpoint should be requested equally to preserve load for the initial adopters.
* We may want to make number of subtrees a consensus parameter, just like chunk size. Maybe keep it simple for start...

martintomazic:
Closes #6219 

I run new chunking algorithm for a few days in the cloud, increasing checkpoint creation frequence 10x, and did a state sync from this node a few times.

This was done by having local and remote node both use new `checkpointsync` p2p protocol. Since this were the only two peers on the new protocol they automatically discovered each other.

### DIFF
--- a/go/worker/storage/config/config.go
+++ b/go/worker/storage/config/config.go
@@ -55,7 +55,7 @@ func DefaultConfig() Config {
 		Checkpointer: CheckpointerConfig{
 			Enabled:         false,
 			CheckInterval:   1 * time.Minute,
-			ParallelChunker: false,
+			ParallelChunker: true,
 		},
 	}
 }
--- a/.changelog/6219.internal.md
+++ b/.changelog/6219.internal.md
@@ -1,0 +1,1 @@
+go/worker/storage/config: Default to parallel checkpoint creation


go/worker/storage/config/config.go
```diff
@@ -55,7 +55,7 @@ func DefaultConfig() Config {
 		Checkpointer: CheckpointerConfig{
 			Enabled:         false,
 			CheckInterval:   1 * time.Minute,
-			ParallelChunker: false,
+			ParallelChunker: true,
```
martintomazic, line 58: Initially the plan was to first release it for the runtimes only and then in the cleaning PR remove legacy checkpointer and also default it for the consensus (if no issues with first release).

I just realised checkpointer config is for both consensus and runtimes so no way to distinct this. Simplest would be to hardcode consensus to always use old one if we want this approach. Probably best.

Finally, when later defaulting for the consensus, I would suggest to decrease it from `12` threads to `1` thread  (consensus) given that consensus has `100x` less state than runtimes so no need to parallelize there.

martintomazic:
As discussed today, we should wait with this one until next release and test this with few canary nodes first.

We could also lower [chunker threads for consensus checkpoints](https://github.com/oasisprotocol/oasis-core/blob/6fe595f4375d54de756520aeb1a1ac2d24eb3a67/go/consensus/cometbft/full/full.go#L78) (possibly to 1) given that consensus state  is significantly smaller than e.g. what we have for runtimes.

PR #6280, opened 2025-08-01
go/runtime/txpool: Add support for multiple transactions per sender

peternose:
~~The schedule queue was changed to support multiple transactions per sender.~~

~~The queue uses virtual transaction blocks to improve priority. For example, if a sender sends a transaction A with sequence number 1 and priority 0 and transaction B with sequence number 2 and priority 100, both transactions are put into a virtual block [A, B] with an average priority 50.~~

The main queue was changed to support multiple transactions per sender.

### DIFF
--- a/go/oasis-test-runner/scenario/e2e/runtime/runtime_txs.go
+++ b/go/oasis-test-runner/scenario/e2e/runtime/runtime_txs.go
@@ -1,0 +1,155 @@
+package runtime
+
+import (
+	"context"
+	"fmt"
+	"maps"
+	"slices"
+	"sync"
+	"time"
+
+	"golang.org/x/sync/errgroup"
+
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/env"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario"
+	"github.com/oasisprotocol/oasis-core/go/runtime/client/api"
+)
+
+// RuntimeTxs tests whether multiple transactions from the same sender
+// can be included in a single block.
+var RuntimeTxs scenario.Scenario = newRuntimeTxsImpl()
+
+type runtimeTxsImpl struct {
+	Scenario
+}
+
+func newRuntimeTxsImpl() scenario.Scenario {
+	return &runtimeTxsImpl{
+		Scenario: *NewScenario("runtime-txs", nil),
+	}
+}
+
+func (sc *runtimeTxsImpl) Clone() scenario.Scenario {
+	return &runtimeTxsImpl{
+		Scenario: *sc.Scenario.Clone().(*Scenario),
+	}
+}
+
+func (sc *runtimeTxsImpl) Fixture() (*oasis.NetworkFixture, error) {
+	return sc.Scenario.Fixture()
+}
+
+func (sc *runtimeTxsImpl) Run(ctx context.Context, _ *env.Env) error {
+	if err := sc.Net.Start(); err != nil {
+		return err
+	}
+
+	if err := sc.Net.ClientController().WaitReady(ctx); err != nil {
+		return err
+	}
+
+	// Queue transactions with higher nonces which should not be included
+	// in a block until transactions with lower nonces are submitted.
+	batch := map[string][]uint64{
+		"sender-1": {1},
+		"sender-2": {1, 2},
+		"sender-3": {1, 2, 3},
+	}
+	group, gctx := errgroup.WithContext(ctx)
+	group.Go(func() error {
+		_, err := sc.submitTxs(gctx, batch)
+		return err
+	})
+
+	// Wait a bit to be sure that transactions are waiting in the pool
+	// and are not included in a block.
+	time.Sleep(5 * time.Second)
+
+	// Queue missing transactions.
+	batch = map[string][]uint64{
+		"sender-0": {0},
+		"sender-1": {0},
+		"sender-2": {0},
+		"sender-3": {0},
+	}
+	rounds, err := sc.submitTxs(ctx, batch)
+	if err != nil {
+		return err
+	}
+
+	// Wait for the first batch to complete.
+	if err := group.Wait(); err != nil {
+		return err
+	}
+
+	// Verify that all transactions were included in a block.
+	total := 0
+	for _, round := range rounds {
+		txs, err := sc.Net.ClientController().RuntimeClient.GetTransactions(ctx, &api.GetTransactionsRequest{
+			RuntimeID: KeyValueRuntimeID,
+			Round:     round,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to fetch latest transactions: %w", err)
+		}
+		total += len(txs)
+	}
+	if total != 10 {
+		return fmt.Errorf("invalid number of transactions: %d", total)
+	}
+
+	// Verify state.
+	expectedValues := map[string]string{
+		"sender-0": "0",
+		"sender-1": "1",
+		"sender-2": "2",
+		"sender-3": "3",
+	}
+	for sender, expected := range expectedValues {
+		value, err := sc.submitKeyValueRuntimeGetQuery(ctx, KeyValueRuntimeID, sender, api.RoundLatest)
+		if err != nil {
+			return fmt.Errorf("failed to query key %s: %w", sender, err)
+		}
+		if value != expected {
+			return fmt.Errorf("unexpected value for key %s: got %s, want %s", sender, value, expected)
+		}
+	}
+
+	return nil
+}
+
+func (sc *runtimeTxsImpl) submitTxs(ctx context.Context, batch map[string][]uint64) ([]uint64, error) {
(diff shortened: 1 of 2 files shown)


go/runtime/txpool/local_queue.go
```diff
@@ -15,89 +17,62 @@ var (
 
 // localQueue is a "front of the line" area for txs from our own node. We also keep these txs in order.
```
ptrus, line 16: > We also keep these txs in order.

I suggest updating the comment as this is not true anymore.

go/runtime/txpool/schedule_queue.go
```diff
@@ -1,186 +1,578 @@
...
+			return fmt.Errorf("replacement transaction has lower priority: %s", raw.hash)
 		}
-		sq.removeLocked(etx)
+		q.removeLocked(tx, senderTxs)
 	}
 
-	sq.all[tx.Hash()] = tx
-	sq.bySender[tx.sender] = tx
-	sq.byPriority.ReplaceOrInsert(tx)
+	// If the sender has reach the limit, remove one of his transactions.
```
ptrus:
```suggestion
	// If the sender has reached the limit, remove one of his transactions.
```

go/runtime/txpool/schedule_queue.go
```diff
@@ -1,186 +1,578 @@
...
+	if len(senderTxs) >= q.limit {
+		var last *blockTransaction
+		for _, tx := range senderTxs {
+			if last == nil || last.raw.senderSeq < tx.raw.senderSeq {
+				last = tx
+			}
+		}
+		if last == nil || last.raw.senderSeq <= raw.senderSeq {
```
ptrus: `last == nil` can only happen if `q.limit == 0`. Maybe handle this case separately, to make this slightly clearer (or maybe we could require/ensure that limit is not ever 0?)

go/runtime/txpool/schedule_queue.go
```diff
@@ -1,186 +1,578 @@
...
+			return fmt.Errorf("sender queue is full")
+		}
+		q.removeLocked(last, senderTxs)
```
ptrus: It looks like we could remove a valid pending sender tx here, even if the new one isn’t accepted (e.g. due to the queue being full in the next step). Could this lead to some cases where a pending tx would be dropped unexpectedly without a new one being accpeted?

But actually, can that even happen? Because removing a tx here likely ensures the queue has space, so the capacity check below won’t reject the tx. Might be worth adding a comment to mention this implicit dependency.

go/runtime/txpool/schedule_queue.go
peternose: Yes, if `q.limit <= 0`. I thought it was clear. I do not want to require the `limit` to be over 0, as then I would need to do the same for the `capacity` and I need to throw an error from the constructor. If the caller sets the `limit` to 0, it was his decision. Therefore, I just treat 0 values the same way as "valid" ones.

I added comments to know that `limit` and `capacity` can be zero when rejecting a tx.

go/runtime/txpool/schedule_queue.go
peternose: > It looks like we could remove a valid pending sender tx here, even if the new one isn’t accepted (e.g. due to the queue being full in the next step).

Since the number of txs never exceeds the capacity, the new tx will always be accepted in the next step as we remove one tx here to make space for the new one.

> Could this lead to some cases where a pending tx would be dropped unexpectedly without a new one being accepted?

> But actually, can that even happen? Because removing a tx here likely ensures the queue has space, so the capacity check below won’t reject the tx. Might be worth adding a comment to mention this implicit dependency.

No, as described above. Added a comment and moved `if`s to a `switch`.

peternose:
I just figured out that one could prepare future transactions with very high priority which will occupy (the top of) the queue forever. As a consequence, the queue could return empty batches.

For some runtimes, this is also possible now, as we accept a transaction from a sender even if the `seqNum` is higher than `stateSeqNum`. For our runtimes this is not true, as the runtime's `check` will reject these kind of transactions.

Possible solutions:
- We are very strict and we demand that transactions start with the `stateSeqNum` and arrive in sequential order.
- When the queue is full, I remove the last transaction of the lowest priority block that has a gap to the previous one. If none exists, I try to remove one from other blocks iff the new tx will extend a block that starts with `stateSeqNum`.

go/runtime/txpool/schedule_queue.go
ptrus: Thanks, the `switch` makes this clear now 👍

go/runtime/txpool/main_queue.go
```diff
@@ -13,6 +13,10 @@ var (
 	_ RepublishableTransactionSource = (*mainQueue)(nil)
 )
 
+// maxTransactionsPerSender is the maximum number of transactions a single
+// sender can have queued at any given time.
+const maxTransactionsPerSender = 10
```
kostko: This is just the global hard limit and the runtime can still enforce a lower limit via transaction checks, right?

kostko:
> I just figured out that one could prepare future transactions with very high priority which will occupy (the top of) the queue forever. As a consequence, the queue could return empty batches.
> 
> For some runtimes, this is also possible now, as we accept a transaction from a sender even if the `seqNum` is higher than `stateSeqNum`. For our runtimes this is not true, as the runtime's `check` will reject these kind of transactions.
> 
> Possible solutions:
> 
> * We are very strict and we demand that transactions start with the `stateSeqNum` and arrive in sequential order.
> * When the queue is full, I remove the last transaction of the lowest priority block that has a gap to the previous one. If none exists, I try to remove one from other blocks iff the new tx will extend a block that starts with `stateSeqNum`.

Can you check what the [Go Ethereum transaction pool](https://github.com/ethereum/go-ethereum/blob/master/core/txpool/legacypool/legacypool.go) does? AFAIK they have explicit provisions for handling nonce gaps.

go/consensus/cometbft/db/badger/badger.go
```diff
@@ -65,8 +64,6 @@ func New(fn string, noSuffix bool) (dbm.DB, error) {
 
 	opts := badger.DefaultOptions(fn) // This may benefit from LSMOnlyOptions.
 	opts = opts.WithLogger(cmnBadger.NewLogAdapter(logger))
-	opts = opts.WithSyncWrites(false)
-	opts = opts.WithCompression(options.Snappy)
```
kostko, line 69: This seems like an unrelated change, why is this here?

go/oasis-test-runner/scenario/e2e/runtime/runtime_txs.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// Wait for transactions to be included in a block.
+	if err := group.Wait(); err != nil {
+		return err
+	}
+
+	// Verify that all transactions were included in a block.
+	txs, err := sc.Net.ClientController().RuntimeClient.GetTransactions(ctx, &api.GetTransactionsRequest{
+		RuntimeID: KeyValueRuntimeID,
+		Round:     api.RoundLatest,
```
kostko: This seems fragile? What if other blocks are emitted due to lag? We should probably get the inclusion round from tx submission and query an exact block?

kostko:
I assume the maximum amount of out-of-order transactions that can be queued per sender is defined by the runtime?

go/consensus/cometbft/db/badger/badger.go
peternose: Wanted to remove default settings. Will remove this change.

go/oasis-test-runner/scenario/e2e/runtime/runtime_txs.go
peternose: Good idea. Will do that for every sender.

peternose:
> I assume the maximum amount of out-of-order transactions that can be queued per sender is defined by the runtime?

Yes, I removed the limit because this can be defined by the runtime. We can add it latter if needed.

peternose (commented):
> Can you test this works correctly also with a runtime from the SDK?

Yes, will do that.

kostko (approved):

peternose:
> Can you test this works correctly also with a runtime from the SDK?

Tested `./tests/run-e2e.sh` locally by replacing binary `oasis-node` under `/tests/untracked/` and all Oasis Core imports `github.com/oasisprotocol/oasis-core/go@eff3d3905b63` in `go.mod` files. Tests pass. 

```bash
 oasis_core_25.5_linux_amd64 git:(peternose/trivial/tests-bump-core) ✗ pwd
/home/oa/oasis-sdk/tests/untracked/oasis_core_25.5_linux_amd64

➜  oasis_core_25.5_linux_amd64 git:(peternose/trivial/tests-bump-core) ✗ oasis-node --version
Software version: 25.5-giteff3d3905
Consensus:
  Consensus protocol version: 7.0.0
Runtime:
  Host protocol version:      5.1.0
  Committee protocol version: 5.0.0
Go toolchain version: 1.25.0
```

Will test again once we merge and release 25.6.

peternose:
> Will test again once we merge and release 25.6.

E2E tests pass also with 25.6.

PR #6285, opened 2025-08-06
go/consensus/cometbft/light: Use persistent store for stateless client

peternose:


peternose:
> Is there any pruning for the persistent store? 

Added pruning to all stores. We can add pruning config to the light client configuration later if we think it's necessary for node operators.

PR #6289, opened 2025-08-12
go/consensus/cometbft/abci/system: Change events root

peternose:


go/consensus/cometbft/abci/mux.go
```diff
@@ -515,6 +521,11 @@ func (mux *abciMux) ProcessProposal(req types.RequestProcessProposal) (resp type
 		}
 	}
 
+	// Validate system transactions included by the proposer.
+	if err := mux.validateSystemTxs(); err != nil {
+		panic(fmt.Errorf("proposed block has invalid system transactions: %w", err))
```
kostko: Should this just REJECT the proposal instead of panicking?

go/consensus/cometbft/abci/mux.go
```diff
@@ -808,11 +819,6 @@ func (mux *abciMux) EndBlock(req types.RequestEndBlock) types.ResponseEndBlock {
 		},
 	}
 
-	// Validate system transactions included by the proposer.
```
kostko, line 811: Just to confirm, is `EndBlock` ever called outside `PrepareProposal`/`ProcessProposal` (e.g. by non-validator clients)?

go/consensus/cometbft/abci/system.go
```diff
@@ -148,11 +167,37 @@ func (mux *abciMux) validateSystemTxs() error {
...
+
+	return merkle.RootHash(events), nil
+}
+
+func (mux *abciMux) computeResultsHash() []byte {
+	if !mux.state.ConsensusParameters().IsFeatureVersion(migrations.Version256) {
+		return nil
 	}
-	return merkle.RootHash(provableEvents), nil
+	return cmttypes.NewResults(mux.state.proposal.resultsDeliverTx).Hash()
```
kostko, line 207: Only `DeliverTx` events?

go/consensus/cometbft/abci/system.go
```diff
@@ -148,11 +167,37 @@ func (mux *abciMux) validateSystemTxs() error {
...
+		// Events root used to be computed only from provable events.
+		provable := mux.state.blockCtx.ProvableEvents
+		provableEvents := make([][]byte, len(provable))
+		for i, pe := range provable {
+			provableEvents[i] = cbor.Marshal(pe.ProvableRepresentation())
+		}
+		return merkle.RootHash(provableEvents), nil
+	}
+
+	// Events root is now computed from all emitted events.
```
kostko, line 186: I remember you mentioned missing domain separation for events?

go/consensus/cometbft/abci/mux.go
peternose: We cannot guarantee this as the method is implementation of cometbft `Application`. Will remove this commit and find another solution.

go/consensus/cometbft/abci/system.go
peternose: Yes, this is how [cometbft](https://github.com/oasisprotocol/cometbft/blob/v0.37.9-oasis/state/store.go#L382) works.

go/consensus/cometbft/abci/system.go
peternose: Yes, if we were using our structs (like it is done for provable events). But here. we are using cometbft type `Event` which has field `Type` which is unique for every app.

go/consensus/cometbft/abci/mux.go
peternose: Reverted this commit and changed how `proposal` is updated. Now, it is on every `BeginBlock`, `DeliverTx` and `EndBlock` call, so that at the end of `EndBlock` we have all data to compute events root and results hash.

go/consensus/cometbft/abci/system.go
kostko: How do you prove other events?

kostko (commented):
Would be good to validate this does not introduce any issues on Testnet and Mainnet.

go/consensus/cometbft/abci/system.go
peternose: Sorry, my answer was wrong. What I meant was that since `NewResults` strips all "non-deterministic" fields, no events can be verified using this hash. That holds also for events from `DeliverTx`. However, you can verify transaction results but only fields `Code, Data, GasWanted, GasUsed`.

To prove events, one must use our Merkle root, which will change if we change events. Observe that this root doesn't hold information in which phase of the block execution event occurred (begin, ..., end) as all current consumers don't care about that, and the implementation simplifies as we don't need to create root for every phase.

PR #6294, opened 2025-08-13
go: Remove tx hash from events

peternose:
Since no one is using the transaction hash of an event, we could simplify the code and make it faster by removing it. Does that make sense?

Hm, ... I think that nexus needs this :thinking:


ptrus:
I think most clients/indexers (for sure oasis-rosseta, nexus, oasis-web3-gateway, likely also others in the wild) use this to link events to transactions.

PR #6298, opened 2025-08-18
keymanager/src/policy: Stop saving policy locally

peternose:
I think that storing the key manager policy has no real effect, because:
- A compromised node operator can always remove it.
- It doesn’t improve performance, since the key manager policy is always verified against consensus during initialization.

This may simply be a legacy leftover.

### DIFF
--- a/keymanager/src/policy/cached.rs
+++ b/keymanager/src/policy/cached.rs
@@ -7,19 +7,11 @@ use std::{
 
 use anyhow::Result;
 use lazy_static::lazy_static;
-use sgx_isa::Keypolicy;
 use tiny_keccak::{Hasher, Sha3};
 
 use oasis_core_runtime::{
-    common::{
-        namespace::Namespace,
-        sgx::{
-            seal::{seal, unseal},
-            EnclaveIdentity,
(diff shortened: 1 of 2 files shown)


keymanager/src/policy/cached.rs
```diff
@@ -172,82 +152,51 @@ impl Policy {
...
 #[derive(Clone, Default, Debug)]
 struct CachedPolicy {
     pub checksum: Vec<u8>,
     pub serial: u32,
-    pub runtime_id: Namespace,
     pub may_query: HashMap<Namespace, HashSet<EnclaveIdentity>>,
     pub may_replicate: HashSet<EnclaveIdentity>,
     pub may_replicate_from: HashSet<EnclaveIdentity>,
-    pub master_secret_rotation_interval: EpochTime,
-    pub max_ephemeral_secret_age: EpochTime,
```
kostko, line 206: Was this unused?

peternose: Yes, this is only used by consensus and the key manager worker, to know when to schedule things.

PR #6299, opened 2025-08-19
go/worker/storage: Refactor state sync worker

martintomazic:
**Motivation**:
Whilst working on https://github.com/oasisprotocol/oasis-core/pull/6239, it became clear the runtime state sync could be optimized. More important;u the worker in the current form is very hard to read, maintain and reason about.

**What was done**:
This PR improves some of the existing issues, such as panicking, and unlikely go routine leaks upon termination/cleanup. More importantly it sets the stage for further refactors.

Please consider this as incremental improvement. The code after this refactor is still far from optimal in terms of readability and maintenance.

**Follow-up**:
1. Make diff sync independent worker, that could be tested and maintained in isolation.
2. Optimize state sync (if possible and sensible).


go/worker/storage/statesync/state_sync.go
```diff
@@ -362,6 +359,8 @@ func (w *Worker) fetchDiff(ctx context.Context, round uint64, prevRoot, thisRoot
 
 // getDiff fetches writelog using diff sync p2p protocol client.
 //
+// The request relies on the default timeout of the underlying p2p protocol clients.
+//
```
martintomazic: How about we move this constant out of the p2p protocol client and make timeout responsibility of the client?

go/worker/storage/statesync/state_sync.go
```diff
@@ -118,8 +120,21 @@ type finalizeResult struct {
...
+// In addition this worker is responsible for:
+//  1. Initializing the runtime state, possibly using checkpoints (if configured).
+//  2. Pruning the state as specified by the configuration.
+//  3. Optionally creating runtime state checkpoints (used by other nodes) for the state sync.
+//  4. Creating (and optionally advertising) statesync p2p protocol clients and servers.
+//  5. Registering node availability when it has synced sufficiently close to
+//     the latest known block header.
+//
+// Suggestion: This worker should not be responsible for creating and advertising p2p related stuff.
```
peternose, line 73: I would remove this since it’s already obvious from a "good programming" perspective. If this is a TODO, then the comment should be a bit different.

go/worker/storage/statesync/state_sync.go
```diff
@@ -300,100 +316,95 @@ func (n *Node) newCheckpointer(ctx context.Context, commonNode *committee.Node,
 
 // Service interface.
 
-// Name returns the service name.
-func (n *Node) Name() string {
-	return "committee node"
```
peternose: Why was this removed? Now , if we change common node's name, the name of this worker might be incorrect.

go/worker/storage/statesync/state_sync.go
```diff
@@ -755,155 +766,155 @@ func (n *Node) consensusCheckpointSyncer() {
...
-	case <-n.ctx.Done():
-		close(n.initCh)
+	case <-w.commonNode.Initialized():
+	case <-w.ctx.Done():
+		close(w.initCh)
 		return
 	}
 
-	n.logger.Info("starting committee node")
+	w.logger.Info("starting state sycne worker")
```
peternose:
```suggestion
	w.logger.Info("starting state sync worker")
```
I prefer only `starting`, as the logger should in general contain the name of the worker.

go/worker/storage/worker.go
```diff
@@ -44,14 +44,14 @@ func New(
...
-		runtimes:     make(map[common.Namespace]*committee.Node),
+		runtimes:     make(map[common.Namespace]*statesync.Worker),
 	}
 
 	if !enabled {
 		return s, nil
 	}
 
-	// Start storage node for every runtime.
+	// Start state sync worker for every runtime.
```
peternose, line 63: Can you improve this comment, as is doesn't make sense since we only register runtime.

go/worker/storage/statesync/prune.go
```diff
@@ -0,0 +1,48 @@
...
+	lastSycnedRound, _, _ := p.worker.GetLastSynced()
+
+	for _, round := range rounds {
+		if round >= lastSycnedRound {
+			return fmt.Errorf("worker/storage: tried to prune past last synced round (last synced: %d)",
+				lastSycnedRound,
+			)
+		}
+
+		// Old suggestion: Make sure we don't prune rounds that need to be checkpointed but haven't been yet.
```
peternose, line 26: I prefer move in one commit, and changes in another.

go/worker/storage/statesync/prune.go
```diff
@@ -0,0 +1,48 @@
...
+
+		p.logger.Debug("pruning storage for round", "round", round)
+
+		// Prune given block.
+		err := p.worker.localStorage.NodeDB().Prune(round)
+		switch err {
+		case nil:
+		case api.ErrNotEarliest:
```
peternose, line 34: I would keep import rename, as `api` is so general and is not the best name for a package.

go/worker/storage/worker.go
```diff
@@ -27,6 +30,9 @@ type Worker struct {
 	quitCh chan struct{}
 
 	runtimes map[common.Namespace]*statesync.Worker
+
+	ctx    context.Context
```
peternose, line 34: You should not [store the context inside a struct](https://pkg.go.dev/context#:~:text=Do%20not%20store%20Contexts%20inside%20a%20struct%20type%3B%20instead%2C%20pass%20a%20Context%20explicitly%20to%20each%20function%20that%20needs%20it.). However, since this is the top level service with `Start` and `Stop`, this is acceptable as this can be latter removed. However, all sub-services should implement `Service` interface and not store `ctx` locally.

go/worker/storage/worker.go
```diff
@@ -142,46 +152,53 @@ func (w *Worker) Start() error {
...
 
 	return nil
 }
 
+// Serve starts running state sync worker for every configured runtime.
+//
+// In case of an error from one of the state sync workers it cancels the remaining
+// ones and waits for all of them to finish. The error from the first worker
+// that failed is returned.
+func (w *Worker) Serve() error {
```
peternose, line 176: I would expect `Serve` to take `ctx` as a parameter.

go/worker/storage/statesync/state_sync.go
```diff
@@ -161,16 +159,12 @@ type Worker struct { // nolint: maligned
...
-	ctxCancel context.CancelFunc
-
-	quitCh chan struct{}
-
 	initCh chan struct{}
 }
 
 // New creates a new state sync worker.
 func New(
+	ctx context.Context,
```
peternose, line 111: This is not fine. You should aim to pass `ctx` to `Serve` function.

go/worker/storage/p2p/diffsync/client.go
```diff
@@ -20,6 +20,8 @@ const (
 type Client interface {
 	// GetDiff requests a write log of entries that must be applied to get from the first given root
 	// to the second one.
+	//
+	// The request times out in [MaxGetDiffResponseTime].
```
peternose, line 24: Avoid being too specific in an interface. Those details belong to the implementation, not the interface itself.

go/worker/storage/statesync/checkpoint_sync.go
```diff
@@ -486,6 +486,7 @@ func (w *Worker) syncCheckpoints(ctx context.Context, genesisRound uint64, wantO
 			}
 		}
 
+		// Suggestion: Limit the max time for restoring checkpoint.
```
peternose, line 489: Isn't this a TODO and could be done now?

go/worker/storage/statesync/state_sync.go
peternose: Which constant? This comment doesn't belong here, as a method should not rely on internal implementation of its parameters, only on their interface.

go/worker/storage/statesync/state_sync.go
```diff
@@ -1022,12 +1022,7 @@ mainLoop:
 				var oldBlock *block.Block
 				oldBlock, err = w.commonNode.Runtime.History().GetCommittedBlock(ctx, i)
 				if err != nil {
-					w.logger.Error("can't get block for round",
-						"err", err,
-						"round", i,
-						"current_round", blk.Header.Round,
-					)
-					panic("can't get block in storage worker")
+					return fmt.Errorf("getting block for round %d (current round: %d): %w", i, blk.Header.Round, err)
```
peternose: Should this be `failed to ...`

go/worker/storage/statesync/state_sync.go
```diff
@@ -604,45 +557,47 @@ func (w *Worker) nudgeAvailability(lastSynced, latest uint64) {
 	}
 }
 
-func (w *Worker) worker() { // nolint: gocyclo
-	defer close(w.quitCh)
+// Run runs state sync worker.
+func (w *Worker) Run(ctx context.Context) error { // nolint: gocyclo
```
peternose, line 252: Rename to `Serve` to implement `Service` interface.

go/worker/storage/statesync/state_sync.go
```diff
@@ -1022,12 +1022,7 @@ mainLoop:
 				var oldBlock *block.Block
 				oldBlock, err = w.commonNode.Runtime.History().GetCommittedBlock(ctx, i)
 				if err != nil {
-					w.logger.Error("can't get block for round",
-						"err", err,
-						"round", i,
-						"current_round", blk.Header.Round,
-					)
-					panic("can't get block in storage worker")
```
peternose: It would be nice to know why we panicked here and not return an error.

go/worker/storage/worker.go
```diff
@@ -142,46 +152,53 @@ func (w *Worker) Start() error {
...
 	}
 
-	// Wait for all runtimes to terminate.
 	go func() {
 		defer close(w.quitCh)
-
-		for _, r := range w.runtimes {
-			<-r.Quit()
-		}
+		_ = w.Serve() // error logged as part of Serve already.
```
peternose, line 157: You should move everything from `Start` and `Stop` to `Serve`, so that we can latter replace the service manager with the one that accepts `Service`s.

martintomazic: I see, I focused on making the common committee node stateless with regards to context, will also do explicit context passing for the storage worker here. +1.

go/worker/storage/statesync/state_sync.go
martintomazic: I think this was to point out impossible scenario, that is really not that impossible to happen.

>Imagine a scenario where you run a paratime node and you have to run oasis-node unsafe-reset. After that, you sync consensus again via checkpoint sync.
Unfortunately, it could happen that the checkpoint height (and thus earliest block header) is greater than you latest runtime state height. Therefore, the storage state sync worker will panic as it will encounter a missing runtime block header.
In this case you are left in a very annoying situation, as you have to either 1. use consensus snapshots provided by devops, or 2. you are forced to sync consensus from genesis just to fill the missing gap of block headers (likely a very small one).
A [realistic scenario](https://oasisprotocol.slack.com/archives/C075WCGDNJ0/p1754561274560769?thread_ts=1754561251.386359&cid=C075WCGDNJ0).

go/worker/storage/statesync/state_sync.go
martintomazic: https://github.com/oasisprotocol/oasis-core/blob/0ff844e85cebaa8d52218dcd71b02eb7e90e8aaa/go/worker/storage/p2p/diffsync/client.go#L34

I suggest to remove `MaxGetDiffResponseTime` from the p2p package and make client responsibility to define the context timeout. This is how is currently done for fetching chunks and is imo also idiomatic/correct.

Then I can remove all the comments that about this constant that you correctly pointed out were off?

go/worker/storage/statesync/checkpoint_sync.go
martintomazic: I thought about this yes. 

Any idea what would be a good context timeout that we don't shoot ourselves in the foot if the state becomes bigger and bigger and thus restoration longer... 

Definitely would prefer to not make this configurable, but we should have sufficient extra time. I think currently it takes around 10-20 min on my machine to restore from the checkpoint...

go/worker/storage/statesync/state_sync.go
martintomazic: Yes this is annoying, to make this work I would also have to refactor ` checkpoint.NewCheckpointer` to not require `ctx` on creation but split it in `NewCheckpointer` and `Serve`. 

Do you want me to do this here also? Imo makes sense...

go/worker/storage/statesync/state_sync.go
peternose: Yes, I guess you could remove it. But current solution is also fine in general, the p2p layer could define its own timeout and state sync could lower it with context if needed. 

Talking about internal implementation of another struct in comment is probably not the best, so I would remove it. Or instead of the comment, I would use context with deadline and use a similar timeout.

go/worker/storage/statesync/checkpoint_sync.go
peternose: I have no idea. You can start with a loose timeout and we can stricken it latter.

go/worker/storage/statesync/checkpointer.go
```diff
@@ -0,0 +1,215 @@
...
+//
+// The reason why we do this is to make it faster for storage nodes that use consensus state sync
+// to catch up as exactly the right checkpoint will be available.
+func (w *Worker) createCheckpoints(ctx context.Context) {
+	consensusCp := w.commonNode.Consensus.Checkpointer()
+	if consensusCp == nil {
+		return
+	}
+
+	// Wait for the common node to be initialized.
```
peternose, line 90: This can be removed.

go/worker/storage/statesync/state_sync.go
```diff
@@ -684,100 +676,46 @@ func (w *Worker) Run(ctx context.Context) error { // nolint: gocyclo
...
-	var wg sync.WaitGroup
+func (w *Worker) sync(
+	ctx context.Context,
+	lastFinalizedRound uint64,
+) error {
 	syncingRounds := make(map[uint64]*inFlight)
 	summaryCache := make(map[uint64]*blockSummary)
-	// Create the fetcher pool.
+	pendingApply := &minRoundQueue{}
+	pendingFinalize := &minRoundQueue{} // Suggestion: slice would suffice given that application must happen in order.
```
peternose: This comment appears to make sense only from the refactorer's perspective, and may be confusing to others. I recommend removing it or addressing the slice change in a separate commit.

go/worker/storage/statesync/state_sync.go
```diff
@@ -684,100 +676,46 @@ func (w *Worker) Run(ctx context.Context) error { // nolint: gocyclo
...
-					wg.Add(1)
-					fetchPool.Submit(func() {
-						defer wg.Done()
-						w.fetchDiff(ctx, this.Round, prevRoots[i], this.Roots[i])
-					})
-				}
-			}
-		}
-	}
+	heartbeat := heartbeat{}
```
peternose:
```suggestion
var heartbeat heartbeat
```

go/worker/storage/statesync/state_sync.go
```diff
@@ -788,32 +726,7 @@ mainLoop:
...
-				default:
-					w.logger.Error("can't apply write log",
-						"err", err,
-						"old_root", lastDiff.prevRoot,
-						"new_root", lastDiff.thisRoot,
-					)
-					lastDiff.pf.RecordSuccess()
-				}
-			}
+			err := w.apply(ctx, lastDiff)
```
peternose: Move 1 line down and into `if`.

go/worker/storage/statesync/state_sync.go
```diff
@@ -832,10 +745,12 @@ mainLoop:
 			delete(summaryCache, lastDiff.round-1)
 			lastFullyAppliedRound = lastDiff.round
 
+			// Suggestion: Rename to lastAppliedRoundMetric, as synced is often synonim for finalized in this code.
```
peternose: These suggestions are the same as TODO. Either you do them or not.

go/worker/storage/statesync/state_sync.go
```diff
@@ -845,81 +760,39 @@ mainLoop:
...
-				oldBlock, err = w.commonNode.Runtime.History().GetCommittedBlock(ctx, i)
-				if err != nil {
-					return fmt.Errorf("getting block for round %d (current round: %d): %w", i, blk.Header.Round, err)
-				}
-				summaryCache[i] = summaryFromBlock(oldBlock)
-			}
-			if _, ok := summaryCache[blk.Header.Round]; !ok {
-				summaryCache[blk.Header.Round] = summaryFromBlock(blk)
-			}
+			// Fixme: If block channel has many pending blocks (e.g. after checkpoint sync),
```
peternose: Can't we just move this to `triggerRoundFetches` and nudge when a round processing is done.

go/worker/storage/statesync/state_sync.go
```diff
@@ -928,56 +801,158 @@ mainLoop:
...
+	if _, ok := summaryCache[blk.Header.Round]; !ok {
+		summaryCache[blk.Header.Round] = summaryFromBlock(blk)
+	}
 	return nil
 }
 
-func (w *Worker) fetchDiff(ctx context.Context, round uint64, prevRoot, thisRoot storageApi.Root) {
+func (w *Worker) triggerRoundFetches(
+	ctx context.Context,
+	wg *sync.WaitGroup,
```
peternose: I avoid passing a wait groups and channels around because it makes it unclear which parts of the code are blocking/using it.

I would try to create two fetchers, a rounder and a differ, which would accept tasks and do their job. The rounder would internally use wait group, while the differ would expose diff channel.

go/worker/storage/statesync/diff_sync.go
```diff
@@ -0,0 +1,462 @@
...
+
+// syncDiffs is responsible for fetching, applying and finalizing storage diffs
+// as the new runtimes block headers arrive from the consensus service.
+//
+// In addition, it is also responsible for updating availability of the registration
+// service and notifying block history and checkpointer of the newly finalized rounds.
+//
+// Suggestion: Ideally syncDiffs is refactored into independent worker and made only
+// responsible for the syncing.
+func (w *Worker) syncDiffs(
```
peternose, line 81: While it is nice to separate diff and state, I don't prefer having worker functions scattered across multiple files. Having a sub-workers for this would solve the problem, and the main worker would just delegate tasks.

go/worker/storage/statesync/diff_sync.go
```diff
@@ -93,6 +93,8 @@ func (w *Worker) syncDiffs(
 	fetchPool := workerpool.New("storage_fetch/" + w.commonNode.Runtime.ID().String())
 	fetchPool.Resize(config.GlobalConfig.Storage.FetcherCount)
 	defer fetchPool.Stop()
+	fetchCtx, cancel := context.WithCancel(ctx)
```
peternose, line 96: Shouldn't we just do this on the top `ctx, cancel := context.WithCancel(ctx)`?

go/worker/storage/statesync/diff_sync.go
```diff
@@ -330,9 +329,7 @@ func (w *Worker) triggerRoundFetches(
 			rootType := prevRoots[i].Type
 			if !syncing.outstanding.contains(rootType) && syncing.awaitingRetry.contains(rootType) {
 				syncing.scheduleDiff(rootType)
-				wg.Add(1)
```
peternose: I don't like leaving go routines to be killed once the program exists as they might corrupt something.

go/worker/storage/statesync/state_sync.go
martintomazic: No because nudge should also deregister a node. So this should be also called everytime a new block arrives to see if we should deregister.

go/worker/storage/statesync/state_sync.go
martintomazic: >I would try to create two fetchers, a rounder and a differ, which would accept tasks and do their job. The rounder would internally use wait group, while the differ would expose diff channel.

Can you elaborate a bit more?

The final refactor I had in mind was done here: https://github.com/oasisprotocol/oasis-core/pull/6242/commits/c7f230b7f54e318fc18403d1a74e3f91dafd445c. See 3x `g.Go(...)`. Arguably could be improved further.

If you can write few sentences about how many workers and high level responsibilities of each that would be awesome! :)

go/worker/storage/statesync/diff_sync.go
martintomazic: >Even though you refactored the state sync worker, I think that there is still a lot of things to be done as the code is still very unclear in functions are way too long. I would try to break the worker into sub-workers, to break down the code into smaller pieces which are easier to understand.

I guess its mostly about getting there incrementally. Anyways, I see your point also so there is a proposal below.


>While it is nice to separate diff and state,  I don't prefer having worker functions scattered across multiple files.

Not sure I agree with this else why we already have `checkpoint_sync.go` and `checkpointer.go` files. I do find the code organised in this way (even if part of the same worker) way easier to navigate and reason.

Ideally, we should also factor out state initialization out of the main worker into separate file (or at least function). Finally `CheckpointSyncRetry` could be moved to `checkpoint_sync.go`. But this would really be out of scope, as I am focusing on the diff sync here mostly.


>Having a sub-workers for this would solve the problem, and the main worker would just delegate tasks.

I tried to do this in the follow-up as I think this would be way to much for one PR.  See https://github.com/oasisprotocol/oasis-core/pull/6242/commits/9a149798e8b87061a2149d592c8893a0e9689240, where this becomes an independent worker.

**How about I cut this PR into two:**
1. All commits from the start all the way to avoiding panic (including).
2. Make storage worker stateless with regards to context/use adapter you suggested (new)
3. Add timeout to checkpoint restoration (new)
4. Avoid potential deadlock on the clean-up.
5. Make `checkpoint.Checkpointer` not require to take `ctx` - https://github.com/oasisprotocol/oasis-core/pull/6299#discussion_r2290350653

The follow-up can then be creating diff sync independent worker which on a high level consists of:
1. Refactor - 213b17fa1f2290f527e527ae25209109d99d494d.
2. Move to separate file
3. Make independent worker - https://github.com/oasisprotocol/oasis-core/pull/6242/commits/9a149798e8b87061a2149d592c8893a0e9689240


Possibly refactor - 213b17fa1f2290f527e527ae25209109d99d494d, could be done as part of the first PR already given that it makes code easier to reason about and is useful in itself.


Maybe you would prefer that?

go/worker/storage/statesync/state_sync.go
peternose: I was thinking something like this, plus maybe some additional structs for `Apply` and `Finalize`. Simple structs, as little fields as possible, easy to test.

Note that this code might have errors.

```go
type Worker struct { 
	nudger *availabilityNudger
	header *headerFetcher
	differ *diffFetcher
	...
}    
```

```go
// availabilityNudger tracks the progress of last and last synced rounds
// and “nudges” role providers to mark themselves available or unavailable
// based on how closely the node is keeping up with consensus.
type availabilityNudger struct {
	roleProvider    registration.RoleProvider
	rpcRoleProvider registration.RoleProvider
	roleAvailable   bool

	lastRound       uint64
	lastSyncedRound uint64
}

// newAvailabilityNudger creates a new availability nudger.
func newAvailabilityNudger(localProvider, rpcProvider registration.RoleProvider) *availabilityNudger {
	return &availabilityNudger{
		roleProvider:    localProvider,
		rpcRoleProvider: rpcProvider,
		lastRound:       math.MaxUint64,
		lastSyncedRound: math.MaxUint64,
	}
}

// setLastRound updates the last round number.
func (m *availabilityNudger) setLastRound(round uint64) {
	m.lastRound = round
}

// setLastSyncedRound updates the most recently synced round number.
func (m *availabilityNudger) setLastSyncedRound(round uint64) {
	m.lastSyncedRound = round
}

// updateAvailability updates the role's availability based on the gap
// between the last round and the last synced round.
func (m *availabilityNudger) updateAvailability() {
	if m.lastRound == math.MaxUint64 || m.lastSyncedRound == math.MaxUint64 {
		return
	}
	if m.lastRound > m.lastSyncedRound {
		return
	}

	switch roundLag := m.lastRound - m.lastSyncedRound; {
	case roundLag < maximumRoundDelayForAvailability:
		m.markAvailable()
	case roundLag > minimumRoundDelayForUnavailability:
		m.markUnavailable()
	}
}

// markAvailable sets the role as available if it is not already.
func (m *availabilityNudger) markAvailable() {
	if m.roleAvailable {
		return
	}
	m.roleAvailable = true

	m.roleProvider.SetAvailable(func(*node.Node) error { return nil })
	if m.rpcRoleProvider != nil {
		m.rpcRoleProvider.SetAvailable(func(*node.Node) error { return nil })
	}
}

// markUnavailable sets the role as unavailable if it is currently available.
func (m *availabilityNudger) markUnavailable() {
	if !m.roleAvailable {
		return
	}
	m.roleAvailable = false

	m.roleProvider.SetUnavailable()
	if m.rpcRoleProvider != nil {
		m.rpcRoleProvider.SetUnavailable()
	}
}
```

```go
// summaryCache is a concurrent-safe cache for block summaries.
type summaryCache struct {
	mu    sync.Mutex
	cache map[uint64]*blockSummary
}

// newSummaryCache creates a new summary cache.
func newSummaryCache() *summaryCache {
	return &summaryCache{
		cache: make(map[uint64]*blockSummary),
	}
}

// set adds the given summary to the cache.
func (s *summaryCache) set(round uint64, summary *blockSummary) {
	s.mu.Lock()
	defer s.mu.Unlock()
	s.cache[round] = summary
}

// get returns a summary from the cache.
func (s *summaryCache) get(round uint64) (*blockSummary, bool) {
	s.mu.Lock()
	defer s.mu.Unlock()
	summary, ok := s.cache[round]
	return summary, ok
}

// delete removes a summary from the cache.
func (s *summaryCache) delete(round uint64) {
	s.mu.Lock()
	defer s.mu.Unlock()
	delete(s.cache, round)
}
```

```go
// headerFetcher is responsible for fetching block headers and populating the summary cache.
type headerFetcher struct {
	history   history.History
	summaries *summaryCache
}

func newHeaderFetcher(history history.History, summaries *summaryCache) *headerFetcher {
	return &headerFetcher{
		history:   history,
		summaries: summaries,
	}
}

// fetch fetches the block header for the given round and populates the summary cache.
func (f *headerFetcher) fetch(ctx context.Context, round uint64, blk *block.Block) error {
	if _, ok := f.summaries.get(round); !ok && round == math.MaxUint64 {
		dummy := blockSummary{
			Namespace: blk.Header.Namespace,
			Round:     round + 1,
			Roots: []api.Root{
				{
					Version: round + 1,
					Type:    api.RootTypeIO,
				},
				{
					Version: round + 1,
					Type:    api.RootTypeState,
				},
			},
		}
		dummy.Roots[0].Empty()
		dummy.Roots[1].Empty()
		f.summaries.set(round, &dummy)
	}
	// Determine if we need to fetch any old block summaries. In case the first
	// round is an undefined round, we need to start with the following round
	// since the undefined round may be unsigned -1 and in this case the loop
	// would not do any iterations.
	startSummaryRound := round
	if startSummaryRound == math.MaxUint64 {
		startSummaryRound++
	}
	for i := startSummaryRound; i < blk.Header.Round; i++ {
		if _, ok := f.summaries.get(i); ok {
			continue
		}
		oldBlock, err := f.history.GetCommittedBlock(ctx, i)
		if err != nil {
			return fmt.Errorf("getting block for round %d (current round: %d): %w", i, blk.Header.Round, err)
		}
		summary := summaryFromBlock(oldBlock)
		f.summaries.set(i, summary)
	}
	if _, ok := f.summaries.get(blk.Header.Round); !ok {
		summary := summaryFromBlock(blk)
		f.summaries.set(blk.Header.Round, summary)
	}
	return nil
}
```

```go
// diffFetcher is responsible for fetching storage diffs.
type diffFetcher struct {
	diffSync          diffsync.Client
	legacyStorageSync synclegacy.Client

	localStorage storageApi.LocalBackend

	syncingRounds map[uint64]*inFlight
	summaries     *summaryCache

	pool *workerpool.Pool

	ch chan *fetchedDiff
}

// fetch fetches the storage diff for the given rounds.
func (f *diffFetcher) fetch(ctx context.Context, start uint64, end uint64) {
	for round := start; round <= end; round++ {
		f.fetchRound(ctx, round)
	}
}

// fetch fetches the storage diff for the given round.
func (f *diffFetcher) fetchRound(ctx context.Context, round uint64) {
	syncing, ok := f.syncingRounds[round]
	if !ok {
		if len(f.syncingRounds) >= maxInFlightRounds {
			return
		}

		syncing = &inFlight{
			startedAt:     time.Now(),
			awaitingRetry: outstandingMaskFull,
		}
		f.syncingRounds[round] = syncing
	}
	if syncing.outstanding.hasAll() {
		return
	}

	prev, _ := f.summaries.get(round - 1)
	curr, _ := f.summaries.get(round)

	prevRoots := make([]api.Root, len(prev.Roots))
	copy(prevRoots, prev.Roots)

	for i := range prevRoots {
		if prevRoots[i].Type == api.RootTypeIO {
			// IO roots aren't chained, so clear it (but leave cache intact).
			prevRoots[i] = api.Root{
				Namespace: curr.Namespace,
				Version:   curr.Round,
				Type:      api.RootTypeIO,
			}
			prevRoots[i].Hash.Empty()
			break
		}
	}

	for i := range prevRoots {
		rootType := prevRoots[i].Type
		if syncing.outstanding.contains(rootType) {
			continue
		}
		if !syncing.awaitingRetry.contains(rootType) {
			continue
		}
		syncing.scheduleDiff(rootType)

		f.pool.Submit(func() {
			f.retryDiff(ctx, curr.Round, prevRoots[i], curr.Roots[i])
		})

	}
}

// fetch fetches the storage diff for the given round and schedules a retry on failure.
func (f *diffFetcher) retryDiff(ctx context.Context, round uint64, prevRoot, thisRoot api.Root) {
	diff, err := f.getDiff(ctx, round, prevRoot, thisRoot)
	if err != nil {
		f.syncingRounds[round].retry(thisRoot.Type)
		return
	}

	select {
	case f.ch <- diff:
	case <-ctx.Done():
	}
}

// getDiff fetches the storage diff for the local storage or remote peers.
func (f *diffFetcher) getDiff(ctx context.Context, round uint64, prevRoot, thisRoot api.Root) (*fetchedDiff, error) {
	result := &fetchedDiff{
		pf:       rpc.NewNopPeerFeedback(),
		round:    round,
		prevRoot: prevRoot,
		thisRoot: thisRoot,
	}

	// Check if the new root doesn't already exist.
	if !f.localStorage.NodeDB().HasRoot(thisRoot) {
		return result, nil
	}

	result.fetched = true

	// Even if HasRoot returns false the root can still exist if it is equal
	// to the previous root and the root was emitted by the consensus committee
	// directly (e.g., during an epoch transition).
	if thisRoot.Hash.Equal(&prevRoot.Hash) {
		result.writeLog = api.WriteLog{}
		return result, nil
	}

	wl, pf, err := f.fetchDiff(ctx, prevRoot, thisRoot)
	if err != nil {
		return nil, err
	}

	result.pf = pf
	result.writeLog = wl

	return result, nil
}

// fetchDiff fetches writelog using diff sync p2p protocol client.
//
// In case of no peers or error, it fallbacks to the legacy storage sync protocol.
func (w *diffFetcher) fetchDiff(ctx context.Context, start, end api.Root) (api.WriteLog, rpc.PeerFeedback, error) {
	rsp1, pf, err := w.diffSync.GetDiff(ctx, &diffsync.GetDiffRequest{
		StartRoot: start,
		EndRoot:   end,
	})
	if err == nil { // if NO error
		return rsp1.WriteLog, pf, nil
	}

	rsp2, pf, err := w.legacyStorageSync.GetDiff(ctx, &synclegacy.GetDiffRequest{
		StartRoot: start,
		EndRoot:   end,
	})
	if err != nil {
		return nil, nil, err
	}
	return rsp2.WriteLog, pf, nil
}
```

go/worker/storage/statesync/diff_sync.go
peternose: > Not sure I agree with this else why we already have `checkpoint_sync.go` and `checkpointer.go` files. I do find the code organised in this way (even if part of the same worker) way easier to navigate and reason.

If you want to split a struct in two files, this means that it has at least two responsibilities and could probably be refactored.

> Ideally, we should also factor out state initialization out of the main worker into separate file (or at least function). Finally `CheckpointSyncRetry` could be moved to `checkpoint_sync.go`. But this would really be out of scope, as I am focusing on the diff sync here mostly.

Probably yes. 

> **How about I cut this PR into two:**

Yes, that would be nice and we could review and merge quicker.

go/worker/storage/statesync/state_sync.go
martintomazic: Nice! Indeed this approach is more readable compared to what I tried in https://github.com/oasisprotocol/oasis-core/commit/9a149798e8b87061a2149d592c8893a0e9689240 (i.e. I tried avoding mutex for any cost - bad).

Anyways, will prepare one PR in-front and then I co-author you for this part :)

go/worker/storage/statesync/state_sync.go
martintomazic: I don't think this method was ever used, as was here to only implement service interface, but this was already nested inside storage worker and thus not used/registered as a service directly?

go/worker/storage/statesync/state_sync.go
martintomazic: We should also create the _checkpointer_ struct, that requires it's last finalized round to be updated, just like the nudger.

But should state sync worker really be responsible for this orchestration and know about role providers, consensus checkpoints, etc.

How about we make it only responsible for the state initialization, checkpoint sync and diff sync. Moreover the diff sync is another worker, consisting of few structs that you provided above.

_Nudger_, _checkpointer_ and _pruner_ should instead be moved out, (possible each into independent package under `/storage/*`, like is the case with `statesync` rn). Instead of pushing the updates to them, they can `watchFinalizedRounds`.

Then inside `storage.Worker` where we [`registerRuntime`](https://github.com/oasisprotocol/oasis-core/blob/d681f4cd48ed1f7beba4d518015c93fd1700ef1f/go/worker/storage/worker.go#L67:L115), we could have per runtime orchestration worker (_nudger, checkpointer, statesync_) that also implements  `hook` interface + fan-out of new blocks coming from the hook subscription.

Over-engineering? Code wise looks even simpler to me, maybe a bit unsure how to organize/name the packages if we go this way...

peternose: > We should also create the _checkpointer_ struct, that requires it's last finalized round to be updated, just like the nudger.

Probably, I just gave you few examples.

> But should state sync worker really be responsible for this orchestration and know about role providers, consensus checkpoints, etc.

Maybe not, unless we rename it.

> How about we make it only responsible for the state initialization, checkpoint sync and diff sync. Moreover the diff sync is another worker, consisting of few structs that you provided above.

Can we do this in 2 steps, where the second step would be what you have written above?

PR #6300, opened 2025-08-19
go/consensus/cometbft/apps/keymanager: Apply policy at epoch boundary

peternose:


keymanager/src/policy/cached.rs
```diff
@@ -55,40 +55,56 @@ impl Policy {
...
+            match old_policy.serial.cmp(&policy.serial) {
                 Ordering::Greater => return Err(KeyManagerError::PolicyRollback.into()),
                 Ordering::Equal => {
-                    if old_policy.checksum != new_policy.checksum {
+                    if old_policy.checksum != policy.checksum {
                         // Policy should be identical.
                         return Err(KeyManagerError::PolicyChanged.into());
                     }
-                    return Ok(new_policy.checksum);
+                    return Ok(policy.checksum);
```
kostko: Should this check if there is a next policy? Otherwise the below will not execute until the next epoch.

kostko:
Can you check that this works as expected when syncing Testnet and Mainnet over some past key manager policy upgrades?

keymanager/src/policy/cached.rs
peternose: Nice catch, I removed this line so now the policy is in this case replaced with identical one.

peternose:
> Can you check that this works as expected when syncing Testnet and Mainnet over some past key manager policy upgrades?

Sure, will test and let you know.

go/upgrade/migrations/consensus_242.go
```diff
@@ -13,6 +13,8 @@ import (
 // This upgrade includes:
 //   - The `MayQuery“ field in the CHURP SGX policy, which defines which enclave identities
 //     are allowed to query runtime key shares.
+//   - An updated key manager policy update transaction that applies a new policy at the epoch
+//     boundary.
 const Consensus242 = "consensus242"
```
kostko, line 18: Why is this still 242?

peternose: Because it is bumped in https://github.com/oasisprotocol/oasis-core/pull/6289.

peternose:
> Can you check that this works as expected when syncing Testnet and Mainnet over some past key manager policy upgrades?

Synced mainnet (polices with serial from 9 to 19) and testnet (policies with serials from 25 to 51 and from 1 to 4).

PR #6301, opened 2025-08-21
go/common/service: Convert background services

peternose:


go/common/service/convert.go
```diff
@@ -0,0 +1,34 @@
+package service
+
+import (
+	"context"
+)
+
+// Convert transforms a background service into a blocking service.
+func Convert(service BackgroundService) Service {
```
kostko: Maybe `service.Convert` is too general?

PR #6303, opened 2025-08-22
go/consensus/cometbft: Refactor code and improve consistency

peternose:
A few minor changes to the code:
- Replaced block height with last height for clarity, since height always refers to a block but it’s not always clear whether it belongs to the last, current, or next block.
- Changed block height to use `int64` and version to use `uint64`, for consistency.
- Revised and removed few TODO comments.
- …


go/consensus/cometbft/apps/roothash/roothash.go
```diff
@@ -85,155 +86,168 @@ func (app *Application) OnCleanup() {
...
-				return fmt.Errorf("unknown runtime governance model on runtime %s: %s", rt.ID, rt.GovernanceModel)
-			}
+	// Prepare new runtime committee based on what the scheduler did.
+	committee, err := scheduler.Committee(ctx, schedulerAPI.KindComputeExecutor, rt.ID)
+	if err != nil {
+		logger.Error("failed to get executor committee from scheduler", "err", err)
+		return err
+	}
+	if committee == nil {
+		logger.Warn("no executor committee")
```
kostko: This could also move into the switch below?

kostko (approved):

PR #6306, opened 2025-08-25
go/worker/storage: Refactor state sync worker pt1

martintomazic:
**What was done:**

1. **Pass context explicitly.**
2. Fix:
    * Panicking.
    * Deadlocking on the clean-up.
    * Triggering genesis checkpoint to early.
3. Minor refactors.

**Follow-up**:

I propose factoring out _checkpointer_ and _availability nudger_ (https://github.com/oasisprotocol/oasis-core/pull/6308) into separate workers. Could be one PR for each.
This will make _diff sync part_ easier to reason about and thus optimize.

### DIFF
--- a/go/consensus/cometbft/abci/state.go
+++ b/go/consensus/cometbft/abci/state.go
@@ -782,10 +782,11 @@ func newApplicationState(ctx context.Context, upgrader upgrade.Backend, cfg *App
 				}, nil
 			},
 		}
-		s.checkpointer, err = checkpoint.NewCheckpointer(s.ctx, ndb, ldb.Checkpointer(), checkpointerCfg)
-		if err != nil {
-			return nil, fmt.Errorf("state: failed to create checkpointer: %w", err)
-		}
+		s.checkpointer = checkpoint.NewCheckpointer(ndb, ldb.Checkpointer(), checkpointerCfg)
+		go func() {
+			err := s.checkpointer.Serve(ctx)
+			s.logger.Error("checkpointer stopped", "err", err)
+		}()
 	}
 
 	go s.metricsWorker()
--- a/go/storage/mkvs/checkpoint/checkpointer.go
+++ b/go/storage/mkvs/checkpoint/checkpointer.go
@@ -72,7 +72,7 @@ type CreationParameters struct {
 	ChunkerThreads uint16
 }
 
-// Checkpointer is a checkpointer.
+// Checkpointer is responsible for creating the storage snapshots (checkpoints).
 type Checkpointer interface {
 	// NotifyNewVersion notifies the checkpointer that a new version has been finalized.
 	NotifyNewVersion(version uint64)
@@ -95,6 +95,9 @@ type Checkpointer interface {
 	// intervals; after unpausing, a checkpoint won't be created immediately, but the checkpointer
 	// will wait for the next regular event.
 	Pause(pause bool)
+
+	// Serve starts running the checkpointer.
+	Serve(ctx context.Context) error
 }
 
 type checkpointer struct {
@@ -285,7 +288,7 @@ func (c *checkpointer) maybeCheckpoint(ctx context.Context, version uint64, para
 	return nil
 }
 
-func (c *checkpointer) worker(ctx context.Context) {
+func (c *checkpointer) Serve(ctx context.Context) error {
 	c.logger.Debug("storage checkpointer started",
 		"check_interval", c.cfg.CheckInterval,
 	)
@@ -310,7 +313,7 @@ func (c *checkpointer) worker(ctx context.Context) {
 
 		select {
 		case <-ctx.Done():
-			return
+			return ctx.Err()
 		case <-time.After(interval):
 		case <-c.flushCh.Out():
 		case paused = <-c.pausedCh:
@@ -323,7 +326,7 @@ func (c *checkpointer) worker(ctx context.Context) {
 		)
 		select {
 		case <-ctx.Done():
-			return
+			return ctx.Err()
 		case v := <-c.notifyCh.Out():
 			version = v.(uint64)
 		case v := <-c.forceCh.Out():
@@ -387,13 +390,8 @@ func (c *checkpointer) worker(ctx context.Context) {
 
 // NewCheckpointer creates a new checkpointer that can be notified of new finalized versions and
 // will automatically generate the configured number of checkpoints.
-func NewCheckpointer(
-	ctx context.Context,
-	ndb db.NodeDB,
-	creator Creator,
-	cfg CheckpointerConfig,
-) (Checkpointer, error) {
-	c := &checkpointer{
+func NewCheckpointer(ndb db.NodeDB, creator Creator, cfg CheckpointerConfig) Checkpointer {
+	return &checkpointer{
 		cfg:        cfg,
 		ndb:        ndb,
 		creator:    creator,
@@ -405,6 +403,4 @@ func NewCheckpointer(
 		cpNotifier: pubsub.NewBroker(false),
 		logger:     logging.GetLogger("storage/mkvs/checkpoint/"+cfg.Name).With("namespace", cfg.Namespace),
 	}
-	go c.worker(ctx)
-	return c, nil
 }
--- a/go/storage/mkvs/checkpoint/checkpointer.go
+++ b/go/storage/mkvs/checkpoint/checkpointer.go
@@ -115,6 +115,23 @@ type checkpointer struct {
 	logger *logging.Logger
 }
 
+// NewCheckpointer creates a new checkpointer that can be notified of new finalized versions and
+// will automatically generate the configured number of checkpoints.
+func NewCheckpointer(ndb db.NodeDB, creator Creator, cfg CheckpointerConfig) Checkpointer {
+	return &checkpointer{
+		cfg:        cfg,
+		ndb:        ndb,
+		creator:    creator,
+		notifyCh:   channels.NewRingChannel(1),
+		forceCh:    channels.NewRingChannel(1),
+		flushCh:    channels.NewRingChannel(1),
+		statusCh:   make(chan struct{}),
+		pausedCh:   make(chan bool),
+		cpNotifier: pubsub.NewBroker(false),
+		logger:     logging.GetLogger("storage/mkvs/checkpoint/"+cfg.Name).With("namespace", cfg.Namespace),
+	}
+}
+
 // Implements Checkpointer.
 func (c *checkpointer) NotifyNewVersion(version uint64) {
 	c.notifyCh.In() <- version
@@ -141,6 +158,106 @@ func (c *checkpointer) Flush() {
 
 func (c *checkpointer) Pause(pause bool) {
 	c.pausedCh <- pause
+}
+
+func (c *checkpointer) Serve(ctx context.Context) error {
+	c.logger.Debug("storage checkpointer started",
+		"check_interval", c.cfg.CheckInterval,
+	)
+	defer func() {
+		c.logger.Debug("storage checkpointer terminating")
+	}()
+
+	paused := false
+
+	for {
+		var interval time.Duration
+		switch c.cfg.CheckInterval {
+		case CheckIntervalDisabled:
+			interval = CheckIntervalDisabled
+		default:
+			interval = random.GetRandomValueFromInterval(
+				checkpointIntervalRandomizationFactor,
+				rand.Float64(),
+				c.cfg.CheckInterval,
+			)
+		}
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(interval):
+		case <-c.flushCh.Out():
+		case paused = <-c.pausedCh:
+			continue
+		}
+
+		var (
+			version uint64
+			force   bool
+		)
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case v := <-c.notifyCh.Out():
+			version = v.(uint64)
+		case v := <-c.forceCh.Out():
+			version = v.(uint64)
+			force = true
+		}
+
+		// Fetch current checkpoint parameters.
+		params := c.cfg.Parameters
+		if params == nil && c.cfg.GetParameters != nil {
+			var err error
+			params, err = c.cfg.GetParameters(ctx)
+			if err != nil {
+				c.logger.Error("failed to get checkpoint parameters",
+					"err", err,
+					"version", version,
+				)
+				continue
+			}
+		}
+		if params == nil {
+			c.logger.Error("no checkpoint parameters")
+			continue
+		}
+
+		// Don't checkpoint if checkpoints are disabled.
+		switch {
+		case force:
+			// Always checkpoint when forced.
+		case paused:
+			continue
+		case params.Interval == 0:
+			continue
+		case c.cfg.CheckInterval == CheckIntervalDisabled:
+			continue
+		default:
+		}
+
+		var err error
+		switch force {
+		case false:
+			err = c.maybeCheckpoint(ctx, version, params)
+		case true:
+			err = c.checkpoint(ctx, version, params)
+		}
+		if err != nil {
+			c.logger.Error("failed to checkpoint",
+				"version", version,
+				"err", err,
+			)
+			continue
+		}
+
+		// Emit status update if someone is listening. This is only used in tests.
+		select {
+		case c.statusCh <- struct{}{}:
+		default:
+		}
+	}
 }
 
 func (c *checkpointer) checkpoint(ctx context.Context, version uint64, params *CreationParameters) (err error) {
@@ -286,121 +403,4 @@ func (c *checkpointer) maybeCheckpoint(ctx context.Context, version uint64, para
 	}
 
 	return nil
-}
-
-func (c *checkpointer) Serve(ctx context.Context) error {
-	c.logger.Debug("storage checkpointer started",
-		"check_interval", c.cfg.CheckInterval,
-	)
-	defer func() {
-		c.logger.Debug("storage checkpointer terminating")
-	}()
-
-	paused := false
-
-	for {
-		var interval time.Duration
-		switch c.cfg.CheckInterval {
-		case CheckIntervalDisabled:
-			interval = CheckIntervalDisabled
-		default:
-			interval = random.GetRandomValueFromInterval(
-				checkpointIntervalRandomizationFactor,
-				rand.Float64(),
-				c.cfg.CheckInterval,
-			)
-		}
-
-		select {
-		case <-ctx.Done():
-			return ctx.Err()
-		case <-time.After(interval):
-		case <-c.flushCh.Out():
-		case paused = <-c.pausedCh:
-			continue
-		}
-
-		var (
-			version uint64
-			force   bool
-		)
-		select {
-		case <-ctx.Done():
-			return ctx.Err()
-		case v := <-c.notifyCh.Out():
-			version = v.(uint64)
-		case v := <-c.forceCh.Out():
-			version = v.(uint64)
-			force = true
-		}
-
-		// Fetch current checkpoint parameters.
-		params := c.cfg.Parameters
-		if params == nil && c.cfg.GetParameters != nil {
-			var err error
-			params, err = c.cfg.GetParameters(ctx)
-			if err != nil {
-				c.logger.Error("failed to get checkpoint parameters",
-					"err", err,
-					"version", version,
-				)
-				continue
-			}
-		}
-		if params == nil {
-			c.logger.Error("no checkpoint parameters")
-			continue
-		}
-
-		// Don't checkpoint if checkpoints are disabled.
-		switch {
-		case force:
-			// Always checkpoint when forced.
-		case paused:
-			continue
-		case params.Interval == 0:
-			continue
-		case c.cfg.CheckInterval == CheckIntervalDisabled:
-			continue
-		default:
-		}
-
-		var err error
-		switch force {
-		case false:
-			err = c.maybeCheckpoint(ctx, version, params)
-		case true:
-			err = c.checkpoint(ctx, version, params)
-		}
-		if err != nil {
-			c.logger.Error("failed to checkpoint",
-				"version", version,
-				"err", err,
-			)
-			continue
-		}
-
-		// Emit status update if someone is listening. This is only used in tests.
-		select {
-		case c.statusCh <- struct{}{}:
-		default:
-		}
-	}
-}
-
-// NewCheckpointer creates a new checkpointer that can be notified of new finalized versions and
-// will automatically generate the configured number of checkpoints.
-func NewCheckpointer(ndb db.NodeDB, creator Creator, cfg CheckpointerConfig) Checkpointer {
-	return &checkpointer{
-		cfg:        cfg,
-		ndb:        ndb,
-		creator:    creator,
-		notifyCh:   channels.NewRingChannel(1),
-		forceCh:    channels.NewRingChannel(1),
-		flushCh:    channels.NewRingChannel(1),
-		statusCh:   make(chan struct{}),
-		pausedCh:   make(chan bool),
-		cpNotifier: pubsub.NewBroker(false),
-		logger:     logging.GetLogger("storage/mkvs/checkpoint/"+cfg.Name).With("namespace", cfg.Namespace),
-	}
 }
--- a/go/storage/mkvs/checkpoint/checkpointer_test.go
+++ b/go/storage/mkvs/checkpoint/checkpointer_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sync"
 	"testing"
 	"time"
 
@@ -25,7 +26,12 @@ const (
 
 func testCheckpointer(t *testing.T, factory dbApi.Factory, earliestVersion, interval uint64, preExistingData bool) {
 	require := require.New(t)
-	ctx := context.Background()
+
+	var wg sync.WaitGroup
+	defer wg.Wait()
+
+	ctx, cancel := context.WithCancel(t.Context())
+	defer cancel()
 
 	// Initialize a database.
 	dir, err := os.MkdirTemp("", "mkvs.checkpointer")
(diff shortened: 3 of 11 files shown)


martintomazic:
For the non-synchronized documentation I should prepare a PR and merge intermediately after?

martintomazic:
Let me fix the linting first and then I open it again. :)

go/storage/mkvs/checkpoint/checkpointer.go
```diff
@@ -140,6 +161,106 @@ func (c *checkpointer) Pause(pause bool) {
 	c.pausedCh <- pause
 }
 
+func (c *checkpointer) Serve(ctx context.Context) error {
```
martintomazic, line 163: This is needed as discussed here https://github.com/oasisprotocol/oasis-core/pull/6299#discussion_r2290350653.

The question is does it really have to implement the Serve interface given that is not expected to return an error...?

**update**: this was already returning context error so my bad

go/worker/storage/statesync/state_sync.go
```diff
@@ -1,4 +1,6 @@
-package committee
+// Package statesync defines the logic responsible for initializing, syncing,
+// and pruning of the runtime state using the relevant p2p protocol clients.
+package statesync
```
martintomazic: https://github.com/oasisprotocol/oasis-core/pull/6308#issuecomment-3229639539 is a good preview of the follow-up PR conclusion.

As pointed there, depending on the style we would prefer, renaming this package might be premature. Opinions?

Finally,  [context](https://github.com/oasisprotocol/oasis-core/pull/6308#issue-3359318661) shows the proposed sequence of PRs.

go/worker/storage/statesync/state_sync.go
```diff
@@ -1076,6 +1070,12 @@ func (w *Worker) worker() { // nolint: gocyclo
 	}
 	close(w.initCh)
 
+	// Notify the checkpointer of the genesis round so it can be checkpointed.
+	if w.checkpointer != nil {
```
peternose: Can this be `nil`?

go/storage/mkvs/checkpoint/checkpointer.go
```diff
@@ -95,6 +95,9 @@ type Checkpointer interface {
 	// intervals; after unpausing, a checkpoint won't be created immediately, but the checkpointer
 	// will wait for the next regular event.
 	Pause(pause bool)
+
+	// Serve starts running the checkpointer.
+	Serve(ctx context.Context) error
```
peternose, line 100: Ideally, this should be part of the struct only, not the interface, as some implementations might not need this method at all. I guess we could fix this latter by removing the interface, as we have only one implementation.

go/storage/mkvs/checkpoint/checkpointer.go
```diff
@@ -405,6 +403,5 @@ func NewCheckpointer(
 		cpNotifier: pubsub.NewBroker(false),
 		logger:     logging.GetLogger("storage/mkvs/checkpoint/"+cfg.Name).With("namespace", cfg.Namespace),
 	}
-	go c.worker(ctx)
-	return c, nil
+	return c
```
peternose:
```suggestion
	return &checkpointer{
```

go/storage/mkvs/checkpoint/checkpointer_test.go
```diff
@@ -89,7 +90,12 @@ func testCheckpointer(t *testing.T, factory dbApi.Factory, earliestVersion, inte
 			return ndb.GetRootsForVersion(version)
 		},
 	})
-	require.NoError(err, "NewCheckpointer")
+	go func() {
+		err := cp.Serve(ctx)
```
peternose, line 99: You should gracefully stop this with wait group.

go/worker/storage/statesync/state_sync.go
```diff
@@ -817,11 +796,31 @@ func (w *Worker) worker() { // nolint: gocyclo
...
 	w.logger.Info("starting")
-
 	w.statusLock.Lock()
 	w.status = api.StatusStarting
 	w.statusLock.Unlock()
 
+	ctx, cancel := context.WithCancel(w.ctx)
+	defer cancel()
+
+	go func() {
```
peternose: It would be prettier if you would wrap the `worker` method, e.g.:
```go
func (w * Worker) Serve(ctx context.Context) {
    // Log starting, change status.
    err := w.serve(ctx)
    // Log stopping, change status.
    return err
}

File: go/worker/storage/statesync/state_sync.go [hunk 8]
```diff
@@ -817,11 +796,31 @@ func (w *Worker) worker() { // nolint: gocyclo
...
+		case <-ctx.Done():
+			w.statusLock.Lock()
+			w.status = api.StatusStopping
+			w.statusLock.Unlock()
+		}
+	}()
+	defer w.logger.Info("stopped")
+
+	go func() {
+		err := w.checkpointer.Serve(w.ctx)
```
peternose: You should wait for this to return before exiting using wait group `wg.Go`.

go/worker/storage/statesync/state_sync.go
```diff
@@ -1365,6 +1365,7 @@ mainLoop:
 		}
 	}
 
+	cancel() // Ctx has to be canceled so that fetcher go routines can be emptied.
```
peternose: I would remove the comment and try to refactor the code is such a way that I have the following structure:
```go
// other code 
defer wg,Wait()
// other code 
defer cancel()
// other code 
```

.changelog/6036.trivial.md
peternose, line 1: 6036 or 6306?

go/worker/storage/worker.go
```diff
@@ -133,6 +142,25 @@ func (w *Worker) Initialized() <-chan struct{} {
...
+			w.logger.Error("failed", "error", err)
+		}
+	}()
+	return nil
+}
+
+// Serve starts a state sync worker for each of the configured runtime, unless
+// disabled.
+//
+// Once all workers have been initialized the init channel is closed.
```
peternose: I believe this and the next comments are redundant. If I were to add one, it would describe what the method does rather than how it is implemented (e.g., how workers are created or stopped).

go/worker/storage/worker.go
```diff
@@ -142,34 +170,35 @@ func (w *Worker) Start() error {
...
-	// Wait for all runtimes to terminate.
-	go func() {
-		defer close(w.quitCh)
-
-		for _, r := range w.runtimes {
-			<-r.Quit()
-		}
+	w.logger.Info("starting", "num_runtimes", len(w.runtimes))
+	defer func() {
+		close(w.quitCh)
```
peternose, line 171: Same here, wrap in `Serve`, do in `serve`.

go/worker/storage/worker.go
```diff
@@ -142,34 +170,35 @@ func (w *Worker) Start() error {
...
 
-	return nil
+	return w.serve(ctx)
+}
+
+func (w *Worker) serve(ctx context.Context) error {
+	g, ctx := errgroup.WithContext(ctx)
+	for id, r := range w.runtimes {
+		g.Go(func() error {
+			err := r.Serve(ctx)
```
peternose: Usually, you would but this in `if`.

go/worker/storage/worker.go
```diff
@@ -179,9 +208,9 @@ func (w *Worker) Stop() {
 		return
 	}
 
-	for _, r := range w.runtimes {
-		r.Stop()
-	}
+	w.logger.Info("stopping")
+	w.cancel()
+	<-w.quitCh
```
peternose, line 208:
```suggestion
	close(w.quitCh)
```
Which can be now moved to the top as `defer close(w.quitCh)`.

go/worker/storage/statesync/state_sync.go
```diff
@@ -450,7 +453,7 @@ func (w *Worker) fetchDiff(round uint64, prevRoot, thisRoot storageApi.Root) {
 		"new_root", thisRoot,
 	)
 
-	ctx, cancel := context.WithCancel(w.ctx)
+	ctx, cancel := context.WithTimeout(w.ctx, diffResponseTimeout)
```
peternose: Does it make sense to have timeout here but not for `case w.diffCh <- result:`?

go/worker/storage/statesync/state_sync.go
```diff
@@ -1370,4 +1312,5 @@ mainLoop:
 	// blockCh will be garbage-collected without being closed. It can potentially still contain
 	// some new blocks, but only as many as were already in-flight at the point when the main
 	// context was canceled.
+	return nil
```
peternose: I would expect here to be `ctx.Err()`, unless function finished its work before the `ctx` was canceled. But according to line `break mainLoop` this is not the case.

Using `defer` for `cancel` and `wg.Wait()`, we could return an error in the loop without need for a `break`.

go/worker/storage/statesync/checkpoint_sync.go
```diff
@@ -222,7 +222,7 @@ func (w *Worker) handleCheckpoint(check *checkpointsync.Checkpoint, maxParallelR
 	chunkReturnCh := make(chan *chunk, maxParallelRequests)
 	errorCh := make(chan int, maxParallelRequests)
 
-	ctx, cancel := context.WithCancel(w.ctx)
+	chunkCtx, cancel := context.WithCancel(ctx)
```
peternose:
```suggestion
	var wg sync.WaitGroup
	defer wg.Wait()
	
	chunkCtx, cancel := context.WithCancel(ctx)
	defer cancel()
```
Actually, I think you can remove this `ctx` as it does nothing.

go/worker/storage/statesync/state_sync.go
```diff
@@ -1023,6 +1023,8 @@ func (w *Worker) Serve(ctx context.Context) error { // nolint: gocyclo
 		w.checkpointer.Flush()
 	}
 
+	// Main syncing loop:
```
peternose: This comment wants to enumerate, but there are no lines/bullets afterwards. It looks like something is missing, It also says this is the main loop, but I don't see any `for` loop.

Also, the name of the commit is not well structured.

go/worker/storage/statesync/state_sync.go
```diff
@@ -1023,6 +1023,8 @@ func (w *Worker) Serve(ctx context.Context) error { // nolint: gocyclo
...
+	err = nil
```
peternose: This line is very unusual and not expected.  I also have no idea why we need it. The code should be organized differently, if this is needed.

go/worker/storage/statesync/state_sync.go
```diff
@@ -1312,5 +1311,5 @@ mainLoop:
 	// blockCh will be garbage-collected without being closed. It can potentially still contain
 	// some new blocks, but only as many as were already in-flight at the point when the main
 	// context was canceled.
-	return nil
+	return err
```
peternose: This is not good as I have no idea where this error came from. Using my suggestion from before, this could be solved by moving return values into the loop.

go/worker/storage/statesync/state_sync.go
```diff
@@ -1025,20 +1022,20 @@ func (w *Worker) Serve(ctx context.Context) error { // nolint: gocyclo
...
 	err = nil
-	// Don't register availability immediately, we want to know first how far behind consensus we are.
-	latestBlockRound := w.undefinedRound
-
-	heartbeat := heartbeat{}
-	heartbeat.reset()
-
 	var wg sync.WaitGroup
+
+	latestBlockRound := w.undefinedRound // Don't register availability immediately, we want to know first how far behind consensus we are.
```
peternose: Why long comment in the same line?

go/worker/storage/statesync/metrics.go
```diff
@@ -7,15 +7,15 @@ import (
...
+	storageWorkerLastFinalizedRound = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "oasis_worker_storage_full_round",
 			Help: "The last round that was fully synced and finalized.",
 		},
 		[]string{"runtime"},
 	)
 
-	storageWorkerLastSyncedRound = prometheus.NewGaugeVec(
+	storageWorkerLastFullyAppliedRound = prometheus.NewGaugeVec(
```
peternose: Fully applied to what? I prefer last finalized and last synced.
You can also change the name of the metric, but then you also need to notify devops or change their dashboards/graphs (maybe too much work).

go/worker/storage/statesync/checkpoint_sync.go
```diff
@@ -21,8 +21,15 @@ import (
 const (
 	// cpListsTimeout is the timeout for fetching checkpoints from all nodes.
 	cpListsTimeout = 30 * time.Second
-	// cpRestoreTimeout is the timeout for restoring a checkpoint chunk from the remote peer.
-	cpRestoreTimeout = 60 * time.Second
+
+	// cpRestoreChunkTimeout is the timeout for restoring a checkpoint chunk from the remote peer.
+	cpRestoreChunkTimeout = 60 * time.Second
```
peternose: `time.Minute`

go/worker/storage/statesync/checkpoint_sync.go
```diff
@@ -195,6 +202,8 @@ func (w *Worker) fetchChunk(ctx context.Context, chunk *chunk) ([]byte, rpc.Peer
 }
 
 func (w *Worker) handleCheckpoint(ctx context.Context, check *checkpointsync.Checkpoint, maxParallelRequests uint) (cpStatus int, rerr error) {
+	ctx, cancel := context.WithTimeout(ctx, cpRestoreTimeout)
+	defer cancel()
```
peternose:
```suggestion
	defer cancel()

```

go/worker/storage/statesync/checkpoint_sync.go
```diff
@@ -21,8 +21,15 @@ import (
...
-	cpRestoreTimeout = 60 * time.Second
+
+	// cpRestoreChunkTimeout is the timeout for restoring a checkpoint chunk from the remote peer.
+	cpRestoreChunkTimeout = 60 * time.Second
+
+	// cpRestoreTimeout is the timeout for restoring the whole checkpoint from the remote peers.
+	//
+	// As of now it takes ~10-30 min to restore the state from the checkpoint, however the timeout
+	// should be significantly higher to account for the growing state.
+	cpRestoreTimeout = 12 * time.Hour
```
peternose: Is this timeout even useful? It’s so large, who would actually wait 12 hours just to see a program return an error?

go/storage/mkvs/checkpoint/checkpointer.go
peternose: If it is not expected to return an error, why does it need `ctx`? If it needs `ctx`, the method should return an error.

go/worker/storage/statesync/checkpoint_sync.go
martintomazic: Probably not, maybe we just remove this change given there was no timeout before?

I am only happy with very loose timeout as the state is growing, and we would have no way to notify us this should be updated? Otherwise it could happen 5 years from now that this would start timeouting and we would be forced to do patch release...

go/worker/storage/statesync/metrics.go
martintomazic: >Fully applied to what

Last round that has all the roots applied to nodedb.

Solution may be to refactor `lastFullyApplied` to `lastSynced` consistently across this worker as you suggest.

There is also `flushSyncedState`, `w.syncedState`, `GetLastSynced` that refers to the last finalized. Maybe this should be also refactored to the `*finalized`, and then existing metrics names are sound.

I guess synced from the client perspective may make slighlty more sense, but we get consistency at least, so I am for this change.

**Update**: If no simple solution I prefer to drop this commit altogether as this PR already does enough.

go/worker/storage/statesync/checkpoint_sync.go
martintomazic:
```go
	// Cancel on exit and wait for the worker pool to drain so that the abort
	// above can proceed safely.
	defer func() {
		cancel()
		<-doneCh
	}()
```

not sure the one at the very top of this function is needed:
```go
ctx, cancel := context.WithTimeout(ctx, cpRestoreTimeout)
defer cancel()
```

go/worker/storage/statesync/state_sync.go
martintomazic: You have a context cancellation there. This is now consistent with `cpRestoreChunkTimeout`.

go/worker/storage/worker.go
martintomazic: ?

go/worker/storage/worker.go
martintomazic: I will remove, albeit this is struct not interface so in this sense documenting impl is valid.

**update**: I find this documentation useful especially given this is implementation, also not sure about suggestion of moving all this into `serve`? Now we have half half of the code logically split.

go/worker/storage/worker.go
martintomazic: No this is blocking until quitCh is finished, whilst you are proposing close.

This has to stay here so that service manager will wait for this worker to terminate.

go/worker/storage/statesync/state_sync.go
martintomazic: **Update**: you meant `WaitGroup` not `errgroup`. all good.

~I guess this would change the semantic slightly as `wg.Go` would trigger shutdown of other goroutines if this fails first. In practice this won't happen right now as `checkpointer.Serve` currently only return `context.Error` but this might change in the future.~

~The idea is checkpointing is less critical part of the code, and error here should probably not close the state sync worker? (at least this was previous semantics).~

~Alternative is to move `WaitGroup` from below up and use it to wait for this go routine also. Not sure maybe I am complicating here, opinions?~

go/worker/storage/statesync/state_sync.go
martintomazic: True, but my understanding status stopping is used when stoping is triggered and not finished, as it could take sometime. Not sure this is best.

**update**: Also this would require me to wrap 2000LC in the serve function so would prefer to keep it as is for now, and do this once things become smaller.

go/storage/mkvs/checkpoint/checkpointer.go
martintomazic: Agree, even if multiple implementation we should follow "return struct, accept interfaces". As you point out let's keep this for now given this is a design problem of existing and not related code (out of scope)?

go/worker/storage/statesync/state_sync.go
martintomazic: Not really, as checkpointer is always created. Should I add a separate commit for that, given this was here redundantly before?

**update**: I suggest fixing this in the follow-up when I will take out the checkpointer. This was existing code anyways.

go/worker/storage/statesync/state_sync.go
martintomazic: >update: Renaming to statesync is definitely premature given that likely we would have diffsync worker. I would however keep refactor that internally renames Node to Worker as this is not node...

If you agree I am removing package rename, but will keep `node.go` -> `worker.go` and `Node` to `Worker` as this is not a node :).

Remaining parts can be discussed on the follow-up PR.

peternose: > If you agree I am removing package rename, but will keep `node.go` -> `worker.go` and `Node` to `Worker` as this is not a node :).

Fine with me.

go/worker/storage/statesync/state_sync.go
peternose: > Should I add a separate commit for that, given this was here redundantly before?
> 
> **update**: I suggest fixing this in the follow-up when I will take out the checkpointer. This was existing code anyways.

Whatever you want. Since the comment is in this PR, it would maybe be better to do it here.

go/worker/storage/statesync/state_sync.go
peternose: It doesn't matter what you use, as long as you wait for the `Serve` to finish.

go/worker/storage/worker.go
peternose: > I will remove, albeit this is struct not interface so in this sense documenting impl is valid.

I’d rather avoid these kinds of comments, since if someone changes the code, they might forget to update the comment. This type of comment just describes what the code already does, which is unnecessary because one can simply read the code itself.

go/worker/storage/worker.go
peternose: Usually, you would put this in `if`.

```go
if err := r.Serve(ctx); err != nil {}
```

go/worker/storage/worker.go
peternose: You are right but the code is still too complicated, as if `enabled==false` you close this in `Stop`, otherwise in `Serve`.

go/worker/storage/statesync/checkpoint_sync.go
peternose: I don't now about the top one, but I think that `chunkCtx, cancel := context.WithCancel(ctx)` does nothing and that `defer` also.

go/worker/storage/statesync/metrics.go
peternose: You can also drop it. I just wanted to say that the name is a bit long and could be improved.

go/worker/storage/worker.go
martintomazic: I find the current organisation more logical and easier to follow.

go/worker/storage/worker.go
martintomazic: I removed explaining about quit and init channel, but kept the the behaviour of the termination.

Generally, If someone changes the code it should also change the comment. Given this is not an interface, describing the actual impl would be also just fine (see `http.Serve`). Indeed one can read the code but the style still requires all the exported methods to be documented.

go/worker/storage/worker.go
martintomazic: This is another issue of existing code, which I don't find a way to fix easily and find  the current approach best.

Ideally, if storage worker is not enabled we should not register it and have it as `nil`. This, however would require to change `startRuntimeWorkers` to also check for nils. Given we are cleaning the code from the inside out, this is incremental step.

I propose to change this once we will be refactoring the background service management. Primarily this PR is about `go/worker/storage/committe/node.go`.

go/worker/storage/statesync/checkpoint_sync.go
martintomazic: No this is needed else in case of an error from the `erroCh` below you would not cancel the go routines so waitgroup would wait indefinitely.

The pattern you proposed originally works, assuming we also drop:

```go
	case <-doneCh:
			// No usable workers left, move on to the next checkpoint.
			return checkpointStatusNext, fmt.Errorf("no usable workers")
		}
```

Which generally should be safe as the only time `checkpointChunkFetcher` returns (worker count decreasing) is upon error or finished, which always implies termination so this case is no longer possible likely is a legacy path from the past when the storage nodes were independent entity?

I am happy to refactor if you confirm my understanding is correct, else we keep it as is.

go/worker/storage/committee/worker.go
```diff
@@ -450,7 +453,7 @@ func (w *Worker) fetchDiff(round uint64, prevRoot, thisRoot storageApi.Root) {
 		"new_root", thisRoot,
 	)
 
-	ctx, cancel := context.WithCancel(w.ctx)
+	ctx, cancel := context.WithTimeout(w.ctx, diffResponseTimeout)
```
martintomazic: Added this one as we discussed in the other threads by simply copying the client internal timeout (defensive).

But this should probably be reset per call or increased 2x given that is now used for calling legacy and new protocol internally...

**update**: tempted to drop this commit alltogether :).

go/consensus/cometbft/abci/state.go
```diff
@@ -782,10 +782,11 @@ func newApplicationState(ctx context.Context, upgrader upgrade.Backend, cfg *App
...
 			},
 		}
-		s.checkpointer, err = checkpoint.NewCheckpointer(s.ctx, ndb, ldb.Checkpointer(), checkpointerCfg)
-		if err != nil {
-			return nil, fmt.Errorf("state: failed to create checkpointer: %w", err)
-		}
+		s.checkpointer = checkpoint.NewCheckpointer(ndb, ldb.Checkpointer(), checkpointerCfg)
+		go func() {
+			err := s.checkpointer.Serve(ctx)
+			s.logger.Error("checkpointer failed", "err", err)
```
peternose:
```suggestion
			s.logger.Error("checkpointer stopped", "err", err)
```

I prefer stopped, as you are not sure if it really failed (assuming context cancellation is not a failure).

go/worker/storage/committee/worker.go
```diff
@@ -755,155 +713,166 @@ func (n *Node) consensusCheckpointSyncer() {
...
+			w.statusLock.Lock()
+			w.status = api.StatusStopping
+			w.statusLock.Unlock()
+		}
+	}()
+	defer w.logger.Info("stopped")
+
+	wg.Go(func() {
+		err := w.checkpointer.Serve(ctx)
+		w.logger.Error("checkpointer failed", "err", err)
```
peternose: Same here.

go/worker/storage/committee/checkpoint_sync.go
```diff
@@ -21,8 +21,8 @@ import (
 const (
 	// cpListsTimeout is the timeout for fetching checkpoints from all nodes.
 	cpListsTimeout = 30 * time.Second
-	// cpRestoreTimeout is the timeout for restoring a checkpoint chunk from a node.
-	cpRestoreTimeout = 60 * time.Second
+	// cpRestoreTimeout is the timeout for restoring a checkpoint chunk from the remote peer.
+	cpRestoreTimeout = 1 * time.Minute
```
peternose:
```suggestion
	cpRestoreTimeout = time.Minute
```

go/worker/storage/committee/worker.go
```diff
@@ -755,155 +713,166 @@ func (n *Node) consensusCheckpointSyncer() {
...
-	n.statusLock.Lock()
-	n.status = api.StatusStarting
-	n.statusLock.Unlock()
+	var wg sync.WaitGroup
+	defer wg.Wait()
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	go func() {
```
peternose: Why not `wg.Go`?

go/worker/storage/worker.go
```diff
@@ -133,6 +142,21 @@ func (w *Worker) Initialized() <-chan struct{} {
 
 // Start starts the storage service.
 func (w *Worker) Start() error {
+	go func() {
+		if err := w.Serve(w.ctx); err != nil {
+			w.logger.Error("failed", "error", err)
```
peternose:
```suggestion
			w.logger.Error("worker stopped", "err", err)
```

go/worker/storage/committee/worker.go
martintomazic: I also removed redundant `select` in this go routine (captured by static check not part of our CI lint).

go/worker/storage/committee/worker.go
martintomazic: I moved this down to `getDiff` now and reset it.

Issue #6307, opened 2025-08-25
Refactor runtime storage committee worker into smaller and independent workers


martintomazic:
~We should also move out things that this worker should not be responsible for such as updating availability, checkpointing and pruning the state.~

Depending on the config storage committee worker should only orchestrate:
1. Diff sync worker (fetches and applies storage diffs).
2. Light sync worker (populates light history - either uses indexer, or eventually p2p or both).
3. Checkpointer (responsible for creating storage checkpoints)
4. "Availability nudger" (old name) responsible for registering availability.
5. Checkpoint sync worker - responsible for runtime state sync:    
    * Currently we are leaking resources, as checkpoints sync p2p clients persists idly (peer tracking etc).

martintomazic:
I believe we should resume work on this as it would enable us to implement **#6356**, #6241 and finally #6400.

## Motivation
First two issues should produce around 2x speed-up alone for the state sync. E.g. to sync consensus with paratime from the genesis it currently takes at least two weeks :/. 

Last issue is about simplifying our abstractions and making future extensions easier to implement.

## Additional requirement
Benchmarking of the state sync should be reproducible, i.e. it should be independent of the p2p layer. This will also enable us to mock and test some of the unhappy paths of this worker.

## Proposed next step 

With #6306  and #6411 merged I suggest to start working on:

> 1. Diff sync worker (fetches and applies storage diffs).

_Diff sync worker_ should define an interface:

```go
type Diff struct {
	round    uint64
	prevRoot storageApi.Root
	thisRoot storageApi.Root
	writeLog storageApi.WriteLog
}

type Fetcher interface {
	Next(ctx context.Context) (Diff, error)
	Accept()
	Reject()
}
```

and we should have various implementations:
```go
// P2PFetcher initializes p2p clients and ensures up to configured number of prefetched storage diffs.
type P2PFetcher struct {
    history history.History
	// other fields
}

// Filefetcher fetches storage diffs from a pre-populated binary file.
// Useful for benchmarking.
type FileFetcher struct {
    path string
	// fields
}
```

[...]

PR #6308, opened 2025-08-27
State sync worker should be only responsible for initializing and syncing the state

martintomazic:
WIP

Follow-up of #6306, inspired by https://github.com/oasisprotocol/oasis-core/pull/6299#discussion_r2296812186, part of the #6307.

After that we will refactor state sync worker into smaller workers (#6307) = https://github.com/oasisprotocol/oasis-core/pull/6299.

Finally, we will have simple PR that optimizes things.

TODO:
- [ ] Move out pruner
- [ ] Move out p2p server registration
- [ ] Add tests
- [ ] Checkpoint sync is independent worker and once it finishes we should close the checkpointsync p2p client, else we are leaking peermgr goroutine...

**End goal:**
Runtime storage worker should orchestrate workers, implement hooks and be responsible for the status:
1. Init state, by filling the possible missing rounds in nodedb and reindexed history.
2. Depending on the result from **1.** start checkpoint sync worker, that also handles lifetime of the underlying checkpoint sync clients(!).
3. State is now fully initialized (update the status)
4. Start remaining workers:
    * Checkpointer, also responsible for genesis checkpoint, even if not enabled or could be separate task.
    * Availability nudger
    * Diff sync worker.
5. Register p2p protocol servers
6. Register prune handler.

This way all workers that actually do work can be easily tested in isolation.

### DIFF
--- a/go/worker/storage/availabilitynudger/availability.go
+++ b/go/worker/storage/availabilitynudger/availability.go
@@ -1,0 +1,144 @@
+// Package availabilitynudger defines logic for updating the role providers.
+package availabilitynudger
+
+import (
+	"context"
+	"fmt"
+	"math"
+
+	"github.com/eapache/channels"
+
+	"github.com/oasisprotocol/oasis-core/go/common"
+	"github.com/oasisprotocol/oasis-core/go/common/logging"
+	"github.com/oasisprotocol/oasis-core/go/common/node"
+	"github.com/oasisprotocol/oasis-core/go/roothash/api/block"
+	"github.com/oasisprotocol/oasis-core/go/worker/registration"
+	"github.com/oasisprotocol/oasis-core/go/worker/storage/statesync"
+)
+
+const (
+	// The maximum number of rounds the worker can be behind the chain before it's sensible for
+	// it to register as available.
+	maximumRoundDelayForAvailability = uint64(10)
+
+	// The minimum number of rounds the worker can be behind the chain before it's sensible for
+	// it to stop advertising availability.
+	minimumRoundDelayForUnavailability = uint64(15)
+)
+
+// Worker tracks the progress of last and last synced rounds
+// and “nudges” role providers to mark themselves available or unavailable
+// based on how closely the node is keeping up with consensus.
+type Worker struct {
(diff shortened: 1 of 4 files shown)


go/worker/storage/checkpointer/checkpointer.go
```diff
@@ -0,0 +1,323 @@
...
+		}
+	}()
+
+	if err := w.ensureGenesisCheckpoint(ctx); err != nil {
+		return fmt.Errorf("failed to ensure genesis checkpoint was created: %w", err)
+	}
+
+	if !w.cfg.CheckpointerEnabled {
+		return nil // We can return safely after creating the genesis checkpoint.
+	}
```
martintomazic, line 149: Will debug this tmr, I think I block here forever. Should be an easy fix.

martintomazic:
On one hand I like granular packages, so that I can reuse simple words `[package_name].New`, and conveniently return `Worker` type. I can use `[package_name].Config` etc. If package has a good name this becomes very useful... Generally, the namespace is very clean.

On the other hand having workers nested under single package, I can make them private, and thus they are not seen in a documentation, which in this context is probably a good thing. **update**: Not true, as documentation should then be moved to the worker orchestrating... Finally, all this packages should be under `internal`,  to make it clear sub-packaging is only for namespacing and API is not stable (e.g. `storage/committee` even `storage`). I think `pkgsite` documentation correctly hides everything under `internal` and by default shows only what the client (someone importing oasis protocol) should use and know about. Still not convinced entirely with neither of the approaches :).

If we decide to go with the latter approach, then `runtime_worker.go`  from this PR should probably be moved inside `committee` package. In such case renaming `committee` to `statesync` was premature, given we will have many workers not directly related to the state sync inside this package. Regardless, we should definitely find a better name, but maybe later. **update**: Renaming to `statesync` is definitely premature given that likely we would have `diffsync` worker. I would however keep refactor that internally renames `Node` to `Worker` as this is not node...

The good point it is relatively easy to refactor between any of the approaches. Finally, this PR reinforced that _State sync worker should be only responsible for initializing and syncing the state_ so factoring out checkpointer and availability nudger after #6306 is the right step to do.

Issue #6309, opened 2025-09-01
Add new command to oasis-node that flattens badger DB


martintomazic:
**Motivation**:
We should [flatten](https://github.com/oasisprotocol/oasis-core/issues/6286#issuecomment-3239940655) `BadgerDB` instances before exporting them for the snapshots, to prevent needless storage, download bandwidth, and replicated pending compaction on all the nodes using it.

Moreover, it would be useful as general purpose maintenance command and should as such be added under appropriate doc section. 

**Solution space**:
Add a new command to oasis node e.g. `oasis-node compact databases` that would flatten all the databases.

This would be a long running operation so cannot be triggered on the node shutdown. As from the [Flatten](https://pkg.go.dev/github.com/dgraph-io/badger/v4@v4.5.1#DB.Flatten) documentation, ideally there should be no writes happening, so this should be called once the node is stopped or syncing is paused.

**Follow-up**:

It would be useful to have a separate command for pruning consensus blocks, and another for consensus and runtime state, without starting the full-node.

Not sure this is even possible, given that ABCI `Commit` with `lastRetainedHeight` defines what can be pruned and what not (prune handlers).

peternose:
It would be nice to test this on devops snapshots/backups.

martintomazic:
>It would be nice to test this on devops snapshots/backups.

Agree.

Will open a PR for that and try it on my machine that has a lot of non-reclaimed disk space. After that we can try it on the devops backups.

There is also this db instance (block sync from genesis, pruning set to 1000), that I am also quite confident would benefit from this command.
```bash
68G	/node/data/consensus/data/state.badger.db
```

This is not to be confused with the mkvs state (that would also benefit from it):
```bash
22G	/node/data/consensus/state/mkvs_storage.pathbadger.db
```

cc @anzoman can you start tracking that on the devops side? I assume currently before exporting snapshots, you don't run any helper script/program to compact the data?

anzoman:
@martintomazic, thanks for initiating this!

On our side, we currently use [zstd](https://github.com/facebook/zstd) compression algorithm before exporting the snapshots, but reducing size even more would come in handy for the restore process.

martintomazic:
@anzoman here are the results.

Downloaded[ consensus snapshot](https://snapshots.oasis.io/#node/mainnet/2025-07-15-eden/).

Decompressing and extracting to my data dir this corresponds to:
```bash
du -h data
[...]

PR #6311, opened 2025-09-03
go/oasis-node/cmd/storage: Add command that flattens consensus dbs

martintomazic:
Part of #6309.

This command has been thoroughly tested and is able to immediately reclaim all the disk space even if enabling pruning late.

Unfortunately, there is a known corner case where disk space used during the peak of the compaction may be double the original size - [see](https://github.com/oasisprotocol/oasis-core/issues/6309#issuecomment-3339126719). It is for this reason that I suggest to _release this as experimental command_.

**Performance:**
Currently it takes less then `1h` to compact all the consensus databases (with all the state/blocks from the genesis) on my `2TB` playground.

 I haven't observed that increasing number of the `Flatten` workers would increase the overall disk usage during the peak of the compaction (as warned by the badger cli documentation). How about adding this as an optional parameter? 

This way users with spare disk space could e.g. configure it to use 8 workers, and thus speed-up compaction significantly. `3-5` times as from my experiments.

We may also explore the impact of increasing `BlockCacheSize`...

**Follow-up**
Given that I was testing this only on the consensus DB instances, I suggest to add compaction of the runtime DB instances as a follow-up. The testing there is also complicated as runtime state pruning is very slow, making it hard to prepare realistic samples.

I suggest the following order of follow-ups:
1. https://github.com/oasisprotocol/oasis-core/pull/6341 (only consensus)
3. Enable compaction for the runtime databases.
3. E2e test for the `compact` command.
4. Enable offline pruning for the runtime databases.

### DIFF
--- a/docs/oasis-node/cli.md
+++ b/docs/oasis-node/cli.md
@@ -331,3 +331,36 @@ response:
 ```
 oasis1qqncl383h8458mr9cytatygctzwsx02n4c5f8ed7
 ```
+
+## storage
+
+### compact-experimental
+
+Run
+
+```sh
+oasis-node storage compact-experimental --config /path/to/config/file
+```
+
+to trigger manual compaction of consensus database instances:
+
+```sh
+{"caller":"storage.go:310","level":"info","module":"cmd/storage", \
+"msg":"Starting database compactions. This may take a while...", \
+"ts":"2025-10-08T09:18:22.185451554Z"}
+```
+
+If pruning was not enabled from the start or was recently increased, then even
+after successful pruning, the disk usage may stay the same.
+
+This is due to the LSM-tree storage design that BadgerDB uses. Concretely,
+deleting a key only marks it as ready to be deleted (a tombstone entry). The
+actual removal of the stale data happens later during the compaction.
+
+During normal operation, compaction happens in the background. However, BadgerDB
+is intentionally lazy, trading write throughput for disk space among other
+things. Therefore it is expected that in case of late pruning, the disk space
+may stay constant or not be reclaimed for a very long time.
+
+This command gives operators manual control to release disk space during
+maintenance periods.


go/oasis-node/cmd/storage/storage.go
```diff
@@ -283,12 +297,83 @@ func doRenameNs(_ *cobra.Command, args []string) error {
 	return nil
 }
 
+func doDbCompactions(_ *cobra.Command, args []string) error {
+	dataDir := cmdCommon.DataDir()
```
martintomazic, line 308: How about this command to require the data dir to be provided explicitly. Although 

 Is it practical to require whole node config file just for this command? Although root flag (`--config`) is always visible so maybe this is not best either.

go/oasis-node/cmd/storage/storage.go
```diff
@@ -283,12 +297,83 @@ func doRenameNs(_ *cobra.Command, args []string) error {
...
+}
+
+func findDBInstances(dataDir string) ([]string, error) {
+	var dbDirs []string
+	err := filepath.WalkDir(dataDir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() && strings.HasSuffix(d.Name(), ".db") {
+			dbDirs = append(dbDirs, path)
```
martintomazic, line 360: This is definitely not robust, although I like the simplicity.

Alternative is to specify the paths explicitly, but this would require making internal paths exported in our packages :/. This still won't work for the cometbft instances where db internal paths are not exported.

1. What if we start using different suffix?
2. Could this be too eager, e.g. doing unexpected changes later on, when we add additional databases? Probably not as from the client perspective compacted and non-compacted database should be the same else you are doing smtg very wrong. 
    * Benefit is we will not forget to update it here.

Suggestions?

go/oasis-node/cmd/storage/storage.go
```diff
@@ -283,12 +297,83 @@ func doRenameNs(_ *cobra.Command, args []string) error {
...
+func compactDB(path string) error {
+	logger = logger.With("path", path)
+
+	logger.Info("compacting")
+	if pretty {
+		fmt.Printf("Compacting %s\n", path)
+	}
+
+	opts := badgerDB.DefaultOptions(path).WithLogger(cmnBadger.NewLogAdapter(logger))
+	db, err := badgerDB.Open(opts)
```
martintomazic: Possibly, this could be backend agnostic command, and we should be calling `optimize`, which may internally do different things if e.g. using B-trees DB. So probably this should accept some kind of interface that both `NodeDB` and key value store as passed to the cometbft would implement.

_I think this is over-engineering and we should handle this later, and only if we actually plan to support multiple backend._

go/oasis-node/cmd/storage/storage.go
```diff
@@ -283,12 +297,86 @@ func doRenameNs(_ *cobra.Command, args []string) error {
...
+
+func compactDB(path string) error {
+	logger = logger.With("path", path)
+
+	logger.Info("compacting")
+	if pretty {
+		fmt.Printf("Compacting %s\n", path)
+	}
+
+	opts := badgerDB.DefaultOptions(path).WithLogger(cmnBadger.NewLogAdapter(logger))
```
kostko: Can passing incompatible options cause any problems? For some databases (eg. state) we use managed mode with explicit timestamp management. Make sure that opening a database in a different manner does not corrupt anything.

go/oasis-node/cmd/storage/storage.go
kostko: I think that passing the config file makes things more robust as then we are sure that the right paths are used if in the future we need to derive further paths.

go/oasis-node/cmd/storage/storage.go
kostko: Let's keep it simple for now.

go/oasis-node/cmd/storage/storage.go
kostko: I think this is fine, we just need to make sure that just opening databases with default options doesn't break anything (see my other comment).

go/oasis-node/cmd/storage/storage.go
kostko: Maybe we should just always open the database in managed mode, it seems like it would be safer.

martintomazic: I tried with non-managed (client node) and the node synced without problems after compaction. I will change this to non-managed as you suggest as is more robust, also safer...

go/oasis-node/cmd/storage/storage.go
```diff
@@ -283,12 +297,86 @@ func doRenameNs(_ *cobra.Command, args []string) error {
...
+	db, err := badgerDB.Open(opts)
+	if err != nil {
+		return fmt.Errorf("failed to open db: %w", err)
+	}
+	defer db.Close()
+
+	if err := db.Flatten(8); err != nil {
+		return fmt.Errorf("failed to flatten db: %w", err)
+	}
```
martintomazic:
```bash
oasis@martintomazic-playground:/node$ ./bin/oasis-node storage compact --config etc/config.yml
Starting database compactions. This may take a while...
Compacting /node/data/consensus/data/blockstore.badger.db
Killed
```

My process was killed when using this script in the 2TB playground instance after pruning the state restored from the [ consensus snapshot](https://snapshots.oasis.io/#node/mainnet/2025-07-15-eden/) (`keep_n=1000`).

~I think~ What happened is is that I ran out of disk space, as compaction likely duplicated the data at peak time. Reducing the number of workers to 1 is sensible, but I still observed some issues. Exploring.

Generally, we should add a recommendation into our docs, that operators should try to preserve at least half of the space empty, if they want to prune in the future or even have reliable compactions.

go/oasis-node/cmd/storage/storage.go
martintomazic: You were right, we should be very careful about the options. E.g. opening the nodedb in the normal mode and calling compaction would result in data loss `300+Gb`  -> `800Mb` :/ Even though I can see all the roots I think what happened it that only latest version (height) was preserved due to the default configs...

I suggest to follow `migrate`, `check` and `rename-ns` pattern, where we added a new method to nodedb interface for each of the commands. Internally they all  [call](https://github.com/oasisprotocol/oasis-core/blob/363ae9e8b5c29b0061ee64a7e88453f43b703b09/go/storage/mkvs/db/badger/helpers.go#L30), thus keeping the options identical.  Meaning we should probably add a new `Compact` method to `NodeDb`.

For the badger instances that cometbft takes, we should grep for all the databases under the corresponding cometbft consensus folder, opening the DB with same settings that [should be made public](https://github.com/oasisprotocol/oasis-core/blob/master/go/consensus/cometbft/db/badger/badger.go#L66:L70).

Are  `migrate`, `check` and `rename-ns` still needed?

go/oasis-node/cmd/storage/storage.go
```diff
@@ -283,12 +295,101 @@ func doRenameNs(_ *cobra.Command, args []string) error {
 	return nil
 }
 
+func doDBCompactions(_ *cobra.Command, args []string) error {
+	if err := cmdCommon.Init(); err != nil {
+		cmdCommon.EarlyLogAndExit(err)
+	}
```
martintomazic, line 306: Commands above don't initialize this, thus all the logs are lost as logger is not initialized. Note this redirects logs to `stdout`, which depending on the configuration may produce a lot of logs on the operator terminal. 

Should this be actually written to `stderr` by default?

update: related to https://github.com/oasisprotocol/oasis-core/pull/6311#discussion_r2397876615

go/oasis-node/cmd/storage/storage.go
```diff
@@ -283,12 +295,101 @@ func doRenameNs(_ *cobra.Command, args []string) error {
...
+
+	return dbDirs, nil
+}
+
+func flattenBadgerDB(db *badgerDB.DB, logger *logging.Logger, path string) error {
+	logger.Info("compacting")
+	if pretty {
+		fmt.Printf("Compacting %s\n", path)
+	}
```
martintomazic: This pattern as can be also find up in this file is redundant, given that logger (at least with current config) already logs to `stdout`...?

go/oasis-node/cmd/storage/storage.go
```diff
@@ -383,6 +389,29 @@ func flattenBadgerDB(db *badgerDB.DB, logger *logging.Logger, path string) error
...
+func compactConsensusNodeDB(dataDir string) error {
+	ldb, ndb, _, err := abci.InitStateStorage(
+		&abci.ApplicationConfig{
+			DataDir:             filepath.Join(dataDir, cmtCommon.StateDir),
+			StorageBackend:      config.GlobalConfig.Storage.Backend,
+			MemoryOnlyStorage:   false,
+			ReadOnlyStorage:     false,
+			DisableCheckpointer: true,
+			// ChainContext:     doc.ChainContext(), TODO: Should we read this from the doc?
+		},
```
martintomazic, line 395: Omitting this for consensus works, but probably I should read a genesis document file provider and get the chaincontext there?

go/oasis-node/cmd/storage/storage.go
martintomazic: This was now fixed by calling exported methods so that exact same options are guaranteed.

As stated we may want to increase the `BlockCacheSize` to (possibly) speed things up.  Given that not whole node is running during this command, every operator has some spare ram.

go/oasis-node/cmd/storage/storage.go
kostko: What is the idea with duplicating the output? I would only use one.

go/oasis-node/cmd/storage/storage.go
kostko: I don't think storage initialization needs the chain context.

go/storage/mkvs/db/badger/badger.go
```diff
@@ -101,6 +101,17 @@ func New(cfg *api.Config) (api.NodeDB, error) {
...
 
+	// Setting a discard timestamp of the BadgerDB is not persistent and is currently
+	// only done during the prune operation.
+	//
+	// Imagine a scenario where during the previous boot of the BadgerDB, data was successfully pruned,
+	// but not yet compacted. Then the NodeDB is restarted, only this time with pruning disabled.
+	// Unless setting discard timestamp to the earliest version manually, the data stored for the
+	// already pruned versions may never be compacted, resulting in redundant disk usage.
+	if earliest := db.GetEarliestVersion(); earliest > tsMetadata {
+		db.db.SetDiscardTs(earliest - 1)
```
kostko: Shouldn't this use `versionToTs` instead?

go/storage/mkvs/db/pathbadger/pathbadger.go
```diff
@@ -58,6 +58,17 @@ func New(cfg *api.Config) (api.NodeDB, error) {
...
 
+	// Setting a discard timestamp of the BadgerDB is not persistent and is currently
+	// only done during the prune operation.
+	//
+	// Imagine a scenario where during the previous boot of the BadgerDB, data was successfully pruned,
+	// but not yet compacted. Then the NodeDB is restarted, only this time with pruning disabled.
+	// Unless setting discard timestamp to the earliest version manually, the data stored for the
+	// already pruned versions may never be compacted, resulting in redundant disk usage.
+	if earliest := db.GetEarliestVersion(); earliest > tsMetadata {
+		db.db.SetDiscardTs(earliest - 1)
```
kostko: Shouldn't this use `versionToTs` instead?

go/oasis-node/cmd/storage/storage.go
martintomazic: I am now only using logging, and this is the output the caller would get if the logger is set to `Info`:

<details>
<summary><b>see logs</b></summary>

```bash
oasis@martintomazic-playground:/node$ oasis-node storage compact-experimental --config etc/config.yml
{"caller":"storage.go:310","level":"info","module":"cmd/storage","msg":"Starting database compactions. This may take a while...","ts":"2025-10-08T09:18:22.185451554Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"All 293 tables opened in 17ms","path":"/node/data/consensus/data/blockstore.badger.db","ts":"2025-10-08T09:18:22.352477401Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Discard stats nextEmptySlot: 0","path":"/node/data/consensus/data/blockstore.badger.db","ts":"2025-10-08T09:18:22.353622596Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Set nextTxnTs to 60475633","path":"/node/data/consensus/data/blockstore.badger.db","ts":"2025-10-08T09:18:22.35390657Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Deleting empty file: /node/data/consensus/data/blockstore.badger.db/000054.vlog","path":"/node/data/consensus/data/blockstore.badger.db","ts":"2025-10-08T09:18:22.354319055Z"}
{"caller":"storage.go:376","level":"info","module":"cmd/storage","msg":"compacting","path":"/node/data/consensus/data/blockstore.badger.db","ts":"2025-10-08T09:18:22.361439547Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"","path":"/node/data/consensus/data/blockstore.badger.db","ts":"2025-10-08T09:18:22.361577945Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 0.      0 B Size.      0 B Max.","path":"/node/data/consensus/data/blockstore.badger.db","ts":"2025-10-08T09:18:22.361698813Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 1.      0 B Size.   10 MiB Max.","path":"/node/data/consensus/data/blockstore.badger.db","ts":"2025-10-08T09:18:22.361744756Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 2.      0 B Size.   10 MiB Max.","path":"/node/data/consensus/data/blockstore.badger.db","ts":"2025-10-08T09:18:22.361780617Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 3.      0 B Size.   10 MiB Max.","path":"/node/data/consensus/data/blockstore.badger.db","ts":"2025-10-08T09:18:22.361815111Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 4.      0 B Size.   38 MiB Max.","path":"/node/data/consensus/data/blockstore.badger.db","ts":"2025-10-08T09:18:22.36184898Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 5.      0 B Size.  383 MiB Max.","path":"/node/data/consensus/data/blockstore.badger.db","ts":"2025-10-08T09:18:22.361881113Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 6.  3.7 GiB Size.  3.7 GiB Max.","path":"/node/data/consensus/data/blockstore.badger.db","ts":"2025-10-08T09:18:22.361918691Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"All tables consolidated into one level. Flattening done.","path":"/node/data/consensus/data/blockstore.badger.db","ts":"2025-10-08T09:18:22.361957829Z"}
{"caller":"storage.go:382","level":"info","module":"cmd/storage","msg":"compaction completed","path":"/node/data/consensus/data/blockstore.badger.db","ts":"2025-10-08T09:18:22.362004339Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"All 2 tables opened in 0s","path":"/node/data/consensus/data/evidence.badger.db","ts":"2025-10-08T09:18:22.379588237Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Discard stats nextEmptySlot: 0","path":"/node/data/consensus/data/evidence.badger.db","ts":"2025-10-08T09:18:22.380499838Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Set nextTxnTs to 6","path":"/node/data/consensus/data/evidence.badger.db","ts":"2025-10-08T09:18:22.380566003Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Deleting empty file: /node/data/consensus/data/evidence.badger.db/000053.vlog","path":"/node/data/consensus/data/evidence.badger.db","ts":"2025-10-08T09:18:22.380746196Z"}
{"caller":"storage.go:376","level":"info","module":"cmd/storage","msg":"compacting","path":"/node/data/consensus/data/evidence.badger.db","ts":"2025-10-08T09:18:22.386150685Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"","path":"/node/data/consensus/data/evidence.badger.db","ts":"2025-10-08T09:18:22.386213759Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 0.  1.4 KiB Size.      0 B Max.","path":"/node/data/consensus/data/evidence.badger.db","ts":"2025-10-08T09:18:22.386297484Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 1.      0 B Size.   10 MiB Max.","path":"/node/data/consensus/data/evidence.badger.db","ts":"2025-10-08T09:18:22.386331717Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 2.      0 B Size.   10 MiB Max.","path":"/node/data/consensus/data/evidence.badger.db","ts":"2025-10-08T09:18:22.386352963Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 3.      0 B Size.   10 MiB Max.","path":"/node/data/consensus/data/evidence.badger.db","ts":"2025-10-08T09:18:22.386372825Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 4.      0 B Size.   10 MiB Max.","path":"/node/data/consensus/data/evidence.badger.db","ts":"2025-10-08T09:18:22.386390678Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 5.      0 B Size.   10 MiB Max.","path":"/node/data/consensus/data/evidence.badger.db","ts":"2025-10-08T09:18:22.386409444Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 6.      0 B Size.   10 MiB Max.","path":"/node/data/consensus/data/evidence.badger.db","ts":"2025-10-08T09:18:22.386428892Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"All tables consolidated into one level. Flattening done.","path":"/node/data/consensus/data/evidence.badger.db","ts":"2025-10-08T09:18:22.386450792Z"}
{"caller":"storage.go:382","level":"info","module":"cmd/storage","msg":"compaction completed","path":"/node/data/consensus/data/evidence.badger.db","ts":"2025-10-08T09:18:22.386482627Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"All 122 tables opened in 7ms","path":"/node/data/consensus/data/state.badger.db","ts":"2025-10-08T09:18:22.469747854Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Discard stats nextEmptySlot: 6","path":"/node/data/consensus/data/state.badger.db","ts":"2025-10-08T09:18:22.470751194Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Set nextTxnTs to 49682633","path":"/node/data/consensus/data/state.badger.db","ts":"2025-10-08T09:18:22.470878904Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Deleting empty file: /node/data/consensus/data/state.badger.db/000053.vlog","path":"/node/data/consensus/data/state.badger.db","ts":"2025-10-08T09:18:22.471123423Z"}
{"caller":"storage.go:376","level":"info","module":"cmd/storage","msg":"compacting","path":"/node/data/consensus/data/state.badger.db","ts":"2025-10-08T09:18:22.476430064Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"","path":"/node/data/consensus/data/state.badger.db","ts":"2025-10-08T09:18:22.47649888Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 0.      0 B Size.      0 B Max.","path":"/node/data/consensus/data/state.badger.db","ts":"2025-10-08T09:18:22.476534877Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 1.      0 B Size.   10 MiB Max.","path":"/node/data/consensus/data/state.badger.db","ts":"2025-10-08T09:18:22.476559671Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 2.      0 B Size.   10 MiB Max.","path":"/node/data/consensus/data/state.badger.db","ts":"2025-10-08T09:18:22.476578393Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 3.      0 B Size.   10 MiB Max.","path":"/node/data/consensus/data/state.badger.db","ts":"2025-10-08T09:18:22.47659683Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 4.      0 B Size.   10 MiB Max.","path":"/node/data/consensus/data/state.badger.db","ts":"2025-10-08T09:18:22.47661545Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 5.      0 B Size.   88 MiB Max.","path":"/node/data/consensus/data/state.badger.db","ts":"2025-10-08T09:18:22.47664698Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"Level: 6.  876 MiB Size.  876 MiB Max.","path":"/node/data/consensus/data/state.badger.db","ts":"2025-10-08T09:18:22.476668319Z"}
{"caller":"helpers.go:41","level":"info","module":"cmd/storage","msg":"All tables consolidated into one level. Flattening done.","path":"/node/data/consensus/data/state.badger.db","ts":"2025-10-08T09:18:22.476690328Z"}
{"caller":"storage.go:382","level":"info","module":"cmd/storage","msg":"compaction completed","path":"/node/data/consensus/data/state.badger.db","ts":"2025-10-08T09:18:22.476715725Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"All 537 tables opened in 34ms","ts":"2025-10-08T09:18:23.030183539Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Discard stats nextEmptySlot: 0","ts":"2025-10-08T09:18:23.032503185Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Set nextTxnTs to 26752296","ts":"2025-10-08T09:18:23.032983936Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Deleting empty file: /node/data/consensus/state/mkvs_storage.pathbadger.db/000054.vlog","ts":"2025-10-08T09:18:23.033752967Z"}
{"caller":"pathbadger.go:741","level":"info","module":"mkvs/db/pathbadger","msg":"compacting","ts":"2025-10-08T09:18:23.039423388Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"","ts":"2025-10-08T09:18:23.039511911Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 0.      0 B Size.      0 B Max.","ts":"2025-10-08T09:18:23.039575414Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 1.      0 B Size.   10 MiB Max.","ts":"2025-10-08T09:18:23.039606073Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 2.      0 B Size.   10 MiB Max.","ts":"2025-10-08T09:18:23.039626897Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 3.      0 B Size.   10 MiB Max.","ts":"2025-10-08T09:18:23.039652958Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 4.      0 B Size.   71 MiB Max.","ts":"2025-10-08T09:18:23.039671905Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 5.   11 MiB Size.  712 MiB Max.","ts":"2025-10-08T09:18:23.039691392Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 6.  7.0 GiB Size.  7.0 GiB Max.","ts":"2025-10-08T09:18:23.039710501Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Attempting to compact with {level:5 score:1.71 adjusted:0 dropPrefixes:[] t:{baseLevel:0 targetSz:[] fileSz:[]}}","ts":"2025-10-08T09:18:23.039740446Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"1 compactor(s) succeeded. One or more tables from level 5 compacted.","ts":"2025-10-08T09:18:23.145678494Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"","ts":"2025-10-08T09:18:23.145741218Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 0.      0 B Size.      0 B Max.","ts":"2025-10-08T09:18:23.145789741Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 1.      0 B Size.   10 MiB Max.","ts":"2025-10-08T09:18:23.145815022Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 2.      0 B Size.   10 MiB Max.","ts":"2025-10-08T09:18:23.145842633Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 3.      0 B Size.   10 MiB Max.","ts":"2025-10-08T09:18:23.145870657Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 4.      0 B Size.   71 MiB Max.","ts":"2025-10-08T09:18:23.145893081Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 5.  3.5 MiB Size.  712 MiB Max.","ts":"2025-10-08T09:18:23.145914811Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 6.  7.0 GiB Size.  7.0 GiB Max.","ts":"2025-10-08T09:18:23.145936938Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Attempting to compact with {level:5 score:1.71 adjusted:0 dropPrefixes:[] t:{baseLevel:0 targetSz:[] fileSz:[]}}","ts":"2025-10-08T09:18:23.145967509Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"1 compactor(s) succeeded. One or more tables from level 5 compacted.","ts":"2025-10-08T09:18:23.207991222Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"","ts":"2025-10-08T09:18:23.208057806Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 0.      0 B Size.      0 B Max.","ts":"2025-10-08T09:18:23.208082957Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 1.      0 B Size.   10 MiB Max.","ts":"2025-10-08T09:18:23.208106752Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 2.      0 B Size.   10 MiB Max.","ts":"2025-10-08T09:18:23.208125343Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 3.      0 B Size.   10 MiB Max.","ts":"2025-10-08T09:18:23.208143639Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 4.      0 B Size.   71 MiB Max.","ts":"2025-10-08T09:18:23.208162819Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 5.      0 B Size.  712 MiB Max.","ts":"2025-10-08T09:18:23.208182284Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level: 6.  7.0 GiB Size.  7.0 GiB Max.","ts":"2025-10-08T09:18:23.208205483Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"All tables consolidated into one level. Flattening done.","ts":"2025-10-08T09:18:23.208227909Z"}
{"caller":"pathbadger.go:747","level":"info","module":"mkvs/db/pathbadger","msg":"compaction completed","ts":"2025-10-08T09:18:23.208253797Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Lifetime L0 stalled for: 0s","ts":"2025-10-08T09:18:23.20829223Z"}
{"caller":"helpers.go:41","level":"info","module":"mkvs/db/pathbadger","msg":"Level 0 [ ]: NumTables: 00. Size: 0 B of 0 B. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 64 MiB\nLevel 1 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB\nLevel 2 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB\nLevel 3 [ ]: NumTables: 00. Size: 0 B of 10 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 2.0 MiB\nLevel 4 [ ]: NumTables: 00. Size: 0 B of 71 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 4.0 MiB\nLevel 5 [B]: NumTables: 00. Size: 0 B of 714 MiB. Score: 0.00->0.00 StaleData: 0 B Target FileSize: 8.0 MiB\nLevel 6 [ ]: NumTables: 537. Size: 7.0 GiB of 7.0 GiB. Score: 0.00->0.00 StaleData: 1.3 GiB Target FileSize: 16 MiB\nLevel Done","ts":"2025-10-08T09:18:23.213098532Z"}
oasis@martintomazic-playground:/node$
```

</details>

I guess the idea of the author above in the previous commands was to split between user facing information and the actual logs, but was not done thoroughly as pointed out https://github.com/oasisprotocol/oasis-core/pull/6311#discussion_r2397872650.

I would prefer to keep it simple and consistent with other commands, e.g. starting `oasis-node` command will split all the logs in the `stdout`, although from the ux perspective this is not best.

martintomazic:
Added a docs section as [requested](https://github.com/oasisprotocol/docs/issues/1469#issuecomment-3377180909) yesterday.

go/oasis-node/cmd/storage/storage.go
```diff
@@ -283,12 +300,119 @@ func doRenameNs(_ *cobra.Command, args []string) error {
...
+			ReadOnlyStorage:     false,
+			DisableCheckpointer: true,
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("failed to initialize ABCI storage backend: %w", err)
+	}
+
+	// Close the resources. Both Close and Cleanup only close NodeDB.
+	// Closing both here, to prevent resource leaks in things change in the future.
```
kostko:
```suggestion
	// Closing both here, to prevent resource leaks if things change in the future.
```

Issue #6327, opened 2025-09-18
Runtime Host Protocol: Propagate write failures


ptrus:
If writing the RHP message fails (e.g. due to message being too large), the requester is never notified and the query appears stuck:
https://github.com/oasisprotocol/oasis-core/blob/22e8b6f2aff6dc525cb13e2542fbb70470b4736b/runtime/src/protocol.rs#L250-L260

Update the function so that when a message write fails, a follow-up error response is sent with the same message ID, indicating the failure to the caller. This should be done for both requests and responses.

PR #6330, opened 2025-09-19
runtime/protocol: Propagate write failures

ptrus:
Fixes https://github.com/oasisprotocol/oasis-core/issues/6327

Example client side response with the update:
```go
	addrss, err := rt.Accounts.Addresses(context.Background(), uint64(height), types.NativeDenomination)
	if err != nil {
		println("Failed to fetch addresses:", err.Error())
	}
	// prints: Failed to fetch addresses: Failed to write response: message too large
```

while before the query would just appear to be stuck on the client.

### DIFF
--- a/runtime/src/protocol.rs
+++ b/runtime/src/protocol.rs
@@ -250,9 +250,15 @@ impl Protocol {
     fn io_write(self: &Arc<Protocol>) {
         info!(self.logger, "Starting protocol writer thread");
 
-        while let Ok(message) = self.outgoing_rx.recv() {
+        while let Ok(
+            message @ Message {
+                id, message_type, ..
+            },
+        ) = self.outgoing_rx.recv()
+        {
             if let Err(error) = self.write_message(message) {
                 warn!(self.logger, "Failed to write message"; "err" => %error);
+                self.handle_write_failure(id, message_type, error);
             }
         }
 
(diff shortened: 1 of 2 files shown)


runtime/src/protocol.rs
```diff
@@ -250,9 +250,11 @@ impl Protocol {
     fn io_write(self: &Arc<Protocol>) {
         info!(self.logger, "Starting protocol writer thread");
 
-        while let Ok(message) = self.outgoing_rx.recv() {
+        while let Ok(message @ Message { id, message_type, .. }) = self.outgoing_rx.recv() {
```
peternose:
```suggestion
        while let Ok(
            message @ Message {
                id, message_type, ..
            },
        ) = self.outgoing_rx.recv()
```
Should this be formatted like this?

runtime/src/protocol.rs
```diff
@@ -250,9 +250,11 @@ impl Protocol {
...
             if let Err(error) = self.write_message(message) {
                 warn!(self.logger, "Failed to write message"; "err" => %error);
+                // Handle write failure by sending error response.
```
peternose:
```suggestion
```

runtime/src/protocol.rs
```diff
@@ -345,6 +347,54 @@ impl Protocol {
         Ok(())
     }
 
+    fn handle_write_failure(
+        &self,
+        message_id: u64,
+        message_type: MessageType,
+        error: &anyhow::Error,
```
peternose:
```suggestion
        error: anyhow::Error,
```

runtime/src/protocol.rs
```diff
@@ -345,6 +347,54 @@ impl Protocol {
...
+                    body: Body::Error(Error::new(
+                        "rhp/write",
+                        1,
+                        &format!("Failed to write response: {error}"),
+                    )),
+                };
+
+                // Try to send the error response; if this also fails, just log it.
+                if let Err(retry_error) = self.write_message(error_response) {
+                    warn!(self.logger, "Failed to write message";
```
peternose:
```suggestion
                    warn!(self.logger, "Failed to write error message"; "err" => %error);
```
No need to log the original error twice.

runtime/src/protocol.rs
```diff
@@ -345,6 +347,54 @@ impl Protocol {
...
+                    )),
+                };
+
+                // Try to send the error response; if this also fails, just log it.
+                if let Err(retry_error) = self.write_message(error_response) {
+                    warn!(self.logger, "Failed to write message"; 
+                          "original_err" => %error, "retry_err" => %retry_error);
+                }
+            }
+            _ => warn!(self.logger, "Write failure for invalid message type"; "err" => %error),
```
peternose: Could also log `message_type`.

runtime/src/protocol.rs
```diff
@@ -345,6 +347,54 @@ impl Protocol {
...
+                    id: message_id,
+                    message_type: MessageType::Response,
+                    body: Body::Error(Error::new(
+                        "rhp/write",
+                        1,
+                        &format!("Failed to write response: {error}"),
+                    )),
+                };
+
+                // Try to send the error response; if this also fails, just log it.
```
peternose:
```suggestion
```

runtime/src/protocol.rs
```diff
@@ -345,6 +347,54 @@ impl Protocol {
...
+
+                if let Some(response_sender) = response_sender {
+                    let error_body = Body::Error(Error::new(
+                        "rhp/write",
+                        1,
+                        &format!("Failed to write request: {error}"),
+                    ));
+
+                    if response_sender.send(error_body).is_err() {
+                        warn!(self.logger, "Failed to deliver response to local handler");
```
peternose:
```suggestion
                        warn!(self.logger, "Failed to deliver error response to local handler");
```

PR #6331, opened 2025-09-21
go/common/sgx/pcs/policy: Add FMSP whitelist to quote policy

peternose:


go/consensus/cometbft/apps/registry/messages.go
```diff
@@ -52,3 +57,30 @@ func (app *Application) changeParameters(ctx *api.Context, msg any, apply bool)
...
+	if teeFeatures == nil {
+		return nil
+	}
+
+	defaultPolicy := teeFeatures.SGX.DefaultPolicy
+	if defaultPolicy == nil {
+		return nil
+	}
+
+	// Allow non-empty `fmpcs_whitelist` when this feature is available.
```
kostko:
```suggestion
	// Allow non-empty `fmspc_whitelist` when this feature is available.
```

go/common/sgx/quote/quote.go
```diff
@@ -64,3 +64,16 @@ type Policy struct {
 	IAS *ias.QuotePolicy `json:"ias,omitempty" yaml:"ias,omitempty"`
 	PCS *pcs.QuotePolicy `json:"pcs,omitempty" yaml:"pcs,omitempty"`
 }
+
+// VerifyFMSPCWhitelist verifies that the FMSPC whitelist is empty.
+func (p *Policy) VerifyFMSPCWhitelist() error {
```
kostko: Seems like a strange name/way to do validation? Should this be something like a general `Validate`?

go/registry/api/runtime.go
```diff
@@ -600,6 +600,13 @@ func (r *Runtime) ValidateDeployments(now beacon.EpochTime, params *ConsensusPar
 			if len(cs.Enclaves) == 0 {
 				return fmt.Errorf("%w: invalid SGX TEE constraints", ErrNoEnclaveForRuntime)
 			}
+			if !isFeatureVersion242 {
+				if cs.Policy != nil {
+					if err := cs.Policy.VerifyFMSPCWhitelist(); err != nil {
```
kostko: Should this be part of `cs.ValidateBasic` which would take this extra flag? Note that before we kept these flags in `params.TEEFeatures` as there can be more in the future which would grow this "flag list". Or we could just pass `ctx` to it.

go/common/sgx/quote/quote.go
peternose: I didn't want to add a general validation method because this should be removed in the future. We only need this method to forbid adding a non-empty whitelist to the consensus before we do feature version bump.

Renamed to `Validate`.

go/registry/api/runtime.go
peternose: Changed. Now all validations take `isFeatureVersion242` parameter which can be removed once we upgrade.

go/common/node/node.go
```diff
@@ -590,7 +590,7 @@ func (c *CapabilityTEE) Verify(teeCfg *TEEFeatures, ts time.Time, height uint64,
 		if err := cbor.Unmarshal(constraints, &sc); err != nil {
 			return fmt.Errorf("node: malformed SGX constraints: %w", err)
 		}
-		if err := sc.ValidateBasic(teeCfg); err != nil {
+		if err := sc.ValidateBasic(teeCfg, isFeatureVersion242); err != nil {
```
kostko, line 594: How will this look when other features are added? Will there be additional version flags like these? Should we instead pass context or somehow make this part of `teeCfg`?

peternose: > How will this look when other features are added? Will there be additional version flags like these? Should we instead pass context or somehow make this part of `teeCfg`?


- If other features are added in the same version, than this remains the same.
- If another feature is added in e.g. 24.3, than we would need to add a new flag `isFeatureVersion243`, pass context or something else. But the trick is that we don't need to do this as this code should be removed in version 24.3, after we do the upgrade, because state sync will work without these changes, as they are here just to prevent filling the whitelist.

kostko: But I guess this can only be removed if there are no transactions posted that would violate this constraint? As otherwise iterative block sync would fail.

kostko (approved):

go/common/node/node.go
peternose: > But I guess this can only be removed if there are no transactions posted that would violate this constraint? As otherwise iterative block sync would fail.

No, wrong thinking. This is here to forbid posting transactions that would violate this constraint (non-empty whitelist) until we do the feature version bump.

This validation ensures we don't have consensus problems, as once we merge this PR in oasis core version 25.6, a new node running version 25.6 could post a policy with non-empty whitelist while an old node with version 25.5 would reject it. Having this check ensures that new nodes must submit empty whitelists until we do the feature version bump. And once we do the bump, this validation can be removed as it is not needed anymore as all nodes should be running at least 25.6. version.

kostko: But what happens if someone posted a transaction with a non-empty whitelist _right now_ and the transaction is included in a block, but fails (because currently it is an unknown field)? Then once this check is removed in a future version, syncing from genesis will cause divergence when processing this transaction (which will now succeed during iterative block sync)? So this works as long as no such transaction is included in a block before the governance vote.

peternose: Oh, now I see. Yes, that would be a problem.

I suggest we merge these changes, and after we upgrade, we do state sync to check that there are no such transactions (this means that there could still be a tx with a whitelist but also with another unknown field, and that field would still fail the tx). If there are, we prettify the code, if not, we can remove these changes.

kostko: Sounds good to me.

martintomazic: >I suggest we merge these changes, and after we upgrade, we do state sync to check that there are no such transactions.

Looks safe to me and indeed simpler. 

If we find such transactions then probably `SGXConstraintsV2` behind feature flag might be cleaner on the long run (unlikely to happen so lets ignore this).

martintomazic (approved):
Is the plan to release this as part of the `25.9` and bind all 4 breaking changes under feature flag  (assuming #6410 will be ready in time as well)?

peternose:
> Is the plan to release this as part of the `25.9` and bind all 4 breaking changes under feature flag (assuming #6410 will be ready in time as well)?

Yes, 25.9 or latter. There are still some breaking changes we would like to merge, before we enable all of them.

peternose:
Mainnet genesis synced successfully, so merging. Will test latter also testnet.

peternose:
> Mainnet genesis synced successfully, so merging. Will test latter also testnet.

Testnet also synced successfully.

Issue #6334, opened 2025-09-24
Pruning of runtime state is too slow


martintomazic:
**Update:** **_`NodeDB.Prune` becomes impractically slow as the  number of versions grow_**.

**Initial problem description**:
Using default settings, pruning of the runtime state happens [every `2` minutes](https://github.com/oasisprotocol/oasis-core/blob/a17cbda4ee5f4ee2c74c9798fb2a49c797e9932d/go/runtime/history/history.go#L322:L341), in a batch of [`64` rounds](https://github.com/oasisprotocol/oasis-core/blob/a17cbda4ee5f4ee2c74c9798fb2a49c797e9932d/go/runtime/history/prune.go#L110). The rate of pruning is therefore by default bounded by `2 rounds/s`.

This is way too slow during initial sync (e.g. if syncing from the genesis), resulting in the similar issues as we observed with CometBFT managed databases only starting with the pruning once synced. I.e. during initial sync of runtime state 1. the disk space would grow way too much, and 2. later BadgerDB is unable to reclaim it.

I tried increasing `runtime.prune.interval` to `1s` from the default 2 minutes, the only thing I achieved was that the rate of pruning is now `1 round/s` :/. Exploring...

martintomazic:
cc @gw0 for visibility and thanks for pointing this out.

martintomazic:
>I tried increasing runtime.prune.interval to 1s from the default 2 minutes, the only thing I achieved was that the rate of pruning is now 1 round/s :/. Exploring...

**Experiment 1:**
Disabled pruning and successfully synced first `1M` rounds. Then enabled pruning again and observed that pruning became significantly slower (few versions per minute at best). Furthermore, the syncing was effectively blocked. By disabling pruning again, the node starts syncing with a rate of `8 versions/s`.

**Sanity 1:**
I wrote a simple script that manually prunes versions offline, to determine where is the bottleneck:
* For consensus with around `550GB` of the state, it took around `50ms` to prune `1000` earliest versions.
    * This corresponds to `1M versions/min` rate of pruning that was already observed during previous experiments.
* For runtime state with only `53GB`  (Sapphire Mainnet, synced up to `2167250`), the rate of the pruning is around 1 version every 30s!


<details>
<summary>See code used</summary>

```go
func openStateDB(dataDir string) (api.NodeDB, error) {
	return pathbadger.New(&db.Config{
		DB:               dataDir,
		NoFsync:          true,
[...]

PR #6339, opened 2025-09-30
go: Bump default block cache size of NodeDB badger instances to 256 MiB

martintomazic:
As from the experiments, increasing block cache size significantly increases the pruning speed of the runtime state ([see](https://github.com/oasisprotocol/oasis-core/issues/6334#issuecomment-3347762369)).

Looking at the badgerDB [documentation](https://github.com/hypermodeinc/badger/blob/64b2f3736e24c16219b0e74f826d46c7187d6f46/options.go#L685), it is recommended to use cache if using compression. The current default value is `256 MB`, thus bumping to it from existing `64 MB`.

Finally, given that we are highly data intensive application we may set it to even more, e.g. `1 GB`.

Might be also a good idea to bump cometbft managed databases (i.e. blockstore, state, evidence) albeit feels wasteful for the evidence, and currently there is no granularity. Given we don't have performance bottlenecks there for now we could keep it as is.

Will add a changelog once we align on the final size and scope of the changes.

### DIFF
--- a/go/consensus/cometbft/abci/state.go
+++ b/go/consensus/cometbft/abci/state.go
@@ -655,7 +655,7 @@ func InitStateStorage(cfg *ApplicationConfig) (storage.LocalBackend, storage.Nod
 	db, err := storageDB.New(&storage.Config{
 		Backend:          cfg.StorageBackend,
 		DB:               filepath.Join(baseDir, storageDB.DefaultFileName(cfg.StorageBackend)),
-		MaxCacheSize:     64 * 1024 * 1024, // TODO: Make this configurable.
+		MaxCacheSize:     256 * 1024 * 1024, // TODO: Make this configurable.
 		DiscardWriteLogs: true,
 		NoFsync:          true, // This is safe as CometBFT will replay on crash.
 		MemoryOnly:       cfg.MemoryOnlyStorage,
--- a/go/storage/mkvs/db/badger/helpers.go
+++ b/go/storage/mkvs/db/badger/helpers.go
@@ -33,7 +33,7 @@ func commonConfigToBadgerOptions(cfg *api.Config, db *badgerNodeDB) badger.Optio
 	opts = opts.WithSyncWrites(!cfg.NoFsync)
 	opts = opts.WithCompression(options.Snappy)
 	if cfg.MaxCacheSize == 0 {
-		opts = opts.WithBlockCacheSize(64 * 1024 * 1024)
+		opts = opts.WithBlockCacheSize(256 * 1024 * 1024)
 	} else {
 		opts = opts.WithBlockCacheSize(cfg.MaxCacheSize)
 	}
--- a/go/worker/storage/config/config.go
+++ b/go/worker/storage/config/config.go
@@ -48,7 +48,7 @@ func (c *Config) Validate() error {
 func DefaultConfig() Config {
 	return Config{
 		Backend:                "auto",
-		MaxCacheSize:           "64mb",
+		MaxCacheSize:           "256mb",
 		FetcherCount:           4,
 		PublicRPCEnabled:       false,
 		CheckpointSyncDisabled: false,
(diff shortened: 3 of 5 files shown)


go/consensus/cometbft/abci/state.go
```diff
@@ -655,7 +655,7 @@ func InitStateStorage(cfg *ApplicationConfig) (storage.LocalBackend, storage.Nod
 	db, err := storageDB.New(&storage.Config{
 		Backend:          cfg.StorageBackend,
 		DB:               filepath.Join(baseDir, storageDB.DefaultFileName(cfg.StorageBackend)),
-		MaxCacheSize:     64 * 1024 * 1024, // TODO: Make this configurable.
+		MaxCacheSize:     256 * 1024 * 1024, // TODO: Make this configurable.
```
martintomazic, line 658: Should we create an issue for TODO and add this config option for consensus as well?

go/worker/storage/config/config.go
```diff
@@ -48,7 +48,7 @@ func (c *Config) Validate() error {
 func DefaultConfig() Config {
 	return Config{
 		Backend:                "auto",
-		MaxCacheSize:           "64mb",
+		MaxCacheSize:           "256mb",
```
martintomazic, line 51: This changes hardware requirements. But not for much given that this is only ~`200MB` more for every affected database? Given that we have consensus and runtime state this is `0.5GB` more per paratime node with the current bump?

As stated in the context I am open to increasing it more also.

go/consensus/cometbft/abci/state.go
kostko: Should this just use the same config option or is there a good reason to have two?

kostko: I would change this to use the same option for now.

go/worker/storage/config/config.go
kostko: Let's start with this bump.

kostko (approved):

go/storage/mkvs/db/badger/helpers.go
```diff
@@ -33,7 +33,7 @@ func commonConfigToBadgerOptions(cfg *api.Config, db *badgerNodeDB) badger.Optio
 	opts = opts.WithSyncWrites(!cfg.NoFsync)
 	opts = opts.WithCompression(options.Snappy)
 	if cfg.MaxCacheSize == 0 {
-		opts = opts.WithBlockCacheSize(64 * 1024 * 1024)
+		opts = opts.WithBlockCacheSize(256 * 1024 * 1024)
```
peternose, line 36: This line could also be omitted as this is the default value, unless we want to be explicit.

go/consensus/cometbft/abci/state.go
martintomazic: >I would change this to use the same option for now.

I would avoid that as `storage` paragraph in the config currently encapsulates only runtime storage worker configuration. Encouraging users to set this field when only running consensus node might be confusing? 

Existing solutions is also strange as `storage` should probably be under `runtime` and we should have `consensus`, `runtime`, and `common` (common config for both), where `keymanager` is just a special case of the `runtime`?

My suggestion is to open a follow-up and do it there under `consesus` section of the config. Possibly we could consider moving this configuration to the `common`, if you want to configure it only once. Obviously this would be breaking.

What do you think?

go/storage/mkvs/db/badger/helpers.go
martintomazic: I would keep it as badger often changes the defaults between versions (e.g. I think in older versions is was `64` actually). This way upgrading does not silently increase hardware requirements.

go/consensus/cometbft/abci/state.go
martintomazic: Tracking it https://github.com/oasisprotocol/oasis-core/issues/6344 and closing.

Issue #6321, opened 2025-09-09
go/oasis-node: Previously synced node starts lagging after enabling aggressive pruning


martintomazic:
When doing experiments with aggressive pruning (state from the genesis -> keep only 1000 latest blocks), I managed to overwhelm already synced node, making `latest_height` be `2h` behind the actual latest block as observed on the block explorer.

To my surprise, the status was still marked as ready:
```bash
 "consensus": {
    "status": "ready",
    ....
}
```

Worst, we have parts of the runtime code that never assume a consensus node may fall behind once marked as ready. E.g.  https://github.com/oasisprotocol/oasis-core/blob/49023d0c895ef6f1eead7292dbe44d5e19a23dea/go/worker/storage/committee/worker.go#L1197:L1199.

martintomazic:
**Explanation**:
Consensus state pruning starts with few minutes lag, but once it does the pruning is extremely fast (e.g. 6 million blocks in 8 minutes when no runtime configured, else prune handlers slow things down...). Then there is also cometbft blockstore pruning that lags behind a bit with pruning. Both processes effectively **starve the progress of syncing**... So if you had synced node with lots of state, stop it and configure pruning to keep only few blocks, and finally re-start immediately, this is quite possible.


As can be seen [state pruning](https://github.com/oasisprotocol/oasis-core/blob/master/go/consensus/cometbft/abci/prune.go#L119:L176) is not rate limited. Same applies for the blockstore, where pruning of the blocks is not rate limited.

In essence this is reverse of the issue https://github.com/oasisprotocol/oasis-core/issues/6334, where pruning is rate limited too much.

TODO: Also note that once the cometbft catches up via the block sync reactor, it is[ switched to consensus reactor](https://github.com/oasisprotocol/cometbft/blob/3988ee533111308ed2eedaa5614bcde6075d7ce1/blocksync/reactor.go#L331:L350). I would have to check but maybe this one is slower so once the node falls behind and was switched here it has trubles catching up again.

martintomazic:
This will be mitigated via https://github.com/oasisprotocol/oasis-core/pull/6341 and documented as part of the https://github.com/oasisprotocol/docs/issues/1469.

PR #6341, opened 2025-10-01
go/oasis-node/cmd/storage: Add command for offline pruning (POC) 

martintomazic:
**POC: quickly coded this up, will prepare it for review tmr, but amazingly it works as intended.** 

When enabling aggressive pruning node may fall behind (#6321). 

Whilst unlikely, the consequence may be the node breaking its commitment, which could results in lost rewards, possibly registering its availability when it is actually not ready, or worst case penalties (don't think this would happen actually).

To prevent it, we should offer validators a maintenance command that should be called if pruning is enabled later on, before starting the node. This way freshly started node is guaranteed to be healthy.

As with compaction command, I have started with consensus databases only and suggest to add runtime databases for the follow-up. Moreover, I quickly tested that this command does not prune past the minimum reindexed height, in case of configured runtimes.

As stated in the code, pruning CometBFT managed DB instances manually, will unfortunately always be a bit hacky (code wise). Also not sure this is actually intended use.

Definitely my suggestion would be to make this command experimental if we decide to proceed.

**_Calling `oasis-node storage compact`, immediately after `oasis-node storage prune` works as expected. I.e. it quickly removes redundant disk space._**


go/consensus/cometbft/db/init.go
```diff
@@ -2,23 +2,61 @@
...
 
 // New constructs a new CometBFT DB with the configured backend.
 func New(fn string, noSuffix bool) (dbm.DB, error) {
 	return badger.New(fn, noSuffix)
 }
+
+// OpenBlockstoreDB opens a CometBFT managed blockstore DB.
+//
+// This function is a hack as CometBFT does not expose a way to access the underlying databases.
+func OpenBlockstoreDB(provider cmtnode.DBProvider, cfg *cmtconfig.Config) (dbm.DB, error) {
```
martintomazic, line 34: Tentative commit: Not sure `init.go` is the best place?

go/oasis-node/cmd/storage/storage.go
```diff
@@ -395,15 +421,207 @@ func compactConsensusNodeDB(dataDir string) error {
...
+	// consensus height, so that light history can be populated correctly.
+	minReindexed, err := minReindexedHeight(dataDir, runtimes)
+	if err != nil {
+		return fmt.Errorf("failed to fetch earliest reindexed consensus height: %w", err)
+	}
+
+	retainHeight := min(
+		latest-numKept, // underflow not possible due to if above.
+		uint64(minReindexed),
+	)
```
martintomazic, line 509: This is still not optimal:

Assume you have 1 year of the data, but want to add a paratime node, so you need more space. You decide to change the config, by keeping only one month of the consensus data and configure the sapphire paratime. You start offline pruning, but as `minReindexed=0` (empty reindexed history), offline pruning will be skipped.

**Update**: I wouldn't complicate, first you prune, then you change the configuration again and add a paratime. We can reflect this in the documentation maybe.

go/oasis-node/cmd/storage/storage.go
```diff
@@ -395,15 +421,207 @@ func compactConsensusNodeDB(dataDir string) error {
...
+//
+// In case of no configured runtimes it returns max int64.
+func minReindexedHeight(dataDir string, runtimes []common.Namespace) (int64, error) {
+	fetchLastReindexedHeight := func(runtimeID common.Namespace) (int64, error) {
+		rtDir := runtimeConfig.GetRuntimeStateDir(dataDir, runtimeID)
+
+		history, err := history.New(runtimeID, rtDir, history.NewNonePrunerFactory(), true)
+		if err != nil {
+			return 0, fmt.Errorf("failed to open new light history: %w", err)
+		}
```
martintomazic, line 562: NIT: related to https://github.com/oasisprotocol/oasis-core/pull/6355#issuecomment-3390339488. Ideally opening runtime light history would not start or require to configure background workers.

go/oasis-node/cmd/storage/storage.go
```diff
@@ -395,15 +421,207 @@ func compactConsensusNodeDB(dataDir string) error {
...
+			minH = h
+		}
+	}
+
+	return minH, nil
+}
+
+func pruneCometDBs(dataDir string, retainHeight int64) error {
+	cmtConfig := cmtconfig.DefaultConfig()
+	_ = viper.Unmarshal(&cmtConfig)
```
martintomazic: Not sure why both `full` an `archive` node do viper unmarshal, nor why the error is ignored? So copying here but probably this should be fixed...

go/consensus/cometbft/db/init.go
martintomazic: This was added initially, naively thinking I would get around `wrapDbProvider` inside `full.go` and also be able to remove it, thus config was passed explicitly.

Since now this is only used in two places we may simply repeat the logic twice instead. We could also keep it but not pass provider and cfg explicitly, although I am trying to avoid of the functions that rely on the global state as much as possible.

Not strong opinion..

docs/oasis-node/cli.md
```diff
@@ -364,3 +364,26 @@ may stay constant or not be reclaimed for a very long time.
 
 This command gives operators manual control to release disk space during
 maintenance periods.
+
+### prune-experimental
+
+Run
+
+```sh
+oasis-node storage prune-experimental --config /path/to/config/file
```
ptrus, line 373: This should maybe mention that it requires that the node is stopped.

go/consensus/cometbft/db/init.go
```diff
@@ -2,23 +2,61 @@
...
+		return nil, fmt.Errorf("failed to open blockstore: %w", err)
+	}
+
+	return db, nil
+}
+
+// OpenStateDB opens a CometBFT managed state DB.
+//
+// This function is a hack as CometBFT does not expose a way to access the underlying databases.
+func OpenStateDB(provider cmtnode.DBProvider, cfg *cmtconfig.Config) (dbm.DB, error) {
```
ptrus, line 48: This method doesn't need to be exported. If i see correctly, it's always used like:

```go
stateDB, err := db.OpenStateDB(dbProvider, cmtConfig)
...
stateStore = db.OpenStateStore(stateDB)
```

So can probably just export `OpenStateStore`, unless i missed something? So clients would then just do:

```go
stateStore, err = db.OpenStateStore(dbProvider, cmtConfig)
```

go/oasis-node/cmd/storage/storage.go
```diff
@@ -395,15 +421,207 @@ func compactConsensusNodeDB(dataDir string) error {
...
+		logger.Info("skipping consensus pruning since disabled in the config")
+		return nil
+	}
+
+	runtimes, err := registry.GetConfiguredRuntimeIDs()
+	if err != nil {
+		return fmt.Errorf("failed to get configured runtimes: %w", err)
+	}
+
+	logger.Info("Starting consensus databases pruning. This may take a while...")
```
ptrus, line 468: Could we check/ensure somehow that the node is not running? Maybe could check socket location since we have the full config? And return a nice error message if the node is up, that this should be run with node stopped.

I assume that currently opening the DB would fail if the node is runnning and it might not be clear to the user what's the issue.

go/oasis-node/cmd/storage/storage.go
ptrus: Good question.

go/consensus/cometbft/db/init.go
martintomazic: Unfortunately you need `stateDB` so that you register it with the [`dbCloser`](https://github.com/oasisprotocol/oasis-core/blob/91d21cf783b21d66f63a3bb6cbfc79d2c47df8f7/go/consensus/cometbft/full/archive.go#L220).

martintomazic:
Rebased (no changes). 

Too see how this would fit in the big picture check the new (pending) pruning section: https://github.com/oasisprotocol/docs/pull/1526

go/oasis-node/cmd/storage/storage.go
```diff
@@ -395,15 +430,216 @@ func compactConsensusNodeDB(dataDir string) error {
...
+	stateDB, err := cdb.OpenStateDB(dbProvider, cmtConfig)
+	if err != nil {
+		return fmt.Errorf("failed to open state db: %w", err)
+	}
+	state := cdb.OpenStateStore(stateDB)
+	defer state.Close()
+
+	logger.Info("pruning consensus states", "base", base, "retain_height", retainHeight)
+	if err := state.PruneStates(base, retainHeight); err != nil {
+		return fmt.Errorf("failed to prune state db (start: %d, end: %d)", base, retainHeight)
```
ptrus: Maybe also wrap the error here.

go/oasis-node/cmd/storage/storage.go
```diff
@@ -6,26 +6,35 @@ import (
...
 
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
 	"github.com/oasisprotocol/oasis-core/go/config"
 	"github.com/oasisprotocol/oasis-core/go/consensus/cometbft/abci"
 	cmtCommon "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/common"
+	cmtConfig "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/config"
+	cdb "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/db"
+	dbProvider "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/db"
```
ptrus: `cdb` and `dbProvider` is duplicate import.

docs/oasis-node/cli.md
```diff
@@ -364,3 +364,26 @@ may stay constant or not be reclaimed for a very long time.
...
+
+to trigger manual pruning of consensus database instances:
+
+```sh
+{"caller":"storage.go:433","level":"info","module":"cmd/storage", \
+"msg":"Starting consensus databases pruning. This may take a while...", \
+"ts":"2025-10-23T11:02:11.129822974Z"}
+```
+
+Operatos should run this whenever they change pruning configuration, e.g. when
```
ptrus: Operators

go/oasis-node/cmd/storage/storage.go
ptrus: We can probably just remove this viper.Unmarshal here then?

ptrus (approved):

PR #6343, opened 2025-10-02
go/p2p: Add observer nodes to the target set of registered nodes

### LINKED ISSUE #6342: go/p2p: Add Observer nodes to the target set of registered nodes for the runtime protocols
As [pointed out](https://github.com/oasisprotocol/oasis-core/pull/6262#discussion_r2213076523) we should also include `Observer` nodes, under the set of target registered nodes to track.

martintomazic:
Closes #6342.

Not sure this will have much impact as these nodes were likely found via discovery (seed nodes, or possibly via libp2p if already connected with a given peer due to other protocol)...

### DIFF
--- a/go/worker/storage/p2p/checkpointsync/protocol.go
+++ b/go/worker/storage/p2p/checkpointsync/protocol.go
@@ -65,7 +65,7 @@ type GetCheckpointChunkResponse struct {
 func init() {
 	peermgmt.RegisterNodeHandler(&peermgmt.NodeHandlerBundle{
 		ProtocolsFn: func(n *node.Node, chainContext string) []core.ProtocolID {
-			if !n.HasRoles(node.RoleComputeWorker | node.RoleStorageRPC) {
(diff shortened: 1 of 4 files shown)


go/worker/storage/p2p/checkpointsync/protocol.go
```diff
@@ -65,7 +65,7 @@ type GetCheckpointChunkResponse struct {
 func init() {
 	peermgmt.RegisterNodeHandler(&peermgmt.NodeHandlerBundle{
 		ProtocolsFn: func(n *node.Node, chainContext string) []core.ProtocolID {
-			if !n.HasRoles(node.RoleComputeWorker | node.RoleStorageRPC) {
+			if !n.HasRoles(node.RoleComputeWorker | node.RoleStorageRPC | node.RoleObserver) {
```
peternose:
```suggestion
			if !n.HasRoles(node.RoleComputeWorker | node.RoleObserver | node.RoleStorageRPC) {
```
If possible, I try to keep the same order as in enum, so that it is easier to check which roles are missing. Huge nit :D

peternose (approved):

PR #6345, opened 2025-10-03
buildkite: Require Sonatype OSS Index auth for nancy audit

anzoman:
Implementing authentication for our Nancy audit Buildkite step, as Sonatype OSS Index now requires authentication.

### DIFF
--- a/.buildkite/go/nancy_audit.sh
+++ b/.buildkite/go/nancy_audit.sh
@@ -5,16 +5,40 @@
 # reported security vulnerabilities.
 #
 # Usage:
-# nancy_audit.sh
+# nancy_audit.sh [path_to_sonatype_oss_index_username] [path_to_sonatype_oss_index_token]
+#
+# path_to_sonatype_oss_index_username - Absolute or relative
+#     path to a file that contains the Sonatype OSS Index
+#     account username. Defaults to "~/.sonatype/oss_index_username".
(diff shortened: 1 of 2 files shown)


.buildkite/go/nancy_audit.sh
```diff
@@ -5,16 +5,40 @@
...
+sonatype_oss_index_token=$(cat ${path_to_sonatype_oss_index_token})
+set -x
+
 ########################################
 # Check dependencies for vulnerabilities
 ########################################
 pushd go/oasis-node
-    go list -json -deps | nancy sleuth -x ../.nancy-ignore
+    go list -json -deps | nancy sleuth \
+        --username ${sonatype_oss_index_username} \
```
ptrus: wont tracing leak these? it would be better if secrets are read into env variables in the part where we disable tracing (assuming nancy supports it)

peternose: Already did :)

kostko: Token has been rotated.

anzoman: Ugh, sorry, forgot to do `set +x` before and `set -x` after this command 🙃.
Anyways, using env vars `OSSI_USERNAME` and `OSSI_TOKEN` seems better and nancy supports it, so I used that.
I have also added a new token to Builkite.

PR #6346, opened 2025-10-06
go: Bump github.com/spf13/viper to v1.21.0

martintomazic:
**Motivation**: Fixes [Audit go dependencies]( https://buildkite.com/oasisprotocol/oasis-core-ci/builds/16027#0199b91d-a891-4015-a484-b38ea8de3758). See [vulnerability](https://ossindex.sonatype.org/vulnerability/CVE-2025-11065?component-type=golang&component-name=github.com%2Fmitchellh%2Fmapstructure&utm_source=nancy-client&utm_medium=integration&utm_content=1.0.33).

**Consideration**:
`v.1.20` was [breaking](https://github.com/spf13/viper/blob/master/UPGRADE.md#v120x), after a quick glance we should not be affected.


go/go.mod
```diff
@@ -3,12 +3,6 @@ module github.com/oasisprotocol/oasis-core/go
 replace (
 	github.com/cometbft/cometbft => github.com/oasisprotocol/cometbft v0.37.15-oasis2
 
-	// v1.5.0 has broken uint parsing, use my commit with fixes instead until
-	// the maintainers merge my PR: https://github.com/spf13/cast/pull/144
-	github.com/spf13/cast => github.com/oasisprotocol/cast v0.0.0-20220606122631-eba453e69641
-	// v1.18.0+ has broken handling of bound flags.
-	github.com/spf13/viper => github.com/spf13/viper v1.17.0
```
martintomazic, line 10: Quickly looking at the `cast` PR mentioned this was fixed upstream.

Not sure about bound flags error (pending a check)

martintomazic:
Anything in particular you would recommend to test/could be broken after this upgrade?

martintomazic:
^^ Nice CI caught smtg, will explore :):
```bash
level=error ts=2025-10-06T13:09:56.471894282Z caller=root.go:448 module=test-runner msg="failed to run scenario" err="root: failed to run scenario: failed to register runtime: failed to submit tx: exit status 1" scenario=e2e/runtime/trust-root/change run_id=0
```

go/go.mod
peternose: `cast`: The minimal working example now works, so it looks that it was fixed.
`viper`: Unit tests with v18.0 pass, so this problem is not captured with our tests. @kostko any idea what was wrong and if this has been fixed?

peternose (approved):

go/go.mod
martintomazic: >Not sure about bound flags error (pending a check)

This is still an issue in the newer versions. E.g. running `e2e/runtime/runtime-dynamic` scenario previously ~`MaxInRuntimeMessages` was set to `128` but after this change is set to `0`, thus [this](https://github.com/oasisprotocol/oasis-core/blob/fa873145e655e946bda7f84d5df2634862502e7a/go/roothash/api/api.go#L660) check fails.~
`rt.Executor.MaxMessages` and `params.MaxRuntimeMessages` were set to `0` and `128` respectively, but after this change this value is swapped, thus [this](https://github.com/oasisprotocol/oasis-core/blob/fa873145e655e946bda7f84d5df2634862502e7a/go/roothash/api/api.go#L657) check fails.

Happy to work on that but I assume this will not be a trivial fix, else it would have been done at that time already?

martintomazic: >Unit tests with v18.0 pass, so this problem is not captured with our tests.

It seems this is only relevant for e2e tests, where we do all the config gymnastics... Maybe is not even that hard to fix... update: or not captured via unit as you say.

kostko: This was more relevant in the past where flags were used for everything. Now most things are passed via the YAML config file. So it shouldn't be that hard to work around?

martintomazic: > Now most things are passed via the YAML config file. So it shouldn't be that hard to work around?

We still do this for the  [genesis command](https://github.com/oasisprotocol/oasis-core/blob/master/go/oasis-node/cmd/genesis/genesis.go#L786:L881) which seems to be the reason why `e2e/runtime/runtime-dynamic scenario` is failing. Moreover, searching for `viper.BindPFlags`, it seems there are quite many places that rely on this pattern.

I tried debugging [example](https://github.com/oasisprotocol/oasis-core/pull/6346#discussion_r2408679236) above, to see why the default value is no longer used after this upgrade:
1. During package `init` we store the [default value](https://github.com/oasisprotocol/oasis-core/blob/fa873145e655e946bda7f84d5df2634862502e7a/go/oasis-node/cmd/genesis/genesis.go#L846) into `initGenesisFlags`.
3. [Bind](https://github.com/oasisprotocol/oasis-core/blob/fa873145e655e946bda7f84d5df2634862502e7a/go/oasis-node/cmd/genesis/genesis.go#L876)  `initGenesisFlags` to viper.
4. [Add](https://github.com/oasisprotocol/oasis-core/blob/fa873145e655e946bda7f84d5df2634862502e7a/go/oasis-node/cmd/genesis/genesis.go#L766) `initGenesisFlags` to `initGenesisCmd` flags.
5. [Read](https://github.com/oasisprotocol/oasis-core/blob/fa873145e655e946bda7f84d5df2634862502e7a/go/oasis-node/cmd/genesis/genesis.go#L526) `CfgRoothashMaxRuntimeMessages` during genesis init command execution.
    * (Why not pass init genesis flagset explicitly and read from there?
    
I see two possible reasons why this upgrade broke things:
* >// v1.18.0+ has broken handling of bound flags. 
    * Most likely reason, but I am still lacking some context/understanding.
* https://github.com/spf13/viper/issues/233 
    * Also saw this one - not sure it applies here.

Related to above, (ideally) we should try to avoid global state (config) and `init` functions as they makes code hard to reason and test.  Furthermore, as from the documentation `viper` global instance [may get deprecated ](https://github.com/spf13/viper?tab=readme-ov-file#viper-or-vipers).

Given my limited knowledge of `pflags`, `viper` and `cobra` this seems far from trivial task, taking the above anti-patterns into consideration. _(That does not mean I am not willing to spent time fixing it)._

Since this is blocking `master`, my suggestion would be to temporary add `CVE-2025-11065` to `.nancy-ignore`, unless we find a simple solution for it tmr.

kostko: Since we use the config file for most things now, why use viper at all? Just use the flags directly. It shouldn't be too hard to change that, at least in the affected locations.

kostko: Added CVE-2025-11065 to .nancy-ignore until 2025-12-01 as the vulnerability seems quite irrelevant so we can do this bump in peace.

peternose: So the last commit in this PR should also undo commit [de3cc372d902ceec8f04c8e40b74676e580d345d](https://github.com/oasisprotocol/oasis-core/pull/6348/commits/de3cc372d902ceec8f04c8e40b74676e580d345d) from #6348.

Issue #6305, opened 2025-08-25
Fill runtime block headers (history reindex) in reverse


martintomazic:
**Feature**:
Assuming the consensus state is synced using the checkpoint sync, the earliest known runtime block header will be the one that corresponds to the checkpoint height. 

We could fill missing block headers in reverse, up until the specified height. This should be easy to proof, given that the current block header contains a hash that points to the previous block header.

This is expected to be slow, but should be significantly faster and less resource intensive than doing block sync (possibly from the genesis).

**Motivation**:
I see at least two scenarios:
1. You want paratime node with the whole runtime history. If you don't wan't to use [snapshots](https://snapshots.oasis.io/), you are forced to do block sync from the genesis just to reindex history.
2. Paratime node with corrupted consensus state. You run `oasis-node unsafe-reset`, and sync consensus again using checkpoint sync. Now it could happen that your last retained height (checkpoint height), is higher than your latest runtime height. Since you have a missing gap of block headers, your node gets [stuck](https://github.com/oasisprotocol/oasis-core/blob/2a01cbe4668099eaf028e73e21fe35db2c94b879/go/worker/storage/committee/node.go#L1282).

**Solution space**:
1. Runtime block history should be adapted slightly to allow adding blocks in reverse.
2. Fetch block headers from the peers (in parallel).

Priority very low for now?

peternose:
I believe there’s also an option to fill runtime block headers using a stateless client. For example, you could choose any consensus height you want and start reindexing runtime blocks from that point onward. This process could even run in parallel. Maybe this will already be a part of the stateless client once we implement configurable levels of statelessness (e.g. stateless consensus, state-full runtimes).

martintomazic:
>For example, you could choose any consensus height you want and start reindexing runtime blocks from that point onward.

Shouldn't the trust age be less than the debonding period else you would be risking a long range attack? 

So the stateless client security is similar to that of the light clients, i.e. your node can be off for sometime but should generally catch up at least once in a while, and downtime should be less than debonding period (also taking catch up + time to submit evidence in the consideration).

martintomazic:
Thought a bit more and I believe this would significantly speed-up the scenario of setting-up runtime observer node with all historical state. Also only syncing block headers is significantly less resources intensive, at least for the initial consensus sync.

[...]

PR #6355, opened 2025-10-10
go: Fix runtime pruning

### LINKED ISSUE #6352: go/runtime: NodeDB pruning nested in the light history prune transaction slows down the node
Part of #6334.

**Context**:
1.  Pruning of runtime light history (`History`) and corresponding state (`NodeDB`) is orchestrated from the light history worker (breaking SRP) - [see](https://github.com/oasisprotocol/oasis-core/blob/e88b25b60618409e3aee983464ea5d9d305749b1/go/runtime/history/history.go#L319:L352). 
2. The `pruneHandler`'s`Prune` operation is responsible for pruning runtime history, and relies on the [transaction rollback ](https://github.com/oasisprotocol/oasis-core/blob/e88b25b60618409e3aee983464ea5d9d305749b1/go/runtime/history/prune.go#L139), in case of any of the prune handlers failing.
3. There are two registered prune handlers:
    a. Executor prune handler that ensures we don't prune past last normal round - [see](https://github.com/oasisprotocol/oasis-core/blob/e88b25b60618409e3aee983464ea5d9d305749b1/go/worker/compute/executor/committee/prune.go#L16).
    b. Storage worker that in addition to to checking we don't prune past last synced round also [does the pruning](https://github.com/oasisprotocol/oasis-core/blob/e88b25b60618409e3aee983464ea5d9d305749b1/go/worker/storage/committee/prune.go#L31) of the `NodeDB` (again this breaks SRP).

**Problem**:
This is all nested into a single BadgerDB [write transaction](https://github.com/oasisprotocol/oasis-core/blob/e88b25b60618409e3aee983464ea5d9d305749b1/go/runtime/history/prune.go#L101:L157).

As pointed out in the parent issue, `NodeDB.Prune` becomes very slow for the runtime state with moderate number of versions. I believe this explains why we see no progress when it comes to syncing runtime state observed in the parent issue.

Observe increasing slowness of  `NodeDB.Prune` as number of versions grow is independent problem.

**Proposed solution:**
1. Move `Pruner` out of  `History` interface. Ideally the worker that initializes runtime's `NodeDB` and `History`, should be responsible for pruning. 
    * I.e. we register prune handlers to it and call prune of the corresponding databases.
2. Add `Prune` method to `History` interface.
3. Rename `Prune` to `CanPrune` where applicable and move the actual pruning out of the **3.b** (storage worker).

**Consideration**:
1. This will not be trivial as we currently don't have a shared worker/container where both `NodeDB` and `History` are both accessible at the same time.
    a. Who should be responsible for it in the first place? common worker? runtime registry? Should storage worker nest state and light history handling?
2. By having everything in the single transaction we had strong atomicity and isolation guarantees.
    b. We should be careful to not break those if we make this a multi-step process.

martintomazic:
Closes #6352.

Considerations  - https://github.com/oasisprotocol/oasis-core/pull/6355#issuecomment-3390393757

### DIFF
--- a/go/runtime/history/prune.go
+++ b/go/runtime/history/prune.go
@@ -128,30 +128,36 @@ func (p *keepLastPruner) Prune(latestRound uint64) error {
 			pruned = append(pruned, round)
 		}
 
-		// If there is nothing to prune, do not call any handlers.
-		if len(pruned) == 0 {
+		if len(pruned) == 0 { // nothing to prune
 			return nil
 		}
 
-		// Before pruning anything, run all prune handlers. If any of them
-		// fails we abort the prune.
-		p.mu.RLock()
-		defer p.mu.RUnlock()
-
-		for _, ph := range p.handlers {
-			if err := ph.CanPruneRuntime(pruned); err != nil {
-				p.logger.Debug("prune handler failed, aborting prune",
-					"err", err,
-					"round_count", len(pruned),
-					"round_min", pruned[0],
-					"round_max", pruned[len(pruned)-1],
-				)
-				return fmt.Errorf("runtime/history: prune handler failed: %w", err)
-			}
+		if err := p.canPrune(pruned); err != nil {
+			p.logger.Debug("prune handler blocked pruning, aborting prune",
+				"err", err,
+				"round_count", len(pruned),
+				"round_min", pruned[0],
+				"round_max", pruned[len(pruned)-1], // safe due to length check above
+			)
+			return fmt.Errorf("prune handler blocked pruning: %w", err)
 		}
 
 		return nil
 	})
+}
+
+// canPrune checks if all prune handlers allow pruning specified rounds.
+func (p *keepLastPruner) canPrune(rounds []uint64) error {
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+
+	for _, ph := range p.handlers {
+		if err := ph.CanPruneRuntime(rounds); err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 // PruneInterval implements Pruner.


martintomazic:
**This PR tries to fix existing problem simply, by avoiding being "perfect".**

However, for an ideal solution, we should first answer the question below, which also serves as a complementary context to what we are trying to solve here:

# Who should be responsible for pruning of the runtime state?
## Context
Seemingly trivial refactor clashes heavily with our current design/code organization.

Since it is not the first time that we are clashing with this specific abstractions, I want this change to also be a step towards making our code more maintainable instead of complicating it further.

## How things work currently

There are two databases used for maintaining runtime state:
 1. _Light history_ (`history.History`) - stores runtime block headers.
 2. _State DB_ (`NodeDB`) - stores runtime state.

### Who populates state instances?
_Light history_ is populated by the `indexer` that continuously updates it by (re)indexing consensus blocks.

_State DB_ is is populated by the storage committee worker by subscribing to incoming block headers.

The incoming block headers are populated by the common committee worker that the storage committee worker subscribes to via `NodeHook` implementation. Common committee worker gets block headers by subscribing to [ `History.WatchCommittedBlocks`](https://github.com/oasisprotocol/oasis-core/blob/a8f77ec7e0c5d1f2fb8d271cc6005849434f5bd4/go/worker/common/committee/node.go#L703). Moreover, storage committee worker has a direct access to the light history, to fallback on it in case of missing block headers. _This raises a question why the storage committee worker does not watch light history directly, making node hook redundant..._ (out of scope).

### Who handles the lifetime of the state instances?
The lifetime of the light history and corresponding indexer is currently owned by the runtime registry ([see](https://github.com/oasisprotocol/oasis-core/blob/a8f77ec7e0c5d1f2fb8d271cc6005849434f5bd4/go/runtime/registry/registry.go#L552:L566)).

_State DB_, on the other hand is created by the storage worker ([see](https://github.com/oasisprotocol/oasis-core/blob/a8f77ec7e0c5d1f2fb8d271cc6005849434f5bd4/go/worker/storage/worker.go#L97)), yet [closed](https://github.com/oasisprotocol/oasis-core/blob/a8f77ec7e0c5d1f2fb8d271cc6005849434f5bd4/go/runtime/registry/registry.go#L300) by the runtime registry.  

### Who is responsible for the pruning.

Currently pruning of the runtime state (both `History` and `NodeDB`) is the light history responsibility. This is the main reason why:
1. `History` API has out-grown its scope.
2.  #6352 bug.
3. Workaround where storage worker prune handler also has a side effect of actual pruning (causing another issue https://github.com/oasisprotocol/oasis-core/issues/6352#issuecomment-3390264966).

## Steps towards cleaner design
_We should get there incremental...._

1. **Light history shouldn't know anything about state syncing nor about pruning of the runtime state.**
    *  Thin interface over underlying DB, possibly implementing minimal business logic/caching.
    *  `StorageSyncCheckpoint`, `LastStorageSyncedRound`, `WaitRoundSynced`, `GetBlock`, `GetAnnotatedBlock`, `WatchBlocks`, `Prune` and `Pruner` are out of scope.
    * Factories are anti-patterns.
2. Pruning of the runtime state is independent process.
    * Accept pruning config, `NodeDB` and `History` and start pruning accordingly:    
    * Only started if `keep_n` strategy (no-op is an anti-pattern).
    * Later if things change (e.g. https://github.com/oasisprotocol/oasis-core/issues/6305#issuecomment-3329944917), with the following design its easy to change pruning to be more granular.
  3. Transparent lifetime of resources and processes:
      * A process opening a resource or starting a sub-process should be also responsible for closing it.
      * E.g. opening and closing of `NodeDB` via two different workers is problematic. 
  4. State syncing should be independent worker.
     * Accept `history.History` interface to populate `NodeDB` by fetching storage diffs.
  5. Storage committee worker should orchestrate many smaller workers ([#6307](https://github.com/oasisprotocol/oasis-core/issues/6307)).

## So who should be responsible for runtime state pruning?

Looking how closely related _Light history_ and _State DB_ are, especially as we have the cases where you need both, I believe there should be a single process responsible for instantiating and closing both. Other processes can receive it as a parameter.


### Storage committee worker

Observe that the storage committee worker is only started when you have [a local storage](https://github.com/oasisprotocol/oasis-core/blob/a8f77ec7e0c5d1f2fb8d271cc6005849434f5bd4/go/config/config.go#L62) and configured runtimes - [see](https://github.com/oasisprotocol/oasis-core/blob/a8f77ec7e0c5d1f2fb8d271cc6005849434f5bd4/go/worker/storage/worker.go#L45). So the first question should be do we have or envision a scenario where light history is being indexed but the storage worker should be disabled? Even if this is the case we could work around this. **update**: we have such scenario: https://github.com/oasisprotocol/oasis-core/pull/6385 (e.g. stateless client with configured paratime).

I am inclined towards making the storage worker responsible for Light history and State DB lifetime, populating and pruning them. Especially if we refactor the storage worker into many smaller ones. Note this would also mean moving the indexer into it.

Finally, I believe this worker should export methods like `LastStorageSyncedRound`, `WaitRoundSynced`, `GetBlock`, `GetAnnotatedBlock`, `WatchBlocks`, given it would have a direct access to the underlying resources  and the actual workers that populate and prune them. Observe that methods like  `History.StorageSyncCheckpoint`, becomes un-needed in this setup.

Other workers that rely on theses methods should define an interface and accept the storage committee worker as a parameter (therefore casting it to the subset of interface methods).

### Runtime registry

This is alternative to the above. Currently we don't have direct access to the `NodeDB` in the runtime registry unless we cast the [`Storage` backend ](https://github.com/oasisprotocol/oasis-core/blob/a8f77ec7e0c5d1f2fb8d271cc6005849434f5bd4/go/runtime/registry/registry.go#L115), that is actually a `LocalBackend` as can be [seen](https://github.com/oasisprotocol/oasis-core/blob/a8f77ec7e0c5d1f2fb8d271cc6005849434f5bd4/go/worker/storage/worker.go#L116) from its registration at storage worker.

Furthermore I am not really convinced by the idea of registry being responsible for pruning.

## Conslusion
_None of the solutions will be trivial to refactor to, so would appreciate any input here to get there incrementally._

martintomazic:
Tried to solve both problems described in https://github.com/oasisprotocol/oasis-core/issues/6352.

Opened a simple solution, that also suffers from some compromises, described commit by commit. For the clean solution we probably have to solve problems described in the [mini design doc](https://github.com/oasisprotocol/oasis-core/pull/6355#issuecomment-3390339488) first :/.

Prior to making this PR ready for review, writing tests and ensuring it works as expected (CI is green but haven't tested) I would appreciate if we could first discuss the high level direction.

go/runtime/history/db.go
```diff
@@ -270,6 +270,34 @@ func (d *DB) getLastBlock() (*roothash.AnnotatedBlock, error) {
...
+
+		var blk roothash.AnnotatedBlock
+		if err := item.Value(func(val []byte) error {
+			return cbor.UnmarshalTrusted(val, &blk)
+		}); err != nil {
+			return fmt.Errorf("failed to unmarshal earliest block: %w", err)
+		}
+
+		if blk.Block.Header.Round != round {
+			return fmt.Errorf("deleting non-earliest version (earliest: %d)", blk.Block.Header.Round)
```
ptrus:
```suggestion
			return fmt.Errorf("trying to delete non-earliest version (earliest: %d)", blk.Block.Header.Round)
```

martintomazic:
Freshly rebased on top of #6384, that also briefly touched this issue.

go/consensus/api/api.go
```diff
@@ -425,18 +425,15 @@ type StatePruner interface {
...
 
-// StatePruneHandler is a handler that is called when heights are pruned
+// StatePruneHandler is a handler that is called before heights are pruned
 // from history.
 type StatePruneHandler interface {
-	// Prune is called before the specified height is pruned.
-	//
-	// If an error is returned, pruning is aborted and the height is
-	// not pruned from history.
+	// CanPruneConsensus is called to check if the specified height can be pruned.
```
peternose, line 434: Why not just `CanPrune`? From consensus point of view, adding `Consensus` is just redundant.

go/consensus/cometbft/abci/prune.go
```diff
@@ -195,6 +190,16 @@ PruneLoop:
 	return nil
 }
 
+// canPrune checks if any of the prune handlers does not allow pruning of the given version.
```
peternose:
```suggestion
// canPrune checks if all prune handlers allow pruning the given version.
```
Comment without negation is simpler.

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +27,75 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
 		}
 
 		// Old suggestion: Make sure we don't prune rounds that need to be checkpointed but haven't been yet.
+	}
+
+	return nil
+}
+
+// statePruner is responsible for pruning of the runtime state
```
peternose:
```suggestion
// statePruner is responsible for pruning of the runtime state.
```

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +27,75 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
+//
+// Everytime pruning is triggered, the pruner checks for the earliest height
+// in the runtime light history and removes any older versions stored in the state db.
+//
+// TOD: This is not the most robust solution as developer changing the pruning of the
```
peternose:
```suggestion
// TODO: This is not the most robust solution as developer changing the pruning of the
```

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +27,75 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
+	logger       *logging.Logger
+}
+
+// newPruner creates new runtime state pruner.
+func newPruner(ndb mkvsDB.NodeDB, history history.History, interval time.Duration) *statePruner {
+	return &statePruner{
+		state:        ndb,
+		lightHistory: history,
+		interval:     interval,
+		logger:       logging.GetLogger("/worker/storage/committee/state-pruner").With("runtime_ID", history.RuntimeID()),
```
peternose:
```suggestion
		logger:       logging.GetLogger("/worker/storage/state-pruner").With("runtime_ID", history.RuntimeID()),
```

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +27,75 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
 
-		p.logger.Debug("pruning storage for round", "round", round)
+// serve periodically triggers the pruning of the runtime state db.
+func (sp *statePruner) serve(ctx context.Context) error {
+	sp.logger.Info("starting")
+	defer sp.logger.Info("stopped")
 
-		// Prune given block.
-		err := p.worker.localStorage.NodeDB().Prune(round)
+	ticker := time.NewTicker(sp.interval)
```
peternose:
```go
	for {
		select {
		case <-ctx.Done():
			return ctx.Err()
		case <-time.After(sp.interval):
		}

		if err := sp.prune(ctx); err != nil {
			sp.logger.Warn("failed to prune", "err", err)
		}
	}
```
Ticker doesn't take into account how long the pruning will last.

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +27,75 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
+		state:        ndb,
+		lightHistory: history,
+		interval:     interval,
+		logger:       logging.GetLogger("/worker/storage/committee/state-pruner").With("runtime_ID", history.RuntimeID()),
+	}
+}
 
-		p.logger.Debug("pruning storage for round", "round", round)
+// serve periodically triggers the pruning of the runtime state db.
+func (sp *statePruner) serve(ctx context.Context) error {
```
peternose:
```suggestion
func (p *statePruner) serve(ctx context.Context) error {
```

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +27,75 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
+
+func (sp *statePruner) prune(ctx context.Context) error {
+	blk, err := sp.lightHistory.GetEarliestBlock(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get earliest block from runtime light history: %w", err)
+	}
+
+	earliest := sp.state.GetEarliestVersion()
+
+	for v := earliest; v < blk.Header.Round; v++ {
```
peternose: I would keep using name round.

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +27,75 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
 
 		// Old suggestion: Make sure we don't prune rounds that need to be checkpointed but haven't been yet.
+	}
+
+	return nil
+}
+
+// statePruner is responsible for pruning of the runtime state
+//
+// Everytime pruning is triggered, the pruner checks for the earliest height
```
peternose: This should probably be in the method's comment.

go/worker/storage/worker.go
```diff
@@ -109,6 +110,7 @@ func (w *Worker) registerRuntime(commonNode *committeeCommon.Node) error {
 			Disabled:          config.GlobalConfig.Storage.CheckpointSyncDisabled,
 			ChunkFetcherCount: config.GlobalConfig.Storage.FetcherCount,
 		},
+		max(config.GlobalConfig.Runtime.Prune.Interval, time.Second),
```
peternose: I would move this to pruner as the pruner can dictate what is the minimum interval.

.changelog/6352.trivial.md
peternose, line 1: In commit message change description to: `There is additional change of semantics: Previously if youR state`.

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +27,75 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
+// in the runtime light history and removes any older versions stored in the state db.
+//
+// TOD: This is not the most robust solution as developer changing the pruning of the
+// runtime light history may also unexpectedly change the pruning behaviour of the state db.
+type statePruner struct {
```
peternose, line 46: Should we name it `runtimeStatePruner`?

go/runtime/history/db.go
```diff
@@ -270,6 +270,34 @@ func (d *DB) getLastBlock() (*roothash.AnnotatedBlock, error) {
 	return &blk, nil
 }
 
+func (d *DB) deleteBlock(round uint64) error {
+	return d.db.Update(func(tx *badger.Txn) error {
+		it := tx.NewIterator(badger.IteratorOptions{
+			Prefix: blockKeyFmt.Encode(),
```
peternose: This should not work. You should pass `round` and check if `it.Item()` returns the right key. 

Why don't you just do `return tx.Delete(blockKeyFmt.Encode(round))`?

go/runtime/history/history.go
```diff
@@ -32,6 +32,9 @@ type History interface {
 	roothash.BlockHistory
 	consensus.StatePruneHandler
 
+	// DeleteBlock deletes a block at a specified round if for the earliest round.
```
peternose: This comment is confusing.

peternose (commented):
I think that first 5 commits look good, even if we change them latter. So we could merge and continue in another PR.

go/worker/storage/committee/prune.go
martintomazic: I would keep it short given that `go/worker/storage/committee/prune.go` is runtime worker. 

Ideally yes this would be reflected in the package path already.

go/worker/storage/committee/prune.go
martintomazic: The reason I moved it at the top is that `prune` is somehow internal, and at the call site you will likely only read `new` and the type it returns...

go/worker/storage/committee/prune.go
martintomazic: Probably, albeit ticker is nice because it does not allocate new timer every loop (probably premature optimization though). Do you want me to also fix this for the runtime's light history where we use same pattern?

go/worker/storage/committee/prune.go
martintomazic: I suggest to add a follow-up issue and link it here? Also it would be good to add a comment where runtime's light history pruning is defined that changing it would also change runtime state pruning...?

go/consensus/api/api.go
martintomazic: True, although is easier to make sense of when implementing it inside `go/runtime/history/history.go` and also "symmetric" with `CanPruneRuntime`, so either or...

My suggestion would be to also get rid of
```go
    // Note that this can be called for the same height multiple
	// times (e.g., if one of the handlers fails but others succeed
	// and pruning is later retried).
```
for both given that `func Can*() error` implies this should be safe...

martintomazic:
>I think that first 5 commits look good

Agree and kept only those.

**Sanity checking semantic change (thinking aloud):**

Previously it could happen that state was pruned before the runtime light history  (or possibly missing, when e.g. doing runtime state sync from the checkpoint earlier than the last reatained light header block - [see](https://github.com/oasisprotocol/oasis-core/blob/3f3d39030275ddd2313bf971926c331bcc76d966/go/worker/client/service.go#L142:L151)). However, it never happened the other way.

This change makes this less likely (except for the late checkpoint scenario), as light history is now always pruned before the State DB, changing the assumptions the old code may have. Therefore we should double check that none of the existing code first queries state / last retained version and then automatically assume light history has corresponding light block for the queried height. I did not find any such place, nor would it be sign of the robust code had I found it.

Double check appreciated. 

**Follow-up (won't be actively worked on)** 

Opened https://github.com/oasisprotocol/oasis-core/issues/6400, and also referenced it in the code. As stated in the issue I believe the answer to _"So who should be responsible for runtime state pruning"_ is clear: the storage committee worker.

martintomazic:
> Previously it could happen that state was pruned before the runtime light history...  However, it never happened the other way. ... Therefore we should double check that other parts of the code not rely on this somewhere.
> I did not find any such place, nor would it be sign of the robust code had I found it.

https://github.com/oasisprotocol/oasis-core/issues/6403 made me think I have to check the storage committee worker once more:
* https://github.com/oasisprotocol/oasis-core/blob/c4cdfca1d4a17b67181a2911dcc07b18a31a65cb/go/worker/storage/committee/worker.go#L802-L819 is not promising as it seems in the corner case we could enter into `initGenesis`: https://github.com/oasisprotocol/oasis-core/blob/c4cdfca1d4a17b67181a2911dcc07b18a31a65cb/go/worker/storage/committee/worker.go#L829-L844

**Update**: false alarm

martintomazic:
^^ Tested / sanity checked everything yesterday and I believe if anything the code is more robust now. 

_Ready for the final review._

go/worker/storage/committee/prune.go
peternose: You could create a timer and reset it after competition. If you think `sp.interval` is long enough that there will extra time after pruning, this is also fine, or you can increase the interval also to ensure that.

I don't remember why I got the feeling that this extra time could be zero :thinking:

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +29,79 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
+}
+
+// statePruner handles pruning of the runtime state.
+//
+// Everytime pruning is triggered, the pruner removes rounds that are older than the earliest
+// round in the runtime’s light history.
+//
+// TODO: https://github.com/oasisprotocol/oasis-core/issues/6400
+//
+//	This is not robust as developer changing the pruning of the runtime light history
```
peternose:
```suggestion
// This is not robust as developer changing the pruning of the runtime light history
```
Replace tabs with spaces.

You should not start a paragraph with this, as you don't know to what it refers.

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +29,79 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
+
+	return nil
+}
+
+// statePruner handles pruning of the runtime state.
+//
+// Everytime pruning is triggered, the pruner removes rounds that are older than the earliest
+// round in the runtime’s light history.
+//
+// TODO: https://github.com/oasisprotocol/oasis-core/issues/6400
```
peternose:
```suggestion
TODO: This is not robust as developer ... . See https://github.com/oasisprotocol/oasis-core/issues/6400.
```

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +29,79 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
+//
+//	This is not robust as developer changing the pruning of the runtime light history
+//	may also unexpectedly change the pruning behaviour of the state db.
```
peternose:
```suggestion
//	may also unexpectedly change the pruning behavior of the state db.
```

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +29,79 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
+type statePruner struct {
+	state        mkvsDB.NodeDB
+	lightHistory history.History
+	interval     time.Duration
+	logger       *logging.Logger
+}
+
+// newPruner creates new runtime state pruner.
+//
+// The interval specifies how often pruning occurs and is capped at a minimum of one second.
```
peternose: I would remove this comment, as the minimum is so low, that people with not notice it. I would add it if the minimum was 1 minute or hour.

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +29,79 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
+//
+// TODO: https://github.com/oasisprotocol/oasis-core/issues/6400
+//
+//	This is not robust as developer changing the pruning of the runtime light history
+//	may also unexpectedly change the pruning behaviour of the state db.
+//	This could be fixed by making the storage committee worker responsible for both
+//	syncing and pruning of the light history and state DB.
+type statePruner struct {
+	state        mkvsDB.NodeDB
+	lightHistory history.History
```
peternose:
```suggestion
	history history.History
```

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +29,79 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
+	defer ticker.Stop()
 
-		// Prune given block.
-		err := p.worker.localStorage.NodeDB().Prune(round)
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+			if err := p.prune(ctx); err != nil {
```
peternose: Removing one indent is always good. Less indents, more readable code.

```go
	for {
		select {
		case <-ctx.Done():
			return ctx.Err()
		case <-ticker.C:
		}

		if err := p.prune(ctx); err != nil {
			p.logger.Warn("failed to prune", "err", err)
		}
	}
```

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +29,79 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
+				p.logger.Warn("failed to prune", "err", err)
+			}
+		}
+	}
+}
+
+func (p *statePruner) prune(ctx context.Context) error {
+	blk, err := p.lightHistory.GetEarliestBlock(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get earliest block from runtime light history: %w", err)
```
peternose:
```suggestion
		return fmt.Errorf("failed to get earliest block from runtime history: %w", err)
```
I would not use word `light`, as it is also not used in the description of the `History.

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +29,79 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
+	}
+
+	earliest := p.state.GetEarliestVersion()
+
+	for round := earliest; round < blk.Header.Round; round++ {
+		p.logger.Debug("pruning storage", "round", round)
```
peternose:
```suggestion
		p.logger.Debug("pruning", "round", round)
```

Name `storage` contradicts logger's name `worker/storage/state-pruner`, so I would just remove it as is redundant. So is it state or storage pruner?

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +29,79 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
+		err := p.state.Prune(round)
```
peternose:
```suggestion
		switch err := p.state.Prune(round); err {
```

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +29,79 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
 		switch err {
 		case nil:
 		case mkvsDB.ErrNotEarliest:
-			p.logger.Debug("skipping non-earliest round",
-				"round", round,
-			)
+			p.logger.Debug("skipping non-earliest round", "round", round)
 			continue
```
peternose:
```suggestion
```
There is no continue in the previous case.

go/worker/storage/committee/worker.go
```diff
@@ -166,6 +166,8 @@ type Worker struct {
 	diffCh     chan *fetchedDiff
 	finalizeCh chan finalizeResult
 
+	pruningInterval time.Duration
```
peternose: This could also be `pruneInterval`, as it is shorter and more related to config `...Prune.Interval`.

go/runtime/history/prune.go
```diff
@@ -142,8 +139,8 @@ func (p *keepLastPruner) Prune(latestRound uint64) error {
 		defer p.mu.RUnlock()
 
 		for _, ph := range p.handlers {
-			if err := ph.Prune(pruned); err != nil {
-				p.logger.Error("prune handler failed, aborting prune",
+			if err := ph.CanPruneRuntime(pruned); err != nil {
```
peternose: You could also move this to a helper function, like done before, together with locking.

go/worker/storage/committee/prune.go
```diff
@@ -12,8 +15,10 @@ type pruneHandler struct {
 	worker *Worker
 }
 
-func (p *pruneHandler) Prune(rounds []uint64) error {
-	// Make sure we never prune past what was synced.
+// CanPruneRuntime return no error when pruning would not go past last synced round.
```
peternose:
```suggestion
// CanPruneRuntime returns no error when pruning would not go past last synced round.
```

go/worker/storage/committee/prune.go
martintomazic: Removed it in all places including commit messages, as I agree with you we should keep things consistent.

Nevertheless, the  original structure (`History`) has a very poor name  as `history` is opaque for such high level concept. `LightHistory` on the other hand would capture exactly what this is and has a definitive meaning in the the blockchain domain. As a consequence this terminology then surfaces in our oasis docs/operators which is not best. Out of scope. Closing.

go/worker/storage/committee/prune.go
martintomazic: Actually, non-earliest can no longer happen, as we start from the earliest? Assuming we are the only one doing the pruning. See further simplification 1a35307c33f4050ce678865491ee2f33d4a475cb.

**update/sanity**: Also there is additional assumption for the upper bound (above is lower bound), i.e. earliest light history block is never pruned past the latest state db round. This includes empty node db (`0` earliest version) not producing warnings if pruning is triggered.

go/worker/storage/committee/prune.go
```diff
@@ -24,23 +29,68 @@ func (p *pruneHandler) Prune(rounds []uint64) error {
...
+	logger   *logging.Logger
+}
+
+// newPruner creates new runtime state pruner.
+func newPruner(ndb mkvsDB.NodeDB, history history.History, interval time.Duration) *statePruner {
+	return &statePruner{
+		state:    ndb,
+		history:  history,
+		interval: max(interval, time.Second),
+		logger:   logging.GetLogger("/worker/storage/state-pruner").With("runtime_ID", history.RuntimeID()),
```
peternose:
```suggestion
		logger:   logging.GetLogger("/worker/storage/state-pruner").With("runtime_id", history.RuntimeID()),
```

PR #6375, opened 2025-10-29
go/runtime/manager: Fix temporary bundle path

ptrus:
`root.Name()` already returns the full path. 

This might have caused some issues on the testnet provider after node upgrade:

```
 net.oasis.error: module: unknown code: 1 message: failed to open bundle: runtime/bundle: failed to open bundle: open /node/data/runtimes/tmp/bundles/node/data/runtimes/tmp/bundles/d8f764ac129c15684291a35ce606b0dbf6aeebd6205dbf2e3a69d41994b57598/instance-00000000000003a9: no such file or directory
```

`root.Name()` test:
```go
package main

import (
	"fmt"
	"os"
)

func main() {
	_ = os.MkdirAll("/tmp/x/y", 0o755)

	root, err := os.OpenRoot("/tmp/x/y")
	if err != nil {
		panic(err)
	}
	defer root.Close()

	fmt.Println("root.Name() =", root.Name())
}

// outputs: root.Name() = /tmp/x/y
```

### DIFF
--- a/.changelog/6375.bugfix.md
+++ b/.changelog/6375.bugfix.md
@@ -1,0 +1,1 @@
+go/runtime/manager: Fix temporary bundle path
--- a/go/runtime/bundle/manager.go
+++ b/go/runtime/bundle/manager.go
@@ -350,7 +350,7 @@ func (m *Manager) AddTemporary(tmpPath string, opts ...AddOption) error {
 	if err != nil {
 		return err
 	}
-	path := filepath.Join(m.tmpBundleDir, root.Name(), fi.Name())
+	path := filepath.Join(root.Name(), fi.Name())
 
 	return m.Add(path, opts...)
 }


kostko (approved):
Hm, how did this work before?

ptrus:
> Hm, how did this work before?

No idea. A symlink on the node for the duplicated dir fixes this. Maybe such a workaround was applied in the past?

peternose:
The fix is not fine. The example you gave is for one root only. If you use two roots, things are different, and you need to prefix.

```go

func main() {
	// First root.
	_ = os.MkdirAll("/tmp/x/y", 0o755)

	root, err := os.OpenRoot("/tmp/x/y")
	if err != nil {
		panic(err)
	}
	defer root.Close()

	fmt.Println("root.Name() =", root.Name())

	// Second root.
	_ = root.MkdirAll("a/b", 0o755)

	root2, err := root.OpenRoot("a/b")
	if err != nil {
		panic(err)
	}

	fmt.Println("root2.Name() =", root2.Name())
}

// outputs:
// root.Name() = /tmp/x/y
// root2.Name() = a/b
```

ptrus:
Hm yeah you are right, will revert and include a test for this.

PR #6376, opened 2025-10-29
revert: bea2709d6dc617de370019172149aa64e3f1f059

ptrus:
See: https://github.com/oasisprotocol/oasis-core/pull/6375#issuecomment-3460292501

### DIFF
--- a/.changelog/6375.bugfix.md
+++ b/.changelog/6375.bugfix.md
@@ -1,1 +1,0 @@
-go/runtime/manager: Fix temporary bundle path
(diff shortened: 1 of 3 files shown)


peternose (approved):

kostko (approved):

kostko:
But where does the duplicate path come from then?

peternose:
~~Maybe from label `net.oasis.origin`, but not sure how~~ :thinking:

peternose:
> But where does the duplicate path come from then?

Problem is in go 1.25.2+, while in 1.25.1 everything works as expected.

peternose:
https://github.com/golang/go/issues/75139

ptrus:
Can confirm the test with old code passes on 1.25.0.  And with the updated code passes on 1.25.2 😁

ptrus:
Replaced by https://github.com/oasisprotocol/oasis-core/pull/6377

PR #6381, opened 2025-10-30
go/runtime/txpool: Remove local queue

peternose:
Removes the local queue, which doesn’t support multiple transactions per sender. All incoming transactions are now added to the main queue.

### DIFF
--- a/go/runtime/txpool/txpool.go
+++ b/go/runtime/txpool/txpool.go
@@ -27,7 +27,7 @@ const (
 	// checkTxTimeout is the maximum time the runtime can spend checking transactions.
 	checkTxTimeout = 15 * time.Second
 	// checkTxRetryDelay is the time to wait before queuing a check tx retry.
-	checkTxRetryDelay = 1 * time.Second
+	checkTxRetryDelay = time.Second
 	// checkTxWaitRoundSyncedTimeout is the time to wait for block to be
 	// synced with storage at the start of check tx.
 	checkTxWaitRoundSyncedTimeout = 5 * time.Second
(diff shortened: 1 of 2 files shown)


go/runtime/txpool/txpool.go
```diff
@@ -627,12 +614,12 @@ func (t *txPool) checkTxBatch(ctx context.Context) error {
 			// Mark new transactions as never having been published. The republish worker will
 			// publish these immediately.
 			publishTime := time.Time{}
```
kostko: Why is this here given it is always immediately overwritten?

peternose: The linter also complains about this. I think I still need to separate local and non-local txs so that this publish time is set correctly. Still working on this.

go/runtime/txpool/txpool.go
```diff
@@ -107,10 +107,6 @@ type TransactionPool interface {
 	// and only the following transactions will be returned.
 	GetSchedulingExtra(offset *hash.Hash, limit int) []*TxQueueMeta
 
-	// FinishScheduling finishes a scheduling session, which resumes transaction rechecking and
-	// republishing.
-	FinishScheduling()
```
kostko, line 112: Why was this locking done before? Could there be issues now that there is no locking?

peternose: I think the main reason for the lock in the previous version of the main queue was that the tree the scheduler iterated over had to remain unchanged. Any modification could have caused the scheduler to produce unexpected results. However, in the current version, the queue supports adding and removing transactions even during scheduling.

peternose: For example, in the previous version, if the last scheduled transaction was replaced with a newer one, the scheduling of additional transactions would fail because the last scheduled transaction could no longer be found in the tree. As a result, the main queue wouldn’t know where to continue from.

The new version tracks nonces for each sender during scheduling. When additional transactions are needed, the queue simply checks whether a transaction with the next expected nonce is available, ignoring any changes to previously processed transactions.

PR #6385, opened 2025-11-02
client-stateless: GetLastRetainedHeight method returns light client's oldest height

### LINKED ISSUE #6382: oasis-node/stateless:  Incorrect GetLastRetainedHeight may block runtime history reindex
Configuring Paratime node using stateless client node for the consensus backend may cause the runtime history reindex to become stuck.

This is because [GetLastRetainedHeight](https://github.com/oasisprotocol/oasis-core/blob/b298a95240d74769e6f1e6ab9946575608bb39b0/go/consensus/cometbft/stateless/core.go#L158C16-L158C37) returns remote provider data. 

Instead, we should return the height of the earliest locally retained consensus light header. This way [reindexTo](https://github.com/oasisprotocol/oasis-core/blob/b298a95240d74769e6f1e6ab9946575608bb39b0/go/runtime/history/indexer.go#L294), won't start reindexing the range that is impossible to verify, avoiding the inifinite loop. (~In my case `start_height` was somehow `0`~, e.g. public grcp start height is `17751681`). 

Moreover, we should change the consensus light client pruning, to not prune past the last reindexed height (in case of configured runtimes). This would be similar to how is done right now for the full node...?

~Finally, might be a good idea to by default preserve sufficient consensus light block headers, to have sufficiently old runtime light history, so that paratime checkpoints could be found.~ update: this is not needed as stateless client does not have storage worker (stateless :) ).

martintomazic:
Closes #6382 

## How to test:
Run stateless client node with at least one configured runtime (triggers history reindex).

Prior to this fix, if you set your trust root to be younger than your configured provider's last retained height, your indexer will get stuck. This PR should fix this.

### DIFF
--- a/.changelog/6382.bugfix.md
+++ b/.changelog/6382.bugfix.md
@@ -1,0 +1,8 @@
+stateless-client: Fix incorrect GetLastRetainedHeight method
+
+Previously, this method would return one of the provider's last retained
+heights which could cause issues with runtime light history reindexing.
+
+This has been fixed, by setting the last retained height to the
+light client's last retained height, as this is the oldest height
+the stateless client can use to verify consensus data against.
--- a/go/consensus/cometbft/stateless/core.go
+++ b/go/consensus/cometbft/stateless/core.go
@@ -156,9 +156,9 @@ func (c *Core) GetGenesisDocument(context.Context) (*genesisAPI.Document, error)
 
 // GetLastRetainedHeight implements api.Backend.
 func (c *Core) GetLastRetainedHeight(ctx context.Context) (int64, error) {
-	// The last retained height cannot be verified, nor does it need to be,
-	// as it is not sensitive information.
-	return c.provider.GetLastRetainedHeight(ctx)
+	// The last retained height equals to the last retained light client height
+	// as this is the oldest height stateless client can verify the data against.
+	return c.lightClient.FirstTrustedHeight()
 }
 
 // GetLatestHeight implements api.Backend.


go/consensus/cometbft/light/client.go
```diff
@@ -122,6 +122,18 @@ func (c *Client) LastTrustedHeight() (int64, error) {
 	return height, nil
 }
 
+// FirstTrustedHeight returns the first (oldest) trusted height.
+func (c *Client) FirstTrustedHeight() (int64, error) {
```
ptrus, line 126: I think we've also been using `Earliest` in some other similar contexts in the code for the "oldest" height.

martintomazic: True, but not consistently. I think there are cases in our code or possibly Cometbft where `Earliest` effectively means "youngest".

I like `LastRetained` more in this context, but used the same terminology here as the underlying cometbft light client does.

Any preference/suggestion? Happy to standardize...

ptrus: First is fine as well

martintomazic:
This may cause reverse issues now, e.g. if your provider has less history then your configured trust.

Log when configuring trust height that is 9 days old against my localhost provider that has history from 8 days ago:

```bash
{"caller":"indexer.go:237","err":"failed to reindex batch: failed to get runtime state: failed to get state root: failed to get transactions from any provider: rpc error: code = Unknown desc = cometbft: block query failed: height 29025440 is not available, lowest height is 29051682","height":29165710,"level":"warn","module":"runtime/history/indexer","msg":"failed to reindex blocks","runtime_id":"000000000000000000000000000000000000000000000000a6d1e3ebf60dff6c","ts":"2025-11-03T10:07:47.573477627Z"}
```

_I believe this change is still correct, and the log is expected._ 

It is your responsibility to configure correct trust against the configured providers. Sure we could find a minimum of your trust and all providers last retained height. But then there is composed provider, scorer etc. Also what if the one provider with most retained history suddenly stops serving... Worst case you again rely on this log so we don't solve anything with complicating things.

Nevertheless, my suggestion is to reflect this requirement under _Setting trust root for stateless client node_ - "configured trust should be younger than your providers last retained height, taking their pruning and your speed of reindexing into account". See existing discussion https://github.com/oasisprotocol/docs/pull/1472#discussion_r2451969014. Realistically if you configure a recent/latest trust as [suggested](https://github.com/oasisprotocol/docs/pull/1472#discussion_r2485785109) this is not a problem at all.

peternose:
Yes, I was thinking about that during review, and the node should probably switch to another provider if data cannot be found. Another option would be to take `max`, but that raises problems as you would need to contact all providers.

martintomazic:
>Another option would be to take max, but that raises problems as you would need to contact all providers.

And even with that you are not safe as the max may prune right after. I suggest merging this for now, and encourage to use very recent if not latest trust root given that anything else is impractical and slow.

PR #6393, opened 2025-11-11
go/runtime/txpool: Add workaround for future txs bug

peternose:


go/runtime/txpool/txpool.go
```diff
@@ -571,25 +571,39 @@ func (t *txPool) checkTxBatch(ctx context.Context) error {
 	)
 
 	// Queue checked transactions for scheduling.
+	stateSeqNums := make(map[string]uint64)
 	for i, pct := range goodPcts {
-		if err = t.mainQueue.Add(pct.TxQueueMeta, results[batchIndices[i]].Meta); err != nil {
+		idx := batchIndices[i]
+		res := results[idx]
+
+		// XXXX: Temporary workaround for an Oasis SDK bug that incorrectly
```
kostko, line 579: I guess this only works if all of the transactions are in the same check batch. I assume this will catch most cases in practice though.

peternose: No, it should work otherwise too. Unless we can do sequential checks of multiple batches, where the next batch is checked from the state of the previous batch. But I don't think we do it that way.

For example, if I currently submit transactions with nonces 1,5,2,3 and my next nonce is 0, the check will succeed for all txs, but the result of the first transaction will say that my next nonce is 0 (sender state seq number), the next that is 1, ..., and the last that is 3. Once all these transactions are added to the main queue, the queue will think that the next nonce is 3 and will remove transactions with nonces 1 and 2. And now we are stuck as we will never accept transaction with nonce 0. If transactions with nonces 1,5,2,3 are split into two parts, e.g. 1,5 and 2,3, things are the same. Just that in this case the next nonce will be 1.

kostko: Ah, you're right.

PR #6397, opened 2025-11-13
go/consensus/api: Support fetching validator set

peternose:


go/consensus/cometbft/stateless/core.go
```diff
@@ -675,6 +708,23 @@ func (c *Core) verifyProof(ctx context.Context, proof *transaction.Proof, tx *tr
 	return nil
 }
 
+func (c *Core) verifyNextValidators(validators *consensusAPI.Validators, lb *cmttypes.LightBlock) error {
+	if validators.Height != lb.Height+1 {
```
peternose, line 712: Fixed to +1, as light block for next validator set is from previous height.

PR #6399, opened 2025-11-13
go/consensus/cometbft/full: Always return full light blocks

peternose:
This should be merged after runtimes support fetching of validator set.
- #6397


peternose:
Before we merge this PR, we need to:
- update Oasis SDK with Oasis Core 25.8 (already done with update to 25.9),
- release a new version of `runtime-sdk`, as `runtime-sdk/v0.16.2` doesn't contain these changes, but the next version will,
- update runtimes with the latest `runtime-sdk` version and deploy them on mainnet and testnet.

PR #6410, opened 2025-11-24
go: Per-role quote policies 

### LINKED ISSUE #6387: Support for compute runtime quote policy
Add support for per-role quote policies in addition to the existing general policy (which would become the "default" policy for roles that don't have an override). This would allow a more relaxed general policy but stricter requirements for nodes that can access the key manager (eg. compute/observer nodes).

martintomazic:
Wip.

Closes #6387, follows #6331

**Status**:
This is ready for preliminary high level review, mainly I would like us to focus on the TODO, that I also explained under _Challenges_ in the[ design doc](https://github.com/oasisprotocol/oasis-core/issues/6387#issuecomment-3744476044).

Once we align on the requirements, and finalize some implementation/design decision I will write the final code and thorough test suite, given this is security critical PR. Please ignore style and code quality until then.

### DIFF
--- a/go/registry/api/api.go
+++ b/go/registry/api/api.go
@@ -619,8 +619,10 @@ func VerifyRegisterNodeArgs( // nolint: gocyclo
 			// both validators and compute nodes and have out of date attestation evidence. Removing
 			// such nodes could lead to consensus not having the proper majority. This is safe as
 			// attestation evidence is independently verified before scheduling committees.
-			if err := VerifyNodeRuntimeEnclaveIDs(logger, &n, rt, regRt, params.TEEFeatures, now, height, isFeatureVersion242); err != nil && !isSanityCheck && !isGenesis {
-				return nil, nil, err
+			if !isSanityCheck && !isGenesis {
+				if err := VerifyNodeRuntimeEnclaveIDs(logger, &n, rt, regRt, params.TEEFeatures, now, height, isFeatureVersion242); err != nil {
+					return nil, nil, err
+				}
 			}
 
 			// Enforce what kinds of runtimes are allowed.


go/common/node/sgx.go
```diff
@@ -224,7 +227,7 @@ func (sa *SGXAttestation) Verify(
 	sc *SGXConstraints,
 	rak signature.PublicKey,
 	rek *x25519.PublicKey,
-	nodeID signature.PublicKey,
+	n *Node,
```
martintomazic, line 268: NIT: Followed the pattern from the admission policy. Alternative is to pass `nodeID` and `roles` (bitmask) by value. 

Passing whole struct by reference feels dangerous as accidental mutations in the helper functions (possibly in the future) may affect the logic upstream.

runtime/src/identity.rs
```diff
@@ -72,6 +72,10 @@ struct Inner {
     quote: Option<Arc<Quote>>,
     quote_timestamp: Option<i64>,
     quote_policy: Option<Arc<QuotePolicy>>,
+    // TODO Add per role quote policy.
+    // But if the runtime cannot trust the host how can it obtain the "role" that is being used for?
+    // Query consensus registry using `node_id`, to obtain the current node struct and thus it's roles?
+    // But the host could set any node_id...
```
martintomazic: ?

martintomazic: cc @peternose | @kostko.

kostko: It needs to query verified consensus state via the consensus verifier.

The host cannot spoof the node ID as it is included in the TEE capability which includes a node ID which is checked during on-chain registration.

Binding generated during attestation:
https://github.com/oasisprotocol/oasis-core/blob/f73f73c3ca2f3001df1b65dba72c8fa9c1d2fa25/runtime/src/attestation.rs#L143-L150

And verified by consensus during each registration refresh:
https://github.com/oasisprotocol/oasis-core/blob/f73f73c3ca2f3001df1b65dba72c8fa9c1d2fa25/go/common/node/sgx.go#L250

We could also change the way the host returns its identity by having the runtime produce a nonce which the host must sign using a special context and return the signature together with the public key. This would prove the host has access to the private key.

go/common/node/tee.go
```diff
@@ -34,18 +34,7 @@ type TEEFeaturesSGX struct {
...
-			sc.Policy = &quote.Policy{}
-		}
-		if sc.Policy.IAS == nil {
-			sc.Policy.IAS = fs.DefaultPolicy.IAS
-		}
-		if sc.Policy.PCS == nil && fs.PCS {
-			sc.Policy.PCS = fs.DefaultPolicy.PCS
-		}
-	}
+	sc.Policy = sc.Policy.ApplyDefault(fs.DefaultPolicy, fs.PCS)
```
peternose, line 37: There are too many polices in this statement, which means that noting is applied at all. I like previous version better.

Observe also that if `fs.DefaultPolicy==nil` and `sc.Policy==nil`, then `sc.Policy` is not `nil` as before. So you changed method `ApplyDefaultConstraints`.

go/common/sgx/quote/quote.go
```diff
@@ -81,3 +81,22 @@ func (p *Policy) Validate(isFeatureVersion242 bool) error {
 
 	return fmt.Errorf("fmspc whitelist should be empty")
 }
+
+func (p *Policy) ApplyDefault(d *Policy, pcsEnabled bool) *Policy {
```
peternose, line 85: A comment is missing. The reason why this functions accepts parameter `pcsEnabled` is because of `fs.PCS`, which is not the best. The signature should be `func (p *Policy) ApplyDefault(d *Policy) *Policy`, which means that this function is not well structured and probally not needed. What you need in this case is `ApplyDefaultIAS`  and `ApplyDefaultPCS`, and let the caller check `pcsEnabled`. However, this complicates things so I would just stick with the original version.

go/common/sgx/quote/quote.go
```diff
@@ -81,3 +81,22 @@ func (p *Policy) Validate(isFeatureVersion242 bool) error {
...
+	if d == nil {
```
peternose, line 86: I hate if parameters can be `nil`, especially the first one, which is the most important. You have this `if` just because `fs.DefaultPolicy` can be `nil`, which is a wrong reason.

go/common/sgx/quote/quote.go
```diff
@@ -81,3 +81,22 @@ func (p *Policy) Validate(isFeatureVersion242 bool) error {
...
+		return p
+	}
+
+	if p == nil {
+		p = &Policy{}
```
peternose, line 91: Returning a new struct in one place and receiver in the other, is unusual.

go/common/node/sgx.go
```diff
@@ -111,14 +111,20 @@ func (sc *SGXConstraints) ValidateBasic(cfg *TEEFeatures, isFeatureVersion242 bo
 		return fmt.Errorf("unsupported SGX constraints version: %d", sc.V)
 	}
 
-	// Check for TDX enablement.
-	if !cfg.SGX.TDX && sc.Policy.PCS != nil && sc.Policy.PCS.TDX != nil {
-		return fmt.Errorf("TDX policy not supported")
+	validatePolicy := func(policy *quote.Policy) error {
```
peternose, line 114: Why the need for a helper function, if you can just put it under `if`?

I think that this is a breaking change. But it should be fine, as older nodes will panic and stop.

go/common/node/sgx.go
```diff
@@ -111,14 +111,20 @@ func (sc *SGXConstraints) ValidateBasic(cfg *TEEFeatures, isFeatureVersion242 bo
...
+		}
+		return nil
 	}
 
-	// Check that policy is compliant with the current feature version.
+	// Check default policy.
 	if sc.Policy != nil {
-		if err := sc.Policy.Validate(isFeatureVersion242); err != nil {
+		if err := validatePolicy(sc.Policy); err != nil {
 			return fmt.Errorf("invalid policy: %w", err)
```
peternose, line 128: Now you return error `invalid policy: invalid policy: ....`.

go/common/sgx/pcs/policy.go
```diff
@@ -28,6 +30,68 @@ type QuotePolicy struct {
...
+
+	merged := &QuotePolicy{
+		Disabled:                   p.Disabled || o.Disabled,
+		TCBValidityPeriod:          min(p.TCBValidityPeriod, o.TCBValidityPeriod),
+		MinTCBEvaluationDataNumber: max(p.MinTCBEvaluationDataNumber, o.MinTCBEvaluationDataNumber),
+	}
+
+	func() {
+		if len(p.FMSPCWhitelist) == 0 {
+			merged.FMSPCWhitelist = o.FMSPCWhitelist
```
peternose, line 58: I would create a copy of the list, in case someone alters one of the policies.

go/common/sgx/pcs/policy.go
```diff
@@ -28,6 +30,68 @@ type QuotePolicy struct {
...
+		for _, fmspc := range p.FMSPCWhitelist {
+			intersect[fmspc] = struct{}{}
+		}
+		for _, fmspc := range o.FMSPCWhitelist {
+			if _, ok := intersect[fmspc]; ok {
+				merged.FMSPCWhitelist = append(merged.FMSPCWhitelist, fmspc)
+			}
+		}
+
+		// Preventing no intersection meaning allow any.
```
peternose, line 77: Confusing comment. Maybe is `// Reject always if there is no intersection` better.

go/common/sgx/pcs/policy.go
```diff
@@ -28,6 +30,68 @@ type QuotePolicy struct {
...
+		return p
+	}
+
+	merged := &QuotePolicy{
+		Disabled:                   p.Disabled || o.Disabled,
+		TCBValidityPeriod:          min(p.TCBValidityPeriod, o.TCBValidityPeriod),
+		MinTCBEvaluationDataNumber: max(p.MinTCBEvaluationDataNumber, o.MinTCBEvaluationDataNumber),
+	}
+
+	func() {
```
peternose, line 56: Could also use `switch`, might be more readable. Any maybe create helper functions for joining whitelists and blacklists.

go/common/sgx/pcs/policy.go
```diff
@@ -28,6 +30,68 @@ type QuotePolicy struct {
...
 
+// Merge merges two QuotePolicies into one, taking more restrictive configuration into account.
+//
+// TODO:
+//   - What if FMSCPWhitelist has no intersection (same applies for TDXQuotePolicy)?
+//   - Should we even allow registration of such runtime descriptor?
+//   - Finish TDX quote policy merge (ugly).
+//   - Unit tests.
+//   - Merge should produce independent copies to not accidentally mutate stuff.
+func (p *QuotePolicy) Merge(o *QuotePolicy) *QuotePolicy {
```
peternose, line 41: Could also have a more general function `Merge(policies ...*QuotePolicy)` or `Merge(policies []*QuotePolicy)`.

Is this method even needed, as you can just check if all policies holds instead of merging them into one and checking if that one holds?

go/common/sgx/pcs/policy.go
```diff
@@ -28,6 +30,68 @@ type QuotePolicy struct {
...
+		}
+
+	}()
+
+	union := make(map[string]struct{}, len(p.FMSPCBlacklist)+len(o.FMSPCBlacklist))
+	for _, fmspc := range append(p.FMSPCBlacklist, o.FMSPCBlacklist...) {
+		union[fmspc] = struct{}{}
+	}
+	merged.FMSPCBlacklist = slices.Collect(maps.Keys(union))
+	slices.Sort(merged.FMSPCBlacklist)
```
peternose, line 89: You should add a comment why we sort blacklist. Should we also sort whitelist, so that the order of policies to be merged is not important?

go/common/sgx/pcs/policy.go
```diff
@@ -28,6 +30,68 @@ type QuotePolicy struct {
...
+
+	merged.TDX = p.TDX.Merge(o.TDX)
```
peternose, line 91: I prefer preparing all variables and in the end creating return value, so that it is more transparent which fields were set and which not.

go/common/node/sgx.go
```diff
@@ -138,6 +157,25 @@ func (sc *SGXConstraints) ContainsEnclave(eid sgx.EnclaveIdentity) bool {
...
+// EffectivePolicy returns a combined policy. The combined policy may in addition to the default policy,
+// (depending on the specified roles) also include additional per-role policies.
+func (sc *SGXConstraints) EffectivePolicy(roles RolesMask) *quote.Policy {
+	effectivePolicy := sc.Policy // TODO: Make this a deep copy or better yet ensure Merge produces it.
+	if effectivePolicy == nil {
+		effectivePolicy = &quote.Policy{}
+	}
+
+	for role, policy := range sc.PerRolePolicy {
+		if role&roles == 0 {
```
peternose, line 169: Could add method `roles.Has(role)`, `roles.Is(role)` or `roles.Contains(role)`.

go/common/node/sgx.go
```diff
@@ -129,6 +132,22 @@ func (sc *SGXConstraints) ValidateBasic(cfg *TEEFeatures, isFeatureVersion242 bo
 		}
 	}
 
+	// Check per-role policies.
+	if !isFeatureVersion242 && sc.PerRolePolicy != nil {
+		return fmt.Errorf("per role policy should be empty")
```
peternose, line 137:
```suggestion
		return fmt.Errorf("per role policy should be nil")
```

go/common/node/sgx.go
```diff
@@ -235,8 +274,9 @@ func (sa *SGXAttestation) Verify(
 	// Use defaults from consensus parameters.
 	cfg.SGX.ApplyDefaultConstraints(sc)
 
-	// Verify the quote.
-	verifiedQuote, err := sa.Quote.Verify(sc.Policy, ts)
+	// Verify the quote againt the effective policy.
+	policy := sc.EffectivePolicy(n.Roles)
```
peternose, line 278: Would it be better to just verify for every policy? Because in general, policies could not be merged efficiently.

PR #6411, opened 2025-11-26
go/worker/storage: Remove hook for watching committed blocks

peternose:
The goal is to make the storage worker less dependent on the common worker to make it easier to modify each independently.

### DIFF
--- a/go/worker/storage/committee/worker.go
+++ b/go/worker/storage/committee/worker.go
@@ -11,8 +11,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/eapache/channels"
-
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
 	"github.com/oasisprotocol/oasis-core/go/common/node"
 	"github.com/oasisprotocol/oasis-core/go/common/pubsub"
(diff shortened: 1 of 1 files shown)


go/worker/storage/committee/worker.go
```diff
@@ -336,9 +332,8 @@ func (w *Worker) GetLocalStorage() storageApi.LocalBackend {
 // NodeHooks implementation.
```
martintomazic, line 332:
```suggestion
```
And remove both hook methods below and don't add this hook anymore?

martintomazic (approved):
Semantically only synchronization changed. Previously we were guaranteed the common committee worker would process the block before ~remaining~ this worker~s~ would start processing it. Looks safe here.

Do you see any place where we relied on this synchronization (maybe we were not aware of this)? Relevant if you start doing this for other workers.

go/worker/storage/committee/worker.go
peternose: No, as we still use `Initialized`.

peternose:
> Do you see any place where we relied on this synchronization (maybe we were not aware of this)?

I couldn't find any.
- The executor worker uses `WaitRoundSynced` to wait until the round is synced, so it doesn't need to be synced with the storage worker through the common worker. They also both get blocks at the same time and queue them in a channel, so they already work async.
- The common worker updates current descriptor, changes committee during epoch transitions, and sends block and incoming messages to the transaction pool, before calling `HandleNewBlockLocked`. None of which rely or are/should be connected to storage sync. So if this is done after block is synced, I don't see any problem.

PR #6412, opened 2025-11-26
go/worker/client/committee: Refactor code

peternose:


go/worker/compute/executor/committee/node.go
```diff
@@ -400,8 +400,6 @@ func (n *Node) scheduleBatch(ctx context.Context, round uint64, force bool) {
 		// We have queued incoming runtime messages (and batch timeout expired), schedule batch.
 	case n.rtState.LastNormalRound == n.rtState.GenesisBlock.Header.Round:
 		// This is the runtime genesis, schedule batch.
-	case n.rtState.LastNormalHeight < n.epoch.GetEpochHeight():
-		// No block in this epoch processed by runtime yet, schedule batch.
```
peternose, line 403: I cannot find a reason why this was added. We already create a new empty block every epoch, and even if we didn’t, this change does not guarantee that there will be one block every epoch.

go/worker/common/committee/group.go
```diff
@@ -34,107 +35,42 @@ func TagForCommittee(kind scheduler.CommitteeKind) string {
...
-
-	executorCommittee *CommitteeInfo
 
 	nodes nodes.VersionedNodeDescriptorWatcher
 }
 
-// IsValid checks whether the given epoch snapshot is valid (represents an actual epoch).
-func (e *EpochSnapshot) IsValid() bool {
-	return e.identity != nil
+// IsMember checks if the current node is a member of the executor committee.
```
kostko:
```suggestion
// IsMember checks if the current node is a member of the committee.
```

go/worker/common/committee/group.go
```diff
@@ -34,107 +35,42 @@ func TagForCommittee(kind scheduler.CommitteeKind) string {
...
+func (ci *CommitteeInfo) IsMember() bool {
+	return len(ci.Roles) > 0
 }
 
-// GetRuntime returns the current runtime descriptor.
-func (e *EpochSnapshot) GetRuntime() *registry.Runtime {
-	return e.runtime
+// IsWorker checks if the current node is a worker of the executor committee.
```
kostko:
```suggestion
// IsWorker checks if the current node is a worker of the committee.
```

go/worker/common/committee/group.go
```diff
@@ -34,107 +35,42 @@ func TagForCommittee(kind scheduler.CommitteeKind) string {
...
-
-// IsExecutorBackupWorker checks if the current node is a backup worker of the executor
-// committee in the current epoch.
-func (e *EpochSnapshot) IsExecutorBackupWorker() bool {
-	if e.executorCommittee == nil {
-		return false
-	}
-	return e.executorCommittee.HasRole(scheduler.RoleBackupWorker)
+// IsBackupWorker checks if the current node is a backup worker of the executor
+// committee.
```
kostko:
```suggestion
// IsBackupWorker checks if the current node is a backup worker of the committee.
```

go/worker/compute/executor/committee/node.go
kostko: Yeah I believe this was before empty blocks were created all the time.

PR #6413, opened 2025-11-27
registry/api: Use nil instead of empty slices

peternose:
Reverting slice changes from:
- https://github.com/oasisprotocol/oasis-core/pull/6383

### DIFF
--- a/.changelog/6413.bugfix.md
+++ b/.changelog/6413.bugfix.md
@@ -1,0 +1,1 @@
+registry/api: Use nil instead of empty slices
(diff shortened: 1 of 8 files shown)


ptrus (approved):

peternose:
> Can we add some tests to make sure this doesn't happen again?

Added tests for the two problematic functions. Other three are not, as we only iterate over returned slices. 

Observe that in theory we could have `nil` and empty slices in the consensus, if there is some other function out there that returns empty instead of `nil` (the other way around). That is not breaking, just not consistent.

The best solution would be to create a breaking change and always ignore empty thresholds slices, but that is probably too much and will make code uglier.

PR #6417, opened 2025-11-28
changelog: Assemble changes for 25.9 release

peternose:


go/common/version/version.go
```diff
@@ -148,7 +148,7 @@ var (
 	// the runtime.
 	//
 	// NOTE: This version must be synced with runtime/src/common/version.rs.
-	RuntimeHostProtocol = Version{Major: 5, Minor: 1, Patch: 0}
+	RuntimeHostProtocol = Version{Major: 6, Minor: 0, Patch: 0}
```
kostko: This will cause issues as old runtimes will not work AFAIK.

peternose: Yes, checked the code, the major needs to be the same. Removing this version bump.

PR #6418, opened 2025-11-28
go/common/version: Bump runtime host protocol version to v6.0.0

peternose:
When fetching a light block from consensus, the runtime host protocol must support validator set retrieval, which was already introduced in version v25.8.


kostko (commented):
This will break existing runtimes if I remember correctly.

PR #6424, opened 2025-12-03
go/worker/client/committee/node: Allow unlimited messages in queries

peternose:
I think we don't gain anything by limiting the number of messages in queries:
- Node operators can overwrite this value, as the value is not verified in the runtime.
- Some queries might last longer and not throw out-of-messages error before they finish. And because we don’t meter gas for queries, any query can still become a potential DoS vector regardless of this limit.

### DIFF
--- a/go/worker/client/committee/node.go
+++ b/go/worker/client/committee/node.go
@@ -3,6 +3,7 @@ package committee
 import (
 	"context"
 	"fmt"
+	"math"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


peternose:
> What is the reasoning behind removing this?

Just trying to decouple common, client and executor worker by removing active descriptor from the client node.

> I believe this is needed for correct gas estimation, otherwise price of messages may be incorrect. Please verify with the SDK before removing this.

Does this hold for queries? I thought that gas estimation is done via simulate calls.

peternose:
> Does this hold for queries? I thought that gas estimation is done via simulate calls.

Oh, I see that gas estimation is done via queries, and that the number of max messages influences the price of one message. But the calculation is not intuitive, especially not [these lines](https://github.com/oasisprotocol/oasis-sdk/blob/main/runtime-sdk/src/modules/core/mod.rs#L1199-L1202).

peternose:
Closing as affects gas estimation.

kostko:
What are simulate calls? Simulation is usually invoked via queries.

peternose:
> What are simulate calls? Simulation is usually invoked via queries.

I was talking about `evm.SimulateCall`. But yes, I checked the code and saw what you wrote, that simulation for gas estimation is invoked via `core.EstimageGas` query.

kostko:
Yeah but even `evm.SimulateCall` is a query :-)

PR #6425, opened 2025-12-04
go/worker/common: Replace epoch transitions with committee transitions

peternose:
In the future there will be no empty `EpochTransition` runtime blocks, so we need to change the committee worker to be able to run without relying on `HeaderType` of runtime blocks.

### DIFF
--- a/go/worker/compute/executor/committee/node.go
+++ b/go/worker/compute/executor/committee/node.go
@@ -344,14 +344,21 @@ func (n *Node) scheduleBatch(ctx context.Context, round uint64, force bool) {
 
 	// If the next block will be an epoch transition block, do not propose anything as it will be
 	// reverted anyway (since the committee will change).
-	epochState, err := n.commonNode.Consensus.Beacon().GetFutureEpoch(ctx, n.dispatchInfo.BlockInfo.ConsensusBlock.Height)
+	height, err := n.commonNode.Consensus.Core().GetLatestHeight(ctx)
+	if err != nil {
+		n.logger.Error("failed to fetch latest height",
+			"err", err,
+		)
+		return
+	}
+	epochState, err := n.commonNode.Consensus.Beacon().GetFutureEpoch(ctx, height)
 	if err != nil {
 		n.logger.Error("failed to fetch future epoch state",
 			"err", err,
 		)
 		return
 	}
-	if epochState != nil && epochState.Height == n.dispatchInfo.BlockInfo.ConsensusBlock.Height+1 {
+	if epochState != nil && epochState.Height == height+1 {
 		n.logger.Debug("not scheduling, next consensus block is an epoch transition")
 		return
 	}


go/worker/common/committee/node.go
```diff
@@ -654,9 +459,12 @@ func (n *Node) worker() {
 		case <-n.stopCh:
 			n.logger.Info("termination requested")
 			return
+		case cm := <-cmCh:
+			n.handleCommittee(n.ctx, cm)
```
kostko, line 463: Is it ok that this is now async with respect to blocks?

go/worker/common/committee/node.go
```diff
@@ -690,75 +498,135 @@ func (n *Node) worker() {
...
+	if committee.Kind != scheduler.KindComputeExecutor {
+		return
+	}
+	if committee.RuntimeID != n.Runtime.ID() {
+		return
+	}
+
+	rs, err := n.Consensus.RootHash().GetRuntimeState(ctx, &roothash.RuntimeRequest{
+		RuntimeID: n.Runtime.ID(),
+		Height:    consensus.HeightLatest,
```
kostko, line 511: Should this be synced by height instead of using "latest"?

go/worker/common/committee/node.go
peternose: > Is it ok that this is now async with respect to blocks?

It is not perfect, but should be fine.

Ideally, we would track consensus blocks, and for every block we would check if the committee has changed and if a new runtime block has been produced. However, this requires a lot of state queries and doesn't use available functions.

So what can go wrong?
- If the committee channel is behind the block channel, the node could schedule blocks or send commitments even though it should not (e.g. not in committee). This is not that big of a deal as the consensus will block these transactions. The node could also propose blocks that don't respect the active runtime descriptor and could be punished. But this is the same as if the node would be lagging behind.
- If ahead, the dispatch info handler function will skip old blocks, so here there are no problems.

go/worker/common/committee/node.go
peternose: > Should this be synced by height instead of using "latest"?

We don't know at which height the committee was elected. Actually, all channels that we expose will never tell you on which height something happened, which is troublesome if you want to be in sync. Since the height is unknown, we can always use the latest. This is not a problem because:
- If the committee hasn't changed, things will work just fine.
- If the committee has changed, we will use a newer committee here, so scheduling will work with the latest committee. However, a new committee event will in that case arrive soon, which will invalidate everything and use again the latest committee. In this case all work (block scheduling) will probably be thrown away, but again, I don't see any problems.

kostko: > all channels that we expose will never tell you on which height something happened

Except the annotated roothash blocks.

But sounds good, we should just make sure to test this well, maybe run some long-term tests.

kostko (approved):

Issue #6315, opened 2025-09-04
Remove usage of github.com/eapache/channels


martintomazic:
As can be see from the [recent commit](https://github.com/eapache/channels/commit/b10d48acd4cfb693c919f5ef3d5df40eaaa48f86), this library is no longer maintained and has bugs.

I suggest for the new code to stop using this package and all libraries that internally depend on it, to ease  the eventual removal.

Related to this, we could deprecate some methods inside `go/common/pubsub`. Concretely, `Subscribe` and `SubscribeEx`. Infinite channels are prone to memory leaks and overall encourage to write less robust clients. Finally, silently pilling the data is always a bad idea.

Instead we should favor `SubscribeBuffered`, forcing the client to define buffer size and strategies if data is dropped. This should be easily refactored internally to only use buffered channel per subscriber, and to drop data for the slow consumers, avoiding the need for the above package altogether.

 When dropping the data is not an option we should use a back-pressure, e.g. buffered channel only.

PR #6444, opened 2026-01-18
go/worker/storage/committee: Fix teardown

martintomazic:
https://github.com/oasisprotocol/internal-ops/issues/1317#issuecomment-3765815368 showed that in case of corrupted storage, our worker teardown might get stuck.

## How to test

Start your node with paratime configured and return a dummy error [here](https://github.com/oasisprotocol/oasis-core/blob/4aca0773e2257c619233f3dfc649faf73168b39a/go/worker/storage/committee/worker.go#L1280).

Prior to this change indexer would continue, whilst storage worker would get stuck at teardown.

It would be nice to have a test for this, but we would need to completely refactor storage worker first. Mainly all the state DB, p2p and other stuff should be passed as parameters, so that errors can be mocked in the "integration" tests.

### DIFF
--- a/go/worker/storage/committee/worker.go
+++ b/go/worker/storage/committee/worker.go
@@ -1046,6 +1046,8 @@ func (w *Worker) Serve(ctx context.Context) error { // nolint: gocyclo
 	fetchPool := workerpool.New("storage_fetch/" + w.commonNode.Runtime.ID().String())
 	fetchPool.Resize(config.GlobalConfig.Storage.FetcherCount)
 	defer fetchPool.Stop()
+	poolCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
 
 	triggerRoundFetches := func() {
 		for i := lastFullyAppliedRound + 1; i <= latestBlockRound; i++ {
@@ -1096,11 +1098,8 @@ func (w *Worker) Serve(ctx context.Context) error { // nolint: gocyclo
 				rootType := prevRoots[i].Type
 				if !syncing.outstanding.contains(rootType) && syncing.awaitingRetry.contains(rootType) {
 					syncing.scheduleDiff(rootType)
-					doneCh := fetchPool.Submit(func() {
-						w.fetchDiff(ctx, this.Round, prevRoots[i], this.Roots[i])
-					})
-					wg.Go(func() {
-						<-doneCh
+					_ = fetchPool.Submit(func() {
+						w.fetchDiff(poolCtx, this.Round, prevRoots[i], this.Roots[i])
 					})
 				}
 			}
--- a/.changelog/6444.bugfix.md
+++ b/.changelog/6444.bugfix.md
@@ -1,0 +1,1 @@
+go/worker/storage/committee: Fix worker teardown


go/common/workerpool/workerpool.go
```diff
@@ -92,7 +96,7 @@ func (p *Pool) Quit() <-chan struct{} {
 }
 
 // Submit adds a task to the pool's queue and returns a channel that will be closed
-// once the task is complete.
+// once the task is complete or the pool is stopped.
 func (p *Pool) Submit(job func()) <-chan struct{} {
```
martintomazic: I believe correct way (over-engineering) would be:
`func (p *Pool) Submit(jobCtx context.Context, job func(ctx context.Context)) <-chan struct{} {`

First parameter would be the job context. Pool would also have internal context. The job should be canceled either if the job context was canceled or if the pool context was canceled. Pool context would be canceled by the `pool.Stop` (which should possibly block until all jobs have finished, or separate method for that).

Current pattern e.g.:
```go
	doneCh := fetchPool.Submit(func() {
	    w.fetchDiff(ctx, this.Round, prevRoots[i], this.Roots[i])
	})
	wg.Go(func() {
	    <-doneCh
	})
``` 
I find non-idiomatic and error prone.

_Regardless I would stick to simpler solution either commit 1 or 2._:

In general go routines are cheap so worker pools shouldn't by typical in go? You have counting semaphore pattern and or buffered channels instead to limit concurrency. Worker pool internally also uses `apache.channels` which are [deprecated](https://github.com/oasisprotocol/oasis-core/issues/6315), and also suffers from unbounded submit issue. Given that we only use it in two places, am not sure this package is actually needed long-term thus the preference for the simpler solutions.

go/worker/storage/committee/worker.go
```diff
@@ -762,6 +762,14 @@ func (w *Worker) Serve(ctx context.Context) error { // nolint: gocyclo
 	w.status = api.StatusStarting
 	w.statusLock.Unlock()
 
+	var fetchPool *workerpool.Pool
```
peternose: These changes don't change anything, or am I missing something? And why is this change better? You just introduced that the pool can be `nil`, which in previous case was not possible.

go/common/workerpool/workerpool.go
```diff
@@ -81,7 +81,11 @@ func (p *Pool) Stop() {
 		close(p.stopCh)
 	})
 
-	for range p.jobCh.Out() { //nolint:revive
+	for item := range p.jobCh.Out() { //nolint:revive
+		job := item.(*jobDescriptor)
+		if job.completeCh != nil {
+			close(job.completeCh)
```
peternose: Now, you are using `completeCh` for 2 things: job completion and job cancellation, and users cannot distinguish them.

martintomazic: Agree. First commit technically mitigates the issue. The correct solution imo would be to fix fetch pool as [suggested](https://github.com/oasisprotocol/oasis-core/pull/6444#discussion_r2702789448), or even better refactor this worker and get rid of the worker pool all-together. 

Given priorities, I would like to take the simplest solution that fixes the issue. So first commit only.

go/worker/storage/committee/worker.go
martintomazic: Previously `defer fetchPool.Stop` happened prior to `defer wg.Wait`. As fetch pool stops all the go routines without closing `doneCh`, `wg.Wait` may now be stuck at waiting for the [following](https://github.com/oasisprotocol/oasis-core/blob/4aca0773e2257c619233f3dfc649faf73168b39a/go/worker/storage/committee/worker.go#L1092) go routines. 

By switching defer statement order I fix this issue of stuck teardwon. I introduced nil pool (which I don't like), because maybe the worker would error prior to even starting fetching stuff so I want to initialize it once we actually know we are ready to start fetching.

peternose: Thank you for explanation. Now I understand the problem. Unfortunately, your solution is not the best as no one will figure out that `pool.Stop` needs to happen after `wg.Wait`, so we can expect this bug also in the future.

A better solution would be to force the pool to wait for job completion, and not the wait group. This way the order of `wg.Wait` and `pool.Stop` doesn't matter.
- Revert code changes regarding `fetchPool`.
- Change round fetches to:
```go
					fetchPool.Submit(func() {
						w.fetchDiff(ctx, this.Round, prevRoots[i], this.Roots[i])
					})
					// wg.Go(func() {
					// 	<-doneCh
					// })
```
- Add `p.wg.Wait()` to the end of `pool.Stop`, or introduce `Wait`, and defer call it after `Stop`.

You still don't have the ability to stop running jobs when the pool is stopped, but as you mentioned, this can only be achieved if the pools accepts jobs as `func(ctx context.Context)`.

go/common/workerpool/workerpool.go
peternose: I agree with everything except with parameter `jobCtx context.Context`. However, as you concluded the `pool` is not ideal solution and should be avoided, eventually its usage refactored and in the end removed. So I would treat it as deprecated and avoid any improvements.

peternose:
> Start your node with paratime configured and return a dummy error [here](https://github.com/oasisprotocol/oasis-core/blob/4aca0773e2257c619233f3dfc649faf73168b39a/go/worker/storage/committee/worker.go#L1280).

Yes, any error returned in the last `for` loop can cause this problem.

go/worker/storage/committee/worker.go
martintomazic: We complicated our life, `pool.Stop`  already ensures graceful shutdown (see the commit message).

martintomazic: This still won't work as `fetchDiff`'s:
```go
	defer func() {
		select {
		case w.diffCh <- result:
		case <-ctx.Done():
		}
	}()
```

would block as `ctx` won't be canceled, and the `w.diffCh` would no longer accept as we exited for loop already.

To make it work we either:
1. Swap the order as temp solution.
2. New fetchpool context and cancel prior to `pool.Stop`
    ``` go 
    fetchPool := workerpool.New("storage_fetch/" + w.commonNode.Runtime.ID().String())
	fetchPool.Resize(config.GlobalConfig.Storage.FetcherCount)
	defer fetchPool.Stop()
	poolCtx, cancel := context.WithCancel(ctx)
	defer cancel()

    .....  

           	_ = fetchPool.Submit(func() {
		          w.fetchDiff(poolCtx, this.Round, prevRoots[i], this.Roots[i])
		     })
    ```
3. Change `Submit` signature.
4. Refactor this worker so that all this is irrelavant.

update: I have slight preference for **2.**

martintomazic: I implemented **2.**. **1.** with good comment is also fine given this is temp solution.

#6451 (**4.**) shows why **3.** is irrelavant.

peternose (approved):

PR #6449, opened 2026-01-26
go/oasis-node: Bump default pruning config to 250_000 versions

martintomazic:
Let's use a sane defaults for the pruning config ([context](https://github.com/oasisprotocol/docs/pull/1526#discussion_r2727234906)).

Update defaults to 250_000 versions, so that node has enough consensus and runtime state to create at least two chekpoints if the checkpoint interval is 100_000. At the same time this is still far from the number of versions, that makes Sapphire pruning prohibitively slow.

## Follow-ups
* In the future to make code more robust we might consider prune handlers that  prevent pruning past pending checkpoint creation. (out of scope).
* Enable checkpoint creation by default to ensure long term data availability of the checkpoints.
* Probably we should enable pruning by default.

### DIFF
--- a/go/consensus/cometbft/config/config.go
+++ b/go/consensus/cometbft/config/config.go
@@ -235,7 +235,7 @@ func DefaultConfig() Config {
 		UpgradeStopDelay: time.Minute,
 		Prune: PruneConfig{
 			Strategy: PruneStrategyNone,
-			NumKept:  3600,
+			NumKept:  250_000,
 			Interval: 2 * time.Minute,
 		},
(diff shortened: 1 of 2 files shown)


martintomazic:
Bumping default configs might break configs that rely on the default values, so should the changelog be breaking? In practice the default value of `600` was already not practical.

martintomazic:
CI was [green](https://buildkite.com/oasisprotocol/oasis-core-ci/builds/16446) meaning no e2e test relied on this defaults. :)

Issue #6419, opened 2025-12-02
docs: Document sync modes


martintomazic:
Add documentation about different syncing modes, syncing stages and corresponding databases.

Include ETA and which stage is blocked on which. Finally, describe how to monitor the progress of each stage using node's status.

## Motivation

Having common terminology and overview of the syncing stages would significantly ease the node troubleshooting with e.g. node operators, devops etc. 

E.g. troubleshooting and planning of https://github.com/oasisprotocol/internal-ops/issues/1221 would be much easier using this documentation.

## Proposed direction

**Syncing modes**
* _Block sync_ for consensus.
* _Diff sync_ for runtimes.
* _State sync / Checkpoint sync (?)_ for both consensus and runtimes:
    * Document checkpoints as independent concept, so that we can reference them elsewhere.
    * Genesis file is technically a checkpoint as well.

**Stages**:
* _Consensus state sync_.
* _History (re)-index_ (in parallel to consensus state sync).
* _Runtime state sync_ (blocked by consensus state sync and or history (re)-index).
 
**Databases used**:
 * _State DB_ (`NodeDB`) for both consensus and runtimes.
 *  Consensus specific DBs:  _Blockstore_, evidence, state...
 * _Runtime (light) history_

Bonus: explain the importance of having no gaps between latest round found inside _Runtime State DB_ and earliest round inside _Runtime Light History_ - relates to https://github.com/oasisprotocol/oasis-core/issues/6403.

## End goal 
Consistent terminology, also reflected in our code, metrics, status.

## Challenge
Find good names that are not far off from existing terminology used in code, documentation, metrics and status output.

PR #6452, opened 2026-02-05
go/oasis-node/storage: Add new inspect command

### LINKED ISSUE #6427: go/oasis-node: Add new storage inspect command
Add new `oasis-node storage inspect` command that displays last retained and latest version for all databases. 

E.g.:
```bash
Consensus:
    State DB:
        Latest height: [uin64]
        Last retained height: [uin64]
    Block store/history: 
        Latest height: [uin64]
        Last retained height: [uin64]
Runtimes:
    000000000000000000000000000000000000000000000000f80306c9858e7279:
        State DB:
            Latest round: [uin64]
            Last retained round: [uin64]
        Light History:
             Latest round: [uin64]
             Last retained round: [uin64]         
```

## Motivation

Currently to obtain this you need to start the node. 

There are cases where you want you data to be static, moreover it takes time for node to actually sync and display this info. Finally, for such simple query you don't want to possibly register and or start serving p2p to then stop the node.

Use cases:
* Sanity checking state of the node DBs and _troubleshooting_ (!) corrupted state.
* Useful for creating snapshots with exact start version:
    * Inspect consensus DB offline, calculate number of versions to keep,  followed by offline pruning.
    * Relevant for #6423 and https://github.com/oasisprotocol/internal-ops/issues/1221.
* Sanity checking downloaded snapshots, prior to starting the node.

## Implementation space
Add to `go/oasis-node/cmd/storage`. 

Should be "trivial" as we already have helpers there for manually opening databases (with exact same config the actual node has).

martintomazic:
Closes #6427 

## Example output

```bash

oasis-node storage inspect --config /path/to/config

....
logs
....

Consensus:
    State DB:
        Ok:  true
        Latest height:  28491162
        Last retained height:  25467042
    Block history:
        Ok:  true
        Latest height:  28491162
        Last retained height:  25467042
Runtimes:
     000000000000000000000000000000000000000000000000e2eaa99fc008f87f
        State DB:
            Ok:  true
            Latest round:  12308254
            Last retained round:  11616199
        Light History:
            Ok:  true
            Latest round:  19489969
            Last retained round:  12066873
     000000000000000000000000000000000000000000000000f80306c9858e7279
        State DB:
            Ok:  true
            Latest round:  4754365
            Last retained round:  4741035
        Light History:
            Ok:  true
            Latest round:  12554133
            Last retained round:  4741166
     000000000000000000000000000000000000000000000000e199119c992377cb
        State DB:
            Ok:  true
            Latest round:  11552194
            Last retained round:  5070614
        Light History:
            Ok:  true
            Latest round:  11552194
            Last retained round:  5070614
```

## How to test:
Check if it works as expected.  If databases are missing it should not return eagerly, json etc.

Writing dedicated e2e test is probably not worth it given this is auxiliary command that does not affect existing consensus code at all?

## Considerations:
1. We should log to `Stderr` so that logs can be redirected separately from the output (e.g. if you output json).
     * Any objections to changing [defaults](https://github.com/oasisprotocol/oasis-core/blob/3182a76ebb128fa8cf6283988f9e1dd282671b8e/go/oasis-node/cmd/common/logging.go#L39)? This would be also helpful for other commands.
2. We need to align on good user facing names, including capitalization. Overlaps with https://github.com/oasisprotocol/oasis-core/issues/6419.:
    * StateDB or NodeDB or MKVSDB or mkvsDB
    * Blockstore, Block store, Block History, block history?
    * Runtime light history?  existing "history" is to opaque...
3. I have omitted cometbft evidence and state store intentionally in the status.

### DIFF
--- a/go/oasis-node/cmd/storage/storage.go
+++ b/go/oasis-node/cmd/storage/storage.go
@@ -176,9 +176,8 @@ func parseRuntimes(args []string) ([]common.Namespace, error) {
 	return runtimes, nil
 }
 
-func doMigrate(_ *cobra.Command, args []string) error {
+func doMigrate(cmd *cobra.Command, args []string) error {
 	dataDir := cmdCommon.DataDir()
-	ctx := context.Background()
 
 	runtimes, err := parseRuntimes(args)
 	cobra.CheckErr(err)
@@ -201,7 +200,7 @@ func doMigrate(_ *cobra.Command, args []string) error {
 			}
 
 			helper := &migrateHelper{
-				ctx:     ctx,
+				ctx:     cmd.Context(),
 				history: history,
 				roots:   map[hash.Hash]node.RootType{},
 			}
@@ -224,9 +223,8 @@ func doMigrate(_ *cobra.Command, args []string) error {
 	return nil
 }
 
-func doCheck(_ *cobra.Command, args []string) error {
+func doCheck(cmd *cobra.Command, args []string) error {
 	dataDir := cmdCommon.DataDir()
-	ctx := context.Background()
 
 	runtimes, err := parseRuntimes(args)
 	cobra.CheckErr(err)
@@ -245,7 +243,7 @@ func doCheck(_ *cobra.Command, args []string) error {
 
 			display := &displayHelper{}
 
-			err := badger.CheckSanity(ctx, nodeCfg, display)
+			err := badger.CheckSanity(cmd.Context(), nodeCfg, display)
 			if err != nil {
 				return fmt.Errorf("node database checker returned error: %w", err)
 			}


martintomazic:
Minor: `IsNodeRunning` check was off. Fixed now. 

**Update**:
* `openRuntimeStateDB` was not working with storage set to `auto`. The correct way is to use `storage.NewLocalBackend` similar to what we already do for consensus.
* Also added a documentation to `cli.md`.

go/oasis-node/cmd/storage/inspect.go
```diff
@@ -0,0 +1,201 @@
...
+				fmt.Println(string(prettyStatus))
+				return nil
+			case "text", "":
+				// Fall through to text output.
+			default:
+				return fmt.Errorf("unsupported output format: %s (supported: text, json)", outputFormat)
+			}
+
+			fmt.Println("Consensus:")
+			fmt.Println("    State DB:")
```
matevz: We use two-spaces indentation in the Oasis CLI. Perhaps we should make this consistent.

```suggestion
			fmt.Println("  State DB:")
```

go/oasis-node/cmd/storage/inspect.go
```diff
@@ -0,0 +1,201 @@
...
+			fmt.Println("        Ok: ", status.Consensus.StateDB.Ok)
+			fmt.Println("        Latest height: ", status.Consensus.StateDB.LatestVersion)
```
matevz: Should we alight numbers on top of each other?

```suggestion
			fmt.Println("        Latest height:        ", status.Consensus.StateDB.LatestVersion)
```

go/oasis-node/cmd/storage/inspect.go
```diff
@@ -0,0 +1,201 @@
...
+type DBStatus struct {
+	// Ok is true when the db exists and there was no error inspecting it.
+	Ok bool `json:"ok"`
+	// LatestVersion is the version of the most recent item stored in the corresponding db.
+	LatestVersion uint64 `json:"latest_version"`
+	// LastRetainedVersion is the version of the oldest item stored in the corresponding db.
+	LastRetainedVersion uint64 `json:"last_retained_version"`
+}
+
+// Inspect inpects node's databases and returns a corresponding storage Status.
```
peternose:
```suggestion
// Inspect inspects node's databases and returns a corresponding storage Status.
```

go/oasis-node/cmd/storage/inspect.go
```diff
@@ -0,0 +1,201 @@
...
+
+	return status, nil
+}
+
+func newInspectCmd() *cobra.Command {
+	var outputFormat string
+
+	cmd := &cobra.Command{
+		Use:   "inspect",
+		Short: "inpect storage",
```
peternose:
```suggestion
		Short: "inspect storage",
```

go/oasis-node/cmd/storage/inspect.go
```diff
@@ -0,0 +1,201 @@
...
+			rtStatus.LightHistory.Ok = true
+		}()
+
+		status.Runtimes[rt] = rtStatus
+	}
+
+	return status, nil
+}
+
+func newInspectCmd() *cobra.Command {
```
peternose, line 122: Why is this not in the same format as other commands (e.g. `doInspect`)?

go/oasis-node/cmd/storage/inspect.go
```diff
@@ -0,0 +1,201 @@
...
+			}
+
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			runtimes, err := registry.GetConfiguredRuntimeIDs()
+			if err != nil {
+				return fmt.Errorf("failed to get configured runtimes: %w", err)
+			}
+			status, err := Inspect(cmd.Context(), cmdCommon.DataDir(), runtimes)
```
peternose: I would replace in a separate commit `background.Context` with `cmd.Context` in other commands, to be consistent.

go/oasis-node/cmd/storage/inspect.go
```diff
@@ -0,0 +1,201 @@
...
+			fmt.Println("    State DB:")
+			fmt.Println("        Ok: ", status.Consensus.StateDB.Ok)
+			fmt.Println("        Latest height: ", status.Consensus.StateDB.LatestVersion)
+			fmt.Println("        Last retained height: ", status.Consensus.StateDB.LastRetainedVersion)
+			fmt.Println("    Block history:")
+			fmt.Println("        Ok: ", status.Consensus.BlockHistory.Ok)
+			fmt.Println("        Latest height: ", status.Consensus.BlockHistory.LatestVersion)
+			fmt.Println("        Last retained height: ", status.Consensus.BlockHistory.LastRetainedVersion)
+
+			if len(status.Runtimes) <= 0 {
```
peternose:
```suggestion
			if len(status.Runtimes) == 0 {
```

go/oasis-node/cmd/storage/inspect.go
```diff
@@ -0,0 +1,201 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/oasisprotocol/oasis-core/go/common"
+	cmdCommon "github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common"
+	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
+	"github.com/oasisprotocol/oasis-core/go/runtime/registry"
+)
+
+// Status contains results of inspecting all of the node's databases.
+type Status struct {
```
peternose, line 16: Maybe this name is too general.

go/oasis-node/cmd/storage/inspect.go
```diff
@@ -0,0 +1,201 @@
...
+	// LatestVersion is the version of the most recent item stored in the corresponding db.
+	LatestVersion uint64 `json:"latest_version"`
+	// LastRetainedVersion is the version of the oldest item stored in the corresponding db.
+	LastRetainedVersion uint64 `json:"last_retained_version"`
+}
+
+// Inspect inpects node's databases and returns a corresponding storage Status.
+//
+// The node should not be running once this command is called.
+func Inspect(ctx context.Context, dataDir string, runtimes []common.Namespace) (Status, error) {
```
peternose: I think that this method never returns an error 🤔

go/oasis-node/cmd/storage/inspect.go
martintomazic: Calling with package `storage.Status` is probably not that bad? I thought a while about smtg better, but couldn't find it especially since you have nested structs with similar names then. Any suggestion?

go/oasis-node/cmd/storage/inspect.go
martintomazic: Existing pattern is not optimal since it uses global variables and init functions. Moreover the scope of the variables is too big, variables are hard to navigate and overall the code is less testable.

E.g. see  [Github cli](https://github.com/cli/cli/blob/4e656374719659ab60d096afec12afac9acab463/pkg/cmd/status/status.go#L41) for the same pattern. Also see how nice [create/import checkpoint](https://github.com/oasisprotocol/oasis-core/blob/ac8bfb5e815fcd6e16f21f2094bd93b9963f275d/go/oasis-node/cmd/storage/checkpoint.go#L29:L167) definitions are.

_That said it breaks consistency._ 

My personal stance is that if it improves the overall quality and is a better pattern we should start using it and gradually get rid of the old one. Fixing all commands to new format is out of scope. 

Happy to revert to existing pattern also.

go/oasis-node/cmd/storage/inspect.go
martintomazic: I find it easier to read if it is next to text?

go/oasis-node/cmd/storage/inspect.go
peternose: No suggestion, just a remark that `Status` is too general as this struct contains only db status. You can leave it, and we change latter if we add another status.

PR #6454, opened 2026-02-07
go/oasis-node/cmd/storage: Add create and import checkpoint cmd

### LINKED ISSUE #6423: go/oasis-node: Enable snapshot creation with exact start version
## Motivation
Storing all historical data for all paratimes from genesis on has exceeded `2TB`, hence we are moving towards incremental snapshot creation.

Creating a snapshot with exact start day is currently practically impossible (e.g. [see](https://github.com/oasisprotocol/internal-ops/issues/1221)). Moreover, it takes too much time.

## Additional Challenges
1. As the network has no intermediate state checkpoints, we are forced into syncing all the data from the genesis onward which can easily take 2 weeks.
2. Runtime state pruning becomes impractically slow as state grows: https://github.com/oasisprotocol/oasis-core/issues/6334

### Add new create/import checkpoint commands
Add new  `oasis-node storage create-checkpoint --version:[uint64] --runtime[hex id]`, that creates checkpoints for all storage roots of the specified _State DB_. If `--runtime` flag is omitted than command creates a checkpoint for the consensus _State DB_. 

To import the checkpoint we could use symmetric `oasis-node storage import-checkpoint` 

#### Pros
* Trivial to implement, as we already have all building blocks (see `checkpoint` package).
* Creating a snapshot will no longer require genesis sync, but instead you will start _Consensus Block Sync_ and later followed with _Runtime Diff Sync_ from the exact checkpoint height (!).

#### Cons
* You need access to a node that has at least as old state as your snapshot start date. Then you have to create a checkpoint on the remote node, copy it to the new node, and finally import it.

#### Bonus
* Since checkpoints are currently < `10Gb`, we could in addition to providing snapshots offer historical checkpoints (e.g. every half a year). 
* Somehow related to history expiry: https://github.com/oasisprotocol/oasis-core/issues/6422.

### Alternative `keep_from` pruning configuration.

> As suggested by @peternose, probably the cleanest way would be to add keep_from strategy to the pruning config.
>
>This way you could specify height/round from which the data should be preserved.
>Assuming you would have pruning enabled from the genesis onward this would work out of the box for you, as internally we could implement very aggressive pruning until this height/round - [thread](https://github.com/oasisprotocol/internal-ops/issues/1221#issuecomment-3603154634).

#### Pros
* Relatively easy to implement as this is just another pruner kind for consensus and runtime.

#### Cons
* You will still have to sync from the genesis onwards.
* Need to extend config, increasing LOC of main oasis logic.

We could also implement both.

martintomazic:
Closes #6423 

- [x] Write our own version of  [BootstrapState](https://github.com/oasisprotocol/cometbft/blob/08e22df73d354512fc27bd0c5731b3dcf1f8fef7/node/node.go#L198)
    * The node creating checkpoint can dump untrusted metadata, that will be used for initializing cometbft stores after importing the checkpoint.
    * We could also make it trustless, if we want to get rid of the snapshots entirely and only store checkpoints + storage diffs  and blocks. :)  
- [x] Fix TODOs depending on the review feedback.

### DIFF
--- a/go/oasis-node/cmd/storage/checkpoint.go
+++ b/go/oasis-node/cmd/storage/checkpoint.go
@@ -1,0 +1,623 @@
+package storage
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"runtime"
+
+	cmtCfg "github.com/cometbft/cometbft/config"
+	cmtProtoState "github.com/cometbft/cometbft/proto/tendermint/state"
+	cmtstore "github.com/cometbft/cometbft/proto/tendermint/store"
+	cmtProto "github.com/cometbft/cometbft/proto/tendermint/types"
+	cmtState "github.com/cometbft/cometbft/state"
+	"github.com/cometbft/cometbft/store"
+	cmttypes "github.com/cometbft/cometbft/types"
+	"github.com/cometbft/cometbft/version"
+	"github.com/cosmos/gogoproto/proto"
+	"github.com/spf13/cobra"
+
+	"github.com/oasisprotocol/oasis-core/go/common"
+	"github.com/oasisprotocol/oasis-core/go/common/cbor"
+	cmdCommon "github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/checkpoint"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/db/api"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/node"
+)
+
+const (
+	consensusSubdir = "consensus"
+	runtimesSubdir  = "runtimes"
+
+	consensusMetaFilename = "bootstrap.cbor"
+)
+
+func newCheckpointCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "checkpoint",
+		Short: "create and import storage checkpoints",
+		PersistentPreRunE: func(_ *cobra.Command, args []string) error {
+			if err := cmdCommon.Init(); err != nil {
+				cmdCommon.EarlyLogAndExit(err)
+			}
+			running, err := cmdCommon.IsNodeRunning()
+			if err != nil {
+				return fmt.Errorf("failed to ensure the node is not running: %w", err)
+			}
+			if running {
+				return fmt.Errorf("checkpoint operations can only be done when the node is not running")
+			}
+			return nil
+		},
+	}
+
+	cmd.AddCommand(newCreateCmd())
+	cmd.AddCommand(newImportCmd())
+
+	return cmd
+}
+
+func newCreateCmd() *cobra.Command {
+	var (
+		height    uint64
+		runtimeID string
+		round     uint64
+		outDir    string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "create",
+		Args:  cobra.NoArgs,
+		Short: "create storage checkpoints",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+			dataDir := cmdCommon.DataDir()
+
+			// Consensus checkpoint:
+			createConsensusCp := func(outputDir string) error {
+				ndb, close, err := openConsensusNodeDB(dataDir)
+				if err != nil {
+					return fmt.Errorf("failed to open consensus state DB: %w", err)
+				}
+				defer close()
+				return createCheckpoints(ctx, ndb, common.Namespace{}, height, outputDir)
+			}
+			if height != 0 {
+				consensusOutDir := filepath.Join(outDir, consensusSubdir)
+				logger.Info("creating consensus checkpoint",
+					"height", height,
+					"output_dir", consensusOutDir,
+				)
+				if err := createConsensusCp(consensusOutDir); err != nil {
+					return fmt.Errorf("failed to create consensus checkpoint (height: %d): %w", height, err)
+				}
+				logger.Info("consensus checkpoint created")
+
+				if err := createCometBFTBootstrapMeta(dataDir, height, consensusOutDir); err != nil {
+					return fmt.Errorf("failed to write bootstrap metadata: %w", err)
+				}
+				logger.Info("bootstrap metadata created")
+			}
+
+			// Runtime Checkpoints:
+			createRuntimeCps := func(ns common.Namespace, outputDir string) error {
+				ndb, err := openRuntimeStateDB(dataDir, ns)
+				if err != nil {
+					return fmt.Errorf("failed to open runtime state DB: %w", err)
+				}
+				defer ndb.Close()
+				return createCheckpoints(ctx, ndb, ns, round, outputDir)
+			}
+			if runtimeID != "" {
+				var ns common.Namespace
+				if err := ns.UnmarshalHex(runtimeID); err != nil {
+					return fmt.Errorf("malformed source runtime ID: %q: %w", runtimeID, err)
+				}
+				rtOutDir := filepath.Join(outDir, runtimesSubdir, ns.Hex())
+				logger.Info("creating runtime checkpoints (may take minutes to hours depending on the db size)",
+					"round", round,
+					"output_dir", rtOutDir,
+					"runtime_id", ns,
+				)
+				if err := createRuntimeCps(ns, rtOutDir); err != nil {
+					return fmt.Errorf("failed to create runtime checkpoints (runtime: %s, round: %d): %w", runtimeID, round, err)
+				}
+				logger.Info("runtime checkpoints created", "runtime_id", ns)
+			}
+
+			return nil
+		},
+	}
+
+	cmd.Flags().Uint64Var(&height, "height", 0, "consensus height")
+	cmd.Flags().StringVar(&runtimeID, "runtime", "", "hex encoded runtime ID")
+	cmd.Flags().Uint64Var(&round, "round", 0, "round for which checkpoints will be created")
+	cmd.Flags().StringVar(&outDir, "output-dir", "", "output directory")
+
+	return cmd
+}
+
+func newImportCmd() *cobra.Command {
+	var inputDir string
+
+	cmd := &cobra.Command{
+		Use:   "import",
+		Args:  cobra.NoArgs,
+		Short: "import storage checkpoints",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+			dataDir := cmdCommon.DataDir()
+
+			// Consensus checkpoint:
+			importConsensusCp := func(inputDir string) error {
+				ndb, close, err := openConsensusNodeDB(dataDir)
+				if err != nil {
+					return fmt.Errorf("failed to open consensus state DB: %w", err)
+				}
+				defer close()
+
+				return importCheckpoints(ctx, ndb, common.Namespace{}, inputDir)
+			}
+			consensusInputDir := filepath.Join(inputDir, consensusSubdir)
+			_, err := os.ReadDir(consensusInputDir)
+			switch {
+			case err == nil:
+				logger.Info("importing consensus checkpoint", "input_dir", inputDir)
+				if err := importConsensusCp(consensusInputDir); err != nil {
+					return fmt.Errorf("failed to import consensus checkpoint (input dir: %s): %w", consensusInputDir, err)
+				}
+				logger.Info("consensus checkpoint imported")
+
+				meta, err := readCometBFTBootstrapMeta(consensusInputDir)
+				if err != nil {
+					return fmt.Errorf("failed to read bootstrap metadata: %w", err)
+				}
+				if err := bootstrapTrustedState(dataDir, meta); err != nil {
+					return fmt.Errorf("failed to bootstrap CometBFT trusted state: %w", err)
+				}
+				logger.Info("bootstrap metadata imported")
+			case os.IsNotExist(err):
+				// No consensus checkpoints to import
+			default:
+				return fmt.Errorf("failed to stat: %w", err)
+			}
+
+			// Runtime checkpoints:
+			importRuntimeCp := func(inputDir string, ns common.Namespace) error {
+				ndb, err := openRuntimeStateDB(dataDir, ns)
+				if err != nil {
+					return err
+				}
+				defer ndb.Close()
+
+				return importCheckpoints(ctx, ndb, ns, inputDir)
+			}
+			runtimeInputDir := filepath.Join(inputDir, runtimesSubdir)
+			entries, err := os.ReadDir(runtimeInputDir)
+			switch {
+			case err == nil:
+				for _, entry := range entries {
+					var ns common.Namespace
+					if err := ns.UnmarshalHex(entry.Name()); err != nil {
+						return fmt.Errorf("malformed source runtime ID: %s: %w", entry.Name(), err)
+					}
+					rtCpsDir := filepath.Join(runtimeInputDir, entry.Name())
+
+					logger.Info("importing runtime checkpoints (may take few minutes)",
+						"runtime_id", ns,
+						"input_dir", inputDir,
+					)
+					if err := importRuntimeCp(rtCpsDir, ns); err != nil {
+						return fmt.Errorf("failed to import checkpoints (checkpoint dir: %s): %w", rtCpsDir, err)
+					}
+					logger.Info("runtime checkpoints imported", "runtime_id", ns)
+				}
+			case os.IsNotExist(err):
+				// No runtime checkpoints to import
+			default:
+				return fmt.Errorf("failed to stat: %w", err)
+			}
+
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&inputDir, "input-dir", "", "directory with checkpoints to import")
+
+	return cmd
+}
+
+func createCheckpoints(ctx context.Context, ndb api.NodeDB, ns common.Namespace, version uint64, outputDir string) error {
+	if err := ensureEmptyDir(outputDir); err != nil {
+		return err
+	}
+
+	latest, ok := ndb.GetLatestVersion()
+	if !ok {
+		return fmt.Errorf("empty state DB")
+	}
+	earliest := ndb.GetEarliestVersion()
+	if version < earliest || version > latest {
+		return fmt.Errorf("version not found (available range: %d-%d)", earliest, latest)
+	}
+
+	roots, err := ndb.GetRootsForVersion(version)
+	if err != nil {
+		return fmt.Errorf("failed to get roots %w", err)
+	}
+	if len(roots) == 0 {
+		// Empty roots are implicit in NodeDB and therefore not returned by GetRootsForVersion.
+		// If at this height state DB has an empty state, create a checkpoint for the empty state root,
+		// so that importing it will still finalize this version (making NodeDB non-empty).
+		//
+		// In case of state DB with multiple empty roots (e.g. state and IO root both empty)
+		// this will create only one empty checkpoint (duplicate hash), which is safe as we only need
+		// it to finalize version with implicitly empty state and IO root.
+		stateRoot := emptyRoot(ns, version, node.RootTypeState)
+		roots = []node.Root{stateRoot}
+	}
+
+	creator, err := checkpoint.NewFileCreator(outputDir, ndb)
+	if err != nil {
+		return fmt.Errorf("failed to create file creator: %w", err)
+	}
+
+	const chunkSize uint64 = 8 * 1024 * 1024 // 8MiB chunks
+	// Use parallel chunking algorithm, but limit concurrency.
+	chunkerThreads := uint16(runtime.GOMAXPROCS(0) / 2)
+	for _, root := range roots {
+		_, err := creator.CreateCheckpoint(ctx, root, chunkSize, chunkerThreads)
+		if err != nil {
+			return fmt.Errorf("failed to create checkpoint (rootType: %s): %w", root.Type, err)
+		}
+	}
+
+	return nil
+}
+
+func importCheckpoints(ctx context.Context, ndb api.NodeDB, ns common.Namespace, inputDir string) error {
+	isEmpty, err := isEmptyDir(inputDir)
+	if err != nil {
+		return fmt.Errorf("failed to inspect input directory: %w", err)
+	}
+	if isEmpty {
+		return fmt.Errorf("input directory is empty: %s", inputDir)
+	}
+	if _, ok := ndb.GetLatestVersion(); ok {
+		return fmt.Errorf("state db not empty")
+	}
+
+	provider, err := checkpoint.NewFileCreator(inputDir, nil) // ndb = nil since we will only import checkpoints.
+	if err != nil {
+		return fmt.Errorf("failed to create checkpoint file creator: %w", err)
+	}
+	cps, err := provider.GetCheckpoints(ctx, &checkpoint.GetCheckpointsRequest{Version: 1, Namespace: ns})
+	if err != nil {
+		return fmt.Errorf("failed to read checkpoints: %w", err)
+	}
+
+	var roots []node.Root
+	for _, cp := range cps {
+		if err := importCp(ctx, ndb, provider, cp); err != nil {
+			return fmt.Errorf("failed to import checkpoint (root hash: %s): %w", cp.Root.Hash, err)
+		}
+		roots = append(roots, cp.Root)
+	}
+	if err := ndb.Finalize(roots); err != nil {
+		return fmt.Errorf("failed to finalize: %w", err)
+	}
+
+	return nil
+}
+
+func importCp(ctx context.Context, ndb api.NodeDB, provider checkpoint.Creator, cp *checkpoint.Metadata) error {
+	if err := ndb.StartMultipartInsert(cp.Root.Version); err != nil {
+		return fmt.Errorf("failed to start multipart insert: %w", err)
+	}
+	defer func() {
+		if err := ndb.AbortMultipartInsert(); err != nil {
+			logger.Error("failed to abort multi-part insert", "err", err)
+		}
+	}()
+
+	for idx := range cp.Chunks {
+		chunk, err := cp.GetChunkMetadata(uint64(idx))
+		if err != nil {
+			return fmt.Errorf("failed to get chunk metadata: %w", err)
+		}
+		var buf bytes.Buffer
+		if err := provider.GetCheckpointChunk(ctx, chunk, &buf); err != nil {
+			return fmt.Errorf("failed to read checkpoint chunk (idx: %d): %w", idx, err)
+		}
+		if err := checkpoint.RestoreChunk(ctx, ndb, chunk, &buf); err != nil {
+			return fmt.Errorf("failed to restore chunk: %w", err)
+		}
+	}
+	return nil
+}
+
+type bootstrapMeta struct {
+	State  []byte `json:"state"`
+	Commit []byte `json:"commit"`
+}
+
+func createCometBFTBootstrapMeta(dataDir string, height uint64, outputDir string) error {
+	stateStore, err := openConsensusStatestore(dataDir)
+	if err != nil {
+		return fmt.Errorf("failed to open cometbft state store: %w", err)
+	}
+	defer stateStore.Close()
+
+	blockStore, err := openConsensusBlockstore(dataDir)
+	if err != nil {
+		return fmt.Errorf("failed to open consensus blockstore: %w", err)
+	}
+	defer blockStore.Close()
+
+	state, err := getState(height, stateStore, blockStore)
+	if err != nil {
+		return fmt.Errorf("failed to load consensus state at height %d: %w", height, err)
+	}
+	statePB, err := state.ToProto()
+	if err != nil {
+		return fmt.Errorf("failed to convert consensus state to proto: %w", err)
+	}
+	stateBytes, err := proto.Marshal(statePB)
+	if err != nil {
+		return fmt.Errorf("failed to marshal consensus state: %w", err)
+	}
+
+	commit, err := getCommit(blockStore, height)
+	if err != nil {
+		return fmt.Errorf("failed to load consensus commit at height %d: %w", height, err)
+	}
+	commitBytes, err := proto.Marshal(commit.ToProto())
+	if err != nil {
+		return fmt.Errorf("failed to marshal consensus commit: %w", err)
+	}
+
+	meta := bootstrapMeta{
+		State:  stateBytes,
+		Commit: commitBytes,
+	}
+	if err := os.WriteFile(filepath.Join(outputDir, consensusMetaFilename), cbor.Marshal(meta), 0o600); err != nil {
+		return fmt.Errorf("failed to write bootstrap metadata: %w", err)
+	}
+
+	return nil
+}
+
+func readCometBFTBootstrapMeta(inputDir string) (bootstrapMeta, error) {
+	data, err := os.ReadFile(filepath.Join(inputDir, consensusMetaFilename))
+	if err != nil {
+		return bootstrapMeta{}, err
+	}
+
+	var meta bootstrapMeta
+	if err := cbor.Unmarshal(data, &meta); err != nil {
+		return bootstrapMeta{}, fmt.Errorf("failed to decode bootstrap metadata: %w", err)
+	}
+
+	return meta, nil
+}
+
+// bootstrapTrustedState synchronizes the cometbft databases after the state sync
+// has been performed offline.
+//
+// It is expected that the block store and state store are empty at the time the
+// function is called.
+//
+// Adapted from https://github.com/oasisprotocol/cometbft/blob/08e22df73d354512fc27bd0c5731b3dcf1f8fef7/node/node.go#L198.
+func bootstrapTrustedState(dataDir string, meta bootstrapMeta) error {
+	stateDB, err := openConsensusStateDB(dataDir)
+	if err != nil {
+		return fmt.Errorf("failed to open cometbft state store: %w", err)
+	}
+	defer stateDB.Close()
+
+	blockStoreDB, err := openConsensusBlockstoreDB(dataDir)
+	if err != nil {
+		return fmt.Errorf("failed to open consensus blockstore: %w", err)
+	}
+	defer blockStoreDB.Close()
+	blockStore := store.NewBlockStore(blockStoreDB)
+
+	if !blockStore.IsEmpty() {
+		return fmt.Errorf("blockstore not empty, trying to initialize non empty state")
+	}
+
+	stateStore := cmtState.NewBootstrapStore(stateDB, cmtState.StoreOptions{
+		DiscardABCIResponses: cmtCfg.DefaultConfig().Storage.DiscardABCIResponses,
+	})
+	defer stateStore.Close()
+
+	state, err := stateStore.Load()
+	if err != nil {
+		return err
+	}
+
+	if !state.IsEmpty() {
+		return fmt.Errorf("state not empty, trying to initialize non empty state")
+	}
+
+	var statePB cmtProtoState.State
+	if err := proto.Unmarshal(meta.State, &statePB); err != nil {
+		return fmt.Errorf("failed to unmarshal consensus state: %w", err)
+	}
+	metaState, err := cmtState.FromProto(&statePB)
+	if err != nil {
+		return fmt.Errorf("failed to parse consensus state: %w", err)
+	}
+
+	var commitPB cmtProto.Commit
+	if err := proto.Unmarshal(meta.Commit, &commitPB); err != nil {
+		return fmt.Errorf("failed to unmarshal consensus commit: %w", err)
+	}
+	commit, err := cmttypes.CommitFromProto(&commitPB)
+	if err != nil {
+		return fmt.Errorf("failed to parse consensus commit: %w", err)
+	}
+
+	if err = stateStore.Bootstrap(*metaState); err != nil {
+		return err
+	}
+
+	err = blockStore.SaveSeenCommit(metaState.LastBlockHeight, commit)
+	if err != nil {
+		return err
+	}
+
+	// SaveSeenCommit (as used by the upstream) does not persist the BlockStoreState (height/base),
+	// so we save it explicitly here. This way the blockstore reports the correct height after bootstrap,
+	// so that status works as expected. If this is not done, this blockstore base is kept as uninitialized,
+	// and is initialized as soon as the first block is fetched. However, we would cut last retained height by 1,
+	// which is the current case for consensus checkpoint sync.
+	store.SaveBlockStoreState(
+		&cmtstore.BlockStoreState{
+			Base:   metaState.LastBlockHeight,
+			Height: metaState.LastBlockHeight,
+		},
+		blockStoreDB,
+	)
+
+	// Once the stores are bootstrapped, we need to set the height at which the node has finished
+	// statesyncing. This will allow the blocksync reactor to fetch blocks at a proper height.
+	// In case this operation fails, it is equivalent to a failure in online state sync where the operator
+	// needs to manually delete the state and blockstores and rerun the bootstrapping process.
+	err = stateStore.SetOfflineStateSyncHeight(metaState.LastBlockHeight)
+	if err != nil {
+		return fmt.Errorf("failed to set synced height: %w", err)
+	}
+
+	return nil
+}
+
+// getCommit is adapted and simplified and mimics StateProvider behaviour used in the upstream BootstrapState.
+func getCommit(blockStore *store.BlockStore, height uint64) (*cmttypes.Commit, error) {
+	commit := blockStore.LoadBlockCommit(int64(height))
+	if commit == nil {
+		return nil, fmt.Errorf("commit not found at height %d", height)
+	}
+	return commit, nil
+}
+
+// getState is adapted and mimics StateProvider behaviour used in the upstream BootstrapState.
+func getState(height uint64, stateStore cmtState.Store, blockStore *store.BlockStore) (cmtState.State, error) {
+	// The snapshot height maps onto the state heights as follows:
+	//
+	// height: last block, i.e. the snapshotted height
+	// height+1: current block, i.e. the first block we'll process after the snapshot
+	// height+2: next block, i.e. the second block after the snapshot
+	//
+	// We need to fetch the NextValidators from height+2 because if the application changed
+	// the validator set at the snapshot height then this only takes effect at height+2.
+	h := int64(height)
+	lastMeta := blockStore.LoadBlockMeta(h)
+	if lastMeta == nil {
+		return cmtState.State{}, fmt.Errorf("block meta not found at height %d", h)
+	}
+	currentMeta := blockStore.LoadBlockMeta(h + 1)
+	if currentMeta == nil {
+		return cmtState.State{}, fmt.Errorf("block meta not found at height %d", h+1)
+	}
+
+	lastVals, err := stateStore.LoadValidators(h)
+	if err != nil {
+		return cmtState.State{}, err
+	}
+	currentVals, err := stateStore.LoadValidators(h + 1)
+	if err != nil {
+		return cmtState.State{}, err
+	}
+	nextVals, err := stateStore.LoadValidators(h + 2)
+	if err != nil {
+		return cmtState.State{}, err
+	}
+
+	consensusParams, err := stateStore.LoadConsensusParams(h + 1)
+	if err != nil {
+		return cmtState.State{}, err
+	}
+
+	storeState, err := stateStore.Load()
+	if err != nil {
+		return cmtState.State{}, err
+	}
+	if storeState.IsEmpty() {
+		return cmtState.State{}, fmt.Errorf("state store is empty")
+	}
+
+	state := cmtState.State{
+		ChainID: storeState.ChainID,
+		Version: cmtProtoState.Version{
+			Consensus: currentMeta.Header.Version,
+			Software:  version.TMCoreSemVer,
+		},
+		InitialHeight: storeState.InitialHeight,
+	}
+	if state.InitialHeight == 0 {
+		state.InitialHeight = 1
+	}
+
+	state.LastBlockHeight = lastMeta.Header.Height
+	state.LastBlockTime = lastMeta.Header.Time
+	state.LastBlockID = lastMeta.BlockID
+	state.AppHash = currentMeta.Header.AppHash
+	state.LastResultsHash = currentMeta.Header.LastResultsHash
+	state.LastValidators = lastVals
+	state.Validators = currentVals
+	state.NextValidators = nextVals
+	state.LastHeightValidatorsChanged = h + 2
+	state.ConsensusParams = consensusParams
+	state.LastHeightConsensusParamsChanged = currentMeta.Header.Height
+
+	return state, nil
+}
+
+func isEmptyDir(dir string) (bool, error) {
+	// Avoid using os.ReadDir as it reads the whole dir.
+	f, err := os.Open(dir)
+	if err != nil {
+		return false, fmt.Errorf("failed to open directory %q: %w", dir, err)
+	}
+	defer f.Close()
+
+	switch _, err = f.Readdir(1); {
+	case err == nil:
+		return false, nil
+	case errors.Is(err, io.EOF):
+	default:
+		return false, fmt.Errorf("failed to read directory %q: %w", dir, err)
+	}
+	return true, nil
+}
+
+func ensureEmptyDir(dir string) error {
+	if err := os.MkdirAll(dir, 0o700); err != nil {
+		return fmt.Errorf("failed to create output directory: %w", err)
+	}
+
+	isEmpty, err := isEmptyDir(dir)
+	if err != nil {
+		return err
+	}
+	if !isEmpty {
+		return fmt.Errorf("output directory is not empty: %s", dir)
+	}
+
+	return nil
+}
+
+func emptyRoot(ns common.Namespace, version uint64, rootType node.RootType) node.Root {
+	root := node.Root{
+		Namespace: ns,
+		Version:   version,
+		Type:      rootType,
+	}
+	root.Hash.Empty()
+	return root
+}
--- a/go/oasis-node/cmd/storage/checkpoint.go
+++ b/go/oasis-node/cmd/storage/checkpoint.go
@@ -300,6 +300,18 @@ func importCheckpoints(ctx context.Context, ndb api.NodeDB, ns common.Namespace,
 		return fmt.Errorf("failed to read checkpoints: %w", err)
 	}
 
+	if len(cps) == 0 {
+		return fmt.Errorf("no checkpoints found")
+	}
+
+	if err := ndb.StartMultipartInsert(cps[0].Root.Version); err != nil {
+		return fmt.Errorf("failed to start multipart insert: %w", err)
+	}
+	defer func() {
+		if err := ndb.AbortMultipartInsert(); err != nil {
+			logger.Error("failed to abort multi-part insert", "err", err)
+		}
+	}()
 	var roots []node.Root
 	for _, cp := range cps {
 		if err := importCp(ctx, ndb, provider, cp); err != nil {
@@ -315,15 +327,6 @@ func importCheckpoints(ctx context.Context, ndb api.NodeDB, ns common.Namespace,
 }
 
 func importCp(ctx context.Context, ndb api.NodeDB, provider checkpoint.Creator, cp *checkpoint.Metadata) error {
-	if err := ndb.StartMultipartInsert(cp.Root.Version); err != nil {
-		return fmt.Errorf("failed to start multipart insert: %w", err)
-	}
-	defer func() {
-		if err := ndb.AbortMultipartInsert(); err != nil {
-			logger.Error("failed to abort multi-part insert", "err", err)
-		}
-	}()
-
 	for idx := range cp.Chunks {
 		chunk, err := cp.GetChunkMetadata(uint64(idx))
 		if err != nil {
--- a/go/oasis-node/cmd/storage/checkpoint_test.go
+++ b/go/oasis-node/cmd/storage/checkpoint_test.go
@@ -1,0 +1,323 @@
+package storage
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/oasisprotocol/oasis-core/go/common"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs"
+	dbAPI "github.com/oasisprotocol/oasis-core/go/storage/mkvs/db/api"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/db/pathbadger"
+	"github.com/oasisprotocol/oasis-core/go/storage/mkvs/node"
+)
+
+const testVersion uint64 = 10
+
+var testNs common.Namespace = common.NewTestNamespaceFromSeed([]byte("test namespace"), 0)
+
+func TestCreateCheckpoints(t *testing.T) {
+	t.Run("fails on non-empty output dir", func(t *testing.T) {
+		t.Parallel()
+
+		ctx := t.Context()
+		ns := common.Namespace{}
+		ndb, err := newTestNodeDB(t, ns)
+		require.NoError(t, err)
+		defer ndb.Close()
+
+		stateRoot := commitRoot(ctx, t, ndb, ns, testVersion, node.RootTypeState, map[string]string{
+			"key": "value",
+		})
+		require.NoError(t, ndb.Finalize([]node.Root{stateRoot}))
+
+		outDir := filepath.Join(t.TempDir(), "checkpoints")
+		require.NoError(t, os.Mkdir(outDir, 0o700))
+		require.NoError(t, os.WriteFile(filepath.Join(outDir, "existing"), []byte{}, 0o600))
+
+		err = createCheckpoints(ctx, ndb, ns, testVersion, outDir)
+		require.ErrorContains(t, err, "output directory is not empty")
+	})
+
+	t.Run("fails on empty node DB", func(t *testing.T) {
+		t.Parallel()
+
+		ctx := t.Context()
+		ns := common.Namespace{}
+		ndb, err := newTestNodeDB(t, ns)
+		require.NoError(t, err)
+		defer ndb.Close()
+
+		cpDir := filepath.Join(t.TempDir(), "checkpoint")
+		err = createCheckpoints(ctx, ndb, ns, testVersion, cpDir)
+		require.ErrorContains(t, err, "empty state DB")
+	})
+
+	t.Run("fails when version is before earliest", func(t *testing.T) {
+		t.Parallel()
+
+		ctx := t.Context()
+		ns := common.Namespace{}
+		ndb, err := newTestNodeDB(t, ns)
+		require.NoError(t, err)
+		defer ndb.Close()
+
+		stateRoot := commitRoot(ctx, t, ndb, ns, testVersion, node.RootTypeState, map[string]string{
+			"key": "value",
+		})
+		require.NoError(t, ndb.Finalize([]node.Root{stateRoot}))
+
+		cpDir := filepath.Join(t.TempDir(), "checkpoint")
+		err = createCheckpoints(ctx, ndb, ns, testVersion-1, cpDir)
+		require.ErrorContains(t, err, "version not found")
+	})
+
+	t.Run("fails when version is after latest", func(t *testing.T) {
+		t.Parallel()
+
+		ctx := t.Context()
+		ns := common.Namespace{}
+		ndb, err := newTestNodeDB(t, ns)
+		require.NoError(t, err)
+		defer ndb.Close()
+
+		stateRoot := commitRoot(ctx, t, ndb, ns, testVersion, node.RootTypeState, map[string]string{
+			"key": "value",
+		})
+		require.NoError(t, ndb.Finalize([]node.Root{stateRoot}))
+
+		cpDir := filepath.Join(t.TempDir(), "checkpoint")
+		err = createCheckpoints(ctx, ndb, ns, testVersion+1, cpDir)
+		require.ErrorContains(t, err, "version not found")
+	})
+}
+
+func TestImportCheckpoints(t *testing.T) {
+	t.Run("fails on non-empty destination DB", func(t *testing.T) {
+		t.Parallel()
+
+		ctx := t.Context()
+		ns := common.Namespace{}
+
+		srcNDB, err := newTestNodeDB(t, ns)
+		require.NoError(t, err)
+		defer srcNDB.Close()
+
+		stateRoot := commitRoot(ctx, t, srcNDB, ns, testVersion, node.RootTypeState, map[string]string{
+			"key": "value",
+		})
+		require.NoError(t, srcNDB.Finalize([]node.Root{stateRoot}))
+
+		cpDir := filepath.Join(t.TempDir(), "checkpoint")
+		require.NoError(t, createCheckpoints(ctx, srcNDB, ns, testVersion, cpDir))
+
+		dstNDB, err := newTestNodeDB(t, ns)
+		require.NoError(t, err)
+		defer dstNDB.Close()
+
+		// NodeDB allows importing checkpoint even if the target NodeDB is not empty,
+		// as long as the checkpoint is younger than the the last finalized version.
+		// We don't want that, thus finalizing testVersion-1 to ensure createCheckpoints
+		// guard is tested and that the test does not rely on mkvs version already finalized.
+		err = dstNDB.Finalize([]node.Root{emptyRoot(ns, testVersion-1, node.RootTypeState)})
+		require.NoError(t, err)
+
+		err = importCheckpoints(ctx, dstNDB, ns, cpDir)
+		require.ErrorContains(t, err, "state db not empty")
+	})
+
+	t.Run("fails on empty input dir", func(t *testing.T) {
+		t.Parallel()
+
+		ctx := t.Context()
+		ns := common.Namespace{}
+		ndb, err := newTestNodeDB(t, ns)
+		require.NoError(t, err)
+		defer ndb.Close()
+
+		err = importCheckpoints(ctx, ndb, ns, t.TempDir())
+		require.ErrorContains(t, err, "input directory is empty")
+	})
+}
+
+func TestCreateImportRoundtrip(t *testing.T) {
+	t.Run("single non-empty checkpoint", func(t *testing.T) {
+		t.Parallel()
+
+		ctx := t.Context()
+		ns := common.Namespace{}
+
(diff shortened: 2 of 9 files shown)


go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,548 @@
...
+
+func createRuntimeCheckpoints(ctx context.Context, ndb api.NodeDB, round uint64, outputDir string) error {
+	roots, err := ndb.GetRootsForVersion(round)
+	if err != nil {
+		return fmt.Errorf("failed to get roots for round %d: %w", round, err)
+	}
+	if lenRoots := len(roots); 0 == lenRoots || lenRoots > 2 { // Empty IO root is implicitly present.
+		return fmt.Errorf("unexpected number of roots: got %d", lenRoots)
+	}
+	return createCheckpoints(ctx, ndb, roots, outputDir)
```
martintomazic: Runtime light history always return 2 roots (possibly empty), whereas `NodeDB.GetRootForVersion` doesn't have such guarantee. Since checkpoint for empty io root is non-empty, we should handle this corner case here. 

The current solution is not ok. If you only have one root, we need to create another that is empty and pass it to `createCheckpoints` method. I don't want to rely on the runtime light history, since this command should work even if light history is empty, i.e. it should only query state db.

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,548 @@
...
+}
+
+// bootstrapTrustedState synchronizes the cometbft databases after the state sync
+// has been performed offline.
+//
+// It is expected that the block store and state store are empty at the time the
+// function is called.
+//
+// Adapted from https://github.com/oasisprotocol/cometbft/blob/08e22df73d354512fc27bd0c5731b3dcf1f8fef7/node/node.go#L198.
+func bootstrapTrustedState(ctx context.Context, dataDir string, meta bootstrapMeta) error {
```
martintomazic: Might be reasonable to write this function next to `BootstrapState` in the `oasis-cometbft`...

Pros: Logically fits there + less things to import.
Cons: You can change things here, if we move it there you need to bump cometbft which is not that practical.

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,548 @@
...
+	// the validator set at the snapshot height then this only takes effect at height+2.
+	h := int64(height)
+	lastMeta := blockStore.LoadBlockMeta(h)
+	if lastMeta == nil {
+		return cmtState.State{}, fmt.Errorf("block meta not found at height %d", h)
+	}
+	currentMeta := blockStore.LoadBlockMeta(h + 1)
+	if currentMeta == nil {
+		return cmtState.State{}, fmt.Errorf("block meta not found at height %d", h+1)
+	}
```
martintomazic, line 529: This follows upstream where current is `h+1`. In practice (confirmed) if you do state sync, either managed by cometbft or using the new import command, the blockstore will start from `h+1` then, meaning you shadowed start height.

~I am inclined to deviate here and use `h-1`, `h` and `h+1` to work around this.~

martintomazic:
Works!  :) 

The only thing that is impractical is finding corresponding runtime rounds to given consensus height and the fact that bootstrap "eats" one height as [described](https://github.com/oasisprotocol/oasis-core/pull/6454#discussion_r2792168225). 

Finally, one should be very careful with creation/import height/rounds so that you have all relevant light history for the runtime checkpoints you are importing.

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,549 @@
...
+				}
+				defer ndb.Close()
+
+				rtOutDir := filepath.Join(outDir, runtimesSubdir, ns.Hex())
+				return createRuntimeCheckpoints(cmd.Context(), ndb, round, rtOutDir)
+			}
+
+			if height != 0 { // TODO handle zero value vs not set correctly.
+				if err := createConsensusCp(); err != nil {
+					return fmt.Errorf("failed to create consensus checkpoint (height: %d): %w", height, err)
```
martintomazic: Maybe use default undefined round (aka max uint64), alternative is `cmd.Flags().Changed("height")`.

**Update**: Alternative is explicit `--consensus` flag or possible `consensus/runtime` sub-commands. No height/round could also mean latest height. ` -all` flag with `--height` would be also interesting if it would find corresponding runtime rounds for the given height.

martintomazic:
Creating checkpoints from the penultimate [snapshot](https://snapshots.oasis.io/#node/mainnet/20231129-20250715/), is dominated by the Sapphire checkpoint creation.

With 6 chunker threads current projection is 5-7 hours (will update). Import is a matter of minutes.

martintomazic:
Added unit and e2e tests, fixed empty state corner case and improved code quality.

Two minor things left to discuss:
- Exact API of this command / default values actions: see https://github.com/oasisprotocol/oasis-core/pull/6454#discussion_r2793549110
- Consider refactoring checkpoint abstraction as proposed: https://github.com/oasisprotocol/oasis-core/issues/6467 (follow-up).

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+					return fmt.Errorf("failed to open runtime state DB: %w", err)
+				}
+				defer ndb.Close()
+
+				return createCheckpoints(ctx, ndb, ns, round, outputDir)
+			}
+			if runtimeID != "" {
+				var ns common.Namespace
+				if err := ns.UnmarshalHex(runtimeID); err != nil {
+					return fmt.Errorf("malformed source runtime ID: %q: %w", runtimeID, err)
```
peternose, line 118:
```suggestion
					return fmt.Errorf("malformed source runtime ID: %s: %w", runtimeID, err)
```

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+	state.NextValidators = nextVals
+	state.LastHeightValidatorsChanged = nextMeta.Header.Height
+	state.ConsensusParams = consensusParams
+	state.LastHeightConsensusParamsChanged = currentMeta.Header.Height
+
+	return state, nil
+}
+
+func isEmptyDir(dir string) (bool, error) {
+	// Over-engineering?
```
peternose: This comment should be removed if the code is ready for review. Instead, add a brief explanation of why this approach was chosen over `os.ReadDir`.

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+	f, err := os.Open(dir)
+	if err != nil {
+		return false, fmt.Errorf("failed to open directory %q: %w", dir, err)
+	}
+	defer f.Close()
+
+	_, err = f.Readdir(1)
```
peternose:
```suggestion
	switch _, err = f.Readdir(1); {
	case err == nil:
		return false, nil
	case errors.Is(err, io.EOF):
	default:
		return false, fmt.Errorf("failed to read directory %q: %w", dir, err)
	}

	return true, nil
```

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+		return err
+	}
+
+	latest, ok := ndb.GetLatestVersion()
+	if !ok {
+		return fmt.Errorf("empty state DB")
+	}
+	earliest := ndb.GetEarliestVersion()
+	if version < earliest || version > latest {
+		return fmt.Errorf("version not finalized (finalized range: %d-%d)", earliest, latest)
```
peternose:
```suggestion
		return fmt.Errorf("version not found")
```

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+	}
+
+	roots, err := ndb.GetRootsForVersion(version)
+	if err != nil {
+		return fmt.Errorf("failed to get finalized roots %w", err)
```
peternose:
```suggestion
		return fmt.Errorf("failed to get roots %w", err)
```

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+	}
+	if len(roots) == 0 {
+		// Empty roots are implicit in NodeDB and therefore not returned by GetRootsForVersion.
+		// If at this height state DB has an empty state, create a checkpoint for the empty state root,
+		// so that importing it will still finalize this version (making NodeDB non-empty).
+		//
+		// In case of state DB with with multiple empty roots (e.g. state and IO root both empty)
```
peternose:
```suggestion
		// In case of state DB with multiple empty roots (e.g. state and IO root both empty)
```

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+			Version:   version,
+			Type:      node.RootTypeState,
+		}
+		stateRoot.Hash.Empty()
+		roots = []node.Root{stateRoot}
+	}
+
+	creator, err := checkpoint.NewFileCreator(outputDir, ndb)
+	if err != nil {
+		return fmt.Errorf("failed to create checkpoint file creator: %w", err)
```
peternose:
```suggestion
		return fmt.Errorf("failed to create file creator: %w", err)
```

I prefer error `failed to create checkpoint: failed to create file creator: ...`.

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+					return fmt.Errorf("failed to open consensus state DB: %w", err)
+				}
+				defer close()
+
+				return createCheckpoints(ctx, ndb, common.Namespace{}, height, outputDir)
+			}
+			if height != 0 { // TODO handle zero value vs not set correctly.
+				consensusOutDir := filepath.Join(outDir, consensusSubdir)
+				if err := createConsensusCp(consensusOutDir); err != nil {
+					return fmt.Errorf("failed to create consensus checkpoint (height: %d): %w", height, err)
```
peternose, line 96:
```suggestion
					return fmt.Errorf("failed to create consensus checkpoint: %w, err)
```

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+		},
+	}
+
+	cmd.Flags().StringVar(&inputDir, "input-dir", "", "directory with checkpoints to import")
+
+	return cmd
+}
+
+func createCheckpoints(ctx context.Context, ndb api.NodeDB, ns common.Namespace, version uint64, outputDir string) error {
+	if err := ensureEmptyDir(outputDir); err != nil {
```
peternose, line 235: I would expect one could create checkpoints for multiple versions in the same folder.

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+func newCreateCmd() *cobra.Command {
+	var (
+		height    uint64
+		runtimeID string
+		round     uint64
+		outDir    string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "create",
```
peternose, line 73: If I run `oasis-node storage checkpoint create` nothing is printed to the console.

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+			createRuntimeCps := func(ns common.Namespace, outputDir string) error {
+				ndb, err := openRuntimeStateDB(dataDir, ns)
+				if err != nil {
+					return fmt.Errorf("failed to open runtime state DB: %w", err)
+				}
+				defer ndb.Close()
+
+				return createCheckpoints(ctx, ndb, ns, round, outputDir)
+			}
+			if runtimeID != "" {
```
peternose, line 115: What happens if all parameters are set? I guess consensus checkpoint is created and runtime checkpoint creation fails as the directory is not empty. Should we check that only one checkpoint is requested at a time?

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+		roots = []node.Root{stateRoot}
+	}
+
+	creator, err := checkpoint.NewFileCreator(outputDir, ndb)
+	if err != nil {
+		return fmt.Errorf("failed to create checkpoint file creator: %w", err)
+	}
+
+	for _, root := range roots {
+		_, err := creator.CreateCheckpoint(ctx, root, 8*1024*1024, 8) // 8MiB chunks
```
peternose: Could move to constants. Could also make this configurable. The number of threads could also be computed based on the CPU.

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+
+	cmd.Flags().Uint64Var(&height, "height", 0, "consensus height")
+	cmd.Flags().StringVar(&runtimeID, "runtime", "", "hex encoded runtime ID")
+	cmd.Flags().Uint64Var(&round, "round", 0, "round for which checkpoints will be created")
+	cmd.Flags().StringVar(&outDir, "output-dir", "", "output directory")
+
+	return cmd
+}
+
+func newImportCmd() *cobra.Command {
```
peternose, line 144: Why don't you just use `restore` or `import` everywhere?

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+				ndb, close, err := openConsensusNodeDB(dataDir)
+				if err != nil {
+					return fmt.Errorf("failed to open consensus state DB: %w", err)
+				}
+				defer close()
+
+				return restoreCheckpoints(ctx, ndb, common.Namespace{}, inputDir)
+			}
+			consensusInputDir := filepath.Join(inputDir, consensusSubdir)
+			_, err := os.ReadDir(consensusInputDir)
```
peternose, line 166: Why is this not part of `restoreCheckpoints`?

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+		}
+	}
+
+	return nil
+}
+
+func restoreCheckpoints(ctx context.Context, ndb api.NodeDB, ns common.Namespace, inputDir string) error {
+	isEmpty, err := isEmptyDir(inputDir)
+	if err != nil {
+		return fmt.Errorf("aborting checkpoint restoration: failed to inspect input directory: %w", err)
```
peternose:
```suggestion
		return fmt.Errorf("failed to inspect input directory: %w", err)
```
This does not belong here.

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+	}
+	if _, ok := ndb.GetLatestVersion(); ok {
+		return fmt.Errorf("aborting checkpoint restoration: state db not empty")
+	}
+
+	provider, err := checkpoint.NewFileCreator(inputDir, nil) // ndb = nil since we will only restore checkpoints.
+	if err != nil {
+		return fmt.Errorf("failed to create checkpoint file creator: %w", err)
+	}
+	cps, err := provider.GetCheckpoints(ctx, &checkpoint.GetCheckpointsRequest{Version: 1, Namespace: ns})
```
peternose, line 298:
```suggestion
	cps, err := provider.GetCheckpoints(ctx, &checkpoint.GetCheckpointsRequest{Version: checkpoint.V1, Namespace: ns})
```

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+		roots = append(roots, cp.Root)
+	}
+	if err := ndb.Finalize(roots); err != nil {
+		return fmt.Errorf("failed to finalize: %w", err)
+	}
+
+	return nil
+}
+
+func importCp(ctx context.Context, ndb api.NodeDB, provider checkpoint.Creator, cp *checkpoint.Metadata) error {
```
peternose, line 329: The name is not consistent with other functions.
Why don't we use `Restorer`?

go/oasis-node/cmd/storage/checkpoint_test.go
```diff
@@ -0,0 +1,332 @@
...
+		require.True(t, dstNDB.HasRoot(stateRoot))
+		require.True(t, dstNDB.HasRoot(ioRoot))
+	})
+}
+
+func newTestNodeDB(t *testing.T, ns common.Namespace) (dbAPI.NodeDB, error) {
+	t.Helper()
+
+	return pathbadger.New(&dbAPI.Config{
+		DB:         filepath.Join(t.TempDir(), "db"),
```
peternose: Why is sub-directory `db` needed?

go/oasis-node/cmd/storage/checkpoint_test.go
```diff
@@ -0,0 +1,332 @@
...
+	rootType node.RootType,
+	data map[string]string,
+) node.Root {
+	t.Helper()
+
+	tree := mkvs.New(nil, ndb, rootType)
+	defer tree.Close()
+
+	for k, v := range data {
+		require.NoError(t, tree.Insert(ctx, []byte(k), []byte(v)))
```
peternose:
```suggestion
		err := tree.Insert(ctx, []byte(k), []byte(v))
		require.NoError(t, err)
```

go/oasis-node/cmd/storage/checkpoint_test.go
```diff
@@ -0,0 +1,332 @@
...
+
+	return node.Root{
+		Namespace: ns,
+		Version:   version,
+		Type:      rootType,
+		Hash:      rootHash,
+	}
+}
+
+func emptyRoot(ns common.Namespace, version uint64, rootType node.RootType) node.Root {
```
peternose: You already use this in `checkpoint.go`, so maybe we could add a public constructor `node.NewEmptyRoot` or `node.NewRoot` and `root.Empty`.

go/oasis-node/cmd/storage/checkpoint_test.go
```diff
@@ -0,0 +1,332 @@
...
+	t.Run("fails on non-empty output dir", func(t *testing.T) {
+		t.Parallel()
+
+		ctx := t.Context()
+		ns := common.Namespace{}
+		ndb, err := newTestNodeDB(t, ns)
+		require.NoError(t, err)
+		defer ndb.Close()
+
+		stateRoot := commitRoot(ctx, t, ndb, ns, testVersion, node.RootTypeState, map[string]string{
```
peternose, line 32: I would probably create a helper struct with method `func (db *testNodeDB) Commit(ctx context.Context, version uint64, rootType node.RootType, data map[string]string) node.Root` and try to avoid passing `t` to helper functions, as then you have just too many arguments.

go/oasis-node/cmd/storage/checkpoint_test.go
```diff
@@ -0,0 +1,332 @@
...
+			"key": "value",
+		})
+		require.NoError(t, ndb.Finalize([]node.Root{stateRoot}))
+
+		outDir := filepath.Join(t.TempDir(), "checkpoints")
```
peternose, line 37: You should use a constant for directory name.

go/oasis-node/cmd/storage/checkpoint_test.go
```diff
@@ -0,0 +1,332 @@
...
+		require.NoError(t, os.Mkdir(outDir, 0o700))
+		require.NoError(t, os.WriteFile(filepath.Join(outDir, "existing"), []byte{}, 0o600))
+
+		err = createCheckpoints(ctx, ndb, ns, testVersion, outDir)
+		require.Error(t, err, "createCheckpoints should fail for non-empty output directory")
```
peternose:
```suggestion
		require.ErrorContains(t, err, "output directory is not empty")
```

You should check if tested function fails for the right reason.

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,584 @@
...
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&inputDir, "input-dir", "", "directory with checkpoints to import")
+
+	return cmd
+}
+
+func createCheckpoints(ctx context.Context, ndb api.NodeDB, ns common.Namespace, version uint64, outputDir string) error {
```
peternose, line 234: Maybe creating a struct `checkpointer` would be better, as you could create multiple checkpoints with the same struct, e.g.
```
cp.Create(ctx, 1, "/checkpoints/1")
cp.Create(ctx, 2, "/checkpoints/2")
...
```
or even without `outputDir` if accepted in the constructor. The new struct might also be easier to test and could be decoupled from the commands.

go/oasis-test-runner/scenario/e2e/runtime/checkpoint_create_import.go
```diff
@@ -0,0 +1,149 @@
...
+	consensus "github.com/oasisprotocol/oasis-core/go/consensus/api"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/env"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis/cli"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario"
+	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
+)
+
+// CheckpointCreateImport is the checkpoint create/import e2e scenario.
+var CheckpointCreateImport scenario.Scenario = newCheckpointCreateImportImpl()
```
peternose, line 17: One could argue that this functionality should not be tested with E2E tests as a bug in the functions doesn't affect nodes running the buggy version of the `oasis-node`.

go/oasis-node/cmd/storage/checkpoint.go
peternose: Currently things are not fine, as you can do consensus and runtime checkpoints at the same time. Furthermore, this might be confusing for users, e.g.., do they need to set `height`, `round`, both?

peternose (commented):
When I import a consensus checkpoint, I get few lines of the following error. Afterwards, blocks execute normally. 

```
{"caller":"grpc.go:194","err":"failed to get consensus status: failed to fetch current block: cometbft: block query failed: height 28800866 must be less than or equal to the current blockchain height 0","level":"error","method":"/oasis-core.NodeController/GetStatus","module":"grpc/internal","msg":"request failed","req_seq":15,"ts":"2026-02-24T13:00:28.934344662Z"}
```

go/oasis-test-runner/scenario/e2e/runtime/checkpoint_create_import.go
martintomazic: I see your point but also observe this is still part of the `oasis-node` binary.

As pointed in the commit message introducing this I would prefer to:

>Furthermore, given that e2e tests are expensive and meant to
test complex scenarios, my suggestions would be to also run
prune, compact, and inspect command on the source node prior
to creating a checkpoint. This way we would "smoke test"
remaining storage commands, and the scenario could be called
storaged_cmds instead


So e2e test for complex scenarios + smoke testing command wiring. Everything else should be tested via unit+integration.

With the new `new*Cmd` pattern you can also easily test cmd argument validation/combinations. The reason why we need e2e here is 1. smoke testing wiring of the existing storage commands + boostrap cometbft DBs would be very complex to mock and test otherwise.


Update: We could also make e2e tests like this only run prior to release.

go/oasis-node/cmd/storage/checkpoint.go
martintomazic: **update**:
>The new struct might also be easier to test and could be decoupled from the commands.

`newCheckpointer` + `cp.create` (proposed) = `createCheckpoints` (current) so this is only a matter of style. 

As usual I prefer an explicit function over abstractions until multiple methods share same parameter set.

-------------------------------------
The question is do we want to allow multiple checkpoint heights/rounds for the same NodeDB.

If you want to make this generic helper I think this would fit inside `checkpoint` package. 
See ( https://github.com/oasisprotocol/oasis-core/issues/6467):

```go
// Consider using functional optional arguments to shorten args list.
CreateCheckpoint(ctx context.Context, ndb db.NodeDB, store Store, root node.Root, chunkSize uint64, chunkerThreads uint16) (*Metadata, error)

// Maybe add CreateAllCheckpoints as well and avoid passing root there.
```

go/oasis-node/cmd/storage/checkpoint_test.go
martintomazic: Probably you mean this for other tests as well? In general I wanted to avoid it as errors are not part of the public API, but then yes the test can fail for a wrong reason. Will add `ErrorContains` as you suggest.

go/oasis-node/cmd/storage/checkpoint_test.go
martintomazic: In the target test, with a short scope, where in-lining reads nicer?

go/oasis-node/cmd/storage/checkpoint_test.go
martintomazic: Will refactor and return `error` instead.

I prefer explicit functions over objects if the benefit is only one argument less, as this only adds complexity. Of course if you can consistently safe on multiple arguments then struct is more suitable.

**Update**: I find the current version simpler and easier taking into account we are using assertion library. My personal preference is no assertion libraries and use of `go-cmp` library.

go/oasis-node/cmd/storage/checkpoint.go
martintomazic: ~import only imports it, restore will import checkpoints and also finalize them. Maybe explicit comment would be better?~ Used consistent naming now.

>Why don't we use Restorer?

Because it only adds complexity, since it only helps you when your chunks are added in non-sequential logic. Looping over chunks still has to be done manually, as well as `Start/AbortMultipartInsert`.

go/oasis-node/cmd/storage/checkpoint.go
martintomazic: Because I want to keep `restoreCheckpoint` generic function that restores a checkpoint from the existing  `checkpoint.FileStore`   (#6467). 


In general I think core business logic should be separate from the command dir parsing.

go/oasis-node/cmd/storage/checkpoint.go
martintomazic: >I guess consensus checkpoint is created and runtime checkpoint creation fails as the directory is not empty.
 
No since every nodeDB has a separate dir and I only prevent creating a checkpoint in the non-empty _checkpoint file store_. With #6467 and explicit `checkpoint.FileStore` I think this would be clearer so maybe we should implement it first?

go/oasis-node/cmd/storage/checkpoint.go
martintomazic: I am fine with that, but then we should also reflect this in the import command. Take latest? Pass expected height/round to the the import command as well? 

Personally, I prefer existing constraint and fail fast to keep other parts simpler. But probably I should document it.


Update: Also this check is to that the use doesn't accidentally write to unexpected dirs (obviously not thorough).

go/oasis-node/cmd/storage/checkpoint.go
martintomazic: Go style discourages _opaque_ errors.

I believe good error trace gives you context to debug the issues just looking at the error trace. E.g. adding a range here, makes the root command return the exact issue to the caller, saving him debug time and calling other commands such as inspect to find a valid range.

E.g. recent error from the CI made is so much simpler to be sure what went wrong: 
```bash
Error: failed to create runtime checkpoints (runtime: 8000000000000000000000000000000000000000000000000000000000000000, round: 15): version not finalized (finalized range: 0-14)
```

In general following this practice makes your life so much easier when an error happens in the flaky/hard to reproduce env.

go/oasis-node/cmd/storage/checkpoint.go
martintomazic: Similar to the range error above, this adds a necessarily context, so that one can quickly see for which height this was called or possibly spot an error when wiring of the command args to this function is wrong.

go/oasis-node/cmd/storage/checkpoint.go
martintomazic: Yes I left it intentional for now. I can easily allow only one at the time. The question is would using sub-commands make things even clearer? Also any preference for what should omitting height/round do? 

Observe also that if you want to be able to create checkpoints for multiple heights/versions (same NodeDB) the import command also grows in complexity.

Finally, one should be very careful for which height to specific runtime round combinations you are creating as you can easily get left missing missing runtime light blocks, and therefore stuck runtime state restore.

We could also make import command use `consensus/runtime`/`height/round` and import the actual directories created one by one to make it symmetric to create.

Any pref?

go/oasis-node/cmd/storage/checkpoint.go
martintomazic: Yes this should be done, but `checkpoint.V1` does not exist. Since there are many other callsites I believe this is out of scope.

martintomazic:
>When I import a consensus checkpoint, I get few lines of the following error. Afterwards, blocks execute normally.

Nice catch. Yes this is also how CometBFT checkpoint import works, but found a fixup regardless :).

The more annoying thing that I find is that you technically cannot import a checkpoint for the latest height, so probably adding extra validation + documenting this in the command would be beneficial, instead of unexpected error.

martintomazic:
Ready for a second review.

As you spotted I am "abusing" `checkpoint.FileStore` abstraction (https://github.com/oasisprotocol/oasis-core/issues/6467)  bound to one concrete NodeDB, that should be able to create and import checkpoints for different versions.

However, this command is technically not `create/import checkpoint` but rather `create/import state/snapshot`. Indeed 99% is creating and importing checkpoints and using `checkpoint.FileStore` for encoding the snapshot structure. Nevertheless, CometBFT bootstrap metadata has nothing to do with the checkpoints alone.

 For this reason I have created my helpers (stateless), so that they can also be easily refactored, possibly moved to `checkpoint` package one day. Until this command is the only client I am not sure it makes sense. Improving `checkpoint` package abstractions probably does. :) 

Let's align on the user facing API and sanity checking the inputs:
1. https://github.com/oasisprotocol/oasis-core/pull/6454#discussion_r2847877470
    * The simpler the better given we will the only user of this command:
3. Should we find a better name for this command given that is not just creating/importing checkpoints? Not sure.

go/oasis-node/cmd/storage/checkpoint.go
peternose: The problem that I saw here was that users have no idea what a finalized version is. It is too specific for them. For us, this error should be fine.

go/oasis-node/cmd/storage/checkpoint.go
peternose: If height was given as a parameter to a function call, I would exclude it from the error. But it doesn't matter that much.

go/oasis-node/cmd/storage/checkpoint.go
peternose: Checkpoint providers will probably want to have multiple versions in the same folder, right?

go/oasis-node/cmd/storage/checkpoint.go
peternose: You are right, it will not fail. However, it is still strange that one command can do 2 checkpoints at once (one for consensus and one for runtime).

go/oasis-node/cmd/storage/checkpoint.go
peternose: But we can change `v1` to be public?

go/oasis-node/cmd/storage/checkpoint_test.go
peternose: I thought that if we change name `checkpoints`, then this test will fail if we don't change the name here also.

go/oasis-node/cmd/storage/checkpoint_test.go
peternose: > Probably you mean this for other tests as well?

Yes. If you are testing that a function fails for a specific reason, you should check the error message. If you only check it fails, then your solution is fine.

go/oasis-node/cmd/storage/checkpoint.go
peternose: Yes, it is a matter of style by the quality/complexity of the code. And more parameters you have, the harder is to follow. And as suggested, compare code if you need to create multiple checkpoints for different versions with a function or with a method call.

go/oasis-test-runner/scenario/e2e/runtime/checkpoint_create_import.go
peternose: You can leave it as it is. I just wanted to say that I would probably not make an e2e test here, only unit or integration test.

peternose:
> Ready for a second review.

Will have a look. Merge after we release 26.0.

go/oasis-node/cmd/storage/checkpoint.go
martintomazic: I see your point but in practice we are talking about:
1. Checkpoint providers or simply `checkpoint.FileStore` ((https://github.com/oasisprotocol/oasis-core/issues/6467)
2. The original command due to the CometBFT metada is no longer create/import checkpoint but instead create/import snapshot....
    * Indeed it uses a subset of functionality from **1.**
    * Creating a snapshot is normally a single operation, not multiple calls. 
    * We could add `--with-cometbft-meta` flag, but this is adding complexity.

My goal was to make **1.** independent, but still internal because I don't want to expose this functionality since we don't need it, i.e. we don't envision checkpoint provider role.

So: 
1. Let's refactor abstractions to use `checkpoint.FileStore` implementation, possibly expose commands that create and import checkpoints into/from the `checkpoint.Store`. Probably over-engineering until we actually have a client/need for it.
2. Align on the snapshot create/import that uses `1.`, and keep existing sanity validation for the empty dirs etc to enforce snapshot structure. I could also be in tar, custom binary encoding (`.cp`). I prefer to keep it simple.

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,625 @@
...
+					"output_dir", consensusOutDir,
+				)
+				if err := createConsensusCp(consensusOutDir); err != nil {
+					return fmt.Errorf("failed to create consensus checkpoint (height: %d): %w", height, err)
+				}
+				logger.Info("consensus checkpoint created")
+
+				if err := createCometBFTBootstrapMeta(dataDir, height, consensusOutDir); err != nil {
+					return fmt.Errorf("failed to write bootstrap metadata: %w", err)
+				}
```
peternose, line 102:
```suggestion
				}
				
				logger.Info("bootstrap metadata created")
```

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,625 @@
...
+				logger.Info("consensus checkpoint imported")
+
+				meta, err := readCometBFTBootstrapMeta(consensusInputDir)
+				if err != nil {
+					return fmt.Errorf("failed to read bootstrap metadata: %w", err)
+				}
+				if err := bootstrapTrustedState(dataDir, meta); err != nil {
+					return fmt.Errorf("failed to bootstrap CometBFT trusted state: %w", err)
+				}
```
peternose, line 181:
```suggestion
				}
				
				logger.Info("bootstrap metadata imported")
```

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,625 @@
...
+			case os.IsNotExist(err):
+				// No consensus checkpoints to import
```
peternose, line 184: Should we log that there is nothing to import?

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,625 @@
...
+	err = stateStore.SetOfflineStateSyncHeight(metaState.LastBlockHeight)
+	if err != nil {
+		return fmt.Errorf("failed to set synced height: %w", err)
+	}
+
+	return nil
+}
+
+// commit is adapted and simplified and mimics StateProvider behaviour used in the upstream BootstrapState.
+func commit(blockStore *store.BlockStore, height uint64) (*cmttypes.Commit, error) {
```
peternose: Can you rename to something like `getCommit`, `fetchCommit`, or `loadCommit`?

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,625 @@
...
+	}
+	cps, err := provider.GetCheckpoints(ctx, &checkpoint.GetCheckpointsRequest{Version: 1, Namespace: ns})
+	if err != nil {
+		return fmt.Errorf("failed to read checkpoints: %w", err)
+	}
+
+	var roots []node.Root
+	for _, cp := range cps {
+		if err := importCp(ctx, ndb, provider, cp); err != nil {
+			return fmt.Errorf("failed to import checkpoint (root hash: %s): %w", cp.Root.Hash, err)
```
peternose, line 318: If one fails because of file corruption, we cannot repeat the process as the checkpoints were partially imported. Should we add a command which cleans/removes partially imported stuff?

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,625 @@
...
+	h := int64(height)
+	lastMeta := blockStore.LoadBlockMeta(h)
+	if lastMeta == nil {
+		return cmtState.State{}, fmt.Errorf("block meta not found at height %d", h)
+	}
+	currentMeta := blockStore.LoadBlockMeta(h + 1)
+	if currentMeta == nil {
+		return cmtState.State{}, fmt.Errorf("block meta not found at height %d", h+1)
+	}
+	nextMeta := blockStore.LoadBlockMeta(h + 2)
```
peternose: This is not needed.

go/oasis-node/cmd/storage/checkpoint.go
```diff
@@ -0,0 +1,625 @@
...
+		return err
+	}
+
+	err = blockStore.SaveSeenCommit(metaState.LastBlockHeight, commit)
+	if err != nil {
+		return err
+	}
+
+	// SaveSeenCommit (as used by the upstream) does not persist the BlockStoreState (height/base),
+	// so we save it explicitly here. This ways the blockstore reports the correct height after bootstrap,
```
peternose:
```suggestion
	// so we save it explicitly here. This way the blockstore reports the correct height after bootstrap,
```

go/oasis-test-runner/scenario/e2e/runtime/checkpoint_create_import.go
```diff
@@ -0,0 +1,149 @@
...
+func (sc *checkpointCreateImportImpl) Run(ctx context.Context, childEnv *env.Env) error {
+	// Start the network and run the test client to populate state.
+	if err := sc.StartNetworkAndTestClient(ctx, childEnv); err != nil {
+		return err
+	}
+	if err := sc.WaitTestClient(); err != nil {
+		return err
+	}
+
+	// Use height - 3 so that blocks at h, h+1, h+2 all exist in the block store.
```
peternose, line 57: Are we sure there will always be 3 blocks at this point or should we wait for e.g. block number 10.

go/oasis-test-runner/scenario/e2e/runtime/checkpoint_create_import.go
```diff
@@ -0,0 +1,149 @@
...
+		return fmt.Errorf("failed to restart source compute worker: %w", err)
+	}
+
+	// Stop compute worker 2 (target node).
+	target := sc.Net.ComputeWorkers()[2]
+	if err := target.StopGracefully(); err != nil {
+		return fmt.Errorf("failed to stop target compute worker: %w", err)
+	}
+
+	// Reset the target node's state completely. Ideally we would use NoAutoStart,
```
peternose, line 123: Maybe we should start the network for 10 blocks, stop the target node, wait for 20 blocks more, stop the source node, do snapshots, start the target node, and import them afterwards.

go/oasis-test-runner/scenario/e2e/runtime/checkpoint_create_import.go
```diff
@@ -0,0 +1,149 @@
...
+
+	sc.Logger.Info("checkpoints imported, starting target node")
+
+	// Start the target node.
+	if err := target.Start(); err != nil {
+		return fmt.Errorf("failed to start target node: %w", err)
+	}
+
+	// Wait for the target node to sync.
+	sc.Logger.Info("waiting for target node to sync")
```
peternose: Can we be sure that the target node synced from checkpoint and not from genesis?

go/oasis-test-runner/scenario/e2e/runtime/checkpoint_create_import.go
```diff
@@ -0,0 +1,196 @@
...
+	})
+	if err != nil {
+		return fmt.Errorf("failed to get runtime state for height %d: %w", candidateHeight, err)
+	}
+
+	// Pick runtime state's LastBlockHeight as the consensus checkpoint height else
+	// runtime light history indexer might miss authoritative light block for the
+	// corresponding runtime round.
+	cpRound := rtState.LastBlock.Header.Round
+	cpHeight := rtState.LastBlockHeight
```
martintomazic, line 75: I believe we have same issue with our checkpoint sync.

Consensus might create a checkpoint, and we would trigger checkpoint creation for the corresponding round for configured runtimes. The problem is that if this round was not created at given height, but before, indexer would skip it:
https://github.com/oasisprotocol/oasis-core/blob/59b8f325ffd4d3ab4455b0495b78552a7b4d739e/go/runtime/history/indexer.go#L357
Which would cause the corresponding runtime checkpoint sync to fail due to a missing authoritative light header.  Out of scope, will open an issue.

go/oasis-test-runner/scenario/e2e/runtime/checkpoint_create_import.go
martintomazic: Starting test client guarantees this practically, but I am also open to calling `WaitForConsensuHeight` before that if you see this as needed.

go/oasis-node/cmd/storage/checkpoint.go
martintomazic: I would not since only creating/importing runtime checkpoints is perfectly valid. 

We do log/error in case of the consensus checkpoint and missing metadata which doesn't make sense.

go/oasis-test-runner/scenario/e2e/runtime/checkpoint_create_import.go
```diff
@@ -0,0 +1,196 @@
...
+	consensus "github.com/oasisprotocol/oasis-core/go/consensus/api"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/env"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis/cli"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario"
+	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
+)
+
+// CheckpointCreateImport is the checkpoint create/import e2e scenario.
+var CheckpointCreateImport scenario.Scenario = newCheckpointCreateImportImpl()
```
martintomazic, line 17: There was one (flaky?) outcome in the CI that I am trying to reproduce locally (might have seen it once locally before).
https://buildkite.com/oasisprotocol/oasis-core-ci/builds/16664#019d43f3-6e51-4ddc-b8a4-a0ab5da22719

```bash
{"caller":"worker.go:1155","err":"storage/database: failed to Apply: mkvs: node not found in node db","level":"error","module":"worker/storage/committee","msg":"can't apply write log","new_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=13 type=state-root hash=4848fde3109b8a49e559a3a10040e25af7c5d79f8eeeecbd55d30738f3baec55>","old_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=12 type=state-root hash=963bc9c0eaf31ad37a8c9a25146d047fdef7d468b89f19a301ab28e76f84ee65>","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:20:47.085392644Z"}
```

_The imported checkpoint was corrupted_ as storage committee worker was unable to apply next storage diff. Might be also committee worker issue:

<details>
<summary>Storage committee logs from the CI </summary>

```bash
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{state-root}","round":11,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:31.02065292Z"}
{"caller":"worker.go:1175","level":"debug","module":"worker/storage/committee","msg":"finished syncing round","round":11,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:31.02067514Z"}
{"caller":"worker.go:439","level":"debug","module":"worker/storage/committee","msg":"storage round finalized","round":11,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:31.020769116Z"}
{"caller":"worker.go:1204","last_finalized":11,"last_synced":11,"level":"debug","module":"worker/storage/committee","msg":"incoming block","round":12,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:33.737682523Z"}
{"awaiting_retry":"outstanding_mask{state-root}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":12,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:33.73770249Z"}
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":12,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:33.737737249Z"}
{"caller":"worker.go:1175","level":"debug","module":"worker/storage/committee","msg":"finished syncing round","round":12,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:33.737816314Z"}
{"caller":"worker.go:439","level":"debug","module":"worker/storage/committee","msg":"storage round finalized","round":12,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:33.737930232Z"}
{"caller":"worker.go:1204","last_finalized":12,"last_synced":12,"level":"debug","module":"worker/storage/committee","msg":"incoming block","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:35.059190353Z"}
{"awaiting_retry":"outstanding_mask{state-root}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:35.059224542Z"}
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{state-root}","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:35.059398047Z"}
{"caller":"worker.go:1175","level":"debug","module":"worker/storage/committee","msg":"finished syncing round","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:35.05943708Z"}
{"caller":"worker.go:439","level":"debug","module":"worker/storage/committee","msg":"storage round finalized","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:35.059690354Z"}
{"caller":"worker.go:1204","last_finalized":13,"last_synced":13,"level":"debug","module":"worker/storage/committee","msg":"incoming block","round":14,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:36.462700172Z"}
{"awaiting_retry":"outstanding_mask{state-root}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":14,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:36.462737431Z"}
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{state-root}","round":14,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:36.462816046Z"}
{"caller":"worker.go:1175","level":"debug","module":"worker/storage/committee","msg":"finished syncing round","round":14,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:36.462842958Z"}
{"caller":"worker.go:439","level":"debug","module":"worker/storage/committee","msg":"storage round finalized","round":14,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:36.463146676Z"}
{"caller":"worker.go:1204","last_finalized":14,"last_synced":14,"level":"debug","module":"worker/storage/committee","msg":"incoming block","round":15,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:37.789492905Z"}
{"awaiting_retry":"outstanding_mask{state-root}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":15,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:37.789509283Z"}
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{state-root}","round":15,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:37.789540837Z"}
{"caller":"worker.go:1175","level":"debug","module":"worker/storage/committee","msg":"finished syncing round","round":15,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:37.789551362Z"}
{"caller":"worker.go:439","level":"debug","module":"worker/storage/committee","msg":"storage round finalized","round":15,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:37.78963392Z"}
{"caller":"worker.go:789","err":"context canceled","level":"error","module":"worker/storage/committee","msg":"checkpointer stopped","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:38.169306642Z"}
{"caller":"worker.go:1318","level":"info","module":"worker/storage/committee","msg":"stopped","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:38.169313829Z"}
{"caller":"worker.go:768","level":"info","module":"worker/storage/committee","msg":"starting","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.835974673Z"}
{"caller":"worker.go:948","genesis_round":0,"last_synced":11,"level":"info","module":"worker/storage/committee","msg":"worker initialized","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.836554655Z"}
{"caller":"worker.go:1021","level":"info","module":"worker/storage/committee","msg":"initialized","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.83663171Z"}
{"caller":"worker.go:1204","last_finalized":11,"last_synced":11,"level":"debug","module":"worker/storage/committee","msg":"incoming block","round":16,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.836767167Z"}
{"awaiting_retry":"outstanding_mask{state-root}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":12,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.83698815Z"}
{"awaiting_retry":"outstanding_mask{state-root}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.837095354Z"}
{"caller":"worker.go:401","level":"debug","module":"worker/storage/committee","msg":"calling GetDiff","new_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=13 type=state-root hash=4848fde3109b8a49e559a3a10040e25af7c5d79f8eeeecbd55d30738f3baec55>","old_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=12 type=state-root hash=963bc9c0eaf31ad37a8c9a25146d047fdef7d468b89f19a301ab28e76f84ee65>","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.837161045Z"}
{"awaiting_retry":"outstanding_mask{state-root}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":14,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.837259592Z"}
{"awaiting_retry":"outstanding_mask{state-root}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":15,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.837338432Z"}
{"awaiting_retry":"outstanding_mask{state-root}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":16,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.837388739Z"}
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":12,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.837444513Z"}
{"caller":"worker.go:401","level":"debug","module":"worker/storage/committee","msg":"calling GetDiff","new_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=14 type=io-root hash=7fa1c8d40fdd82e2c0af8ffd3009890a4d5cc1109e1b36789b7fd283a95bf07e>","old_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=14 type=io-root hash=c672b8d1ef56ed28ab87c3622c5114069bdd3ad7b8f9737498d0c01ecef0967a>","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.837508641Z"}
{"caller":"worker.go:401","level":"debug","module":"worker/storage/committee","msg":"calling GetDiff","new_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=14 type=state-root hash=9314bf6ac6112131839cae80d8452dfaabcb0e408413b402e6573eb53eed3333>","old_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=13 type=state-root hash=4848fde3109b8a49e559a3a10040e25af7c5d79f8eeeecbd55d30738f3baec55>","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.837837373Z"}
{"caller":"worker.go:1175","level":"debug","module":"worker/storage/committee","msg":"finished syncing round","round":12,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.838913637Z"}
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.838980235Z"}
{"caller":"worker.go:1155","err":"storage/database: failed to Apply: mkvs: node not found in node db","level":"error","module":"worker/storage/committee","msg":"can't apply write log","new_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=13 type=state-root hash=4848fde3109b8a49e559a3a10040e25af7c5d79f8eeeecbd55d30738f3baec55>","old_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=12 type=state-root hash=963bc9c0eaf31ad37a8c9a25146d047fdef7d468b89f19a301ab28e76f84ee65>","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.8390787Z"}
{"caller":"worker.go:439","level":"debug","module":"worker/storage/committee","msg":"storage round finalized","round":12,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.839273082Z"}
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.839260743Z"}
{"caller":"worker.go:401","level":"debug","module":"worker/storage/committee","msg":"calling GetDiff","new_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=15 type=io-root hash=a15d4949610bf4c365b0a75368b6e79bf751c57067fd456ed34f1a5693e00ee6>","old_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=15 type=io-root hash=c672b8d1ef56ed28ab87c3622c5114069bdd3ad7b8f9737498d0c01ecef0967a>","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.839331111Z"}
{"caller":"worker.go:401","level":"debug","module":"worker/storage/committee","msg":"calling GetDiff","new_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=15 type=state-root hash=b9f885bdbbeb1504c9b0e46d186ed90af8fed7f88b0c8772058a483ee0efd9b0>","old_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=14 type=state-root hash=9314bf6ac6112131839cae80d8452dfaabcb0e408413b402e6573eb53eed3333>","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.839631516Z"}
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.839724966Z"}
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.839799029Z"}
{"caller":"worker.go:401","level":"debug","module":"worker/storage/committee","msg":"calling GetDiff","new_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=13 type=state-root hash=4848fde3109b8a49e559a3a10040e25af7c5d79f8eeeecbd55d30738f3baec55>","old_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=12 type=state-root hash=963bc9c0eaf31ad37a8c9a25146d047fdef7d468b89f19a301ab28e76f84ee65>","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.839847976Z"}
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.840478217Z"}
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.840622243Z"}
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.841242907Z"}
{"caller":"worker.go:1155","err":"storage/database: failed to Apply: mkvs: node not found in node db","level":"error","module":"worker/storage/committee","msg":"can't apply write log","new_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=13 type=state-root hash=4848fde3109b8a49e559a3a10040e25af7c5d79f8eeeecbd55d30738f3baec55>","old_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=12 type=state-root hash=963bc9c0eaf31ad37a8c9a25146d047fdef7d468b89f19a301ab28e76f84ee65>","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.841528927Z"}
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.841592563Z"}
{"caller":"worker.go:401","level":"debug","module":"worker/storage/committee","msg":"calling GetDiff","new_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=13 type=state-root hash=4848fde3109b8a49e559a3a10040e25af7c5d79f8eeeecbd55d30738f3baec55>","old_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=12 type=state-root hash=963bc9c0eaf31ad37a8c9a25146d047fdef7d468b89f19a301ab28e76f84ee65>","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.841656563Z"}
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.842566191Z"}
{"caller":"worker.go:1155","err":"storage/database: failed to Apply: mkvs: node not found in node db","level":"error","module":"worker/storage/committee","msg":"can't apply write log","new_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=13 type=state-root hash=4848fde3109b8a49e559a3a10040e25af7c5d79f8eeeecbd55d30738f3baec55>","old_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=12 type=state-root hash=963bc9c0eaf31ad37a8c9a25146d047fdef7d468b89f19a301ab28e76f84ee65>","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:06:55.842681114Z"}
{"caller":"worker.go:1261","in_flight_rounds":4,"level":"debug","module":"worker/storage/committee","msg":"heartbeat","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:07:03.261447066Z"}
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:07:03.261550558Z"}
{"caller":"worker.go:401","level":"debug","module":"worker/storage/committee","msg":"calling GetDiff","new_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=13 type=state-root hash=4848fde3109b8a49e559a3a10040e25af7c5d79f8eeeecbd55d30738f3baec55>","old_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=12 type=state-root hash=963bc9c0eaf31ad37a8c9a25146d047fdef7d468b89f19a301ab28e76f84ee65>","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:07:03.26165467Z"}
{"awaiting_retry":"outstanding_mask{}","caller":"worker.go:1074","level":"debug","module":"worker/storage/committee","msg":"preparing round sync","outstanding_mask":"outstanding_mask{}","round":13,"runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:07:03.263252444Z"}
{"caller":"worker.go:1155","err":"storage/database: failed to Apply: mkvs: node not found in node db","level":"error","module":"worker/storage/committee","msg":"can't apply write log","new_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=13 type=state-root hash=4848fde3109b8a49e559a3a10040e25af7c5d79f8eeeecbd55d30738f3baec55>","old_root":"<Root ns=8000000000000000000000000000000000000000000000000000000000000000 version=12 type=state-root hash=963bc9c0eaf31ad37a8c9a25146d047fdef7d468b89f19a301ab28e76f84ee65>","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:07:03.26339852Z"}
{"caller":"worker.go:1261","in_flight_rounds":4,"level":"debug","module":"worker/storage/committee","msg":"heartbeat","runtime_id":"8000000000000000000000000000000000000000000000000000000000000000","ts":"2026-03-31T13:07:12.7026433Z"}
```
</details>

^^ Worker synced from rounds 0-15, then state was reset. After that round 11 is imported, round 12 root does not change hence no need to fetch it but we still see finalized round. Finally round 13 fails.

All other local/CI runs have no such issues whatsoever, including testing this command on the real mainnet data.

Submitting more runtime transactions once the checkpoint was created, so that restarted node has to catch-up with more (and new) state, did not trigger this outcome either.

PR #6460, opened 2026-02-11
Bump Rust toolchain to 2026-02-11

kostko:
Based on #6442

### DIFF
--- a/.changelog/6460.internal.md
+++ b/.changelog/6460.internal.md
@@ -1,0 +1,1 @@
+Bump Rust toolchain to 2026-02-11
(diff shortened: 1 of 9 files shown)


peternose:
Similar to https://github.com/oasisprotocol/oasis-core/pull/6442. I think the problem is in the `runtime-loader`, which doesn't work after the upgrade.

kostko:
Oh sorry, didn't see you already did that, I'll take a look.

We need to bump this for the new evm crate in the SDK and probably other things.

kostko:
And this was supposed to be draft ;)

PR #6466, opened 2026-02-22
go/registry/api: Prevent registering Observer node with 0 runtimes

martintomazic:
The change is under feature flag so should not be breaking.

Observe registration worker will block registering Observer node with 0 runtimes immediatelly. This should be safe as it is already the intent of existing node binaries and cannot cause consensus divergence.

### DIFF
--- a/go/registry/api/api.go
+++ b/go/registry/api/api.go
@@ -145,6 +145,12 @@ var (
 
 	// RuntimesRequiredRoles are the Node roles that require runtimes.
 	RuntimesRequiredRoles = node.RoleComputeWorker |
+		node.RoleObserver |
+		node.RoleKeyManager |
+		node.RoleStorageRPC
+
+	// DeprecatedRuntimesRequiredRoles are the legacy Node roles that require runtimes.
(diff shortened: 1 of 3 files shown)


go/registry/api/api.go
```diff
@@ -145,6 +145,13 @@ var (
 
 	// RuntimesRequiredRoles are the Node roles that require runtimes.
 	RuntimesRequiredRoles = node.RoleComputeWorker |
+		node.RoleObserver |
+		node.RoleKeyManager |
+		node.RoleStorageRPC
+
+	// RuntimesRequiredRolesBefore242 are the Node roles that require runtimes
```
peternose: Can you change the comment and vars so that value 242 is not used, as it will be renamed latter to 261 or something similar, and we might miss renaming this. For example, can you use `DeprecatedRuntimesRequiredRoles` instead?

peternose (commented):
LGTM. Can you also update the comment in `go/upgrade/migrations/consensus_242.go` with a new bullet?

PR #6471, opened 2026-03-04
Per role quote policies

martintomazic:
Replaces https://github.com/oasisprotocol/oasis-core/pull/6410 following simplifications agreed in the #6387.

**Considerations**:
- Can be simplified further: https://github.com/oasisprotocol/oasis-core/pull/6471#discussion_r2883283444
- We can make the runtime sign the role information as part of the attestation (similar to the node ID), so if the host lied about its role during RHP it can be later verified.
    - New RHP method, technically should not be breaking, add role as part of the hash structure. Breaking?
-  Architectural smell exposed: https://github.com/oasisprotocol/oasis-core/issues/6387#issuecomment-3923908101.
- _e2e test or at least sanity that everything is wired correctly and access gated_.

### DIFF
--- a/go/oasis-node/cmd/node/node.go
+++ b/go/oasis-node/cmd/node/node.go
@@ -11,6 +11,7 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/common/grpc"
 	"github.com/oasisprotocol/oasis-core/go/common/identity"
 	"github.com/oasisprotocol/oasis-core/go/common/logging"
+	"github.com/oasisprotocol/oasis-core/go/common/node"
 	"github.com/oasisprotocol/oasis-core/go/common/persistent"
 	"github.com/oasisprotocol/oasis-core/go/common/version"
 	"github.com/oasisprotocol/oasis-core/go/config"
@@ -205,12 +206,29 @@ func (n *Node) initRuntimeWorkers(genesisDoc *genesisAPI.Document) error {
 	}
 	n.svcMgr.Register(n.RuntimeRegistry)
 
+	// Determine runtime role.
+	runtimeRoles := node.RoleEmpty
+	switch config.GlobalConfig.Mode {
+	case config.ModeCompute:
+		runtimeRoles = node.RoleComputeWorker
+	case config.ModeKeyManager:
+		runtimeRoles = node.RoleKeyManager
+	case config.ModeClient, config.ModeStatelessClient:
+		if config.GlobalConfig.Registration.Entity != "" || config.GlobalConfig.Registration.EntityID != "" {
+			runtimeRoles |= node.RoleObserver
+		}
+		if config.GlobalConfig.Storage.PublicRPCEnabled {
+			runtimeRoles |= node.RoleStorageRPC
+		}
+	}
+
 	// Initialize the common worker.
 	n.CommonWorker, err = workerCommon.New(
 		n,
 		n.dataDir,
 		n.chainContext,
 		n.Identity,
+		runtimeRoles,
 		n.Consensus,
 		n.LightService,
 		n.P2P,
--- a/go/runtime/host/host.go
+++ b/go/runtime/host/host.go
@@ -22,6 +22,10 @@ type Config struct {
 	// ID is the runtime identifier.
 	ID common.Namespace
 
+	// RuntimeRoles is the runtime role mask the node will register RONL component with.
+	// In case of ROFL only nodes (no registration) this is expected to be empty role.
+	RuntimeRoles node.RolesMask
+
 	// Component is the component that should be provisioned.
 	Component *bundle.ExplodedComponent
 
--- a/go/runtime/host/sgx/common/common.go
+++ b/go/runtime/host/sgx/common/common.go
@@ -48,7 +48,7 @@ func GetQuotePolicy(
 				return nil, fmt.Errorf("malformed runtime SGX constraints: %w", err)
 			}
 
-			return sc.Policy, nil
+			return sc.PolicyFor(cfg.RuntimeRoles), nil
 		}
 		return fallbackPolicy, nil
 	case component.ROFL:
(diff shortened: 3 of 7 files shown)


go/common/node/sgx.go
```diff
@@ -32,9 +32,17 @@ type SGXConstraints struct {
...
+	// Policy is the default quote policy. The default policy must be satisfied
+	// unless there exists a corresponding per-role policy.
 	Policy *quote.Policy `json:"policy,omitempty"`
 
+	// PerRolePolicy defines additional role specific quote policies, that overwrite
+	// Policy when node with these roles does an attestation.
+	//
+	// A valid entry is for either [RoleComputeWorker] or [RoleObserver]. Single entry
+	// should not encode multiple roles.
+	PerRolePolicy map[RolesMask]quote.Policy `json:"per_role_policy,omitempty"`
```
martintomazic, line 44: Given that:
*  We only allow it for observer and compute roles
*  Keymanager runtime kind does not allow it  
*  We don't see discriminating observer/compute access, as they both have access to same secrets,

How about simplifying things further and instead have `ComputePolicy *quote.Policy` that only applies for the compute runtimes for the observer/compute roles. 

This should simplify assumptions, tests, comments, validations and avoid the need for new "at most one runtime sgx role" invariant.

go/common/node/node.go
```diff
@@ -593,7 +593,7 @@ func HashRAK(rak signature.PublicKey) hash.Hash {
 }
 
 // Verify verifies the node's TEE capabilities, at the provided timestamp and height.
-func (c *CapabilityTEE) Verify(teeCfg *TEEFeatures, ts time.Time, height uint64, constraints []byte, nodeID signature.PublicKey, isFeatureVersion242 bool) error {
+func (c *CapabilityTEE) Verify(teeCfg *TEEFeatures, ts time.Time, height uint64, constraints []byte, nodeID signature.PublicKey, nodeRoles RolesMask, isFeatureVersion242 bool) error {
```
martintomazic, line 596: NIT: All this functions should accept validation options structs?

go/common/node/node.go
```diff
@@ -332,7 +344,7 @@ func (n *Node) UnmarshalCBOR(data []byte) error {
 }
 
 // ValidateBasic performs basic descriptor validity checks.
-func (n *Node) ValidateBasic(strictVersion bool) error {
+func (n *Node) ValidateBasic(strictVersion bool, isFeatureVersion242 bool) error {
```
martintomazic, line 347: Nit: There is not such thing as `ValidateBasic`. Only `Validate` with clear invariants and tests for the invariants the `Validate` will check. Moreover, the validation params should probably be passes as part of the validation options struct. Possibly this check could also be part of the `VerifyRegisterNodeArgs` directly which avoids some additional changes, but feels off there.

go/registry/api/runtime.go
```diff
@@ -597,6 +597,9 @@ func (r *Runtime) ValidateDeployments(now beacon.EpochTime, params *ConsensusPar
 			if err := cs.ValidateBasic(params.TEEFeatures, isFeatureVersion242); err != nil {
 				return fmt.Errorf("%w: invalid SGX TEE constraints", ErrInvalidArgument)
 			}
+			if r.Kind == KindKeyManager && cs.PerRolePolicy != nil {
+				return fmt.Errorf("%w: invalid SGX TEE constraints: keymanager runtime with per-role policies", ErrInvalidArgument)
+			}
```
martintomazic, line 602: Or make it part of the `ValidateBasic`, again this should be `Validate` with validate option struct so that it is clear what we are validating.

go/common/node/sgx.go
```diff
@@ -96,6 +104,19 @@ func (sc *SGXConstraints) MarshalCBOR() ([]byte, error) {
 	}
 }
 
+// PolicyFor returns a matching per-role policy when present, or otherwise falls back to the default policy.
+//
+// This function expects role mask that has at most one runtime SGX role.
+func (s *SGXConstraints) PolicyFor(roles RolesMask) *quote.Policy {
```
martintomazic: It would be so much nicer if we would have `RuntimeSGXRole` type so that comments, invariants and corner cases are instead captured by the type system. Even per role policies could use this type as a key. Not sure there is practical / idiomatic way to achieve this in go.

martintomazic:
Ready for the first round of reviews.

Please focus on the considerations outlined in the context and existing threads.

Given motivation and intended usage of this feature, I am in favor of replacing per-role policies with a single compute runtime policy that applies to both observer and compute roles.

.changelog/6387.internal.0.md
```diff
@@ -0,0 +1,4 @@
+go/registry/api: Allow at most one runtime SGX role
```
kostko, line 1:
```suggestion
go/registry/api: Allow at most one runtime TEE role
```

go/common/node/node.go
```diff
@@ -230,6 +230,18 @@ func (m RolesMask) IsSingleRole() bool {
 	return m != 0 && m&(m-1) == 0 && m&RoleReserved == 0
 }
 
+// AtMostOneRuntimeSGXRole returns true when RoleMask has at most one SGX runtime role.
+func (m RolesMask) AtMostOneRuntimeSGXRole() bool {
+	sgxRoles := m & (RoleComputeWorker | RoleObserver | RoleKeyManager)
```
kostko, line 235: The set of roles should be a top-level constant. Also note that these are not SGX-specific so you should use "TEE" instead of "SGX".

go/common/node/node.go
kostko: Agreed, this is why we introduced the `TEEFeatures` struct a while back (which is now somewhat obsolete due to the use of feature versions). There are already too many parameters everywhere.

go/common/node/sgx.go
kostko: That makes sense to me, no need to have a general map that introduces unnecessary complexity.

martintomazic:
Closed in favor of #6474.

Issue #6459, opened 2026-02-10
Runtime quote verification should use defaults from the consensus params


martintomazic:
Validation on the runtime side is not symmetric due to ignoring [defaults from the consensus parameters](https://github.com/oasisprotocol/oasis-core/blob/7fe1cbbfeaad548d7041d54c37522b223ba07e46/go/common/node/sgx.go#L229:L230) applied on the consensus side. Currently, `TEEFeaturesSGX.DefaultPolicy` is set to `nil` so this is harmless for now.

Do we even see this params getting used anytime in the future or should we deprecate?

For the sake of consistency, there is another [hardcoded policy](https://github.com/oasisprotocol/oasis-core/blob/master/go/common/sgx/pcs/quote.go#L143:L148) that I don't see nowhere in the runtime code.

martintomazic:
I see two options:
1. Verify this parameter was always nil since genesis, drop it and there is no asymmetry left. Implement https://github.com/oasisprotocol/oasis-core/issues/5935#issuecomment-5239461259 in parallel.
2. Use this parameter to set defaut ias policy to disabled. Consensus verifier inside enclave should read mvks authenticated consensus params, and apply them over the runtime policies as is currently done for the Go codebase. Keymanager access policy (once implemented) should also apply those defaults, on both Go and Rust side (see https://github.com/oasisprotocol/oasis-core/pull/6474#discussion_r3744863003).

Both options achieve that IAS corner cases / bypass is not possible structurally and code becomes symmetric.

Issue #6317, opened 2025-09-05
Key manager should use runtime quote policy


kostko:
Currently the key manager runtime takes its own quote policy into account for all incoming EnclaveRPC connections. It should instead support using runtime quote policies in order to allow independent policies (eg. maybe a stricter quote policy for the key manager runtime itself, and more relaxed quote policy for runtimes).

martintomazic:
>Currently the key manager runtime takes its own quote policy into account for all incoming EnclaveRPC connections. It should instead support using per-runtime quote policies in order to allow independent policies...

Initially, I thought that if key-manager serves a secret for a runtime `A`, then all it cares during EnclaveRPC is that client (e.g. compute node) provides attestation that passes SGX constraints defined in the consensus for runtime `A`. This should be safe as secrets are derived per-runtime, moreover runtimes should have freedom (and responsibility) for setting suitable trust.

Looking at #6387 motivation:
>This would allow a more relaxed general policy but stricter requirements for nodes that can access the key manager (e.g. compute/observer nodes).

Would it make sense to have it implemented this way? In fact when I started working on https://github.com/oasisprotocol/oasis-core/pull/6410, this was my initial idea/understanding.

kostko:
That is exactly the idea mentioned in the issue -- eg. the key manager should use the _runtime_ policies when deriving keys for downstream compute runtimes while it currently uses its own policy.

martintomazic:
So this will also make [`MayQuery`](https://github.com/oasisprotocol/oasis-core/blob/3182a76ebb128fa8cf6283988f9e1dd282671b8e/go/keymanager/secrets/policy_sgx.go#L43) field from the key manager's  `EnclavePolicySGX` struct redundant, given this will be now fetched form the consensus sgx constraints? 

My initial understanding (wrong) was that this issues was about adding per-runtime sgx-constraints into the `EnclavePolicySGX` which obviously wouldn't make sense :).

Still we probably need to keep at least runtime ID, to give key-manager owner ability to choose, which runtime it is willing to serve.

Issue #5935, opened 2024-11-14
Remove EPID/IAS attestation support


kostko:
As EPID/IAS will be EOLed in 2025 and no production deployment is using it, we should remove support for attestation and the IAS proxy implementation.

martintomazic:
I suggest to do this now, as current EPID/IAS policy and quote verification is the source of complexity and edge cases. With nil values sometimes using hard-coded defaults, other times bypassing validation etc, relying on the fact that one can no longer create valid IAS bundles or that IAS is explicitly disabled feels error prone.

It would be nice if:
1. After feature flag `26.3` `SGXAttestation` with `ias.AVRBundle` is rejected.
2. After feature flag `26.3` `SGXConstraints` with `ias.QuotePolicy` is rejected or at least we no longer allow registering new descriptors with it  (some runtimes still have both ias and pcs configured but use only pcs, as one cannot create valid IAS bundle anymore).
    * Since this will be breaking change that will be voted on, we might as well expect the runtime owners to migrate in time and make this breaking.
    *  Depending on the approach and the nil semantic we want to have, we might also enforce that new descriptors should have ias policy set as non-nil and disabled.

We could introduce `SGXConstraints` `v2`, and similar `SGXAttestation` `v2` and reject < `v2` after the feature flag so that this is structurally not possible anymore.

Overlaps with https://github.com/oasisprotocol/oasis-core/issues/6459, also see this problem manifest in real code: https://github.com/oasisprotocol/oasis-core/pull/6474#discussion_r3744863003.

PR #6474, opened 2026-03-16
Add support for key manager access quote policy

### LINKED ISSUE #6387: Support for compute runtime quote policy

martintomazic:
Replaces #6471 closes #6387.

### DIFF
--- a/.buildkite/code.pipeline.yml
+++ b/.buildkite/code.pipeline.yml
@@ -357,7 +357,10 @@ steps:
     command:
       - trap 'buildkite-agent artifact upload "coverage-merged-e2e-*.txt;/tmp/e2e/**/*.log;/tmp/e2e/**/genesis.json;/tmp/e2e/**/runtime_genesis.json"' EXIT
       - .buildkite/scripts/download_e2e_test_artifacts_mocksgx.sh
-      - .buildkite/scripts/test_e2e.sh --timeout 20m --scenario e2e/runtime/runtime-encryption
+      - .buildkite/scripts/test_e2e.sh --timeout 20m
+        --scenario e2e/runtime/runtime-encryption
+        --scenario e2e/runtime/keymanager-access-policy
+
     env:
       OASIS_TEE_HARDWARE: intel-sgx
       OASIS_UNSAFE_MOCK_TEE: "1"
--- a/go/common/node/node.go
+++ b/go/common/node/node.go
@@ -549,6 +549,27 @@ func (h *TEEHardware) FromString(str string) error {
 	return nil
 }
 
+// CapabilityTEEVerifyParams contains parameters required to verify a TEE attestation.
+type CapabilityTEEVerifyParams struct {
+	// Features are the TEE features and defaults advertised by the consensus layer.
+	Features *TEEFeatures
+
+	// Time is the current consensus time.
+	Time time.Time
+
+	// Height is the current consensus height.
+	Height uint64
+
+	// Constraints are the serialized TEE constraints.
+	Constraints []byte
+
+	// NodeID is the node identity the TEE capability must be bound to.
+	NodeID signature.PublicKey
+
+	// IsFeatureVersion261 is true for consensus at version 26.1 or higher.
+	IsFeatureVersion261 bool
+}
+
 // CapabilityTEE represents the node's TEE capability.
 type CapabilityTEE struct {
 	// TEE hardware type.
@@ -565,7 +586,7 @@ type CapabilityTEE struct {
 }
 
 // Verify verifies the node's TEE capabilities, at the provided timestamp and height.
-func (c *CapabilityTEE) Verify(teeCfg *TEEFeatures, ts time.Time, height uint64, constraints []byte, nodeID signature.PublicKey, isFeatureVersion261 bool) error {
+func (c *CapabilityTEE) Verify(params CapabilityTEEVerifyParams) error {
 	switch c.Hardware {
 	case TEEHardwareIntelSGX:
 		// Parse SGX remote attestation.
@@ -573,21 +594,29 @@ func (c *CapabilityTEE) Verify(teeCfg *TEEFeatures, ts time.Time, height uint64,
 		if err := cbor.Unmarshal(c.Attestation, &sa); err != nil {
 			return fmt.Errorf("node: malformed SGX attestation: %w", err)
 		}
-		if err := sa.ValidateBasic(teeCfg); err != nil {
+		if err := sa.ValidateBasic(params.Features); err != nil {
 			return fmt.Errorf("node: malformed SGX attestation: %w", err)
 		}
 
 		// Parse SGX constraints.
 		var sc SGXConstraints
-		if err := cbor.Unmarshal(constraints, &sc); err != nil {
+		if err := cbor.Unmarshal(params.Constraints, &sc); err != nil {
 			return fmt.Errorf("node: malformed SGX constraints: %w", err)
 		}
-		if err := sc.ValidateBasic(teeCfg, isFeatureVersion261); err != nil {
+		if err := sc.ValidateBasic(params.Features, params.IsFeatureVersion261); err != nil {
 			return fmt.Errorf("node: malformed SGX constraints: %w", err)
 		}
 
 		// Verify SGX remote attestation.
-		return sa.Verify(teeCfg, ts, height, &sc, c.RAK, c.REK, nodeID)
+		return sa.Verify(
+			params.Features,
+			params.Time,
+			params.Height,
+			&sc,
+			c.RAK,
+			c.REK,
+			params.NodeID,
+		)
 	default:
 		return ErrInvalidTEEHardware
 	}
--- a/go/common/node/node.go
+++ b/go/common/node/node.go
@@ -14,7 +14,6 @@ import (
 	beacon "github.com/oasisprotocol/oasis-core/go/beacon/api"
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
-	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
 	"github.com/oasisprotocol/oasis-core/go/common/prettyprint"
 	"github.com/oasisprotocol/oasis-core/go/common/version"
@@ -563,14 +562,6 @@ type CapabilityTEE struct {
 
 	// Attestation.
 	Attestation []byte `json:"attestation"`
-}
-
-// HashRAK computes the expected report data hash bound to a given public RAK.
-func HashRAK(rak signature.PublicKey) hash.Hash {
-	hData := make([]byte, 0, len(teeHashContext)+signature.PublicKeySize)
-	hData = append(hData, teeHashContext...)
-	hData = append(hData, rak[:]...)
-	return hash.NewFromBytes(hData)
 }
 
 // Verify verifies the node's TEE capabilities, at the provided timestamp and height.
--- a/go/common/node/sgx.go
+++ b/go/common/node/sgx.go
@@ -286,6 +286,14 @@ func (sa *SGXAttestation) verifyAttestationSignature(
 	return nil
 }
 
+// HashRAK computes the expected report data hash bound to a given public RAK.
+func HashRAK(rak signature.PublicKey) hash.Hash {
+	hData := make([]byte, 0, len(teeHashContext)+signature.PublicKeySize)
+	hData = append(hData, teeHashContext...)
+	hData = append(hData, rak[:]...)
+	return hash.NewFromBytes(hData)
+}
+
 // HashAttestation hashes the required data that needs to be signed by RAK producing the attestation
 // signature. The hash is computed as follows:
 //
--- a/go/common/node/sgx.go
+++ b/go/common/node/sgx.go
@@ -32,8 +32,20 @@ type SGXConstraints struct {
 	// Enclaves is the allowed MRENCLAVE/MRSIGNER pairs.
 	Enclaves []sgx.EnclaveIdentity `json:"enclaves,omitempty"`
 
-	// Policy is the quote policy.
+	// Policy is the default quote policy.
 	Policy *quote.Policy `json:"policy,omitempty"`
+
+	// KeyManagerAccessPolicy is the compute runtime quote policy for the enclaves,
+	// requesting runtime key derivation from the key manager.
+	//
+	// If not nil, it replaces the default quote policy for compute and observer nodes.
+	//
+	// NOTE: Enforced directly during compute/observer registration and committee election.
+	// Key manager EnclaveRPC access is gated transitively through registered runtime membership.
+	// Key manager clients must still satisfy the key manager runtime's own quote policy. Making
+	// the key manager runtime resolve caller runtime policies itself is tracked in:
+	// https://github.com/oasisprotocol/oasis-core/issues/6317
+	KeyManagerAccessPolicy *quote.Policy `json:"kma_policy,omitempty"`
 
 	// MaxAttestationAge is the maximum attestation age (in blocks).
 	MaxAttestationAge uint64 `json:"max_attestation_age,omitempty"`
@@ -111,21 +123,53 @@ func (sc *SGXConstraints) ValidateBasic(cfg *TEEFeatures, isFeatureVersion261 bo
 		return fmt.Errorf("unsupported SGX constraints version: %d", sc.V)
 	}
 
-	if sc.Policy == nil {
-		return nil
+	validatePolicy := func(policy *quote.Policy) error {
+		if policy == nil {
+			return nil
+		}
+
+		// Check for TDX enablement.
+		if !cfg.SGX.TDX && policy.PCS != nil && policy.PCS.TDX != nil {
+			return fmt.Errorf("TDX policy not supported")
+		}
+
+		// Check that policy is compliant with the current feature version.
+		return policy.Validate(isFeatureVersion261)
 	}
 
-	// Check for TDX enablement.
-	if !cfg.SGX.TDX && sc.Policy.PCS != nil && sc.Policy.PCS.TDX != nil {
-		return fmt.Errorf("TDX policy not supported")
+	validateKMAPolicy := func(policy *quote.Policy) error {
+		if policy == nil {
+			return nil
+		}
+		if !isFeatureVersion261 {
+			return fmt.Errorf("policy should be nil")
+		}
+		if policy.IAS != nil {
+			return fmt.Errorf("IAS not allowed")
+		}
+		return validatePolicy(policy)
 	}
 
-	// Check that policy is compliant with the current feature version.
-	if err := sc.Policy.Validate(isFeatureVersion261); err != nil {
+	if err := validatePolicy(sc.Policy); err != nil {
 		return fmt.Errorf("invalid policy: %w", err)
 	}
 
+	if err := validateKMAPolicy(sc.KeyManagerAccessPolicy); err != nil {
+		return fmt.Errorf("invalid key manager access policy: %w", err)
+	}
+
 	return nil
+}
+
+// ResolvePolicy returns the effective policy.
+//
+// TODO: Handle default constraints overwrite consistently.
+// See https://github.com/oasisprotocol/oasis-core/issues/6459.
+func (sc *SGXConstraints) ResolvePolicy(useKMAPolicy bool) *quote.Policy {
+	if sc.KeyManagerAccessPolicy != nil && useKMAPolicy {
+		return sc.KeyManagerAccessPolicy
+	}
+	return sc.Policy
 }
 
 // ContainsEnclave returns true iff the allowed enclave list in SGX constraints contain the given
@@ -229,10 +273,17 @@ func (sa *SGXAttestation) Verify(
 	}
 
 	// Use defaults from consensus parameters.
+	// TODO: Handle default constraints overwrite consistently.
+	// See https://github.com/oasisprotocol/oasis-core/issues/6459.
 	cfg.SGX.ApplyDefaultConstraints(sc)
 
+	// Verify function is only called for compute and keymanager nodes (empty keymanager access policy),
+	// therefore we always set it to true.
+	useKMAPolicy := true
+	policy := sc.ResolvePolicy(useKMAPolicy)
+
 	// Verify the quote.
-	verifiedQuote, err := sa.Quote.Verify(sc.Policy, ts)
+	verifiedQuote, err := sa.Quote.Verify(policy, ts)
 	if err != nil {
 		return err
 	}
--- a/go/common/node/sgx.go
+++ b/go/common/node/sgx.go
@@ -32,8 +32,14 @@ type SGXConstraints struct {
 	// Enclaves is the allowed MRENCLAVE/MRSIGNER pairs.
 	Enclaves []sgx.EnclaveIdentity `json:"enclaves,omitempty"`
 
-	// Policy is the quote policy.
+	// Policy is the default quote policy.
 	Policy *quote.Policy `json:"policy,omitempty"`
+
+	// KeyManagerAccessPolicy is the compute runtime quote policy for the enclaves,
+	// requesting runtime key derivation from the key manager.
+	//
+	// If not nil, it replaces the default quote policy for compute and observer nodes.
+	KeyManagerAccessPolicy *quote.Policy `json:"kma_policy,omitempty"`
 
 	// MaxAttestationAge is the maximum attestation age (in blocks).
 	MaxAttestationAge uint64 `json:"max_attestation_age,omitempty"`
@@ -111,21 +117,50 @@ func (sc *SGXConstraints) ValidateBasic(cfg *TEEFeatures, isFeatureVersion261 bo
 		return fmt.Errorf("unsupported SGX constraints version: %d", sc.V)
 	}
 
-	if sc.Policy == nil {
-		return nil
+	validatePolicy := func(policy *quote.Policy) error {
+		if policy == nil {
+			return nil
+		}
+
+		// Check for TDX enablement.
+		if !cfg.SGX.TDX && policy.PCS != nil && policy.PCS.TDX != nil {
+			return fmt.Errorf("TDX policy not supported")
+		}
+
+		// Check that policy is compliant with the current feature version.
+		return policy.Validate(isFeatureVersion261)
 	}
 
-	// Check for TDX enablement.
-	if !cfg.SGX.TDX && sc.Policy.PCS != nil && sc.Policy.PCS.TDX != nil {
-		return fmt.Errorf("TDX policy not supported")
+	validateKMAPolicy := func(policy *quote.Policy) error {
+		if policy == nil {
+			return nil
+		}
+		if !isFeatureVersion261 {
+			return fmt.Errorf("policy should be nil")
+		}
+		if policy.IAS != nil {
+			return fmt.Errorf("IAS not allowed")
+		}
+		return validatePolicy(policy)
 	}
 
-	// Check that policy is compliant with the current feature version.
-	if err := sc.Policy.Validate(isFeatureVersion261); err != nil {
+	if err := validatePolicy(sc.Policy); err != nil {
 		return fmt.Errorf("invalid policy: %w", err)
 	}
 
+	if err := validateKMAPolicy(sc.KeyManagerAccessPolicy); err != nil {
+		return fmt.Errorf("invalid key manager access policy: %w", err)
+	}
+
 	return nil
+}
+
+// ResolvePolicy returns the effective policy.
+func (sc *SGXConstraints) ResolvePolicy(useKMAPolicy bool) *quote.Policy {
+	if sc.KeyManagerAccessPolicy != nil && useKMAPolicy {
+		return sc.KeyManagerAccessPolicy
+	}
+	return sc.Policy
 }
 
 // ContainsEnclave returns true iff the allowed enclave list in SGX constraints contain the given
@@ -229,10 +264,14 @@ func (sa *SGXAttestation) Verify(
 	}
 
 	// Use defaults from consensus parameters.
+	// TODO: Handle default constraints overwrite consistently.
+	// See https://github.com/oasisprotocol/oasis-core/issues/6459.
 	cfg.SGX.ApplyDefaultConstraints(sc)
 
+	policy := sc.ResolvePolicy(true)
+
 	// Verify the quote.
-	verifiedQuote, err := sa.Quote.Verify(sc.Policy, ts)
+	verifiedQuote, err := sa.Quote.Verify(policy, ts)
 	if err != nil {
 		return err
 	}
--- a/go/common/node/sgx.go
+++ b/go/common/node/sgx.go
@@ -144,8 +144,11 @@ func (sc *SGXConstraints) ValidateBasic(cfg *TEEFeatures, isFeatureVersion261 bo
 		if !isFeatureVersion261 {
 			return fmt.Errorf("policy should be nil")
 		}
-		if policy.IAS != nil {
-			return fmt.Errorf("IAS not allowed")
+
+		// TODO: Remove IAS support entirely.
+		// See https://github.com/oasisprotocol/oasis-core/issues/5935.
+		if policy.IAS == nil || !policy.IAS.Disabled {
+			return fmt.Errorf("IAS must be disabled")
 		}
 		return validatePolicy(policy)
 	}
--- a/go/common/node/sgx.go
+++ b/go/common/node/sgx.go
@@ -39,6 +39,12 @@ type SGXConstraints struct {
 	// requesting runtime key derivation from the key manager.
 	//
 	// If not nil, it replaces the default quote policy for compute and observer nodes.
+	//
+	// NOTE: Enforced directly during compute/observer registration and committee election.
+	// Key manager EnclaveRPC access is gated transitively through registered runtime membership.
+	// Key manager clients must still satisfy the key manager runtime's own quote policy. Making
+	// the key manager runtime resolve caller runtime policies itself is tracked in:
+	// https://github.com/oasisprotocol/oasis-core/issues/6317
 	KeyManagerAccessPolicy *quote.Policy `json:"kma_policy,omitempty"`
 
 	// MaxAttestationAge is the maximum attestation age (in blocks).
--- a/go/oasis-node/cmd/node/node.go
+++ b/go/oasis-node/cmd/node/node.go
@@ -199,7 +199,16 @@ func (n *Node) initRuntimeWorkers(genesisDoc *genesisAPI.Document) error {
 	}
 
 	// Initialize runtime provisioner.
-	policyProvider := &runtimeRegistry.QuotePolicyProvider{Consensus: n.Consensus}
+	var useKMAPolicy bool
+	switch config.GlobalConfig.Mode {
+	case config.ModeCompute, config.ModeObserver:
+		useKMAPolicy = true
+	default:
+	}
+	policyProvider := &runtimeRegistry.QuotePolicyProvider{
+		Consensus:    n.Consensus,
+		UseKMAPolicy: useKMAPolicy,
+	}
 	var err error
 	n.Provisioner, err = provisioner.New(
 		n.dataDir,
--- a/go/oasis-node/cmd/node/node.go
+++ b/go/oasis-node/cmd/node/node.go
@@ -212,9 +212,19 @@ func (n *Node) initRuntimeWorkers(genesisDoc *genesisAPI.Document) error {
 	}
 	n.svcMgr.Register(n.RuntimeRegistry)
 
+	// Determine whether hosted RONL components will be registered on the
+	// consensus layer.
+	var willRegisterComputeRuntime bool
+	switch config.GlobalConfig.Mode {
+	case config.ModeCompute, config.ModeObserver:
+		willRegisterComputeRuntime = true
+	default:
+	}
+
 	// Initialize the common worker.
 	commonCfg := workerCommon.Config{
-		TxPool: config.GlobalConfig.Runtime.TxPool,
+		TxPool:                     config.GlobalConfig.Runtime.TxPool,
+		WillRegisterComputeRuntime: willRegisterComputeRuntime,
 	}
 	n.CommonWorker, err = workerCommon.New(
 		commonCfg,
--- a/go/oasis-test-runner/scenario/e2e/runtime/kma_policy.go
+++ b/go/oasis-test-runner/scenario/e2e/runtime/kma_policy.go
@@ -1,0 +1,54 @@
+package runtime
+
+import (
+	"context"
+
+	"github.com/oasisprotocol/oasis-core/go/common/sgx/pcs"
+	"github.com/oasisprotocol/oasis-core/go/common/sgx/quote"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/env"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/oasis"
+	"github.com/oasisprotocol/oasis-core/go/oasis-test-runner/scenario"
+)
+
+// KeyManagerAccessPolicy is the key manager access policy e2e test scenario.
+var KeyManagerAccessPolicy scenario.Scenario = newKeyManagerAccessPolicyImpl()
+
+type keyManagerAccessPolicyImpl struct {
+	Scenario
+}
+
+func newKeyManagerAccessPolicyImpl() scenario.Scenario {
+	return &keyManagerAccessPolicyImpl{
+		Scenario: *NewScenario("keymanager-access-policy", NewTestClient().WithScenario(SimpleScenario)),
+	}
+}
+
+func (sc *keyManagerAccessPolicyImpl) Clone() scenario.Scenario {
+	return &keyManagerAccessPolicyImpl{
+		Scenario: *sc.Scenario.Clone().(*Scenario),
+	}
+}
+
+func (sc *keyManagerAccessPolicyImpl) Fixture() (*oasis.NetworkFixture, error) {
+	f, err := sc.Scenario.Fixture()
+	if err != nil {
+		return nil, err
+	}
+
+	f.Runtimes[1].KeyManagerAccessPolicy = &quote.Policy{
+		PCS: &pcs.QuotePolicy{
+			TCBValidityPeriod:          90,
+			MinTCBEvaluationDataNumber: 12,
+		},
+	}
+
+	return f, nil
+}
+
+func (sc *keyManagerAccessPolicyImpl) Run(ctx context.Context, childEnv *env.Env) error {
+	if err := sc.StartNetworkAndTestClient(ctx, childEnv); err != nil {
+		return err
+	}
+
+	return sc.WaitTestClientAndCheckLogs()
+}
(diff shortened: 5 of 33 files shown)


.buildkite/code.pipeline.yml
```diff
@@ -356,7 +356,10 @@ steps:
     command:
       - trap 'buildkite-agent artifact upload "coverage-merged-e2e-*.txt;/tmp/e2e/**/*.log;/tmp/e2e/**/genesis.json;/tmp/e2e/**/runtime_genesis.json"' EXIT
       - .buildkite/scripts/download_e2e_test_artifacts_mocksgx.sh
-      - .buildkite/scripts/test_e2e.sh --timeout 20m --scenario e2e/runtime/runtime-encryption
+      - .buildkite/scripts/test_e2e.sh --timeout 20m
+        --scenario e2e/runtime/runtime-encryption
+        --scenario e2e/runtime/compute-policy
+
```
martintomazic: e2e test with mocked TEE was instrumental to realizing that without  `compute_policy` on the rust side of the protocol, RHP would fail due to the missing field.

To test that compute policy is applied e2e you can disable pcs in the quote policy (fixture) and run:
```export OASIS_UNSAFE_SKIP_AVR_VERIFY=1
export OASIS_UNSAFE_ALLOW_DEBUG_ENCLAVES=1
export OASIS_UNSAFE_MOCK_TEE=1
export OASIS_TEE_HARDWARE=intel-sgx

make
.buildkite/scripts/test_e2e.sh --scenario e2e/runtime/compute-policy
```

The test would gets stuck (as expected).

Full SGX test would be desired here (easy). It would be also nice to add another one, where we e.g. set `MinTCBEvaluationDataNumber` and verify via the control status that the registration failed... Still not optimal as technically we would catch node local attestation to halt early and not consensus rejecting byzantine compute.

Testing that rofl nodes pass on the lower TCB but compute node is rejected would be even harder as we would need two different SGX environments to make it work.

_What do you think?_

go/oasis-node/cmd/node/node.go
```diff
@@ -205,12 +205,25 @@ func (n *Node) initRuntimeWorkers(genesisDoc *genesisAPI.Document) error {
...
+		}
+	}
+
 	// Initialize the common worker.
 	n.CommonWorker, err = workerCommon.New(
 		n,
 		n.dataDir,
 		n.chainContext,
 		n.Identity,
+		willRegisterComputeRuntime,
```
martintomazic: Previously I was passing runtime roles ([see](https://github.com/oasisprotocol/oasis-core/pull/6471/changes/efbc5bc807436471498b489ed46258eba255fd4a)). Now we only need bool flag.

No matter what name I choose here, passing the variable through workers, feels like abstraction leak.

I could avoid all this and use the `config.GlobalConfig` directly in the `common.GetQuotePolicy`. I hate global state though, but here might be acceptable with a TODO issue given this shows abstractions are not best in the first place?

runtime/src/consensus/registry.rs
```diff
@@ -703,6 +703,10 @@ pub enum SGXConstraints {
         #[cbor(optional)]
         policy: sgx::QuotePolicy,
 
+        /// The compute runtime quote policy.
+        #[cbor(optional)]
+        compute_policy: sgx::QuotePolicy,
```
martintomazic: This field is ignored for now and only here to pass e2e test. 

In practice we probably need new RHP method (similar to node identity), to signal whether compute policy should be respected or not. Technically not breaking?

The runtime should sign with its RAK whether the intention of this enclave is to be used as the one registering on the consensus (having access to keymanager) or is the intention to be used in the less constrained environments (default policy only). Similar solution to what is done for the `node_id`, i.e. make it part of the hash structure and sign with RAK and validate signature on the consensus side.
_Do we really need this / what are possible attack vectors without it?_

.buildkite/code.pipeline.yml
kostko: Yes, it would be great if we can test it in real SGX.

For the two different environments, let's open a separate issue to test that E2E.

go/common/node/node.go
```diff
@@ -573,30 +573,64 @@ func HashRAK(rak signature.PublicKey) hash.Hash {
 	return hash.NewFromBytes(hData)
 }
 
+// CapabilityTEEVerifyParams contains parameters required to verify a TEE attestation.
+type CapabilityTEEVerifyParams struct {
+	// Features are the TEE features and defaults advertised by the consensus layer.
+	Features *TEEFeatures
+
+	// Timestamp is the current consensus time.
+	Timestamp time.Time
```
kostko: Let's call this `Now` or similar.

go/common/node/sgx.go
```diff
@@ -32,9 +32,13 @@ type SGXConstraints struct {
 	// Enclaves is the allowed MRENCLAVE/MRSIGNER pairs.
 	Enclaves []sgx.EnclaveIdentity `json:"enclaves,omitempty"`
 
-	// Policy is the quote policy.
+	// Policy is the default quote policy.
 	Policy *quote.Policy `json:"policy,omitempty"`
 
+	// ComputePolicy is additional compute runtime policy that replaces (default) Policy
+	// if not nil.
```
kostko: Make it more clear that this only replaces the policy for compute/observer nodes.

go/oasis-node/cmd/node/node.go
kostko: Yes, this seems awkward. For example the common worker constructor is already doing config lookups so it seems fine to have them there.

Removing the use of global config and explicitly passing it instead is a different issue.

runtime/src/consensus/registry.rs
kostko: I would avoid this extra verification for now as I don't see what attack this would prevent. It is other runtimes and the consensus layer that are enforcing this policy.

go/oasis-node/cmd/node/node.go
martintomazic: #6481 followed by #6489 will simply this. The current code was also off a bit as it ignored debug test entity set via env variable.

.buildkite/code.pipeline.yml
martintomazic: > Yes, it would be great if we can test it in real SGX.

Done https://github.com/oasisprotocol/oasis-core/pull/6474/changes/9675e140de159c551e4e2ce5314b81e9473fbc0c :).

I had to comment `UNSAFE_SKIP_AVR_VERIFY` which caused most of the policy verification to be ignored ([see](https://github.com/oasisprotocol/oasis-core/blob/59b8f325ffd4d3ab4455b0495b78552a7b4d739e/go/common/sgx/pcs/quote.go#L196)) and set a more relaxed verification because our runner has outdated TCB.

My suggestion is to add another env variable: `OASIS_UNSAFE_SKIP_LOCAL_VERIFY`, which skips local verification prior to submitting attestation via transaction. Then I can verify from the node control status consensus rejected attestation for the right reason. The test is currently stopped at the local verification which is not ideal given this is critical code path that we want to have tested.

```bash
{"caller":"service.go:122","err":"quote verification failed (fresh bundle): pcs/quote: failed to verify TCB bundle: pcs/tcb: failed to verify TCB info: pcs/tcb: invalid TCB info: pcs/tcb: FMSPC is not whitelisted","level":"warn","module":"common/sgx/pcs/cqs","msg":"error verifying downloaded TCB refresh","tcb_evaluation_data_number":21,"ts":"2026-03-30T14:04:50.6645918Z"}
```


Optional:
- I have happy and unhappy path. If we whitelist `InvalidFMSCP` under the default policy and the one the current runner is using under compute policy, this will also verify FMSCP whitelist e2e.
- We can also play with `MinTCBEvaluationDataNumber`, possibly downgrade it to make current runner pass without `LAX_VERIFY`.

martintomazic: `OASIS_UNSAFE_SKIP_LOCAL_VERIFY`: technically there is nothing unsafe about it as this is local sanity check prior to _trying_ to register.

martintomazic: >My suggestion is to add another env variable: OASIS_UNSAFE_SKIP_LOCAL_VERIFY, which skips local verification prior to submitting attestation via transaction.

We also have RHP, which currently passes with empty default policy (used there) and relaxed TCB level. So this flag will not be that useful for other tests in the future, unless we rely on the default/compute policy.

In practice this variable should gate:
* https://github.com/oasisprotocol/oasis-core/blob/59b8f325ffd4d3ab4455b0495b78552a7b4d739e/go/common/sgx/pcs/service.go#L118
* https://github.com/oasisprotocol/oasis-core/blob/59b8f325ffd4d3ab4455b0495b78552a7b4d739e/runtime/src/identity.rs#L273

It cannot live inside those functions as it should only apply for local verification, but should NOT pass remote verification, unless specified by other environment variables. The easiest way would be to add another parameter to  both `verify` functions specifying if verification should be ignored, in addition to internal global state flags (e.g. [see]( https://github.com/oasisprotocol/oasis-core/blob/59b8f325ffd4d3ab4455b0495b78552a7b4d739e/runtime/src/common/sgx/pcs/quote.rs#L46)) that currently define whether verification should be ignored/relaxed regardless whether this is local or remote/consensus verification.

runtime/src/consensus/registry.rs
martintomazic: Agree as long as we do 2 side verification on both enclave_rpc/consensus side (confirmed myself again) exploiting this is not possible. Moreover,  currently the keymanager policy must be respected on both sides when interacting with it.

If however, at an point we change keymanager-RONL interaction, so that RONL must only satisfy its policy when querying keymanager for its secrets, we should be careful, not to accidentally use the default one, especially if we introduce remote RONLs. The problem there would be if the ROFL would want stricter policy then the default RONL, but again this would be mitigated by the enclave rpc using compute policy.

Having ROFL requiring stricter policy than the keymanager/compute nodes wouldn't make sense given it depends on them directly, so ignoring this corner case.

.buildkite/code.pipeline.yml
kostko: I don't like this additional variable as it seems it would introduce additional complexity around checks. I would skip it for now.

go/common/node/node.go
```diff
@@ -573,30 +573,64 @@ func HashRAK(rak signature.PublicKey) hash.Hash {
...
 			return fmt.Errorf("node: malformed SGX constraints: %w", err)
 		}
-		if err := sc.ValidateBasic(teeCfg, isFeatureVersion242); err != nil {
+		if err := sc.ValidateBasic(params.Features, params.IsFeatureVersion242); err != nil {
 			return fmt.Errorf("node: malformed SGX constraints: %w", err)
 		}
 
 		// Verify SGX remote attestation.
-		return sa.Verify(teeCfg, ts, height, &sc, c.RAK, c.REK, nodeID)
+		return sa.Verify(
```
kostko, line 624: Could you just pass `params` here instead of so many parameters?

go/common/node/sgx.go
```diff
@@ -32,9 +32,14 @@ type SGXConstraints struct {
 	// Enclaves is the allowed MRENCLAVE/MRSIGNER pairs.
 	Enclaves []sgx.EnclaveIdentity `json:"enclaves,omitempty"`
 
-	// Policy is the quote policy.
+	// Policy is the default quote policy.
 	Policy *quote.Policy `json:"policy,omitempty"`
 
+	// ComputePolicy is additional compute runtime quote policy.
+	//
+	// If not nil, it replaces (default) Policy for the compute and observer nodes.
```
kostko:
```suggestion
	// ComputePolicy is the compute runtime quote policy.
	//
	// If not nil, it replaces the default quote policy for compute and observer nodes.
```

go/common/node/sgx.go
```diff
@@ -32,9 +32,14 @@ type SGXConstraints struct {
...
+	ComputePolicy *quote.Policy `json:"compute_policy,omitempty"`
```
kostko, line 41: Not sure if "compute policy" is the best name for this given that both policies apply to compute runtimes, just different levels of access. Any other suggestions?

go/common/node/node.go
martintomazic: I would avoid that because they are two different functions, i.e. `sa.Verify` uses parsed constrains, whereas the `c.Verify` doe not. 

Since most of the params is shared this could be solved by a shared struct?  `sa.Verify` is also called only once  inside same package so not sure it has to be public/having clean params is that critical.

Additional smells I see:
1.  _`sa.Verify` should not be responsible for applying default constrains, instead `c.Verify` should apply them, before passing._
2.  By passing `allowSignedAttestations` and **1.** resolved, `sa.Verify` no longer needs `teeCfg`, although other `ValidateBasic` methods accept it so questionable.
3. Does `c.Verify` need to call `sc.ValidateBasic` as consensus already validate this (could omit `isFeatureVersion`)?
4. Go side of verification only calls this from the consensus, that ensures keymanger policy is always empty, in that sense `useComputePolicy` is actually redundant. The rust side during RHP initialization only verifies against the default policy and relies on the consensus/enclave rpc verification there instead.

go/common/node/sgx.go
martintomazic: I am not 100% fond of the name either.

`StrictPolicy`? and `useStrictPolicy` to configure runtime provisioner early validation? 

Cons: now is less clearer why keymanager cannot have strict policy, also strictness is not actually enforced, e.g. with current code you can actually set looser policy for the compute one (but that would be violation of the intent making this name more suitable).

go/common/node/node.go
kostko: Ok.

martintomazic: >Go side of verification only calls this from the consensus, that ensures keymanger policy is always empty, in that sense useComputePolicy is actually redundant.

The reason I added is because one day someone may want to verify in the non-strict mode from the go side as well.

go/oasis-node/cmd/node/node.go
martintomazic: This is now simpler, but kept passing this variable as will help us finding a better name as you suggested https://github.com/oasisprotocol/oasis-core/pull/6474#discussion_r3059395354.

The early local verification could also simply use just the default policy (and we rely on the consensus catching strict one), which would also enable writting e2e test with unhappy path. Imo also not best, so closing for now.

.buildkite/code.pipeline.yml
martintomazic: Removed this for now as I agree this adds too much complexity.

One option would be to only use default policy during local validation, like is currently done for RHP, as consensus would catch wrong attestation anyways. This would simplify our code further (removing the need for ff833fcf5c6d4319da0490a17efb6daf90012d76) and enable us to keep previous unhappy [test](https://github.com/oasisprotocol/oasis-core/blob/7466948cac1fdc33b068195059a106fbb1e65c5f/go/oasis-test-runner/scenario/e2e/runtime/compute_policy_rejects.go). On the other hand this may cause honest nodes to post invalid registration transactions. Also not most robust/maintainable in the long run but it achieves the goal.

 If anyone sees a clean way to test unhappy path is rejected, I am willing to implement it here.

go/common/node/sgx.go
kostko: This is basically the policy for requesting runtime key derivation from the key manager.

So maybe something related to that? KeyManagerAccessPolicy (`kma_policy`)?

go/oasis-node/cmd/node/node.go
```diff
@@ -232,6 +241,7 @@ func (n *Node) initRuntimeWorkers(genesisDoc *genesisAPI.Document) error {
 		n.Consensus.KeyManager(),
 		n.RuntimeRegistry,
 		n.Provisioner,
+		willRegisterComputeRuntime,
```
kostko, line 244: Should this be in the `commonCfg` instead of being a separate argument?

go/worker/common/committee/node.go
```diff
@@ -471,7 +475,8 @@ func (n *Node) worker() { //nolint: gocyclo
 			switch {
 			case compNotify.Added != nil:
 				// Received a new version of a runtime component.
-				if err := n.ProvisionHostedRuntimeComponent(compNotify.Added); err != nil {
+				attestCfg := n.attestationCfg(compNotify.Added)
+				if err := n.ProvisionHostedRuntimeComponent(compNotify.Added, attestCfg); err != nil {
```
martintomazic: NIT: Not every component is TEE enabled, so maybe attestCfg should be optional here if `comp.TEEKind != component.TEEKindNone`...

martintomazic (commented):
Should be ready for a final review.
- Added one more check: runtime without key manager cannot have kma_policy.
- Removed `UseKMAPolicy` verification flag (tentative).
- Note CI is currently failing because of runners running out of disk space (so ignore this here).

runtime/src/consensus/registry.rs
martintomazic: I guess this willl used/done as part of the https://github.com/oasisprotocol/oasis-core/issues/6317 then.

go/worker/common/committee/node.go
```diff
@@ -698,13 +716,13 @@ func NewNode(
...
 	// Prepare transaction pool.
-	n.TxPool = txpool.New(runtime.ID(), txPoolCfg, rhn.GetHostedRuntime(), runtime.History(), n)
+	n.TxPool = txpool.New(runtime.ID(), cfg.TxPool, rhn.GetHostedRuntime(), runtime.History(), n)
 
 	// Register transaction message handler as that is something that all workers must handle.
 	p2pHost.RegisterHandler(txTopic, &txMsgHandler{n})
 
 	// Register transaction sync service.
-	p2pHost.RegisterProtocolServer(txsync.NewServer(chainContext, runtime.ID(), n.TxPool))
+	p2pHost.RegisterProtocolServer(txsync.NewServer(n.ChainContext, runtime.ID(), n.TxPool))
```
kostko: Seems strange that somewhere you are using `cfg.ChainContext`/`cfg.TxPool` but elsewhere `n.ChainContext`/`n.TxPool`. I would unify these.

kostko (commented):
Not part of this issue, but we need to change the 24.2 (242) upgrade identifier to something else like 26.1 as currently multiple previous versions have the same handler.

go/worker/common/committee/node.go
martintomazic: using `cfg.ChainContext` now consistently in the `new` method. Notice `cfg.TxPool` is a config but `n.TxPool` is and tx pool instance.

kostko (approved):

go/common/node/node.go
```diff
@@ -573,30 +573,59 @@ func HashRAK(rak signature.PublicKey) hash.Hash {
 	return hash.NewFromBytes(hData)
 }
 
+// CapabilityTEEVerifyParams contains parameters required to verify a TEE attestation.
+type CapabilityTEEVerifyParams struct {
```
peternose, line 551: The location of this struct is unfortunate, as it splits `CapabilityTEE` and its functions. Therefore, I would move it up or down. Same holds for the existing function `HashRAK`.

go/common/node/node.go
```diff
@@ -573,30 +573,59 @@ func HashRAK(rak signature.PublicKey) hash.Hash {
...
+	// Features are the TEE features and defaults advertised by the consensus layer.
+	Features *TEEFeatures
+
+	// Now is the current consensus time.
+	Now time.Time
```
peternose: I would rename this to `Time` or `Timestamp`, as you never can be sure if `Now` is really now, and `Params` to `State` or `Env`, as otherwise is too general, and reorganize fields (node fields, deployment fields, consensus fields), e.g.:
```go


// CapabilityTEEVerifyState contains ....
type CapabilityTEEVerifyState struct {
	// Node is ...
	Node NodeState

	// Deployment is ...
	Deployment DeploymentState

	// Consensus is ...
	Consensus ConsensusState
}

type NodeState struct {
	// ID is the node identifier.
	ID signature.PublicKey
}

// DeploymentState contains the deployment state relevant for TEE attestation
// verification.
type DeploymentState struct {
	// Constraints are the serialized TEE constraints.
	Constraints []byte
}

// ConsensusState contains the consensus state relevant for TEE attestation.
type ConsensusState struct {
	// Height is the current consensus height.
	Height uint64

	// Time is the current consensus time.
	Time time.Time

	// IsFeatureVersion261 is true for consensus at version 26.1 or higher.
	IsFeatureVersion261 bool

	// Features are the TEE features and defaults advertised by the consensus layer.
	Features *TEEFeatures
}
```

go/common/node/node.go
```diff
@@ -573,30 +573,59 @@ func HashRAK(rak signature.PublicKey) hash.Hash {
...
+
+	// Height is the current consensus height.
+	Height uint64
+
+	// Constraints are the serialized TEE constraints.
+	Constraints []byte
+
+	// NodeID is the node identity the attestation must be bound to.
```
peternose: I would not use word `attestation` as that is `TEEHardwareIntelSGX` specific.

go/common/node/sgx.go
```diff
@@ -111,23 +117,52 @@ func (sc *SGXConstraints) ValidateBasic(cfg *TEEFeatures, isFeatureVersion261 bo
...
-	if !cfg.SGX.TDX && sc.Policy.PCS != nil && sc.Policy.PCS.TDX != nil {
-		return fmt.Errorf("TDX policy not supported")
+	validateKMAPolicy := func(policy *quote.Policy) error {
+		if policy == nil {
+			return nil
+		}
+		if !isFeatureVersion261 {
+			return fmt.Errorf("policy should be nil")
+		}
+		if policy.IAS != nil {
```
peternose: Do we need this? If we keep this, after we upgrade, we cannot remove feature version check and use the same validation function for both policies.

go/common/node/sgx.go
```diff
@@ -111,23 +117,52 @@ func (sc *SGXConstraints) ValidateBasic(cfg *TEEFeatures, isFeatureVersion261 bo
...
+			return fmt.Errorf("IAS not allowed")
+		}
+		return validatePolicy(policy)
+	}
+
+	if err := validatePolicy(sc.Policy); err != nil {
+		return fmt.Errorf("invalid default policy: %w", err)
```
peternose:
```suggestion
		return fmt.Errorf("invalid policy: %w", err)
```

go/common/node/sgx.go
```diff
@@ -32,9 +32,15 @@ type SGXConstraints struct {
...
 
-	// Policy is the quote policy.
+	// Policy is the default quote policy.
 	Policy *quote.Policy `json:"policy,omitempty"`
 
+	// KeyManagerAccessPolicy is the compute runtime quote policy for the enclaves,
+	// requesting runtime key derivation from the key manager.
+	//
+	// If not nil, it replaces the default quote policy for compute and observer nodes.
+	KeyManagerAccessPolicy *quote.Policy `json:"kma_policy,omitempty"`
```
peternose, line 48: I would name this `KeyManagerPolicy` and add word `access` in the comment, as the name is complicated.

go/common/node/sgx.go
```diff
@@ -32,9 +32,15 @@ type SGXConstraints struct {
...
 	Enclaves []sgx.EnclaveIdentity `json:"enclaves,omitempty"`
 
-	// Policy is the quote policy.
+	// Policy is the default quote policy.
 	Policy *quote.Policy `json:"policy,omitempty"`
 
+	// KeyManagerAccessPolicy is the compute runtime quote policy for the enclaves,
+	// requesting runtime key derivation from the key manager.
+	//
+	// If not nil, it replaces the default quote policy for compute and observer nodes.
```
peternose, line 41: You are too specific with compute and observer nodes here. The comment should state that this is an optional quote policy that replaces the default policy for nodes that query the key manager. You might also improve the comment for the default policy, i.e. adding description for what is used.

go/common/node/sgx.go
```diff
@@ -111,23 +117,52 @@ func (sc *SGXConstraints) ValidateBasic(cfg *TEEFeatures, isFeatureVersion261 bo
...
-		return fmt.Errorf("invalid policy: %w", err)
+	if err := validateKMAPolicy(sc.KeyManagerAccessPolicy); err != nil {
+		return fmt.Errorf("invalid key manager access policy: %w", err)
 	}
 
 	return nil
 }
 
+// ResolvePolicy returns the effective policy.
+func (sc *SGXConstraints) ResolvePolicy(useKMAPolicy bool) *quote.Policy {
```
peternose, line 171: I would get rid of this method, as it is too complicated. The method returns the key manager policy, if it is set, unless the parameter says otherwise. Complicated an not intuitive. 

I would keep it if it looked like this:
```go
if sc.KeyManagerAccessPolicy != nil {
	return sc.KeyManagerAccessPolicy
}
return sc.Policy
```

go/registry/api/api.go
```diff
@@ -840,12 +840,17 @@ func VerifyNodeRuntimeEnclaveIDs(
...
+		// This is safe because only nodes with TEE capabilities reach this point,
+		// which implies a compute or observer role.
+		useKMAPolicy := regRt.Kind == KindCompute
 		if err := rt.Capabilities.TEE.Verify(node.CapabilityTEEVerifyParams{
 			Features:            teeCfg,
 			Now:                 ts,
 			Height:              height,
 			Constraints:         rtVersionInfo.TEE,
 			NodeID:              nodeID,
+			UseKMAPolicy:        useKMAPolicy,
```
peternose: Using `NodeState`, one could get rid of this `bool` and add node kind to the state. But since this is latter removed, it doesn't matter.

go/registry/api/runtime.go
```diff
@@ -597,6 +597,14 @@ func (r *Runtime) ValidateDeployments(now beacon.EpochTime, params *ConsensusPar
 			if err := cs.ValidateBasic(params.TEEFeatures, isFeatureVersion261); err != nil {
 				return fmt.Errorf("%w: invalid SGX TEE constraints", ErrInvalidArgument)
 			}
+			if cs.KeyManagerAccessPolicy != nil {
+				if r.Kind == KindKeyManager {
```
peternose, line 601: Not sure if this limitations are needed. So what if a key manager defines this policy? And a runtime might define this policy before choosing a key manager.

go/oasis-node/cmd/node/node.go
```diff
@@ -212,9 +212,19 @@ func (n *Node) initRuntimeWorkers(genesisDoc *genesisAPI.Document) error {
 	}
 	n.svcMgr.Register(n.RuntimeRegistry)
 
+	// Determine whether hosted RONL components will be registered on the
+	// consensus layer. This is the case for compute and observer nodes.
```
peternose:
```suggestion
	// consensus layer.
```

go/oasis-node/cmd/node/node.go
```diff
@@ -212,9 +212,19 @@ func (n *Node) initRuntimeWorkers(genesisDoc *genesisAPI.Document) error {
 	}
 	n.svcMgr.Register(n.RuntimeRegistry)
 
+	// Determine whether hosted RONL components will be registered on the
```
peternose: You are stating here that this holds only for RONL, but the name of the variable doesn't say this. Which might lead to future problems.

Therefore, it would be better to solve this some other way, e.g. by passing node kind (which will be too specific and to much information for the code that uses this), or by sending implementation of some kind of interface. I guess the worker could know which kind of node is using running it, but not the code that the worker is using.

go/worker/common/committee/node.go
```diff
@@ -388,7 +391,8 @@ func (n *Node) worker() { //nolint: gocyclo
 
 	// Provision all known components.
 	for _, comp := range bundleRegistry.Components(n.Runtime.ID()) {
-		if err := n.ProvisionHostedRuntimeComponent(comp); err != nil {
+		attestCfg := n.attestationCfg(comp)
```
peternose: Could this be removed by using a struct that implements `GetQuotePolicy`. This ways, `consensus.Service` could be removed from some structs.

go/worker/common/committee/node.go
```diff
@@ -47,6 +47,14 @@ type NodeHooks interface {
 	Initialized() <-chan struct{}
 }
 
+// Config contains committee node specific configuration.
+type Config struct {
```
peternose, line 50: Is this really a config? I would say that this is more of a state or an environment, as mentioned somewhere else. And if we are defining this, why are we not including also other fields from parameters of `committee.NewNode`.

go/oasis-node/cmd/node/node.go
martintomazic: I could also call it `RegisterHostedRONL` (what I meant with compute runtime).

I see your point and the code is currently ugly but I feel alternatives make it even more complex?

go/worker/common/committee/node.go
martintomazic: >And if we are defining this, why are we not including also other fields from parameters of committee.NewNode.

config is static, other fields are services / runtime dependencies.

But yes maybe identity and chaincontext don't necessarily fit. I would avoid state and env as this becomes vague, and favor that every worker have config and typed services.

go/registry/api/runtime.go
martintomazic: > So what if a key manager defines this policy?

I am trying to reduce the amount of valid states, which may simplify the complexity and number of corner cases for our protocol. The cost is slightly more complex validation code.

>And a runtime might define this policy before choosing a key manager.

I find this misleading though, because you are adding a policy for something you don't use. In such case this is no longer keymanager access policy only, or this is an accident and you forgot something so this should error.

Therefore the runtime should define the policy at the time of choosing keymanager iff it choses one.

go/registry/api/api.go
martintomazic: >Using NodeState, one could get rid of this bool and add node kind to the state. 

Ture but here what you really want is runtime kind (keymanager or compute).

go/common/node/sgx.go
martintomazic: Maybe, notice we have two callsites and getting this right is critical so I was able to also write some tests for it. 

But maybe in other threads we find alternative to useKMAPolicy flag, although it feels this will just propagate up.

go/common/node/sgx.go
martintomazic: https://github.com/oasisprotocol/oasis-core/pull/6474#discussion_r2978615065, we can probably get rid of it yes.

go/common/node/sgx.go
martintomazic: I feel this might be confusing as it sound like a policy for the keymanager, not access to it. See existing discussion https://github.com/oasisprotocol/oasis-core/pull/6474#discussion_r305939535

go/common/node/sgx.go
martintomazic: IAS is deprecated, so when I am adding new features I prefer to not have this as an option. Similar to other thread this way protocol has less invariants, corner (dead) cases at the cost of slightly more complex code. Having lean protocol means less surprises in the future.

go/common/node/node.go
martintomazic: Renamed `Now` to `Time`. 

 `Params`,  `State` or `Env` to me are equally general, and I think we use `Params` in the codebase for similar pattern where the goal is to have named params not to confuse order of same type (intent of this fix).
 
 I like your `ConsensusState` (this one is indeed a state) separation, though exported `node.ConsensusState` type feels a bit off?
 
 `DeploymentState` I would avoid and instead have `RuntimeState: Kind, Constraints`, (if kind is ever needed as you [suggested](https://github.com/oasisprotocol/oasis-core/pull/6474#discussion_r3232636748)). Same problem with exported public type.

In general one field nested structs I would avoid, until there is more fields and do refactor at that time.

go/worker/common/committee/node.go
martintomazic: So what you propose is
```go
func (n *RuntimeHostNode) ProvisionHostedRuntimeComponent(comp *bundle.ExplodedComponent, provider host.ConsensusQuotePolicyProvider) error {

```
?

## Concern
On the type system we cannot guarantee component and provider match. 

This might be solved with:
```go
type ConsensusQuotePolicyFactory func(comp, attestationCfg) host.ConsensusQuotePolicyProvider
```

passed to `NewRuntimeHostNodeand` and keeping existing `ProvisionHostedRuntimeComponent` signature.


**Pros**:  we currently create attestation cfg/provisioner for the components that are not sgx enabled - see https://github.com/oasisprotocol/oasis-core/pull/6474#discussion_r3124256192. This solves it.
**Cons**: I hate factories as they make code hard to read and complex.


## Next step
If we decide for this approach I would suggest to create one PR infront fixing the following:
1. Change existing `GetQuotePolicy` to `GetConsensusQuotePolicy`, as epid and ecdsa both pass nil fallback. Then in case of the rofl component, this method would return nil policy as there exist no consensus policy for it. In the case of tdx rofl (qemu) we can then do client side override with tdx defaults, "so that we can provision something" as the comment says.
2. Provider interface, so that `consensus.Service` could be removed from some structs.

martintomazic: https://github.com/oasisprotocol/oasis-core/pull/6531 is now tracking this and can probably be also done as a follow-up. :)

martintomazic:
Following https://github.com/oasisprotocol/oasis-core/pull/6531 I have rebased this PR so that we use new `QuotePolicyProvider`.

I feel existing 
```go
// ResolveQuote resolves a given raw quote into a full bundle with the required collateral.
ResolveQuote(ctx context.Context, rawQuote []byte, quotePolicy *QuotePolicy) (*QuoteBundle, error)
```
should not receive policy, and that instead this whole quote policy provider could be removed from the local provisioner. I assume that if the highest TCB does not pass policy, then any TCB bundle for the lower TCB eval number will also not pass the policy, so passing policy to this function had not extra functionality, except for catching outdated TCB early on.

Instead we could rely on the enclave rejecting quote that does not pass relevant policy, but we need to let it know whether kma policy should be taken into account or ignored.

This could be done e.g. by new RHP method, to communicate whether enclave should use keymanager access policy from the host, like is currently done for the nodeID. Similar to what what was already discussed https://github.com/oasisprotocol/oasis-core/pull/6474#discussion_r2951973041. Note I am not planning to sign this flag with RAK, as is currently done for the nodeID. Here we can assume this is untrusted, and is only used for early validation to simplify host side code. Later we can also do that, albeit consensus anyways validates against the correct policy so not sure this would give us any extra benefit.

_Not sure the code would be simpler after that._

go/upgrade/migrations/consensus_261.go
```diff
@@ -15,6 +15,9 @@ import (
 //     are allowed to query runtime key shares.
 //   - The `FMSPCWhitelist` field in the quote policy, which defines which processor packages
 //     and platform instances are allowed.
+//   - The `KeyManagerAccessPolicy` field in the `SGXConstraints`, which defines additional
+//     policy that may overwrite the default policy when verifying TEE attestations for nodes
```
peternose:
```suggestion
//     policy that overwrites the default policy when verifying TEE attestations for nodes
```

go/worker/common/config.go
```diff
@@ -7,4 +7,8 @@ import (
 // Config contains common worker config.
 type Config struct {
 	TxPool tpConfig.Config
+
+	// WillRegisterComputeRuntime specifies whether hosted RONL components will
```
peternose:
```suggestion
	// WillRegisterComputeRuntime specifies whether the node hosting RONL components will be registered on the consensus layer as compute/observer node.
```

go/oasis-test-runner/scenario/e2e/runtime/kma_policy.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+}
+
+func (sc *keyManagerAccessPolicyImpl) Fixture() (*oasis.NetworkFixture, error) {
+	f, err := sc.Scenario.Fixture()
+	if err != nil {
+		return nil, err
+	}
+
+	f.Runtimes[1].KeyManagerAccessPolicy = &quote.Policy{
```
peternose, line 50:
```suggestion
	f.Runtimes[0].KeyManagerAccessPolicy = &quote.Policy{
```
If I use the above change, the test will still pass, right? So it is hard to know if we are testing the right thing. So I would optionally do two things:
- Add a comment before this line to explain what this test is doing, as code in `Run` tells you nothing.
- Modify the test so that we test what happens if this policy is set and and what happens if it is not. So without access policy the test client should pass and fail otherwise.
  -  This way you don't need all these functions in the test runner, as you can start the network normally, run test client, set the access policy with consensus transaction, run test client again and catch the error.

peternose (approved):

go/oasis-test-runner/scenario/e2e/runtime/kma_policy.go
martintomazic: >If I use the above change, the test will still pass, right? 

No, the test will not pass, as keymanager cannot have keymanager access policy. But I understand your comment is broader.

>Modify the test so that we test what happens if this policy is set and and what happens if it is not. So without access policy the test client should pass and fail otherwise.

The problem is that unless we add additional variable (that disables local validation), you cannot assert consensus rejected the node from registering for the right reason. See https://github.com/oasisprotocol/oasis-core/pull/6474#discussion_r3010346254. 

_If you see a way around this without introducing too much complexity, I would love to implement it. Like you, I am not particularly fond that critical unhappy path is not tested e2e ._

The current value of this test is quoted below, but leaves much to be desired:
>e2e test with mocked TEE was instrumental to realizing that without compute_policy on the rust side of the protocol, RHP would fail due to the missing field.

NIT: existing pattern in the e2e test suite of hardcoding runtime indices is bad, and should be improved in a separate PR.

go/common/node/sgx.go
```diff
@@ -39,6 +39,12 @@ type SGXConstraints struct {
 	// requesting runtime key derivation from the key manager.
 	//
 	// If not nil, it replaces the default quote policy for compute and observer nodes.
+	//
+	// NOTE: Enforced directly during compute/observer registration and committee election.
+	// Key manager EnclaveRPC access is gated transitively through registered runtime membership.
+	// Key manager clients must still satisfy the key manager runtime's own quote policy. Making
+	// the key manager runtime resolve caller runtime policies itself is tracked in:
+	// https://github.com/oasisprotocol/oasis-core/issues/6317
```
martintomazic, line 47: Added a TODO for the follow-up.

NIT:  If/when #6317 changes the key manager to verify callers against caller runtime policies, we need to make sure it does not accidentally call the default runtime policy accessor.  Maybe I should leave additional comment somewhere inside enclave RPC/responder, or refactor and choose policy explicitly instead. The current policy fallback inside responder (prior to this PR) was already error prone and am afraid now is even worse.

go/common/node/sgx.go
martintomazic: I believe this achieved the opposite of the intent and we should harden to:
```go
if policy.IAS == nil || !policy.IAS.Disabled {
    return fmt.Errorf("IAS must be explicitly disabled")
}
```

Otherwise  malicious node passing `ias.AVRBundle` would trigger default IAS verification bypassing most of the validation.

Intel retired EPID/IAS attestation, so fresh AVRs generally aren't obtainable. The code has no idea that's true and is furthermore not documented anywhere.

This comes from the asymetry we have:
* Nil PCS/IAS policy means use hardcoded defaults, whilst nil tdx policy means reject.

runtime/src/consensus/registry.rs
```diff
@@ -703,6 +703,10 @@ pub enum SGXConstraints {
         #[cbor(optional)]
         policy: sgx::QuotePolicy,
 
+        /// The key manager access policy.
+        #[cbor(optional)]
+        kma_policy: sgx::QuotePolicy,
```
martintomazic: Started prototyping https://github.com/oasisprotocol/oasis-core/issues/6317 and realized this has to be `Option<sgx::QuotePolicy>`, else `QuotePolicy::Default` applies.

This way one can distinguish between optional value (nil pointer in go) and empty value. :)

go/oasis-test-runner/scenario/e2e/runtime/kma_policy.go
martintomazic: >The problem is that unless we add additional variable (that disables local validation), you cannot assert consensus rejected the node from registering for the right reason. See https://github.com/oasisprotocol/oasis-core/pull/6474#discussion_r3010346254.

Lets start with valid kma policy, update the descriptor and set policy that no-one can satisfy, wait 1-2 epochs and assert no compute/observer nodes for the given runtime is still registered. 

I believe since failed re-attestation does not clear the previously generated CapabilityTEE the node should re-register using its cached capability, while consensus verifies it against the newly restrictive KMA policy and returns the exact error through Registration.LastAttemptErrorMessage.

martintomazic (commented):
**Update**: This PR was temporary on hold due to other priorities. I did a final pass  once again and explored https://github.com/oasisprotocol/oasis-core/issues/6317 (follow-up) to convince myself this is the right direction. As a result two "minor" issues were found.

My concern is that the code this PR touches is very subtle and full of undocumented corner cases. We have already introduced 4 preparatory PRs that did clean-up and significantly decreased original design complexity but I am still not satisfied. In its current form, this PR adds another layer of subtlety to an area that is already difficult to reason about and review confidently. I would push for simplifying further or document the non-obvious invariants better.

I propose the following steps:
1. Extract trivial PR in-front to keep existing focused. Include:
    * [go/common/node: Move HashRAK function to a better place](https://github.com/oasisprotocol/oasis-core/pull/6474/changes/010a75db0e3c5b41b9ba6871777d0aca3d8dbf3a)
    * [go: Use CapabilityTEEVerifyParams instead of long param list](https://github.com/oasisprotocol/oasis-core/pull/6474/changes/04f7648b40233a1785e3e0166f94676130be68dd)
    * [go/worker/common/committee: Add config parameter](https://github.com/oasisprotocol/oasis-core/pull/6474/changes/5c656bdbc6fcfb5876e872ff460978628e3f4691)
    * Move quote policy selection from the shared SGX/TDX provisioners into each runtime component’s configuration (extract from a2becf169a268ddeff934ac9fea6e60bf5493615).
2. Harden tests, fix existing issues and improve/document (this PR):
    * Asserting unhappy path should be possible e2e without extra variables see https://github.com/oasisprotocol/oasis-core/pull/6474#discussion_r3595582414
    * New quote policy provider probably also enables integration tests.
    * Document some assumptions the code has, or ideally refactor so that assumptions are not needed
3. It would be nice to fix https://github.com/oasisprotocol/oasis-core/issues/6459 in parallel as introducing kma policy makes another place for the asymmetry to manifest.

go/common/node/sgx.go
martintomazic: https://github.com/oasisprotocol/oasis-core/issues/5935#issuecomment-5239461259 - related. It would be nice to solve this beforehand or at least align on the direction and document with TODOs in the code.

peternose:
> Intel retired EPID/IAS attestation, so fresh AVRs generally aren't obtainable. The code has no idea that's true and is furthermore not documented anywhere.

We should probably disable IAS in another PR, so that the code is aware of that, and that we don't need to worry about it anymore.

martintomazic:
>We should probably disable IAS in another PR, so that the code is aware of that, and that we don't need to worry about it anymore.

@peternose  see existing proposal https://github.com/oasisprotocol/oasis-core/issues/5935#issuecomment-5239461259. Agree this has to be done separately.

go/common/node/sgx.go
```diff
@@ -144,8 +144,11 @@ func (sc *SGXConstraints) ValidateBasic(cfg *TEEFeatures, isFeatureVersion261 bo
 		if !isFeatureVersion261 {
 			return fmt.Errorf("policy should be nil")
 		}
-		if policy.IAS != nil {
-			return fmt.Errorf("IAS not allowed")
+
+		// TODO: Remove IAS support entirely.
+		// See https://github.com/oasisprotocol/oasis-core/issues/5935.
+		if policy.IAS == nil || !policy.IAS.Disabled {
```
peternose, line 150: Is this correct?

peternose (commented):
You should organize commits before merging.

PR #6489, opened 2026-03-27
Config: Add explicit observer mode and storage kind 

martintomazic:
Closes #6488, follows #6481.

### DIFF
--- a/.changelog/6488.cfg.1.md
+++ b/.changelog/6488.cfg.1.md
@@ -1,0 +1,9 @@
+go/config: Add explicit local storage config
+
+Explicit `config.Config.Consensus.LocalStorage` (default `true`) was added.
+When set to `false`, remote authenticated storage is used and verified
+by the node's light client.
+
+Mode `stateless-client` no longer exists. It should be instead configured by
+`config.Config.Mode: client` and
+`config.Config.Consensus.LocalStorage: false`.
--- a/go/config/config.go
+++ b/go/config/config.go
@@ -87,6 +87,17 @@ func (c *Config) Validate() error {
 		return fmt.Errorf("unknown node mode: %s", c.Mode)
 	}
 
+	if !c.Consensus.LocalStorage {
+		if c.Consensus.Validator {
+			return fmt.Errorf("local storage not available in specified mode")
+		}
+		switch c.Mode {
+		case ModeClient, ModeObserver:
+		default:
+			return fmt.Errorf("local storage not available in specified mode")
+		}
+	}
+
 	if err = c.Common.Validate(); err != nil {
 		return fmt.Errorf("common: %w", err)
 	}
--- a/go/config/config.go
+++ b/go/config/config.go
@@ -31,40 +31,21 @@ const (
 	ModeValidator NodeMode = "validator"
 	// ModeCompute is the name of the compute node mode.
 	ModeCompute NodeMode = "compute"
+	// ModeObserver is the name of the observer node mode.
+	ModeObserver NodeMode = "observer"
 	// ModeKeyManager is the name of the key manager node mode.
 	ModeKeyManager NodeMode = "keymanager"
 	// ModeClient is the name of the client node mode.
 	ModeClient NodeMode = "client"
-	// ModeStatelessClient is the name of the stateless client node mode.
-	ModeStatelessClient NodeMode = "client-stateless"
 	// ModeSeed is the name of the seed node mode.
 	ModeSeed NodeMode = "seed"
 	// ModeArchive is the name of the archive node mode.
 	ModeArchive NodeMode = "archive"
 )
 
-// IsClientOnly returns true iff the mode is one that has the node running
-// as a client for all configured runtimes.
-func (m NodeMode) IsClientOnly() bool {
-	switch m {
-	case ModeClient, ModeStatelessClient:
-		return true
-	}
-	return false
-}
-
 // IsArchive returns true iff the mode is set to archive node mode.
 func (m NodeMode) IsArchive() bool {
 	return m == ModeArchive
-}
-
-// HasLocalStorage returns true iff the mode is one that has local storage.
-func (m NodeMode) HasLocalStorage() bool {
-	switch m {
-	case ModeClient, ModeCompute, ModeArchive:
-		return true
-	}
-	return false
 }
 
 // GlobalConfig holds the global configuration options.
@@ -97,9 +78,9 @@ func (c *Config) Validate() error {
 	switch c.Mode {
 	case ModeValidator:
 	case ModeCompute:
+	case ModeObserver:
 	case ModeKeyManager:
 	case ModeClient:
-	case ModeStatelessClient:
 	case ModeSeed:
 	case ModeArchive:
 	default:
--- a/go/config/config_test.go
+++ b/go/config/config_test.go
@@ -1,0 +1,39 @@
+package config
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestRemoteStorageValidation(t *testing.T) {
+	for _, tc := range []struct {
+		name      string
+		mode      NodeMode
+		validator bool
+		mustErr   bool
+	}{
+		{name: "client", mode: ModeClient},
+		{name: "observer", mode: ModeObserver},
+		{name: "validator", mode: ModeValidator, mustErr: true},
+		{name: "compute", mode: ModeCompute, mustErr: true},
+		{name: "keymanager", mode: ModeKeyManager, mustErr: true},
+		{name: "seed", mode: ModeSeed, mustErr: true},
+		{name: "archive", mode: ModeArchive, mustErr: true},
+		{name: "consensus validator and observer", mode: ModeObserver, validator: true, mustErr: true},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			cfg := DefaultConfig()
+			cfg.Mode = tc.mode
+			cfg.Consensus.LocalStorage = false
+			cfg.Consensus.Validator = tc.validator
+
+			err := cfg.Validate()
+			if tc.mustErr {
+				require.ErrorContains(t, err, "local storage not available in specified mode")
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+}
--- a/go/consensus/cometbft/cometbft.go
+++ b/go/consensus/cometbft/cometbft.go
@@ -37,26 +37,27 @@ func New(
 		return nil, err
 	}
 
-	switch config.GlobalConfig.Mode {
-	case config.ModeArchive:
+	if config.GlobalConfig.Mode == config.ModeArchive {
 		node, err := createArchiveNode(ctx, dataDir, identity, genesis, doc, genesisDoc)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create archive node: %w", err)
 		}
 		return node, nil
-	case config.ModeStatelessClient:
+	}
+
+	if !config.GlobalConfig.Consensus.LocalStorage {
 		node, err := createStatelessNode(ctx, dataDir, identity, genesis, doc, genesisDoc, p2p)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create stateless node: %w", err)
 		}
 		return node, nil
-	default:
-		node, err := createFullNode(ctx, dataDir, identity, genesis, doc, genesisDoc, upgrader, p2p)
-		if err != nil {
-			return nil, fmt.Errorf("failed to create full node: %w", err)
-		}
-		return node, nil
 	}
+
+	node, err := createFullNode(ctx, dataDir, identity, genesis, doc, genesisDoc, upgrader, p2p)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create full node: %w", err)
+	}
+	return node, nil
 }
 
 func createArchiveNode(
--- a/go/consensus/cometbft/config/config.go
+++ b/go/consensus/cometbft/config/config.go
@@ -39,6 +39,12 @@ type Config struct {
 	// Average amount of time to delay shutting down the node on upgrade.
 	UpgradeStopDelay time.Duration `yaml:"upgrade_stop_delay,omitempty"`
 
+	// LocalStorage specifies whether consensus state is stored locally.
+	//
+	// If false and the node has additional runtime storage configured,
+	// that runtime storage is also accessed remotely.
+	LocalStorage bool `yaml:"local_storage,omitempty"`
+
 	// ABCI state pruning configuration.
 	Prune PruneConfig `yaml:"prune,omitempty"`
 
@@ -233,6 +239,7 @@ func DefaultConfig() Config {
 			MaxFee: 10_000_000_000,
 		},
 		UpgradeStopDelay: time.Minute,
+		LocalStorage:     true,
 		Prune: PruneConfig{
 			Strategy: PruneStrategyNone,
 			NumKept:  250_000,
--- a/go/oasis-node/cmd/node/node.go
+++ b/go/oasis-node/cmd/node/node.go
@@ -277,7 +277,11 @@ func (n *Node) initRuntimeWorkers(genesisDoc *genesisAPI.Document) error {
 	n.svcMgr.Register(n.BeaconWorker)
 
 	// Initialize the storage worker.
+	storageCfg := workerStorage.Config{
+		HasLocalStorage: config.GlobalConfig.Consensus.LocalStorage,
+	}
 	n.StorageWorker, err = workerStorage.New(
+		storageCfg,
 		n.grpcInternal,
 		n.CommonWorker,
 		n.RegistrationWorker,
--- a/go/oasis-node/cmd/node/node.go
+++ b/go/oasis-node/cmd/node/node.go
@@ -479,6 +479,20 @@ func NewNode() (node *Node, err error) { // nolint: gocyclo
 	}
 	node.EntityID = entityID
 
+	requiresEntity := config.GlobalConfig.Consensus.Validator
+	switch config.GlobalConfig.Mode {
+	case config.ModeValidator, config.ModeCompute, config.ModeKeyManager, config.ModeObserver:
+		requiresEntity = true
+	}
+
+	if requiresEntity && entityID == nil {
+		return nil, fmt.Errorf("node has no entity configured but expects one")
+	}
+
+	if config.GlobalConfig.Mode == config.ModeClient && entityID != nil {
+		return nil, fmt.Errorf("client node must not have entity configured, try using observer mode instead")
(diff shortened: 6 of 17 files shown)


go/consensus/cometbft/config/config.go
```diff
@@ -39,6 +39,9 @@ type Config struct {
 	// Average amount of time to delay shutting down the node on upgrade.
 	UpgradeStopDelay time.Duration `yaml:"upgrade_stop_delay,omitempty"`
 
+	// StateKind specifies whether state is stored locally or fetched from the network.
+	StateKind string `yaml:"state_kind,omitempty"`
```
martintomazic: Or simply `LocalStorage bool`.

go/consensus/cometbft/config/config.go
```diff
@@ -195,6 +206,12 @@ func (c *Config) Validate() error {
 		return fmt.Errorf("only one of {halt_epoch, halt_height} can be set")
 	}
 
+	switch c.StateKind {
+	case StateKindFull, StateKindStateless:
+	default:
+		return fmt.Errorf("invalid state kind: %s", c.StateKind)
+	}
```
martintomazic: NIT: prevent invalid configs e.g. compute node with stateless storage configured..

go/worker/common/committee/p2p.go
```diff
@@ -36,19 +37,18 @@ func (h *txMsgHandler) HandleMessage(ctx context.Context, _ signature.PublicKey,
 
 	tx := msg.([]byte) // Ensured by DecodeMessage.
 
-	switch config.GlobalConfig.Mode {
-	case config.ModeStatelessClient:
-		// Ignore transactions on stateless clients.
+	if config.GlobalConfig.Consensus.StateKind == consensusCfg.StateKindStateless {
+		return nil // Ignore transactions on stateless clients.
+	}
```
martintomazic: I can bring this in using `cfg.accept_gossiped_txs` to avoid global state. 

The location of this handler is weird though, as this protocol is anyways enabled only for the compute nodes, therefore it doe not fit into `common` package.

go/worker/client/worker.go
```diff
@@ -158,8 +156,12 @@ func (w *Worker) registerRuntime(commonNode *committeeCommon.Node) error {
...
 	}
 
-	// If we are running in stateless client mode, register remote storage.
-	if config.GlobalConfig.Mode == config.ModeStatelessClient {
+	// If we have stateless storage, register remote storage.
+	// TODO: Client worker should not care about the storage kind.
+	// Instead storage initialization should be done during node initialization.
+	// Same applies for storage committee worker that initializes and registers local storage, but
+	// should intead receive it.
+	if config.GlobalConfig.Consensus.StateKind == consensusCfg.StateKindStateless {
```
martintomazic: We can do this in two ways:
1. Add new `registry.Runtime.GetLocalStorage` that returns local storage if it exits.
2. Initialize local storage at the node startup and then pass it to common worker (to set it into Runtime container) and to storage worker.

Both options are nice because we can completely decouple storage committee and common worker. Out of scope of this PR, but let's open a separate issue for that...

go/oasis-node/cmd/node/node.go
```diff
@@ -469,6 +469,17 @@ func NewNode() (node *Node, err error) { // nolint: gocyclo
...
+	if config.GlobalConfig.Mode == config.ModeObserver && entityID == nil {
+		return nil, fmt.Errorf("observer node has no entity configured")
+	}
+
+	if config.GlobalConfig.Mode == config.ModeClient && entityID != nil {
+		logger.Warn("client node must not have entity configured",
+			"hint", `to run observer node use "observer" mode with entity configured`,
+		)
+		return nil, fmt.Errorf("client node must not have entity configured")
+	}
```
martintomazic: The fact that CI didn't fail means we had no observer e2e test in the past?

go/worker/storage/worker.go
```diff
@@ -42,7 +42,7 @@ func New(
 	registration *registration.Worker,
 ) (*Worker, error) {
 	ctx, cancel := context.WithCancel(context.Background())
-	enabled := config.GlobalConfig.Mode.HasLocalStorage() && len(commonWorker.GetRuntimes()) > 0
+	enabled := config.GlobalConfig.HasLocalStorage() && len(commonWorker.GetRuntimes()) > 0
```
martintomazic: Follow-up: This worker shouldn't even be started if no local storage || zero runtimes.

go/oasis-node/cmd/node/node_control.go
```diff
@@ -270,7 +270,7 @@ func (n *Node) getRuntimeStatus(ctx context.Context) (map[common.Namespace]contr
 		}
 
 		// Take storage into account for last retained round.
-		if config.GlobalConfig.Mode.HasLocalStorage() {
+		if config.GlobalConfig.HasLocalStorage() {
 			lsb, ok := rt.Storage().(storage.LocalBackend)
 			switch ok {
 			case false:
```
martintomazic: As [pointed](https://github.com/oasisprotocol/oasis-core/pull/6489/changes/329f7f5965a3b36128074421c3b856a2c6007d2a#r3010045635), local storage should be passed explicitly or we add `func (r *Runtime) LocalStorage () (storage.LocalBakend, error)`  that returns optional local storage.

.changelog/6488.cfg.1.md
```diff
@@ -0,0 +1,9 @@
+go/config: Add explicit storage kind and remove `stateless-client` mode
+
+An explicit storage kind was added to `config.Config.Consensus.StateKind`.
+It can be either `full` (default, local storage), or `stateless` which
+uses remote (authenticated) storage verified by the node's light client.
+
+`stateless-client` mode no longer exists. It should be instead configured by
+`config.Config.Mode: client` and
+`config.Config.Consensus.StateKind: stateless`.
```
kostko: But there is also stateless storage client, how is that configured now?

Edit: I see below this is mentioned in one of the TODOs. Ideally there should be a similar configuration option for runtimes, it could even be a per-runtime configuration option. It is fine if you leave the TODO and open a separate issue.

go/consensus/cometbft/config/config.go
kostko: Do you think there will be other kinds of state storage?

go/consensus/cometbft/cometbft.go
```diff
@@ -37,26 +39,27 @@ func New(
...
+	if config.GlobalConfig.Mode == config.ModeArchive {
 		node, err := createArchiveNode(ctx, dataDir, identity, genesis, doc, genesisDoc)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create archive node: %w", err)
 		}
 		return node, nil
-	case config.ModeStatelessClient:
+	}
+
+	if config.GlobalConfig.Consensus.StateKind == consensusCfg.StateKindStateless {
```
kostko: So now a stateless archive node is possible? Are these configurations tested anywhere or should they be disabled?

go/oasis-node/cmd/node/node.go
kostko: We did have observer tests AFAIK.

go/consensus/cometbft/config/config.go
martintomazic: If we only want global config no.**

That would also fix weird imports:
```go
config.GlobalConfig.Consensus.StateKind == consensusCfg.StateKindStateless
config.GlobalConfig.Runtime.StateKind == runtimeCfg.StateKindStateless
```
that could be mitigated by placing in enum in some shared package e.g. `go/storage/api` or `go/common/statekind`? The former probably not as this is the actual backend config. Meh.

_Will refactor to `storage.local bool` as also more understandable for the operators, less code and overall cleaner._

**If you need per runtime override we need 3 values: either `string` or use `*bool` for the local override.

.changelog/6488.cfg.1.md
martintomazic: I can do it here since should be trivial. 

This would also open possibility for stateless consensus + stateful runtime. Not sure this setup is operationally practical as setting aggressive pruning for full consensus client, makes consensus part already negligible compared to the statefull runtime part. But lets do it for the sake of completeness. Per-runtime I would avoid until needed - [see](https://github.com/oasisprotocol/oasis-core/pull/6489#discussion_r3056677181).

Cons: Now you need to set two variables but imo worth it.

go/oasis-node/cmd/node/node.go
martintomazic: Will write one ;)

.changelog/6488.cfg.1.md
kostko: Let's avoid it for now.

go/config/config_test.go
```diff
@@ -0,0 +1,39 @@
...
+		mode      NodeMode
+		validator bool
+		mustErr   bool
+	}{
+		{name: "client", mode: ModeClient},
+		{name: "observer", mode: ModeObserver},
+		{name: "validator", mode: ModeValidator, mustErr: true},
+		{name: "compute", mode: ModeCompute, mustErr: true},
+		{name: "keymanager", mode: ModeKeyManager, mustErr: true},
+		{name: "seed", mode: ModeSeed, mustErr: true},
```
martintomazic, line 21: Key-manager is technically stateless. Keymanager and seed should probably have separate config, as most of the configs that apply to client/compute are possible (not rejected) but simply ignored.

Out of scope, just this test makes it clearer.

go/oasis-node/cmd/node/node.go
```diff
@@ -475,6 +479,17 @@ func NewNode() (node *Node, err error) { // nolint: gocyclo
 	}
 	node.EntityID = entityID
 
+	if config.GlobalConfig.Mode == config.ModeObserver && entityID == nil {
```
kostko: What about other kinds of nodes? Validator, compute, key manager? These should all have an entity configured?

go/oasis-node/cmd/node/node.go
```diff
@@ -277,7 +277,11 @@ func (n *Node) initRuntimeWorkers(genesisDoc *genesisAPI.Document) error {
 	n.svcMgr.Register(n.BeaconWorker)
 
 	// Initialize the storage worker.
+	storageCfg := workerStorage.Config{
+		HasLocalStorage: config.GlobalConfig.Consensus.LocalStorage,
```
kostko, line 281: Is there a reason why the storage worker doesn't read this from the global config directly?

martintomazic: As I am refactoring the code (and if trivial) I am trying to push out the global config from the workers as much as possible so that the final refactor will be easier. This also encourages me to write better documentation, tests and abstractions.

E.g. similar thread: https://github.com/oasisprotocol/oasis-core/pull/6481#discussion_r3028011467.

go/oasis-node/cmd/node/node.go
martintomazic: Added the logic. Notice I also added fixup commit that allows you to run validator + observer, just like is possible to be validator + compute. Previously this was not possible due to abusing client role + entity for observer. Whether it is good to run ~consensus~ validator + runtime is a different question :).

.changelog/6488.cfg.1.md
martintomazic: The only reason to do it now is that if we do it later this would likely be breaking cfg again, unless we make it `*bool`, and by default inherit consensus value.

I find this sensible even if we do it now, in such case no need to do it here. :)

kostko (approved):

.changelog/6488.cfg.1.md
```diff
@@ -0,0 +1,9 @@
+go/config: Add explicit local storage config
+
+Explicit `config.Config.Consensus.LocalStorage` (default `true`) was added.
+When set to `false`, remote authenticated storage is used and verified
+by the node's light client.
+
+`stateless-client` mode no longer exists. It should be instead configured by
```
peternose:
```suggestion
Mode `stateless-client` no longer exists. It should be instead configured by
```
Don't start sentences this way.

go/consensus/cometbft/config/config.go
```diff
@@ -39,6 +39,12 @@ type Config struct {
 	// Average amount of time to delay shutting down the node on upgrade.
 	UpgradeStopDelay time.Duration `yaml:"upgrade_stop_delay,omitempty"`
 
+	// LocalStorage specifies whether consensus state is stored locally.
+	//
+	// If false and the node has additional runtime storage configured,
+	// that runtime storage is also accessed remotely.
+	LocalStorage bool `yaml:"local_storage"`
```
peternose:
```suggestion
	LocalStorage bool `yaml:"local_storage,omitempty"`
```

go/oasis-test-runner/oasis/oasis.go
```diff
@@ -303,12 +303,12 @@ func (n *Node) Start() error {
...
 
-	switch n.Config.Mode {
-	case config.ModeStatelessClient:
+	if !n.Config.Consensus.LocalStorage {
 		for _, worker := range n.net.computeWorkers {
 			n.Config.Consensus.Providers = append(n.Config.Consensus.Providers, "unix:"+worker.SocketPath())
 		}
-	case config.ModeSeed:
+	}
+	if n.Config.Mode == config.ModeSeed {
```
peternose, line 311: Won't seed nodes by default have local storage enabled?

go/oasis-test-runner/oasis/stateless_client.go
```diff
@@ -58,7 +57,7 @@ func (client *StatelessClient) ModifyConfig() error {
 
 	client.Config.P2P.Port = client.p2pPort
 
-	client.Config.Mode = config.ModeStatelessClient
+	client.Config.Consensus.LocalStorage = false
```
peternose, line 62: Do we need to add line `client.Config.Mode = config.ModeClient`?

go/worker/client/worker.go
```diff
@@ -158,8 +155,12 @@ func (w *Worker) registerRuntime(commonNode *committeeCommon.Node) error {
 		return err
 	}
 
-	// If we are running in stateless client mode, register remote storage.
-	if config.GlobalConfig.Mode == config.ModeStatelessClient {
+	// If we have stateless storage, register remote storage.
+	// TODO: Client worker should not care about the storage kind.
```
peternose: Nobody will fix this, so I would remove text, reword, or create an issue with this description.

go/oasis-node/cmd/node/node.go
```diff
@@ -475,6 +479,23 @@ func NewNode() (node *Node, err error) { // nolint: gocyclo
...
+		requiresEntity = true
+	}
+
+	if requiresEntity && entityID == nil {
+		return nil, fmt.Errorf("node has no entity configured but expects one")
+	}
+
+	if config.GlobalConfig.Mode == config.ModeClient && !config.GlobalConfig.Consensus.Validator && entityID != nil {
+		logger.Warn("client node must not have entity configured",
+			"hint", `to run observer node use "observer" mode with entity configured`,
```
peternose: This is new. I would remove this log as the error should tell you what is wrong. For solution, see the docs. Unless this hint is temporary.

go/oasis-node/cmd/node/node.go
```diff
@@ -475,6 +479,23 @@ func NewNode() (node *Node, err error) { // nolint: gocyclo
...
+	switch config.GlobalConfig.Mode {
+	case config.ModeValidator, config.ModeCompute, config.ModeKeyManager, config.ModeObserver:
+		requiresEntity = true
+	}
+
+	if requiresEntity && entityID == nil {
+		return nil, fmt.Errorf("node has no entity configured but expects one")
+	}
+
+	if config.GlobalConfig.Mode == config.ModeClient && !config.GlobalConfig.Consensus.Validator && entityID != nil {
```
peternose, line 492: Why do we allow validator + observer?

go/config/config.go
```diff
@@ -97,15 +78,26 @@ func (c *Config) Validate() error {
...
-	case ModeStatelessClient:
 	case ModeSeed:
 	case ModeArchive:
 	default:
 		return fmt.Errorf("unknown node mode: %s", c.Mode)
 	}
 
+	if !c.Consensus.LocalStorage {
+		if c.Consensus.Validator {
+			return fmt.Errorf("consensus.local_storage: false, but should be true for validator")
```
peternose: I still hate variables in comments, error msgs, .... What happens if someone renames the variable? This msg won't change.
I would also change the error so that it is general, e.g. `local storage not available in specified mode`.

peternose (commented):
Are you planning to also update the docs, e.g. for Stateless Client Node?

go/oasis-test-runner/oasis/stateless_client.go
martintomazic: client mode is actually the default value, added it explicitly now.

go/oasis-test-runner/oasis/oasis.go
martintomazic: What do you mean? this was mechanical refactor but maybe I missed smtg?

go/oasis-node/cmd/node/node.go
martintomazic: The hint is there to ease the life of operators that might be affected by the new breaking cfg change. I would not make it as part of error, as error should tell what went wrong not how to fix it. Warning log seemed the most reasonable place.

Yes we can remove it later, but I feel issue/follow-up PR might be an overkill? Happy to do it though.

go/oasis-node/cmd/node/node.go
martintomazic: For the same reason that we allow validator + compute. See https://github.com/oasisprotocol/oasis-core/pull/6489#discussion_r3086940250.

go/oasis-test-runner/oasis/oasis.go
peternose: Before, seed nodes would never execute line `for _, worker := range n.net.computeWorkers {`, but now they can if local storage is disabled. Maybe for test runner this doesn't matter.

go/oasis-test-runner/oasis/stateless_client.go
peternose: Yes, I prefer explicit.

go/oasis-node/cmd/node/node.go
peternose: I would warn with the following msg: `client node must not have entity configured, try using observer mode instead`, to prevent using `hint` as a key.

go/oasis-test-runner/oasis/oasis.go
martintomazic: https://github.com/oasisprotocol/oasis-core/pull/6489#discussion_r3080852415 it cannot be disabled. Not optimal either but this is a separate problem.

peternose (approved):

PR #6497, opened 2026-04-12
oasis-core/go/oasis-test-runner: Add observer fixture and e2e test

### LINKED ISSUE #6495: oasis-core/go/oasis-test-runner: Add observer fixture and e2e observer test
## Motivation
Ability to write e2e tests for #6488, #6387 and have existing critical code path tested.

##  Solution
1. Add new `oasis.Observer` and `ObserverFixture`. 
2. Meaningful e2e test:
    * Instead of  just smoke testing observer configuration and registration assert behavior.
    * Suggested candidate: Make sure only whitelisted observer can call `enc_get`.  This call should be rejected for non-whitelisted observer, and normal clients. 

## Considerations
E2e tests entities are indexed based and only resolved during network initialization. Since for whitelist we need entity public keys, that are not know yet during fixture specification, we would also need `RuntimeAdmissionPolicyFixture` to  specify whitelist using indexes:

```go

type RuntimeAdmissionPolicyFixture struct {
	AnyNode         bool                                             `json:"any_node,omitempty"`
	EntityWhitelist *EntityWhitelistRuntimeAdmissionPolicyFixture    `json:"entity_whitelist,omitempty"`
	PerRole         map[node.RolesMask]PerRoleAdmissionPolicyFixture `json:"per_role,omitempty"`
}

type EntityWhitelistRuntimeAdmissionPolicyFixture struct {
	Entities map[int]registry.EntityWhitelistConfig `json:"entities"`
}

type PerRoleAdmissionPolicyFixture struct {
	EntityWhitelist *EntityWhitelistRoleAdmissionPolicyFixture `json:"entity_whitelist,omitempty"`
}

type EntityWhitelistRoleAdmissionPolicyFixture struct {
	Entities map[int]registry.EntityWhitelistRoleConfig `json:"entities"`
}

func (f RuntimeAdmissionPolicyFixture) Resolve(net *Network) (registry.RuntimeAdmissionPolicy, error) {
...
}
```

This results in around 70 LOC of boilerplate code.

## Nice to have
* As `Client`, `Observer` and `Compute` share a lot of fixture fields and initialization (especially related to hosted components) which is also opportunity to extract this.

martintomazic:
Closes #6495

### DIFF
--- a/.buildkite/code.pipeline.yml
+++ b/.buildkite/code.pipeline.yml
@@ -262,6 +262,7 @@ steps:
         --scenario e2e/runtime/rofl
         --scenario e2e/runtime/trust-root/.+
         --scenario e2e/runtime/keymanager-.+
+        --scenario e2e/runtime/observer
     env:
       # Unsafe flags needed as the trust-root test rebuilds the enclave with embedded trust root data.
       OASIS_UNSAFE_SKIP_AVR_VERIFY: "1"
--- a/go/oasis-test-runner/oasis/fixture.go
+++ b/go/oasis-test-runner/oasis/fixture.go
@@ -28,6 +28,7 @@ type NetworkFixture struct {
 	Keymanagers        []KeymanagerFixture       `json:"keymanagers,omitempty"`
 	KeymanagerPolicies []KeymanagerPolicyFixture `json:"keymanager_policies,omitempty"`
 	ComputeWorkers     []ComputeWorkerFixture    `json:"compute_workers,omitempty"`
+	Observers          []ObserverFixture         `json:"observers,omitempty"`
 	Sentries           []SentryFixture           `json:"sentries,omitempty"`
 	Clients            []ClientFixture           `json:"clients,omitempty"`
 	StatelessClients   []StatelessClientFixture  `json:"stateless_clients,omitempty"`
@@ -102,6 +103,13 @@ func (f *NetworkFixture) Create(env *env.Env) (*Network, error) {
 
 	// Provision the compute workers.
 	for _, fx := range f.ComputeWorkers {
+		if _, err = fx.Create(net); err != nil {
+			return nil, err
+		}
+	}
+
+	// Provision the observers.
+	for _, fx := range f.Observers {
 		if _, err = fx.Create(net); err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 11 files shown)


go/oasis-test-runner/oasis/fixture.go
```diff
@@ -249,7 +250,7 @@ type RuntimeFixture struct {
 	TxnScheduler registry.TxnSchedulerParameters `json:"txn_scheduler"`
 	Storage      registry.StorageParameters      `json:"storage"`
 
-	AdmissionPolicy registry.RuntimeAdmissionPolicy                                               `json:"admission_policy"`
```
martintomazic, line 244: NIT: is this actually breaking for anyone?

kostko: Could be for any already generated fixtures. But if so, it will be handled during a version bump, so it shouldn't be problematic.

martintomazic:
SGX tests seems to be failing because of the CI runner issue: [see](https://oasisprotocol.slack.com/archives/C01KA8RRJBU/p1776017009853899?thread_ts=1775934491.024479&cid=C01KA8RRJBU)

**Update**: was fixed.

.buildkite/code.pipeline.yml
```diff
@@ -262,6 +262,7 @@ steps:
         --scenario e2e/runtime/rofl
         --scenario e2e/runtime/trust-root/.+
         --scenario e2e/runtime/keymanager-.+
+        --scenario e2e/runtime/observer
```
martintomazic, line 265: To be able to run this in non-sgx mode we would have to fix:
https://github.com/oasisprotocol/oasis-core/blob/master/go/worker/keymanager/secrets.go#L245:L264. I.e. in non-sgx mode access list obtained from the consensus registered nodes is not applied.

Imo out of scope, also fine with stronger test albeit less practical to test locally.

go/oasis-test-runner/oasis/fixture.go
```diff
@@ -458,6 +466,53 @@ func (f *ComputeWorkerFixture) Create(net *Network) (*Compute, error) {
...
+	}
+
+	return net.NewObserver(&ObserverCfg{
+		NodeCfg: NodeCfg{
+			Name:                        f.Name,
+			Consensus:                   f.Consensus,
+			AllowErrorTermination:       f.AllowErrorTermination,
+			AllowEarlyTermination:       f.AllowEarlyTermination,
+			NoAutoStart:                 f.NoAutoStart,
+			SupplementarySanityInterval: f.Consensus.SupplementarySanityInterval,
```
peternose, line 574: Do we also need `Consensus: f.Consensus,`?

go/oasis-test-runner/scenario/e2e/runtime/observer.go
```diff
@@ -0,0 +1,214 @@
...
+
+	// Sanity: ensure whitelisted observer has registered and non-whitelisted failed to prevent false positives.
+	if err = sc.waitObserverRegistration(ctx, whitelistedCtrl, true); err != nil {
+		return err
+	}
+	if err = sc.waitObserverRegistration(ctx, nonWhitelistedCtrl, false); err != nil {
+		return err
+	}
+
+	sc.Logger.Info("ensuring client cannot query encrypted key")
```
peternose:
```suggestion
	sc.Logger.Info("ensuring client cannot query encrypted value")
```

go/oasis-test-runner/scenario/e2e/runtime/observer.go
```diff
@@ -0,0 +1,214 @@
...
+func (sc *observerModeImpl) waitObserverRegistration(ctx context.Context, ctrl *oasis.Controller, mustRegister bool) error {
+	sc.Logger.Info("waiting for observer registration result", "must_register", mustRegister)
+
+	reg, err := sc.waitObserverRegistrationAttempt(ctx, ctrl)
+	if err != nil {
+		return err
+	}
+
+	// Must not register
+	if !mustRegister {
```
peternose, line 173: Could also solve with ` mustRegister != reg.LastAttemptSuccessful `.

go/oasis-test-runner/scenario/e2e/runtime/observer.go
```diff
@@ -0,0 +1,214 @@
...
+	value, err := sc.queryEncryptedKey(ctx, whitelistedCtrl, observerTestKey)
+	if err != nil {
+		return fmt.Errorf("whitelisted observer failed querying encrypted key: %w", err)
+	}
+	if value != observerTestValue {
+		return fmt.Errorf("whitelisted observer query returned unexpected value (got: %s, want: %s)", value, observerTestValue)
+	}
+
+	sc.Logger.Info("checking non-whitelisted observer cannot query encrypted key")
+	if _, err = sc.queryEncryptedKey(ctx, nonWhitelistedCtrl, observerTestKey); err == nil {
```
peternose, line 138: I usually use `switch` when checking if `err == nil` or I add a comment at the end, as this is not a normal error check.

peternose (approved):
LGTM 👍

peternose:
The file `.changelog/6495.trivial.md` could also be renamed to `.changelog/6497.trivial.md`, to point to PR not the issue. As an issue could have multiple PRs, with different commits which are seen in the changelog.

go/oasis-test-runner/oasis/fixture.go
martintomazic: Was there already, but followed client style. No constructor follows `NodeCfg` ordering, so ignoring for now.

go/oasis-test-runner/scenario/e2e/runtime/observer.go
martintomazic: True but then some logs are missing / cannot easily check observer role so keeping as is.

PR #6512, opened 2026-04-22
go/registry/api: Improve VerifyRuntime function signature and test suite

martintomazic:
**Trivial**

Doing this here to avoid clutter of #6474 and further duplication of code in the test suite.

### DIFF
--- a/go/registry/api/api.go
+++ b/go/registry/api/api.go
@@ -1101,21 +1101,29 @@ func VerifyNodeUpdate(
 	return nil
 }
 
+// VerifyRuntimeOptions contains the options for VerifyRuntime.
+type VerifyRuntimeOptions struct {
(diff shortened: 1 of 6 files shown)


go/registry/api/api.go
```diff
@@ -1102,20 +1102,28 @@ func VerifyNodeUpdate(
...
+
 func VerifyRuntime( // nolint: gocyclo
 	params *ConsensusParameters,
 	logger *logging.Logger,
 	rt *Runtime,
-	isGenesis bool,
-	isSanityCheck bool,
 	now beacon.EpochTime,
-	isFeatureVersion242 bool,
+	opts VerifyRuntimeOptions,
```
martintomazic, line 1120: NIT: technically an API change, but don't think this was ever meant as an API/there are downstream users of this function...

peternose (approved):

PR #6520, opened 2026-05-04
go/oasis-node/cmd/common: Fix isNodeRunning function

martintomazic:
## What
Storage commands gated by this check are sometimes blocked due to the stale internal socket. This has now also surfaced in the e2e test I was writing about offline pruning/compaction.

## How to test locally
A non-graceful shutdown of the node tends to keep stale `internal.sock`.

## Observation
Seem that `internal.sock`  is sometimes not cleaned even with graceful shutdown. Meaning a possible bug during shutdown/cleanup (out of scope).

### DIFF
--- a/go/oasis-node/cmd/common/common.go
+++ b/go/oasis-node/cmd/common/common.go
@@ -6,9 +6,12 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"net"
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
+	"time"
 
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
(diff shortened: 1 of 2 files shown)


go/oasis-node/cmd/common/common.go
```diff
@@ -77,14 +80,17 @@ func InternalSocketPath() string {
 func IsNodeRunning() (bool, error) {
 	path := InternalSocketPath()
 
-	if _, err := os.Stat(path); err != nil {
-		if errors.Is(err, fs.ErrNotExist) {
-			return false, nil
-		}
-		return false, fmt.Errorf("stat %s: %w", path, err)
+	conn, err := net.DialTimeout("unix", path, time.Second)
+	if err == nil { // NO error
```
peternose: Could respect the happy path.

```go
switch {
    case err == nil:
        _ = conn.Close()
    case errors.Is(err, fs.ErrNotExist) :
    ...
}

return true
```

peternose (approved):

peternose (approved):

PR #6521, opened 2026-05-06
Release oasis storage prune and compact subcomands

### LINKED ISSUE #6519: Release offline pruning and compact command
With #6334 merged, sub-commands `prune-experimental` and `compact-experimental` should be replaced with `prune` and `compact` respectively. Both commands should also apply to runtime history and state.

martintomazic:
Closes #6519.

Relatively trivial to review. LOC is big because I extracted commands to separate files and added missing tests (separate commits).

## Motivation
1. Offline pruning/compaction was introduced because late pruning 1. may not reclaim disk space, 2. may cause node to fall behind even once registered as ready (pending compaction load).
2. We could also use it to create https://github.com/oasisprotocol/oasis-core/issues/6423, serving as alternative to https://github.com/oasisprotocol/oasis-core/pull/6454.

## How to test locally

For testing I used latest Sapphire  [snapshot](https://snapshots.oasis.io/#node/mainnet/20250722-20260403/) (cca 7 months of data, `keep_n = 100_000`). It took 10min to prune and 6min to compact. For consensus I used [snapshot-old](https://snapshots.oasis.io/#node/mainnet/20250722-20251105/) (cca 3 months of data, `keep_n = 100_000`), due to limited disk space on my machine. Pruning took 2min and compaction 3min.

Node started syncing normally after prune/compact commands.

## Possible follow-up
For https://github.com/oasisprotocol/oasis-core/pull/6454, we may want to add additional flag to the pruning command, e.g. `retain_height`, that ignores pruning config and calculates corresponding rounds at this height of every runtime, implementing `keep_from` [proposal](https://github.com/oasisprotocol/oasis-core/issues/6423#issue-3689723722).

### DIFF
--- a/docs/oasis-node/cli.md
+++ b/docs/oasis-node/cli.md
@@ -365,19 +365,19 @@ may stay constant or not be reclaimed for a very long time.
 This command gives operators manual control to release disk space during
 maintenance periods.
 
-### prune-experimental
+### prune
 
 Run (when the node is not running):
 
 ```sh
-oasis-node storage prune-experimental --config /path/to/config/file
+oasis-node storage prune --config /path/to/config/file
 ```
 
-to trigger manual pruning of consensus database instances:
+to trigger manual pruning of all configured database instances:
 
 ```sh
-{"caller":"storage.go:433","level":"info","module":"cmd/storage", \
-"msg":"Starting consensus databases pruning. This may take a while...", \
+{"caller":"prune.go:47","level":"info","module":"cmd/storage", \
+"msg":"Starting databases pruning. This may take a while...", \
 "ts":"2025-10-23T11:02:11.129822974Z"}
 ```
 
--- a/docs/oasis-node/cli.md
+++ b/docs/oasis-node/cli.md
@@ -334,15 +334,15 @@ oasis1qqncl383h8458mr9cytatygctzwsx02n4c5f8ed7
 
 ## storage
 
-### compact-experimental
+### compact
 
 Run (when the node is not running):
 
 ```sh
-oasis-node storage compact-experimental --config /path/to/config/file
+oasis-node storage compact --config /path/to/config/file
 ```
 
-to trigger manual compaction of consensus database instances:
+to trigger manual compaction of all configured database instances:
 
 ```sh
 {"caller":"storage.go:310","level":"info","module":"cmd/storage", \
@@ -385,8 +385,8 @@ Operators should run this whenever they change pruning configuration, e.g. when
 enabling it for the first time, or later changing it to retain less data. This
 way they guarantee the node is healthy when it starts.
 
-Following successful pruning, to release disk space, they are encouraged to run
-[the compaction command](#compact-experimental).
+After the pruning operators should run [the compaction command](#compact)
+to release disk space.
 
 ### inspect
 
--- a/go/oasis-node/cmd/storage/prune.go
+++ b/go/oasis-node/cmd/storage/prune.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"math"
 
+	"github.com/cometbft/cometbft/state"
+	"github.com/cometbft/cometbft/store"
 	"github.com/spf13/cobra"
 
 	"github.com/oasisprotocol/oasis-core/go/common"
@@ -14,8 +16,6 @@ import (
 	"github.com/oasisprotocol/oasis-core/go/runtime/registry"
 	db "github.com/oasisprotocol/oasis-core/go/storage/mkvs/db/api"
 )
-
-var pruneDiskSyncInterval uint64 = 10_000
 
 func newPruneCmd() *cobra.Command {
 	cmd := &cobra.Command{
@@ -101,14 +101,33 @@ func pruneConsensusDBs(ctx context.Context, dataDir string, numKept uint64, runt
 	}
 	logger.Info("pruning of consensus node DB successful", "pruned", pruned)
 
-	if err := pruneCometDBs(dataDir, int64(retainHeight)); err != nil {
+	if err := pruneCometDBs(ctx, dataDir, int64(retainHeight)); err != nil {
 		return fmt.Errorf("failed to prune CometBFT managed databases: %w", err)
 	}
 
 	return nil
 }
 
-func pruneBefore(ctx context.Context, ndb db.NodeDB, version uint64) (uint64, error) {
+type pruneOption func(*pruneOptions)
+
+type pruneOptions struct {
+	diskSyncInterval uint64
+}
+
+func withPruneDiskSyncInterval(interval uint64) pruneOption {
+	return func(opts *pruneOptions) {
+		opts.diskSyncInterval = interval
+	}
+}
+
+func pruneBefore(ctx context.Context, ndb db.NodeDB, version uint64, options ...pruneOption) (uint64, error) {
+	opts := pruneOptions{
+		diskSyncInterval: 10_000,
+	}
+	for _, option := range options {
+		option(&opts)
+	}
+
 	earliest := ndb.GetEarliestVersion()
 
 	if version <= earliest {
@@ -127,7 +146,7 @@ func pruneBefore(ctx context.Context, ndb db.NodeDB, version uint64) (uint64, er
 		}
 		pruned++
 
-		if pruneDiskSyncInterval != 0 && h%pruneDiskSyncInterval == 0 { // periodically sync to disk
+		if opts.diskSyncInterval != 0 && h%opts.diskSyncInterval == 0 { // periodically sync to disk
 			if err := ndb.Sync(); err != nil {
 				return 0, fmt.Errorf("failed to sync NodeDB: %w", err)
 			}
@@ -177,34 +196,12 @@ func minReindexedHeight(dataDir string, runtimes []common.Namespace) (int64, err
 	return minH, nil
 }
 
-func pruneCometDBs(dataDir string, retainHeight int64) error {
+func pruneCometDBs(ctx context.Context, dataDir string, retainHeight int64) error {
 	blockstore, err := openConsensusBlockstore(dataDir)
 	if err != nil {
 		return fmt.Errorf("failed to open consensus blockstore: %w", err)
 	}
 	defer blockstore.Close()
-
-	// Mimic the upstream pruning logic from CometBFT
-	// (see https://github.com/oasisprotocol/cometbft/blob/653c9a0c95ac0f91a0c8c11efb9aa21c98407af6/state/execution.go#L655):
-	// 1. Get the base from the blockstore
-	// 2. Prune blockstore
-	// 3. Prune statestore
-	//
-	// This ordering is problematic: if the blockstore pruning succeeds (updating the base) but
-	// state DB pruning fails or is interrupted, a subsequent pruning run will skip already
-	// pruned blocks while leaving part of the state DB unpruned.
-	base := blockstore.Base()
-	if retainHeight <= base {
-		logger.Info("blockstore and state db already pruned")
-		return nil
-	}
-
-	logger.Info("pruning consensus blockstore", "base", base, "retain_height", retainHeight)
-	n, err := blockstore.PruneBlocks(retainHeight)
-	if err != nil {
-		return fmt.Errorf("failed to prune blocks (retain height: %d): %w", retainHeight, err)
-	}
-	logger.Info("blockstore pruning finished", "pruned", n)
 
 	state, err := openConsensusStatestore(dataDir)
 	if err != nil {
@@ -212,11 +209,57 @@ func pruneCometDBs(dataDir string, retainHeight int64) error {
 	}
 	defer state.Close()
 
-	logger.Info("pruning consensus states", "base", base, "retain_height", retainHeight)
-	if err := state.PruneStates(base, retainHeight); err != nil {
-		return fmt.Errorf("failed to prune state db (start: %d, end: %d): %w", base, retainHeight, err)
+	logger.Info("pruning block and state store", "retain_height", retainHeight)
+
+	const cometPruneBatchSize int64 = 10_000
+	var totalPruned uint64
+	for pruneUntil := blockstore.Base(); pruneUntil < retainHeight; {
+		if err = ctx.Err(); err != nil {
+			return err
+		}
+
+		pruneUntil = min(pruneUntil+cometPruneBatchSize, retainHeight)
+
+		pruned, err := pruneBlocks(blockstore, state, pruneUntil)
+		if err != nil {
+			return fmt.Errorf("failed to prune block and state store: %w", err)
+		}
+		totalPruned += pruned
 	}
-	logger.Info("state db pruning finished")
+
+	logger.Info("successfully pruned block and state store", "pruned", totalPruned)
 
 	return nil
+}
+
+// pruneBlocks mimics the upstream pruning logic from CometBFT
+// (see https://github.com/oasisprotocol/cometbft/blob/653c9a0c95ac0f91a0c8c11efb9aa21c98407af6/state/execution.go#L655):
+// 1. Get the base from the blockstore
+// 2. Prune blockstore
+// 3. Prune statestore
+//
+// This ordering is problematic: if the blockstore pruning succeeds (updating the base) but
+// state DB pruning fails or is interrupted, a subsequent pruning run will skip already
+// pruned blocks while leaving part of the state DB unpruned.
+//
+// Best way to mitigate the size of stale state (if it happens, very unlikely) is to prune in small batches.
+func pruneBlocks(blockstore *store.BlockStore, statestore state.Store, retainHeight int64) (uint64, error) {
+	base := blockstore.Base()
+	if retainHeight <= base {
+		logger.Info("blockstore and state db already pruned")
+		return 0, nil
+	}
+
+	logger.Debug("pruning consensus blockstore", "base", base, "retain_height", retainHeight)
+	pruned, err := blockstore.PruneBlocks(retainHeight)
+	if err != nil {
+		return 0, fmt.Errorf("failed to prune blocks (retain height: %d): %w", retainHeight, err)
+	}
+
+	logger.Debug("pruning consensus states", "base", base, "retain_height", retainHeight)
+	if err := statestore.PruneStates(base, retainHeight); err != nil {
+		return 0, fmt.Errorf("failed to prune state db (start: %d, end: %d): %w", base, retainHeight, err)
+	}
+
+	return pruned, nil
 }
--- a/go/oasis-node/cmd/storage/prune.go
+++ b/go/oasis-node/cmd/storage/prune.go
@@ -14,48 +14,51 @@ import (
 	db "github.com/oasisprotocol/oasis-core/go/storage/mkvs/db/api"
 )
 
-var pruneCmd = &cobra.Command{
-	Use:   "prune-experimental",
-	Args:  cobra.NoArgs,
-	Short: "EXPERIMENTAL: trigger pruning for all consensus databases",
-	RunE:  doPrune,
-}
+func newPruneCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "prune-experimental",
+		Args:  cobra.NoArgs,
+		Short: "EXPERIMENTAL: trigger pruning for all consensus databases",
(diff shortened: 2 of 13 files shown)


go/oasis-node/cmd/storage/prune.go
```diff
@@ -0,0 +1,326 @@
...
+
+	if latest < numKept {
+		logger.Info("skipping runtime pruning as the latest version is smaller than the number of versions to keep")
+		return nil
+	}
+
+	// By calculating retain round from the runtime state DB latest round,
+	// we ensure light history is never pruned past the latest synced runtime
+	// round.
+	retainRound := latest - numKept
```
martintomazic, line 316: In practice we should probably also respect executor prune handler that should prevent pruning past last normal round.

```go
// runtimeLastNormalRound returns the last normal round for the given runtime.
func runtimeLastNormalRound(ctx context.Context, ndb db.NodeDB, runtimeID common.Namespace) (uint64, error) {
	latest, ok := ndb.GetLatestVersion()
	if !ok {
		return 0, fmt.Errorf("consensus node DB is empty")
	}

	roots, err := ndb.GetRootsForVersion(latest)
	if err != nil {
		return 0, fmt.Errorf("failed to get roots for consensus version %d: %w", latest, err)
	}
	if len(roots) == 0 {
		return 0, fmt.Errorf("no roots found for consensus version %d", latest)
	}

	tree := mkvs.NewWithRoot(nil, ndb, roots[0], mkvs.WithoutWriteLog())
	defer tree.Close()

	s := roothashState.NewImmutableState(tree)
	rtState, err := s.RuntimeState(ctx, runtimeID)
	if err != nil {
		return 0, fmt.Errorf("failed to get runtime state: %w", err)
	}

	return rtState.LastNormalRound, nil
}

```

We have two options:
1. Pass consensus node DB to `pruneRuntimeDBs`.
    * Or just open and close it there (simplest and least changes).
    *  This feels weird as this function should probably accept `retainRound`, like `pruneConsensusDBs` should accept `retainHeight` instead of `runtimeIds` (solution **2.**). 
2. Add new `consensusRetainHeight` and `runtimeRetainRound` functions that precompute this limits, possibly write unit tests for those two functions. The annoying part is that those two functions consume runtime histories, nodedbs and consensus nodedbs, meaning we loose resource encapsulation and need to keep them open throughout the whole command. Feels like a better direction, but requires a thorough refactor + makes logging and orchestration incredibly messy.

In addition adding this handler requires consensus state to be always present. Without it you can prune runtime state without having a consensus state locally (e.g. when hacking state if imported using snapshots, don't think this is needed though).

go/oasis-node/cmd/storage/prune.go
```diff
@@ -0,0 +1,326 @@
...
+	logger.Info("pruning runtime history", "retain_round", retainRound)
+	history, err := openRuntimeLightHistory(dataDir, runtimeID)
+	if err != nil {
+		return fmt.Errorf("failed to open runtime light history: %w", err)
+	}
+	defer history.Close()
+	pruned, err := pruneRuntimeHistory(context.TODO(), history, retainRound)
+	if err != nil {
+		return fmt.Errorf("failed to prune runtime history before round %d: %w", retainRound, err)
+	}
```
martintomazic, line 329: The reason I haven't wired the context yet was that many long commands, e.g. badgerDB flatten, `blockstore.PruneBlocks` and `state.PruneStates` do not expose context cancellation api.

As a workaround, we could loop blockstore and statestore pruning in batch like we currently do for the runtime history, where the goal is to prevent too big badger db transaction. This would be also a nice workaround for the upstream issues described under `pruneCometDBs`.

Notice we would still have to connect cobra context to `	signal.Notify(sigCh, os.Interrupt, syscall.SIGTERM)
` if we wanted the graceful shutdown during this command.

Currently, shutdown non-graceful, which might leave corrupted state (didn't happen yet to me). Pruning runs cca 15 min and compaction approximately the same for the latest snapshot.  Obviously function of the db sizes, but generally does not exceed 1-2h with 2 years of the state (mainly compaction takes time).

go/oasis-node/cmd/storage/prune.go
```diff
@@ -0,0 +1,326 @@
...
+	// consensus height, so that light history can be populated correctly.
+	minReindexed, err := minReindexedHeight(dataDir, runtimes)
+	if err != nil {
+		return fmt.Errorf("failed to fetch earliest reindexed consensus height: %w", err)
+	}
+
+	retainHeight := min(
+		latest-numKept, // underflow not possible due to if above.
+		uint64(minReindexed),
+	)
```
martintomazic, line 121: E.g. this could be `func consensusRetainHeight(ndb db.NodeDB, histories []history.Histories)(uint64, bool, error)` and this function takes `retainHeight: uint64` instead of the last two params, which would also allow us to unit test the business logic. As stated [above](https://github.com/oasisprotocol/oasis-core/pull/6521#discussion_r3201554302) this complicates the orchestration a lot though :(

go/oasis-node/cmd/storage/prune.go
```diff
@@ -0,0 +1,356 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			runtimes, err := registry.GetConfiguredRuntimeIDs()
+			if err != nil {
+				return fmt.Errorf("failed to get configured runtimes: %w", err)
+			}
+
+			logger.Info("Starting databases pruning. This may take a while...")
+
+			dataDir := cmdCommon.DataDir()
+			ctx := cmd.Context()
```
martintomazic, line 52: NIT: We should probably wire

```go
 signal.Notify(sigCh, os.Interrupt, syscall.SIGTERM) .
```

As cobra context does not capture this signals by default, resulting in non-graceful shutdown (never had an issue so fat though). Still it is annoying that BadgerDB command does not expose cancelation api, so even with this fixed it won't be complete.

martintomazic:
Ready for a review. 

Please check the PR context, the main question is how much do we want to complicate our life with prune handlers - [comment](https://github.com/oasisprotocol/oasis-core/pull/6521#discussion_r3201554302).

Since the PR is already rather big, I am open to push one PR infront that does:
*  factors prune/compact commands to separate files,
*  adds missing tests
*  minor fixes
*  prune block and state store in small intervals.

Then this PR becomes much smaller and focused only one new runtime logic and releasing the commands.

docs/oasis-node/cli.md
```diff
@@ -334,15 +334,15 @@ oasis1qqncl383h8458mr9cytatygctzwsx02n4c5f8ed7
 
 ## storage
 
-### compact-experimental
+### compact
 
 Run (when the node is not running):
```
matevz, line 339: Can it automatically check whether the node is not running before compacting the storage? Same for the prune below.

docs/oasis-node/cli.md
```diff
@@ -386,7 +386,7 @@ enabling it for the first time, or later changing it to retain less data. This
 way they guarantee the node is healthy when it starts.
 
 Following successful pruning, to release disk space, they are encouraged to run
```
matevz: So many words :)

```suggestion
After the pruning operators should run [the compaction command](#compact) to release disk space.
```

docs/oasis-node/cli.md
martintomazic: Already the case for both commands, see `PreRunE`.

go/oasis-test-runner/oasis/controller.go
```diff
@@ -46,6 +50,52 @@ func (c *Controller) Close() {
 	c.conn.Close()
 }
 
+// WaitConsensusHeight waits until the controller observes at least specified consensus height.
+func (c *Controller) WaitConsensusHeight(ctx context.Context, height int64) error {
```
peternose:
```suggestion
func (c *Controller) WaitHeight(ctx context.Context, height int64) error {
```
It is obvious that height belongs to consensus and round to runtime. I would also shorten names in the offline pruning scenario.

These two methods should be in the file `helpers_consensus.go`, together with other wait functions.

go/oasis-test-runner/oasis/controller.go
```diff
@@ -46,6 +50,52 @@ func (c *Controller) Close() {
...
+				return nil
+			}
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+	}
+}
+
+// WaitRuntimeRound waits until the controller observes at least specified runtime round.
+func (c *Controller) WaitRuntimeRound(ctx context.Context, runtimeID common.Namespace, round uint64) error {
```
peternose:
```suggestion
func (c *Controller) WaitRound(ctx context.Context, runtimeID common.Namespace, round uint64) error {
```

go/oasis-test-runner/scenario/e2e/runtime/offline_pruning.go
```diff
@@ -0,0 +1,209 @@
...
+	if err := sc.runOfflineCompaction(childEnv, client.Node); err != nil {
+		return err
+	}
+
+	afterPrune, err := sc.fetchStorageInspectStatus(childEnv, client.Node)
+	if err != nil {
+		return err
+	}
+
+	ensurePruned := func(want, got uint64) error {
```
peternose, line 113: Maybe this is over-engineering :)

go/oasis-node/cmd/storage/prune.go
```diff
@@ -0,0 +1,211 @@
...
+
+	"github.com/oasisprotocol/oasis-core/go/common"
+	"github.com/oasisprotocol/oasis-core/go/config"
+	cmtConfig "github.com/oasisprotocol/oasis-core/go/consensus/cometbft/config"
+	cmdCommon "github.com/oasisprotocol/oasis-core/go/oasis-node/cmd/common"
+	"github.com/oasisprotocol/oasis-core/go/runtime/registry"
+	db "github.com/oasisprotocol/oasis-core/go/storage/mkvs/db/api"
+)
+
+func newPruneCmd() *cobra.Command {
```
peternose, line 23: You should have done this in 2 steps. First commit copies, second adds `newPruneCmd`.

go/oasis-node/cmd/storage/prune_test.go
```diff
@@ -0,0 +1,120 @@
...
+	addVersion := func(version uint64) node.Root {
+		root := commitRoot(ctx, t, ndb, ns, lastRoot, version, node.RootTypeState, map[string]string{
+			"key": "value " + strconv.FormatUint(version, 10),
+		})
+		require.NoError(t, ndb.Finalize([]node.Root{root}))
+		lastRoot = root
+		return root
+	}
+
+	// Add two versions
```
peternose:
```suggestion
```

go/oasis-node/cmd/storage/prune_test.go
```diff
@@ -0,0 +1,120 @@
...
+	t.Run("prune with retain version before earliest is no-op", func(t *testing.T) {
+		pruned, err := pruneBefore(ctx, ndb, 1)
+		require.NoError(t, err)
+		require.EqualValues(t, 0, pruned)
+		require.Equal(t, uint64(2), ndb.GetEarliestVersion())
+	})
+
+	t.Run("prune with periodic disk sync", func(t *testing.T) {
+		// Force disk sync.
+		oldSyncInterval := pruneDiskSyncInterval
```
peternose: Variable `pruneDiskSyncInterval` should be a constant and you should never do ugly stuff like this. If you want this to be configurable, find some other solution (e.g. optional config).

go/oasis-node/cmd/storage/prune_test.go
```diff
@@ -0,0 +1,120 @@
...
+		pruneDiskSyncInterval = 2
+
+		defer func() {
+			pruneDiskSyncInterval = oldSyncInterval
+		}()
+
+		for version := uint64(3); version <= 20; version++ {
```
peternose: You should create 20 versions on startup, and then play with them.

go/oasis-node/cmd/storage/prune_test.go
```diff
@@ -0,0 +1,120 @@
...
+func TestPruneNodeDB(t *testing.T) {
+	ctx := context.Background()
+	ns := common.NewTestNamespaceFromSeed([]byte("storage prune test ns"), 0)
+
+	ndb, err := newTestNodeDB(t, ns)
+	require.NoError(t, err)
+	defer ndb.Close()
+
+	lastRoot := newEmptyRoot(node.RootTypeState)
+	addVersion := func(version uint64) node.Root {
```
peternose, line 30: Strange that this function is inlined, while `commitRoot` is not.

go/oasis-node/cmd/storage/prune_test.go
```diff
@@ -0,0 +1,120 @@
...
+	t.Helper()
+
+	return pathbadger.New(&dbAPI.Config{
+		DB:        filepath.Join(t.TempDir()),
+		Namespace: ns,
+		NoFsync:   true,
+	})
+}
+
+func commitRoot(
```
peternose: Should we rename this helper functions or inline them, so that developers know that these are test methods and should not be used.

go/oasis-node/cmd/storage/prune.go
```diff
@@ -102,33 +108,38 @@ func pruneConsensusDBs(dataDir string, numKept uint64, runtimes []common.Namespa
 	return nil
 }
 
-func pruneConsensusNodeDB(ndb db.NodeDB, retainHeight uint64) error {
-	startHeight := ndb.GetEarliestVersion()
+func pruneBefore(ctx context.Context, ndb db.NodeDB, retainVersion uint64) (uint64, error) {
```
peternose: This renaming is not the best.

go/runtime/history/db.go
```diff
@@ -243,6 +243,40 @@ func (d *DB) getEarliestBlock() (*roothash.AnnotatedBlock, error) {
 	return &blk, nil
 }
 
+func (d *DB) pruneBefore(roundLimit uint64) (uint64, error) {
```
peternose: I don't understand the signature: `prune before round limit`. Can we rename the parameter to `round`?

go/oasis-node/cmd/storage/prune.go
```diff
@@ -13,17 +14,22 @@ import (
...
+	roothash "github.com/oasisprotocol/oasis-core/go/roothash/api"
+	"github.com/oasisprotocol/oasis-core/go/runtime/history"
 	"github.com/oasisprotocol/oasis-core/go/runtime/registry"
 	db "github.com/oasisprotocol/oasis-core/go/storage/mkvs/db/api"
 )
 
-var pruneDiskSyncInterval uint64 = 10_000
+var (
+	pruneDiskSyncInterval uint64 = 10_000
+	pruneBatchSize        uint64 = 10_000
```
peternose: Same here.

go/oasis-node/cmd/storage/prune.go
martintomazic: renamed `retainVersion` to `version` like was done for the simlar runtime history prune before method.

go/oasis-node/cmd/storage/prune_test.go
martintomazic: One is closure, the other is generic function that may be reused later. Inlining would also make this function hard to read. I have renamed `commitRoot` to `commitTestRoot`.

go/oasis-node/cmd/storage/prune_test.go
martintomazic: Refactored to the functional optional style to avoid this.

The idea came from the book D&K _The go programming language_ where they show exactly this scenario. I was aware this is ugly as we introduce global state and cannot run tests in parallel, but wanted to keep it simple, as the current table tests are anyways not pure functions.

Now that I think the functional optional parameters were cheap to impl (was afraid is over engineering at the start), so happy with the refactor.

go/oasis-test-runner/oasis/controller.go
martintomazic: Done.

>These two methods should be in the file helpers_consensus.go, together with other wait functions.

Probably not since this is method on the controller pointer receiver? Unless I pass channel of blocks to this function. That said  `WaitRound` shouldn't go into `helpers_consensus.go`. 

Personally I would keep it as is (with shorter names as you suggested)

peternose: Then rename `helpers_consensus.go`. Functions that wait for height, round, block, epoch, ... should be in one place, so I would reorganize code so that is true. Adding these functions to the `Controller` feels like spam and I think they belong somewhere else. I moved them under helpers, but could be probably organized better, maybe in a new struct.

martintomazic: >Functions that wait for height, round, block, epoch, ... should be in one place,

One works on the `Scenario` abstraction (uses default client controller internally) the other works directly on the controller primitive.

> Adding these functions to the Controller feels like spam and I think they belong somewhere else. 

I feel this is the right abstraction, e.g. we also have `IsReady` and `WaitSync` defined on it. Ideally we should also move `scenario.WaitBlocks` and `scenario.WaitEpochs` methods there, and make existing scenario methods delegate the call to them. Eventually if needed, they might become part of the `control.api.NodeController`....

`*_helpers.go` is the biggest smell right now, we could probably go with `controller_wait.go` and `scenario_wait.go`, which I would avoid until number of methods grow.

_I feel fixing all this is out of scope and would keep as is._ :)

go/oasis-node/cmd/storage/prune.go
martintomazic: We can also hardcode sain default e.g. min few days retention, and leave a TODO issue.

go/oasis-test-runner/oasis/controller.go
peternose: File `controller_wait.go ` sounds nice. You can change the code so that it uses the controller directly instead of the `Scenario` abstraction. Should not be hard.

I think this is not out of scope, because (as already mentioned) you are now adding a code similar to the one in helper functions to a completely different location. So this is not an existing problem that should be addressed in another PR, but rather a new problem that this PR introduced.

peternose: I would still not add these wait functions directly to the controller.

martintomazic: I am happy to do the refactor here and especially like `controller_wait.go` idea.

Looking at your latest two comments I am not sure do you want me to add it or not to controller. If not what is alternative?

My stance is it fits on the controller, given we already have existing `Wait*` methods defined on it,  (see embedded `control.NodeController`). I would avoid adding them there for now and prefer non-embedded method for now not to increase API surface.

Also I am fine having similar methods on the scenario as well, that should ideally delegate to controller, as this can be seen as simple wrapper/syntax shortener, but we should probably improve comments.

Given this PR significantly improved existing code, both in terms of functionality, style and testing (yes my original experimental commands were of low quality) I don't see this as blocking.

peternose: I still think these methods do not belong to the controller as they are built on top of consensus services. So maybe we need to create a struct that takes consensus service like `core` and adds extra methods like `WaitBlock` to it. So maybe this doesn't belong to `controller_wait.go`. Since we are using this only in tests, I would still keep this code here and not in the same place where services are defined.

Already defined methods `Wait*` are not the same as we are defining here. Those methods wait for node to do something, not to wait for something in the consensus.

This PR does not improve the existing code. Similar functionality that is already implemented in the helper files is added to the controller defined in another package.

If you want, you can keep these methods in the controller. But then you need to move functions from the helper files to the controller.

martintomazic: > Already defined methods Wait* are not the same as we are defining here. Those methods wait for node to do something, not to wait for something in the consensus.

```go
// WaitNodesRegistered implements control.DebugController.
func (n *Node) WaitNodesRegistered(ctx context.Context, count int) error {
	registry := n.Consensus.Registry()
```
?

>This PR does not improve the existing code. Similar functionality that is already implemented in the helper files is added to the controller defined in another package.

As explained above, these methods do not operate on the same concept. One is on the level of scenario (i.e. something happened in the consensus), the second is that concrete node observed something locally (useful for synchronization in e2e tests).

I introduced this functions in hope we would use them more, as currently we have many flaky tests, or tests with time assumptions (very bad). Therefore I think helper like this would encourage others to write better tests. As more methods are added, this can be refactored to something even better, but I would avoid striving for perfection from the start.

I would prefer keeping it as is, if not I will copy this two helpers into my e2e test, and let others introduce generic helpers one day (I don't want to scope creep this PR further).

peternose (approved):

Issue #6430, opened 2025-12-15
oasis-node: Independent consensus light history and pruning


martintomazic:
## Context
It has become impractical to keep all state from the genesis thus more and more nodes are pruning it.

## Problem
1. Pruning is already causing availability issues:
    * E.g. https://github.com/oasisprotocol/internal-ops/issues/1281.
    * Previous issues with emty validator set because of the missing light blocks.
3. Consensus light blocks are computed ad-hoc, from the full state:   
    * Running observer with historical state requires all consensus data, just to fetch light blocks - [see](https://github.com/oasisprotocol/docs/pull/1526#discussion_r2512094786).

## Solution:

1. Create additional DB for consensus light blocks and query them from there.
2. Independent full-state (NodeDB && blockstore) vs light state pruning.
    * E.g. keep 1 year of light history by default (should be few `GB` max). 

## Follow-ups:
1. Similar thing should be done for the runtime state and its pruning (https://github.com/oasisprotocol/oasis-core/issues/6400).
2. [History expiry](https://github.com/oasisprotocol/oasis-core/issues/6422).

martintomazic:
## Proposed implementation (consensus)

Keep (abusing) [ad-hoc production]( [see](https://github.com/oasisprotocol/oasis-core/blob/5c2b576da712b5d998bdf21b749423a95e8075e8/go/consensus/cometbft/full/common.go#L605)) of light blocks and only write light headers to the new light history before pruning corresponding height.

Pros:
* _Simpler_ as we don't have to implement consensus light history (re)-indexer.
* Easier and safer to synchronize data at the tail of the chain than at the tip.

Cons:
* Internally, to serve light header, two databases need to be queried.
* Having all light headers pre-computed in the separate DB improves the latency for this call and reduces the load on consensus DBs.

### Follow-up:

By default every node fetches consensus light blocks in reverse until certain target  (independent process, not blocking initialization).

This way we can guarantee solid data availability of the whole network. 

Similar could be done for the runtime light history (https://github.com/oasisprotocol/oasis-core/issues/6305).

### Additional consideration:
Light history size should be negligible compared to other consensus DBs, therefore we may not prune it at all. Possibly, we could fetch it in reverse up-to genesis, although I find 1 year a sane default.

Issue #6388, opened 2025-11-04
oasis-node/status: Correctly show light client status


martintomazic:
## What

 `oasis-node control status` should display the[ _local light client_](https://github.com/oasisprotocol/oasis-core/blob/f5a234b5bc24083fe70d6f57110bbbd19a727540/go/consensus/cometbft/light/client.go#L65) status.

This client is created for the:
1. stateless client node  - [see](https://github.com/oasisprotocol/oasis-core/blob/f5a234b5bc24083fe70d6f57110bbbd19a727540/go/consensus/cometbft/cometbft.go#L112)
2. full node's state sync  [see](https://github.com/oasisprotocol/oasis-core/blob/f5a234b5bc24083fe70d6f57110bbbd19a727540/go/consensus/cometbft/full/full.go#L741) 

## Motivation

This would be useful to monitor stateless client (e.g. was missing this during #6385), or possibly consensus state sync.

## Existing status
Current light client status is confusing / incorrect:
```bash
  "light_client": {
    "latest_height": 0,
    "latest_hash": "0000000000000000000000000000000000000000000000000000000000000000",
    "latest_time": "0001-01-01T00:00:00Z",
    "oldest_height": 0,
    "oldest_hash": "0000000000000000000000000000000000000000000000000000000000000000",
    "oldest_time": "0001-01-01T00:00:00Z",
    "peer_ids": [
      "12D3KooWHjqvFidQgDcw24A88KX1oaPjvqiLq6X8nPi4zMnSCDu6",
      "12D3KooWKTvsowgyv3WRTVm2dvEBtxpKsshWKtKJmGriGDPdzRBM"
    ]
  },
```
Concretely this is not even light client, but an actual [wrapper over the remote providers](https://github.com/oasisprotocol/oasis-core/blob/f5a234b5bc24083fe70d6f57110bbbd19a727540/go/consensus/cometbft/light/service.go#L60:L63).

Moreover, the linked documentation is a bit misleading as trusted blocks are not obtained from the local consensus backend?

## Solution:
Two statuses:
1. One for the the actual _local light client_ (if applicable depending on the mode/config).
2. One for the consensus light service (`consensusAPI.LightService`), where we should probably get rid of anything but the remote providers peer ids. 
    * Albeit I am confused why not use full node consensus backend here over remote providers?

Any preference for nesting and naming?

Btw where is the light client serve part? I don't see any like for other protocols so I assume CometBFT full nodes serve this by default?

PR #6526, opened 2026-05-12
go/consensus/cometbft/light: Improve fetching of missing light blocks

martintomazic:
## Problem

```bash
{"err":"builder: StringTracer: I/O error: rpc error: server error: module: unknown code: 1 message: light block fetch failure: call failed on all peers","level":"error","module":"runtime/consensus/cometbft/verifier","msg":"Consensus verifier failed to initialize, retrying","retry":1,"ts":"2026-05-08T01:38:21.870417058Z"}
{"err":"builder: StringTracer: I/O error: rpc error: server error: module: unknown code: 1 message: light block fetch failure: call failed on all peers","level":"error","module":"runtime/consensus/cometbft/verifier","msg":"Consensus verifier failed to initialize, retrying","retry":2,"ts":"2026-05-08T01:38:23.152344571Z"}
{"err":"builder: StringTracer: I/O error: rpc error: server error: module: unknown code: 1 message: light block fetch failure: call failed on all peers","level":"error","module":"runtime/consensus/cometbft/verifier","msg":"Consensus verifier failed to initialize, retrying","retry":3,"ts":"2026-05-08T01:38:24.263100939Z"}
{"level":"error","module":"runtime/consensus/cometbft/verifier","msg":"Failed to start consensus verifier, aborting","ts":"2026-05-08T01:38:25.297222923Z"}
```

Both RONLs and ROFL run consensus verifier inside the enclave, starting at the embedded trust root. In case the light blocks are not locally available,  e.g. ROFL node might have just synced from the recent checkpoint, or is pruning its state, they are fetched from the remote peers ([see](https://github.com/oasisprotocol/oasis-core/blob/2c4a47b51637a6b35631bf64e8cf016c8a558318/go/runtime/registry/handler.go#L267:L283)). Most of the remote peers prune their consensus state, so fetching older blocks (e.g. more than 2 weeks) is in practice not reliable.

Currently, during initialization the enclave will try to initialize consensus verifier 3 times, and then abort ([see](https://github.com/oasisprotocol/oasis-core/blob/2c4a47b51637a6b35631bf64e8cf016c8a558318/runtime/src/consensus/tendermint/verifier/mod.rs#L601:L625)). For every retry in practice we only contact `numWitnesses = 2` (variable abuse fixed in this PR), meaning remote fetch is not really good at finding peers with actual light blocks.

Whilst indeed this only happens during ronl/rofl initialization, rofl app developers might still force storage wipe/redeploy, or spin additional instance much later on. Generally, we don't want to force them to update embedded root (changes enclave measurement), just to have light block availability (either local or remote).

## Solution 

This PR focuses on improving the process of finding suitable remote peers and fetching light blocks from them.

There is orthogonal problem of fixing overall data availability of light blocks (https://github.com/oasisprotocol/oasis-core/issues/6430), possibly fetching them up to target height on every node in reverse (light history expiry), that then also serves as as the oldest possible embedded trust root ronl/rofl developers are expected to maintain.

## How to test?

I tested locally using dummy sgx rofl with old embedded trust root, that only logged runtime light blocks. No issues after the refactor, hard to quantify the improvement on the testnet, where old light block availability seems good.

### DIFF
--- a/go/consensus/cometbft/light/service.go
+++ b/go/consensus/cometbft/light/service.go
@@ -10,11 +10,8 @@ import (
 )
 
 const (
-	// numWitnesses is the number of libp2p backed CometBFT light-block witnesses to be instantiated.
-	numWitnesses = 2
-	// lcMaxRetryAttempts is the number of retry attempts the CometBFT light client does,
-	// before switching the primary provider.
-	lcMaxRetryAttempts = 5
+	// numProviders is the number of libp2p backed CometBFT light-block providers used by the light client service.
+	numProviders = 2
 )
 
 type ClientService struct {
@@ -69,8 +66,8 @@ func (c *ClientService) Validators(ctx context.Context, height int64) (*consensu
 // New creates a new CometBFT light client service backed by remote P2P peers.
 func New(ctx context.Context, chainContext string, p2p rpc.P2P) (*ClientService, error) {
 	pool := NewProviderPool(ctx, chainContext, p2p)
-	providers := make([]*Provider, 0, numWitnesses)
-	for range numWitnesses {
+	providers := make([]*Provider, 0, numProviders)
+	for range numProviders {
 		p := pool.NewProvider()
 		providers = append(providers, p)
 	}
--- a/go/consensus/cometbft/light/service.go
+++ b/go/consensus/cometbft/light/service.go
@@ -23,7 +23,7 @@ type ClientService struct {
 	providers []*Provider
 }
 
-// GetStatus implements api.LightService.
+// GetStatus implements [consensus.LightService].
 func (c *ClientService) GetStatus() (*consensus.LightClientStatus, error) {
 	status := &consensus.LightClientStatus{}
 
@@ -36,7 +36,7 @@ func (c *ClientService) GetStatus() (*consensus.LightClientStatus, error) {
 	return status, nil
 }
 
-// LightBlock implements the LightProvider interface.
+// LightBlock implements the [consensus.LightProvider] interface.
 func (c *ClientService) LightBlock(ctx context.Context, height int64) (*consensus.LightBlock, error) {
 	rsp, _, err := tryProviders(ctx, c.providers, func(p *Provider) (*lightBlock, rpc.PeerFeedback, error) {
 		return p.getLightBlock(ctx, height)
@@ -51,7 +51,7 @@ func (c *ClientService) LightBlock(ctx context.Context, height int64) (*consensu
 	return rsp.lb, nil
 }
 
-// Validators implements the LightProvider interface.
+// Validators implements the [consensus.LightProvider] interface.
 func (c *ClientService) Validators(ctx context.Context, height int64) (*consensus.Validators, error) {
 	validators, _, err := tryProviders(ctx, c.providers, func(p *Provider) (*consensus.Validators, rpc.PeerFeedback, error) {
 		return p.getValidators(ctx, height)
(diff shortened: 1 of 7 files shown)


go/consensus/cometbft/light/service.go
```diff
@@ -72,13 +58,15 @@ func (c *ClientService) Validators(ctx context.Context, height int64) (*consensu
...
 
-// New creates a new CometBFT light client service backed by the local full node.
-//
-// This light client is initialized with a trusted blocks obtained from the local consensus backend.
+// New creates a new CometBFT light client service backed by remote P2P peers.
 func New(ctx context.Context, chainContext string, p2p rpc.P2P) (*ClientService, error) {
+	// The service should be able to fetch light blocks with arbitrary age.
+	// Since most of the nodes are pruning, we try to maintain at least 10 providers,
+	// to increase the likelihood of success.
+	const numProviders = 10
```
martintomazic, line 66: Missing timeout, possibly in parallel. Why not use `CallOne`, or `CallMulti` that goes over all best peers either sequential or in parallel. 

In fact we should just refactor `service.go` to p2p `client.go` (we already have `server.go` and `protocol.go`) like is done with other runtime p2p protocols.

Then in the status we can also add peer count of the light protocol...

kostko:
Cool, this looks like a good idea.

go/consensus/api/light.go
```diff
@@ -69,22 +67,8 @@ type Evidence struct {
 	Meta []byte `json:"meta"`
 }
 
-// LightClientStatus is the current light client status overview.
+// LightClientStatus is the current light client service status overview.
 type LightClientStatus struct {
```
martintomazic, line 73: `cfg.breaking`? Notice we should probably also refactor `LightClientStatus` to `LightServiceStatus**` or even better stop treating this as something special and refactor it to normal p2p client as suggested by https://github.com/oasisprotocol/oasis-core/pull/6526/changes#r3225549584 and only track protocol peer count.

As pointed by @anzoman it would be also nice to see the actual peers connected, so that one can convince himself that setting persistent peers actually worked.


** There is also separate issue for adding the actual light client status (consensus state sync or stateless client), thus the renaming is needed. https://github.com/oasisprotocol/oasis-core/issues/6388

go/consensus/cometbft/light/service.go
kostko: I believe there is already a status for peer count of the light protocol?

For example, one of my nodes shows something like this under `p2p.protocols`:
```
"/oasis/bb3d748def55bdfb797a2ac53ee6ee141e54cd2ab2dc2375f4a0703a178e6e55/light/1.0.0": 48,
```

martintomazic: You are right. 

If this is refactored to the actual `client.go` pattern, then we probably don't need ~this~ separate p2p client service status at all, unless we reuse/refactor it for https://github.com/oasisprotocol/oasis-core/pull/6526#discussion_r3225569309? 

To accomodate @anzoman  idea we might add `--verbose` flag to the status, and then at the bottom next to protocol peer count also print the actual peers IDs. Notice also we should match the format of the config and status in such case, so I suggest to open a separate issue for that.

kostko (approved):

go/consensus/cometbft/light/service.go
```diff
@@ -72,13 +58,15 @@ func (c *ClientService) Validators(ctx context.Context, height int64) (*consensu
 	return validators, nil
 }
 
-// New creates a new CometBFT light client service backed by the local full node.
-//
-// This light client is initialized with a trusted blocks obtained from the local consensus backend.
+// New creates a new CometBFT light client service backed by remote P2P peers.
 func New(ctx context.Context, chainContext string, p2p rpc.P2P) (*ClientService, error) {
+	// The service should be able to fetch light blocks with arbitrary age.
+	// Since most of the nodes are pruning, we try to maintain at least 10 providers,
```
peternose:
```suggestion
	// Since most of the nodes are pruning, we try to maintain more providers,
```

Don't hard code values, as then you need to change them twice.

peternose (approved):

PR #6531, opened 2026-05-21
go/runtime/host: Improve common.GetQuotePolicy function

martintomazic:
Trivial, semantics should be preserved.

Tackles https://github.com/oasisprotocol/oasis-core/pull/6474#discussion_r3232716113.


go/runtime/host/tdx/qemu.go
```diff
@@ -434,18 +434,23 @@ func (p *qemuProvisioner) updateCapabilityTEE(ctx context.Context, hp *sandbox.H
 	rekPub := rspRep.RuntimeCapabilityTEERakReportResponse.RekPub
 	rawQuote := rspRep.RuntimeCapabilityTEERakReportResponse.Report
 
-	// Prepare the quote policy for local verification. In case a policy is not available or it
-	// indicates that TDX is not supported, use the fallback policy so we can provision something.
+	// Use the fallback policy for ROFL components and RONL components with no TDX policy so that provisioning can proceed.
```
martintomazic, line 447: >RONL components with no TDX policy

Or RONL policy with no deployments for given version. RONL tdx is anyways not supported (for now), still raises a question why even pass such things as runtime verifier inside the enclave would then reject anyways.

For ROFL we currently pass because we don't do pre-validation on the host side, and nil tdx policy is not valid (no default override like we have for the nil sgx).

go/runtime/host/sgx/common/common.go
```diff
@@ -8,56 +8,47 @@ import (
...
 	"github.com/oasisprotocol/oasis-core/go/runtime/host/sandbox"
 )
 
-// GetQuotePolicy fetches the quote policy for the given component. In case the policy is
-// not available, return the fallback policy.
-func GetQuotePolicy(
+// GetRONLQuotePolicy fetches the quote policy for the specified RONL deployment.
+//
+// Returns nil if the policy is not available.
+func GetRONLQuotePolicy(
```
peternose: I would rename this back to `GetQuotePolicy` as the parameter `runtimeID` already says that this is for RONL only. But what I really wanted in that comment is to get rid of the parameter `cs`, by creating a struct like `QuotePolicyProvider` (or something better) and calling `provider.Get(ctx, runtimeID, version)`.

go/runtime/host/sgx/ecdsa.go
```diff
@@ -63,13 +64,15 @@ func (ec *teeStateECDSA) Update(ctx context.Context, sp *sgxProvisioner, conn pr
...
 	}
 
-	quotePolicy, err := sgxCommon.GetQuotePolicy(ctx, ec.cfg, sp.consensus, nil)
-	if err != nil {
-		return nil, err
-	}
 	var pcsQuotePolicy *pcs.QuotePolicy
-	if quotePolicy != nil {
-		pcsQuotePolicy = quotePolicy.PCS
+	if ec.cfg.Component.Kind == component.RONL {
```
peternose: I would replace all these `if`s with `switch` for better readability, and in the `default` section add a comment `// No policy.`

go/runtime/host/sgx/common/common.go
martintomazic: > But what I really wanted in that comment is to get rid of the parameter cs, by creating a struct like QuotePolicyProvider (or something better) and calling provider.Get(ctx, runtimeID, version).

We could probably get rid of `epid.go` as epid provisioner is deprecated. Then we could also get rid of fallback check on the consensus params whether pcs is enabled which would remove the need for the consensus service altogether. This should be fine unless sometime later we want to disable pcs again and want to trigger new fallback (unlikely to happen anytime soon, we can pass this as a config instead).

kostko: Removing EPID should be fine, see https://github.com/oasisprotocol/oasis-core/issues/5935.

go/runtime/host/sgx/ecdsa.go
peternose: Was this resolved or are you skipping the comment?

martintomazic:
Things can get further simplified now:
1. `teeState` interface no longer makes sense and mock flag should instead be nested with ifs inside the `ecdsa` like is done elsewhere for the mock flag. Also existing name is confusing, `Attester` would probably be easier to read, but probably no longer needed anyways.
3. Get rid of the common package nested inside `sgx` since it is sgx agnostic and common packages are a smell. E.g. move one level up to `attestation` package:
```go
runtime/host/attestation/policy.go
  QuotePolicyProvider

runtime/host/attestation/??.go
  UpdateRuntimeQuote
  EndorseCapabilityTEE

runtime/host/attestation/worker.go
  AttestationWorker

runtime/host/attestation/metrics.go
  InitMetrics
  UpdateAttestationMetrics
```

go/runtime/host/sgx/ecdsa.go
```diff
@@ -18,15 +17,6 @@ type teeStateECDSA struct {
 }
 
 func (ec *teeStateECDSA) Init(ctx context.Context, sp *sgxProvisioner, cfg *host.Config) ([]byte, error) {
-	// Check whether the consensus layer even supports ECDSA attestations.
-	regParams, err := sp.consensus.Registry().ConsensusParameters(ctx, consensus.HeightLatest)
-	if err != nil {
-		return nil, fmt.Errorf("unable to determine registry consensus parameters: %w", err)
-	}
-	if regParams.TEEFeatures == nil || !regParams.TEEFeatures.SGX.PCS {
```
peternose, line 26: Why is this redundant?

martintomazic: Currently we only support PCS, so no need to check what is implicit. For anything else you will have an error/be rejected by the consensus.

This allows us to simplify code, else this whole PR is not much of the improvement if we still need to pass both consensus service and policy provider struct.

_If_ new provisioning will be added in the future (unlikely) this can be made "complex" again.

peternose: I think this is not fine as someone could use our code to create their own chain without PCS support. And for those chains, this check is not redundant.

martintomazic: > I think this is not fine as someone could use our code to create their own chain without PCS support.

Not really as then nothing would work for them, given that IAS is deprecated (also a problem then).

I see your point in making code extensible, but realistically we are not bulding this repo so that random chains could be spawned. I feel keeping this requirements is _over-engineering._

I think this is safe and if and when new provisioning is added, this should be reevaluated/made better. 

We could probably add comment to `regParams.TEEFeatures.SGX.PCS`,that this is deprecated and is expected to always be true?

go/runtime/host/provisioner/provisioner.go
```diff
@@ -47,8 +48,10 @@ func New(
 		return nil, err
 	}
 
+	policyProvider := sgxCommon.NewQuotePolicyProvider(consensus)
```
peternose: Since `QuotePolicyProvider` is a long word, Go might suggest names like `QuotePolicer`. Not sure if this is better, though.

go/runtime/host/sgx/ecdsa.go
peternose: OK, maybe you are right and we can remove this.

peternose (approved):
> further simplified now

I would do this in another PR, if needed.

PR #6534, opened 2026-05-22
go/worker/storage/p2p: Penalize bad checkpoint sync peers

martintomazic:


go/worker/storage/p2p/synclegacy/client.go
```diff
@@ -77,6 +78,11 @@ func (c *client) GetCheckpoints(ctx context.Context, request *GetCheckpointsRequ
 		peerCps := peerRsp.(*GetCheckpointsResponse).Checkpoints
 
 		for _, cpMeta := range peerCps {
+			if slices.Contains(peerCps, nil) {
```
kostko: Why is this inside the for loop?

go/worker/storage/p2p/checkpointsync/client.go
martintomazic, line 50: Should I export `p2p.rpc.NopClient` and `p2p.rpc.NopPeerManager`, and then via embedding only overwrite `CallMulti`, create custom struct for peer feedback, so that we can mock `client.GetCheckpoints`? and assert expected behaviour? cca 100 LOC.

go/worker/storage/p2p/checkpointsync/client.go
```diff
@@ -59,6 +60,11 @@ func (c *client) GetCheckpoints(ctx context.Context, request *GetCheckpointsRequ
 	for i, peerRsp := range rsps {
 		peerCps := peerRsp.(*GetCheckpointsResponse).Checkpoints
 
+		if slices.Contains(peerCps, nil) {
+			pfs[i].RecordBadPeer()
+			continue
+		}
```
martintomazic: NIT: after this validation `CheckpointMetadata` and `ChunkMetadata` should be passed by value not by reference, as technically relatively small immutable struct?

Go community is a bit split when you have struct with slice that points to mutable array, some prefer pointer semantics there to highlight this, but I feel majority is in favor of pass by value still.

Would be good to align on the best practice here as in the long run this will reduce likelihood of nil deref.

go/worker/storage/p2p/checkpointsync/client.go
kostko: That would be cool.

kostko (approved):

go/worker/storage/p2p/checkpointsync/client.go
peternose: Can't you just pass by value now and skip this validation?

martintomazic: You propose changing API?:
```go
// GetCheckpointsResponse is a response to a GetCheckpoints request.
type GetCheckpointsResponse struct {
	Checkpoints []checkpoint.Metadata `json:"checkpoints,omitempty"`
}
```

which should be safe as wire format would be the same as both pointer and non-pointer values are encoded the same in cbor? Notice that nil is silently decoded as zero value in such case, but peer can also sent zero value checkpoint.

In theory checkpoint sync should later penalize such checkpoints. In practice it does not happen as , [checkCheckpointUsable](https://github.com/oasisprotocol/oasis-core/blob/68559e44dc3910854249df43be6b9b81932bda2c/go/worker/storage/committee/checkpoint_sync.go#L370:L404) would mark it as `false` due to namespace mismatch so peers won't receive bad feedback (actually good one as they returned more then 0 checkpoints).

Probably `checkCheckpointUsable` should also penalize peers that:
1. Advertised wrong namespace.
2. Advertised checkpoints with roots not matching with light history roots.
3. Checkpoints with zero chunks.

Current `checkCheckpointUsable` is bad as it mixes structural validation with scheduling (finding appropriate checkpoint for the given node).

Few options I see:
1. Client does light validation e.g. **1.** and **3.**, but it needs to be constructed with runtime ID.
2. Client receives runtime light history and does full structural validation, i.e. **1-3**
    * If checkpoint is older then earliest light block we simply drop it
    * If hashes don't match this is bad peer and should not get to the checkpoint sync.
4. Inside `checkpointSync.go`:
    i.  extract `validateCheckpoints` out of `checkCheckpointUsable` and penalize there.
    ii. Inline penalization of peers if namespace or root does not match.


_Technically this is orthogonal issues that we found here, harmless in practice._

wdy?

peternose: Maybe 3. option would be the right one, but we are probably getting out of the way here. Why nodes send `nil` in the first place?

martintomazic: >Why nodes send nil in the first place?

Because this value comes from remote peers that may sent anything.

Fixed existing issues, code can be further improved by changing pointer to value semantics but this is out of scope and low priority. Like storage committe worker it would be also nice to refactor checkpoint sync to the worker that is actually tested as currently it is hard to be confident in any changes one makes. Again separate issue.

Closing.

go/worker/storage/p2p/checkpointsync/client.go
```diff
@@ -59,6 +60,11 @@ func (c *client) GetCheckpoints(ctx context.Context, request *GetCheckpointsRequ
 	for i, peerRsp := range rsps {
 		peerCps := peerRsp.(*GetCheckpointsResponse).Checkpoints
```
peternose, line 61: Is this safe?

go/worker/storage/p2p/checkpointsync/client.go
```diff
@@ -59,6 +60,11 @@ func (c *client) GetCheckpoints(ctx context.Context, request *GetCheckpointsRequ
...
 
+		if slices.ContainsFunc(peerCps, func(m *checkpoint.Metadata) bool { return m.Validate() != nil }) {
```
peternose, line 63: This doesn't feel robust. If we're fetching checkpoints from untrusted nodes, I'd expect us to use all checkpoints they provide, skipping `nil` values and penalizing the node for each `nil` encountered.

martintomazic: >This doesn't feel robust.

Why? If peer misbehaves once, I decided to ignore all his remaining input.

go/worker/storage/p2p/checkpointsync/client.go
martintomazic: What are you afraid of? (also existing code)

peternose: Can this code panic?

go/worker/storage/p2p/checkpointsync/client.go
peternose: Sounds strict to me. But since we are probably the only one providing peer implementation, this might be fine.

go/worker/storage/p2p/checkpointsync/client.go
martintomazic: No since:
```go
	var rsp GetCheckpointsResponse
	rsps, pfs, err := c.rc.CallMulti(ctx, c.mgr.GetBestPeers(), MethodGetCheckpoints, request, rsp)
	if err != nil {
		return nil, err
	}
```
and later in `CallMulti`:
```go
rsp := reflect.New(reflect.TypeOf(rspTyp)).Interface()
```

peternose (approved):

PR #6535, opened 2026-05-22
go/consensus/cometbft/api: Remove message execution from context

peternose:
Application `roothash` uses a message execution context when processing and executing runtime messages. This context was introduced solely to reject runtime escrow messages when they are forbidden by consensus. Therefore, it should not be used in other applications. However, since the `vault` application can also process and execute messages, the current naming implies that it should use the aforementioned context as well. To avoid this ambiguity, we are renaming the message execution context to roothash execution context, emphasizing that only roothash messages are executed within this context.

UPDATE: Since renaming was a bad idea, this PR introduces sender of application messages and uses it to figure out the origin of the message and then decide if it comes from roothash or not.

### DIFF
--- a/go/consensus/cometbft/abci/messages.go
+++ b/go/consensus/cometbft/abci/messages.go
@@ -25,13 +25,13 @@ func (md *messageDispatcher) Subscribe(kind any, ms api.MessageSubscriber) {
 }
 
 // Implements api.MessageDispatcher.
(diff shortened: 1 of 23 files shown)


kostko (commented):
I don't think I like this as the context should be more general than an individual app (roothash)?

kostko (approved):

PR #6551, opened 2026-07-09
go/consensus/cometbft/config: Move allow duplicate IP config

### LINKED ISSUE #6550: Make consensus p2p allow duplicate IP option non-debug
Currently the only way to allow multiple peers from the same IP for the consensus P2P configuration is to use `consensus.debug.allow_duplicate_ip` option which is gated behind the unsafe debug flags. This configuration option is actually useful in some local setups so it should be moved to the regular `consensus.p2p.allow_duplicate_ip` and not gated on the unsafe debug flags.

martintomazic:
Closes #6550

### DIFF
--- a/go/consensus/cometbft/config/config.go
+++ b/go/consensus/cometbft/config/config.go
@@ -90,6 +90,15 @@ type P2PConfig struct {
 	DisablePeerExchange bool `yaml:"disable_peer_exchange"`
 	// CometBFT max timeout when redialing a persistent peer (default: unlimited).
 	PersistentPeersMaxDialPeriod time.Duration `yaml:"persistent_peers_max_dial_period"`
+
+	// Allow multiple P2P connections from the same IP.
+	//
+	// WARNING: Enabling this disables duplicate-IP protection, allowing a single
+	// remote host to occupy all inbound peer slots, and is unsafe for public-facing
+	// nodes. Only use this in isolated/private topologies where P2P ingress is
+	// restricted to trusted peers, e.g. local test networks or validators behind
+	// sentries with firewall-restricted access.
+	AllowDuplicateIP bool `yaml:"allow_duplicate_ip"`
 }
 
 // SubmissionConfig is the transaction submission configuration.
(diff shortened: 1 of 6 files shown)


go/consensus/cometbft/config/config.go
```diff
@@ -90,6 +90,9 @@ type P2PConfig struct {
 	DisablePeerExchange bool `yaml:"disable_peer_exchange"`
 	// CometBFT max timeout when redialing a persistent peer (default: unlimited).
 	PersistentPeersMaxDialPeriod time.Duration `yaml:"persistent_peers_max_dial_period"`
+
+	// Allow multiple P2P connections from the same IP.
+	AllowDuplicateIP bool `yaml:"allow_duplicate_ip"`
 }
```
martintomazic, line 102: Maybe we should add extra warning here this is only safe to use in the sentry like setups, e.g. persistent peers + limit peer count setup. So that normal operators don't accidentally enable it.

The defaults intentionally set this to false and this is how it should stay for the public facing nodes.

martintomazic: >e.g. persistent peers + limit peer count setup

I guess blocking all traffic except from the local network is the only safe way in such setup cc @kostko ?

kostko: Hm, I guess you are right. My point was using this with multiple nodes on localhost. But the problem is this also allows external peers with the same.

Maybe just using unconditional peers would work? Would need to check if unconditional peers bypass this requirement.

kostko: I checked and this is not the case so this PR is needed.

kostko: And yes, let's add the warning.

docs/development-setup/deploying-a-runtime.md
```diff
@@ -334,7 +334,7 @@ oasis-node \
   --runtime.supported $RUNTIME_ID \
   --runtime.paths $RUNTIME_ID=$RUNTIME_BINARY \
   --consensus.cometbft.debug.addr_book_lenient \
-  --consensus.cometbft.debug.allow_duplicate_ip \
```
kostko, line 337: This is obsolete and likely doesn't work anymore as you can only pass these options via the configuration file.

kostko (approved):

PR #6562, opened 2026-08-06
go/runtime/txpool: Limit batch size

peternose:
We should limit the batch size to prevent RHP `CheckTx` calls from exceeding the message size limit. Perhaps we should also limit the size of the pool in bytes.

### DIFF
--- a/go/runtime/txpool/check_queue.go
+++ b/go/runtime/txpool/check_queue.go
@@ -7,6 +7,11 @@ import (
 	"github.com/gammazero/deque"
 )
 
+const (
+	// maxBatchBytes is the maximum size of a batch in bytes.
+	maxBatchBytes = 31 * 1024 * 1024 // 31MiB
+)
+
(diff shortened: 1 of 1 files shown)


kostko (approved):

kostko:
> Perhaps we should also limit the size of the pool in bytes.

This sounds reasonable, but should probably be a configuration option with a sane default. Note that `CheckTx` inside the runtime will already do transaction size checks based on the runtime limits.

peternose:
I will create a separate PR for pool limit.

Yes, the runtime will do the check also, but we still need to limit the transaction size as otherwise too big transactions never reach the runtime and are retried indefinitely.

kostko:
Yeah I meant for the pool size limit, since the pool is currently limited based on the number of transactions and the runtime enforces transaction size. But having a node-side upper hard limit makes sense IMO.

PR #6566, opened 2026-08-10
go/storage/mkvs/node: Validate key bit depth

peternose:
Working with `Key` type can be problematic as its bit size can overflow the maximum value when `[]byte` is converted to `Key`.


kostko (approved):

kostko:
Would be great to also re-run the fuzz tests for MKVS.

PR #6568, opened 2026-08-11
go: Clean up before key manager access policy

martintomazic:
**Trivial**.

First 3 commits were extracted from the https://github.com/oasisprotocol/oasis-core/pull/6474.

Since the https://github.com/oasisprotocol/oasis-core/pull/6474/changes/a2becf169a268ddeff934ac9fea6e60bf5493615 was scope creeping keymanager access policy I decided to do part of it here and simplify:
* Per component provisioner was overkill. Your node cannot be keymanager and compute at the same time, so you have one per-node provisioner, and policy provider.  This enables us to stop passing `willRegisterComputeRuntime` through layers once we add kma policy.
    * If/when this changes, e.g. we start passing ROFL policies for the local pre-validation, this can be implemented again.

### DIFF
--- a/go/common/node/node.go
+++ b/go/common/node/node.go
@@ -14,7 +14,6 @@ import (
 	beacon "github.com/oasisprotocol/oasis-core/go/beacon/api"
 	"github.com/oasisprotocol/oasis-core/go/common"
 	"github.com/oasisprotocol/oasis-core/go/common/cbor"
-	"github.com/oasisprotocol/oasis-core/go/common/crypto/hash"
 	"github.com/oasisprotocol/oasis-core/go/common/crypto/signature"
 	"github.com/oasisprotocol/oasis-core/go/common/prettyprint"
 	"github.com/oasisprotocol/oasis-core/go/common/version"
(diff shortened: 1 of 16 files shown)


go/runtime/registry/policy_provider_test.go
```diff
@@ -0,0 +1,75 @@
...
+	consensus.Service
+
+	registry registry.Backend
+}
+
+func (m *mockConsensus) Registry() registry.Backend {
+	return m.registry
+}
+
+func TestConsensusQuotePolicyProvider(t *testing.T) {
```
martintomazic, line 44: NIT: Existing issue. The test raises a question whether in case of missing version we should warn or error instead of silently defaulting to nil. Not critical as this is local pre-validation only (sanity check on the go side).

go/common/node/node.go
```diff
@@ -565,38 +585,38 @@ type CapabilityTEE struct {
 	Attestation []byte `json:"attestation"`
 }
 
-// HashRAK computes the expected report data hash bound to a given public RAK.
-func HashRAK(rak signature.PublicKey) hash.Hash {
```
peternose, line 569: You should also move `teeHashContext`.

peternose (approved):

go/common/node/node.go
martintomazic: Done,  I made it function local constant, as it has single caller and was package private.

PR #6571, opened 2026-08-16
go/runtime/registry: Isolate ROFL EnclaveRPC session budgets

### LINKED ISSUE #6563: go/runtime: ROFL components share one EnclaveRPC session budget potentially evicting each other
## Problem
ROFL to RONL EnclaveRPC path doesn't set PeerID ([see](https://github.com/oasisprotocol/oasis-core/blob/b7fce3cb10734806f088696f649a9eff67897116/go/runtime/registry/handler_rofl.go#L145)), so every ROFL component on a node is accounted under the same empty peer and they all compete for `RPC_MAX_SESSIONS_PER_PEER = 8`. With more than 8 components (e.g. a marketplace provider serving many ROFL apps), they start force-closing each other's sessions, each eviction costing a fresh attested handshake.

## Solution
Trivial. Give peer IDs an explicit namespacing scheme, e.g. `p2p:<peer-address>` for network peers, `local:<component-id>` for local inter-enclave calls.

martintomazic:
Closes #6563


go/runtime/registry/handler_rofl_test.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (rt *capturingRuntime) Call(_ context.Context, request *protocol.Body) (*protocol.Body, error) {
+	rt.request = request
+	return &protocol.Body{
+		RuntimeRPCCallResponse: &protocol.RuntimeRPCCallResponse{},
+	}, nil
+}
+
+func TestROFLHostHandlerEnclaveRPCPeerID(t *testing.T) {
```
martintomazic, line 29: Not sure this test adds much value...

kostko (commented):
Should we make these prefixes constants and define them in one place?

martintomazic:
>Should we make these prefixes constants and define them in one place?

Made them public. Initial reasoning was not to include them in the API as they are opaque anyways.

kostko (approved):

PR #6585, opened 2026-09-23
go/runtime/txpool: Fix republish debounce

peternose:
Also bumping yanked rust dependencies and improving runtime scenario to wait for the transaction pool to be ready. This will make sure that transactions with higher nonces are in the main queue waiting for transactions with lower ones. If we don't wait for the transaction pool to be ready, all transactions might end up in the check queue, which would not test the main queue at all.


go/runtime/txpool/txpool.go
```diff
@@ -816,10 +827,13 @@ func (t *txPool) republishWorker() {
...
 		ticker.Reset(nextPendingRepublish)
 
-		t.logger.Debug("republished transactions",
+		t.logger.Debug(
+			"republished transactions",
 			"num_txs", republishedCount,
 			"next_republish", nextPendingRepublish,
 		)
+
+		lastRepublish = time.Now()
```
kostko, line 836: So the issue was that republish requests were not debounced at all?

kostko (approved):

go/runtime/txpool/txpool.go
peternose: There was no issue in production at all. AI found this bug, so we are fixing it. But yes, these requests were never debounced at all.

